PR #157, opened 2021-08-24
Fix tag reading exceptions

oischinger:
The disc and track numbers are often stored as strings containing e.g.
"1/16" which leads to exceptions.
Also taglib sometimes returns empty lists which also produces an
exception.

This fix avoids some common exceptions.


marcelveldt:
Thanks!

PR #341, opened 2022-05-25
Support for album having multiple artists

marcelveldt:
- Enable support that an album can have multiple artists
- Including backwards compatibility
- Including database migration
- No rescan of existing media, a simple "refresh" on an album will reveal the additional artist(s)


marcelveldt:
fixes https://github.com/music-assistant/hass-music-assistant/issues/220
fixes https://github.com/music-assistant/hass-music-assistant/issues/241

PR #449, opened 2022-08-07
Fix: Ignore diacritic characters when matching on name

marcelveldt:
loosen up the matching on name a small bit by ignoring diacritic characters. as additional safety guard the total difference of characters may not be more than 2


marcelveldt:
fixes https://github.com/music-assistant/hass-music-assistant/issues/811

DutchJaFO:
I think it's not working for me.
image


Upper/lower case matches are finally working as I noticed that it managed to group both tracks in the same album despite different casing for the album title "'t beste van 2 werelden" as evidenced here :
image

PR #454, opened 2022-08-09
Fix Youtube Music support for some countries

quanghd96:
Some countries do not currently support youtube music. I tried using YTM_DOMAIN which is https://www.youtube.com, it works fine

### DIFF
--- a/music_assistant/music_providers/ytmusic/ytmusic.py
+++ b/music_assistant/music_providers/ytmusic/ytmusic.py
@@ -50,6 +50,7 @@ from music_assistant.music_providers.ytmusic.helpers import (
     search,
 )
 
+YT_DOMAIN = "https://www.youtube.com"
 YTM_DOMAIN = "https://music.youtube.com"
 YTM_BASE_URL = f"{YTM_DOMAIN}/youtubei/v1/"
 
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Can you elaborate on this? The function you adjusted is just for grabbing the timestamps to retrieve valid stream urls. Information about Artist, Albums, Playlists, Tracks etc would still come from https://music.youtube.com, so I doubt that will work if YT Music is not supported in your country.

quanghd96:
I checked that the youtube music api and youtube api seem to be the same, you can change the domain music.youtube.com to www.youtube.com and the results will be the same.
For example: 
+ https://www.youtube.com/youtubei/v1/browse
+ https://music.youtube.com/youtubei/v1/browse
image
image

MarvinSchenkel:
I can imagine they would be similar for 'music videos' that are coming from youtube, but what about actual music coming from https://music.youtube.com? Did you test this with for example with albums coming from YT Music (not youtube)?

quanghd96:
But here I just use youtube domain instead of youtube music to get base.js file. All other functions use youtube music domain. I tested it and it works fine

image

marcelveldt (approved):
I don't know why this works but the Youtube API has suprised us more often.

PR #455, opened 2022-08-12
Improve performance of YT Music search

MarvinSchenkel:
This PR removes additional search API calls and enhances the parse-functions to deal with the sad search results coming from YT Music.


marcelveldt (approved):
Thanks, not sure if there are side effects but we'll see about that later

PR #569, opened 2023-03-24
"fix" Soundcloud not starting

gieljnssns:


marcelveldt:
Sorry, I fixed this already but did not yet checkin. I wil now release a new version with the patch included. I did still not test it as the auth header is missing for me but looks like I need to recreate my account

PR #579, opened 2023-03-27
Refactor image handling a bit

marcelveldt:
Refactors the logic to resolve images that are not (or may be not) remotely accessible. Instead of limiting that functionality to the file providers, this is now useable for other providers too (e.g. Plex)

### DIFF
--- a/music_assistant/server/controllers/metadata.py
+++ b/music_assistant/server/controllers/metadata.py
@@ -261,8 +261,7 @@ class MetaDataController:
         self,
         media_item: MediaItemType,
         img_type: ImageType = ImageType.THUMB,
-        resolve_local: bool = True,
-        allow_local: bool = True,
+        resolve: bool = True,
     ) -> str | None:
         """Get url to image for given media media_item."""
         if not media_item:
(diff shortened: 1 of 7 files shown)


marcelveldt:
I'll go ahead and merge this and adjust the frontend so you can give it a test and we address any issues you may find then.

music_assistant/server/controllers/metadata.py
```diff
@@ -273,35 +272,35 @@ async def get_image_url_for_item(
...
+                if img.provider != "url" and not resolve:
                     continue
-                if img.source != "http" and resolve_local:
-                    # return imageproxy url for local filesystem items
+                if img.provider != "url" and resolve:
+                    # return imageproxy url for images that need to be resolved
                     # the original path is double encoded
                     encoded_url = urllib.parse.quote(urllib.parse.quote(img.url))
                     return f"{self.mass.webserver.base_url}/imageproxy?path={encoded_url}"
                 return img.url
```
micha91, line 282: I think this should be img.path, right?

marcelveldt: oops, yes and thats why we invented code reviews :-)

marcelveldt: fixed it on main

PR #586, opened 2023-03-28
First approach of a Plex integration

micha91:
We start with a small feature set. A Plex provider is always bound to a user account and a library. If a user has multiple libraries, the user has to add them as separate providers. I am using an async version of the lru_cache to cache artist and album information, because there is only a reference delivered by the API whenever you request a track. Maybe it would be better to use the MA integrated cache for that instead - I am absolutely open to discuss that 👍 

In the next steps, I want to add a flag to the config to let the user decide, whether a library contains audio books or music and I will add the chapter meta data for audio books.


music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+
+class PlexProvider(MusicProvider):
+    """Provider for a plex music library."""
+
+    _plex_server: PlexServer = None
+    _plex_library: PlexMusicSection = None
+
+    async def handle_setup(self) -> None:
+        """Set up the music provider by connecting to the server."""
+        if not self.config.get_value(CONF_AUTH_TOKEN):
```
marcelveldt, line 82: maybe move this to the module level setup ?

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+            raise LoginFailed("Invalid login credentials")
+
+        def connect():
+            plex_account = MyPlexAccount(token=self.config.get_value(CONF_AUTH_TOKEN))
+            return plex_account.resource(self.config.get_value(CONF_SERVER_NAME)).connect()
+
+        self._plex_server = await self._run_async(connect)
```
marcelveldt, line 89: Is there some un-initialization needed when closing reloading the provider ?
In that case also implement a unload (async) method to do any cleanup

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+            ProviderFeature.ARTIST_ALBUMS,
+        )
+
+    async def _run_async(self, call: Callable, *args, **kwargs):
+        return await self.mass.create_task(call, *args, **kwargs)
+
+    async def _get_data(self, key, cls=None):
+        return await self._run_async(self._plex_library.fetchItem, key, cls)
+
+    @alru_cache(maxsize=128, ttl=300)
```
marcelveldt, line 117: I'm not so sure about this. We should either switch all in-memory caching to lru_cache or keep with the music assistant cache, otherwise its hard to keep track of memory consumption. 

I'm leaning towards using the music assistant cache as that can also store longer TTL in a db but I'd like to hear your vision.

marcelveldt: for the record, the MA cache also has a decorator, just import use_cache from the cache module

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+    async def _search_album_advanced(self, limit, **kwargs) -> list[PlexAlbum]:
+        return await self._run_async(self._plex_library.searchAlbums, filters=kwargs, limit=limit)
+
+    async def _search_artist_advanced(self, limit, **kwargs) -> list[PlexArtist]:
+        return await self._run_async(self._plex_library.searchArtists, filters=kwargs, limit=limit)
+
+    async def _search_playlist_advanced(self, limit, **kwargs) -> list[PlexPlaylist]:
+        return await self._run_async(self._plex_library.playlists, filters=kwargs, limit=limit)
+
+    async def _search_and_parse(self, search_coro: Coroutine, parse_coro: Callable):
```
marcelveldt, line 165: missing return type annotation

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        return await self._run_async(self._plex_library.fetchItem, key, cls)
+
+    @alru_cache(maxsize=128, ttl=300)
+    async def _get_data_cached(self, key, ma_cls):
+        if ma_cls == Artist:
+            return await self._parse_artist(await self._get_data(key, PlexArtist))
+        if ma_cls == Album:
+            return await self._parse_album(await self._get_data(key, PlexAlbum))
+
+    async def _parse(self, plex_media):
```
marcelveldt, line 124: missing return type annotation

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+
+    async def _search_artist(self, search_query, limit) -> list[PlexArtist]:
+        return await self._run_async(
+            self._plex_library.searchArtists, title=search_query, limit=limit
+        )
+
+    async def _search_playlist(self, search_query, limit) -> list[PlexPlaylist]:
+        return await self._run_async(self._plex_library.playlists, title=search_query, limit=limit)
+
+    async def _search_track_advanced(self, limit, **kwargs) -> list[PlexTrack]:
```
marcelveldt, line 153: whats the advanced for ?

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        return await self._run_async(self._plex_library.searchAlbums, filters=kwargs, limit=limit)
+
+    async def _search_artist_advanced(self, limit, **kwargs) -> list[PlexArtist]:
+        return await self._run_async(self._plex_library.searchArtists, filters=kwargs, limit=limit)
+
+    async def _search_playlist_advanced(self, limit, **kwargs) -> list[PlexPlaylist]:
+        return await self._run_async(self._plex_library.playlists, filters=kwargs, limit=limit)
+
+    async def _search_and_parse(self, search_coro: Coroutine, parse_coro: Callable):
+        tasks = []
```
marcelveldt, line 166: can you rename this to task_results ?
I was confused about the tasks and was first reading it wrong

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+            results.append(task.result())
+
+        return results
+
+    async def _parse_album(self, plex_album: PlexAlbum) -> Album:
+        """Parse a Plex Album response to an Album model object."""
+        album_id = plex_album.key
+        album = Album(
+            item_id=album_id,
+            name=plex_album.title,
```
marcelveldt, line 182: can you swap the name and provider so item_id and provider are always the first entries ?

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+            provider=self.domain,
+        )
+        if plex_album.year:
+            album.year = plex_album.year
+        if thumb := plex_album.firstAttr("thumb", "parentThumb", "grandparentThumb"):
+            album.metadata.images = [MediaItemImage(ImageType.THUMB, thumb, self.instance_id)]
+        if plex_album.summary:
+            album.metadata.description = plex_album.summary
+
+        album.artist = await self._get_data_cached(plex_album.parentKey, Artist)
```
marcelveldt, line 192: please do not do any caching in your module, this is all handled in the outer controllers itself. You should just return the most recent data when requested. Or did you mean to implement a very short cache just to return recurring items such as an album artist in the listing ? How much data do you have here from an artist within album level ? Because we do support returning a simple ItemMapping object here to just return a reference to another nested object.

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+            )
+        )
+        return playlist
+
+    async def _parse_track(self, plex_track: PlexTrack) -> Track:
+        """Parse a Plex Track response to a Track model object."""
+        track = Track(item_id=plex_track.key, provider=self.domain, name=plex_track.title)
+
+        if plex_track.grandparentKey:
+            track.artist = await self._get_data_cached(plex_track.grandparentKey, Artist)
```
marcelveldt, line 249: same comment as for the album - you may just return an ItemMapping here if you have at least some data

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        media_type = ContentType.try_parse(media.container)
+
+        stream_details = StreamDetails(
+            item_id=plex_track.key,
+            provider=self.domain,
+            content_type=ContentType.try_parse(media.container),
+            duration=plex_track.duration,
+            channels=media.audioChannels,
+            data=plex_track,
+            loudness=audio_stream.loudness,
```
marcelveldt, line 437: scrap if this is not an EBU-R128 measurement, normally this is measured in the stream controller.
Only when you are certain you can return the EBU-R128 average loudness as float, you may set it here yourself, otherwise leave it out and it will be auto filled.

music_assistant/server/providers/plex/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        else:
+            url = plex_track.getStreamURL()
+            media_info = await parse_tags(url)
+
+            stream_details.channels = media_info.channels
+            stream_details.content_type = ContentType.try_parse(media_info.format)
+            stream_details.sample_rate = media_info.sample_rate
+            stream_details.bit_depth = media_info.bits_per_sample
+
+        return stream_details
```
marcelveldt, line 456: if you have just an url to stream, you'd better set it as "direct" attribute of the streamdetails so url is fed directly into ffmpeg, saves you form having to chunk the audio yourself with the get_audio_stream

music_assistant/server/providers/plex/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "music",
+  "domain": "plex",
+  "name": "Plex Media Server Library",
+  "description": "Support for the Plex streaming provider in Music Assistant.",
+  "codeowners": ["@micha91"],
+  "requirements": ["plexapi==4.13.2", "async-lru==2.0.2"],
```
marcelveldt, line 7: as discussed, let's have a look if we should switch to async_lru completely or adjust your implementation to MA cache, I'd rather not have 2 cache implementations. That said,  maybe you can even skip the cache completely if you could use ItemMapping objects ?

marcelveldt (changes requested):
Overall looking good!
I have left you some small comments

music_assistant/server/providers/plex/__init__.py
micha91: I do have the key of the artist here. The key is unique on a Plex server and I am also always using the key as id when I create a MA Artist object. But I don't have access to the name of the artist or so here. Would that work or would we have to get the artist from the DB first to generate a mapping?

music_assistant/server/providers/plex/__init__.py
micha91: I did not find any recommendations regarding un-initialization. As we don't provide a requests.Session object, the library creates its own session and should implicitly delete it

music_assistant/server/providers/plex/__init__.py
micha91: Plex offers a really cool and complex query language, where you can build full decision trees. But for now these search methods are all not used. Does it make sense to keep them because MA supports that kind of search, too or is it more like we have a search string an a media_type?

music_assistant/server/providers/plex/__init__.py
micha91: If it is not an m4a, I set the direct attribute. If it is an m4a, we use the Plex transcoding and in that case the seek position has to be included in the URL, which I build in the get_audio_stream method. And there I have to return the binary data like a do it right now, haven't I? Is there a better way to solve that?

music_assistant/server/providers/plex/__init__.py
micha91: It is, but I just figured out that the analysis seems to done for plex pass owners only, so I will remove it

music_assistant/server/providers/plex/__init__.py
marcelveldt: Cool, yes MA supports these kind of queries too (for all the sorting and stuff in the frontend) but this is done on MA's own database as abstraction layer because there is no way we can find common ground between all the different providers.

But yeah, keep it in here because maybe we find a future use ;-)

music_assistant/server/providers/plex/__init__.py
marcelveldt: Oh sorry, in that case I misread the code and its all good now.

music_assistant/server/providers/plex/__init__.py
micha91: I will do some research, but at least I did not find anything directly in the libraries documentation

micha91: Oh damned - in German we say, sometimes you don't see the forest because of all the trees. There is a parentTitle in addition to the parentKey and its the same for all the stuff in tracks. What shall I set for sort_name, uri and version? I just found examples in the code where the mapping was created from an existing MediaItem

music_assistant/server/providers/plex/__init__.py
marcelveldt: ah thats a bummer. But if it is EBU-R128 we could set it if exists ?

marcelveldt: At least I assume it has a value when the user has a plex pass and no value if otherwise

music_assistant/server/providers/plex/__init__.py
marcelveldt: Cool! That saves a lot of performance. 
uri is auto generated if you leave it blank from the provider and item_id

Copy the __post_init__ from MediaItem to ItemMapping and you are all set

marcelveldt: https://github.com/music-assistant/server/blob/main/music_assistant/common/models/media_items.py#L176

marcelveldt: The version you can leave out if you don't have it, that's just for textual reasons. When a full object is retrieved, it fetches all info and only then is any matching performed.

music_assistant/server/providers/plex/__init__.py
micha91: removed in favor of ItemMapping

music_assistant/server/providers/plex/__init__.py
micha91: Okay, I think I have got it :-)
Ps. Do I have to set my domain or my instance_id as provider? I am just a bit confused because I just realized that the key is only on a server unique - So if someone configures multiple servers (and not just multiple libraries on the same server, like I do), I can't guarantee that the combination of the domain "plex" and the key is unique. It would just be unique in the plex instance

music_assistant/server/providers/plex/__init__.py
micha91: Okay, I added it if present 👍

music_assistant/server/providers/plex/__init__.py
marcelveldt: In your case set the instance_id. It accepts both but setting the domain is more suitable for multiple instances of e.g. spotify where an item exists on all accounts. For filebased providers and plex, set it to the instance id.

marcelveldt:
Let's merge so we have a base to iterate over. Improvements can be made later and on base of user feedback/testing.
Great!!

PR #592, opened 2023-03-28
Make base url settable by providing MASS_BASE_URL

bennnnnnnn:
Hi Marcel,

Tiny change to be able to use MASS_BASE_URL to set the Music assistant base url, similar to MA1.

Kind regards,
Ben

### DIFF
--- a/music_assistant/server/controllers/webserver.py
+++ b/music_assistant/server/controllers/webserver.py
@@ -35,8 +35,8 @@ class WebserverController:
 
     @property
     def base_url(self) -> str:
-        """Return the (web)server's base url."""
-        return f"http://{self.mass.base_ip}:{self.port}"
+        """Return the (web)server's base url, use the one from env var if set"""
+        return os.getenv("MASS_BASE_URL", f"http://{self.mass.base_ip}:{self.port}")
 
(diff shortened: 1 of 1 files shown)


marcelveldt:
Sorry, that is not going to work this time. devices must be able to reach the MA server on IP address, no https, no dns name. The default 8095 port is also used for all kind of callbacks.

What is your usecase ? Reaching the web server ? Because I'm going to provide an option to start a second interface of the webserver for other purposes but first I need to finish the HA auth on the websocket.

marcelveldt:
Can't you use a reverse proxy btw ? Or maybe I should include a base config in the container ?

bennnnnnnn:
My use case is that I'm running a k3s cluster. Ideally I'd like to expose the webserver frontend through traefik (with ssl offloading) and expose the streaming webserver directly through a service IP, I'd like to specify that service IP as the base url for the streaming clients.

marcelveldt:
As discussed on discord we're going to implement it a bit more flexible. I'll pick it up in between somewhere next days.


https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=24191084

PR #594, opened 2023-03-29
Fix remove playlist tracks

jozefKruszynski:
Fix GenericAlias error when trying to remove playlist tracks


marcelveldt (changes requested):
I think you can better change line 96 

if origin in (tuple, list) so you catch both in one

music_assistant/server/helpers/api.py
```diff
@@ -93,7 +93,7 @@ def parse_value(name: str, value: Any, value_type: Any, default: Any = MISSING)
     if value is None and value_type is NoneType:
         return None
     origin = get_origin(value_type)
-    if origin is list:
+    if origin in (tuple, list):
         return [
```
marcelveldt, line 97:
```suggestion
        return origin(
```

marcelveldt: and ) on line 101

music_assistant/server/helpers/api.py
```diff
@@ -93,12 +93,12 @@ def parse_value(name: str, value: Any, value_type: Any, default: Any = MISSING)
     if value is None and value_type is NoneType:
         return None
     origin = get_origin(value_type)
-    if origin is list:
-        return [
+    if origin in (tuple, list):
+        return (
```
marcelveldt, line 97:
```suggestion
        return origin(
```
this should make a tuple become a tuple again and a list a list

marcelveldt (approved):

PR #597, opened 2023-03-29
Prefer instance id over domain

marcelveldt:
Preparation of https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=24283728

This makes sure that we do lookups all on provider instance first and only fallback to domain.
Allows setting the provider attribute on the media item to the provider instance.


micha91 (commented):
Looks good, but I would recommend to review the doc string for the is_unique flag and maybe it could also be a property, couldn't it?

marcelveldt:
> We have one more error, you wanted to fix in here, right?

Yeah, I'm working on it, found some other rabbit holes

PR #602, opened 2023-04-01
Soundcloud guard

gieljnssns:
Add a guard to get_library_playlists 
Also added time debug as in spotify


music_assistant/server/providers/soundcloud/__init__.py
```diff
@@ -146,10 +161,17 @@ async def get_library_artists(self) -> AsyncGenerator[Artist, None]:
...
+        )
         for item in playlists["collection"]:
-            playlist_id = item["playlist"]["id"]
-            playlist_obj = await self._soundcloud.get_playlist_details(playlist_id=playlist_id)
+            if "playlist" in item and "id" in item["playlist"]:
+                playlist_obj = await self._soundcloud.get_playlist_details(
+                    playlist_id=item["playlist"]["id"]
+                )
             try:
                 yield await self._parse_playlist(playlist_obj)
```
marcelveldt, line 176:
```suggestion
            try:
                playlist_obj = await self._soundcloud.get_playlist_details(
                    playlist_id=item["playlist"]["id"]
                )
                yield await self._parse_playlist(playlist_obj)
```

marcelveldt: either put the whole block into the try..except clause or add a guard upfront

PR #617, opened 2023-04-07
Add Plex Track Artist for compilations

micha91:
Even if the track's artist is different from the album's artist, we were only using the album's artist by now. This changes with this PR. Now we will be use the track's artist for tracks if it is different from the album's artist. As plex does not create artists in their database, if they don't have any albums, we have to create fake/MA-database-only artists on our side.


marcelveldt (approved):
creative solution but very good for UX !

PR #626, opened 2023-04-14
Add Tidal Music provider

jozefKruszynski:
I think everything is working as it should. Although that is a bold statement


music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+class TidalProvider(MusicProvider):
+    """Implementation of a Tidal MusicProvider."""
+
+    async def handle_setup(self) -> None:
+        """Handle async initialization of the provider."""
+        self._cache_dir = CACHE_DIR
+        self._tidal_session = TIDAL_SESSION
+        self._tidal_user_id = self.config.get_value(CONF_USER_ID)
+        # try to get a token, raise if that fails
+        self._cache_dir = os.path.join(CACHE_DIR, self.instance_id)
```
marcelveldt, line 152: is the CACHE_DIR some leftover old code ?

music_assistant/server/providers/tidal/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "music",
+  "domain": "tidal",
+  "name": "Tidal",
+  "description": "Support for the Tidal streaming provider in Music Assistant. Hit the button below to authenticate with Tidal, and check that the redirect is not blocked by a Pop-up Blocker!",
```
marcelveldt, line 5:
```suggestion
  "description": "Support for the Tidal streaming provider in Music Assistant.",
```

music_assistant/server/providers/tidal/manifest.json
```diff
@@ -0,0 +1,10 @@
...
+  "codeowners": ["@jozefKruszynski"],
+  "requirements": ["git+https://github.com/jozefKruszynski/python-tidal.git"],
```
marcelveldt, line 7: is temporary unversioned ?

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: Yes it is indeed

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,390 @@
...
+            action=CONF_ACTION_AUTH,
+            action_label="Authenticate on Tidal.com",
+            value=values.get(CONF_AUTH_TOKEN) if values else None,
+        ),
+        ConfigEntry(
+            key=CONF_REFRESH_TOKEN,
+            type=ConfigEntryType.SECURE_STRING,
+            label="Refresh token for Tidal",
+            description="You need to link Music Assistant to your Tidal account.",
+            action_label="Authenticate on Tidal.com",
```
marcelveldt, line 121:
```suggestion
            hidden=True,
```

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,390 @@
...
+            value=values.get(CONF_REFRESH_TOKEN) if values else None,
+        ),
+        ConfigEntry(
+            key=CONF_EXPIRY_TIME,
+            type=ConfigEntryType.STRING,
+            label="Expiry time of auth token for Tidal",
+            action_label="Authenticate on Tidal.com",
```
marcelveldt, line 128:
```suggestion
            hidden=True,
```

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,390 @@
...
+            value=values.get(CONF_EXPIRY_TIME) if values else None,
+        ),
+        ConfigEntry(
+            key=CONF_USER_ID,
+            type=ConfigEntryType.STRING,
+            label="Your Tidal User ID",
+            description="This is your unique Tidal user ID.",
+            action_label="Authenticate on Tidal.com",
```
marcelveldt, line 136:
```suggestion
            hidden=True,
```

music_assistant/server/providers/tidal/helpers.py
```diff
@@ -0,0 +1,493 @@
...
+            image_url,
+        )
+    ]
+    metadata.copyright = album_obj.copyright
+    metadata.explicit = album_obj.explicit
+    metadata.popularity = album_obj.popularity
+    return metadata
+
+
+async def parse_track(tidal_provider, track_obj: TidalTrack) -> Track:
```
marcelveldt, line 401: you can drop the async methods if all you do is set some attributes

music_assistant/server/providers/tidal/helpers.py
```diff
@@ -0,0 +1,493 @@
...
+        artist = await parse_artist(tidal_provider=tidal_provider, artist_obj=track_artist)
+        track.artists.append(artist)
+    available = track_obj.available
+    track.add_provider_mapping(
+        ProviderMapping(
+            item_id=track_id,
+            provider_domain=tidal_provider.domain,
+            provider_instance=tidal_provider.instance_id,
+            content_type=ContentType.FLAC,
+            bit_rate=1411,
```
marcelveldt, line 432: is the contenttype flac by default ? and is the sample rate and bit depth limited to redbook (CD quality) 44.1khz/16 bits ? Tidal also has some hi res material right ? Or is that limited to MQA ?

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: I tested this and if I set it as hidden the auth button also becomes hidden.
The other entries I will hide though.

music_assistant/server/providers/tidal/manifest.json
jozefKruszynski: Honestly, I'm not sure what you mean here... Yes is my temporary answer to a "hopefully" temporary issue

music_assistant/server/providers/tidal/helpers.py
jozefKruszynski: The higher quality stuff I thought I would add later. The plan is to give the option during setup of choosing quality. Now assuming that our token and the user account have access to that quality and the track eexists also in that quality... I would set that at parsing time

music_assistant/server/providers/tidal/manifest.json
marcelveldt: Yeah, well, by putting only the github version here its no longer pinned to a specific version but guess that is not a big deal for now. You could also pin it so a specific branch, tag or even commit if I'm correct but as you are in control of the repo yourself it shouldn't matter that much.

music_assistant/server/providers/tidal/__init__.py
marcelveldt: Yeah, only hide the entries you (ab)use for the refresh token etc. and you do not want to have visible in the UI

music_assistant/server/providers/tidal/manifest.json
jozefKruszynski: Got ya, I was trying to create a tagged version last night, or perhaps this morning, but I had been out having "fun" for a bit too long to be capable of any code related activities

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,393 @@
...
+class TidalProvider(MusicProvider):
+    """Implementation of a Tidal MusicProvider."""
+
+    _tidal_session_manager: TidalSessionManager | None = None
+
+    async def handle_setup(self) -> None:
+        """Handle async initialization of the provider."""
+        async with TidalSessionManager(
+            self.instance_id, self.config, self.mass
+        ) as tidal_session_manager:
```
marcelveldt, line 144: this is a bit weird to use a contextmanager. I was under the impression that you needed to validate the session each call. In that case you can perhaps store the session as attribute on your class attribute and use the contextmanager  on every call that needs the session. Within the context manager you can then perform a quick check if the token needs to be refreshed, cleanup resources etc.

There is no actual need for a contextmanager but its just comfortable. You can even turn a method within your class into a contextmanager: https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager

so you can store/cache the session as instance attribute.
and then in every function that needs access to the session you do something like this:

session = self.get_tidal_session()
and then do something with that session
or the contextmanager approach:

async with self.get_tidal_session() as session:
   session.do_something()

and in the "get_tidal_session" you do all the checks needed if the session is still valid, auth token fresh etc.

You can also create a decorator and decorate all functions.

or the lazy approach is to just call whatever on the todil session and catch the exception that the token expired, refresh it then and then retry.

So enough approaches ;-)

jozefKruszynski: To be honest, I think I was going round in circles today.
I already started changing this whole context manager approach as it also seemed completely strange to me.
Should I convert this PR to a draft for the time being?

music_assistant/server/providers/tidal/manifest.json
jozefKruszynski: Tagged version now added

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: (Ab)used entries hidden

music_assistant/server/providers/tidal/helpers.py
jozefKruszynski: I was giving this some thought, and I had a couple of ideas for this. However, I'll need to upgrade my account to implement them.
However, the HI RES recordings could be somewhat interesting, in so much as the bit rate, mode (stereo, Atmos, etc.) and codec) on the HI RES versions varies and may not be immediately available when parsing the track.
Bitrate varies between 2304-9216 Kbps
Not sure about codecs used yet.
Mode can be MQA, Reality360 or Atmos.

I think I can only get the codec at streaming time, although one could assume that the codec will be directly related to the audio mode.
Bit rate is not included in any response per se, however, I assume that MA will calculate the bitrate during streaming? Do I assume correctly here and If so, should I just skip setting it in the provider mapping altogether?

marcelveldt: bitrate is not important, bit depth and sample rate is, that should be set correctly, as well as the codec. The more accurate quality details you provide, the better. For the streamdetails these details NEED to be accurate, for just the regular track item it MAY be accurate.

jozefKruszynski: Sadly again, Tidal are rather vague about the bit depth and sample rate.
The say typically HI RES is 96KHz/24bit, but that leaves room for interpretation.
I'll dig into web console and see if I can find the info anywhere in some obscure call that isn't mapped in the tidalapi.

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        async for track in self.get_playlist_tracks(prov_playlist_id):
+            if track.position in positions_to_remove:
+                prov_track_ids.append(track.item_id)
+            if len(prov_track_ids) == len(positions_to_remove):
+                break
+        return await add_remove_playlist_tracks(
+            self._tidal_session, prov_playlist_id, prov_track_ids, add=False
+        )
+
+    async def create_playlist(self, name: str) -> Playlist:  # type: ignore[return]
```
marcelveldt, line 317: why the type ignore here ?I think that can be omitted

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        track = await get_track(self._tidal_session, item_id)
+        url = await get_track_url(self._tidal_session, item_id)
+        if not track:
+            raise MediaNotFoundError(f"track {item_id} not found")
+        # make sure that the token is still valid by just requesting it
+        await validate_token_and_refresh(self._tidal_session, self.mass, self.config)
+        return StreamDetails(
+            item_id=track.id,
+            provider=self.instance_id,
+            content_type=ContentType.FLAC,
```
marcelveldt, line 335: for a follow up PR: make sure that the sample rate and bit depth is correct here

music_assistant/server/providers/tidal/helpers.py
marcelveldt: Let's do that in a follow-up PR

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        self, search_query: str, media_types=list[MediaType] | None, limit: int = 5
+    ) -> SearchResults:
+        """Perform search on musicprovider.
+
+        :param search_query: Search query.
+        :param media_types: A list of media_types to include. All types if None.
+        :param limit: Number of items to return in the search (per type).
+        """
+        search_query = search_query.replace("'", "")
+        results = await search(self._tidal_session, search_query, media_types, limit)
```
marcelveldt, line 186: replace the "self._tidal_session" with your helper method that checks the token and returns the session

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+            for playlist in results["playlists"]:
+                parsed_results.playlists.append(await parse_playlist(playlist))
+        if results["tracks"]:
+            for track in results["tracks"]:
+                parsed_results.tracks.append(await parse_track(track))
+        return parsed_results
+
+    async def get_library_artists(self) -> AsyncGenerator[Artist, None]:
+        """Retrieve all library artists from Tidal."""
+        artists_obj = await get_library_artists(self._tidal_session, self._tidal_user_id)
```
marcelveldt, line 204: replace the "self._tidal_session" with your helper method that checks the token and returns the session

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        for artist in artists_obj:
+            yield parse_artist(tidal_provider=self, artist_obj=artist)
+
+    async def get_library_albums(self) -> AsyncGenerator[Album, None]:
+        """Retrieve all library albums from Tidal."""
+        albums_obj = await get_library_albums(self._tidal_session, self._tidal_user_id)
```
marcelveldt, line 210: replace the "self._tidal_session" with your helper method that checks the token and returns the session

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        for album in albums_obj:
+            yield parse_album(tidal_provider=self, album_obj=album)
+
+    async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
+        """Retrieve library tracks from Tidal."""
+        tracks_obj = await get_library_tracks(self._tidal_session, self._tidal_user_id)
```
marcelveldt, line 216: replace the "self._tidal_session" with your helper method that checks the token and returns the session

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -0,0 +1,382 @@
...
+        for track in tracks_obj:
+            if track.available:
+                yield parse_track(tidal_provider=self, track_obj=track)
+
+    async def get_library_playlists(self) -> AsyncGenerator[Playlist, None]:
+        """Retrieve all library playlists from the provider."""
+        playlists_obj = await get_library_playlists(self._tidal_session, self._tidal_user_id)
```
marcelveldt, line 223: you get the point by now ;-)

music_assistant/server/providers/tidal/helpers.py
jozefKruszynski: Yup

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: Removed, and yes, remnant of copy paste from other provider

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: Will do

music_assistant/server/providers/tidal/__init__.py
jozefKruszynski: Created a decorator function and wrapped all methods

marcelveldt (approved):
Well done, very nice!

PR #637, opened 2023-04-19
Fix large Tidal listings

marcelveldt:
Instead of grabbing insane large lists of 99999 items, chunk them up and yield from the smaller chunks.

@jozefKruszynski I changed the helpers, now all is needed is adjusting the implementation detail within the musicprovider.
The first one (library artists) is already done as example.


music_assistant/server/providers/tidal/__init__.py
```diff
@@ -180,94 +224,119 @@ async def search(
...
+        while True:
+            chunk = await get_playlist_tracks(
+                tidal_session,
+                limit=DEFAULT_LIMIT,
+                offset=offset,
+                prov_playlist_id=prov_playlist_id,
+            )
+            for track_obj in chunk:
+                if track_obj.available:
+                    track = self._parse_track(track_obj=track_obj)
```
marcelveldt, line 321: this needs to run in the executor too

PR #650, opened 2023-04-25
Plex provider: Ensure that we only connect to servers

micha91:
By now we tried to connect to any resource. A resource can also be a PlexClient (e.g. a player/app etc.), so this can lead to multiple different problems and errors, as these devices do not have any libraries or something like that.

### DIFF
--- a/music_assistant/server/providers/plex/helpers.py
+++ b/music_assistant/server/providers/plex/helpers.py
@@ -27,7 +27,8 @@ async def get_libraries(mass: MusicAssistant, auth_token: str) -> set[str]:
         all_libraries: list[str] = []
         plex_account = MyPlexAccount(token=auth_token)
         for resource in plex_account.resources():
-            if "server" in resource.provides:
+            if "server" not in resource.provides:
+                continue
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/plex/helpers.py
```diff
@@ -26,17 +26,18 @@ def _get_libraries():
...
-        for server_resource in plex_account.resources():
-            try:
-                plex_server: PlexServer = server_resource.connect(None, 10)
-            except plexapi.exceptions.NotFound:
-                continue
-            for media_section in plex_server.library.sections():
-                media_section: PlexLibrarySection  # noqa: PLW2901
-                if media_section.type != PlexMusicSection.TYPE:
+        for resource in plex_account.resources():
+            if "server" in resource.provides:
```
marcelveldt, line 30: better use a guard to prevent the deeper nesting.

marcelveldt, line 30:
```suggestion
            if "server" not in resource.provides:
                continue
```

PR #672, opened 2023-05-16
Bump Deezer-python to 5.12, optimizations, quality of life changes. 

Un10ck3d:
This PR adds a few optimizations, bumps deezer-python to 5.12 and a few quality of life changes.
Optimizations include:
- Setting throttler to 50 requests per 5 seconds instead of 4 requests per 1 second. 
- Not checking if track is available that much and only when the track is getting parsed. 

Other changes:
- Setting name of the provider according to account name. This makes it easier to distinguish if multiple are added. 
- Bump deezer-python to 5.12.0 from 5.10.0
- Stop using eval to convert string to integer because its a security issue.

### DIFF
--- a/music_assistant/server/providers/deezer/__init__.py
+++ b/music_assistant/server/providers/deezer/__init__.py
@@ -342,7 +342,7 @@ class DeezerProvider(MusicProvider):
     async def add_playlist_tracks(self, prov_playlist_id: str, prov_track_ids: list[str]):
         """Add tra ck(s) to playlist."""
         await self.client.add_playlist_tracks(
-            playlist_id=prov_playlist_id, tracks=[eval(i) for i in prov_track_ids]
+            playlist_id=prov_playlist_id, tracks=[int(i) for i in prov_track_ids]
         )
 
     async def remove_playlist_tracks(
--- a/music_assistant/server/providers/deezer/__init__.py
+++ b/music_assistant/server/providers/deezer/__init__.py
@@ -119,7 +119,7 @@ async def get_config_entries(
     )
 
 
-class DeezerProvider(MusicProvider):
+class DeezerProvider(MusicProvider):  # pylint: disable=W0223
     """Deezer provider support."""
 
     client: DeezerClient
--- a/music_assistant/server/providers/deezer/__init__.py
+++ b/music_assistant/server/providers/deezer/__init__.py
@@ -129,7 +129,7 @@ class DeezerProvider(MusicProvider):
 
     async def handle_setup(self) -> None:
         """Set up the Deezer provider."""
-        self._throttler = Throttler(rate_limit=4, period=1)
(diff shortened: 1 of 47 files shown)


music_assistant/server/providers/deezer/__init__.py
```diff
@@ -143,6 +143,7 @@ async def handle_setup(self) -> None:
 
         self.gw_client = GWClient(self.mass.http_session, self.config.get_value(CONF_ACCESS_TOKEN))
         await self.gw_client.setup()
+        self.config.name = f"Deezer - {self.client.user.name}"
```
marcelveldt, line 146: this is actually not supposed to be changed this way, you are overriding the user's config.
instead override the name property:

```
@property
    def name(self) -> str:
        """Return (custom) friendly name for this provider instance."""
       if client := getattr(self, 'client', None):
         return f"Deezer - {client.user.name}"
       return super().name
```

music_assistant/server/providers/deezer/gw_client.py
```diff
@@ -150,4 +150,54 @@ async def get_deezer_track_urls(self, track_id):
         if error := result_json["data"][0].get("errors"):
             raise DeezerGWError("Received an error from API", error)
 
+        print(await self.mark_track_played(track_id=track_id))
```
marcelveldt, line 153: please either drop the non working code or finish it

Un10ck3d: Yeah I was not planning to already do a PR but then the deezer-python version came out and and I wasn't home so I can't fix it yet.

Un10ck3d:
With 50 requests per 5 seconds we can for example do 40 requests in 2 seconds and then only do 10 after that. But yes its true I don't use the throttler... I thought it would magically throttle stuff but yeah that dosnt make much sense i see.
But in general I made this pr way to early and just mashed thing together I think..

marcelveldt:
@Un10ck3d I think at least we should make sure we fix the eventloop lockups asap as I still see a lot of sync requests blocking the loop. In fact I noticed it even hurting playback.

So, if you need some more time for the other stuff in the PR, maybe break it down into smaller chunks and PR now what is ready ?

Un10ck3d:
So sorry I didn't see your message, I will do it tomorrow ASAP. 
About blocking sync code; yeah I have been trying to find the blocking come but without luck.

Maybe I should just port deezer-python to aiohttp 🤔

Un10ck3d:
Closing this in favor of my new pr's

jozefKruszynski:
> So sorry I didn't see your message, I will do it tomorrow ASAP. About blocking sync code; yeah I have been trying to find the blocking come but without luck.
> 
> Maybe I should just port deezer-python to aiohttp 🤔 Edit: Ok i did it now anyway: #681 #682 #683

I'm thinking about trying to do the same for tidalapi. 😬

PR #682, opened 2023-05-26
[Deezer] Quality of life and refactor

Un10ck3d:
- Adds `fd` to whitelisted protocols
- Adds documentation to manifest
- Updates throttler limits
- Stop using eval to convert to int
- Names the provider based on Deezer username
- Adds limit parameter to get_artist_top function


marcelveldt (dismissed):
great!

PR #685, opened 2023-05-27
Disable Youtube Music cipher caching

MarvinSchenkel:
This PR disables caching of ciphers to prevent weird playback errors. It also adds some extra debugging to catch edge cases related to streaming.


marcelveldt (approved):
only comments I have is that we should ask users to set debug level for YTM to debug if they have issues.

PR #712, opened 2023-06-13
Change radio endpoint + oauth fixes

MarvinSchenkel:
This PR:
- Changes the radio endpoint
- Fixes OAuth issues


marcelveldt:
Does this fix the unavailable tracks issue ?

marcelveldt (approved):

MarvinSchenkel:
> Does this fix the unavailable tracks issue ?

It might, since we use a different endpoint now to obtain radio tracks. Hopefully YT will not return unavailable tracks, but if they do, we need to build on top of this anyway. Radio mode ensures that the generated radio is unique rather than static

PR #736, opened 2023-07-04
Add chapters for plex audio books

micha91:
Finally the chapters are now available in the plexapi library 🥳


marcelveldt:
Great, now we have a reason to actually use this :-)

PR #738, opened 2023-07-06
Fix tidal issue with 'gender'

jozefKruszynski:
Fix for the recent tidalapi change


marcelveldt (approved):
at some point we should upstream the changes

jozefKruszynski:
There is still one PR that is waiting to be merged at the python-tidal project, it has been approved, but they haven't got round to merging it for some reason, they already merged a couple of other PRs from me. Once that one is merged then we can finally switch to the official lib.

PR #801, opened 2023-07-29
[Deezer] Switch to async library

Un10ck3d:
Huge performance upgrade to Deezer. Especially search which is now roughly 25 times faster than before.


music_assistant/server/providers/deezer/__init__.py
```diff
@@ -71,6 +72,22 @@
...
+
+@dataclass
+class Credential:
+    """Class for storing credentials."""
+
+    def __init__(self, app_id: int, app_secret: str, access_token: str):
+        """Set the correct things."""
+        self.app_id = app_id
+        self.app_secret = app_secret
+        self.access_token = access_token
```
marcelveldt, line 84: this is redundant in case of a dataclass

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -71,6 +72,22 @@
     ProviderFeature.RECOMMENDATIONS,
 )
 
+
+@dataclass
+class Credential:
```
marcelveldt, line 77: Maybe rename to DeezerCredentials ?

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -147,11 +158,16 @@ async def handle_setup(self) -> None:
             access_token=self.config.get_value(CONF_ACCESS_TOKEN),  # type: ignore
         )
         try:
-            deezer_client = await DeezerClient.get_deezer_client(self=None, creds=self.creds)
-            self.client = DeezerClient(creds=self.creds, client=deezer_client)
+            self.client = deezer.Client(
```
marcelveldt, line 161: there is no hidden IO in the instantiating of the Client ?

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -273,26 +289,30 @@ async def get_playlist_tracks(
     ) -> AsyncGenerator[PlaylistTrack, None]:
         """Get all tracks in a playlist."""
         playlist = await self.client.get_playlist(playlist_id=prov_playlist_id)
-        for count, deezer_track in enumerate(playlist.tracks, start=1):
-            track = self.parse_track(track=deezer_track, user_country=self.gw_client.user_country)
+        for count, deezer_track in enumerate(await playlist.get_tracks(), start=1):
```
marcelveldt, line 292:
```suggestion
        playlist_tracks = await playlist.get_tracks()
        for count, deezer_track in enumerate(playlist_tracks, start=1):
```

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -518,16 +533,46 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
+    async def parse_track(
         self,
         track: deezer.Track,
         user_country: str,
         extra_init_kwargs: dict[str, Any] | None = None,
-    ) -> Track | PlaylistTrack:
+    ) -> Track | PlaylistTrack | AlbumTrack:
         """Parse the deezer-python track to a MASS track."""
+        try:
+            deezer_artist = await track.get_artist()
```
marcelveldt, line 554: only return ItemMapping here if you get some limited object here.
If you actually get back a Artist object here with some details filled in already (like image etc), just use that.

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -518,16 +533,46 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
+                deezer_artist.name,
+            )
+        except deezer.exceptions.DeezerErrorResponse:
+            artist = ItemMapping(
+                MediaType.ARTIST,
+                self.instance_id,
+            )
+        try:
+            deezer_album = await track.get_album()
+            album = ItemMapping(
```
marcelveldt, line 568: only return ItemMapping here if you get some limited object here.
If you actually get back a Album object here with some details filled in already (like image etc), just use that.

music_assistant/server/providers/deezer/manifest.json
```diff
@@ -5,6 +5,6 @@
   "description": "Support for the Deezer streaming provider in Music Assistant.",
   "codeowners": ["@Un10ck3d",  "@micha91"],
   "documentation": "https://github.com/orgs/music-assistant/discussions/1245",
-  "requirements": ["deezer-python==6.0.0", "pycryptodome==3.18.0"],
+  "requirements": ["git+https://github.com/music-assistant/deezer-python-async", "pycryptodome==3.18.0"],
```
marcelveldt, line 8: at some point maybe publish it to pypi ?

marcelveldt (dismissed):
see inline feedback

music_assistant/server/providers/deezer/manifest.json
Un10ck3d: Yeah good idea. But i should probably talk to the owner of deezer-python first

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: No, but that makes me wonder why i ever added the try catch thing...

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: Yup good idea

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: Oh yeah..

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: I ended up not fetching either artist or album since its slow

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: I ended up not fetching either artist or album since its slow

marcelveldt:
is this one ready to go now ?

Un10ck3d:
Yes now it is

marcelveldt:
@micha91 as you have more experience with deezer, can you do the review and merge of this one ?

Un10ck3d:
The changed code that probably need review the most is this: https://github.com/music-assistant/deezer-python-async

micha91 (changes requested):
First of all: I can confirm that this increases the performance a lot - great job! Until now the search functionality was that slow that I nearly never used it, but this will change now.
Beside that I have a few comments, especially regarding the attributes of Deezer data classes. In addition I am wondering, whether we really want to have the parse functions async. As far as I see, we don't perform any requests in there so we don't have a benefit when we implement them async.

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -526,21 +531,56 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
+            owner=creator.name,
         )
 
-    def parse_track(
+    def get_playlist_creator(self, playlist: deezer.Playlist):
+        """See https://twitter.com/Un10cked/status/1682709413889540097."""
+        try:
+            return playlist.creator
+        except AttributeError:
+            return playlist.user
```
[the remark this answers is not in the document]
Un10ck3d: >imho it's always better to check a condition instead of catching an exception, so I would change it as suggested

Yeah thats very true i didnt even think of that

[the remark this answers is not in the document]
Un10ck3d: There is probably alot of other places we can change that as well

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -275,29 +280,31 @@ async def get_playlist_tracks(
     ) -> AsyncGenerator[PlaylistTrack, None]:
         """Get all tracks in a playlist."""
         playlist = await self.client.get_playlist(playlist_id=prov_playlist_id)
-        for count, deezer_track in enumerate(playlist.tracks, start=1):
-            track = self.parse_track(
+        playlist_tracks = await playlist.get_tracks()
+        count = 1
+        async for deezer_track in playlist_tracks:
```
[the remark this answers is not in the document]
Un10ck3d: Marcel actually suggested the opposite 😆 https://github.com/music-assistant/server/pull/801#discussion_r1281125542
I dont know what is better 🤔

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -445,18 +450,15 @@ def parse_metadata_track(self, track: deezer.Track) -> MediaItemMetadata:
         """Parse the track metadata."""
         try:
             return MediaItemMetadata(
-                preview=track.preview,
```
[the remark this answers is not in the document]
Un10ck3d: I dont remember, probably because its one of those things that only sometimes gets fetched by the library

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -526,21 +531,56 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
+                media_type=MediaType.ARTIST,
+                item_id=str(track.artist.id),
+                provider=self.instance_id,
+                name=track.artist.name,
+            )
+        except (deezer.exceptions.DeezerErrorResponse, AttributeError):
+            artist = ItemMapping(
+                media_type=MediaType.ARTIST,
+                provider=self.instance_id,
+            )
```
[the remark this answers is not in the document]
Un10ck3d: Yeah i agree didnt think of that

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -554,57 +594,81 @@ def parse_track(
...
+        index = 0
+        async for playlist in deezer_playlists:
+            playlists.append(self.parse_playlist(playlist))
+            index += 1
+            if index >= limit:
+                return playlists
+        return playlists
 
     ### OTHER PARSING FUNCTIONS ###
     def _get_album(self, track: deezer.Track) -> Album | None:
```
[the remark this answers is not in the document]
Un10ck3d: Yeah.. It was used earlier but now its unused. Im deleting it

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -659,9 +723,14 @@ async def get_track_content_type(self, gw_client: GWClient, track_id: int):
...
-    def track_available(self, track: deezer.Track, user_country: str) -> bool:
+    def track_available(
+        self, track: deezer.Track, user_country: str, track_type: Track | PlaylistTrack | AlbumTrack
+    ) -> bool:
         """Check if a given track is available in the users country."""
-        return user_country in track.available_countries
+        if isinstance(track_type, Track):
+            return user_country in track.available_countries
+        else:
+            return True
```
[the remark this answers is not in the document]
Un10ck3d: Uhm.. i dont remember why i did that... i suppose i will just make it check it always :shrug:

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -554,57 +594,81 @@ def parse_track(
...
-                    name=track.artist.name,
-                )
-            ],
-            album=ItemMapping(
-                media_type=MediaType.ALBUM,
-                item_id=str(track.album.id),
-                provider=self.domain,
-                name=track.album.title,
-            ),
+            artists=[artist],
```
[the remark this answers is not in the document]
Un10ck3d: Yeah good idea

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -526,21 +531,56 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
+            )
+        try:
+            album = ItemMapping(
+                media_type=MediaType.ALBUM,
+                item_id=str(track.album.id),
+                provider=self.instance_id,
+                name=track.album.name,
+            )
+        except (deezer.exceptions.DeezerErrorResponse, AttributeError):
+            album = None
```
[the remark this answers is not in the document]
Un10ck3d: Yeah

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -526,21 +531,56 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
...
         track: deezer.Track,
         user_country: str,
         extra_init_kwargs: dict[str, Any] | None = None,
-    ) -> Track | PlaylistTrack:
+    ) -> Track | PlaylistTrack | AlbumTrack:
         """Parse the deezer-python track to a MASS track."""
+        try:
+            isrc = track.isrc
+        except AttributeError:
+            isrc = None
```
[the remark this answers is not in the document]
Un10ck3d: Yup when thinking about i definetly also prefer hasattr i just didnt think of it..

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
Un10ck3d: earlier it was doing async work but i removed it

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
Un10ck3d: It throws an attribute error if i remove it

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
Un10ck3d: I readded it along with a bunch of other metadata

Un10ck3d:
These are the changes made in the deezer library: https://github.com/music-assistant/deezer-python-async/pull/1

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
micha91: hmm, I just wanted to keep it consistent to the way we do it in `get_library_albums` and other functions. But in the end it's more a decision of personal preferences than right/wrong 😉

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -275,29 +280,31 @@ async def get_playlist_tracks(
...
         artist = await self.client.get_artist(artist_id=int(prov_artist_id))
-        top_tracks = await self.client.get_artist_top(artist=artist, limit=25)
+        top_tracks = await artist.get_top()
         return [
             self.parse_track(track=track, user_country=self.gw_client.user_country)
-            for track in top_tracks
+            for track in islice(top_tracks, 0, 25)
         ]
 
     async def library_add(self, prov_item_id: str, media_type: MediaType) -> bool:
```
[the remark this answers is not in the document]
micha91: same for the calls performed in `library_remove`

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -363,23 +370,21 @@ async def recommendations(self) -> list[BrowseFolder]:
 
     async def add_playlist_tracks(self, prov_playlist_id: str, prov_track_ids: list[str]):
         """Add tra ck(s) to playlist."""
-        await self.client.add_playlist_tracks(
-            playlist_id=prov_playlist_id, tracks=[int(i) for i in prov_track_ids]
-        )
+        playlist = await self.client.get_playlist(prov_playlist_id)
```
micha91, line 373: `get_playlist` expects an int, but a str is provided

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -363,23 +370,21 @@ async def recommendations(self) -> list[BrowseFolder]:
...
         async for track in self.get_playlist_tracks(prov_playlist_id):
             if track.position in positions_to_remove:
-                playlist_track_ids.append(track.id)
+                playlist_track_ids.append(int(track.item_id))
             if len(playlist_track_ids) == len(positions_to_remove):
                 break
-        await self.client.remove_playlist_tracks(
-            playlist_id=prov_playlist_id, tracks=list(playlist_track_ids)
-        )
+        playlist = await self.client.get_playlist(prov_playlist_id)
```
micha91, line 386: `get_playlist` expects an int, but a str is provided

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -554,64 +596,80 @@ def parse_track(
...
-            ),
+            artists=[artist] if artist else [],
+            album=album,
             provider_mappings={
                 ProviderMapping(
                     item_id=str(track.id),
                     provider_domain=self.domain,
                     provider_instance=self.instance_id,
-                    available=self.track_available(track, user_country),
+                    available=self.track_available(track, user_country, track_class),
```
micha91, line 608:
```suggestion
                    available=self.track_available(track, user_country),
```

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -554,64 +596,80 @@ def parse_track(
...
+        index = 0
+        async for playlist in deezer_playlists:
+            playlists.append(self.parse_playlist(playlist))
+            index += 1
+            if index >= limit:
+                return playlists
+        return playlists
 
     ### OTHER FUNCTIONS ###
     async def update_access_token(self, app_id, app_secret, code, http_session=None) -> str:
```
micha91, line 675: Why are `update_access_token` and `_post_http` part of the class? `_post_http` is used in the `update_access_token` function only and that function is used from outside the class only as far as I can see. For that reason it makes much more sense to move them out of the class. This way you can also use them more easily in the `get_config_entries` function 🙂

micha91 (changes requested):
Some more comments

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
Un10ck3d: Hmm, where do you see that? https://github.com/music-assistant/deezer-python-async/blob/main/src/deezer/client.py

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: I think i was moving things around and didnt think about it. But good point im moving it now

music_assistant/server/providers/deezer/__init__.py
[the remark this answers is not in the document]
micha91: oh, then it's just about typos: You are calling `add_user_playlists`, but you should call `add_user_playlist` and the same for artist, track, album and all the remove functions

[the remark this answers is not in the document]
Un10ck3d: Oh yeah, didnt see that. Changed in newest commit

Un10ck3d:
This should be ready right?

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -83,6 +92,24 @@
 DEEZER_APP_SECRET = app_var(7)
 
 
+async def update_access_token(
+    app_id: str, app_secret: str, code: str, http_session: ClientSession
+) -> str:
+    """Update the access_token."""
+    response = await http_session.post(
+        "https://connect.deezer.com/oauth/access_token.php",
+        json={"code": code, "app_id": app_id, "secret": app_secret},
```
micha91, line 101:
```suggestion
        params={"code": code, "app_id": app_id, "secret": app_secret},
```
Otherwise it does not work for me on a clean installation. Until now we set `params` and `json`, but params seems to be good enough.

micha91 (changes requested):
Beside the login thing, it all seems to be working and looking good. Imho there is nothing urgent left to be changed 🙂

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: Oh yeah forgot to test it, I changed it and then wanted to test if it worked later but forgot..

Un10ck3d: Random question: Is it the reviewer or the pr author that is supposed to click "Resolve conversation"? :laughing: Till now i just did it myself but i guess the reviewer should be the one to decide whether it has been resolved... I dont know, im still fairly new to all of this i guess

micha91: In most cases I do it the same way. I think it highly depends on whether you apply a requested change. If you do so, I would say it's okay to resolve it on your own. If not, you should answer the comment and discuss to find the best solution before someone resolves the comment

music_assistant/server/providers/deezer/manifest.json
```diff
@@ -5,6 +5,6 @@
   "description": "Support for the Deezer streaming provider in Music Assistant.",
   "codeowners": ["@Un10ck3d",  "@micha91"],
   "documentation": "https://github.com/orgs/music-assistant/discussions/1245",
-  "requirements": ["deezer-python==6.0.0", "pycryptodome==3.18.0"],
+  "requirements": ["git+https://github.com/music-assistant/deezer-python-async@convert_to_async", "pycryptodome==3.18.0"],
```
micha91, line 8: I think, we should merge convert_to_async before we merge this PR here in MA. Then we could set a tag on main in deezer-python-async and reference that here in the requirements. This way we ensure that we are only using a tested version of deezer-python-async as a requirement. Otherwise we could run into issues in MA whenever we change something in the other repo, without even realizing it here.

music_assistant/server/providers/deezer/__init__.py
Un10ck3d: Alright thanks 🙏

music_assistant/server/providers/deezer/manifest.json
Un10ck3d: Just made the [tag on deezer-python-async](https://github.com/music-assistant/deezer-python-async/releases/tag/v0.1.0). Im now changing the dependency in the manifest and will test if everything still works

Un10ck3d: Yes i just commited the change in deps. 
The [itemmapping issue](https://discord.com/channels/753947050995089438/1004642314968367124/1139145197314379877) is still present. But that a music assistant general error. This pr is good to go from me

micha91 (approved):
Awesome work. Also had a quick end2end test right after these changes and everything still seems to work as expected

marcelveldt:
Amazing work @Un10ck3d, congrats in getting this done!

Feel free to merge and include it in the next release if you see fit

Un10ck3d:
Alright very cool! This was a bit of a fight :laughing: Feels good to finally have it working! And thanks for reviewing my bad code :rofl:

PR #824, opened 2023-08-07
Readme: Add data directory mount to docker run command and explanatory note

krisnoble:
Include a mount for the data store in the `docker run` command so that data persists between updates, and an explanatory note below. 

Could also just have the note and not include the `-v ...` but I feel this is an important thing to include since missing it out means data loss - potentially a lot of time wasted and users lost if someone sets up their entire library, setup etc first time, then updates and loses everything.


marcelveldt (approved):
thanks!

PR #831, opened 2023-08-08
Deezer provider: Use kwargs only to create ItemMapping

micha91:
fixes a bunch of exceptions occurring during library sync.


Un10ck3d:
All of this is actually already fixed in this pr: https://github.com/music-assistant/server/pull/801 🙂 I just didn't bother fixing it in the non async version since it's double work and we are hopefully merging the async version before next beta releases. Sorry I should have mentioned that to you...

micha91:
No worries. I will test that branch later today and give feedback 🙂

Un10ck3d:
Depends on if the async version makes it in before next beta

PR #848, opened 2023-08-17
Fix similar tracks function

jozefKruszynski:
Fixes issue: https://github.com/music-assistant/hass-music-assistant/issues/1475

### DIFF
--- a/music_assistant/server/providers/tidal/__init__.py
+++ b/music_assistant/server/providers/tidal/__init__.py
@@ -310,13 +310,14 @@ class TidalProvider(MusicProvider):
             )
             yield track
 
-    async def get_similar_tracks(self, prov_track_id: str, limit=25) -> list[Track]:
+    async def get_similar_tracks(self, prov_track_id: str, limit=25) -> list[Track]:  # noqa: ARG002
         """Get similar tracks for given track id."""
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Ruff is failing so we cannot merge this :(. Probably also need to remove the limit here: https://github.com/music-assistant/server/blob/01d3bfe982499da3107cbdf660a26f427b50ed65/music_assistant/server/providers/tidal/__init__.py#L319

jozefKruszynski:
> Ruff is failing so we cannot merge this :(. Probably also need to remove the limit here:
> 
> https://github.com/music-assistant/server/blob/01d3bfe982499da3107cbdf660a26f427b50ed65/music_assistant/server/providers/tidal/__init__.py#L319
> 
> and in the helper function to keep ruff happy.

Ooooh, this is a real stinker as I would also have to remove it from the function in __init__.py. This won't work, so I'll have to add an exemption to the method for now I think

MarvinSchenkel:
> 

I think just removing it from the helper function will be fine for now.

MarvinSchenkel:
> 

Yes you are right, I just edited my answer :)

MarvinSchenkel (approved):

PR #849, opened 2023-08-17
Re-add addon update action to ci

jozefKruszynski:
Re-added the addon upadte action, this time with a depdencey to both prior jobs. The addon update action no longer sorts the yaml keys incorrectly and the mismatch of version/tag has been corrected here:
https://github.com/music-assistant/addon-update-action/commit/e9441782b6a4f929c6dd797ec72923919d6d1524

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,12 +101,12 @@ jobs:
           build-args: |
             "MASS_VERSION=${{ needs.build-and-publish-pypi.outputs.version }}"
 
-  #addon-version-update:
-  #  name: Server repo PR
-  #  needs: build-and-publish-pypi
-  #  runs-on: ubuntu-latest
-  #  steps:
-  #    - uses: music-assistant/addon-update-action@main
-  #      with:
-  #        github_token: ${{ secrets.PRIVILEGED_GITHUB_TOKEN }}
-  #        new_release_version: ${{ needs.build-and-publish-pypi.outputs.version }}
+  addon-version-update:
+    name: Server repo PR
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -101,12 +101,12 @@ jobs:
...
-  #  needs: build-and-publish-pypi
-  #  runs-on: ubuntu-latest
-  #  steps:
-  #    - uses: music-assistant/addon-update-action@main
-  #      with:
-  #        github_token: ${{ secrets.PRIVILEGED_GITHUB_TOKEN }}
-  #        new_release_version: ${{ needs.build-and-publish-pypi.outputs.version }}
+  addon-version-update:
+    name: Server repo PR
+    needs: [ build-and-publish-pypi, build-and-push-container-image ]
```
MarvinSchenkel, line 106: I think the `build-and-publish-pypi` bit is redundant and can be removed, since `build-and-push-container-image` already depends on it.

jozefKruszynski: I kept it in this case rather than adding a new output to the build-and-push-container-image job. I was tossing up what would be the best.
Let me double check what the default outputs are from that docker stuff

jozefKruszynski: From double checking a few sources, it seems that as I need the output created in the first job, I have to keep it as a dependency in the add-on update job.
It also feels cleaner to do this, rather than essentially read that output as an input in the docker build job and then output the same value again to use in the add-on update job.

MarvinSchenkel (approved):

PR #857, opened 2023-08-27
Feature/tidal-quality-updates

jozefKruszynski:
Update tidal api
Enable HI RES
Add Quality selector to config map
Re-add track limit for similar tracks function
Add parsing to determine stream quality

### DIFF
--- a/music_assistant/server/providers/tidal/__init__.py
+++ b/music_assistant/server/providers/tidal/__init__.py
@@ -46,6 +46,7 @@ from music_assistant.common.models.media_items import (
     Track,
 )
 from music_assistant.server.helpers.auth import AuthenticationHelper
+from music_assistant.server.helpers.tags import AudioTags, parse_tags
 from music_assistant.server.models.music_provider import MusicProvider
 
 from .helpers import (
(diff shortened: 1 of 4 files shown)


music_assistant/server/providers/tidal/__init__.py
```diff
@@ -89,11 +96,12 @@ async def setup(
     return prov
 
 
-async def tidal_code_login(auth_helper: AuthenticationHelper) -> TidalSession:
+async def tidal_code_login(auth_helper: AuthenticationHelper, quality: str) -> TidalSession:
     """Async wrapper around the tidalapi Session function."""
 
     def inner() -> TidalSession:
-        config = TidalConfig(quality=TidalQuality.lossless, item_limit=10000, alac=False)
+        print(quality)
```
Un10ck3d, line 103: Is this still supposed to be here?

jozefKruszynski: Good spot. No it isn't

marcelveldt:
What's the breaking change here ?

jozefKruszynski:
> What's the breaking change here ?

New config property, that does not exist prior to this version

marcelveldt:
But you have a default value right ? So then it should not break ?!

jozefKruszynski:
True.

jozefKruszynski:
Going from previous setup provider on main branch and then switching to this feature branch, there was indeed no error. Everything plays nicely.

PR #862, opened 2023-09-01
Fix itemmapping issue on theaudiodb

arctixdev:


arctixdev:
Closing since this isnt a good fix nor am i sure its actually still a problem

PR #867, opened 2023-09-06
Fix child players get stuck on zero volume 

OzGav:
Originally raised here https://github.com/music-assistant/hass-music-assistant/issues/1504

The suggested fix however would result in child volumes not going below 1. Looking at the current behavior the use of average of the powered child volumes for the group is good however the percentage change gives rise to unexpected results. Apart fro the zero case I also tried setting one speaker to 46 and another to 10 which gave a group volume of 28. Changing the group volume to 56 resulted in the first player going to 98 and the other to 14. I don't think this is expected. Therefore this PR suggests changing to relative movement. With the player volumes limited to 0 and 100 I think the result should be OK but I haven't tested it to see what the UX is like.

### DIFF
--- a/music_assistant/server/controllers/players.py
+++ b/music_assistant/server/controllers/players.py
@@ -464,6 +464,8 @@ class PlayerController(CoreController):
         for child_player in self._get_child_players(group_player, True):
             cur_child_volume = child_player.volume_level
             new_child_volume = int(cur_child_volume + volume_dif)
+            new_child_volume = max(0, new_child_volume)
+            new_child_volume = min(100, new_child_volume)
             coros.append(self.cmd_volume_set(child_player.player_id, new_child_volume))
         await asyncio.gather(*coros)
 


marcelveldt:
This fix is not good as a group volume change always has to be relative. It just needs an adjustment to prevent the zero

OzGav:
Why do you think it has to be relative? I thought my example of 

> I also tried setting one speaker to 46 and another to 10 which gave a group volume of 28. Changing the group volume to 56 resulted in the first player going to 98 and the other to 14.

is valid as an unexpected outcome. I doubled the group volume and one volume more than doubles and the other less than a 50% increase.

marcelveldt:
hmm, I'll play with it a bit but usually group volume adjustments are relative. You increase volume with 30%, then each individual player should be increased with 30%, regardless of its current value. It should just be shifted.

marcelveldt:
Actually, tested a bit with your proposal to just use the relative absolute value instead of the percentage and its working just fine. Not sure why I overcomplicated my thinking with the percentage as this works just fine. Only added a small clamping to min/max values and all set. Thanks!

PR #868, opened 2023-09-08
FIx cumbersome restart, add idx to album tracks

jozefKruszynski:
Due to config values being altered, this is a breaking change.
Remove and re-add plex provider
Added idx and enumeration to get_album_tracks function to avoid the `None`or `0` track number issue
We should do a follow up PR to remove this assert in albums.py. I don't have time right now to ensure all providers uphold this contract.

### DIFF
--- a/music_assistant/server/providers/plex/__init__.py
+++ b/music_assistant/server/providers/plex/__init__.py
@@ -560,11 +560,12 @@ class PlexProvider(MusicProvider):
         plex_album: PlexAlbum = await self._get_data(prov_album_id, PlexAlbum)
         tracks = []
         for idx, plex_track in enumerate(await self._run_async(plex_album.tracks), 1):
+            track_number = plex_track.trackNumber if plex_track.trackNumber != None else idx
             track = await self._parse_track(
                 plex_track,
                 {
                     "disc_number": plex_track.parentIndex,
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/plex/__init__.py
```diff
@@ -556,10 +559,13 @@ async def get_album_tracks(self, prov_album_id: str) -> list[AlbumTrack]:
...
         plex_album: PlexAlbum = await self._get_data(prov_album_id, PlexAlbum)
         tracks = []
-        for plex_track in await self._run_async(plex_album.tracks):
+        for idx, plex_track in enumerate(await self._run_async(plex_album.tracks), 1):
             track = await self._parse_track(
                 plex_track,
-                {"disc_number": plex_track.parentIndex, "track_number": plex_track.trackNumber},
+                {
+                    "disc_number": plex_track.parentIndex,
+                    "track_number": idx,
```
marcelveldt, line 567: Is this on purpose ? Now you are overwriting any value plex had already
maybe change to "track_number": plex_track.trackNumber or idx

marcelveldt:
Did you test this PR ? Safe to merge ?

jozefKruszynski:
Quick and dirty tests yes, extensive no.

PR #916, opened 2023-11-09
Snapcast Provider

SantiagoSotoC:
Allows snapcast to be used as a provider.
Ports 4953 and above must be open on the snapacast server. For the moment I use snapcast as tcp server and music assistant writes to the socket.


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+        player = self.mass.players.get(player_id, raise_unavailable=False)
+        if not player:
+            player = Player(
+                player_id=player_id,
+                provider=self.domain,
+                type=PlayerType.PLAYER,
+                name=snap_client.friendly_name,
+                available=True,
+                powered=snap_client.connected,
+                device_info=DeviceInfo(),
```
marcelveldt, line 117: Maybe provide some default name here like SnapCast Player ?

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+        return [ele for ele in group.clients if ele != player_id]
+
+    async def cmd_sync(self, player_id: str, target_player: str) -> None:
+        """Sync Snapcast player."""
+        child_player = self.mass.players.get(player_id)
+        assert child_player  # guard
+        parent_player = self.mass.players.get(target_player)
+        assert parent_player  # guard
+        # always make sure that the parent player is part of the sync group
+        # parent_player.group_childs.add(parent_player.player_id)
```
marcelveldt, line 245: is this leftover ?

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+            await group.set_stream(stream_id)
+
+        stream = self._get_snapstream(player_id)
+
+        stream_host = stream._stream.get("uri").get("host")
+        ffmpeg = (
+            FFmpeg()
+            .option("y")
+            .input(url)
+            .output(f"tcp://{stream_host}", f="u16le", acodec="pcm_s16le", ac=2, ar=48000)
```
marcelveldt, line 180: does this need to be configurable for other sample rates ?

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+        """Unsync Snapcast player."""
+        group = self._get_snapgroup(player_id)
+        await group.remove_client(player_id)
+        group = self._get_snapgroup(player_id)
+        stream_id = await self._get_empty_stream(player_id)
+        await group.set_stream(stream_id)
+        self._handle_update()
+
+    def _get_snapgroup(self, player_id):
+        client = self._snapserver.client(player_id)
```
marcelveldt, line 265: All methods below are mnissing docstrings

marcelveldt: and missing return annotations

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+        group = self._get_snapgroup(player_id)
+        return self._snapserver.stream(group.stream)
+
+    def _synced_to(self, player_id):
+        ret = None
+        snap_group = self._get_snapgroup(player_id)
+        snap_clients = list(filter(lambda x: x != player_id, snap_group.clients))
+        if player_id == snap_group.clients[0]:  # Player is a Sync group master
+            player = self.mass.players.get(player_id)
+            player.group_childs.clear()
```
marcelveldt, line 278: there is a side effect in a method to get the synced groups ?

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+            for snap_client in snap_clients:
+                player.group_childs.add(snap_client)
+        elif len(snap_clients) > 0:
+            ret = snap_clients[0]  # Return sync group master id
+        return ret
+
+    async def _get_empty_stream(self, player_id) -> str:
```
marcelveldt, line 285: what does this do ? maybe add a comment why this is here, why its needed ?

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: I can provide more specific information about the device

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: Yes, I remove it

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: This is the default sample rates in snapcast when creating a stream. I will check if I have a way to get it from the api.

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: Ok

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: ah ok all clear then. For some future iteration we might want to skip the intermediate ffmpeg in between and just forward the raw pcm bytes.

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: OK, please do so in a follow up PR then.

marcelveldt (changes requested):
see some in-line comments that are still open.

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: This is what I found the most accurate. 

```
 device_info=DeviceInfo(
                    model=snap_client._client.get("host").get("os"),
                    address=snap_client._client.get("host").get("ip"),
                    manufacturer=snap_client._client.get("host").get("arch"),
                ),
```

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: I removed it and changed the code so that these events are also processed through a callback.

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: I re-implemented this method, split it into 2 and removed the side effects. 
```
    def _synced_to(self, player_id) -> str | None:
        """Return player_id of the player this player is synced to."""
        snap_group = self._get_snapgroup(player_id)
        if player_id != snap_group.clients[0]:
            return snap_group.clients[0]

    def _group_childs(self, player_id) -> set[str]:
        """Return player_ids of the players synced to this player."""
        snap_group = self._get_snapgroup(player_id)
        snap_clients = [
            snap_client for snap_client in snap_group.clients if snap_client != player_id
        ]
        return set(snap_clients)
```

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: This looks for a stream in snapcast to send the audio to in case they are all used, it creates a new one, the default snapcast configuration is a named pipe.
This method tells snapcast to create a tcp socket as a server to send the audio to.  
Port 4953 is not chosen randomly. It is the default when creating a tcp socket in snapcast. 

[https://github.com/badaix/snapcast/blob/develop/doc/configuration.md#tcp-server](url)

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -0,0 +1,300 @@
...
+        if player_id != snap_group.clients[0]:
+            return snap_group.clients[0]
+
+    def _group_childs(self, player_id) -> set[str]:
+        """Return player_ids of the players synced to this player."""
+        snap_group = self._get_snapgroup(player_id)
+        snap_clients = [
+            snap_client for snap_client in snap_group.clients if snap_client != player_id
+        ]
+        return set(snap_clients)
```
marcelveldt, line 279:
```suggestion
        return {
            snap_client for snap_client in snap_group.clients if snap_client != player_id
        }
```

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: If you consider that it should go in the library it is not a problem, I will release a new version to simplify the imports. I could also add it to that release 
I can add parameters like codec, samplerate, bit depth, channels. 
I would look for one that is free and meets the criteria or create a new one. 
The codec is for distribution from the snapcast server to the clients. There is a client for esp32 without PSRAM that only supports Opus and pcm at the moment.

marcelveldt: OK, all clear! Let's go from here and improve/iterate later.

marcelveldt (approved):

marcelveldt (approved):

marcelveldt:
let's go and release this!
Awesome work @SantiagoSotoC

abelmatser:
I have not used this yet, but @SantiagoSotoC great work! Thanks @marcelveldt as well for the speedy reviews.

I am very much looking forward to this functionality.

PR #920, opened 2023-11-12
[soundcloud] bug fix: add limit parameter to get_recommended function

domanchi:
This PR fixes the error that occurs when attempting to start a radio for a given soundcloud track.

https://github.com/music-assistant/server/blob/ea1143408ef90637ae3508ee55905f78dbf87834/music_assistant/server/providers/soundcloud/__init__.py#L269

The caller specifies the `limit` parameter, but the underlying function doesn't accept it, causing the code to panic and crash.

The `limit` parameter isn't used for now, but it seems better to include it in the `get_recommended` function (to align with the other methods in the same class), rather than remove the parameter from the call site.


marcelveldt:
Always good to see a new contributor, thanks for spotting and fixing the issue!

marcelveldt (dismissed):

domanchi:
Thanks for the quick review!

PR #922, opened 2023-11-12
[soundcloud] improved track iteration

domanchi:
#### Problem

The soundcloud integration only loads the first 50 tracks from the user's list after setup. This is primarily due to the default limit imposed on the underlying API functionality:

https://github.com/music-assistant/server/blob/7b08dcf9e5d82884ee4500b6e4b8056d00c19229/music_assistant/server/providers/soundcloud/soundcloudpy/asyncsoundcloudpy.py#L121

For users with more than 50 liked tracks, this results in an incomplete import of such tracks into the Music Assistant.

#### Solution

We convert the soundcloud API library into an async generator, rather than a standard API fetcher. Pagination is implemented, and the `limit` parameter is made optional so that all tracks can be fetched if no limit is specified.

### Testing Done

This was manually tested with an explicit limit, as well as without.

### DIFF
--- a/music_assistant/server/providers/soundcloud/__init__.py
+++ b/music_assistant/server/providers/soundcloud/__init__.py
@@ -192,12 +192,7 @@ class SoundcloudMusicProvider(MusicProvider):
     async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
         """Retrieve library tracks from Soundcloud."""
         time_start = time.time()
-        tracks = await self._soundcloud.get_tracks_liked()
-        self.logger.debug(
-            "Processing Soundcloud library tracks took %s seconds",
-            round(time.time() - time_start, 2),
-        )
-        for item in tracks["collection"]:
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
I don't know about the internals of soundcloud but code wise this looks good.

marcelveldt:
Thanks for contributing again and feel free to mention your name in the maintainers section of the manifest.
Also, at some point the async soundcloud library should either be upstreamed or get its own pypy package but that is just thoughts for  later.

domanchi:
> Also, at some point the async soundcloud library should either be upstreamed or get its own pypy package but that is just thoughts for later.

Yeah, makes sense. I thought it was kept in-line here, forked from the original code, mainly for speed of iteration too.

https://github.com/music-assistant/server/blob/6f125311cd0d09c4e5183394ed9c681f5444d239/music_assistant/server/providers/soundcloud/soundcloudpy/asyncsoundcloudpy.py#L5

PR #935, opened 2023-11-25
fix: handle invalid start_index when receiving command player status

wiflye81:
This PR fix the Home Assistant squeezebox integration players discovery, by providing a valid start_index when receiving the command "player: status".
In this case, start_index was set to "status".


marcelveldt (approved):
Thanks!

PR #955, opened 2023-12-06
Fix snapcast provider in the dynamic generation of tcp streams 

SantiagoSotoC:
Snapcast expects ip 127.0.0.1 in tcp server mode, not the host's ip


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -297,7 +297,7 @@ async def _get_empty_stream(self) -> str:
         while True:
             port += 1
             new_stream = await self._snapserver.stream_add_stream(
-                f"tcp://{self.snapcast_server_host}:{port}?name={name}"
+                f"tcp://127.0.0.1:{port}?name={name}"
```
marcelveldt, line 300: is this correct as it now will only work if the snapcast server is on the same host as MA

SantiagoSotoC: If I was reading the documentation and it should be 0.0.0.0.0, it makes more sense.

marcelveldt (approved):

PR #969, opened 2023-12-17
Fixed theaudiodb provider with get_track_metadata issue

celestianx:
Initially, adb_track was initialized to None but not assigned anywhere in the function but called by self.__parse_track(adb_track) causing systematically an error. Just removed adb_track and replaced it by item referencing the matched item content


marcelveldt (approved):
Thanks!

PR #979, opened 2023-12-29
Retain extra url keys

OzGav:
I’m not sure if there is a downside to retaining the extra info but we have two open issues which will be resolved if we do so…?


marcelveldt:
For now this is ok but it will break HA's TTS so we need to deal with that as soon as we finish the announce feature

marcelveldt (approved):

OzGav:
Thanks. For my education, is the TTS passed as a URL?

PR #989, opened 2024-01-02
WIP: SiriusXM provider

scottt732:
An attempt at a provider for SiriusXM based on sxm-client ([source](https://github.com/AngellusMortis/sxm-client), [docs](https://sxm-client.readthedocs.io/en/latest/readme.html), [example player](https://github.com/AngellusMortis/sxm-player))

This requires a valid SiriusXM streaming account (https://www.siriusxm.com/plans) and doesn't attempt to get around any restrictions they put in place. You can get 1 stream per account at a time.

Browsing is working, but I'm having trouble getting the audio to play properly. My python is rusty and I'm pretty new to async/await/async generators in python so... go easy on me or feel free to add commits ;-)


music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+        mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    prov = SiriusXMProvider(mass, manifest, config)
+
+    await prov.handle_setup()
+    return prov
+
+
+# noinspection PyUnusedLocal
```
marcelveldt, line 64: none of our linters should require this ? what editor are you using for Python ?

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+            channels = self._channels_by_category[category_key]
+            for channel in channels:
+                yield self._parse_radio(channel)
+
+    async def get_library_radios(self) -> AsyncGenerator[Radio, None]:
+        """Retrieve library/subscribed playlists from the provider."""
+        for channel_guid, channel in self._channels_by_id.items():
+            yield self._parse_radio(channel)
+
+    async def get_radio(self, channel_id: str) -> Union[Radio, None]:
```
marcelveldt, line 232: you may use new style type annotations..
so instead of Union(Radio, None] you can just use Radio | None

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+
+        self.logger.info(f"Authenticating with SiriusXM...")
+        if not await self._client.authenticate():
+            raise RuntimeError(
+                f"Authentication failed"
+            )
+
+        await self._refresh_channels()
+
+    def _channel_updated(self, live_channel_raw: Dict):
```
marcelveldt, line 135:
```suggestion
    def _channel_updated(self, live_channel_raw: dict[str, Any]) -> None:

```

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+        # Sample Rate: 44100
+        # Bits per sample: 32
+
+        return StreamDetails(
+            provider=self.instance_id,
+            item_id=item_id,
+            audio_format=AudioFormat(
+                sample_rate=44100,
+                channels=2,
+                bit_depth=32,
```
marcelveldt, line 277: it is very, very unlikely that the bit_depth is 32 bits (that is only used in recording studios)
I think you meant 16 bits here ?

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+            ),
+            # stream_title=channel.name,
+            # duration=?,
+            # data={"url": url, "format": url_details["format"]}
+            # expires=time() + 24 * 3600,
+            # size=?,
+            media_type=MediaType.RADIO,
+            # data=live_channel.primary_hls.url,
+            direct=None,
+            # can_seek=False,
```
marcelveldt, line 288: can_seek should be set to False because these streams are not seekable

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+                sample_rate=44100,
+                channels=2,
+                bit_depth=32,
+                content_type=ContentType.M4A,
+            ),
+            # stream_title=channel.name,
+            # duration=?,
+            # data={"url": url, "format": url_details["format"]}
+            # expires=time() + 24 * 3600,
+            # size=?,
```
marcelveldt, line 284:
```suggestion
```

music_assistant/server/providers/siriusxm/__init__.py
scottt732: PyCharm Professional. It showed `mass`, `instance_id`, `action`, `values` as unused params w/warnings. I wasn't sure if I could just remove them or should just ignore.

music_assistant/server/providers/siriusxm/__init__.py
scottt732: Yeah I thought the same thing. I pulled this from VLC media/codec info. I had never seen 32-bit + 44.1KHz in the wild before. Usually it would be like 96KHz or 192KHz when people are going overboard on bit depth.

music_assistant/server/providers/siriusxm/__init__.py
scottt732: I'll set that for now. Their streams do support some DVR-like transport options. You can seek backwards or forward if you're paused a bit (not the future obviously)... but I don't know if I can implement that or not with their client. The JSON payload has like cut lists that tell you when songs/episodes start & stop and you can skip around by those fine/course-grained options a bit. But there are per-channel DRM restrictions in the payload about the number of seconds you can skip by and the maximum number of skips allowed in a certain time window.

I'll go with `can_seek=False` for now. I think music assistant would probably need to extend the provider model a little bit to support these kinds of things.

music_assistant/server/providers/siriusxm/__init__.py
```diff
@@ -0,0 +1,437 @@
...
+        self.logger.info(f"Fetching playlist for channel: {channel_id}")
+        playlist = await self._client.get_playlist(channel_id, use_cache=False)
+
+        playlist_paths = list(filter(lambda l: not l.startswith("#"), playlist.split('\n')))
+        for playlist_path in playlist_paths:
+            self.logger.info(f"Getting segment: {playlist_path}")
+            yield await self._client.get_segment(playlist_path)
+
+    async def get_audio_stream(self, stream_details: StreamDetails, seek_position: int = 0) -> AsyncGenerator[bytes, None]:
+        async for chunk in self.get_playlist_items_generator(stream_details):
```
marcelveldt, line 307: this is not going to work because as far as I can see the "get_playlist_items_generator" returns audio file URLs and not raw audio bytes... You will have to fetch the segment through the http client and then stream the content.

It could be as simple as this:

```
from music_assistant.server.helpers.audio import get_http_stream

async for playlist_chunk_url in self.get_playlist_items_generator(stream_details):
   async for chunk in get_http_stream(
                self.mass, playlist_chunk_url, streamdetails
            ):
                yield chunk
```

music_assistant/server/providers/siriusxm/__init__.py
marcelveldt: You may never remove them but instead you inform the linter that it may be ignored.
We use the linters that are configured in the pre-commit config

marcelveldt: We all use VSCode btw ;-)

music_assistant/server/providers/siriusxm/__init__.py
scottt732: The way the sxm-client works (when you're not relying on their HTTP proxy endpoint) is that you request a playlist for a channel which returns an m3u8 file as a string. Here's a fragment of a random one I grabbed:

```
#EXTM3U
#EXT-X-VERSION:1
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:104665
#EXT-X-ALLOW-CACHE:NO
#EXT-X-KEY:METHOD=AES-128,URI="key/1"
#EXT-X-PROGRAM-DATE-TIME:2023-12-23T14:02:23.561+00:00
#EXTINF:10,
AAC_Data/cnn/HLS_cnn_256k_v3/cnn_256k_1_122350543561_00104665_v3.aac
#EXT-X-PROGRAM-DATE-TIME:2023-12-23T14:02:33.312+00:00
#EXTINF:10,
AAC_Data/cnn/HLS_cnn_256k_v3/cnn_256k_1_122350553312_00104666_v3.aac
#EXT-X-PROGRAM-DATE-TIME:2023-12-23T14:02:43.064+00:00
#EXTINF:10,
AAC_Data/cnn/HLS_cnn_256k_v3/cnn_256k_1_122350563064_00104667_v3.aac
...etc
```

While you can get the root URL from the client (a CDN or something), the HTTP requests that you'd make require session tokens and a bunch of other stuff. Their client handles that for you. So you make a request like this:

`bytes = await sxm_client.get_segment('AAC_Data/cnn/HLS_cnn_256k_v3/cnn_256k_1_122350543561_00104665_v3.aac')`

(via https://github.com/AngellusMortis/sxm-client/blob/dd25b56d7a8a7c33399b25e72a7c2e136fb30e6a/sxm/client.py#L382)

and you get back the bytes from the http response ... where the request had all of the session-related magic taken care of behind the scenes.

music_assistant/server/providers/siriusxm/__init__.py
marcelveldt: well, you would hear mangled audio if its set to the wrong bit depth so you will know soon enough :-)

music_assistant/server/providers/siriusxm/__init__.py
marcelveldt: ooh fancy... yeah that would require extending our models a bit in the future yeah

music_assistant/server/providers/siriusxm/__init__.py
marcelveldt: ah yes, that sounds like exactly what you need then!

scottt732: What I was trying to achieve here is to get music assistant to basically pull the next segment as needed but not try to download them all. Like can I yield the bytes of file #1, then yield the bytes of file #2 when music assistant needs them, etc.? This is what I was trying to do here. Also do you know if I would need to break these segments down into smaller chunks & yield those?

marcelveldt: Yeah, you just need to make sure to grab each segment, yield chunks and then grab next segment yield chunks and so on. We have some helpers in the audio.py module to help with parsing the m3u file if you want that.

and yes, you need to yield the smaller chunks otherwise it will all be loaded in memory
Basically you just reimplement the "get_segment" function from the lib but instead of returning the resp.content, you just yield chunks from the resp.iter_any or resp.iter_chunked

So something along these lines within the get_audio_stream:

```
playlist_file = get_sirius_playlist_m3u8_file
playlist_segments = parse_playlist_into_the segment_parts

for playlist_segment in playlist_segments:
    url = urllib.parse.urljoin(await self.get_hls_root(), path)
    res = await self._session.get(url, params=self._token_params())

        if res.status_code == 403:
            raise SegmentRetrievalException(
                "Received status code 403 on segment, renew session"
            )

        if res.is_error:
            self._log.warn(f"Received status code {res.status_code} on segment")
            return None

        for chunk in res.iter_any():
            yield chunk

```

You could try first by just using the entire contents from the get_segment function btw, maybe its not that big

scottt732:
It turns out I needed to setup that http endpoint to get the sxm-client ~working. I was running into ffmpeg errors trying to calculate loudness and it looks like those segments I thought I was getting back as aac or m4a were actually hls or something along those lines. I verified with ffprobe that the segment data I was getting back was invalid. 

With the latest commit, by passing the m3u8 path as `direct=f"http://127.0.0.1:{self._bind_port}/{channel_id}.m3u8`, I can get it to play the first minute or so (VLC is able to play the whole stream). I assume it's handling the first segment only.  I see this warning coming from code outside my provider:

```
2024-01-04 17:19:55.285 WARNING (MainThread) [py.warnings] /home/scottt732/code/music-assistant/server/music_assistant/server/helpers/audio.py:435: RuntimeWarning: coroutine 'MusicProvider.get_audio_stream' was never awaited
  async for audio_chunk in music_prov.get_audio_stream(streamdetails, seek_pos):
```

I haven't had any luck with getting  `get_audio_stream` to work with a generator yielding chunks of bytes. For some reason, it looks like the `get_playlist` generator called by `get_audio_stream` ends up getting called more than once for some reason (maybe `get_audio_stream` is getting called more than once). This is tricky b/c of the DRM restrictions server-side. I really need to somehow ensure that I'm only asking the client about 1 channel at a time and that I'm not trying to load the segments in parallel or out-of-order.  I left this implementation commented out in case you have any pointers.

One other quick question... There is a callback from the sxm-client where I can be notified when the currently playing track changes. Is there any way I can update the footer of the web UI?  I assumed this would be in/around `self.mass.players` but I wasn't sure how to find the player_id's that happen to be tuned in to this provider.

marcelveldt:
Exceptions for tasks can sometimes bubble up in a somewhat strange fashion with asyncio. In your case you can see the error MusicProvider.get_audio_stream' was never awaited which means something crashed. The code was porbbaly still calling the get_audio_stream in your provider Sure that happened too when you had set the "direct" attribute of the streamdetails.

if you fill streamdetails with the "direct" attribute it means we're not going to intercept with the stream but just feed the url into ffmpeg directly and let that deal with it. we do that for mpeg dash streams for example and also for local files.

You can set the streamdetails.streamtitle if you want for the updated track details from that callback. But those are small finetune details later. First focus on getting playback working. I have no way to test this to know what is going on but its getting harder to follow now that we're also spinning up a webserver to grab the audio bits. I still dont get why we cant just replicate what that webserver is doing and do that directly but for now its maybe good to use this at starting point.

marcelveldt:
OK, so I had a quick look and it looks like there's a lot more needed to get this working due to the whole HLS stuff.
Basically you want to replicate what the xsm-player is doing: https://github.com/AngellusMortis/sxm-player/blob/master/sxm_player/workers/base.py

That is, if you want to feed the audio chunks yourself (in the get_audio_stream) but for now I think its totally fine to use that webserver wrapper for it. 

Unfortunately I cant this myself (maybe I can test with a VPN?!) otherwise I could maybe help you trace the issue.

From what I can see it should work fine, if that m3u8 playlist is also working in VLC so maybe this is ffmpeg related.
Can you test the streaming with the loglevel for the server set to debug ? That will also print all traces from ffmpeg to teh console

marcelveldt:
Sorry for the mega late reply! We had some serious bugs that needed to taken care of.
In the meanwhile I've built a HLS playlist parser into the code so we can now give this another test again.

OzGav:
hey @scottt732 just wondering if you have made any more progress?

bs42:
Here is a different project that I use with my AmpliPi, it's a more recently written interface to SiriusXM, perhaps there's something useful in it that would help with this:

[https://github.com/vszander/Amplipi/blob/main/python/sxm.py](url)

marcelveldt:
Superseded by #1730

PR #991, opened 2024-01-06
Fixes bug where sanpcast does not work

SantiagoSotoC:
Also note that the user can change the default snapcast configuration, causing it to malfunction. 
Add prefix to dynamically created streams.


marcelveldt (approved):
lgtm

PR #1000, opened 2024-01-15
Fixes for plex connection

jozefKruszynski:
I labelled it as breaking change as users will definitely need to start over to ensure a smooth transition.
However, I don't see this as an issue, as I guess for most people it seems to not work at all at the moment.

The myplexaccount.ping() function seems to be all that is required for the refresh of the token
https://github.com/pkkid/python-plexapi/pull/1271


marcelveldt (dismissed):
lgtm

jozefKruszynski:
@marcelveldt Can you double check the new update, I added the GDM discovery.

marcelveldt (approved):

PR #1006, opened 2024-01-18
Add Open Subsonic MusicProvider

khers:
This allows any Open Subsonic project (Goinc, pysonic, Airsonic-Advance, etc) to be configured as a music provider. Podcasts are added as artist  tracks but not imported by album.
 
This provider uses the py-opensonic module to communicate with the  *sonic server.


khers:
I could make an album for the podcast, but I found it strange to see my podcasts in with my albums when I use the Album view in the front end. Podcast channels show up in the artist view and episodes in the track view.

I personally think they need their own top level element in the UI and should be handled differently because they are very different things. I have podcasts that I would say fit the album model (think entire series like Mike Duncan's Revolutions where the content will remain relevant). But I also have Podcasts that don't (ones that are current news, politics, or others) and for these I really don't want episodes to hang around too long after I have listened.

khers:
I need to add a new config option that allows a user to disable podcasts (not all server implementations support them) and this is causing a failure for Navidrome users.

marcelveldt:
OK, so if I understand you correctly I think we should leave out podcasts completely for now from the sync. You can just return them in the browse function if you'd like. So at least they are not read into the database.

Next step would be that I introduce a new MediaType for podcasts but I need to think about it and see some examples.
I'll have a look how they're handled in apps like Spotify as well.

khers:
Okay, I have removed it from the indexing. It is not clear to me how to expose them in the browse feature without making them available to the indexing.

I don't want to block this PR on podcast support. I have left the infrastructure in place so I can make use of it when podcast handling is ready in Mass.

marcelveldt:
I will look into adding podcast support natively soon, I have already looked at the datamodels from both Spotify and Subsonic and I think its good to give them their own dedicated place instead of hacking them in the artist/album/track structure. So please leave the method indeed in the code to retrieve the info as we will need that later.

I will quickly push some fixes to the PR to fix the blocking IO issue in the streamer and then its ready to go to release to the public, nice work!

khers:
I will keep an eye out for the Podcast changes and get working on them when it's ready. Thanks for all your help!

marcelveldt:
I pushed the changes for audio playback, can you quickly test if it works ?

khers:
Yup, it is still working. Thanks!

marcelveldt (approved):

marcelveldt:
Merged! So it will be in the next version of the server, great work !

PR #1007, opened 2024-01-19
Add Jellyfin Music Provider

lokiberra:


marcelveldt:
Please make sure that you lint the files with pre-commit (see our development notes).
Feature wise this looks pretty solid, especially for a first iteration!

Press the "ready for review" button once you are ready for a final review, thanks!

marcelveldt:
Can you have another look? A lot more issues are discovered

marcelveldt:
I've rebased it - now let's see what it finds

lokiberra:
I focused on fixing the issues found by pre-commit in my files, I didn't touch the other ones.

marcelveldt:
Still linting errors are found, can you have another look ?

music_assistant/.vscode/launch.json
marcelveldt, line 1: this file may not be part of the PR

music_assistant/server/controllers/media/albums.py
```diff
@@ -292,7 +292,6 @@ async def _get_provider_album_tracks(
         items = []
         for track in await prov.get_album_tracks(item_id):
             assert isinstance(track, AlbumTrack)
-            assert track.track_number
```
marcelveldt, line 295: this is not allowed

marcelveldt (changes requested):
just 2 small leftover comments and then this is good to go!

music_assistant/server/controllers/media/albums.py
marcelveldt: btw: this assert is here for the reason that we often make this mistake

lokiberra: This was removed per our discussion on Discord because of 0 numbered pre tracks. Before I removed this I was setting the 0 tracks to 99 but that makes them out of order and there could be the case that there is a pre-track (0) and over 98 tracks.

https://discord.com/channels/753947050995089438/1083871908317102150/1198446101376802817

music_assistant/.vscode/launch.json
marcelveldt: I've removed the file from the PR

lokiberra:
Thanks for the help! Happy to contribute to this great project

PR #1009, opened 2024-01-19
Fix volume level calculation in Snapcast Provider

SantiagoSotoC:


marcelveldt:
sorry this is already fixed in an upcoming PR

PR #1056, opened 2024-01-31
Subsonic: Implement features required by Nextcloud Music support

perillamint:
This PR implements features/fallback support required for https://github.com/music-assistant/hass-music-assistant/issues/1856


music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -495,7 +495,11 @@ async def get_album(self, prov_album_id: str) -> Album:
         """Return the requested Album."""
         try:
             sonic_album: SonicAlbum = await self._run_async(self._conn.getAlbum, prov_album_id)
-            sonic_info = await self._run_async(self._conn.getAlbumInfo2, aid=prov_album_id)
+            try:
+                sonic_info = await self._run_async(self._conn.getAlbumInfo2, aid=prov_album_id)
+            except KeyError:
```
khers, line 500: The 'albumInfo' field in the response from the server is required (see: https://opensubsonic.netlify.app/docs/endpoints/getalbuminfo2/). If the server does not implement that endpoint it should return an error saying so.

On top of accommodating it here, please open a bug with the Nextcloud devs (if they claim to support Open Subsonic) about the missing item in the response.

khers (approved):
This looks good to me.

music_assistant/server/providers/opensubsonic/sonic_provider.py
perillamint: I can confirm MASS works without the workaround patch after monkey-patching my Nextcloud Music.

It seems they return `albumInfo2` instead of `albumInfo`.

https://github.com/owncloud/music/blob/master/lib/Controller/SubsonicController.php#L1524
https://github.com/owncloud/music/blob/master/lib/Controller/SubsonicController.php#L327

Also opened issue on the Owncloud/Nextcloud Music repository.

https://github.com/owncloud/music/issues/1123

perillamint: Reverting b952f26f9c0502167fdaebe4d434066e430ec424

khers:
The fall to the legacy auth is good to go, thanks for the contribution!

marcelveldt (dismissed):

marcelveldt:
Please note the lint errors

perillamint:
Sorry for the late reply. @marcelveldt 

Fixed everything and tested using `pre-commit`.

marcelveldt (approved):

PR #1070, opened 2024-02-08
Reconfigure linting,testing and formatting

marcelveldt:
Set-up (pre-commit) linting with a config that suits a 2024 python project.
Tooling gets updated using regular dependabot updates and each tool runs in the venv on execution.

Now, because the linter is setup including pylint and mypy and is more strict in general it finds massive issues.
We should perhaps go ahead and disable tools first to enable them one-by-one later ?


.github/workflows/test.yml
```diff
@@ -53,4 +53,4 @@ jobs:
           python -m pip install --upgrade pip build setuptools
           pip install .[server] .[test]
       - name: Pytest
-        run: pytest --durations 10 --cov-report term-missing --cov=music_assistant --cov-report=xml tests/
+        run: pytest --durations 10 --cov-report term-missing --cov=matter_server --cov-report=xml tests/
```
jozefKruszynski, line 56: Is this one intentional?

marcelveldt: copy paste issue :-)

PR #1086, opened 2024-02-18
Fix Airplay playback on docker/haos installs

marcelveldt:


andreasbrett:
Only macos binary has execution permissions. On Linux aarch64 I had to "docker exec" into my container and chmod a+x the cliraop binaries for AirPlay to work.

marcelveldt:
Ah crap, been too hasty building the binaries. But it did work after setting the execution bit then?

andreasbrett:
Yes, AirPlay is working fine

PR #1137, opened 2024-03-13
Subsonic: Check for transcodeOffset support on startup

khers:
If this support is present, then we can seek using the stream endpoint timeOffset argument. If not we need to fall back to having the player provider do it.

### DIFF
--- a/music_assistant/server/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/server/providers/opensubsonic/sonic_provider.py
@@ -65,6 +65,7 @@ class OpenSonicProvider(MusicProvider):
 
     _conn: SonicConnection = None
     _enable_podcasts: bool = True
+    _seek_support: bool = False
 
     async def handle_async_init(self) -> None:
         """Set up the music provider and test the connection."""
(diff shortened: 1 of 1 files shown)


khers:
Sorry for the PR noise, I was wondering why I could catch IOError and then I realized it was the linter.

music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -97,6 +98,16 @@ async def handle_async_init(self) -> None:
             )
             raise LoginFailed(msg) from e
         self._enable_podcasts = self.config.get_value(CONF_ENABLE_PODCASTS)
+        try:
+            ret = self._conn.getOpenSubsonicExtensions()
```
marcelveldt, line 102: sorry, this is blocking IO again, wrap it in an executor task

khers: Please excuse my Python async I/O ignorance, but even if we make this call run in a different thread/process, this thread will block on the result but it shouldn't hold the GIL while blocking. Does the async I/O library run its own thread scheduler and the await keyword is conditional reschedule or yield by this thread in a way that executing native library code or a system call is not?

marcelveldt: You may simply not run ANY blocking IO code in an eventloop. You are calling (blocking!) code from an non-async library, in those cases you need to wrap it into an executor. I suggest you read a bit into asyncio and its event loop.

Check for example this video: https://www.youtube.com/watch?v=M-UcUs7IMIM

music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -97,6 +98,16 @@ async def handle_async_init(self) -> None:
             )
             raise LoginFailed(msg) from e
         self._enable_podcasts = self.config.get_value(CONF_ENABLE_PODCASTS)
+        try:
+            ret = await self._run_async(self._conn.getOpenSubsonicExtensions)
+            extensions = ret["openSubsonicExtensions"]
+            for entry in extensions:
+                if entry["name"] == "transcodeOffset":
+                    self._seek_support = True
```
marcelveldt, line 106: This could use either a break or it can be replaced by a any statement

marcelveldt (approved):
Thanks!

PR #1148, opened 2024-03-17
Announce support (TTS with automatic resume)

marcelveldt:
Full support of the announce feature:

- If music playing it will be temporary interrupted and restored after the announcement
- will power on the player if needed and off if needed afterwards
- increases the volume slightly while playing the announcement
- prefers native approaches where possible (for example on sonos)
- works with all players
- For TTS announcements, supports prepending it with a "pre announcement sound"

Requires new HA integration as well - breaking changes!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ venv/
 *.spec
 .history
 .idea
+.coverage
--- a/music_assistant/client/players.py
+++ b/music_assistant/client/players.py
@@ -121,6 +121,16 @@ class Players:
         """
         await self.client.send_command("players/cmd/unsync", player_id=player_id)
 
+    async def play_announcement(
+        self,
+        player_id: str,
+        url: str,
+    ) -> None:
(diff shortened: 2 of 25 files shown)


arretx:
Two questions about this.

1.  I use PicorePlayer as my primary sound effects player but I also use a Monoprice 6-zone amplifier which creates additional media_players.  Will the volume boost affect all media players or just the player that generates the audio file?
2. "Requires new HA Integration." Is there something new here that we need to ensure is in place?

arretx:
Never mind on question 2, I see that HACS has a new update as well.

hmorais29:
Hi all, 
good to see that the announcement is back, thanks for it, great work.
quick question please, can one set the volume increase for the announcement like it used to be in the past (maybe I'm just missing where to configure it)

marcelveldt:
Don't post comments on closed issues or merged PRs, create an issue report, thanks!

PR #1155, opened 2024-03-19
Jellyfin Transcoding Fix

tesnos6921:
Jellyfin's universal audio stream API will attempt to transcode any media to mp3 if the application does not specify what containers it supports [1].  If Music Assistant knows that the file in Jellyfin is of a certain container type, and thus expects an audio stream of that type, there may be a playback error when Jellyfin delivers the data in mp3 format instead.  This PR adds information about the containers supported by FFmpeg to the universal audio stream API call so that Jellyfin does not attempt to transcode unnecessarily.

[1] https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/UniversalAudioController.cs/#L269

### DIFF
--- a/music_assistant/server/providers/jellyfin/__init__.py
+++ b/music_assistant/server/providers/jellyfin/__init__.py
@@ -812,7 +812,7 @@ class JellyfinProvider(MusicProvider):
         self, streamdetails: StreamDetails, seek_position: int = 0
     ) -> AsyncGenerator[bytes, None]:
         """Return the audio stream for the provider item."""
-        url = API.audio_url(self._jellyfin_server.jellyfin, streamdetails.item_id, "ogg,flac,mp3,aac,mpeg,alac,wav,aiff,wma,m4a,m4b,dsf,opus,wv")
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/jellyfin/__init__.py
```diff
@@ -812,7 +812,7 @@ async def get_audio_stream(
         self, streamdetails: StreamDetails, seek_position: int = 0
     ) -> AsyncGenerator[bytes, None]:
         """Return the audio stream for the provider item."""
-        url = API.audio_url(self._jellyfin_server.jellyfin, streamdetails.item_id)
+        url = API.audio_url(self._jellyfin_server.jellyfin, streamdetails.item_id, "ogg,flac,mp3,aac,mpeg,alac,wav,aiff,wma,m4a,m4b,dsf,opus,wv")
```
marcelveldt, line 815: maybe move that into a module level constant ?

marcelveldt (approved):
Thanks!

PR #1160, opened 2024-03-21
Add support for Dev Containers/Codespaces

robin-thoni:
This adds support for Dev Containers/Codespaces to provide a supported development environment

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -11,4 +11,7 @@ apt-get install -yq --no-install-recommends \
   ;
 apt-get clean
 rm -rf /var/lib/apt/lists/*
+pip install -U \
+  pre-commit \
+  ;
 EOF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+#syntax=docker/dockerfile:1.6
+
+FROM mcr.microsoft.com/vscode/devcontainers/python:3.12
+
+RUN <<EOF
+set -x
+set -e
+apt-get update
+apt-get install -yq --no-install-recommends \
+  ffmpeg \
+  ;
+apt-get clean
+rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


.devcontainer/post-create.sh
```diff
@@ -0,0 +1,13 @@
+#! /usr/bin/env bash
+
+pre-commit install
```
marcelveldt, line 3:
```suggestion
```

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ ! -d .venv ]
+then
+  python -m venv .venv
+fi
+
+source .venv/bin/activate
+
+pip install -e .[test]
+pip install -r requirements_all.txt
```
marcelveldt, line 13:
```suggestion
pip install -r requirements_all.txt
pre-commit install
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+set -e
+apt-get update
+apt-get install -yq --no-install-recommends \
+  ffmpeg \
+  ;
+apt-get clean
+rm -rf /var/lib/apt/lists/*
+pip install -U \
+  pre-commit \
+  ;
```
marcelveldt, line 16: pre-commit is installed with the test requirements

robin-thoni: Doesn't it need to be installed globally, rather than in the venv? That was the point of this implementation. If not, I'll commit your suggestions

marcelveldt: nope, in the venv.

robin-thoni: Done

robin-thoni:
I just re-validated the creation from scratch. Everything's ok. Ready to be merged if you're happy with it :)

marcelveldt (approved):
Nice, thanks!

PR #1161, opened 2024-03-21
Add support for Spotify Liked Songs

robin-thoni:
Spotify `Liked Songs` is a special feature that is not exposed as a standard playlist. Instead, it has its own endpoints:
- https://developer.spotify.com/documentation/web-api/reference/get-users-saved-tracks
- https://developer.spotify.com/documentation/web-api/reference/save-tracks-user
- https://developer.spotify.com/documentation/web-api/reference/remove-tracks-user
- https://developer.spotify.com/documentation/web-api/reference/check-users-saved-tracks

This PR exposes it as a standard playlist to MASS:

image

image

image

As I am running my dev env in Codespace (no Players available), I couldn't check (yet) that the playlist is actually playing, thus the Draft status.

There's also a dependency on #1160


music_assistant/server/providers/spotify/__init__.py
```diff
@@ -229,8 +229,35 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
...
 
+    async def _get_liked_songs_playlist(self) -> Playlist:
+        liked_songs = Playlist(
+            item_id="tracks",
+            provider=self.domain,
+            name="Liked Songs",  # TODO to be translated
+            owner="Me",  # TODO Get logged in user display name
+            provider_mappings={
+                ProviderMapping(
+                    item_id="tracks",
```
marcelveldt, line 240: please make this a constant (aso I think liked_songs is a better name)

music_assistant/server/providers/spotify/__init__.py
```diff
@@ -252,8 +279,11 @@ async def get_track(self, prov_track_id) -> Track:
 
     async def get_playlist(self, prov_playlist_id) -> Playlist:
         """Get full playlist details by id."""
-        playlist_obj = await self._get_data(f"playlists/{prov_playlist_id}")
-        return await self._parse_playlist(playlist_obj)
+        if prov_playlist_id == "tracks":
```
marcelveldt, line 282: use the constant here again

music_assistant/server/providers/spotify/__init__.py
```diff
@@ -252,8 +279,11 @@ async def get_track(self, prov_track_id) -> Track:
...
+            return await self._get_liked_songs_playlist()
+        else:
```
marcelveldt, line 284: no else after return needed

music_assistant/server/providers/spotify/__init__.py
```diff
@@ -229,8 +229,35 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
...
+                    provider_domain=self.domain,
+                    provider_instance=self.instance_id,
+                    url="https://open.spotify.com/collection/tracks",
+                )
+            },
+        )
+
+        liked_songs.is_editable = False  # TODO Editing requires special endpoints
+
+        liked_songs.metadata.images = [
```
marcelveldt, line 250: don't forget to set the metadata's checksum field otherwise it wont be refreshed on updates.

marcelveldt:
Nice contribution! I left a few very small comments

music_assistant/server/providers/spotify/__init__.py
robin-thoni: Makes me think: does it need to be globally unique? What if I link 2 different Spotify accounts?

music_assistant/server/providers/spotify/__init__.py
robin-thoni: Liked Songs do not have anything like `snapshot_id` in regular playlists. It's really just a plain list of tracks. Checking how other providers are doing it and it's applicable here

marcelveldt: ah in that case just assign a random value like time.time() or a random integer or whatever to invalidate the cache

music_assistant/server/providers/spotify/__init__.py
marcelveldt: In that case append the instance id

marcelveldt: In that case append the instance id

music_assistant/server/providers/spotify/__init__.py
robin-thoni: Done

robin-thoni:
Playback is ok.
However, for some reason, it shows the provider URL twice:
image

robin-thoni:
Hmm, nevermind, I'm pretty sure this is due to renaming `tracks` to `liked_songs`. I just deleted the cache and library, and rebuild the library from scratch to check. Will update.

music_assistant/server/providers/spotify/__init__.py
robin-thoni: Done

robin-thoni:
The provider details issue was indeed due to the renaming and inconsistencies in the DB. No issue here.

That looks good to me. If it's ok to you, you just need to merge @1160 first and I'll rebase this one on `main`

marcelveldt (approved):
Nice work!

PR #1174, opened 2024-03-24
Various fixes for Tidal

jozefKruszynski:


andreasbrett:
Is there a reason to use http and not https for tidal.com? The connection will always be upgraded to https for tidal after all, so why add this unnecessary step?

jozefKruszynski:
No reason, somehow I just never noticed that I http rather than https. One of those times when you can't see the wood for trees

marcelveldt (approved):

PR #1200, opened 2024-04-03
Fixes Spotify API for library Add/Remove of Artist

sprocket-9:
`type=artist` needs to be a URL parameter.


marcelveldt (approved):
Thanks!

PR #1202, opened 2024-04-04
Add videos to YTM search results

MarvinSchenkel:
Make sure to include videos in the search results as 'tracks'. This allows people to search for non-music content like desk concerts, kids music etc.


marcelveldt (approved):
please test this for unexpected side effects

PR #1204, opened 2024-04-04
Tidal: Fix retrieval of images for media items

jozefKruszynski:


marcelveldt (changes requested):
This is not acceptable as it will results in a huge resource hit due to the separate calls.

marcelveldt (approved):
fixed the code so no api calls are needed at all, resulting in a massive speed bump as a bonus as well as working images.

PR #1207, opened 2024-04-06
Ensure favorited item is added to provider library

sprocket-9:
Hey @marcelveldt , adding an item to favourites in MA is adding it to local library, but not to the Spotify library so I've taken a stab at a fix here - is this an acceptable fix?

I've seen bug reports and mentions of this from Spotify users, but from what I can tell this can't have worked for any provider, yet where are the bug reports?  So maybe I'm missing something...

### DIFF
--- a/music_assistant/server/controllers/music.py
+++ b/music_assistant/server/controllers/music.py
@@ -345,6 +345,14 @@ class MusicController(CoreController):
         """Add an item to the favorites."""
         if isinstance(item, str):
             item = await self.get_item_by_uri(item)
+        # ensure item is added to streaming provider library
+        prov_controller = self.mass.get_provider(item.provider)
+        if (
+            prov_controller
+            and prov_controller.is_streaming_provider
+            and prov_controller.library_edit_supported(item.media_type)
+        ):
+            await prov_controller.library_add(item.item_id, item.media_type)
         # make sure we have a full library item
         # a favorite must always be in the library
         full_item = await self.get_item(
@@ -360,14 +368,6 @@ class MusicController(CoreController):
             full_item.item_id,
             True,
         )
-        # ensure item is added to provider library
-        for provider_mapping in full_item.provider_mappings:
-            prov_controller = self.mass.get_provider(provider_mapping.provider_instance)
-            if prov_controller and prov_controller.library_edit_supported(full_item.media_type):
-                with suppress(NotImplementedError):
-                    await prov_controller.library_add(
-                        provider_mapping.item_id, full_item.media_type
-                    )
 
     @api_command("music/favorites/remove_item")
     async def remove_item_from_favorites(
--- a/music_assistant/server/controllers/music.py
+++ b/music_assistant/server/controllers/music.py
@@ -346,13 +346,12 @@ class MusicController(CoreController):
         if isinstance(item, str):
             item = await self.get_item_by_uri(item)
         # ensure item is added to streaming provider library
-        prov_controller = self.mass.get_provider(item.provider)
         if (
-            prov_controller
-            and prov_controller.is_streaming_provider
-            and prov_controller.library_edit_supported(item.media_type)
+            (provider := self.mass.get_provider(item.provider))
+            and provider.is_streaming_provider
+            and provider.library_edit_supported(item.media_type)
         ):
-            await prov_controller.library_add(item.item_id, item.media_type)
+            await provider.library_add(item.item_id, item.media_type)
         # make sure we have a full library item
         # a favorite must always be in the library
         full_item = await self.get_item(
--- a/music_assistant/server/controllers/music.py
+++ b/music_assistant/server/controllers/music.py
@@ -360,6 +360,14 @@ class MusicController(CoreController):
             full_item.item_id,
             True,
         )
+        # ensure item is added to provider library
+        for provider_mapping in full_item.provider_mappings:
+            prov_controller = self.mass.get_provider(provider_mapping.provider_instance)
+            if prov_controller and prov_controller.library_edit_supported(full_item.media_type):
+                with suppress(NotImplementedError):
+                    await prov_controller.library_add(
+                        provider_mapping.item_id, full_item.media_type
+                    )
 
     @api_command("music/favorites/remove_item")
     async def remove_item_from_favorites(


sprocket-9:
Ohhh just had a thought - maybe it actually is working afterall (at least since the spotify API fixes went it), it's just that favouriting adds the item to the local DB but the sync with remote providers doesn't happen immediately but after a set time or at startup?

sprocket-9:
No this isn't the case -  just tried favoriting a track in b131, then a spotify provider-> synchronize now and a MA restart.  The track doesn't appear in spotify.

marcelveldt:
This needs more thought, especially for multiple instances of a provider. I think we would need a toggle if MA is allowed to push back to the provider even.

music_assistant/server/controllers/music.py
```diff
@@ -360,6 +360,14 @@ async def add_item_to_favorites(
             full_item.item_id,
             True,
         )
+        # ensure item is added to provider library
+        for provider_mapping in full_item.provider_mappings:
+            prov_controller = self.mass.get_provider(provider_mapping.provider_instance)
+            if prov_controller and prov_controller.library_edit_supported(full_item.media_type):
+                with suppress(NotImplementedError):
```
marcelveldt, line 367: Do not silently suppress here, that leads to hidden bugs if a (new) provider forgets to implement it.
Use a check if the provider is a streaming provider and has library edit feature.

sprocket-9:
> This needs more thought, especially for multiple instances of a provider. I think we would need a toggle if MA is allowed to push back to the provider even

How about I do the streaming provider and library edit check earlier in the favorite request?  So check on `item` before it gets turned into a `full_item` then add `item` to streaming provider library.  

That way (I'm assuming!) we're dealing with an item from a single provider instance as clicked on  by the user in the search results, rather than `full_item` where `provider=library` and iterating through all its provider mappings to do library adds.

marcelveldt:
Well, we should do some more checks here. Imagine the situation where someone has multiple spotify accounts added to MA... What would be the expected behavior if you mark a track as favorite in MA ? Will it add to library on ALL instances of the spotify provider ? Or should we introduce some finegrained control over this ? That can be doen outside of this PR btw, it just occurred to me while I was looking at your PR and the problem description.

So I can imagine some additional config settings:
- Push library items from Provider to MA
- Push library items from MA to Provider

That way the user can control (per provider instance) if its a 2-way or one way sync and have extra instances of a provider become more passively involved. We can even take it a step further and define this option on a per media type base.

Anyways, its out of scope for this PR but this question will pop up immediately once we fix the bug with this PR.
I will have a look at this before 2.0 stable lands.

marcelveldt:
As for this PR... Just remove that Exception supression and its fine - the "is_library_edit_supported" handles it

music_assistant/server/controllers/music.py
marcelveldt, line 367:
```suggestion
```
Just remove this and fix the indentation and its all fixed already.
With the comment that the item will be added to the library of ALL INSTANCES of a provider and I can imagine people not wanting that

sprocket-9:
I agree, I can imagine people not wanting an item added to all instances.  It's odd behaviour and quite annoying to have items suddenly appear in a provider library you didn't explicitly ask MA to do.

Having those config settings to control this behaviour is definitely needed (or I can see a hard uninstall of MA happening from some users! 😅 ) but that's a feature request which doesn't sound too trivial to implement and could introduce further bugs at a time you're in feature freeze focusing on getting 2.0 out the door.

This isn't what I intended to achieve with this PR, as I'm testing with only a single provider I didn't fully understand the implications of multi-provider and multi-instances.  As I mentioned in the previous comment, looking at the initial `item` request rather than wait until it's in the MA library as a `full_item` will add it to the provider library of the provider the user clicked in the UI, which is what I intended.

I've been testing these changes out today and it's working ok, so I'll push the changes up for you to have a look at and if you're happy with it I suggest going with it and save a lot of extra work right now.  I'm sure you've got plenty on your plate already focusing on 2.0 ☺️

music_assistant/server/controllers/music.py
```diff
@@ -345,6 +345,14 @@ async def add_item_to_favorites(
...
         if isinstance(item, str):
             item = await self.get_item_by_uri(item)
+        # ensure item is added to streaming provider library
+        prov_controller = self.mass.get_provider(item.provider)
+        if (
+            prov_controller
+            and prov_controller.is_streaming_provider
+            and prov_controller.library_edit_supported(item.media_type)
+        ):
+            await prov_controller.library_add(item.item_id, item.media_type)
```
marcelveldt, line 355: Yes, this looks much better!
But... I'm afraid this will fail if the item is already in the library (provider will be 'library' in that case).
So, just to be safe I'd add a check for the provider to have a value, this is also needed for the type checker.

music_assistant/server/controllers/music.py
```diff
@@ -345,6 +345,14 @@ async def add_item_to_favorites(
         """Add an item to the favorites."""
         if isinstance(item, str):
             item = await self.get_item_by_uri(item)
+        # ensure item is added to streaming provider library
+        prov_controller = self.mass.get_provider(item.provider)
```
marcelveldt, line 349:
```suggestion
        if provider := self.mass.get_provider(item.provider):
```

music_assistant/server/controllers/music.py
marcelveldt: Ah I see you do exactly that (I just got distracted by the name prov_controller) - rename that to provider or provider_instance and its all good.

music_assistant/server/controllers/music.py
```diff
@@ -345,6 +345,14 @@ async def add_item_to_favorites(
...
+        if (
+            prov_controller
+            and prov_controller.is_streaming_provider
+            and prov_controller.library_edit_supported(item.media_type)
+        ):
```
marcelveldt, line 354:
```suggestion
        if (
            (provider := self.mass.get_provider(item.provider)):
            and provider.is_streaming_provider
            and provider.library_edit_supported(item.media_type)
        ):
```

marcelveldt: something like this (compacted it a bit with a walrus)

sprocket-9:
Ahh yeah the walrus, I keep forgetting to use that magnificent beast!  I've made those tweaks.

marcelveldt (approved):
Thanks!

marcelveldt:
Thanks for helping out! I'll add an item to the backlog to add those options for more fine grained control over the 2 way sync

PR #1214, opened 2024-04-09
Add support for using streamer provider publicly shareable URLs in search query

sprocket-9:
I often get Spotify URLs sent to me which I'd like to open directly in MA.  Right now I have to go into the Spotify app first to find out what it is, then search for the track in MA for playing.  It would be nice to use these URLs as is, so I've added support for using them in the MA global search.

Conversely I'd like to be able to copy the provider URL in MA so I can share it, but there's some CSS property making it unselectable and on a phone in portrait mode, the provider box contents wrap so the URL disappears, making it unclickable to go to Spotify app.  A button next to the URL to copy it to clipboard would be very nice!   Anyway that's a front end issue, so wrong place to put this I know ☺️ I'll put a feature request in for the new front end design.

Oh I've enabled the use of Qobuz URLs as they seem to use the same format as Spotify, but I'm not able to test Qobuz,

### DIFF
--- a/music_assistant/common/helpers/uri.py
+++ b/music_assistant/common/helpers/uri.py
@@ -1,12 +1,28 @@
 """Helpers for creating/parsing URI's."""
 
 import os
+import re
 
 from music_assistant.common.models.enums import MediaType
-from music_assistant.common.models.errors import MusicAssistantError
+from music_assistant.common.models.errors import InvalidProviderURI, MusicAssistantError
+
+base62_length22_id_pattern = re.compile(r"^[a-zA-Z0-9]{22}$")
 
 
-def parse_uri(uri: str) -> tuple[MediaType, str, str]:
+def valid_base62_length22(item_id) -> bool:
+    """Validate Spotify style ID."""
+    return bool(base62_length22_id_pattern.match(item_id))
+
+
+def valid_id(provider: str, item_id: str) -> bool:
+    """Validate Provider ID."""
+    if provider == "spotify":
+        return valid_base62_length22(item_id)
+    else:
+        return True
+
+
+def parse_uri(uri: str, validate_id: bool = False) -> tuple[MediaType, str, str]:
     """Try to parse URI to Mass identifiers.
 
     Returns Tuple: MediaType, provider_instance_id_or_domain, item_id
@@ -45,6 +61,10 @@ def parse_uri(uri: str) -> tuple[MediaType, str, str]:
     except (TypeError, AttributeError, ValueError, KeyError) as err:
         msg = f"Not a valid Music Assistant uri: {uri}"
         raise MusicAssistantError(msg) from err
+    if validate_id:
+        if not valid_id(provider_instance_id_or_domain, item_id):
+            msg = f"Invalid {provider_instance_id_or_domain} ID: {item_id}"
+            raise InvalidProviderURI(msg)
     return (media_type, provider_instance_id_or_domain, item_id)
 
 
--- a/music_assistant/common/helpers/uri.py
+++ b/music_assistant/common/helpers/uri.py
@@ -4,7 +4,7 @@ import os
 import re
 
 from music_assistant.common.models.enums import MediaType
-from music_assistant.common.models.errors import InvalidProviderURI, MusicAssistantError
+from music_assistant.common.models.errors import InvalidProviderID, InvalidProviderURI
 
 base62_length22_id_pattern = re.compile(r"^[a-zA-Z0-9]{22}$")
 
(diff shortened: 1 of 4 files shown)


music_assistant/server/controllers/music.py
```diff
@@ -157,6 +160,33 @@ async def search(
         :param media_types: A list of media_types to include.
         :param limit: number of items to return in the search (per type).
         """
+        # Check if the search query is a streaming provider public shareable URL
+        try:
+            media_type, provider_instance_id_or_domain, item_id = parse_uri(search_query)
```
marcelveldt, line 165: I think this helper util alone should be enough to parse a shareable URL. Calling get_item is heavy and overkill. We should improve the "parse_uri" if that doesn't already handle it. I know I added some support in there for share URL's.

music_assistant/server/controllers/music.py
```diff
@@ -157,6 +160,33 @@ async def search(
...
+            pass
+        else:
+            if provider_instance_id_or_domain in PROVIDERS_WITH_SHAREABLE_URLS:
+                try:
+                    item = await self.get_item(
+                        media_type=media_type,
+                        item_id=item_id,
+                        provider_instance_id_or_domain=provider_instance_id_or_domain,
+                    )
+                except (MusicAssistantError, aiohttp.client_exceptions.ClientResponseError):
```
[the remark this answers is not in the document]
sprocket-9: get_item with a spotify ID missing a character (like if you miss the last character when copying a shared URL) returns:
aiohttp.client_exceptions.ClientResponseError: 400, message='Bad Request'

and with changing one character so it's a valid spotify ID length, but an unknown ID it returns:
aiohttp.client_exceptions.ClientResponseError: 502, message='Bad Gateway' which is kind of odd.

So it's get_data in the provider that needs modified to handle these two errors and return a MusicAssistantError derivative?  MediaNotFoundError sounds the most suitable I suppose?  However I remember I was getting 400 error codes returned when the add/remove API requests were incorrect.

I can't test Qobuz so I don't know what codes their API will return to these types of incorrect requests.

music_assistant/server/controllers/music.py
sprocket-9: I am using `parse_uri` to parse the url and get the media_type and the item_id from it.  I then use  these in `get_item` to actually retrieve the `MediaType` item from the provider, then stuff the item in a `SearchResults` for return.  So there's no actual searching going on here, I'm just using the search interface to present the "results".

marcelveldt: ah yeah right... So instead of doing a text search on name, you can just paste an uri or share link.

marcelveldt: Still, if you know for certain that the spotify id's should fit a common pattern, I think it would be good to improve the parse_uri helper to catch this as early as possible and save the overhead of the URL call. You could even raise the InvalidUri exception

sprocket-9: Good idea.  I've added validation for the spotify ID and others can be added later once URL search is enabled and provider ID format known.

music_assistant/server/controllers/music.py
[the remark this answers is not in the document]
sprocket-9: I've removed catching the aiohttp exceptions.  With the URI validator I'm now preventing an API call with an invalid Spotify ID occurring, so the 400 shouldn't happen. 

502 will still occur with a valid ID format but and ID which doesn't exist on spotify.  Wonder if this code is also returned for a spotify item which isn't available in your region.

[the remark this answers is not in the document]
marcelveldt: Yeah, let's fix that in the spotify provider

music_assistant/common/helpers/uri.py
```diff
@@ -45,6 +61,10 @@ def parse_uri(uri: str) -> tuple[MediaType, str, str]:
     except (TypeError, AttributeError, ValueError, KeyError) as err:
         msg = f"Not a valid Music Assistant uri: {uri}"
         raise MusicAssistantError(msg) from err
```
marcelveldt, line 63: maybe raise InvalidProviderURI here as well ? That way we dont have to catch the very generic MusicAssistantError anymore

music_assistant/server/controllers/music.py
```diff
@@ -157,6 +162,39 @@ async def search(
...
+            media_type, provider_instance_id_or_domain, item_id = parse_uri(
+                search_query, validate_id=True
+            )
+        except InvalidProviderURI as err:
+            self.logger.warning("%s", str(err))
+            return SearchResults()
+        except MusicAssistantError:
+            pass
+        else:
+            if provider_instance_id_or_domain in PROVIDERS_WITH_SHAREABLE_URLS:
```
marcelveldt, line 176: You could consider moving this block below the parse_uri and just do everything in one GO - catch InvalidProviderURI and let that pass (as it means there was no uri passed) - catch MusicAssistantError and log that one.

music_assistant/common/helpers/uri.py
sprocket-9: I think we still need two different named exceptions to catch from `parse_uri`:
1: This is not a URI: pass through to the regular search block
2: This is a URI but it's an invalid provider ID: log and return empty SearchResults() so no search is performed

They could be better named though.  How about:
1: Use InvalidProviderURI
2: Use InvalidProviderID

music_assistant/server/controllers/music.py
sprocket-9: Hmm yeah need to give this some thought.  If we do as I suggest above and create another exception class, we're now catching three with different handlers.  Moving this block inside the `try:` will look prettier, but will it aid readability when we come back to look at it later and now there's multiple things going on in the `try:` block and we need to work out which thing can raise which exception?  Especially given `get_item` is a multi-pathed beast.  It's up to you though, I'm happy to go with whatever you feel is best ☺️

music_assistant/common/helpers/uri.py
marcelveldt: Yes, that sounds good!

music_assistant/server/controllers/music.py
marcelveldt: Let's go with an iterative approach, fix the issue now and make it prettier later.

marcelveldt (approved):
Great!

PR #1220, opened 2024-04-12
Fix snapcast client_id

SantiagoSotoC:


marcelveldt (changes requested):
you should create a small helper method to get/calculate the player id

SantiagoSotoC:
Yes, that's why I created a draft. I wanted to see if it works in the dev addon, but I couldn't.

SantiagoSotoC:
I will test it well, when I know that everything works fine I mark the pr as ready for review.

marcelveldt:
> Yes, that's why I created a draft. I wanted to see if it works in the dev addon, but I couldn't.

the dev-addon can just checkout a branch you give it. for pr's you need to specify it a bit fancy, I believe you can specify: pr/1220

SantiagoSotoC:
I'll try it now thanks, when I see the volume I'll do it directly in the repo.

SantiagoSotoC:
This way of doing it will not work, some clients use "-" in the identifier (snapweb). I think the solution will be to create a dictionary and generate internal identifiers.

PR #1231, opened 2024-04-17
fix: Several minor fixes for deezer

micha91:
* don't use `seconds_skipped` from StreamDetails anymore
* avoid using `utc_now`
* use `cache_checksum` instead of `checksum` on MediaItemMetadata

### DIFF
--- a/music_assistant/server/providers/deezer/__init__.py
+++ b/music_assistant/server/providers/deezer/__init__.py
@@ -1,6 +1,5 @@
 """Deezer music provider support for MusicAssistant."""
 
-import datetime
 import hashlib
 import uuid
 from asyncio import TaskGroup
@@ -14,6 +13,7 @@ from aiohttp import ClientSession, ClientTimeout
 from Crypto.Cipher import Blowfish
 from deezer import exceptions as deezer_exceptions
 
+from music_assistant.common.helpers.datetime import utc_timestamp
 from music_assistant.common.models.config_entries import (
     ConfigEntry,
     ConfigValueType,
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/deezer/__init__.py
```diff
@@ -473,7 +474,7 @@ async def get_audio_stream(
             headers["Range"] = f"bytes={skip_bytes}-"
 
         buffer = bytearray()
-        streamdetails.data["start_ts"] = datetime.datetime.utcnow().timestamp()
+        streamdetails.data["start_ts"] = datetime.datetime.now().timestamp()
```
marcelveldt, line 477: this will break soon, better use our utcnow helper from utils.datetime

music_assistant/server/providers/deezer/__init__.py
```diff
@@ -618,7 +619,7 @@ def parse_playlist(self, playlist: deezer.Playlist) -> Playlist:
                         remotely_accessible=True,
                     )
                 ],
-                checksum=playlist.checksum,
+                cache_checksum=playlist.checksum,
```
marcelveldt, line 622: oops sorry, missed this one

music_assistant/server/providers/deezer/__init__.py
micha91: Why would this break soon - I thought `utcnow` was deprecated only? 
However it's cleaner to use UTC timestamps anyways, I changed it 👍

marcelveldt: datetime.now() without timezone is deprecated and will be remove din python 13

marcelveldt (approved):
Thanks!

PR #1253, opened 2024-04-25
Better handling of Jellyfin metadata missing

lokiberra:
- Added handling of tracks where the Album key is missing. Uses the parent id to get the album information, if that is not present the track should load but have an unknown album
- Added handling when the item type is MusicArtists instead of MusicArtist.


music_assistant/server/providers/jellyfin/const.py
```diff
@@ -41,6 +41,7 @@
 
 ITEM_TYPE_ALBUM: Final = "MusicAlbum"
 ITEM_TYPE_ARTIST: Final = "MusicArtist"
+ITEM_TYPE_ARTIST_2: Final = "MusicArtists"
```
marcelveldt, line 44: should the name of this constant not just be ITEM_TYPE_ARTISTS or even ITEM_TYPE_MUSICARTISTS ?

lokiberra: I've update it now to ITEM_TYPE_MUSICARTISTS

PR #1264, opened 2024-04-30
Open Subsonic: Fix MediaImageItem constructor call

khers:
The MediaImageItem construtor now requires a provider key word argument.

Previous versions of MA would avoid calling resolve_image if the path was None or empty. This seems to have changed and is causing errors in library syncing. The function now checks if path is empty and returns an empty list if so.

### DIFF
--- a/music_assistant/server/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/server/providers/opensubsonic/sonic_provider.py
@@ -328,7 +328,10 @@ class OpenSonicProvider(MusicProvider):
             if sonic_info.small_url:
                 album.metadata.images.append(
                     MediaItemImage(
-                        type=ImageType.THUMB, path=sonic_info.small_url, remotely_accessible=False
+                        type=ImageType.THUMB,
+                        path=sonic_info.small_url,
+                        remotely_accessible=False,
+                        provider=self.instance_id,
                     )
                 )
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -427,6 +430,11 @@ async def _run_async(self, call: Callable, *args, **kwargs):
 
     async def resolve_image(self, path: str) -> bytes:
         """Return the image."""
+        # We assign None to the image ID if we don't have one and this used to be okay,
+        # however newer MA seems to be unhappy with this and will still ask us to resolve
+        # an image without a path.
```
marcelveldt, line 435: An image without a path ?! That is not good. If an item has no image, just dont set an image, dont set an empty path.

marcelveldt (changes requested):

music_assistant/server/providers/opensubsonic/sonic_provider.py
khers: I think we can actually drop this entirely, the empty path was a knock on from the ProviderMapping bug because I didn't see the failure from this method once that bug was fixed. I will drop this change from the patch and update.

marcelveldt (approved):

PR #1277, opened 2024-05-07
lower throttler log level to info

OzGav:
As a suggestion I am thinking that just doing a retry doesn't need to be a warning. A warning is something that I think should be investigated. So lower that to an info and users will still be alerted if the retries are exhausted.


marcelveldt (approved):
thanks!

PR #1281, opened 2024-05-07
Fix: Deezer albums

arctixdev:
Fixes Deezer album track parsing by adding required disc number

### DIFF
--- a/music_assistant/server/providers/deezer/__init__.py
+++ b/music_assistant/server/providers/deezer/__init__.py
@@ -664,6 +664,7 @@ class DeezerProvider(MusicProvider):  # pylint: disable=W0223
             metadata=self.parse_metadata_track(track=track),
             track_number=position,
             position=position,
+            disc_number=getattr(track, "disk_number", 1),
         )
         if isrc := getattr(track, "isrc", None):
             item.external_ids.add((ExternalID.ISRC, isrc))


music_assistant/server/providers/deezer/__init__.py
```diff
@@ -664,6 +664,7 @@ def parse_track(self, track: deezer.Track, user_country: str, position: int = 0)
             metadata=self.parse_metadata_track(track=track),
             track_number=position,
             position=position,
+            disc_number=getattr(track, "disk_number", 1),
```
marcelveldt, line 667: is it really disk number (with a k) and not disc number ?

arctixdev: Yup, i wondered as well :thinking: https://deezer-python.readthedocs.io/en/stable/api_reference/resources/track.html#deezer.Track.disk_number at least according to the code and the docs. But could be a typo in the original library. Let me double check

arctixdev: Nope, thats also how it is written in the official deezer api documentation: https://developers.deezer.com/api/track

marcelveldt: ok, good that you double checked, all good then!

marcelveldt (approved):

PR #1304, opened 2024-05-14
Subsonic: Set Track.disc_number to 0 for now

khers:
The standard for Open Subsonic is not complete for Songs so we don't know if this field will be supported. Once the standard is ready we can readress this.

Fixes: https://github.com/music-assistant/hass-music-assistant/issues/2247


marcelveldt (approved):
Thanks!

PR #1311, opened 2024-05-16
Fixes the spotify json decode mime-type errors

sprocket-9:
`_post_data` does return json from playlist ops so this is untouched.


marcelveldt (approved):
Thanks for fixing my sloppy mistake ;-)

PR #1314, opened 2024-05-18
Fix owner id mismatch bug in Spotify create playlist API

sprocket-9:
`pl = create_playlist("x") # pl['owner']['id'] not my spotify id`
`In get_playlist(pl["id"]) method: playlist_obj['owner']['id']  my spotify id`

Looks to me there's a bug in the create playlist API which is returning the incorrect owner id, but needs confirmation you're seeing the same thing and not something odd with my account.

Have fixed it by munging the returned playlist data rather than just doing a get_playlist(new_playlist['id']) in create_playlist, in order to save an api call and know when it's fixed.

If it is a confirmed bug then spotify should be notified, but from what I can see from forum posts they're not speedy with responding to API issues!


marcelveldt (approved):
Creative solution ;-)
Nice, thanks!

PR #1322, opened 2024-05-29
Fix snapcast players ids

SantiagoSotoC:
Fixes the following bug [https://github.com/music-assistant/hass-music-assistant/issues/2108](https://github.com/music-assistant/hass-music-assistant/issues/2108)

The repair involves changing the player_ids so there will remain old instances of snapclients


marcelveldt (approved):
Looks fine to me

marcelveldt:
pinging @OzGav here about the player id's that are going to change

PR #1324, opened 2024-06-01
Refactor SnapCastProvider cmd_volume_set method

SantiagoSotoC:
The method in the snapserver class does not call the callback, so the volume was not updated correctly. 
Also the muting logic is eliminated because it is not correct that if the volume of a group is changed, a muted client will sound again.


SantiagoSotoC:
Do not merge yet, there is an error, the library is  broken.

SantiagoSotoC:
I think I found the right method now, I will test it later in the day and report back if it works well.

marcelveldt:
And? any updates?

SantiagoSotoC:
It's not perfect, but it works better than the current one, I'm talking to the people at the library to make some changes, if it doesn't work out I'll have to see if I can do it another way, but this is an improvement.

SantiagoSotoC:
This is a substantial improvement, changing the library takes time and I don't know if it will be possible in the short term.

marcelveldt (approved):

PR #1318, opened 2024-05-22
Turn off Display and visualization by default

cociweb:
Turn off Display support and use None type visualization by default. This PR tries to resolves the periodical "unhandled grfe" messages in the player's log as defined in music-assistant/hass-music-assistant/issues/2118

PR #1331, opened 2024-06-07
Fix None value for Display and visualization

cociweb:
Based on an untested commentary provided in the review (https://github.com/music-assistant/server/pull/1318), the variable was changed from None.value to None. This suggestion has now been reverted and retested. The value of the variable has been set to None.value again.

This PR resolve: https://github.com/music-assistant/hass-music-assistant/issues/2420


marcelveldt:
This is not good. Why is this needed ? Please explain ? Isnt this a stringenum ?

marcelveldt:
I fixed this in the config entry handling code so we can handle other cases as well for enum values.
Thanks anyways!

PR #1335, opened 2024-06-09
Two updates to the Subsonic provider

khers:
The first commit adds the ability to edit playlists from MA, partially addressing the feature request in https://github.com/orgs/music-assistant/discussions/2358 favorite syncing will come later.

The second commit explicitly sets the favorite field in the album and artists constructors to the value in the provider.

### DIFF
--- a/music_assistant/server/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/server/providers/opensubsonic/sonic_provider.py
@@ -244,6 +244,7 @@ class OpenSonicProvider(MusicProvider):
             item_id=sonic_artist.id,
             name=sonic_artist.name,
             provider=self.domain,
+            favorite=bool(sonic_artist.starred),
             provider_mappings={
                 ProviderMapping(
                     item_id=sonic_artist.id,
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -675,6 +684,38 @@ async def get_similar_tracks(self, prov_track_id: str, limit: int = 25) -> list[
...
+
+        Note that the configured user must own the playlist to edit this way.
+        """
+        try:
+            await self._run_async(
+                self._conn.updatePlaylist, lid=prov_playlist_id, songIdsToAdd=prov_track_ids
+            )
+        except SonicError:
+            msg = f"Failed to add songs to {prov_playlist_id}, try again later."
+            raise ResourceTemporarilyUnavailable(msg)
```
marcelveldt, line 703: Isn't this a bit of a strange error to raise on this action ?

ResourceTemporarilyUnavailable is more for rate limiters or other reason why an action can temporary not be performed

khers: I put a new series up that adds a ProviderPermissionDenied error and uses that instead.

marcelveldt (approved):
Nice work @khers !

PR #1338, opened 2024-06-10
Jellyfin: Don't fail entire sync if a single track or album have an invalid mbid

Jc2k:
This is a follow up to https://github.com/music-assistant/server/pull/1332, and handles InvalidDataError in some other cases to ensure a Jellyfin sync does not fail entirely for a single bad file.

(I've not seen these ones in the wild, but there is the same lack of validation on the Jellyfin side so these seem worth adding).


marcelveldt (approved):
LGTM!

PR #1325, opened 2024-06-02
Make Jellyfin provider compatible with Emby

kingy444:
I am currently trying to write the Emby integration using their native API (but getting some strange import issues and my time is limited). In the process of looking at that i have taken a quick look at the JellyFin integration.

JellyFin having forked from an old version of Emby still functions very much the same so with a few tweaks i have been able to get the JellyFin plugin to work while maintaining all JellyFin capabilities.

- [Icon and server display fix](https://github.com/music-assistant/server/pull/1325/commits/f0d2325d27f1fc43035e96f92e170668d35845b2) - Updated Icon to flow with the rest of the icons (removed black background) and removed title which was being tagged into the track screens for title attribute
- [Simplify looping](https://github.com/music-assistant/server/pull/1325/commits/4b93fbed3f8e926c0d332916a99e4b69cc167316) - Simply code clean up
- [Move to Artists Endpoint](https://github.com/music-assistant/server/pull/1325/commits/756eaa4e87a71cf945f9dcbaf2779b3678359d3b) - Needed to handle Emby artists better, and works perfectly with JellyFin. While the current method worked for both it was messy with Emby as Composers are considered an Artist too
- [Handle playlists for both providers](https://github.com/music-assistant/server/pull/1325/commits/f9d9cf8fdb3f6d3eb086c5e6718496312304e9ec) - just some IF logic to handle that JellyFin have improved their playlist functionality. Emby still present only audio playlists
- [Drop ParentId in favor of AlbumId](https://github.com/music-assistant/server/pull/1325/commits/f268d5e6f8034161b8ae736c17bccabb2263a8fc) - The main change required here was to use AlbumId instead of ParentId. Emby has diverged from ParentId relationship JellyFin inherited but the AlbumId relationship remains the same.
- [Filter on ArtistType](https://github.com/music-assistant/server/pull/1325/commits/5ebd2340392d5e12d15604284bf7a92da00e86ee) - Emby allows further filtering, Jellyfin ignores
- [Fix double & when property should be comma delim](https://github.com/music-assistant/server/pull/1325/commits/ea69c727d3ea2ff845eb27cf7a28ac3dbe5819db) - Was creating ``&IncludeItemTypes=MusicArtists&IncludeItemTypes=MusicArtist`` this should be ``&IncludeItemTypes=MusicArtists,MusicArtist``

[...]

PR #1339, opened 2024-06-10
Jellyfin: Fix regression importing track artist mapping

Jc2k:
AFAICT, https://github.com/music-assistant/server/pull/1325 introduced a regression in syncing Jellyfin tracks. After that merge the code started looking for ITEM_KEY_ALBUM_ID in the album metadata. It then tries to create an artist mapping with the ID of the album (rather than an artist id) and the "album artist" text from the album. This creates invalid data. Or it would, but an album doesn't actually have an AlbumId field at all.

This PR allows syncing to complete without exceptions again by using the "AlbumArtists" list from the album object:

```
    "AlbumArtists": [
        {
            "Name": "Papa Roach",
            "Id": "e439648e08ade14e27d5de48fa97c88e"
        }
    ],
```


marcelveldt (approved):
LGTM, thanks!

PR #1347, opened 2024-06-11
Fix snapcast player unsync when master off

SantiagoSotoC:


SantiagoSotoC:
It was helpful that you referred me to the player manager, I think I found the bug, the library callbacks work horrible and the value of synced_to remains None.

marcelveldt (approved):

PR #1350, opened 2024-06-12
Revert "Fix snapcast player unsync when master off"

SantiagoSotoC:
Introduces another bug, where only two clients can be synchronised.


marcelveldt (approved):

marcelveldt:
Looks like perhaps swapping target with the player ?
Target is the master player you want to sync a player to

player.synced_to --> should be set only on sync childs - set to the "master" it joined
player.group_childs --> should be a list of player ids part of the sync group (including the master) - can only be set on a sync master

SantiagoSotoC:
Yes, you're right I got confused, I'll send the bugfix as soon as possible

PR #1351, opened 2024-06-12
Apple Music: Fixed lots of metadata related issues

MarvinSchenkel:
This PR:
- Fixes wrong artist on compilation albums
- Fixes error when scrolling to the bottom of a playlist
- Adds safeguards for 504 errors
- Prefers catalog info over library info for playlists

### DIFF
--- a/music_assistant/server/providers/apple_music/__init__.py
+++ b/music_assistant/server/providers/apple_music/__init__.py
@@ -198,15 +198,29 @@ class AppleMusicProvider(MusicProvider):
     async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
         """Retrieve library tracks from the provider."""
         endpoint = "me/library/songs"
-        for item in await self._get_all_items(endpoint, include="artists,albums,catalog"):
-            if item and item["id"]:
+        song_catalog_ids = []
+        for item in await self._get_all_items(endpoint, include="catalog"):
+            if item and "catalog" in item["relationships"]:
+                song_catalog_ids.append(item["relationships"]["catalog"]["data"][0]["id"])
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/apple_music/__init__.py
```diff
@@ -237,7 +257,16 @@ async def get_album_tracks(self, prov_album_id) -> list[Track]:
         """Get all album tracks for given album id."""
         endpoint = f"catalog/{self._storefront}/albums/{prov_album_id}/tracks"
         response = await self._get_data(endpoint, include="artists")
-        return [self._parse_track(track) for track in response["data"] if track["id"]]
+        # Including albums results in a 504 error, so we need to fetch the album separately
```
marcelveldt, line 260: How does this solve that 504 error ?

MarvinSchenkel: `include="artists,albums"` messes with the API and throws a 504. This should be supported according to the docs and definitely works for other endpoints, but somehow crashes on the `/albums/xxx/tracks` endpoint. Using either `include="artists"` or `include="albums"` seems to work fine. Since we need both, I opted to retrieve the Album object once and set it on each individual Track

marcelveldt: Yeah, my point was that there was no inclusion of both in the call so strictly taken this was not broken ?

MarvinSchenkel: But both are required because tracks were missing an album, causing some weird behavior. To fix that, I needed to include both which threw a consistent 504, so I worked around that.

marcelveldt (approved):

PR #1357, opened 2024-06-13
Enforce line endings

Jc2k:
Noticed git diff output looked weird. The project has mixed line endings. This PR makes "ruff format" enforce line endings, and it fixes the files that aren't consistent with the rest of the project.


marcelveldt (approved):
Thanks!

PR #1362, opened 2024-06-14
Jellyfin: Drop all synchronous HTTP and use aiohttp instread

Jc2k:
Replacement library just implements the 7 or 8 API's the Jellyfin integration currently needs.

It deliberately tries to mirror the existing API, so that the switch can be reviewed more easily.

I think this fixes a few cases of calling blocking code on the event loop.

After it's merged I want to get mypy passing for the provider, and flush out the response schemas on the library side. I didn't want to make this change harder to review by doing that at the same time.

And i'll add a config bool for SSL verification (https://github.com/music-assistant/hass-music-assistant/issues/2025).

### DIFF
--- a/music_assistant/server/providers/jellyfin/__init__.py
+++ b/music_assistant/server/providers/jellyfin/__init__.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import asyncio
 import logging
 import mimetypes
 import socket
@@ -10,11 +9,10 @@ import uuid
 from asyncio import TaskGroup
 from typing import TYPE_CHECKING, Any
 
+from aiojellyfin import SessionConfiguration, Connection, authenticate_by_name
+
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, Callable, Coroutine
-
-from jellyfin_apiclient_python import JellyfinClient
-from jellyfin_apiclient_python.api import API
 
 from music_assistant.common.models.config_entries import (
     ConfigEntry,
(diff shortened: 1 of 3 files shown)


music_assistant/server/providers/jellyfin/__init__.py
```diff
@@ -153,41 +151,26 @@ async def get_config_entries(
 class JellyfinProvider(MusicProvider):
     """Provider for a jellyfin music library."""
 
-    # _jellyfin_server : JellyfinClient = None
-
     async def handle_async_init(self) -> None:
         """Initialize provider(instance) with given configuration."""
+        session_config = SessionConfiguration(
+            session=self.mass.http_session,
```
marcelveldt, line 157: I was thinking if we also need a separate aiohttp session at core level without ssl, just like HA which also has 2.
And then some provider config setting "do not verify ssl" which then uses that ssl-less http session

music_assistant/server/providers/jellyfin/manifest.json
```diff
@@ -4,7 +4,7 @@
   "name": "Jellyfin Media Server Library",
   "description": "Support for the Jellyfin streaming provider in Music Assistant.",
   "codeowners": ["@lokiberra"],
```
marcelveldt, line 6: maybe also put your username here @Jc2k ?

marcelveldt:
Nice work on this @Jc2k !
I leave final approval/review to @lokiberra as he's the maintainer of this provider.

music_assistant/server/providers/jellyfin/__init__.py
Jc2k: I've been lucky and not needed the unverified client session over there. Can look into what HA does and then do similar in a follow up?

marcelveldt: Yeah it was just an idea - there was also a request somewhere from somebody who asked if we can offer an "do not verify ssl" option on the home Assistant provider in MA as he is using a self signed cert.
Definitely for later (just thinking out loud) - now we just have a singe aiohttp session in the server

lokiberra (approved):

lokiberra (approved):

PR #1368, opened 2024-06-16
Fix a typing issue in the client.

jozefKruszynski:


marcelveldt:
Looks like your local ruff version is not the same as what we have on CI.

jozefKruszynski:
Fixed now.

marcelveldt (approved):

PR #1370, opened 2024-06-17
Force comparison of MBID

MelHarbour:
ISRC turns out to be a minefield of inconsistency. MBID is generally a good indicator for tracks, so if that's present on both tracks, default to using that.

### DIFF
--- a/music_assistant/server/helpers/compare.py
+++ b/music_assistant/server/helpers/compare.py
@@ -112,6 +112,9 @@ def compare_track(
     # return early on exact item_id match
     if compare_item_ids(base_item, compare_item):
         return True
+    # return early on MBID match
+    if base_item.mbid and compare_item.mbid:
+        return base_item.mbid == compare_item.mbid
     # return early on (un)matched external id
     external_id_match = compare_external_ids(base_item.external_ids, compare_item.external_ids)
     if external_id_match is not None:


MelHarbour:
See https://github.com/music-assistant/hass-music-assistant/issues/2316 for discussion

marcelveldt:
Nice, easy and clean. Maybe add a link to the issue report in the comment as a reference for the future why we have this early mbid check in the code

MelHarbour:
I did already in the comment just above yours!

MelHarbour:
> Nice, easy and clean. Maybe add a link to the issue report in the comment as a reference for the future why we have this early mbid check in the code

Done.

music_assistant/server/helpers/compare.py
```diff
@@ -112,6 +112,9 @@ def compare_track(
     # return early on exact item_id match
     if compare_item_ids(base_item, compare_item):
         return True
+    # return early on MBID match to avoid issues with ISRC (https://github.com/music-assistant/hass-music-assistant/issues/2316)
```
marcelveldt, line 115:
```suggestion
    # return early on MBID match to avoid issues with ISRC 
    # (https://github.com/music-assistant/hass-music-assistant/issues/2316)
```

PR #1371, opened 2024-06-17
Jellyfin: Typing improvements

Jc2k:
With this PR, mypy is clear for Jellyfin.

* Added missing type defs
* Switched a bunch of things like `dict[str,Any]` to types in aiojellyfin
* Fixed some missing `UniqueList`
* Simplified a few things for the sake of mypy
* Removed some unused bits

### DIFF
--- a/music_assistant/common/models/enums.py
+++ b/music_assistant/common/models/enums.py
@@ -142,12 +142,12 @@ class ContentType(StrEnum):
     UNKNOWN = "?"
 
     @classmethod
-    def _missing_(cls: Self, value: object) -> Self:  # noqa: ARG003
+    def _missing_(cls, value: object) -> Self:  # noqa: ARG003
         """Set default enum member if an unknown value is provided."""
         return cls.UNKNOWN
 
     @classmethod
-    def try_parse(cls: Self, string: str) -> Self:
+    def try_parse(cls, string: str) -> Self:
         """Try to parse ContentType from (url)string/extension."""
         tempstr = string.lower()
         if "audio/" in tempstr:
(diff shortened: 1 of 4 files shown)


music_assistant/common/models/enums.py
```diff
@@ -142,12 +142,12 @@ class ContentType(StrEnum):
     UNKNOWN = "?"
 
     @classmethod
-    def _missing_(cls: Self, value: object) -> Self:  # noqa: ARG003
+    def _missing_(cls, value: object) -> Self:  # noqa: ARG003
```
marcelveldt, line 145:
```suggestion
    def _missing_(cls: type, value: object) -> Self:  # noqa: ARG003
```

I think strictly speaking it should be this ?

music_assistant/server/providers/jellyfin/__init__.py
```diff
@@ -692,7 +661,7 @@ async def get_stream_details(self, item_id: str) -> StreamDetails:
         if ITEM_KEY_MEDIA_CODEC in media_stream:
             content_type = ContentType.try_parse(media_stream[ITEM_KEY_MEDIA_CODEC])
         else:
-            content_type = ContentType.try_parse(mimetype)
+            content_type = ContentType.try_parse(mimetype) if mimetype else ContentType.UNKNOWN
```
marcelveldt, line 664: If you want, we could also just make the try_parse return Unknown when passing in an invalid mimetype

Jc2k: I thought about that but wasn't sure if I liked it or not. It would be stupid to pass an int into that function, so why would I pass None in? But I'm happy to change it if you want.

music_assistant/common/models/enums.py
marcelveldt: well, I think it doesnt matter that much as no type checker is wining about it (guess because its still not clear haha).
Lets keep it just without the type hint

marcelveldt (approved):
Looks good!

PR #1379, opened 2024-06-17
Jellyfin: Make less http requests

Jc2k:
By passing `enableUserData` and `fields` to more requests we can get all the data we need from the `_search` methods without `_parse` methods needing to make additional HTTP requests. This can save hundreds if not thousands of HTTP requests on moderate libraries.

We are able to make _parse_playlist non-async as it already does no I/O, and _parse_album and _parse_artist are (after the change above) also I/O free. 

This just leaves _parse_track. I can easily get rid of 1 extra http request per track. However in the worst case, this function will still make 2 HTTP requests:

* If a track has `AlbumId` but not `Album` (the human readable album name) it does a HTTP request to get it for the ItemMapping
* If a track doesn't have `ArtistItems` but does have `AlbumId` it does a HTTP request to see if the artist data is on the album object.

Both of these are for the same album, so worst case we can get this down to 1 additional HTTP request, but i'lll probably look at that separately.

### DIFF
--- a/music_assistant/server/providers/jellyfin/__init__.py
+++ b/music_assistant/server/providers/jellyfin/__init__.py
@@ -46,6 +46,7 @@ from music_assistant.server.models.music_provider import MusicProvider
 from music_assistant.server.server import MusicAssistant
 
 from .const import (
+    ALBUM_FIELDS,
     ARTIST_FIELDS,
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):
LGTM, the sync is going to fly now ;-)

PR #1380, opened 2024-06-18
Jellyfin: Add pagination for get_library_albums and get_library_artists

Jc2k:
On my test instance I have ~800 artists and ~400 albums, which get fetched in 2 requests. This works and is fast, but I guess some people will have much bigger libraries and that might be a problem. So add some pagination.

After this PR, i'll be switching get_library_tracks to do similar. Hoping that should reduce HTTP calls during get_library_tracks by ~75%.


marcelveldt (approved):
LGTM !

marcelveldt:
Long term solution we can also investigate if we can leverage something like E-Tag to determine if the library actually changed to speedup syncs even further. Just need to figure out a way how to make that opt-in per provider, maybe just by setting a ProviderFeature. Food for thought (for later) again. Anyways, this looks good!

PR #1381, opened 2024-06-18
Add index on timestamp_added

MelHarbour:
The home screen of MA includes a 'recently added' row:

https://github.com/music-assistant/frontend/blob/7553714d748b53ba29b52671586ab38108fa18be/src/components/HomeWidgetRows.vue#L112

My library has ca. 8000s tracks and it takes a fair amount of time to fetch just that row (7-8s I think). This should speed up the fetch and hence the display of the home screen.


marcelveldt:
Thanks!

PR #1383, opened 2024-06-18
Jellyfin: Faster track fetching

Jc2k:
This:

* Fixes a case where the same _parse_track call might call out to get the same album twice
* Fetches all tracks, even if they don't have an album

It also means instead of making a HTTP request for evey album its making one for every 4 or so albums (paginating over the whole library, rather than fetching each albums tracks).

Ideally i'd like to not query the album API at all when parsing the track. But:

* Right now we use the album to find the artist if the track doesn't have an artist directly. This feels weird and edge casey, but it does happen in my library.
* Apparently there might be edge cases where we have an album id but not name. Can I have an item mapping that has an id but not a name?

### DIFF
--- a/music_assistant/server/providers/jellyfin/__init__.py
+++ b/music_assistant/server/providers/jellyfin/__init__.py
@@ -419,31 +419,37 @@ class JellyfinProvider(MusicProvider):
                         artist_item[ITEM_KEY_NAME],
                     )
                 )
-        elif ITEM_KEY_ALBUM_ID in jellyfin_track:
-            parent_album = await self._client.get_item(jellyfin_track[ITEM_KEY_ALBUM_ID])
-            if ITEM_KEY_ALBUM_ARTISTS in parent_album:
-                for artist_item in parent_album[ITEM_KEY_ALBUM_ARTISTS]:
-                    track.artists.append(
-                        self._get_item_mapping(
-                            MediaType.ARTIST,
-                            artist_item[ITEM_KEY_ID],
-                            artist_item[ITEM_KEY_NAME],
-                        )
-                    )
 
-        if ITEM_KEY_ALBUM_ID in jellyfin_track and ITEM_KEY_ALBUM in jellyfin_track:
-            track.album = self._get_item_mapping(
-                MediaType.ALBUM,
-                jellyfin_track[ITEM_KEY_ALBUM_ID],
-                jellyfin_track[ITEM_KEY_ALBUM],
-            )
-        elif ITEM_KEY_ALBUM_ID in jellyfin_track:
-            parent_album = await self._client.get_item(jellyfin_track[ITEM_KEY_ALBUM_ID])
-            track.album = self._get_item_mapping(
-                MediaType.ALBUM,
-                parent_album[ITEM_KEY_ID],
-                parent_album[ITEM_KEY_NAME],
-            )
+        if ITEM_KEY_ALBUM_ID in jellyfin_track:
+            if ITEM_KEY_ALBUM in jellyfin_track:
+                track.album = self._get_item_mapping(
+                    MediaType.ALBUM,
+                    jellyfin_track[ITEM_KEY_ALBUM_ID],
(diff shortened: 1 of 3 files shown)


music_assistant/server/providers/jellyfin/__init__.py
```diff
@@ -419,31 +419,37 @@ async def _parse_track(self, jellyfin_track: JellyMediaItem) -> Track:
...
+        if ITEM_KEY_ALBUM_ID in jellyfin_track:
+            if ITEM_KEY_ALBUM in jellyfin_track:
+                track.album = self._get_item_mapping(
+                    MediaType.ALBUM,
+                    jellyfin_track[ITEM_KEY_ALBUM_ID],
+                    jellyfin_track[ITEM_KEY_ALBUM],
+                )
+
+            # If we have an AlbumID not have not managed to set track.artists or track.album
+            # Let's see if we can pull it from the album
```
marcelveldt, line 432: Isn't this a mega bug in Jellyfin (or its api) ?
Track artists are really something different than album artists.

What will happen with a compilation album for example ?
Are the tracks going to have "Various artists" as artist ?

Jc2k: No, not a jellyfin bug. Jellyfin has separate metadata for both.

Existing code in the provider is trying to handle the case where we have an album artist and not a track artist by assuming the album artist is more useful than "Various".

In my library it would be more right than wrong, but not a fan personally of making fallback guesses. I'd rather fail it or set it to unknown.

This is preserving current behavior. Planning on sorting out unknown/various artists in a follow up PR, if that's ok?

Jc2k: OK, ended up adding the [unknown] artist we talked about on Discord.

music_assistant/server/providers/jellyfin/const.py
```diff
@@ -70,3 +73,7 @@
 
 USER_APP_NAME: Final = "Music Assistant"
 USER_AGENT: Final = "Music-Assistant-1.0"
+
+UNKNOWN_ARTIST_MAPPING = ItemMapping(
+    media_type=MediaType.ARTIST, item_id="[unknown]", provider=DOMAIN, name="[unknown]"
```
Jc2k, line 78: I used '[unknown]' to match musicbrainz. I could use the UNKNOWN_ARTIST constant instead, and i could also update that constant to match musicbrainz.

Jc2k:
> > * Apparently there might be edge cases where we have an album id but not name. Can I have an item mapping that has an id but not a name?
> 
> Technically you can (just set the name to a random string or whatever) but UX is not great as the referenced object will be shown with that name.

OK, if i return an `ItemMapping` with a faked (random) name, then it never shows up in my UI. I assume because the `Album` for that ID was pre-created during the sync already, it doesn't use the name from the `ItemMapping` or some other similiar mechanism.

With both of those changes, `_parse_track` no longer needs to make any HTTP calls 🥳

music_assistant/server/providers/jellyfin/const.py
marcelveldt: If you see improvements in the core/base implementation, always feel free to adjust in your PR or a separate follow-up one. I think it makes sense if we have a global constant value for an unknown artist, including musicbrainz id

Jc2k: Have added the MBID to core, but core has this:

`UNKNOWN_ARTIST: Final[str] = "Unknown Artist"`

So wondered whether to prefer that, or what musicbrainz used.

marcelveldt: I think we should follow musicbrainz

Jc2k: Done - '`music_assistant.server.constants.UNKNOWN_ARTIST`' now `'[unknown]`'

marcelveldt (approved):
Nice work @Jc2k !

PR #1385, opened 2024-06-18
Snapcast fix sync and unsync many

SantiagoSotoC:
The method in the python-snapcast library requires that the sync and unsync be done in series and not in parallel.

Soon I will try to add a method that supports this option since the snapcast api allows these operations in a single call.

### DIFF
--- a/music_assistant/server/models/player_provider.py
+++ b/music_assistant/server/models/player_provider.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import asyncio
 from abc import abstractmethod
 from collections.abc import Iterable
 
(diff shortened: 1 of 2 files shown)


marcelveldt (dismissed):
This is basically the default implemenation of the provider base class but I guess you are future proofing ?

SantiagoSotoC:
The default implementation uses task group, it is done in parallel, while this is done sequentially. 

Fixes a bug in the new UI where if you try to synchronize more than one player at a time it fails.

marcelveldt:
Thinking about this, the Taskgroup may introduce issues in other providers as well maybe ?
Shall we just modify the default implementation instead ?

SantiagoSotoC:
Yes, maybe it is better to change the default implementation and if any provider can take advantage of the task group to do it internally, now I'm on my cell phone but in a while I'll do a new PR, is that ok?

marcelveldt:
> Yes, maybe it is better to change the default implementation and if any provider can take advantage of the task group to do it internally, now I'm on my cell phone but in a while I'll do a new PR, is that ok?

I've already adjusted the PR so its all set

SantiagoSotoC:
Sorry, I had a long day, I thought I was free earlier

PR #1388, opened 2024-06-19
Prevent overwriting album_tracks table

MelHarbour:
If you have the same track on two different albums via the filesystem provider, this code will force an overwrite of the album_tracks table, meaning it only gets shown on a single album, rather than (correctly) in both places.

AIUI the intention of the overwrite is to force rewriting in the case of modified ID3 tags, but the only tags that would be relevant in this case would be disc number and track number, and those should already be taken care of by this line:

https://github.com/music-assistant/server/blob/642ccbef85bd5a190078563e8572a5aa0dbc0e89/music_assistant/server/controllers/media/tracks.py#L453

Which does an insert or update as required.

See https://github.com/music-assistant/hass-music-assistant/issues/2506 for reporting of the issue.

### DIFF
--- a/music_assistant/server/controllers/media/tracks.py
+++ b/music_assistant/server/controllers/media/tracks.py
@@ -410,7 +410,6 @@ class TracksController(MediaControllerBase[Track]):
         For digital releases, the discnumber will be just 0 or 1.
         Track number should start counting at 1.
         """
-        
         db_album: Album | ItemMapping = None
         if album.provider == "library":
             db_album = album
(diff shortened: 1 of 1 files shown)


marcelveldt:
I honestly can't remember why it was implemented this way, there was some vague edge case where someone had really bad tags and then rescanned the collection and that left some leftovers but I guess those situations should just be avoided and we assume the id3 tags to be in somewhat good shape or the user just re-adds the file provider in case of massive re-tagging.

@OzGav I think that will be a good note to add to the docs. We cant recover from every situation where the user starts doing massive rename/retag operations, e.g. renaming album names etc. In case of big operations it may be better to just re-add the file provider.

marcelveldt (approved):
Thanks!

PR #1390, opened 2024-06-19
Jellyfin: Stricter typing

Jc2k:
aiojellyfin now has typed functions for everything, so instead of searches returning a "MediaItem", there is a wrapper that will return an Artist, a wrapper that will return a Track, etc. So you can't pass an album into _parse_track etc. A side effect of that is we can delete some code on the MA side, yay!


marcelveldt (approved):
LGTM!

PR #1393, opened 2024-06-19
Jellyfin: Get more artwork from the Jellyfin instance

Jc2k:
This fetches `Backdrop` and `Logo`, in addition to `Primary`.


marcelveldt (approved):
LGTM !

PR #1397, opened 2024-06-20
Jellyfin: Add some basic parsing tests

Jc2k:
These make sure that we can turn Jellyfin models into MA models.

This breaks the parsers out of the provider class so that they are a bit easier to unittest.

I've included some dodgy data that i've previously fixed handling of to protect us from regressions.


Jc2k:
Test failure is unrelated (not mine :laughing:)

tests/server/providers/jellyfin/test_parsers.py
```diff
@@ -0,0 +1,71 @@
...
+from collections.abc import AsyncGenerator
+
+import aiofiles
+import aiohttp
+import pytest
+from aiojellyfin import Artist, Connection, SessionConfiguration
+from mashumaro.codecs.json import JSONDecoder
+from syrupy.assertion import SnapshotAssertion
+
+from music_assistant.server.providers.jellyfin.parsers import parse_album, parse_artist, parse_track
```
marcelveldt, line 14: this line looks too long to me, is the linter not working on the files in tests ?

tests/server/providers/jellyfin/test_parsers.py
```diff
@@ -0,0 +1,71 @@
...
+
+ARTIST_DECODER = JSONDecoder(Artist)
+
+_LOGGER = logging.getLogger(__name__)
+
+
+@pytest.fixture
+async def connection() -> AsyncGenerator[Connection, None]:
+    """Spin up a dummy connection."""
+    async with aiohttp.ClientSession() as session:
```
marcelveldt, line 29: do we want to globally mock the aiohttp session ?

Jc2k: To be honest i'm a bit annoyed this bit needs a ClientSession at all. its not actually used by the test, I just need one to instance the aiojellyfin objects used in the test. I am probably going to try to refactor it away at some point.

I am playing around with mocking ClientSession generally in another branch. The idea is to provide a fake for the API in aiojellyfin that we can load fixtures into, then have some way to use that fake in the MA tests with the `aiohttp_client` fixture (which does have its own ClientSession). I'll probably combine that with something like the socket blocker from HA's tests.

tests/server/providers/jellyfin/test_parsers.py
Jc2k:
```
ruff format tests/server/providers/jellyfin/test_parsers.py 
1 file left unchanged

ruff check tests/server/providers/jellyfin/test_parsers.py 
All checks passed!
```

Jc2k: And come to think of it, [af7b711](https://github.com/music-assistant/server/pull/1397/commits/af7b711fdf75ab5a91b2a3369f5b84f361756203) was because of the lints running on the tests.

marcelveldt (approved):
Amazing work again, nice!

PR #1398, opened 2024-06-21
Tests: Add a 'mass' fixture

Jc2k:
Adds a fixture `mass`, like the `hass` fixture. Deliberately basic for now we'll mock out more as needed and gradually build out guard rails.

After this i'll work on an example provider test for jellyfin where a fixture adds a provider config for jellyfin to the running `mass`, then does syncs against a mocked out API.


marcelveldt (approved):
Nice, thanks for getting this started!

PR #1405, opened 2024-06-24
Remove some unnecessary async and await in ytmusic

tronikos:


MarvinSchenkel (approved):
Awesome! Completely missed this, thanks for this.

PR #1411, opened 2024-06-25
Double check external ID comparison

MelHarbour:
Additional logic got added to the compare_track function that accounts for ISRCs. When importing, it doesn't make it that far though. This is slightly clunky, but adding in an extra double check should handle it.

### DIFF
--- a/music_assistant/server/controllers/media/base.py
+++ b/music_assistant/server/controllers/media/base.py
@@ -126,7 +126,6 @@ class MediaControllerBase(Generic[ItemCls], metaclass=ABCMeta):
             # existing item match by provider id
             await self._update_library_item(cur_item.item_id, item, overwrite=overwrite_existing)
             return cur_item.item_id
-
         if cur_item := await self.get_library_item_by_external_ids(item.external_ids):
             # existing item match by external id
             # Double check external IDs - if MBID exists, regards that as overriding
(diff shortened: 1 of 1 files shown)


MelHarbour:
Should finally fix https://github.com/music-assistant/hass-music-assistant/issues/2316

music_assistant/server/controllers/media/base.py
```diff
@@ -106,8 +106,12 @@ async def add_item_to_library(
...
         elif cur_item := await self.get_library_item_by_external_ids(item.external_ids):
             # existing item match by external id
-            await self._update_library_item(cur_item.item_id, item, overwrite=overwrite_existing)
-            library_id = cur_item.item_id
+            # Double check external IDs - if MBID exists, regards that as overriding
+            if compare_media_item(item, cur_item):
+                await self._update_library_item(
+                    cur_item.item_id, item, overwrite=overwrite_existing
+                )
+                library_id = cur_item.item_id
```
marcelveldt, line 114: This now discards the cur_item for more comparisons further down below.
Easiest way to fix this code is to change this to guards by doing if's only with return
So in this case, the return will only be after compare_media_item succeeded

```
if cur_item := await self.get_library_item_by_prov_id(item.item_id, item.provider):
   do blah
   return

if cur_item := await self.get_library_item_by_external_ids(item.external_ids):
    do blah
    return

etc.

```

MelHarbour: Ok, I've broken out the comparison - I think that does the right thing now, making sure all possible checks are done.

marcelveldt (approved):
Nice!

PR #1412, opened 2024-06-25
Avoid subshells when using check_output

Jc2k:
A few places were passing `str` to check_output rather then `list[str]`. The list is better - avoids needing to worry about escaping (for security), avoids a few fstrings, and avoids creating /bin/sh instances we don't need.

Hopefully fixes: https://github.com/music-assistant/hass-music-assistant/issues/2234 https://github.com/music-assistant/hass-music-assistant/issues/2266

### DIFF
--- a/music_assistant/server/providers/filesystem_smb/__init__.py
+++ b/music_assistant/server/providers/filesystem_smb/__init__.py
@@ -220,7 +220,10 @@ class SMBFileSystemProvider(LocalFileSystemProvider):
             raise LoginFailed(msg)
 
         self.logger.info("Mounting //%s/%s%s to %s", server, share, subfolder, self.base_path)
-        self.logger.debug("Using mount command: %s", mount_cmd.replace(password, "########"))
+        self.logger.debug(
+            "Using mount command: %s",
+            [m.replace(password, "########") if password else m for m in mount_cmd],
+        )
 
         returncode, output = await check_output(mount_cmd)
         if returncode != 0:
--- a/music_assistant/server/providers/filesystem_smb/__init__.py
+++ b/music_assistant/server/providers/filesystem_smb/__init__.py
@@ -186,7 +186,13 @@ class SMBFileSystemProvider(LocalFileSystemProvider):
 
         if platform.system() == "Darwin":
             password_str = f":{password}" if password else ""
-            mount_cmd = f'mount -t smbfs "//{username}:{password_str}@{server}/{share}{subfolder}" "{self.base_path}"'  # noqa: E501
+            mount_cmd = [
+                "mount",
+                "-t",
+                "smbfs",
+                f"//{username}:{password_str}@{server}/{share}{subfolder}",
+                self.base_path,
+            ]
 
         elif platform.system() == "Linux":
             options = [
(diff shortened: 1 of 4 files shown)


music_assistant/server/providers/filesystem_smb/__init__.py
```diff
@@ -186,7 +186,13 @@ async def mount(self) -> None:
 
         if platform.system() == "Darwin":
             password_str = f":{password}" if password else ""
-            mount_cmd = f'mount -t smbfs "//{username}:{password_str}@{server}/{share}{subfolder}" "{self.base_path}"'  # noqa: E501
+            mount_cmd = [
+                "mount",
+                "-t",
+                "smbfs",
+                f"//{username}:{password_str}@{server}/{share}{subfolder}",
```
MelHarbour, line 193: You're still going to have the escaping issue on this line, aren't you?

music_assistant/server/providers/filesystem_smb/__init__.py
```diff
@@ -199,10 +205,15 @@ async def mount(self) -> None:
...
-                f"mount -t cifs -o {options_str} "
-                f'"//{server}/{share}{subfolder}" "{self.base_path}"'
-            )
+            mount_cmd = [
+                "mount",
+                "-t",
+                "cifs",
+                "-o",
+                options_str,
+                f"//{server}/{share}{subfolder}",
```
MelHarbour, line 214: Also escaping on this line

music_assistant/server/providers/filesystem_smb/__init__.py
Jc2k: Primarily this is about avoiding subshells.

If you pass a `str` to `check_output` it runs `asyncio.create_subprocess_shell`. This is like passing shell=True to the `subprocess` module - rather than creating your target program directly you invoke `/bin/sh -c 'your shell commands here'`. Here, we are subject to shell parsing rules. And that means you can inject a password of `' || rm -rf / || true #` and /bin/sh will return an exit code of 0 and delete your OS. User controlled variables getting near /bin/sh is one of the most common problems you see with modem/router vulnerabilities.

If you pass `list[str]`, it actually invokes `create_subprocess_exec`. Here there is no `/bin/sh`. The target program is invoked directly. There is no interpreting or parsing. `arg[0]` in python becomes `argv[0]` in the target process. It is `/bin/sh` that creates the security issue.

`shlex.quote` was mentioned in one of the tickets. It is way to armor a variable so that when a shell parses it, it arrives at the original variable we had on our side (i.e. to protect it from shell expansion or injection). If there is no /bin/sh, nothing will remove that armor before `mount` is invoked. The target process can't actually can't cope with the armored version.

See also https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_shell vs https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec - the quoting advice only applies to create_subprocess_shell.

So, this PR will stop me from being able to own you by using a dodgey username or password. As a side effect, more symbols in passwords will work. I'd expect `'` to work.

You are correct that we may need to additionally urlencode this URI. For example, an '@' in a password might be a problem. I think you are supposed to use url encoding rules on them (so '@' becomes '%40'). I can't test that, so i haven't done it here.

MelHarbour: Cool - thanks for taking the time to level up my understanding of how python treats the various subprocess commands as well!

marcelveldt (approved):

PR #1415, opened 2024-06-25
Fix unload on snapcast

SantiagoSotoC:
A warning is displayed when deleting the provider, I forgot to change the cmd_stop code here in previous changes.


SantiagoSotoC:
It does not eliminate the error, but still the actual code is incorrect and this is what is expected

marcelveldt (approved):

PR #1413, opened 2024-06-25
Run mypy on CI

Jc2k:
Turns on mypy on CI for tests and jellyfin.

PR #1416, opened 2024-06-26
Mypy: Get mypy passing on music_assistant.client

Jc2k:
I'll add it to CI separately (blocked on https://github.com/music-assistant/server/pull/1413)


marcelveldt (approved):
Nice work!

PR #1418, opened 2024-06-26
YTMusic: Fix missing metadata for Artists on library sync

MarvinSchenkel:
Fixes missing thumbs and descriptions during initial library sync. Previously, Artists were added as ItemMappings because of the missing info. However, those were never refreshed unless the were actively clicked on by the user. With this change, full artist details will be fetched during the library sync from YT Music, also relieving some stress on MusicBrainz.

### DIFF
--- a/music_assistant/server/providers/ytmusic/__init__.py
+++ b/music_assistant/server/providers/ytmusic/__init__.py
@@ -6,7 +6,6 @@ import asyncio
 import logging
 from collections.abc import AsyncGenerator
 from time import time
-from timeit import default_timer as timer
 from typing import TYPE_CHECKING, Any
 from urllib.parse import unquote
 
@@ -621,8 +620,8 @@ class YoutubeMusicProvider(MusicProvider):
             album.year = album_obj["year"]
         if "thumbnails" in album_obj:
             album.metadata.images = self._parse_thumbnails(album_obj["thumbnails"])
-        if "description" in album_obj:
-            album.metadata.description = unquote(album_obj["description"])
+        if description := album_obj.get("description"):
+            album.metadata.description = unquote(description)
         if "isExplicit" in album_obj:
             album.metadata.explicit = album_obj["isExplicit"]
         if "artists" in album_obj:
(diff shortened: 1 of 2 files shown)


marcelveldt (changes requested):
I'm not sure I can agree with this. This will result in a shitshow of requests against their api if I'm reading this right ?

ItemMapping exists to prevent all these additional per-item calls.

MarvinSchenkel:
> I'm not sure I can agree with this. This will result in a shitshow of requests against their api if I'm reading this right ?
> 
> ItemMapping exists to prevent all these additional per-item calls.

True, but the side effect is an artist overview without any thumbs and other Metadata. Somehow a force refresh on these item mappings do not result in resolving the artist from either ytm or music brains, so th user is left with empty Metadata here.

Also, these additional requests are only made during the sync.

marcelveldt:
Let's concentrate on the issue then. By default artists (that go into the db from albums or tracks) wont have metadata unless you first click them. Is that not working ? Then we should fix that. Doing a full artist lookup at every sync is really way too heavy. What we can do however if this is really wanted, direct the request to the artists controller, as that will use the cache.

so instead of grabbing the artist within your own provider, do it by using mass.music.artists.get_provider_item

Another approach we could consider is restoring the background check but with a lot more caution to prevent too many calls.

MarvinSchenkel:
> Let's concentrate on the issue then. By default artists (that go into the db from albums or tracks) wont have metadata unless you first click them. Is that not working ? Then we should fix that. Doing a full artist lookup at every sync is really way too heavy. What we can do however if this is really wanted, direct the request to the artists controller, as that will use the cache.
> 
> so instead of grabbing the artist within your own provider, do it by using mass.music.artists.get_provider_item
> 
> Another approach we could consider is restoring the background check but with a lot more caution to prevent too many calls.

Yeah I have to agree with this. First things first, I'll see if I can figure out why an item mapping is never updated with the full details. I can see the request coming in in the YT Music provider's get_artist but it never reaches the update_db in the artist controller.

marcelveldt:
superseded by https://github.com/music-assistant/server/pull/1480

PR #1419, opened 2024-06-26
Snapcast remove check for 0.27

SantiagoSotoC:
The debian repo was updated to v0.28 that's why snapcast didn't start, I do a more generic check so that in the next update it won't break and it will warn us in the log in case we have to make any changes.


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -109,7 +109,7 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
     returncode, output = await check_output("snapserver -v")
-    snapserver_present = returncode == 0 and "snapserver v0.27.0" in output.decode()
+    snapserver_present = returncode == 0 and "snapserver v0.2" in output.decode()
```
marcelveldt, line 112: I would not do that - we need actually the version to be minimum 0.27 so what I would do here is parse the version and compare it. So first check if the returncode is 0. Next step is parse the version.

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -109,7 +109,14 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
     returncode, output = await check_output(["snapserver", "-v"])
-    snapserver_present = returncode == 0 and "snapserver v0.27.0" in output.decode()
+    snapserver_version: int = int(output.decode().split(".")[1] or -1)
```
marcelveldt, line 112: this will crash if the returncode is not 0 so you will need to move this check after the returncode check or guard for empty output

SantiagoSotoC: Something like this is ok?
` snapserver_version: int = int(output.decode().split(".")[1]) if returncode == 0 else -1`

SantiagoSotoC:
This would not fail, according to the python documentation
[https://docs.python.org/3/reference/expressions.html#conditional-expressions](https://docs.python.org/3/reference/expressions.html#conditional-expressions)

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: yes

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -109,7 +109,14 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
     returncode, output = await check_output(["snapserver", "-v"])
-    snapserver_present = returncode == 0 and "snapserver v0.27.0" in output.decode()
+    snapserver_version: int = int(output.decode().split(".")[1]) if returncode == 0 else -1
+    snapserver_valid_version: bool = snapserver_version >= 27
+    snapserver_present = (
+        returncode == 0 and "snapserver" in output.decode() and snapserver_valid_version
+    )
```
marcelveldt, line 116: I can imagine this being obsolete now that you have an explicit version check ?

SantiagoSotoC: Yes, you're right, in 5 minutes I'll change it.

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -110,7 +110,11 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
     returncode, output = await check_output(["snapserver", "-v"])
-    snapserver_present = returncode == 0 and "snapserver v0.27.0" in output.decode()
+    snapserver_version: int = int(output.decode().split(".")[1]) if returncode == 0 else -1
+    snapserver_valid_version: bool = snapserver_version >= 27
```
marcelveldt, line 114:
```suggestion
    local_snapserver_present = snapserver_version >= 27
```

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -123,7 +127,7 @@ async def get_config_entries(
             "the sample is played-out on the client",
             required=False,
             category="Built-in Snapserver Settings",
-            hidden=not snapserver_present,
+            hidden=not snapserver_valid_version,
```
marcelveldt, line 130:
```suggestion
            hidden=not local_snapserver_present,
```

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: renaming it to local_snapserver_present makes it more implicit imo

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -110,7 +110,11 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
     returncode, output = await check_output(["snapserver", "-v"])
-    snapserver_present = returncode == 0 and "snapserver v0.27.0" in output.decode()
+    snapserver_version: int = int(output.decode().split(".")[1]) if returncode == 0 else -1
```
marcelveldt, line 113:
```suggestion
    snapserver_version = int(output.decode().split(".")[1]) if returncode == 0 else -1
```

marcelveldt: no need to add the type hint as the type can be derived from the input

SantiagoSotoC:
I ran this branch in the dev addon, with main merge, I would have to increase the dryout ms value, it seemed to work fine with 2000 but it also fails, I increase it here to not make a PR just for that ?

marcelveldt (approved):

PR #1425, opened 2024-06-26
Apple Music: Fix tracks that are unavailable in the catalog + Fix tracks with no album data

MarvinSchenkel:
This PR:
- Fixes https://github.com/music-assistant/hass-music-assistant/issues/2431
- Fixes https://discord.com/channels/753947050995089438/1242741503642570752/1253508280802607125

### DIFF
--- a/music_assistant/server/providers/apple_music/__init__.py
+++ b/music_assistant/server/providers/apple_music/__init__.py
@@ -199,9 +199,16 @@ class AppleMusicProvider(MusicProvider):
         """Retrieve library tracks from the provider."""
         endpoint = "me/library/songs"
         song_catalog_ids = []
-        for item in await self._get_all_items(endpoint, include="catalog"):
-            if item and "catalog" in item["relationships"]:
-                song_catalog_ids.append(item["relationships"]["catalog"]["data"][0]["id"])
+        for item in await self._get_all_items(endpoint):
+            catalog_id = item.get("attributes", {}).get("playParams", {}).get("catalogId")
+            if not catalog_id:
+                self.logger.warning(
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/apple_music/__init__.py
```diff
@@ -199,9 +199,16 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
         """Retrieve library tracks from the provider."""
         endpoint = "me/library/songs"
         song_catalog_ids = []
-        for item in await self._get_all_items(endpoint, include="catalog"):
-            if item and "catalog" in item["relationships"]:
-                song_catalog_ids.append(item["relationships"]["catalog"]["data"][0]["id"])
+        for item in await self._get_all_items(endpoint):
+            catalog_id = item.get("attributes", {}).get("playParams", {}).get("catalogId")
+            if not catalog_id:
+                self.logger.warning(
```
marcelveldt, line 205: Does it need to be warning level ? Maybe just use debug level ?

MarvinSchenkel: I was thinking in the lines of how we also warn about songs without artist. The result the catalog_id not being available is that the song does not show up in MA, so I thought a warning was appropriate. If you think it should be debug, I will change it.

marcelveldt: From what I read between the lines of the issue reports is that this happens for some songs that folks uploaded to their library ages ago so maybe debug is just enough. But do that in a follow up PR if folks complain about noisy logs.

marcelveldt (approved):

PR #1428, opened 2024-06-30
Mypy: Add music_assistant.common

Jc2k:
I've added some comments where i'm a bit uncertain on this one.

### DIFF
--- a/music_assistant/client/music.py
+++ b/music_assistant/client/music.py
@@ -3,9 +3,10 @@
 from __future__ import annotations
 
 import urllib.parse
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, cast
 
 from music_assistant.common.models.enums import ImageType, MediaType
+from music_assistant.common.models.errors import UnsupportedFeaturedException
 from music_assistant.common.models.media_items import (
     Album,
     AlbumTrack,
@@ -19,13 +20,27 @@ from music_assistant.common.models.media_items import (
     Radio,
     SearchResults,
     Track,
-    media_from_dict,
 )
 from music_assistant.common.models.provider import SyncTask
 from music_assistant.common.models.queue_item import QueueItem
 
 if TYPE_CHECKING:
     from .client import MusicAssistantClient
+
+
+def media_from_dict(media_item: dict[str, Any]) -> MediaItemType:
+    """Return MediaItem from dict."""
+    if media_item["media_type"] == "artist":
+        return Artist.from_dict(media_item)
+    if media_item["media_type"] == "album":
+        return Album.from_dict(media_item)
+    if media_item["media_type"] == "track":
+        return Track.from_dict(media_item)
+    if media_item["media_type"] == "playlist":
+        return Playlist.from_dict(media_item)
+    if media_item["media_type"] == "radio":
+        return Radio.from_dict(media_item)
+    raise UnsupportedFeaturedException(f"Unknown media_type: {media_item['media_type']}")
 
 
 class Music:
--- a/music_assistant/client/music.py
+++ b/music_assistant/client/music.py
@@ -3,10 +3,9 @@
 from __future__ import annotations
 
 import urllib.parse
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, cast
 
 from music_assistant.common.models.enums import ImageType, MediaType
-from music_assistant.common.models.errors import UnsupportedFeaturedException
 from music_assistant.common.models.media_items import (
     Album,
     AlbumTrack,
@@ -20,27 +19,13 @@ from music_assistant.common.models.media_items import (
     Radio,
     SearchResults,
     Track,
+    media_from_dict,
 )
 from music_assistant.common.models.provider import SyncTask
 from music_assistant.common.models.queue_item import QueueItem
 
 if TYPE_CHECKING:
     from .client import MusicAssistantClient
-
-
-def media_from_dict(media_item: dict[str, Any]) -> MediaItemType:
-    """Return MediaItem from dict."""
-    if media_item["media_type"] == "artist":
-        return Artist.from_dict(media_item)
-    if media_item["media_type"] == "album":
-        return Album.from_dict(media_item)
-    if media_item["media_type"] == "track":
-        return Track.from_dict(media_item)
-    if media_item["media_type"] == "playlist":
-        return Playlist.from_dict(media_item)
-    if media_item["media_type"] == "radio":
-        return Radio.from_dict(media_item)
-    raise UnsupportedFeaturedException(f"Unknown media_type: {media_item['media_type']}")
 
 
 class Music:
@@ -455,7 +440,7 @@ class Music:
         path: str | None = None,
         limit: int | None = None,
         offset: int | None = None,
-    ) -> list[MediaItemType]:
+    ) -> list[MediaItemType | ItemMapping]:
         """Browse Music providers."""
         return [
             media_from_dict(obj)
@@ -469,7 +454,7 @@ class Music:
 
     async def recently_played(
         self, limit: int = 10, media_types: list[MediaType] | None = None
-    ) -> list[MediaItemType]:
+    ) -> list[MediaItemType | ItemMapping]:
         """Return a list of the last played items."""
         return [
             media_from_dict(item)
@@ -481,7 +466,7 @@ class Music:
     async def get_item_by_uri(
         self,
         uri: str,
-    ) -> MediaItemType:
+    ) -> MediaItemType | ItemMapping:
         """Get single music item providing a mediaitem uri."""
         return media_from_dict(await self.client.send_command("music/item_by_uri", uri=uri))
 
@@ -493,7 +478,7 @@ class Music:
         force_refresh: bool = False,
         lazy: bool = True,
         add_to_library: bool = False,
-    ) -> MediaItemType:
+    ) -> MediaItemType | ItemMapping:
         """Get single music item by id and media type."""
         return media_from_dict(
             await self.client.send_command(
@@ -549,7 +534,7 @@ class Music:
     async def refresh_item(
         self,
         media_item: MediaItemType,
-    ) -> MediaItemType | None:
+    ) -> MediaItemType | ItemMapping | None:
         """Try to refresh a mediaitem by requesting it's full object or search for substitutes."""
         if result := await self.client.send_command("music/refresh_item", media_item=media_item):
             return media_from_dict(result)
--- a/music_assistant/common/helpers/datetime.py
+++ b/music_assistant/common/helpers/datetime.py
@@ -27,7 +27,7 @@ def now_timestamp() -> float:
     return now().timestamp()
 
 
-def future_timestamp(**kwargs) -> float:
+def future_timestamp(**kwargs: float) -> float:
     """Return current timestamp + timedelta."""
     return (now() + datetime.timedelta(**kwargs)).timestamp()
 
--- a/music_assistant/common/helpers/util.py
+++ b/music_assistant/common/helpers/util.py
@@ -7,7 +7,8 @@ import os
 import re
 import socket
 from collections.abc import Callable
-from typing import Any, TypeVar
+from collections.abc import Set as AbstractSet
+from typing import Any, Generic, TypeVar
 from urllib.parse import urlparse
 from uuid import UUID
 
@@ -335,7 +336,7 @@ def get_changed_keys(
     dict1: dict[str, Any],
     dict2: dict[str, Any],
     ignore_keys: list[str] | None = None,
-) -> set[str]:
+) -> AbstractSet[str]:
     """Compare 2 dicts and return set of changed keys."""
     return get_changed_values(dict1, dict2, ignore_keys).keys()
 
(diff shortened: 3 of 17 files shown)


music_assistant/common/models/enums.py
```diff
@@ -3,13 +3,25 @@
...
+from enum import EnumType, StrEnum
 
-from music_assistant.common.helpers.util import classproperty
 
+class MediaTypeMeta(EnumType):
+    """Class properties for MediaType."""
 
-class MediaType(StrEnum):
+    @property
+    def ALL(self) -> list[MediaType]:  # noqa: N802
```
Jc2k, line 13: `@classproperty` is a bit broken under mypy, e.g. https://github.com/python/mypy/issues/2563. Properties on the metaclass work like classproperty and have working typing. Maybe a constant of type `Final[tuple[MediaType, ...]]` would be better?

music_assistant/common/models/media_items.py
```diff
@@ -269,20 +274,16 @@ class _MediaItemBase(DataClassDictMixin):
...
     # sort_name and uri are auto generated, do not override unless really needed
     sort_name: str | None = None
     uri: str | None = None
     external_ids: set[tuple[ExternalID, str]] = field(default_factory=set)
     media_type: MediaType = MediaType.UNKNOWN
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         """Call after init."""
-        if self.name is None:
```
Jc2k, line 282: I had to lift this up or make name `str | None`. If the problem was that plex is explicitly passing None, then I think we have to fix that in the provider itself but I don't know if that was the problem or not.

music_assistant/common/models/media_items.py
```diff
@@ -318,16 +319,17 @@ def __hash__(self) -> int:
...
+        if not isinstance(other, MediaItem | ItemMapping):
+            return False
         return self.uri == other.uri
 
 
 @dataclass(kw_only=True)
 class MediaItem(_MediaItemBase):
     """Base representation of a media item."""
 
-    __hash__ = _MediaItemBase.__hash__
```
Jc2k, line 330: I didn't know why these were explicitly assigned like this rather than using the inherited ones, but they break the "self" typing so I had to remove a few.

music_assistant/common/models/media_items.py
```diff
@@ -432,7 +433,11 @@ def has_chapters(self) -> bool:
         This is often an indicator that this track is an episode from a
         Podcast or AudioBook.
         """
-        return self.metadata and self.metadata.chapters and len(self.metadata.chapters) > 1
+        if not self.metadata:
```
Jc2k, line 436: If i didn't break this out like this it technically returned `bool | None`

music_assistant/common/models/media_items.py
```diff
@@ -465,28 +470,22 @@ class AlbumTrack(Track):
...
     @classmethod
     def from_track(
-        cls: type,
+        cls,
         track: Track,
         album: Album | None = None,
         disc_number: int | None = None,
         track_number: int | None = None,
-    ) -> Self:
+    ) -> AlbumTrack:
```
Jc2k, line 478: Again, mypy seems to struggle a bit with classmethods and self. Self is apparently meant for when there might be a subclass, so i'm not too worried that I had to set the type here.

music_assistant/client/music.py
```diff
@@ -28,6 +28,21 @@
     from .client import MusicAssistantClient
 
 
+def media_from_dict(media_item: dict[str, Any]) -> MediaItemType:
```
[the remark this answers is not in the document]
marcelveldt: ItemMappings are also used in the client (e.g. track artists) so we may need to update the type hints to "MediaItemType | ItemMapping" ?

music_assistant/common/helpers/util.py
```diff
@@ -7,14 +7,13 @@
...
 import socket
 from collections.abc import Callable
+from collections.abc import Set as AbstractSet
 from typing import Any, TypeVar
 from urllib.parse import urlparse
 from uuid import UUID
 
 # pylint: disable=invalid-name
 T = TypeVar("T")
-_UNDEF: dict = {}
```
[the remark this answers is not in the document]
marcelveldt: probably just some copy paste leftover mess

music_assistant/common/models/config_entries.py
```diff
@@ -254,7 +257,8 @@ def update(self, update: dict[str, ConfigValueType]) -> set[str]:
             changed_keys.add(key)
 
         # config entry values
-        values = update.get("values", update)
+        values = cast(dict[str, ConfigValueType], update.get("values", update))
```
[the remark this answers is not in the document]
marcelveldt: A config object.dict has values with the actual configentry values.

music_assistant/common/models/enums.py
marcelveldt: yeah, this is only used as a shortcut on some places to reference all media types. feel free to adjust/modify

music_assistant/common/models/media_items.py
marcelveldt: Yeah, this is an annoying one - one of the providers actually has a few edge cases where the name was None, causing a flow of edge case messy hell. Maybe I shouldnt have put this workaround here but explicitly handle it in the provider where it happened. Now, I just need to remember which provider had this issue :-)

music_assistant/common/models/media_items.py
marcelveldt: Otherwise they dont work - they do not inherit otherwise

music_assistant/common/helpers/datetime.py
```diff
@@ -27,7 +27,7 @@ def now_timestamp() -> float:
     return now().timestamp()
 
 
-def future_timestamp(**kwargs) -> float:
+def future_timestamp(**kwargs: float) -> float:
```
marcelveldt, line 30: this feels a bit weird to type the kwargs this way - what are the actual (kw)args on the timedelta method ?

music_assistant/common/models/enums.py
```diff
@@ -56,7 +57,7 @@ class ExternalID(StrEnum):
     UNKNOWN = "unknown"
 
     @classmethod
-    def _missing_(cls: Self, value: object) -> Self:  # noqa: ARG003
+    def _missing_(cls, value: object) -> ExternalID:  # noqa: ARG003
```
marcelveldt, line 60:
```suggestion
    def _missing_(cls: type, value: object) -> Self:  # noqa: ARG003
```

maybe this works ?

music_assistant/common/models/media_items.py
marcelveldt: ok strange, as far as I remember the Self type was explcitly added for this case. But fine to just use the class as type but I wonder why they introduced the Self type then

music_assistant/common/models/media_items.py
Jc2k: This way breaks mypy because it expects self to be MediaItemBase rather than MediaItem, so i'll have to add a function that just calls super() i think.

music_assistant/common/models/media_items.py
Jc2k: The error you get is 

```
error: Incompatible return value type (got "AlbumTrack", expected "Self")  [return-value]
```

Possibly https://github.com/python/mypy/issues/16558?

Which says, if you subclassed `AlbumTrack` and `SomeSubclass.from_track` returned an `AlbumTrack`, rather than `SomeSubclass`, then there would be a type violation.

I ran into this after a similar problem where the mypy issue said to just not use Self for enums (iirc), so assumed this one was similar. But with fresh eye, it also looks like it works to do `cls.from_dict()`. 

One of the issues said `Self` was intended for cases where you were subclassing, as a shortcut for making a method Generic[] over cls or self etc.

music_assistant/common/helpers/datetime.py
Jc2k: Agree. It makes me not want to use kwargs anywhere tbh. But they need a type. I checked before i did this and timedelta has 7 args, and they are all floats.

I could add all 7 args as fully fleshed out kwargs?

music_assistant/common/models/enums.py
Jc2k: https://github.com/python/mypy/issues/15223 maybe?

music_assistant/common/models/media_items.py
```diff
@@ -571,7 +570,7 @@ class SearchResults(DataClassDictMixin):
     radio: list[Radio | ItemMapping] = field(default_factory=list)
 
 
-def media_from_dict(media_item: dict) -> MediaItemType:
+def media_from_dict(media_item: dict[str, Any]) -> MediaItemType | ItemMapping | MediaItem:
```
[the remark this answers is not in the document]
Jc2k: If I remove that I get:

```
error: Incompatible return value type (got "MediaItem", expected "Artist | Album | PlaylistTrack | AlbumTrack | Track | Radio | Playlist | BrowseFolder | ItemMapping")  [return-value]
```

It seems to be a fallback case - if its an unknown media type, load it as a MediaItem. Maybe we just throw in that case?

music_assistant/common/models/media_items.py
Jc2k: I think it was Plex. I can add a similar workaround there.

music_assistant/common/models/config_entries.py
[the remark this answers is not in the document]
Jc2k: So really ConfigValueType needs to include `dict[str, ConfigValueType]` as an allowed type (if they can be recursive).

music_assistant/common/models/media_items.py
marcelveldt: I think that is better - the models should just assume that the provider doesn't do stupid things.

music_assistant/common/models/media_items.py
marcelveldt: I have no idea why it doesn't work though - could be a bug in python 3.12 or it is simply not meant to inherit

music_assistant/common/models/media_items.py
marcelveldt: OK all good then, I actually like the explicit typing more than Self ;-)

music_assistant/common/helpers/datetime.py
marcelveldt: that is sad, is there no generic way to type kwargs ?
I guess dict[str, Any] but that is maybe a tiny bit too wide?

music_assistant/common/models/enums.py
marcelveldt: argh wth is this so convoluted haha. Well, lets just pick what works now and not worry too much about it

music_assistant/common/models/config_entries.py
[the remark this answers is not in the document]
Jc2k: What i'm stuck on here is that update() is typed as `dict[str, ConfigValueType]`, but that doesn't seem to be the case. Because the dict passed in has a key which itself is a `dict[str, ConfigValueType]`, which is why i needed the cast to get this line passing.

So i really need a TypedDict here that has enabled, name and values maybe?

music_assistant/common/models/media_items.py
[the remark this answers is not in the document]
marcelveldt: I can't think of any reason or usecase why to use the MediaItem - I'm suspecting this is just some leftover from early days. We use the explicit mediatype now everywhere (like Track, Album etc.) - thinking about this, you can say that the MediaItem class should perhaps become private (like _MediaItemBase) ?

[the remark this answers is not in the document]
marcelveldt: I'd say we raise indeed if we cant match

music_assistant/common/helpers/datetime.py
Jc2k: I did find https://stackoverflow.com/a/63550734. There's an Unpack that you can use with kwargs to refer to a TypedDict. I can't say I like it that much. Would rather just type out the kwargs I think.

music_assistant/common/models/media_items.py
Jc2k: 4d5e2e47ac58d9cb1986754201662b0bcd0d3cb4

music_assistant/common/models/media_items.py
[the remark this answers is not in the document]
Jc2k: ffea82656b43ae42fbed8a3daacae0b17d687759

music_assistant/common/models/media_items.py
Jc2k: 4f80bdfe8550eae805d8927bbd4751c213b645ae

music_assistant/common/models/config_entries.py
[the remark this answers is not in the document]
Jc2k: d598c3cfb0fadc31c2d25d99bfa7ce7fe015846b seems to work. I could not find a case where it broke, but.. Is it right??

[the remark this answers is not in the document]
marcelveldt: We're actually cheating in the frontend and secretly send the enabled and name within the values ;-)
https://github.com/music-assistant/frontend/blob/main/src/views/settings/EditPlayer.vue#L115

So I think we should either fix that or create an extra type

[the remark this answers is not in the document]
marcelveldt: Let's keep it this way and fix the api to send enabled and name as separate params instead of hiding them in the config values. Either that or make enabled and name actual config entries. Now its a bit mangled which I agree is not great, your confusion here in the typing is a proof of that.

marcelveldt (approved):
Impressive, nice work!

PR #1434, opened 2024-07-01
Snapcast: Add delay, before so that the stream does not idle immediately

SantiagoSotoC:
2000 ms seems to work fine, but I'm worried about a song with more than 2 seconds of silence in the middle.
[https://github.com/music-assistant/hass-music-assistant/issues/2402](https://github.com/music-assistant/hass-music-assistant/issues/2402)


marcelveldt:
Why not just bump it way higher, like 30 seconds ? We control the stream anyway and clean it up when needed right?

SantiagoSotoC:
Yes, we could. I don't know how it will work when I finish the metadata, I'm going slowly, it's tedious.

The other way is to make it configurable and mark in the documentation how to use it.

marcelveldt (approved):

SantiagoSotoC:
I have to change one verification, due to changes in the other PR

PR #1436, opened 2024-07-02
Don't load_provider if provider already deleted

Jc2k:
Our retry logic never stops trying to load a provider until it works. Even after a provider is removed.

Because failures are captured with `self.config.set`, this causes corruption where we re-insert a nearly empty dict for an instance id:

```
  "providers": {
    "snapcast--e8gUtVPG": {
      "last_error": "Unable to start the Snapserver connection ?"
    }
  }
```

This PR changes things to make sure a provider still exists in the config system before trying to load_provider. The corruption never happens, and the looping stops.

The problem with this approach is that it requires a ProviderConfig to be stored *before* we can call load_provider. This is a change in behaviour.

### DIFF
--- a/music_assistant/server/controllers/config.py
+++ b/music_assistant/server/controllers/config.py
@@ -802,11 +802,12 @@ class ConfigController:
         )
         # validate the new config
         config.validate()
-        # try to load the provider first to catch errors before we save it.
-        await self.mass.load_provider(config, raise_on_error=True)
-        # the load was a success, store this config
+
         conf_key = f"{CONF_PROVIDERS}/{config.instance_id}"
         self.set(conf_key, config.to_raw())
+
+        await self.mass.load_provider(instance_id, raise_on_error=True)
+
         return config
 
     async def _load_provider_config(self, config: ProviderConfig) -> None:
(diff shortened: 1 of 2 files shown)


marcelveldt (dismissed):
Nice!

Jc2k:
Are you ok with changing the load/save order?

marcelveldt:
> Are you ok with changing the load/save order?

Yeah, but we maybe have to simulate a few edge cases, such as setting up a provider and then fail in that setup to test if we recover from that. I believe the order was that way to prevent a half baken config to land up in the storage but now you handle that later in the process so see what it brings.

Jc2k:
This actually has a horrible regression, caused by changing the save/load order. To simulate failure I broke jellyfin. Now it created a new save (with a new instance_id) every time i hit save.. So i suddenly had multiple Jellyfin instances. I reworked it a bit to preserve the old order and that is fixed.

So now theres 2 API's - one that takes a ProviderConfig and *does not* retry. It raises, and is meant to be used when adding a new provider instance **only**. Then the 2nd, which takes an instance_id and checks the provider still exists in settings.

marcelveldt:
yeah, this makes a lot of sense!

marcelveldt (approved):

PR #1440, opened 2024-07-03
Resolve radio browse issue

max246:
Part of the issue #2553 I have updated the way the system look for the country code.
Previously the limit was not allowing to search in the whole list of countries.

I have updated the way we retrive the countries to have a true order since the radios package is not doing it correctly, for example United Kingdom was after Switzerland instead of being on the bottom with United states.

I have reduced the cache since it is a bit odd to have a 7 day cache, which could be helpful but since there is not API limit we can query a bit more frequent

### DIFF
--- a/music_assistant/server/providers/radiobrowser/__init__.py
+++ b/music_assistant/server/providers/radiobrowser/__init__.py
@@ -6,20 +6,17 @@ from typing import TYPE_CHECKING
 
 from radios import FilterBy, Order, RadioBrowser, RadioBrowserError
 
-from music_assistant.common.models.enums import LinkType, ProviderFeature, StreamType
-from music_assistant.common.models.media_items import (
-    AudioFormat,
-    BrowseFolder,
-    ContentType,
-    ImageType,
-    MediaItemImage,
-    MediaItemLink,
-    MediaItemType,
-    MediaType,
-    ProviderMapping,
-    Radio,
-    SearchResults,
-)
+from music_assistant.common.models.enums import (LinkType, ProviderFeature,
+                                                 StreamType)
+from music_assistant.common.models.media_items import (AudioFormat,
+                                                       BrowseFolder,
+                                                       ContentType, ImageType,
+                                                       MediaItemImage,
+                                                       MediaItemLink,
+                                                       MediaItemType,
+                                                       MediaType,
+                                                       ProviderMapping, Radio,
+                                                       SearchResults)
 from music_assistant.common.models.streamdetails import StreamDetails
 from music_assistant.server.controllers.cache import use_cache
 from music_assistant.server.models.music_provider import MusicProvider
@@ -27,11 +24,9 @@ from music_assistant.server.models.music_provider import MusicProvider
 SUPPORTED_FEATURES = (ProviderFeature.SEARCH, ProviderFeature.BROWSE)
 
 if TYPE_CHECKING:
-    from music_assistant.common.models.config_entries import (
-        ConfigEntry,
-        ConfigValueType,
-        ProviderConfig,
-    )
+    from music_assistant.common.models.config_entries import (ConfigEntry,
+                                                              ConfigValueType,
+                                                              ProviderConfig)
     from music_assistant.common.models.provider import ProviderManifest
     from music_assistant.server import MusicAssistant
     from music_assistant.server.models import ProviderInstanceType
@@ -94,15 +89,15 @@ class RadioBrowserProvider(MusicProvider):
         result = SearchResults()
         if MediaType.RADIO not in media_types:
             return result
+        
+        search_result = await self.radios.search(name=search_query, limit=limit)
 
-        searchresult = await self.radios.search(name=search_query, limit=limit)
-
-        for item in searchresult:
+        for item in search_result:
             result.radio.append(await self._parse_radio(item))
 
         return result
 
-    @use_cache(86400 * 7)
+    @use_cache(60)
     async def browse(self, path: str, offset: int, limit: int) -> list[MediaItemType]:
         """Browse this provider's items.
 
@@ -162,7 +157,7 @@ class RadioBrowserProvider(MusicProvider):
         if subpath == "country":
             items: list[BrowseFolder | Radio] = []
             for country in await self.radios.countries(
-                order=Order.NAME, hide_broken=True, limit=limit, offset=offset
+                order=Order.NAME, hide_broken=True
             ):
                 folder = BrowseFolder(
                     item_id=country.code.lower(),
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/radiobrowser/__init__.py
```diff
@@ -94,15 +89,15 @@ async def search(
...
-        searchresult = await self.radios.search(name=search_query, limit=limit)
-
-        for item in searchresult:
+        for item in search_result:
             result.radio.append(await self._parse_radio(item))
 
         return result
 
-    @use_cache(86400 * 7)
+    @use_cache(60)
```
marcelveldt, line 100: why ? lets be careful with doing many requests. Is it really true that the output of the radiobrowser api changes often ? Maybe make this a 1 hour cache at least, not 60 seconds!

music_assistant/server/providers/radiobrowser/__init__.py
```diff
@@ -179,21 +174,20 @@ async def browse(self, path: str, offset: int, limit: int) -> list[MediaItemType
                     )
                 ]
                 items.append(folder)
-            return items
-
-        if subsubpath in await self.get_tag_names(limit=limit, offset=offset):
-            return await self.get_by_tag(subsubpath)
+            return items[offset:(offset+limit)]
```
marcelveldt, line 177: this bad - this fetches all data and then truncates it - then you can better leave out all paging completely

music_assistant/server/providers/radiobrowser/__init__.py
```diff
@@ -203,10 +197,10 @@ async def get_tag_names(self, limit: int, offset: int):
             tag_names.append(tag.name.lower())
         return tag_names
 
-    async def get_country_codes(self, limit: int, offset: int):
+    async def get_country_codes(self):
         """Get a list of country names."""
         countries = await self.radios.countries(
-            order=Order.NAME, hide_broken=True, limit=limit, offset=offset
```
marcelveldt, line 209: why did you remove the paging ? Is it broken in the library ?

music_assistant/server/providers/radiobrowser/__init__.py
```diff
@@ -162,7 +157,7 @@ async def browse(self, path: str, offset: int, limit: int) -> list[MediaItemType
         if subpath == "country":
             items: list[BrowseFolder | Radio] = []
             for country in await self.radios.countries(
-                order=Order.NAME, hide_broken=True, limit=limit, offset=offset
```
marcelveldt, line 165: why did you remove the paging ? Is it broken in the library ?

marcelveldt (changes requested):
Thanks for the effort but this really needs more thought

music_assistant/server/providers/radiobrowser/__init__.py
max246: Yeah paging with order Is broken. As I explained in the description it doesn't properly sort the station name. Since it gets cached it should not be a problem unless you see otherwise

music_assistant/server/providers/radiobrowser/__init__.py
max246: True we can increase it, we can even do one day but 7 days in my opinion is a bit too much

music_assistant/server/providers/radiobrowser/__init__.py
max246: Unfortunately the countries method doesn't have a filter so I thought to leave the paging by slicing, but happy to remove it and get it all in once. Does the method need to add a cache decorator or it pass that over from the main method ?

music_assistant/server/providers/radiobrowser/__init__.py
max246: Unfortunately the countries method doesn't have a filter so I thought to leave the paging by slicing, but happy to remove it and get it all in once. Does the method need to add a cache decorator or it pass that over from the main method ?

max246:
> Thanks for the effort but this really needs more thought

No worries, this is working but I am happy to make amends to ensure you are happy with the changes

music_assistant/server/providers/radiobrowser/__init__.py
marcelveldt: there doesn't need to be a filter - the filter is in the frontend - you just need to return a sane amount of chunked data - the frontend will notice that the endpoint doesn't do filtering and grab all items first. For lists that have < 1000 items its fine to just return them all at once and save the paging overhead.

What version did you test this on? I think all this logic is fixed in the recent betas of MA but might be still broken on the stable version.

music_assistant/server/providers/radiobrowser/__init__.py
marcelveldt: See my reply abovve - it doesn't need a filter, we just want all the items in chunks

music_assistant/server/providers/radiobrowser/__init__.py
max246: Latest version. 
The issue is that we have like 200+ countries/tags and the frontend is asking :
"look for X in  offset 0 limit 50", then the rest is not looked up. That is why the ticket above is not showing GB radios

marcelveldt: Nope, that is not how it works. Are you on latest beta or stable ?

music_assistant/server/providers/radiobrowser/__init__.py
marcelveldt: the paging does not need the order. Sorting and filtering is done by the frontend.

max246: bit confused, I put some logging and observed it was sending:
offset=0, limit=50
offset=50, limit=100
...

but if you say the frontend is doing the sorting, there is something not working because it wasnt doing it.  I synced my  fork with main before starting last night, so I am not sure. 
I will have to test again if you think that is true.

music_assistant/server/providers/radiobrowser/__init__.py
max246: I was on the main branch, which I suppose is the stable one? 
I synced to the source main and I just worked from there, let me know if I should work on another base of the branch.
I added logs and that what was happening on my side.

marcelveldt: I just tested and its working fine on my end - hence my question how you tested this

marcelveldt:
I actually found an easier way to solve the issue by just ignoring the paging of RadioBrowser - fixed in next release.
Can you give the next beta a try ?

max246:
> I actually found an easier way to solve the issue by just ignoring the paging of RadioBrowser - fixed in next release. Can you give the next beta a try ?

Ok will try,

bjtho08:
When I go to "browse by country" and select "Denmark" I get a single station, which appears to be from Guatemala, but it has the letters "DK" in its name. I am on 2.1.0rc1. Denmark should have a list of 73 stations.

PR #1442, opened 2024-07-03
Snapcast: Reload built-in server in case of connection loss

SantiagoSotoC:
I have found that sometimes the snapserver can crash, due to lack of ram I think. Use the connection down callback. Allow 1 try. if it fails restart the server.

### DIFF
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -301,8 +301,8 @@ class SnapCastProvider(PlayerProvider):
                 "Started connection to Snapserver %s",
                 f"{self._snapcast_server_host}:{self._snapcast_server_control_port}",
             )
-            if self._use_builtin_server:
-                self._snapserver.set_on_disconnect_callback(self._restart_builtin_server)
+            # if self._use_builtin_server:
+            # self._snapserver.set_on_disconnect_callback(self._restart_builtin_server)
 
         except OSError as err:
             msg = "Unable to start the Snapserver connection ?"
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -284,7 +284,7 @@ class SnapCastProvider(PlayerProvider):
         self._ids_map = bidict({})
 
         if self._use_builtin_server:
-            self._start_builtin_server()
+            await self._start_builtin_server()
         else:
             self._snapserver_runner = None
             self._snapserver_started = None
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -318,10 +320,8 @@ async def unload(self) -> None:
             player_id = self._get_ma_id(snap_client_id)
             await self.cmd_stop(player_id)
         self._snapserver.stop()
-        if self._snapserver_runner and not self._snapserver_runner.done():
-            self._snapserver_runner.cancel()
-        await asyncio.sleep(10)  # prevent race conditions when reloading
-        self._snapserver_started.clear()
+        await asyncio.sleep(10)
```
marcelveldt, line 323: what's up with the sleep here ? That is not so nice as it will delay reload of the provider and server shutdown even.

marcelveldt, line 323:
```suggestion
```

SantiagoSotoC: Yes, my bad

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -670,3 +669,30 @@ async def _builtin_server_runner(self) -> None:
...
+        else:
+            self._builtin_server_retry += 1
+            self.logger.debug("Increase snapcast retry count")
+
+    async def _stop_builtin_server(self) -> None:
+        """Stop the built-in Snapserver."""
+        self.logger.info("Stopping, built-in Snapserver")
+        if self._snapserver_runner and not self._snapserver_runner.done():
+            self._snapserver_runner.cancel()
+            await asyncio.sleep(10)  # prevent race conditions when reloading
```
marcelveldt, line 689:
```suggestion
```

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -670,3 +669,30 @@ async def _builtin_server_runner(self) -> None:
                         # delay init a small bit to prevent race conditions
                         # where we try to connect too soon
                         self.mass.loop.call_later(2, self._snapserver_started.set)
+
+    async def _restart_builtin_server(self) -> None:
+        """Restart the built-in Snapserver."""
+        if self._use_builtin_server and self._builtin_server_retry > 1:
+            self.logger.info("Restarting, built-in Snapserver.")
+            await self._stop_builtin_server()
+            await asyncio.sleep(10)
```
marcelveldt, line 678:
```suggestion
            await asyncio.sleep(10)  # prevent race conditions when reloading
```

marcelveldt (approved):

PR #1446, opened 2024-07-05
Fix some bugs caused by regression

marcelveldt:
- Fix RadioBrowser browsing 
- Fix mounting SMB paths
- Fix Airplay playback

### DIFF
--- a/music_assistant/server/helpers/process.py
+++ b/music_assistant/server/helpers/process.py
@@ -257,20 +257,13 @@ class AsyncProcess:
         return self._returncode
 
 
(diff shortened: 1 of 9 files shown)


music_assistant/server/helpers/process.py
```diff
@@ -257,20 +257,13 @@ async def wait(self) -> int:
         return self._returncode
 
 
-async def check_output(args: str | list[str]) -> tuple[int, bytes]:
+async def check_output(*args: list[str]) -> tuple[int, bytes]:
```
Jc2k, line 260: Think this should be *args: str?

PR #1447, opened 2024-07-05
Reload Snapcast provider when connection to the server gets lost

marcelveldt:
Another implementation for reloading snapcast when the connection to the server gets lost, which will work for both the builtin and external servers. It will also prevent a race condition when the connection drops

### DIFF
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -229,7 +229,6 @@ class SnapCastProvider(PlayerProvider):
     _snapserver_runner: asyncio.Task | None
     _snapserver_started: asyncio.Event | None
     _ids_map: bidict  # ma_id / snapclient_id
-    _builtin_server_retry: int
 
     def _get_snapclient_id(self, player_id: str) -> str:
         search_dict = self._ids_map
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -691,7 +678,14 @@ async def _stop_builtin_server(self) -> None:
...
             self._snapserver_runner = asyncio.create_task(self._builtin_server_runner())
             await asyncio.wait_for(self._snapserver_started.wait(), 10)
+
+    def _handle_disconnect(self, exc: Exception) -> None:
+        """Handle disconnect callback from snapserver."""
+        self.logger.info(
+            "Connection to SnapServer lost, reason: %s. Reloading provider in 5 seconds.", str(exc)
+        )
+        # schedule a reload of the provider
+        self.mass.call_later(5, self.mass.config.reload_provider(self.instance_id))
```
Jc2k, line 691: Bit of an edge case, but if the provider is removed while the timer is waiting, reload_provider explodes. (It calls get_provider_config, which will throw a KeyError if the provider was removed).

I think maybe just make it fail silently?:

```
class ConfigController:
    ...
    @api_command("config/providers/reload")
    async def reload_provider(self, instance_id: str) -> None:
        """Reload provider."""
        try:
            config = await self.get_provider_config(instance_id)
        except KeyError:
            # Provider was removed before we could reload it
            return
        await self._load_provider_config(config)```

?

marcelveldt: yeah - sounds reasonable

PR #1448, opened 2024-07-05
Better handle of SMB mount with special characters in username/password

marcelveldt:
- Fix smb mount on MacOS
- Fix mount command if the username or password includes quotes

To handle special characters in the username and/or password, they need to be quoted correctly but the username/password can also contain quotes on its own, so our previous solution wasnt fail proof

Using environmental variables it works - it accepts all the special characters now
Only MacOS doesn't work but as that is dev only, we shouldn't care

### DIFF
--- a/music_assistant/server/helpers/process.py
+++ b/music_assistant/server/helpers/process.py
@@ -257,12 +257,10 @@ class AsyncProcess:
         return self._returncode
 
 
-async def check_output(*args: str) -> tuple[int, bytes]:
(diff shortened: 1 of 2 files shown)


marcelveldt:
This is still not working ast should - maybe we should just deny special characters like quotes, ampersand and spaces

marcelveldt:
Using environmental variables it works - it accepts all the special characters now
Only MacOS doesn't work but as that is dev only, we shouldn't care

PR #1449, opened 2024-07-05
Add typing for radiobrowser

Jc2k:


music_assistant/server/models/music_provider.py
```diff
@@ -282,7 +283,7 @@ async def get_item(self, media_type: MediaType, prov_item_id: str) -> MediaItemT
             return await self.get_radio(prov_item_id)
         return await self.get_track(prov_item_id)
 
-    async def browse(self, path: str, offset: int, limit: int) -> list[MediaItemType]:
+    async def browse(self, path: str, offset: int, limit: int) -> Sequence[MediaItemType]:
```
Jc2k, line 286: list is invariant, so you can't return a list[Radio] directly from browse. mypy doesn't know that something else won't append to the list, and it wouldn't be safe to append an Artist to a list[Radio]. Sequence is a read only list, so can be treated as covariant. This solves been able to directly return a list[Radio] from browse without doing any weird casting.

I'll need to update all browse() eventually, but will do that as i convert each provider.

marcelveldt (approved):
Nice work again!

PR #1460, opened 2024-07-06
Snapcast: revert dryout and set state by hand

SantiagoSotoC:
Dryout is a poorly documented function, I have suspicions that it does not work correctly with tcp streams.
It is better to ignore the state of the snap_stream and set the state by hand, this ensures that the correct song is displayed.


SantiagoSotoC:
I left music running in the dev addon, if in one hour the information is not delayed I will update here

SantiagoSotoC:
Well it's been playing for 4 hours now and it doesn't lose sync.

marcelveldt (approved):
LGTM

PR #1461, opened 2024-07-07
Add typing for filesystem_local

Jc2k:


music_assistant/server/models/music_provider.py
```diff
@@ -249,6 +251,7 @@ async def get_stream_details(self, item_id: str) -> StreamDetails:
         """Get streamdetails for a track/radio."""
         raise NotImplementedError
 
+    @abc.abstractmethod
```
marcelveldt, line 254: be aware that this is an optional method to implement so marking it as abstractmethod may lead to errors in providers that do not implement it.

Jc2k: Added it to avoid this:

```
music_assistant/server/providers/filesystem_local/base.py:672: error: Return type "AsyncGenerator[bytes, None]" of "get_audio_stream" incompatible with return type "Coroutine[Any, Any, AsyncGenerator[bytes, None]]" in supertype "MusicProvider"  [override]
```

The other suggestion is to use a `def` instead of an `async def` that returns an `AsyncIterable`. (https://github.com/python/mypy/issues/5385). Which is super grim IMO. Or have a dummy yield.

Gone with the dummy yield this time.

marcelveldt: Yeah dummy yield is the way to go for this - I think we've done that in more places

marcelveldt (approved):
Nice!

PR #1462, opened 2024-07-07
Apple Music: Fix albums without name, increase cooldown periode on 504 gateway timeouts.

MarvinSchenkel:
Fix albums without name. Increase cooldown periode on 504 timeout. Also add some more logging to the 504 so we can see the request that is timing out

### DIFF
--- a/music_assistant/server/providers/apple_music/__init__.py
+++ b/music_assistant/server/providers/apple_music/__init__.py
@@ -114,7 +114,7 @@ class AppleMusicProvider(MusicProvider):
     _decrypt_private_key: bytes | None = None
     # rate limiter needs to be specified on provider-level,
     # so make it an instance attribute
-    throttler = ThrottlerManager(rate_limit=1, period=2)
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):

MarvinSchenkel:
Do they fail after 5 retries? Or do you see it picking up after 2 or 3 retries? I have not been able to find a root cause. Might be Apple.. Might be some edge cases in your library (e.g. purchased music, custom uploaded music etc..).

faelenor:
They always fail after 5 retries. I've never been able to sync any track.

MarvinSchenkel:
Alright, could you please open an issue and provide some debug logs? Maybe we can find a pattern and figure out what is going wrong. Please also describe your library, does it just consists of online content, did you upload custom stuff, did you purchase music etc...

faelenor:
Ok, I'll do this, thanks! Do you know why the logs are always truncated? I can only see the last couple of KBs, both in the addon page and the settings/system/log section.

PR #1465, opened 2024-07-07
Fix crash on missing names for album, artist or track

xmirakulix:
Avoid exceptions if apple music library contains tracks, albums or artists without names.

For more info see comment in https://github.com/music-assistant/hass-music-assistant/issues/2431#issuecomment-2212468656


xmirakulix:
duplicate of https://github.com/music-assistant/server/pull/1457

PR #1468, opened 2024-07-07
Fix Jellyfin similar tracks

Jc2k:
Somehow lost this in a merge or a rebase. If we don't specific the "extra" fields there's not enough data for _parse_track to work off.


marcelveldt (approved):
LGTM

PR #1469, opened 2024-07-07
Jellyfin: Skip tracks with no MediaStreams

Jc2k:
Towards https://github.com/music-assistant/hass-music-assistant/issues/2594. It's not clear why this library has tracks with no MediaStreams - maybe API misuse on our part, maybe a corrupt library, maybe a setting. But lets log such tracks and allow the sync to continue.


marcelveldt (approved):
LGTM!

PR #1470, opened 2024-07-07
Add type checking for filesystem_smb

Jc2k:


music_assistant/server/providers/filesystem_smb/__init__.py
```diff
@@ -171,13 +171,13 @@ async def get_audio_stream(
...
-        password: str = self.config.get_value(CONF_PASSWORD)
-        share: str = self.config.get_value(CONF_SHARE)
+        server = str(self.config.get_value(CONF_HOST))
+        username = str(self.config.get_value(CONF_USERNAME))
+        password = str(self.config.get_value(CONF_PASSWORD))
+        share = str(self.config.get_value(CONF_SHARE))
 
         # handle optional subfolder
-        subfolder: str = self.config.get_value(CONF_SUBFOLDER)
+        subfolder = str(self.config.get_value(CONF_SUBFOLDER))
```
marcelveldt, line 180: this will turn None into a string "None"

music_assistant/server/providers/filesystem_smb/__init__.py
```diff
@@ -171,13 +171,13 @@ async def get_audio_stream(
 
     async def mount(self) -> None:
         """Mount the SMB location to a temporary folder."""
-        server: str = self.config.get_value(CONF_HOST)
-        username: str = self.config.get_value(CONF_USERNAME)
-        password: str = self.config.get_value(CONF_PASSWORD)
-        share: str = self.config.get_value(CONF_SHARE)
+        server = str(self.config.get_value(CONF_HOST))
+        username = str(self.config.get_value(CONF_USERNAME))
+        password = str(self.config.get_value(CONF_PASSWORD))
```
marcelveldt, line 176: this will turn None into a string "None"
I'm not sure if we allow empty passwords though

music_assistant/server/providers/filesystem_smb/__init__.py
marcelveldt: Maybe we should simply set the default value to an empty string

Jc2k: Good spot. If I revert the change and the value is None, then it would get included in the f-string on 193 and 209 and turn into the string "None" also. So I think default value of empty string is the way to go. Actually, subfolder already has a default value of "". So all good as is, I think

music_assistant/server/providers/filesystem_smb/__init__.py
Jc2k: Empty passwords seem to be allowed. Have updated.

marcelveldt (approved):
Nice, another one fully typed now!

PR #1471, opened 2024-07-08
Replace windows separators in m3u

MelHarbour:
Fixes https://github.com/music-assistant/hass-music-assistant/issues/2432

Replace instances of backslash with forward slashes when parsing m3u files.


marcelveldt (dismissed):

marcelveldt:
See the lint error, you need to add a ruff exclusion here.

Tip: Please install pre-commit on your dev environment so you catch these errors already upfront.
Use the setupscript (scripts\setup.sh) before developing to get the venv in the correct shape

MelHarbour:
Yeah, I need to get my dev env in shape.

marcelveldt (approved):
Thanks!

PR #1474, opened 2024-07-08
Add typing for test provider

Jc2k:


music_assistant/server/providers/test/__init__.py
```diff
@@ -150,19 +151,17 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
 
     async def get_stream_details(self, item_id: str) -> StreamDetails:
         """Get streamdetails for a track/radio."""
-        media_info = await self._get_media_info(item_id)
```
Jc2k, line 153: Function didn't exist so i've hardcoded some dummy data for now.

marcelveldt: I hacked this provider in a while ago just to simply create a dummy db with 1 million records to do performance testing. I'm surprised this is all you found :-)

Anyways, my idea with this provider was that maybe we could use it to do some unit/integration tests at some point to test he base logic and models or at least use it as some sort of example provider. (hmm, maybe the name should have been example provider instead of test then)

Jc2k: I had the same thought :) want to add a test player as well.

Want to get to tests, but want to try and get providers fully typed first.

marcelveldt (approved):
Bam, yet another one typed. You're on fire.

PR #1482, opened 2024-07-09
Add typing for fullykiosk

Jc2k:


music_assistant/server/providers/fully_kiosk/__init__.py
```diff
@@ -146,22 +146,22 @@ async def loaded_in_mass(self) -> None:
...
-                break
-        current_url = self._fully.deviceInfo.get("soundUrlPlaying")
-        player.current_item_id = current_url
-        if not current_url:
-            player.state = PlayerState.IDLE
-        player.available = True
-        self.mass.players.update(player_id)
-
-    async def get_player_config_entries(self, player_id: str) -> tuple[ConfigEntry]:
+        if player := self.mass.players.get(player_id):
```
Jc2k, line 149: I've made a bunch of those fail silently on the assumption that such the player being missing would happen when we were unloading a player. Maybe raising `PlayerUnavailableError` would be better in some cases? I'm not sure.

marcelveldt: mass.players.get has an optional argument to raise on unavailable but ofc mypy has no knowledge of that

I think just doing this None check is fine - probbaly really some edgecases we're talking about

marcelveldt, line 149: you can also turn this into a guard clause instead:

```suggestion
        if(player := self.mass.players.get(player_id)) is None:
            return
```

marcelveldt (approved):

PR #1485, opened 2024-07-09
Fix MusicBrainz external ID match

marcelveldt:
We were using MusicBrainz ReleaseGroup ID as a strong identifier which caused different versions of the same album to be linked together. The strong identifier for a unique album(version) should be the MusicBrainzAlbumID and the ReleaseGroupId should only be used to lookup external (meta)data. 

This PR makes this more explicit by literally naming the musicbrainz external ID's.

Note that when no MusicBrainz tags are present (e.g. streaming providers), the logic will still fallback to just name comparing to catch different versions


marcelveldt:
@Jc2k the test of jellyfin doesnt seem to like the fact that there are now multiple external id's returned.
Because external_ids is a set, the order will be random and the test will simply fail sometimes when the order is not what is expected in the snapshot file. Any ideas ?

marcelveldt:
For now I've just adjusted the tests to sort the external_ids after converting to dict.
Maybe at some point if we want to make this more future proof we could decide to implement a serialization strategy:
https://github.com/Fatal1ty/mashumaro?tab=readme-ov-file#serialization_strategy-config-option

Jc2k:
I thought syrupy would handle sets out of the box. Will add a proper strategy soon.

PR #1490, opened 2024-07-11
Subsonic: Workaround for Navidrome Various Artists track handling

khers:
Navidrome does not create an artist entry for any artist that does not have an ablum credited to them. This means that for many tracks on a Various Artists album, the instance may not have artist ID's for the artist named on the track. This patch hacks around this by creating a 'fake' artist ID for them and handling it in the provider. If the ID we create conflicts with a subsonic server's ID creation scheme we can revisit.


marcelveldt (approved):
LGTM

PR #1492, opened 2024-07-11
Update name to clarify difference with the integration

OzGav:
Hopefully this is the place make the name change!


marcelveldt (approved):

marcelveldt:
It also needs to be adjusted in the add-on config (for all 3 addons)
https://github.com/music-assistant/home-assistant-addon/blob/main/music_assistant/config.yaml

PR #1493, opened 2024-07-12
YT Music: Fix token expiration on reboot

MarvinSchenkel:
This PR refreshes the cache of YT DLP on startup to create authenticated requests, since the cache gets removed on reboot, causing the provider not to get loaded.

NB: I am planning to write my own yt dlp plugin that accept an access token to authenticate to yt dlp, rather than this 3rd party oauth plugin.


marcelveldt (approved):
LGTM

PR #1497, opened 2024-07-14
YTMusic: Rewrite YTM auth to ytdlp access token plugin

MarvinSchenkel:
This PR replaces the hacky ytdlp oauth plugin with my own plugin. Instead of having to abuse the yt dlp cache to pass an access token to the plugin, we can now pass the MA access token directly to be injected in the yt dlp requests.

### DIFF
--- a/music_assistant/server/providers/ytmusic/manifest.json
+++ b/music_assistant/server/providers/ytmusic/manifest.json
@@ -4,7 +4,7 @@
   "name": "YouTube Music",
   "description": "Support for the YouTube Music streaming provider in Music Assistant.",
   "codeowners": ["@MarvinSchenkel"],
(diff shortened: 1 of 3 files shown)


music_assistant/server/providers/ytmusic/manifest.json
```diff
@@ -4,7 +4,7 @@
   "name": "YouTube Music",
   "description": "Support for the YouTube Music streaming provider in Music Assistant.",
   "codeowners": ["@MarvinSchenkel"],
-  "requirements": ["ytmusicapi==1.7.3", "git+https://github.com/coletdjnz/yt-dlp-youtube-oauth2.git", "yt-dlp==2024.7.7"],
+  "requirements": ["ytmusicapi==1.7.3", "git+https://github.com/MarvinSchenkel/yt-dlp-youtube-accesstoken.git", "yt-dlp==2024.7.7"],
```
marcelveldt, line 7: please convert this to a pypi package before the release

marcelveldt (approved):

PR #1520, opened 2024-07-23
Jellyfin: Fix artist enumeration

Jc2k:
At some point I had switched to `/Items` instead of `/Artists` for consistency with every other object type and it seemed to work. However it was only returning a subset of artists. This reverts that, which restores full sync speed.

https://github.com/Jc2k/aiojellyfin/releases/tag/v0.9.2


marcelveldt (approved):

marcelveldt:
Added it to the 2.1.1 milestone assuming you want this backported to te stable release?

Jc2k:
Yes please!

PR #1521, opened 2024-07-23
Jellyfin: Use aiojellyfin 0.10.0 for stricter typing and more speed

Jc2k:
* The latest aiojellyfin uses the builder pattern. Mostly this is because the number of arguments the `/Items` API has is huuuge and my function bodies were getting ugly.
* Along with that, some of the fields become more strictly typed (so for example the `fields` query arg now has an enum instead of being text).
* Pagination moved out of MA and into the backing library, where its implemented once instead of 4 times. I also added a new prefetching strategy. It takes time to process a page of results and insert them in the MA db. On one of my older dev boxes that takes long enough that we can prefetch the next page.

(This isn't for backporting, though it could be after some time to brew up I guess).


marcelveldt (approved):
Cool!

PR #1523, opened 2024-07-24
Subsonic: Add debug level logging around stream start, end, and details

khers:
We need to try and debug why some M4A files are failing from certain providers and not others and some extra info is needed. Add debug level logging around stream start and end as well as when stream details are requested.

### DIFF
--- a/music_assistant/server/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/server/providers/opensubsonic/sonic_provider.py
@@ -792,6 +792,13 @@ class OpenSonicProvider(MusicProvider):
         if mime_type.endswith("mpeg"):
             mime_type = sonic_song.suffix
 
+        self.logger.debug(
+            "Fetching stream details for id %s '%s' with format '%s'",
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/opensubsonic/sonic_provider.py
```diff
@@ -792,6 +792,11 @@ async def get_stream_details(self, item_id: str) -> StreamDetails:
         if mime_type.endswith("mpeg"):
             mime_type = sonic_song.suffix
 
+        logging.getLogger("libopensonic").debug(
```
Jc2k, line 795: I think this should probably use `self.logger`.

Logging generally does not use f-strings. This is a small efficiency saving as you can defer the string concatenation (and associated memory allocations), not even doing it at all if the debug log is off.

Jc2k: Yep, looks great to me.

marcelveldt (approved):

PR #1524, opened 2024-07-25
Spotify: remove deprecated user-read-birthdate scope

chatziko:
This tiny PR removes the `user-read-birthdate` scope which seems to be deprecated (see for instance https://github.com/jbszczepaniak/spotify-cli/pull/9). Spotify probably ignores it for old apps to avoid breaking them, but if a different client-id is used authentication fails unless `user-read-birthdate` is removed.


marcelveldt (approved):
Thanks!

PR #1525, opened 2024-07-26
Typing: Tidal

Jc2k:


music_assistant/server/providers/tidal/__init__.py
```diff
@@ -606,7 +611,7 @@ async def get_track(self, prov_track_id: str) -> Track:
             track = self._parse_track(track_obj)
             # get some extra details for the full track info
             with suppress(tidal_exceptions.MetadataNotAvailable, AttributeError):
-                lyrics: TidalLyrics = await asyncio.to_thread(track.lyrics)
+                lyrics: TidalLyrics = await asyncio.to_thread(track_obj.lyrics)
```
Jc2k, line 614: I think this is a real bug, so maybe the AttributeError isn't actually needed?

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -864,27 +875,32 @@ def _parse_playlist(self, playlist_obj: TidalPlaylist) -> Playlist:
...
+        func: Callable[_P, list[_R]] | Callable[_P, Awaitable[list[_R]]],
+        *args: _P.args,
+        **kwargs: _P.kwargs,
+    ) -> AsyncGenerator[_R, None]:
         """Yield all items from a larger listing."""
         offset = 0
         while True:
             if asyncio.iscoroutinefunction(func):
-                chunk = await func(*args, **kwargs, offset=offset)
+                chunk = await func(*args, **kwargs, offset=offset)  # type: ignore[arg-type]
```
Jc2k, line 901: `Concatenate` doesn't seem to be flexible enough to handle this case, and there doesn't seem to be a way to add a constraint to the ParamSpec to verify `offset` is actually available in the wrapped function.

music_assistant/server/providers/tidal/__init__.py
```diff
@@ -172,16 +177,18 @@ async def get_config_entries(
     action: [optional] action key called from config entries UI.
     values: the (intermediate) raw values for config entries sent with the action.
     """
+    assert values is not None
```
Jc2k, line 180: Not familiar enough with MA's take on config entries to know if this is sane. Further down it assumes values is not None without testing, and AFAICT if values was None in the next few chunks it would pass invalid data into `tidal_auth_url` and `tidal_pkce_login`. Without this restriction, we end up passing quality (and others) as a `str|None` instead of just `str`.

jozefKruszynski:
Honestly everything looks very sensible, however, I can't test anything for the next couple of weeks whilst I'm away from home.

music_assistant/server/providers/tidal/__init__.py
marcelveldt: yeah this makes sense

marcelveldt:
@jozefKruszynski can we move forward on this one ?

jozefKruszynski:
Hmm, sorry work has been busy busy since getting back from holiday and I have not wanted to face the computer afterwards yet.
I guess we'll need to rebase the branch and run some local tests etc.
I can look at this tomorrow evening as the wife is off out with friends.

jozefKruszynski:
Just did some quick testing and it all appears to be good. I'll do a bit more later tonight actually

jozefKruszynski:
@marcelveldt Everything seems to run very nicely, from my side I would say the changes seem solid. However, I can't speak to the overall code, as I still feel like a complete novice with Python overall.

@Jc2k Can you check the conflicts and fix those up please.

Jc2k:
Done

marcelveldt (approved):
LGTM, Nice work again @Jc2k !

PR #1526, opened 2024-07-27
Throttler: sleep without busy wait, log delayed calls

chatziko:
This PR adds two improvements to throttling:

1. Accurate sleep until the next available time. `asyncio_throttle` by default wakes every 10msecs which is crazy. `ThrottlerManager` sets the wait to 100mecs, which is still unnecessary, but some modules don't use `ThrottlerManager` anyway.

2. Return the delay caused by the throttler (0 means no throttling) and log it in `throttle_with_retries`.

Since `asyncio_throttle` is very simple and looks unmaintained (PR ignored for years) I copied the code in `server/helpers/throttle_retry.py`, no point having an unmaintained dependency for a few lines of code.

### DIFF
--- a/music_assistant/server/helpers/throttle_retry.py
+++ b/music_assistant/server/helpers/throttle_retry.py
@@ -3,10 +3,10 @@
 import asyncio
 import functools
 import logging
+import time
+from collections import deque
 from collections.abc import Awaitable, Callable, Coroutine
 from typing import TYPE_CHECKING, Any, Concatenate, ParamSpec, TypeVar
-
-from asyncio_throttle import Throttler
 
 from music_assistant.common.models.errors import ResourceTemporarilyUnavailable, RetriesExhausted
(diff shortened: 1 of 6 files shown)


marcelveldt (dismissed):
Looks good, nice work!

marcelveldt:
CI found an issue in the code, can you check that please ?

Tip: to setup the dev environment use the scripts\setup.sh

chatziko:
That's strange, the pre-commit checks are passing in my vscode dev container. Even stranger is that the failing line was not changed by me but by the ruff formatter! But now I ran the formatter again and it didn't make the same change. Weird...

Anyway I recommitted without the formatting changes and it seems to pass now.

marcelveldt:
Maybe a version mismatch of the ruff formatter in your dev container and CI ?

marcelveldt (approved):
Nice work, thanks!

PR #1549, opened 2024-08-09
Fix issue with track linking

marcelveldt:
Fix a small oversight in the track linking logic that could result in a track being linked to an alternative version produced by another artist in some conditions.


OzGav:
Is this fixing this https://discord.com/channels/753947050995089438/1098863188193325136/1271104287799251107

PR #1564, opened 2024-08-13
Soundcloud fix

gieljnssns:
I hope this fixes the soundcloud issues


marcelveldt:
I'm not entirely sure if just ignoring HLS streams can be called a fix. The whole world is moving to HLS streams and I can imagine Soundcloud is as well and at some point the HLS/MPD streams will be higher quality than legacy ones or HLS will be the only option left.

So I would propose that its better to fix the issues with HLS (make them actually work) rather than just filtering them out.
As a temporary measure you could buy some time by just changing the order for now and prefer legacy streams over HLS.

marcelveldt:
I just fixed the HLS streams for soundcloud in another PR so this PR is superseded

PR #1570, opened 2024-08-16
Implement Spotify (oauth) PKCE Login flow

marcelveldt:
This implements a full oauth flow for Spotify authentication which is used for both the library management and the streaming (librespot).

Instead of relying on librespot for authentication, we now handle all authentication ourselves using the regular/recommended PKCE flow from Spotify and we pass the Bearer token to Librespot when we need to playback.


marcelveldt:
@OzGav FYI !

PR #1573, opened 2024-08-17
Change logger used for provider

khers:
This provider was using a mix of self.logger as well as creating it's
own outside the MA infrastructure. The correct thing to do is have the
provider log to self.logger so change all instances to that. This way
all provider logging respects the log level set either globally or for
this provider.


marcelveldt (changes requested):
This is not an acceptable solution

khers:
So should all the logging be changed away from this?

marcelveldt (approved):

PR #1599, opened 2024-08-23
Guard against eyed3 returning None

dmcc:
In some cases, `eyed3` will return `None` instead of a file with empty tags. Fixes [issue #2793](https://github.com/music-assistant/hass-music-assistant/issues/2793).


marcelveldt (approved):
Thanks!

PR #1632, opened 2024-08-30
Add dev manual build

jozefKruszynski:
Split the main pypi publish into 2 jobs with artifact upload/download
in-between. Added both jobs as dependencies to the release-notes and
addons update jobs, and made a start on a manually triggered dev build
with git hash as version tag

Added new build args with a MASS_INSTALL_LOCATION and DEV_RELEASE to
manipulate how and where the mass wheel is installed in the Dockerfile. Both have defaults that should cause the standard behaviour.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,6 +95,11 @@ jobs:
       packages: write
     needs: build-artifact
     steps:
+      - name: Retrieve release distributions
+        uses: actions/download-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
       - uses: actions/checkout@v4.1.4
       - name: Download Widevine CDM client files from private repository
         shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -95,7 +127,7 @@ jobs:
           echo "major=${patch%.*.*}" >> $GITHUB_OUTPUT
       - name: Build and Push release
         uses: docker/build-push-action@v6.7.0
-        if: github.event.release.prerelease == false
+        if: !github.event.workflow_dispatch && github.event.release.prerelease == false
```
marcelveldt, line 130: does this work in this syntax here ?

jozefKruszynski: I honestly don't know, and can't seem to find reliable info about it

marcelveldt (approved):
Nice work again!

PR #1634, opened 2024-09-01
Album track sort order - Add track name as fallback when track number is missing

jdeksup:
When using an album without track numbers set, track order is random.
With this fix track name is used as fallback to sort files (generally name starting with 01, 02, 03 etc...)

### DIFF
--- a/music_assistant/server/controllers/media/albums.py
+++ b/music_assistant/server/controllers/media/albums.py
@@ -227,7 +227,7 @@ class AlbumsController(MediaControllerBase[Album]):
         result: UniqueList[Track] = UniqueList(db_items)
         if in_library_only:
             # return in-library items only
-            return sorted(db_items, key=lambda x: (x.disc_number, x.track_number))
+            return sorted(db_items, key=lambda x: (x.disc_number, x.track_number, x.sort_name))
 
         # return all (unique) items from all providers
         # because we are returning the items from all providers combined,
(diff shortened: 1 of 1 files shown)


marcelveldt (changes requested):
I'm sorry but this is not a correct solution. If the track number is missing, it should be added to the source/provider. If the track number is present in the filename, we could perhaps parse it from the track filename.

marcelveldt:
> I understand in a way that it is not a standard solution, i propose that just as an unofficial improvement which has no cost and is still better than a random order based on database row order which is bad (internal database id could be also a solution as it will be the insert order). I saw also in several metadata provider that the track number is sometimes not set.

No, I don't agree on that. You should fix an issue at the source not hack around it.
If you expect disc/track sort and then get name sort, that is just as bad as db insertion order.

If this needs to be fixed, it needs to be fixed at music provider level.
Be aware that the place you now tried to fix this, is no longer aware of the filename.

I assume you are using local file provider and you have some files on disk you want in MA. Apparently you didnt set any/all id3 tags. So, in that case you could maybe extend the tag parser to allow reading the track number from the filename;
https://github.com/music-assistant/server/blob/main/music_assistant/server/helpers/tags.py#L182

marcelveldt:
@jdeksup do you have a few examples of files with the track number prefix ?

And to repeat the question, I assume you are using a local fileprovider in MA right ?

marcelveldt:
superseded by #1663

PR #1642, opened 2024-09-03
Some small bugfixes and tweaks

marcelveldt:


OzGav (commented):
Will be interesting to see if this has any adverse matching effect. Two seconds seems reasonable.

OzGav:
For the radiobrowser change I will add this limit to the docs and as you said in the chat, tell people to use the search for larger lists

PR #1645, opened 2024-09-04
fix: 🐛 move download to after checkout, fix conditionals

jozefKruszynski:
Add if always() to ensure further expression is checked, ensure that
build and push docker job runs after pypi publish is either successful
or skipped.  Move download artifacts job to after checkout so that it
isn't wiped out by the checkout

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,18 +98,14 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       packages: write
-    needs: build-artifact
+    needs: 
+      - build-artifact
+      - pypi-publish
+    if: |
+      always() &&
+      needs.build-artifact.result == 'success' &&
+      (needs.pypi-publish.result == 'success' || needs.pypi-publish.result == 'skipped')  
     steps:
-      - name: Retrieve release distributions
-        uses: actions/download-artifact@v4
-        with:
-          name: release-dists
-          path: dist/
-      - name: Copy Wheel to new location
-        shell: bash
-        run: |
-          mkdir -p pre_built_whl
-          cp dist/*.whl pre_built_whl/ 
       - uses: actions/checkout@v4.1.4
       - name: Download Widevine CDM client files from private repository
         shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -98,18 +98,14 @@ jobs:
...
     permissions:
       packages: write
-    needs: build-artifact
+    needs: 
+      - build-artifact
+      - pypi-publish
+    if: |
+      always() &&
+      needs.build-artifact.result == 'success' &&
+      (needs.pypi-publish.result == 'success' || needs.pypi-publish.result == 'skipped')
```
marcelveldt, line 107: why the check for pypi-publish here ?
dont we always want to grab the release from the prebuilt wheel?

jozefKruszynski: I assumed that we wanted to build from the pypi published artifact IF we actually publish it, if not, then I can just remove that bit.

MarvinSchenkel: I think it makes sense to also put the 'publish-to-pypi' job as a pre-requisite for the docker build job for stable releases. That way we can ensure that the release artifact is actually published to pypi before we build a stable Docker image.

MarvinSchenkel: Also think we should add an additional check to the jobs for bumping the add-on / creating the release notes. They should not run for dev builds.

jozefKruszynski: That would be a tricky condition to write though.

`if: |
      always() &&
      needs.build-artifact.result == 'success' &&
      ((needs.pypi-publish.result == 'success' && GitHub.events.release  == 'true') || (needs.pypi-publish.result ==     'skipped' && GitHub.events.release  == 'false'))`

Or something weird like this

MarvinSchenkel: There's no such thing as weird code if there's a proper comment that goes along with it ;-)

jozefKruszynski: I prefer to avoid code that needs to be commented to explain it, but you're right, waiting for PYPI makes sense for release builds, and a comment at least serves to explain in case of confusion.

marcelveldt: yeah - its okay if we dont install FROM pypi (as that adds an annoying delay) but we must ensure that for stable builds, the pypi task did complete

marcelveldt (approved):

PR #1665, opened 2024-09-15
Adding missing icon for the Soundcloud music provider

robsonke:
It seems like this icon was missing. Since I like to recognise the source of tracks, creating this PR. Let me know if this is correct.


marcelveldt (approved):
Thanks!

PR #1666, opened 2024-09-15
Fix loading state from cache when connecting to slimproto players

kepstin:
The cache keys being used for saving previous state and for loading state when connecting did not match.


marcelveldt (approved):
Thanks!

PR #1681, opened 2024-09-28
Add 'Don't stop the music' feature

MarvinSchenkel:
Automatically play similar music at the end of a queue


music_assistant/server/controllers/player_queues.py
```diff
@@ -345,6 +359,10 @@ async def play_media(
                 else:
                     media_item = item
 
+                # Save requested media item to play on the queue so we can use it as a source
+                # for Don't stop the music
+                queue.played_media_item = media_item
```
marcelveldt, line 364: So if I play a few tracks, it will be based on the last track of the list?

MarvinSchenkel: Yep, we could also change it to be a `list` so that it grabs all the tracks that were added.

marcelveldt: yeah maybe a length limited list of max 10 items or so

T3chArmy: Out of curiosity, would this theoretically resolve my issue with the current implementation where it pulls the same 20(or so) songs from Tidal into the playlist. So instead of pulling songs based on the song the radio station was started on, it would pull a list of songs based on say the most recently played 5-10 songs?

MarvinSchenkel:
> @MarvinSchenkel please mark this PR as ready for review once you are ready for a final review.

I will. Still need some final tweaks + the UI bit. Will try to finalize it this week.

music_assistant/server/controllers/player_queues.py
MarvinSchenkel: No, radio mode bases the radio based on the 'radio media item' (item used when the user clicked 'start radio'). However, I rewrote the logic a few beta releases ago, so that it uses individual songs to search for 'similar songs' instead of an album or an artist. Did you test if this behaviour still occurs with Tidal in the latest beta?

music_assistant/common/models/player_queue.py
```diff
@@ -36,6 +37,7 @@ class PlayerQueue(DataClassDictMixin):
     current_item: QueueItem | None = None
     next_item: QueueItem | None = None
     radio_source: list[MediaItemType] = field(default_factory=list)
+    played_media_item: list[MediaItemType] = field(default_factory=list)
```
marcelveldt, line 40: hmmm, not sure about this variable name as its a bit misleading Can you at least add some surrounding comments what its for.

music_assistant/server/controllers/player_queues.py
```diff
@@ -890,11 +915,17 @@ async def on_player_register(self, player: Player) -> None:
                     str(err),
                 )
         if queue is None:
+            dont_stop_the_music_enabled = self.mass.config.get_raw_core_config_value(
+                self.domain,
+                CONF_DEFAULT_DONT_STOP_THE_MUSIC,
+                True,
```
marcelveldt, line 921: Why not define the configentry as constant and refer to it here ?
Also.. not so sure about the default set to true.

music_assistant/server/controllers/player_queues.py
```diff
@@ -1050,6 +1081,15 @@ def on_player_update(
...
             ):
                 self.mass.create_task(self._fill_radio_tracks(queue_id))
+            elif (
+                # We have received the last item in the queue and Don't stop the music is enabled
+                queue.dont_stop_the_music_enabled
+                and queue.current_index
+                and (queue.items - queue.current_index) <= 1
+            ):
+                # Enable radio mode on the originally requested MediaItem in play_media
+                queue.radio_source = queue.played_media_item
```
marcelveldt, line 1091: Isn't this a super complicated way of doing this ?
We have all the already played tracks available in the queue_items

music_assistant/common/models/player_queue.py
MarvinSchenkel: Agree. How about `enqueued_media_items` ?

music_assistant/server/controllers/player_queues.py
MarvinSchenkel: Well we do not want the played tracks as a source for our radio. We only want to use the _originally enqueued media items_ as a source for the radio, because they represent the closest of what the user intended to play. 

The logic here checks if we are playing the last item in the queue and, if DSTM is enabled, uses those _originally enqueued media items_ as a radio source to dynamically fill the queue.

marcelveldt: Sure but radio mode is always based on the tracks right ?
So radio mode for playlist --> will be based on the tracks
So radio mode for album--> will be based on the tracks

MarvinSchenkel: > Sure but radio mode is always based on the tracks right ? So radio mode for playlist --> will be based on the tracks So radio mode for album--> will be based on the tracks

No, radio mode is based on `MediaItems`. So the `radio_source` can be a list of let's say `[Album, Track, Playlist]`. The radio _logic_ translates this into a list of base tracks (random 5 tracks per `MediaItem`), in this case 5 album tracks + the enqueued track + 5 random playlist tracks. Then it will create the dynamic playlist based on similar tracks for all of those base tracks.

This ensures that the originally enqueued items are remembered so we can always go back to them for generating a fresh list of dynamic tracks.

music_assistant/common/models/player_queue.py
marcelveldt: would it maybe be possible to limit this to only the uri's (or Itemmapping) of the items you want to remember ? 
Otherwise the player_queue object which is already very large, gets even larger. Note that this also has to be serialized over the api

MarvinSchenkel: I am currently limiting this to 10 MediaItems using a FIFO approach here: https://github.com/music-assistant/server/blob/10fb892d31948f1955881d52b985c26126a47a4d/music_assistant/server/controllers/player_queues.py#L369 . Is this what you meant?

marcelveldt: No, What I mean is instead of storing complete MediaItem objects in that list, will it suffice to only store uri's (so a list of string) ?

music_assistant/server/controllers/player_queues.py
MarvinSchenkel: Looked at the default value, but don't see the point in making it a constant. The config is only referred to once, at the time of when the player queue is created. This is the only time we actually need to read the default value from the settings. In all other cases we need the value that is stored on the player_queue. To me it seems similar as for example this setting: https://github.com/music-assistant/server/blob/10fb892d31948f1955881d52b985c26126a47a4d/music_assistant/server/controllers/player_queues.py#L1471

music_assistant/common/models/player_queue.py
MarvinSchenkel: Yea good one, I think we can store the uri's and translate the uri's into MediaItem object at the time when DSTM kicks in?

marcelveldt: yes, that way you can also store more items in the list and it will prevent the queueitem object blowing up too much

music_assistant/server/controllers/player_queues.py
marcelveldt: then at least make it a constant and set the default to false

marcelveldt: or let it say otherwise: the default can be true but then we need to ensure there are providers present that have dynamic tracks feature

music_assistant/common/models/player_queue.py
MarvinSchenkel: Fixed!

music_assistant/server/controllers/player_queues.py
MarvinSchenkel: Fixed, Default is now a constant `True`. I also check whether there is a provider available with the similar tracks feature when using the default value.

music_assistant/common/models/player_queue.py
```diff
@@ -36,6 +37,8 @@ class PlayerQueue(DataClassDictMixin):
     current_item: QueueItem | None = None
     next_item: QueueItem | None = None
     radio_source: list[MediaItemType] = field(default_factory=list)
+    # Use a list[str] here for the uri's to avoid having to store full MediaItem objects
```
marcelveldt, line 40:
```suggestion
    # Use a list of media item uri's here to avoid having to store full MediaItem objects
```

music_assistant/server/controllers/player_queues.py
```diff
@@ -1525,3 +1573,11 @@ def _parse_player_current_item_id(self, queue_id: str, player: Player) -> str |
...
                 return current_item_id
         return None
+
+    async def _schedule_dont_stop_the_music(self, queue: PlayerQueue):
+        """Auto turn on Radio Mode based on enqueued Media Items."""
+        queue.radio_source = [
+            await self.mass.music.get_item_by_uri(uri) for uri in queue.enqueued_media_items
+        ]
+        # schedule the next item in the queue
+        self.mass.create_task(self._fill_radio_tracks(queue.queue_id))
```
marcelveldt, line 1583: this is already executed as background task, why spawn another task ?

marcelveldt, line 1583:
```suggestion
        await self._fill_radio_tracks(queue.queue_id)
```

MarvinSchenkel: Fair point

marcelveldt (approved):
Nice!

PR #1682, opened 2024-09-29
Drop root during base image build and use venv

kbirger:
Addresses https://github.com/music-assistant/hass-music-assistant/issues/2956

To prevent having to run the container as root, this PR creates as an "app" user and group during the build of the base image (per alpine best practice) and additionally sets `/tmp` to be writable by that user. The PR also activates the virtual environment created.


Dockerfile.base
```diff
@@ -29,18 +29,24 @@ RUN set -x \
...
+    && chgrp -R app /tmp
+
+WORKDIR /home/app
+USER app
+
 
 # Configure runtime environmental variables
 ENV LD_PRELOAD="/usr/lib/libjemalloc.so.2"
-ENV VIRTUAL_ENV=/opt/venv
+ENV VIRTUAL_ENV=venv
```
marcelveldt, line 42:
```suggestion
ENV VIRTUAL_ENV=/home/app/venv
```

marcelveldt:
I tested this just now and got the same result as my own testing using a custom user within the container: CIFS mounting doesn't work. 

So my approach was just to keep the default behavior of running as the default root user (and have a bit of faith in the docker security), drop the privileged and fix the permissions on the venv and temp so that you can potentially run the docker container yourself as different user (but just without the samba provider)...

marcelveldt:
Thinking about this more I think this is fine. If a user wants to use the SMB provider to mount a samba share, they can implicitly define --user 0:0 to run the container as the root user again. Just tested that and that works just fine.

marcelveldt (approved):
Thanks!

marcelveldt:
See https://github.com/music-assistant/hass-music-assistant/issues/2956#issuecomment-2381637688

I had to partially revert because it completely broke the HA add-ons (and for now I really want to share the docker layer between regular docker and HA add-on).

I just tested and with the chmods of the workdir and venv, you can now run the container with a custom user.
2.3.0b28 has been released so you can test it

PR #1683, opened 2024-09-29
fix: album directory parsing with album versions

skrobul:
I have found this problem while investigating why some of the covers don't load for me in https://github.com/orgs/music-assistant/discussions/2976.

This fixes how the album directory is parsed when the directory name includes the version.

Prior to this patch:

```
>>> print(get_album_dir(track_dir='/media/download/others/Kiasmos - II - (WEB)', album_name="II"))
None
```

After the patch:

```
>>> print(get_album_dir(track_dir='/media/download/others/Kiasmos - II - (WEB)', album_name="II"))
/media/download/others/Kiasmos - II - (WEB)
```

### DIFF
--- a/music_assistant/server/providers/filesystem_local/helpers.py
+++ b/music_assistant/server/providers/filesystem_local/helpers.py
@@ -79,7 +79,7 @@ def get_album_dir(track_dir: str, album_name: str) -> str | None:
         if compare_strings(album_name, dirname.split(" - ")[-1], False):
             # account for ArtistName - AlbumName format in the directory name
             return parentdir
-        if compare_strings(album_name, dirname.split(" - ")[1].split("(")[0], False):
+        if compare_strings(album_name, dirname.split(" - ")[-1].split("(")[0], False):
             # account for ArtistName - AlbumName (Version) format in the directory name
             return parentdir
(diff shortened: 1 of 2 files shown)


music_assistant/server/providers/filesystem_local/helpers.py
```diff
@@ -79,7 +79,7 @@ def get_album_dir(track_dir: str, album_name: str) -> str | None:
         if compare_strings(album_name, dirname.split(" - ")[-1], False):
             # account for ArtistName - AlbumName format in the directory name
             return parentdir
-        if compare_strings(album_name, dirname.split("(")[0], False):
+        if compare_strings(album_name, dirname.split(" - ")[1].split("(")[0], False):
```
marcelveldt, line 82: this is dangerous - what if there is no "-" in the name ? crash
This needs either an additional guard or just use index -1 to grab the lst part

skrobul: thanks for catching that - it was meant to be `-1`. I have fixed it in the commit below and added some tests just to be sure.

marcelveldt (approved):
Nice, thanks!

PR #1691, opened 2024-10-02
Unfinished change for Snapcast provider

whc2001:
Details in [#2690](https://github.com/music-assistant/hass-music-assistant/issues/2690)


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -668,8 +684,9 @@ async def _builtin_server_runner(self) -> None:
             "--http.port=1780",
             f"--http.doc_root={SNAPWEB_DIR}",
             "--tcp.enabled=true",
-            "--tcp.port=1705",
-            f"--stream.buffer={self._snapcast_server_control_port}",
+            f"--tcp.port={self._snapcast_server_control_port}",
+            f"--stream.buffer={self._snapcast_server_buffer_size}",
```
SantiagoSotoC, line 688: The port control is only intended for an external server, does it change anything to allow the user to change the port of the embedded server? Doesn't it just add complexity?

whc2001: I think when the external server is disabled this value is just set to `DEFAULT_SNAPSERVER_PORT` as in L291-293?

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -133,6 +134,20 @@ async def get_config_entries(
...
+            default_value=20,
+            label="Snapserver chunk size",
+            description="Default source stream read chunk size [ms]. "
+            "The server will continously read this number of milliseconds from the source into buffer and pass this buffer to the encoder."
+            "The encoded buffer is sent to the clients. Some codecs have a higher latency and will need more data, e.g. Flac will need ~26ms chunks",
+            required=False,
+            category="Built-in Snapserver Settings",
+            hidden=not local_snapserver_present,
+            help_link="https://raw.githubusercontent.com/badaix/snapcast/86cd4b2b63e750a72e0dfe6a46d47caf01426c8d/server/etc/snapserver.conf",
+        ),
```
SantiagoSotoC, line 150: How does the chunk_size work in the snapserver? isn't it defined by default different for each codec? i don't know why i thought it was that way but now i can't find anything in the documentation

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: You are right, this is ok

SantiagoSotoC: I don't know how to do it, I do it from vscode with the extension for github

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -133,6 +134,20 @@ async def get_config_entries(
             hidden=not local_snapserver_present,
             help_link="https://raw.githubusercontent.com/badaix/snapcast/86cd4b2b63e750a72e0dfe6a46d47caf01426c8d/server/etc/snapserver.conf",
         ),
+        ConfigEntry(
+            key=CONF_SERVER_CHUNK_MS,
+            type=ConfigEntryType.INTEGER,
+            range=(10, 100),
+            default_value=20,
```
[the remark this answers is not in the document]
SantiagoSotoC: Our default codec is flac, shouldn't we set the chunk size to 26 as described in the description?

music_assistant/server/providers/snapcast/__init__.py
whc2001: To be honest, I am not sure how exactly this parameter is working either. I just know from my experience that if this value is too small the sound will sometimes become choppy. Due to this I think it would be handy to leave out a configurable parameter in case of problems.

music_assistant/server/providers/snapcast/__init__.py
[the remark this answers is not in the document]
whc2001: Makes sense, although I am not sure why the default value in the example config is 20 while the default codec is FLAC. Bumping it up to 26 should not hurt anything.

PR #1692, opened 2024-10-02
Snapcast: Add configuration options 

SantiagoSotoC:
We have a new contributor, @whc2001 thank you for your help 🎉🎉

Some configuration options and some related details are fixed.

Fixes the problem with tts cutting off in the middle of the time, but there is still a problem with very short tts, like “Hi” where it just stops the music that is playing and comes back with the music after a few seconds and you never hear the chime or the “Hi”.

### DIFF
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -141,8 +141,10 @@ async def get_config_entries(
             default_value=26,
             label="Snapserver chunk size",
             description="Default source stream read chunk size [ms]. "
-            "The server will continously read this number of milliseconds from the source into buffer and pass this buffer to the encoder."
-            "The encoded buffer is sent to the clients. Some codecs have a higher latency and will need more data, e.g. Flac will need ~26ms chunks",
+            "The server will continuously read this number of milliseconds from the source into "
+            "buffer and pass this buffer to the encoder."
+            "The encoded buffer is sent to the clients. Some codecs have a higher"
+            "latency and will need more data, e.g. Flac will need ~26ms chunks",
             required=False,
             category="Built-in Snapserver Settings",
             hidden=not local_snapserver_present,
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -588,7 +588,7 @@ class SnapCastProvider(PlayerProvider):
             result = await self._snapserver.stream_add_stream(
                 # NOTE: setting the sampleformat to something else
                 # (like 24 bits bit depth) does not seem to work at all!
-                f"tcp://0.0.0.0:{port}?name={name}&sampleformat=48000:16:2",
+                f"tcp://0.0.0.0:{port}?name={name}&sampleformat=48000:16:2&idle_threshold=60000",
             )
             if "id" not in result:
                 # if the port is already taken, the result will be an error
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -62,11 +62,14 @@ CONF_SERVER_BUFFER_SIZE = "snapcast_server_built_in_buffer_size"
 CONF_SERVER_INITIAL_VOLUME = "snapcast_server_built_in_initial_volume"
 CONF_SERVER_TRANSPORT_CODEC = "snapcast_server_built_in_codec"
 CONF_SERVER_SEND_AUDIO_TO_MUTED = "snapcast_server_built_in_send_muted"
+CONF_STREAM_IDLE_THRESHOLD = "snapcast_stream_idle_threshold"
+
 
 # airplay has fixed sample rate/bit depth so make this config entry static and hidden
 CONF_ENTRY_SAMPLE_RATES_SNAPCAST = create_sample_rates_config_entry(48000, 16, 48000, 16, True)
 
 DEFAULT_SNAPSERVER_PORT = 1705
+DEFAULT_SNAPSTREAM_IDLE_THRESHOLD = 60000
 
 SNAPWEB_DIR: Final[pathlib.Path] = pathlib.Path(__file__).parent.resolve().joinpath("snapweb")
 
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -130,6 +134,22 @@ async def get_config_entries(
...
+            range=(10, 100),
+            default_value=26,
+            label="Snapserver chunk size",
+            description="Default source stream read chunk size [ms]. "
+            "The server will continuously read this number of milliseconds from the source into "
+            "buffer and pass this buffer to the encoder."
+            "The encoded buffer is sent to the clients. Some codecs have a higher"
+            "latency and will need more data, e.g. Flac will need ~26ms chunks",
+            required=False,
+            category="Built-in Snapserver Settings",
```
marcelveldt, line 149: let's not use a full text here. Use an identifier and add it to the translations on the frontend.

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -210,6 +230,14 @@ async def get_config_entries(
             depends_on=CONF_USE_EXTERNAL_SERVER,
             category="advanced" if local_snapserver_present else "generic",
         ),
+        ConfigEntry(
+            key=CONF_STREAM_IDLE_THRESHOLD,
+            type=ConfigEntryType.INTEGER,
+            default_value=DEFAULT_SNAPSTREAM_IDLE_THRESHOLD,
+            label="Snapcast idle threshold stream parameter",
+            required=True,
+            category="Stream Settings",
```
marcelveldt, line 239: same, category is not meant to be a freetext label

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: Where can I find an example of this, I searched in other suppliers and for now this is how it was done before

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: this seems to be some super advanced setting, why not mark the category as "advanced" ?

music_assistant/server/providers/snapcast/__init__.py
whc2001: I also checked the server and frontend repo, not sure how do I make it an identifier...

music_assistant/server/providers/snapcast/__init__.py
whc2001: > this seems to be some super advanced setting, why not mark the category as "advanced" ?

This value is the key to solve the audio cut-out problem, so it might still need some tweaking. But putting it in Advanced makes sense as it's not something you want to adjust every now and then.

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: In the frontend is to add a line to en.js, isn't it ?

whc2001: > In the frontend is to add a line to en.js, isn't it ?

I think that JSON file is autogenerated from Lokalise i18n platform and should not be modified by hand. We need to add a "key" instead of a naive string in the server code somewhere, then put that key into Lokalise and provide translations to it. Still figuring out how to do that

music_assistant/server/providers/snapcast/__init__.py
SantiagoSotoC: I move it to advance

music_assistant/server/providers/snapcast/__init__.py
whc2001: Okay I think I figured out what to do...
image

SantiagoSotoC: I think that permissions are required, I don't have them in lokalise, the option that is in the docs doesn't appear to me

whc2001: #1693 I have removed the description and separated some of the hard coded values

SantiagoSotoC:
@whc2001 are you using pre commit ?

whc2001:
> @whc2001 are you using pre commit ?

Ah sorry I think it's not properly configured on the Windows machine I used to push

music_assistant/server/providers/snapcast/__init__.py
marcelveldt: category should be a (string) key only. All lowercase, no spaces.
The default keys are generic and advanced. If you want to add more, then the key needs to be in the translations/en.json file from frontend:

https://github.com/music-assistant/frontend/blob/main/src/translations/en.json#L291

Only edit the en.json file, the others are maintained by lokalize.

marcelveldt:
> Ah sorry I think it's not properly configured on the Windows machine I used to push

Windows is not recommended for python development, especially not for MA development.
Best is to use MacOS or Linux. Dev container could potentially work but has caveats.

marcelveldt (approved):
Looks good, nice work!

music_assistant/server/providers/snapcast/__init__.py
whc2001: > category should be a (string) key only. All lowercase, no spaces. The default keys are generic and advanced. If you want to add more, then the key needs to be in the translations/en.json file from frontend:
> 
> https://github.com/music-assistant/frontend/blob/main/src/translations/en.json#L291
> 
> Only edit the en.json file, the others are maintained by lokalize.

Got it, I'll change it later as I am a little busy today.

marcelveldt:
Want to merge this now or do you want to adjust those category labels first ?

SantiagoSotoC:
> Want to merge this now or do you want to adjust those category labels first ?

The categories are already set but we are going to do the localization here, otherwise everything will remain without description

SantiagoSotoC:
@whc2001 I am already editing the frontend en.json

SantiagoSotoC:
> @whc2001 I am already editing the frontend en.json

https://github.com/music-assistant/frontend/pull/673

whc2001:
> > @whc2001 I am already editing the frontend en.json
> 
> [music-assistant/frontend#673](https://github.com/music-assistant/frontend/pull/673)

Just got there, I'll check

whc2001:
Already got my latest change in https://github.com/music-assistant/frontend/pull/674 and https://github.com/music-assistant/server/pull/1695

PR #1695, opened 2024-10-03
fix: change category key format

whc2001:


whc2001:
I think this one is still needed, otherwise the `Built-in Snapserver Settings` string will not get a translation?

PR #1705, opened 2024-10-11
Snapcast update 0.29

SantiagoSotoC:
Should we use snapcast as a package or snapcast-server ? 
snapcast includes snapclient as well.


marcelveldt:
snapcast package is fine as we need snapclient as well soon for the local player I'm working

marcelveldt (approved):
how stable is the 0.29 version and what does it (potentially) fix that we want it already ?

SantiagoSotoC:
it is stable, it has 2 months without complaints, version 8 snapweb corresponds to 0.29

SantiagoSotoC:
But with snapcast it is not clear how to test it, the documentation is not very clear.

PR #1718, opened 2024-10-16
Refactor SnapCastProvider cmd_volume_mute method

SantiagoSotoC:
This pull request includes a small but important change to the `cmd_volume_mute` method in the `music_assistant/server/providers/snapcast/__init__.py` file. The change ensures that the player's mute status is correctly updated in the system.

### DIFF
--- a/music_assistant/server/providers/snapcast/__init__.py
+++ b/music_assistant/server/providers/snapcast/__init__.py
@@ -447,8 +447,12 @@ class SnapCastProvider(PlayerProvider):
 
     async def cmd_volume_mute(self, player_id: str, muted: bool) -> None:
         """Send MUTE command to given player."""
+        ma_player = self.mass.players.get(player_id, raise_unavailable=False)
         snap_client_id = self._get_snapclient_id(player_id)
-        await self._snapserver.client(snap_client_id).set_muted(muted)
+        snapclient = self._snapserver.client(snap_client_id)
+        await snapclient.set_muted(muted)
+        ma_player.volume_muted = snapclient.muted
+        self.mass.players.update(player_id)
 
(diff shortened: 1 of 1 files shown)


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -447,8 +447,12 @@ async def cmd_stop(self, player_id: str) -> None:
 
     async def cmd_volume_mute(self, player_id: str, muted: bool) -> None:
         """Send MUTE command to given player."""
+        ma_player = self.mass.players.get(player_id, raise_unavailable=False)
         snap_client_id = self._get_snapclient_id(player_id)
-        await self._snapserver.client(snap_client_id).set_muted(muted)
+        snapclient = self._snapserver.client(snap_client_id)
+        await snapclient.set_muted(muted)
+        ma_player.volume_muted = snapclient.muted
```
marcelveldt, line 454: So, you're setting an optimistic value here. Does that mean snapcast does not do reporting of the mute state ?
Is that a bug perhaps in the api ?

In any case, please add a comment here that you are setting the state optimistically and why that is.

SantiagoSotoC: snapcast does report it, but the library does not.

`async def set_muted(self, status):

        """Set client mute status."""
        new_volume = self._client['config']['volume']
        new_volume['muted'] = status
        self._client['config']['volume']['muted'] = status
        await self._server.client_volume(self.identifier, new_volume)
        _LOGGER.debug('set muted to %s on %s', status, self.friendly_name)`

That is the code in the library, there would be two options either to call the callback or return the response from the server.

marcelveldt (approved):

PR #1721, opened 2024-10-17
YouTube Music: Filter videos from search results and radio

ov1d1u:
This small PR removes videos from search results and radio playlists. I made this change mostly for myself, as I'm annoyed to hear all non-music related noises when I'm listening to music from YouTube Music.

Because the results are being filtered after YouTube returned them, I think this may lead to situations when you start a radio and you get nothing (because everything was filtered out), but I never ran in such an issue.

Ideally this would be an option in the YouTube Music provider, but I don't know Music Assistant's codebase so well yet to implement this.

### DIFF
--- a/music_assistant/server/providers/ytmusic/helpers.py
+++ b/music_assistant/server/providers/ytmusic/helpers.py
@@ -232,7 +232,9 @@ async def get_song_radio_tracks(
             videoId=prov_item_id, playlistId=playlist_id, limit=limit, radio=True
         )
         # Replace inconsistensies for easier parsing
-        for track in result["tracks"]:
+        music_video_types = ["MUSIC_VIDEO_TYPE_ATV", "OFFICIAL_SOURCE_MUSIC"]
+        tracks = [ t for t in result["tracks"] if t["videoType"] in music_video_types ]
+        for track in tracks:
             if track.get("thumbnail"):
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Just out of curiosity, can you link a few videos that would be excluded with this extra filter? I once had a '12 hours calm waves' video included in the radio, but we are currently already restricting radio items to have a max length of 20mins.

Preferably this would be a boolean configuration item 'Exclude videos from search & radio' (default `False`) in the provider. You can have a look [here](https://github.com/music-assistant/server/blob/ada4dceef1f1279dc0d0eb9d9824d7a805497022/music_assistant/server/controllers/player_queues.py#L210) at an example. If you need some more pointers or help, let me know.

ov1d1u:
An example would be https://www.youtube.com/watch?v=8WYHDfJDPDc, which is a music video with some non-music segment at the end. These are the results without this change, with the first result being this video:
Screenshot 2024-10-18 at 09 42 28

And these are the results after this change, the video filtered out:
Screenshot 2024-10-18 at 09 42 33

Thanks for pointing out that code, I will look into it and try to make this a configuration option.

marcelveldt:
Maybe better to point at a provider-level ConfigEntry, for examples see any of the existing providers

PR #1732, opened 2024-10-20
RSS Podcast Feed Provider

saeugetier:
This PR provides a work progress implementation for a RSS Podcast Feed Provider. Podcasts are not officially supported for now. The provider implements the Podcast feed as "Artist" and episodes as "Tracks". 

Playback of online podcast media works. But currently no sophisticated media type detection or error handling mechanism are implemented. It is just enough for playing single episodes.

This work is unfinished, but feedback would be appreciated. "Tracks" are not shown when the "Artist" is selected. I currently don't know, what is going wrong.


marcelveldt:
Nice idea!
Lets wait with this PR until after 2.3 has been released and we start the beta period for 2.4, in which we have planned to add media type support for audiobooks and podcasts. Then this would fit in much better.

So I'll get back to you on this one within 2 weeks.

marcelveldt:
Sorry it got accidentally closed

saeugetier:
Closed because fork will be recreated. Branch names has been changed and it was not easy to fix with my knowledge.

PR #1743, opened 2024-10-23
Snapcast fix short announcements

SantiagoSotoC:


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -559,12 +563,17 @@ async def _streamer() -> None:
...
                 self._set_childs_state(player_id)
             finally:
-                with suppress(TypeError, KeyError, AttributeError):
-                    await self._snapserver.stream_remove_stream(stream.identifier)
+                await self._delete_current_snapstream(stream, media)
 
         # start streaming the queue (pcm) audio in a background task
         self._stream_tasks[player_id] = asyncio.create_task(_streamer())
 
+    async def _delete_current_snapstream(self, stream, media):
```
marcelveldt, line 571: annotations missing here

music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -559,12 +563,17 @@ async def _streamer() -> None:
...
+        with suppress(TypeError, KeyError, AttributeError):
+            if media.duration < 5:
+                await asyncio.sleep(5)
```
marcelveldt, line 574: what's this for ?

SantiagoSotoC: It adds a delay before eliminating the streams when the tracks are short, when they are very short it doesn't have time to play, that's why the short ttls don't work, try several values, it's the minimum that works with only one letter.

marcelveldt (approved):

PR #1764, opened 2024-10-28
Snapcast fix lost state when stop 

SantiagoSotoC:


music_assistant/server/providers/snapcast/__init__.py
```diff
@@ -432,6 +432,7 @@ async def cmd_volume_set(self, player_id: str, volume_level: int) -> None:
     async def cmd_stop(self, player_id: str) -> None:
         """Send STOP command to given player."""
         player = self.mass.players.get(player_id, raise_unavailable=False)
+        player.elapsed_time = time.time()
```
marcelveldt: this doesnt make any sense

SantiagoSotoC: Well then there is a bug somewhere else.
With this it works as it should.
The problem is that when pausing, it goes back to the first song in the queue.
It also clears the list of songs already played.
Do you have any idea what it could be ?

SantiagoSotoC:
I understand that the code does not make sense, but I leave a video showing the error and the “solution”.

Current status in branch dev: 

https://github.com/user-attachments/assets/63962d42-14ac-4858-9428-9702066af5e0

With the change in the pr:


https://github.com/user-attachments/assets/f23aef77-83a9-4ba3-837f-494224a36fb7

SantiagoSotoC:
Does this make more sense?

marcelveldt (approved):
Yes, this fix makes more sense

SantiagoSotoC:
I have been trying this a little more and it does not solve the error, the first solution does, the second does not, it is strange because it certainly does not make sense.

SantiagoSotoC:
Move the change to cmd_stop if it seems to fix it. It shouldn't make a difference though.

PR #1774, opened 2024-11-08
Fix Radio mode for Subsonic and remove Podcast hacks

khers:
Don't stop the music was failing due to a typo when calling get_similar_songs(), fix the typo to enable this mode.

There is a model refactor coming that will add Podcasts as first class objects to MA so none of this code that tries to wedge them into the Artist/Album/Track model is necessary. Tear out the dead code.


marcelveldt (approved):
lgtm!

PR #1777, opened 2024-11-16
Fix: Implement cookie authentication for YTM

MarvinSchenkel:
This PR adds cookie authentication to the YT Music provider. Google unfortunately disabled OAuth, so we have to rely on cookies now to get it working


marcelveldt (approved):
Its unfortunate that we're back to the cookies hack but at least it works again.

marcelveldt:
Ignoring the lint issue as the requirements gen script is a bit confused.

PR #1779, opened 2024-11-17
Fix: Improve accuracy of matching album in directory structure for local filesystem

skrobul:
Hi, this is another PR in my quest to fix the album art not loading correctly on my library. Here's a description of the changes from the first commit:

This code resolves a problem with some of the albums not being matched by other heuristics in the `get_album_dir()` method by taking slightly different approach. It emulates what humans do - mentally break down the name into logical chunks that represent an artist name, album name and
ignoring everything else.

The testing suite has been extended with a few examples of albums exhibiting problems with the way matching was done prior to this commit.

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -1028,7 +1028,7 @@ class LocalFileSystemProvider(MusicProvider):
                 if item.ext != ext:
                     continue
                 # try match on filename = one of our imagetypes
-                if item.name in ImageType:
+                if item.name in ImageType.__members__:
                     images.append(
                         MediaItemImage(
                             type=ImageType(item.name),
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -1028,7 +1028,7 @@ class LocalFileSystemProvider(MusicProvider):
                 if item.ext != ext:
                     continue
                 # try match on filename = one of our imagetypes
-                if item.name in ImageType.__members__:
+                if item.name in ImageType:
                     images.append(
                         MediaItemImage(
                             type=ImageType(item.name),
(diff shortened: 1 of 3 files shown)


music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1028,7 +1028,7 @@ async def _get_local_images(self, folder: str) -> UniqueList[MediaItemImage]:
                 if item.ext != ext:
                     continue
                 # try match on filename = one of our imagetypes
-                if item.name in ImageType:
+                if item.name in ImageType.__members__:
```
```suggestion
                if item.name in ImageType:
```
any reason for doing this as this should not be needed. ImageType is a StrEnum

marcelveldt: Scherm­afbeelding 2024-11-20 om 18 47 28

marcelveldt (approved):
Nice work!

music_assistant/providers/filesystem_local/__init__.py
skrobul: The pyproject.toml suggests that Python 3.11 is still supported, but that does not work in 3.11:


```python
❯ python
Python 3.11.9 (main, Apr  2 2024, 08:25:04) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from music_assistant_models import enums
>>> enums.ImageType
<enum 'ImageType'>
>>> "abc" in enums.ImageType
<stdin>:1: DeprecationWarning: in 3.12 __contains__ will no longer raise TypeError, but will return True or
False depending on whether the value is a member or the value of a member
  File "<stdin>", line 1, in <module>
  File "/run/current-system/sw/lib/python3.11/enum.py", line 742, in __contains__
    raise TypeError(
TypeError: unsupported operand type(s) for 'in': 'str' and 'EnumType'
>>>
```

marcelveldt: ah.. we're using 3.12 in our base images so lets bump that

PR #1797, opened 2024-11-29
Fix: Add logging around Subsonic scrobble calls

khers:
There is some disagreement about if the scrobble calls are actually occuring for some providers. Add logging around these calls to make sure the user knows they happened.

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -711,11 +711,14 @@ class OpenSonicProvider(MusicProvider):
         )
 
     async def _report_playback_started(self, item_id: str) -> None:
+        self.logger.debug("scrobble for now playing called for %s", item_id)
         await self._run_async(self._conn.scrobble, sid=item_id, submission=False)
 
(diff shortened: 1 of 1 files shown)


music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -711,11 +711,14 @@ async def get_stream_details(
         )
 
     async def _report_playback_started(self, item_id: str) -> None:
+        self.logger.debug("scrobble for now playing called for %s", item_id)
         await self._run_async(self._conn.scrobble, sid=item_id, submission=False)
 
     async def on_streamed(self, streamdetails: StreamDetails, seconds_streamed: int) -> None:
         """Handle callback when an item completed streaming."""
+        self.logger.debug("on_streamed called for %s", streamdetails.item_id)
```
marcelveldt, line 719: note that this will be called when the buffer is filled - its not perse the actual playback time, especially not for players that buffer a lot of data ahead such as sonos.

If you want actual played seconds we also have an event called by the queue manager.

marcelveldt (approved):

PR #1800, opened 2024-12-02
Add new file extension

OzGav:
Will this fix this? https://github.com/orgs/music-assistant/discussions/3216

TL;DR user has files with capitalised extension not recognised by MA


marcelveldt:
This is not the way to fix this. The proper fix would be to make the code case insensitive.
First question I have is why you would have upper case file extensions but that aside.

OzGav:
Is it then a change to this line https://github.com/music-assistant/server/blame/c33e766ec4062ac9b812d31968fe3393bada4df6/music_assistant/providers/filesystem_local/__init__.py#L252

to 

            elif item.ext.lower() in TRACK_EXTENSIONS:

PR #1802, opened 2024-12-03
Subsonic: Add Podcast Support

khers:
This is a draft of the PR to add podcast support to the provider. I cannot test this yet as it seems the db tables are not ready.

I do want to call out the changes I had to make to the MusicProvider.


khers:
With these patches, I can sync the podcasts on my gonic server (I see them in the logs) however I don't see anywhere in the UI for them, not sure that has been implemented yet.

marcelveldt (approved):
Nice work!
FYI: I'm adding the UI support now so this should be in the next beta

PR #1803, opened 2024-12-08
fix: Solved a bug for applying genres in Soundcloud tracks

robsonke:
Simple bugfix to treat the genre metadata as an array instead of a string.

Original result after sync:
bug-genre

With fix applied:
bug-genre-solved


OzGav:
Fixes https://github.com/music-assistant/hass-music-assistant/issues/3107

robsonke:
Ah nice, didn't see that issue, but it's indeed the same problem.

marcelveldt (approved):

PR #1806, opened 2024-12-16
Allow Chromecast provider to handle/avoid images which are "too big"

### LINKED ISSUE #3285: Chromecast Player Crashes / Chomecast Reboots

2.3.3

Builtin to HA core 2024.12.1

- [X] Yes

### The problem

When playing back a subset of my albums on a Chromecast Ultra I see a spinning wheel where the artwork would appear and then either the app exits or the chromecast reboots -- I get the "Something went wrong" screen from the Chromecast OS and then it restarts

While the spinning is ongoing music plays ok.

The crashes appear to correlate with higher DPI/larger cover art files. All artwork was scanned locally by me from the physical CD.

Causes the problem (files are 9.6M and 14M respectively):
```
Holy Moses - World Chaos/cover.jpeg:                JPEG image data, JFIF standard 1.01, resolution (DPI), density 1200x1200, segment length 16, baseline, precision 8, 5664x5669, components 3
Winterfylleth - The Dark Hereafter/cover.jpeg:      JPEG image data, JFIF standard 1.01, resolution (DPI), density 1200x1200, segment length 16, baseline, precision 8, 5668x5668, components 3
```

Do not cause the problem (files are 381K, 156K and 475K respectively):
```
The Haunted - The Haunted Made Me Do It/cover.jpeg: JPEG image data, JFIF standard 1.01, resolution (DPI), density 300x300, segment length 16, baseline, precision 8, 1405x1405, components 3
Anathema - Eternity/cover.jpeg:                     JPEG image data, JFIF standard 1.01, resolution (DPI), density 150x150, segment length 16, baseline, precision 8, 826x738, components 3
Winterfylleth - The Threnody of Triumph/cover.jpeg: JPEG image data, JFIF standard 1.01, resolution (DPI), density 300x300, segment length 16, baseline, precision 8, 1417x1417, components 3
```

The music files are all flac files with the same sample rate:
```
Holy Moses - World Chaos/01-World Chaos.flac:                       FLAC audio bitstream data, 16 bit, stereo, 44.1 kHz, 8008560 samples
Winterfylleth - The Dark Hereafter/01-The Dark Hereafter.flac:      FLAC audio bitstream data, 16 bit, stereo, 44.1 kHz, 10829784 samples
The Haunted - The Haunted Made Me Do It/01-Dark Intentions.flac:    FLAC audio bitstream data, 16 bit, stereo, 44.1 kHz, 3993696 samples
Anathema - Eternity/01-Sentient.flac:                               FLAC audio bitstream data, 16 bit, stereo, 44.1 kHz, 7969164 samples
Winterfylleth - The Threnody of Triumph/01-A Thousand Winters.flac: FLAC audio bitstream data, 16 bit, stereo, 44.1 kHz, 20097252 samples
```

### How to reproduce

It appears to deterministically occur with specific artwork for specific albums.

Here are the various good/bad files discussed above: [ma.zip](https://github.com/user-attachments/files/18059974/ma.zip)

### Music Providers

All music from the local provider. Album artwork is a `cover.jpeg` in each folder. Artwork scanned locally by me from CD cover art.

### Player Providers

All my providers are chromecast.

On a Chromecast Ultra I see the issue as described above

On a JBL speaker (no screen => no artwork) I do not see the issue.

Chromecast to an nVidia shield I do not see the issue, the artwork appears ~immediately.

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/18059986/music-assistant.log)

### Additional information

I'm queuing albums via the web UI (from the "Random album" playlist).

I'm running HA and MA in Docker compose stacks on the same machine. Network mode is "host" for both containers. Music content is bind mounted ro into the MA container at `/media`.

Music is all musicbrainz tagged using picard.

Nothing in the troubleshooting guide seems relevant to a chromecast device crash.

2024.12.1

Home Assistant Container

ijc:
My Chromecast Ultra crashes when presented with cover images on the order of 5000x5000 pixels (I've not determined the actual limit, 2048x2048 is ok though).

Allow users to avoid this problem by providing an advanced option per player which downscales the image to a specified size.

Fixes: https://github.com/music-assistant/hass-music-assistant/issues/3285

### DIFF
--- a/music_assistant/providers/chromecast/__init__.py
+++ b/music_assistant/providers/chromecast/__init__.py
@@ -675,7 +675,7 @@ class ChromecastProvider(PlayerProvider):
         # update metadata of current item chromecast
         if media_controller.status.media_custom_data["queue_item_id"] != current_item.queue_item_id:
             image_url = (
-                self.mass.metadata.get_image_url(current_item.image)
+                self.mass.metadata.get_image_url(current_item.image, size=512)
                 if current_item.image
                 else MASS_LOGO_ONLINE
             )
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1261,7 +1261,7 @@ class PlayerQueuesController(CoreController):
                 album.name if (album := getattr(queue_item.media_item, "album", None)) else ""
             )
             if queue_item.image:
-                media.image_url = self.mass.metadata.get_image_url(queue_item.image)
+                media.image_url = self.mass.metadata.get_image_url(queue_item.image, size=512)
         return media
 
     async def get_artist_tracks(self, artist: Artist) -> list[Track]:
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -13,7 +13,7 @@ if [ -d "$env_name" ]; then
   echo "Virtual environment '$env_name' already exists."
 else
   echo "Creating Virtual environment..."
-  python -m venv .venv
+  ${PYTHON:-python} -m venv .venv
 fi
 echo "Activating virtual environment..."
 source .venv/bin/activate


khers (approved):

marcelveldt:
I don't think this should even be an option. Let's just limit all images that are sent to a player directly to 512x512, which is also the max for some other players like airplay

ijc:
My logic was that this was in the same set of things as the "force mp3" behaviour, but that's fine, I can make the change. Do you want this to happen in the core `PlayerQueuesController` level or per-player provider?

marcelveldt:
> My logic was that this was in the same set of things as the "force mp3" behaviour, but that's fine, I can make the change. Do you want this to happen in the core `PlayerQueuesController` level or per-player provider?

Yeah, but enabling the mp3 mode is more a choice where this is more a bug; sending a large image will simply crash the device. 

Let's just adjust the "get_image_url" with a "size=512" parameter in the individual player providers, starting with cast.


Scherm­afbeelding 2024-12-19 om 17 29 34

ijc:
Ah, I misunderstood the mp3 option as being a workaround for buggy players too.

> ... adjust the "get_image_url" with a "size=512" parameter in the individual player providers,...

Changing just that one in `update_flow_metadata` is not sufficient IME.

The one in `PlayerQueuesController.player_media_from_queue_item` ([here](https://github.com/music-assistant/server/blob/73545eef0d98383d79c89ee9a09bda586eeeba78/music_assistant/controllers/player_queues.py#L1264)) also needs to change, the URL from that one ends up in the per player `play_media` and `enqueue_next_media` hooks (as part of the `PlayerMedia` parameter) which is why I was patching in `_handle_limit_image_size` into those locations.

In this PR I didn't touch the one in `update_flow_metadata` at all and things work for me when playing back from the local provider. Looks like `update_flow_metadata` is only used for `queue.flow_mode` (which isn't used in chromecast player?) or for radio streams (which I've never tried).

ijc:
I added some radio stations and those do go via `update_flow_metadata` as expected -- none of the ones I tried actually had an image url, but if they did I'd expect them to be on the small side and not trigger the issue anyway.

music_assistant/providers/chromecast/__init__.py
```diff
@@ -624,19 +624,47 @@ async def _disconnect_chromecast(self, castplayer: CastPlayer) -> None:
...
+            url = url._replace(query=urlencode(qsl))
+            return url.geturl()
+        except Exception as err:
+            self.logger.warning(
+                "Error while clamping image url size: %s",
+                str(err),
+                exc_info=err if self.logger.isEnabledFor(10) else None,
+            )
+
+        return orig_url
```
marcelveldt: This assumes that every image has a query string with a size parameter while that is strictly speaking not true. If you want to absolutely guarantee the image size, you will have to request the image through the image proxy, which can also do resizing. 

I'm ok with this change as a start but you will still run into the issue if you receive an URL to an image that has no size in a query string but is just a super large image

marcelveldt:
> Changing just that one in update_flow_metadata is not sufficient IME.

That was just a code example. It would need to be adjusted in all parts that handle the image from the queue item.
If you want to be 100% certain that a resized image is sent, we route everything through the image proxy.

Idea can be that we go with your patch now and if this pops up again, we know we have to adjust it to the proxy everywhere ?

music_assistant/providers/chromecast/__init__.py
ijc: If you would prefer I could make the change to the call to `get_image_url` in the `PlayerQueuesController` instead, that would catch all (?) URL generation but would be affecting all players not just Chromecast.

marcelveldt: ha, yeah that makes a lot more sense - that is a single line change and fixes it everywhere. Good call

ijc: Done!

I changed the calls to `get_image_url` in `PlayerQueuesController` and the Chromecast player.

I didn't touch the calls in the airplay player (already sets size=500) or slimproto + sonos players (no known issues there). They will get the 512 limit from `PlayerQueuesController` but not in their own calls, that seems a bit inconsistent -- would it be better to add the 512 to those two as well while I'm here?

marcelveldt: yeah, and always feel free to submit/propose optimizations when you see fit.
Thanks!

marcelveldt (approved):
Thanks!

PR #1809, opened 2024-12-18
Make extension check case insensitive

OzGav:
Hopefully this will fix the MP3 extension issue?


OzGav:
Lint suggests this isn’t a solution…

PR #1823, opened 2025-01-02
Fix support links

chatziko:
For some reason I'm used to access the issue page via the README links, and I just realized they stopped working.


marcelveldt (approved):
Thanks!

PR #1839, opened 2025-01-08
Feat: Add multi device DSP support

maximmaxim345:
Requires https://github.com/music-assistant/models/pull/29 to be merged first.

# Overview

This PR primarily adds grouped playback support to the DSP of Music Assistant.
DSP is automatically disabled when grouped playback is enabled on player providers that do not support the new `MULTI_DEVICE_DSP` `PlayerFeature` flag.
If a player provider has this flag, it is responsible for retrieving `get_player_filter_params` for each sub-device separately.
Universal Groups allow all contained devices to apply their own DSP.
So if a user wants to use DSP on two players that do not support MULTI_DEVICE_DSP (like Snapcast), they can still create a Universal Group with them that will apply the DSP separately (at the cost of synchronization).

# Universal Group Adjustments

Universal Groups previously were forced to use MP3 for all devices. This PR decouples the audio processing into two stages:
- The previous speed limiter
- An optional per-client stream with applied DSP and correct output format

This has two main advantages:
- A filter can be applied when necessary, allowing us to use individual DSP on every device in the group
- We are no longer forced to use MP3 for every device, since a device not supporting lossless playback can request its own MP3 stream

Currently, all players except Slimproto, Snapcast, and AirPlay still use MP3, but this can be changed in the future.

# Supported Providers

With this PR, DSP is supported on:
- All player providers with single device playback active
- Grouped AirPlay players
- Grouped Slimproto players
- All players inside a Universal Group

With automatic disablement in unsupported cases.

# Testing

I have verified this PR as follows:
- Snapcast with 2 Snapweb browsers: Since Snapcast does not support multi-device DSP, this PR properly disables DSP when grouped
- Permanent Snapcast group: Works as above, disables the DSP
- Edge cases on startup: Works since reload detection is handled in `update()`
- Snapcast with manually added MULTI_DEVICE_DSP FeatureFlag: Keeps DSP enabled
- AirPlay: Each device gets its own DSP during synced playback
- Universal group with 2 Snapcast players: Individual players' DSPs are applied
- Universal group with 2 DLNA devices: Individual players' DSPs are applied
- Universal group with 2 DLNA devices, 1 Snapcast player, and 1 Chromecast: All 4 players receive their own DSP

# Future Work

- Add UI for multi-device DSP (including a warning that it's disabled!)
- Fix links to DSP Settings in grouped playback UIs
- Disable DSP config UI for Universal Group players


music_assistant/helpers/audio.py
```diff
@@ -836,6 +837,13 @@ def get_player_filter_params(
 
     dsp = mass.config.get_player_dsp_config(player_id)
 
+    if player := mass.players.get(player_id):
+        is_grouped = bool(player.synced_to) or bool(player.group_childs)
+        if is_grouped and PlayerFeature.MULTI_DEVICE_DSP not in player.supported_features:
+            # We can not correctly apply DSP to a grouped player without multi-device DSP support,
+            # so we disable it.
+            dsp.enabled = False
```
marcelveldt, line 845: If you are wondering about a way how we can tell the user what is happening:

I was thinking that we should maybe also extend streamdetails the same way we do now for volume normalization etc. That way the frontend can show the user exactly the whole chain:

- The input file/stream/codec
- Filters like volume normalization
- DSP on/off and what filters
- Output path (incl lossy or lossless flag and sample rate etc)

Related to this one here: https://github.com/music-assistant/frontend/pull/651

music_assistant/providers/player_group/ugp_stream.py
```diff
@@ -71,13 +75,26 @@ async def subscribe(self) -> AsyncGenerator[bytes, None]:
...
+            yield chunk
+
     async def _runner(self) -> None:
         """Run the stream for the given audio source."""
         await asyncio.sleep(0.25)  # small delay to allow subscribers to connect
         async for chunk in get_ffmpeg_stream(
             audio_input=self.audio_source,
             input_format=self.input_format,
-            output_format=self.output_format,
+            output_format=self.input_format,
```
marcelveldt: I was confused for a seconds only realizing that this means that we use the same format for in- and output here only to apply the readrate. Maybe we should rename this to "base_format" or "base_pcm_format" of whatever to make it more clear or should I just get more coffee ?

maximmaxim345: Done

marcelveldt (approved):
LGTM, nice work!

marcelveldt:
models have been bumped btw

PR #1846, opened 2025-01-09
Small fix for checking playlist url

ztripez:
fix: allow m3u/m3u8 detection with query parameters

Previously, URLs containing query parameters were not being correctly detected
as m3u/m3u8 playlists. This patch ensures URLs with query strings are properly
identified as valid playlist formats.


marcelveldt (approved):
Nice one, thanks!

PR #1847, opened 2025-01-09
fix: Make sure that Jellyfin playlists are sorted and paginated

Jc2k:
We were using the wrong endpoint for playlists so they were unsorted. We also need to use the order that they are returned in, rather than any index from the track object itself.

This also fixes pagination for that API endpoint.


Jc2k:
Not certain but this may help with https://github.com/music-assistant/support/issues/3278 as it is using the correct interface for fetching playlist tracks. Whilst the old way "worked" (albeit unsorted) its possible its a fluke of the specific data in my Jellyfin instance. Using a playlist specific endpoint may account for whatever data difference this is.

marcelveldt (approved):

PR #1849, opened 2025-01-09
chore: enable mypy for new modules by default

Jc2k:
All new modules must pass mypy.

Existing modules should be removed from the exclude list as they are fixed.

There was conflicting config in mypy.ini and pyproject.toml, so this merges them to be stricter and to only use pyproject.toml.


marcelveldt (approved):
Nice, appreciate your efforts getting the code in the right shape for testing and linting!

PR #1850, opened 2025-01-09
Add LastFM Audio Scrobbler plugin provider

wjzijderveld:
An initial basic implementation to support scrobbling via Last.fm and Libre.fm (I've not tried it with Libre yet).

It probably could use some better explanation how to create an key and secret.

Also this is my first time hacking on music-assistant and one of the few times I worked in Python, so please let me know if I'm being super dumb somewhere 😅


music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+        self.mass.subscribe(self._on_mass_queue_updated, EventType.QUEUE_UPDATED)
+        self.mass.subscribe(self._on_mass_media_item_played, EventType.MEDIA_ITEM_PLAYED)
+
+    async def _on_mass_queue_updated(self, event: MassEvent) -> None:
+        """Player has updated, update nowPlaying."""
+        if self._network is None:
+            return
+
+        queue: PlayerQueue = event.data
+        if queue.state == PlayerState.PLAYING and queue.current_item.media_type == MediaType.TRACK:
```
wjzijderveld: This needs some better checks, to ideally only trigger when the queue state changes from idle to playing, or when it switches to a new track.

marcelveldt: we now have a beter event for that MEDIA_ITEM_PLAYED
So if you listen for EventType.MEDIA_ITEM_PLAYED you get an event everytime something has been played, including how many seconds and if the item was fully played

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+
+        if self.config.get_value(CONF_SESSION_KEY):
+            self._network = _get_network(self._get_network_config())
+
+        # could be interesting for a "scrobble after X seconds" feature
+        # self.mass.subscribe(self._on_mass_queue_time_updated, EventType.QUEUE_TIME_UPDATED)
+        self.mass.subscribe(self._on_mass_queue_updated, EventType.QUEUE_UPDATED)
+        self.mass.subscribe(self._on_mass_media_item_played, EventType.MEDIA_ITEM_PLAYED)
+
+    async def _on_mass_queue_updated(self, event: MassEvent) -> None:
```
marcelveldt: I wouldt listen to this event for scrobbling at all but only to MEDIA_ITEM_PLAYED

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+                )
+            except Exception as err:
+                self.logger.exception(err)
+
+    async def _on_mass_media_item_played(self, event: MassEvent) -> None:
+        """Media item has finished playing, we'll scrobble the track."""
+        if self._network is None:
+            return
+
+        item = await self.mass.music.get_item_by_uri(event.object_id)
```
marcelveldt: event.data also contains a media_item attribute so you wont have to fetch it , saves a potential api call

OzGav:
Out of interest will you be able to scrobble to last.fm and libre.fm at the same time? Have two instances for example?

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+                    artist=track.artist_str,
+                    title=track.name,
+                    mbid=track.mbid,
+                    album=track.album.name if track.album else None,
+                )
+            except Exception as err:
+                self.logger.exception(err)
+
+    async def _on_mass_media_item_played(self, event: MassEvent) -> None:
+        """Media item has finished playing, we'll scrobble the track."""
```
OzGav: I believe the Subsonic provider scrobbles if half the track is played. Is there a standard for this? (Reference: https://github.com/music-assistant/support/issues/3203#issuecomment-2571387547 )

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+        self.mass.subscribe(self._on_mass_media_item_played, EventType.MEDIA_ITEM_PLAYED)
+
+    async def _on_mass_queue_updated(self, event: MassEvent) -> None:
+        """Player has updated, update nowPlaying."""
+        if self._network is None:
+            return
+
+        queue: PlayerQueue = event.data
+        if queue.state == PlayerState.PLAYING and queue.current_item.media_type == MediaType.TRACK:
+            track: Track = queue.current_item.media_item
```
Jc2k: We have a helper called is_track which implements a typing TypeGuard. If you use it here you won't need the explict type.

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: That's a different context though, there's 2 different calls, `track.updateNowPlaying` ([docs](https://www.last.fm/api/show/track.updateNowPlaying)) is an indicator about the current song that's playing and won't result in an actual scrobble, while `track.scrobble` ([docs](https://www.last.fm/api/show/track.scrobble)) is the actual scrobbling of a played song. 
So one close to what you often see on Discord for example, where people have the current song they're listening to as a status, while the other is historical tracking. And thus both have a different trigger.

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: doh, no idea how I missed that, I'll update that, thanks!

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: It's one of the things I want to look at, 50% is fairly common, but often configurable. 
For now I just focussed on the generic config and basic events, as currently it's not super easy to trigger something based on percentage played.

wjzijderveld:
> Out of interest will you be able to scrobble to last.fm and libre.fm at the same time? Have two instances for example?

Not at the moment and honestly I've no idea what it would involve. I saw there is an option to allow multiple instances of a plugin, so maybe that would be enough, but I would need to some rigorous testing first.

music_assistant/providers/audioscrobbler/__init__.py
marcelveldt: ah! got it! all clear now then and ignore my previous comment

music_assistant/providers/audioscrobbler/__init__.py
marcelveldt: vm as explained above this is for reporting what is in progress

music_assistant/providers/audioscrobbler/__init__.py
marcelveldt: _on_mass_media_item_played will always be triggered and includes the amount of seconds played. 
So you can make your own logic together with the item's duration

Check the event.data of this event, it has all the info you need.

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+
+    _network: pylast._Network = None
+
+    def __init__(
+        self, mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+    ) -> None:
+        """Initialize AudioScrobbler."""
+        super().__init__(mass, manifest, config)
+        if self.logger.level == logging.DEBUG:
+            pylast.logger.setLevel(logging.DEBUG)
```
marcelveldt: Move this into setup

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+
+        try:
+            track: Track = item
+            self._network.scrobble(
+                artist=track.artist_str,
+                title=track.name,
+                timestamp=time.time(),
+                mbid=track.mbid,
+                album=track.album.name if track.album else None,
+            )
```
marcelveldt: this is doing a blocking call - you need to either wrap the calls to this library into an executor or do the calls yourself using an async web client.

requirements_all.txt
```diff
@@ -34,6 +34,7 @@ py-opensonic==5.2.1
 pyblu==2.0.0
 PyChromecast==14.0.5
 pycryptodome==3.21.0
+pylast
```
marcelveldt: pinned versions are required

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: I've added a basic check for now, but usually the approach is a bit different. For example, often you'd actually detect length that's being played, so skipping ahead would not cause a scrobble. 
And with the current setup, when a song gets paused for example, played and paused near the end of the song, it gets scrobbled every time. I could try to detect that, but having that work while also accounting for songs that are on repeat would be annoying.

These are all nice to haves though, we can iterate later if we want.

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: I figured out why I missed that. Because it's not true :P `data["media_item']` contains the reference as well, not the item. I played around a bit with passing the actual item, but I got serialization errors.

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: Done, it's been a while since I worked in an eventlooped system. Combined with my lack of python experience I just saw `async` and assumed all was good 😂

music_assistant/providers/audioscrobbler/__init__.py
wjzijderveld: I've used it in 2 places, but honestly the typing support in Python confuses me a bit and I probably don't have my editor setup correct to understand the TypeGuard, so I still left the `: Track` typehint here.

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,325 @@
...
+                try:
+                    await cb(event)
+                except Exception as e:
+                    self.logger.exception(e)
+
+            return handler
+
+        # subscribe to internal events
+        self.mass.subscribe(wrap(self._on_mass_queue_updated), EventType.QUEUE_UPDATED)
+        self.mass.subscribe(wrap(self._on_mass_media_item_played), EventType.MEDIA_ITEM_PLAYED)
```
marcelveldt: I'm not a fan of this wrapping of event logic tbh
The core is already creating tasks for each event handler. 
Please just subscribe a callback as intended

wjzijderveld: Ah yeah, I forgot to add a comment about it. I'm not a fan either, so I'm happy to remove it.
I added it because I had some trouble with silent exceptions, which was very frustrating.

marcelveldt: hmmm, exceptions should never go silent as we have a global catch-all exception handler for uncaught ones.  
I'll have to look into that, maybe a bug.

wjzijderveld: I don't know if there's a different in Python, but among others I ran into `method not found` kind of exceptions that didn't get logged. I imported `is_track` within the `TYPE_CHECKING` block, which (obviously) didn't work.

marcelveldt: these should be catched by the catch-all exception handler. Are you running the debugger in vscode according to the predefined settings?

music_assistant/providers/audioscrobbler/__init__.py
marcelveldt: I updated the object with the info present. We must prevent fetching full item details where/if we can to save api calls

OzGav:
@wjzijderveld once you get close to merging can you provide some text for the docs? We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    provider = ScrobblerProvider(mass, manifest, config)
+    pylast.logger.setLevel(provider.logger.level)
+    if provider.logger.level == logging.DEBUG:
+        # httpcore is quite spammy without providing useful information 99% of the time
+        logging.getLogger("httpcore").setLevel(logging.INFO)
+
+    return provider
+
+
+class ScrobblerProvider(PluginProvider):
```
marcelveldt: Not sure if the idea is to extend this scrobble provider with other targets but otherwise I think should be renamed to LastFM Scrobble Provider and mention in the readme/description that its also suitable for LibreFM.

marcelveldt:
@wjzijderveld can you mark this PR as "ready for review" if you are ready for the (final?) review and merge ?

OzGav:
@wjzijderveld Just a reminder for the docs info as well please!

OzGav:
@wjzijderveld version 2.4 of MA will ship soon. It would be great if this could be included

wjzijderveld:
Sorry, I've been very pre-occupied. I'll try to make some time for it today or tomorrow to get it over the line!

wjzijderveld:
I've worked on it for a while yesterday. The updated event does give me most information needed without doing an extra API call. Although the 30s interval currently makes it impossible to detect if I should scrobble the track, unless I rely on `fully_played` which I'm doing now. I'll have to look a bit deeper another time to see how we can improve that, what extra info would be useful in the reporting event. 

But I'm having a lot of trouble committing the result, as mypy is complaining about so much stuff it didn't report earlier.

```
music_assistant/providers/lastfm/__init__.py:77: error: Argument 1 to "subscribe" of "MusicAssistant" has incompatible type "Callable[[MassEvent], Coroutine[Any, Any, None]]"; expected "Callable[[MassEvent], None] | Coroutine[MassEvent, Any, None]"  [arg-type]
music_assistant/providers/lastfm/__init__.py:80: error: Argument 1 to "subscribe" of "MusicAssistant" has incompatible type "Callable[[MassEvent], Coroutine[Any, Any, None]]"; expected "Callable[[MassEvent], None] | Coroutine[MassEvent, Any, None]"  [arg-type]
music_assistant/providers/lastfm/__init__.py:218: error: Incompatible types in assignment (expression has type "str | int | float | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None", variable has type "str")  [assignment]
music_assistant/providers/lastfm/__init__.py:252: error: Value of type "dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None" is not indexable  [index]
music_assistant/providers/lastfm/__init__.py:253: error: Argument 1 to "_get_network" has incompatible type "dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None"; expected "dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None]"  [arg-type]
music_assistant/providers/lastfm/__init__.py:273: error: Unsupported target for indexed assignment ("dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None")  [index]
music_assistant/providers/lastfm/__init__.py:274: error: Unsupported target for indexed assignment ("dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None")  [index]
music_assistant/providers/lastfm/__init__.py:282: error: Argument "label" to "ConfigEntry" has incompatible type "tuple[str, str]"; expected "str"  [arg-type]
music_assistant/providers/lastfm/__init__.py:292: error: Unsupported target for indexed assignment ("dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None")  [index]
music_assistant/providers/lastfm/__init__.py:293: error: Unsupported target for indexed assignment ("dict[str, str | int | float | bool | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None] | None")  [index]
music_assistant/providers/lastfm/__init__.py:337: error: Incompatible types in assignment (expression has type "str | int | float | list[str] | list[tuple[int, int]] | tuple[int, int] | list[int] | Enum | None", variable has type "str")  [assignment]
``` 

Having so many union types isn't helping to parse these kind of messages 😅 
Did something change recently that made the rules stricter?

marcelveldt:
> Although the 30s interval currently makes it impossible to detect if I should scrobble the track, unless I rely on fully_played which I'm doing now

The idea is that you decide for yourself if you like to scrobble while playback is busy or you only scrobble when a track has been fully played. For example the audiobook providers use this info to sync progress but for scrobbling i think its more common to scrobble only fully played tracks. 

> But I'm having a lot of trouble committing the result, as mypy is complaining about so much stuff it didn't report earlier.

Looks like you simply need a couple of asserts and casts to help the python linter. Especially with config values as those can be any of a broad range of value types.

music_assistant/providers/audioscrobbler/__init__.py
marcelveldt: this comment remains, the providers is targeted to be a lastfm/librefm scrobble provider

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+        await super().loaded_in_mass()
+
+        if not self.config.get_value(CONF_SESSION_KEY):
+            self.logger.info("No session key available")
+            return
+
+        self._network = _get_network(self._get_network_config())
+
+        # subscribe to internal events
+        self.mass.subscribe(self._on_mass_queue_updated, EventType.QUEUE_UPDATED)
```
marcelveldt: Better use the MEDIA_ITEM_PLAYED event for this as that fires once/after a track has been playing for more than 30 seconds so you ignore fast skipping of tracks

music_assistant/providers/audioscrobbler/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    async def _on_mass_media_item_played(self, event: MassEvent) -> None:
+        """Media item has finished playing, we'll scrobble the track."""
+        if self._network is None:
+            self.logger.error("no network available during _on_mass_media_item_played")
+            return
+
+        item = await self.mass.music.get_item_by_uri(event.data["media_item"])
+        if not is_track(item):
+            self.logger.debug(f"track {event.data['media_item']} skipped, as it is not a track")
+            return
```
marcelveldt: I'm not following why its needed to request a full item details (which is also a potential call to a streaming provider api) - lets try to prevent this and work with the event data we already have

wjzijderveld:
I've not pushed the changes yet, as mypy wouldn't allow me to commit my changes, so you're commenting on old code :)

I'll add more asserts and casts, I already started on it earlier, but it was just confusing as these linter errors didn't pop up earlier and I see very similar code that should also trigger the linter. So I decided to first ask if something changed.

marcelveldt:
I actually found one wrong annotation on the subscribe method itself which is now fixed if you pull the latest from dev. That might already fix one or two of your issues.

If you want some help in fixing the found issues, you can decide to temporary disable pre-commit and commit your changes so I can help commenting on the new code.

Just run "pre-commit uninstall" on your venv to disable it and commit.
Simply run the MA setup script again to reinstate it

wjzijderveld:
Sorry for the delay! I've wrapped up the mypy issues and did a small refactor to only depend on `MEDIA_ITEM_PLAYED` now. 
There's an edge-case that now isn't covered though, but it requires some changes in the queue controller; if a a single song is played on repeat and the length of the song is in specific ranges (f.e. 3m07s), it will never get scrobbled. As the queue item id doesn't change and the `queue.elapsed_time % 30` does not trigger withing 90% of the duration of the song. 

I still don't think that `fully_played` is desirable to based scrobbling on, LastFM defaults to 50% in their own app as well for example, so in the future I'd like to look at changing that. But the current logic is fine in most cases and I don't want to stall this PR even longer 😅

edit: I'll add some copy for the docs in a bit.

wjzijderveld:
@OzGav for the docs:

### Features
Connect to either LastFM or LibreFM and scrobble the songs you are playing. It also supports the now playing feature.

### Configuration
First get your API credentials. For LastFM follow the first two steps to get your credentials: https://www.last.fm/api/authentication
Enter your credentials and click "Authorize with LastFM", this will bring you to LastFM where you need to give permission by clicking "Yes, allow access". 
In Music Assistant it will show you if you're successfully logged in. DON'T FORGET TO HIT SAVE, otherwise your freshly acquired authorization won't be persisted. 

### Known issues / Notes
- Currently songs will only get scrobbled when they're fully played (90+%). 
- Be careful for double scrobbling, which is possible if you use a music provider that also does scrobbling internally. A future improvement is to make it configurable which music providers will be scrobbled.

music_assistant/providers/lastfm/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+                    report.album,
+                    duration=report.duration,
+                    mbid=report.mbid,
+                )
+            except Exception as err:
+                self.logger.exception(err)
+
+        # update now playing if needed
+        if self._currently_playing is None or self._currently_playing != report.uri:
+            self.mass.loop.run_in_executor(None, update_now_playing)
```
marcelveldt:
```suggestion
            await asyncio.to_thread(update_now_playing)
```

music_assistant/providers/lastfm/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+
+        if self.should_scrobble(report):
+            self.mass.loop.run_in_executor(None, scrobble)
```
marcelveldt:
```suggestion
            await asyncio.to_thread(scrobble)
```

music_assistant/providers/lastfm/__init__.py
```diff
@@ -0,0 +1,304 @@
...
+            self._currently_playing = None
+
+    def should_scrobble(self, report: MediaItemPlaybackProgressReport) -> bool:
+        """Determine if a track should be scrobbled, to be extended later."""
+        # ideally we want more precise control
+        # but because the event is triggered every 30s
+        # and we don't have full queue details to determine
+        # the exact context in which the event was fired
+        # we can only rely on fully_played for now
+        return bool(report.fully_played)
```
marcelveldt: you also have a "is_playing" attribute on the event now.
So you could consider the following logic:

scrobble if playing and progress >= 50%
or
scrobble if fully played 

In the future we can also replace fully played and is_playing booleans with some enum value of specific state

music_assistant/providers/lastfm/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "plugin",
+  "domain": "lastfm",
```
marcelveldt: Chances are big that we also want a lastfm metadata provider at some point. Now I'm wondering if we can squeeze that all into one provider (so with a single login, have both metadata retrieval such as recommendations etc and scrobbling). In that case we can keep this the lastfm provider and extend it later.

If that becomes too messy or it just touches completely different aspects of lastfm api, both become a dedicated provider.

Concrete my question is if we should keep the domain just "lastfm" or maybe "lastfm_scrobble"

music_assistant/providers/lastfm/__init__.py
wjzijderveld: oh nice, that does help. Although I'm not sure what the direct implications are, as I then need to keep state to not double trigger a scrobble. Which becomes interesting when considering single songs being on loop (which I personally do quite a lot 😂), as it's hard to detect when it restarts. 

I tried to fix it when I looked at the edge-case I mentioned, but it's not very straightforward currently, so I didn't add it to the MR. But it would be nice if the playback report had an indication that the event specifically indicates start of queued item and end of queued item.

music_assistant/providers/lastfm/manifest.json
wjzijderveld: Does that mean `domain` is a unique identifier? Based on the name I expected it to be a grouping, but honestly I didn't really look into how its being used.
edit: I just checked the docs again and it clearly states it's the unique identifier 🤦

For example, I would assume something like this would be possible:
```
providers/lastfm_scrobble/manifest.json:
  { 
    type: plugin,
    name: lastfm_scrobble,
    domain: lastfm,
  }
providers/lastfm_metadata/manifest.json:
  {
    type: metadata,
    name: lastfm_metadata,
    domain: lastfm
  }
``` 

I don't know if I can answer this properly to be honest. I don't see a direct issue from an API point of view, the API is quite "dumb", there are no scopes for example, so a single set of credentials would work fine and the library I use implements the entire API, so retrieving metadata should be possible.

But when I look at it based on my experience and from a security perspective, I would always try to be as granular as possible (the zero-trust model) and keep those concerns separate.

music_assistant/providers/lastfm/__init__.py
marcelveldt: OK, we can extend/iterate on it later once this is in and we have some feedback and actually daily use. 
Not sure if you're on discord but we could have a chat there how to best move forward on this.

music_assistant/providers/lastfm/manifest.json
marcelveldt: Yeah, let's go with "lastfm_scrobble" then now and add a "lastfm_metadata" later. Although they share the same api (and library) they represent 2 completely different usecases.

music_assistant/providers/lastfm/manifest.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "LastFM Scrobbler",
+  "description": "Scrobble your music to Last.fm and others with a compatible API like Libre.fm",
+  "codeowners": ["@music-assistant"],
+  "documentation": "",
```
OzGav:
```suggestion
  "documentation": "https://beta.music-assistant.io/plugins/lastfm_scrobble/",
```

wjzijderveld: Just checking, will this still be correct with the rename I just did? `lastfm` => `lastfm_scrobble`

music_assistant/providers/lastfm/__init__.py
wjzijderveld: Yeah I am, same username, although not really active besides in one of the LastFM threads 😂

music_assistant/providers/lastfm/manifest.json
marcelveldt: let's keep the name of the docs the same as the domain

music_assistant/providers/lastfm_scrobble/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "plugin",
+  "domain": "lastfm_scrobble",
+  "name": "LastFM Scrobbler",
+  "description": "Scrobble your music to Last.fm and others with a compatible API like Libre.fm",
+  "codeowners": ["@music-assistant"],
+  "documentation": "",
```
marcelveldt:
```suggestion
  "documentation": "https://music-assistant.io/plugins/lastfm_scrobble/",
```

marcelveldt (commented):
Nice work @wjzijderveld !

wjzijderveld:
Thanks for all the help and feedback! :pray: 
When working in new projects and languages its always nice to get some support, so its really appreciated!

PeachiPrism:
Unfortunately this implementation doesn't support album artists being different from artists which is a bit disappointing, is this something that could eventually be added or is this a limitation of Music Assistant or Home Assistant not having that as a line in the medias state?

Quite annoying to have to avoid Various Artist albums using MA or having to pay for Last FM Pro to set up autocorrections for songs in a playlist that have a different album artist from track artist.
Having a scrobbler plugin in MA itself is cool either way but currently because of the album artist thing it doesn't do anything more special than https://github.com/valentin-gosselin/lastfm-scrobbler-ha-integration apart from being less complicated to set up (Thank you for not making us have to run a python script to get it working)

OzGav:
@PeachiPrism I don't know what you mean. I am playing an album right now which has the album artist tagged as "Various Artists" and it is scrobbling each individual track correctly.  Also don't add comments to merged or closed topics as they will get lost

PR #1851, opened 2025-01-09
Mypy fixes

Jc2k:
See https://github.com/music-assistant/models/pull/30. These are the typing failures that have snuck in because py.typed escaped the new models pypi package.

Not sure if this will work without https://github.com/music-assistant/models/pull/30 + tag + bump in here.

There are 2 i'm not sure about:

```
music_assistant/providers/theaudiodb/__init__.py:331: error: Argument 2 to "update_item_in_library" of "MediaControllerBase" has incompatible type "Artist | ItemMapping"; expected "Artist"  [arg-type]
music_assistant/providers/theaudiodb/__init__.py:377: error: Argument 2 to "update_item_in_library" of "MediaControllerBase" has incompatible type "Artist | ItemMapping"; expected "Artist"  [arg-type]
```

### DIFF
--- a/music_assistant/constants.py
+++ b/music_assistant/constants.py
@@ -3,11 +3,8 @@
 import pathlib
 from typing import Final
 
-from music_assistant_models.config_entries import (
-    ConfigEntry,
-    ConfigEntryType,
-    ConfigValueOption,
-)
+from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption
+from music_assistant_models.enums import ConfigEntryType
 
 API_SCHEMA_VERSION: Final[int] = 26
 MIN_SCHEMA_VERSION: Final[int] = 24
--- a/music_assistant/controllers/media/albums.py
+++ b/music_assistant/controllers/media/albums.py
@@ -39,7 +39,7 @@ if TYPE_CHECKING:
     from music_assistant.models.music_provider import MusicProvider
 
 
-class AlbumsController(MediaControllerBase[Album]):
+class AlbumsController(MediaControllerBase[Album, Album]):
     """Controller managing MediaItems of type Album."""
 
     db_table = DB_TABLE_ALBUMS
--- a/music_assistant/controllers/media/artists.py
+++ b/music_assistant/controllers/media/artists.py
@@ -36,7 +36,7 @@ if TYPE_CHECKING:
     from music_assistant.models.music_provider import MusicProvider
 
 
-class ArtistsController(MediaControllerBase[Artist]):
+class ArtistsController(MediaControllerBase[Artist, Artist | ItemMapping]):
     """Controller managing MediaItems of type Artist."""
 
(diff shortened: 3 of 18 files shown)


marcelveldt:
models package is released as 1.1.0 maybe try bumping that ?

Jc2k:
I'll bump and fix any errors I can in next few hours. Cheers for quick release!

Jc2k:
Have rebased and bumped the models package. Just need to figure out these last 2. Looks like all MediaControllerBase must take a MediaItemType, and that is not an ItemMapping? So should I assert that I have an Artist before calling update_item_in_library

marcelveldt:
> Have rebased and bumped the models package. Just need to figure out these last 2. Looks like all MediaControllerBase must take a MediaItemType, and that is not an ItemMapping? So should I assert that I have an Artist before calling update_item_in_library

Yeah, MediaControllerBase only take MediaItemType and that can never ever be an ItemMapping.
We only use ItemMapping for nested objects and incomplete/fast listings (such as search).
Now I did have a fair share of fighting with typing when I implemented that base class with the dynamic type inheritence so there may be something super stupid hidden inside it :-)

marcelveldt:
Darn yeah... Artist is an exception where we sometimes take ItemMapping - I believe that is done to prevent ourselves from having to query an entire chain of objects when we just want one track added to the lib. So we accept the ItemMapping, try to resolve it lazy and if not accept it as input as an artist object has no nested objects.

Long story short: The ArtistController takes an Artist or ItemMapping as input for the add and update methods

Jc2k:
Geez thats a rabbit hole.

I can lie and say all instances of update_library_item take an ItemMapping, which lets the current setup pass. But that's wrong, *only* artist can. But we can't violate the typing of the base class.

I can change the ItemCls bound to accept `MediaItemType | ItemMapping` then back the artist controller take `Artist | ItemMapping`. But that infects everywhere and suddenly loads of stuff starts complaining that i'm trying to pass it an ItemMapping. Hate hate hate.

So that leads us to: https://github.com/music-assistant/server/pull/1851/commits/e49daee0431ac3d68bbd07fdb0831c7a7963c266. Use a different generic parameter just for the library update. That does have the relaxed bound, and its fine for it to be Artist | ItemMapping on the Artist controller and not be ItemMapping anywhere else. I don't know this code very well, so I don't know if there are better names for things. I haven't done much testing yet. But what do you think?

marcelveldt:
Ha creative but genius - yeah this is good, it keeps the strict typing (and hints). Nice work!
The easy way could indeed be to simply accept ItemMapping for the other objects as well just to satisfy typing but that will lead us into a giant shitshow if you try add an itemmapping to the lib or seomething like that. 

ItemMapping's are evil but we need them to be efficient :-)

Thanks!

marcelveldt (approved):

PR #1854, opened 2025-01-10
chore: mypy for sonos s1

Jc2k:


music_assistant/providers/sonos_s1/player.py
```diff
@@ -763,7 +767,7 @@ def _speaker_activity(self, source: str) -> None:
     @soco_error()
     def _join(self, members: list[SonosPlayer]) -> list[SonosPlayer]:
         if self.sync_coordinator:
-            self.unjoin()
+            self._unjoin()
```
Jc2k, line 770: Think this might be an actual bug its found? Can't test this as I only have one sonos speaker.

marcelveldt: yep, just checked it

marcelveldt (approved):
Nice, thanks @Jc2k !

PR #1857, opened 2025-01-12
Add Audiobookshelf audiobooks & podcast provider

fmunkes:
# Initial support for Audiobookshelf
This PR is an implementation of [Audiobookshelf](https://audiobookshelf.org) (abs) as music provider.

## Features
* Populates Audiobooks from all libraries accessible by user.
* Populates Podcasts from all libraries accessible by user.
* Browse Feature shows the libraries names and podcasts/ books within.
* Progress reporting both ways (see issues below).
* API client for mass - I didn't find a suitable python lib, so I implemented what is needed.

## Possible future improvements
* Browse by author, series, narrator etc.
* Search Feature
* Edit Feature

## Known issues:
- Progress reporting:
    * abs to mass seems ok
    * mass to abs:
        * on_streamed function of MusicProvider:
            - only called, when next/ previous used in UI (apparently?).
            - not called on pause
            - not called regularly
        * I.e. abs progress is only updated in first case.
    * Which function is called, when "mark as (un)played" is used in the UI?
- Audiobooks:
    Am I correct, that mass only supports single file based audiobooks? abs may list multiple
    audio files per book, so if this is the case, that's something for the documentation. abs even
    offers an ffmpeg concat via its interface. Code-wise currently only the first file of an audiobook
    is respected.
- mypy:
    Is currently disabled for this provider. Unfortunately I had troubles with some type hints, due to
    a lack of knowledge on my side. Happy to figure them out.

Personal note: I used Python quite often during the years, however, I'm not a professional developer.
If you think the code is not worth pursuing, don't hesitate to let me know :). I still learned something.

Thanks for the amazing project!


pyproject.toml
```diff
@@ -126,6 +126,7 @@ exclude = [
   '^music_assistant/providers/_template_music_provider/.*$',
   '^music_assistant/providers/_template_player_provider/.*$',
   '^music_assistant/providers/apple_music/.*$',
+  '^music_assistant/providers/audiobookshelf/.*$',
```
Jc2k: We are working towards making this list empty, we shouldn't be adding to it. Feel free to ask questions if you need help.

music_assistant/providers/audiobookshelf/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "music",
+  "domain": "audiobookshelf",
+  "name": "Audiobookshelf",
+  "description": "Audiobookshelf (audiobookshelf.org) as audiobook and podcast provider",
+  "codeowners": [
+    "@fmunkes"
+  ],
+  "requirements": [],
+  "documentation": "Link to the documentation on the music-assistant.io helppage (may be added later).",
+  "mdns_discovery": []
```
Jc2k: You can remove this key entirely if the list is blank.

music_assistant/providers/audiobookshelf/abs_client.py
```diff
@@ -0,0 +1,372 @@
...
+            abs_media.chapters.append(abs_chapter)
+
+        abs_media.tracks = []
+        for track in audiobook["media"]["tracks"]:
+            abs_audio_track = await self._map_attributes(ABSAudioTrack, track)
+            abs_media.tracks.append(abs_audio_track)
+
+        return abs_audiobook
+
+    async def _map_attributes(self, dclass, api_dict: dict[str, Any]):
```
Jc2k: This doesn't do any I/O so doesn't need to be async, and shouldn't be awaited.

It looks like the kind of thing where you probably should use https://pypi.org/project/mashumaro/. If your dataclasses are set up properly you'd be able to do `ABSAudioBook.from_json(some_json)` and get rid of most of the function above.

That would solve your typing query, too, as the deserialization would be handled in mashumaro, so typing it wouldn't be your problem.

music_assistant/providers/audiobookshelf/abs_client.py
```diff
@@ -0,0 +1,372 @@
...
+        self.audiobook_libraries: list[ABSLibrary] = []
+        self.user: ABSUser
+        self.check_ssl: bool
+
+    async def init(
+        self, base_url: str, username: str, password: str, check_ssl: bool = True
+    ) -> None:
+        """Initialize."""
+        self.check_ssl = check_ssl
+        self.session = ClientSession(base_url)
```
Jc2k: There is a shared client session at `self.mass.http_session` on the provider that you should use (though that means handling base_url yourself in _get/_post).

music_assistant/providers/audiobookshelf/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "music",
+  "domain": "audiobookshelf",
+  "name": "Audiobookshelf",
+  "description": "Audiobookshelf (audiobookshelf.org) as audiobook and podcast provider",
+  "codeowners": [
+    "@fmunkes"
+  ],
+  "requirements": [],
+  "documentation": "Link to the documentation on the music-assistant.io helppage (may be added later).",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/music-providers/audiobookshelf",
```

OzGav:
@fmunkes can you provide some text for the docs? We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. I could extract most of this from your opening post in the PR but I am conscious of the improvements you may have made since then. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

marcelveldt:
> on_streamed function of MusicProvider - only called, when next/ previous used in UI (apparently?).

We're looking into this one, seems to be a bug. Expect to be solved soon.

> on_streamed function of MusicProvider - not called on pause

No and that is by design. It will be called when playback completed (full item played or playback stopped) which is enough imo.

> on_streamed function of MusicProvider - not called regularly

No and that is by design. Why would we want to report it regularly ?

> Which function is called, when "mark as (un)played" is used in the UI?

None, which should be fixed. I will adjust it so it calls on_streamed with 0 seconds or something (I'll chew on it a bit)

> Am I correct, that mass only supports single file based audiobooks? 

This is still topic of discussion as there seem to be multiple different implementations in the wild but the most common one is a single file per audiobook or podcast-episode with optional chapters. 
What we have now is chapter support in a single audiobook/episode. So if you have a file per chapter that will be a challenge

> mypy Is currently disabled for this provider.

OK, that will have to be fixed before we can merge but just ask for help and we will help you

pyproject.toml
fmunkes: I removed the line.

music_assistant/providers/audiobookshelf/abs_client.py
fmunkes: I've used the suggested library, and it resolved most typing issues I faced. Thank you!

music_assistant/providers/audiobookshelf/abs_client.py
fmunkes: Shared client session is used, and _get, _post, _patch are adjusted accordingly.

fmunkes:
@OzGav documentation:

### FEATURES
- Populates Audiobooks from all libraries accessible by supplied user.
- Populates Podcasts from all libraries accessible by supplied user.
- Browse Feature shows the library names and podcasts/ books within.
- Progress reporting both ways
### CONFIGURATION
- For setup you need
  - the URL of an Audiobookshelf instance
  - username of an Audiobookshelf user
  - password of this user
- You may optionally disable ssl verification.
### KNOWN ISSUES/ NOTES
Audiobookshelf supports multiple files (e.g. one file per chapter) per Audiobook. Music Assistant however expects a single file. It is suggested to merge multiple files into a single file. You can do this in Audiobookshelf by selecting the Audiobook, click edit, then tools and finally "Open Manager". You will see the merge option there. Do not forget to take a backup of your initial files.
### NOT YET SUPPORTED
- Browsing by author
- Browsing by collection
- Browsing by series
- Browsing by playlists

fmunkes:
> > on_streamed function of MusicProvider - only called, when next/ previous used in UI (apparently?).

> We're looking into this one, seems to be a bug. Expect to be solved soon.

Thanks - I think my other "on_streamed remarks" become obsolete when this is resolved.

> > Which function is called, when "mark as (un)played" is used in the UI?

> None, which should be fixed. I will adjust it so it calls on_streamed with 0 seconds or something (I'll chew on it a bit)

Thank you!

> > Am I correct, that mass only supports single file based audiobooks?

> This is still topic of discussion as there seem to be multiple different implementations in the wild but the most common one is a single file per audiobook or podcast-episode with optional chapters. What we have now is chapter support in a single audiobook/episode. So if you have a file per chapter that will be a challenge

I think for the time being it is sufficient to just add a comment to the documentation (maybe also to the overall one of 'Audiobooks'?). In the case of Audiobookshelf, merging into one file can be achieved within the UI.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -0,0 +1,527 @@
...
+        """Return True if the provider is a streaming provider."""
+        # For streaming providers return True here but for local file based providers return False.
+        return True
+
+    async def sync_library(self, media_types: tuple[MediaType, ...]) -> None:
+        """Run library sync for this provider."""
+        await self._client.sync()
+        await super().sync_library(media_types=media_types)
+
+    async def _parse_podcast(self, abs_podcast: ABSPodcast) -> Podcast:
```
Jc2k: This doesn't do network i/o, so doesn't need to be marked as async or awaited.

fmunkes:
There is a currently a single mypy error left. Everything else was resolved by switching to mashumaro as suggested.
I get:
`music_assistant/providers/audiobookshelf/__init__.py:496: error: Incompatible types in assignment (expression has type "ABSAudioBook", variable has type "ABSPodcast")  [assignment]`
Don't see what's wrong there currently. I'll check again tomorrow.

music_assistant/providers/audiobookshelf/abs_client.py
```diff
@@ -0,0 +1,303 @@
...
+    ABSMediaProgress,
+    ABSPodcast,
+    ABSUser,
+)
+
+# use page calls in case of large libraries
+LIMIT_ITEMS_PER_PAGE = 10
+
+
+def camel_to_snake(camel: str) -> str:
```
Jc2k: Is this still used somewhere or can we get rid of it?

music_assistant/providers/audiobookshelf/icon.svg
Jc2k, line 1: Can we get a logo for ABS?

Jc2k (commented):
RE your mypy error, that's a mypy quirk. Not sure whether its a limitation or a deliberate decision, but you can't use a variable twice for different types. Line 493 makes it think `item` is a podcast. Then on line 496 `item` is an audio book. I don't know if there is a better way, but if you had something link `async from audiobook in ...` it will probably work.

music_assistant/providers/audiobookshelf/icon.svg
OzGav: Here is one https://iconduck.com/icons/252885/audiobookshelf

marcelveldt:
If you rebase your branch, you get a new "on_played" callback:

Scherm­afbeelding 2025-01-15 om 01 19 42

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -0,0 +1,534 @@
...
+        Called when provider is deregistered (e.g. MA exiting or config reloading).
+        is_removed will be set to True when the provider is removed from the configuration.
+        """
+        await self._client.logout()
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        # For streaming providers return True here but for local file based providers return False.
+        return True
```
marcelveldt: are you sure about this ?
Shouldnt this be False as the content is unique ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -0,0 +1,534 @@
...
+                    provider_instance=self.instance_id,
+                )
+            },
+        )
+        mass_podcast.metadata.description = abs_podcast.media.metadata.description
+        token = self._client.token
+        url = f"{self.config.get_value(CONF_URL)}/api/items/{abs_podcast.id_}/cover?token={token}"
+        mass_podcast.metadata.images = UniqueList(
+            [MediaItemImage(type=ImageType.THUMB, path=url, provider=self.lookup_key)]
+        )
```
marcelveldt:
```suggestion
        image_url = f"{self.config.get_value(CONF_URL)}/api/items/{abs_podcast.id_}/cover?token={token}"
        mass_podcast.metadata.images = UniqueList(
            [MediaItemImage(type=ImageType.THUMB, path=image_url, provider=self.lookup_key)]
        )
```

fmunkes:
RE: on_played
Thanks a lot! The callback unfortunately doesn't report the correct position for me. I tried the following cases:
- play (no item in queue)
  - callback is called with id of new stream to be played
- replace queue
  - callback is called twice, with id of old stream and new stream 
- next / previous
  - same as above
In all cases, position = 0.

When I try to (un)mark a media item I'm faced with a RuntimeError:
`
ERROR (MainThread) [music_assistant.webserver] Error handling message: music/mark_played: Value None of type <class 'NoneType'> is invalid for media_item
`

Am I missing something here?

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: I wasn't sure, will change

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -0,0 +1,534 @@
...
+        raise MediaNotFoundError("Stream unknown")
+
+    async def _get_stream_details_audiobook(self, audiobook_id: str) -> StreamDetails:
+        """Only single audio file in audiobook."""
+        abs_audiobook = await self._client.get_audiobook(audiobook_id)
+        tracks = abs_audiobook.media.tracks
+        if len(tracks) == 0:
+            raise MediaNotFoundError("Stream not found")
+        if len(tracks) > 1:
+            logging.warning("MASS probably only supports single file base audiobooks")
```
marcelveldt:
```suggestion
            logging.warning("Music Assistant only supports single file base audiobooks")
```

For now, we only support single file audiobooks. Let me know if this is very common or many people run into this, then I can add support to the core.

fmunkes:
Comment on audiobooks split accross files:
I learned today, that abs supports sessions, where the abs instance offers you an hls stream of the audiobook. I.e. multiple files would be hidden from mass. This would change the function returning StreamDetails, and add a few other schema definitions plus code complexity. I only played using curl and api calls, nothing in python yet.
I think it is reasonable to have the current PR in a merge-ready state first. I think it does cover many use cases. When that is done, I will see if I can have that feature added as well, possible via a  config option of the provider, and would open another PR.

marcelveldt:
> Comment on audiobooks split accross files: I learned today, that abs supports sessions, where the abs instance offers you an hls stream of the audiobook. I.e. multiple files would be hidden from mass. This would change the function returning StreamDetails, and add a few other schema definitions plus code complexity. I only played using curl and api calls, nothing in python yet. I think it is reasonable to have the current PR in a merge-ready state first. I think it does cover many use cases. When that is done, I will see if I can have that feature added as well, possible via a config option of the provider, and would open another PR.

OK! Keep in mind that you can simply feed the HLS url as url in streamdetails and it will be automagically sorted.

marcelveldt:
> RE: on_played Thanks a lot! The callback unfortunately doesn't report the correct position for me. 

What player did you use to test it ?

fmunkes:
> OK! Keep in mind that you can simply feed the HLS url as url in streamdetails and it will be automagically sorted.

Ah, excellent, I just had a look at the StreamDetails class. I do have some more time on Saturday/ Sunday, so I could report back by the end of this week on this part. So we could also wait for this, and then continue from there. I guess, I'll return something like this then:
```
return StreamDetails(
            provider=self.instance_id,
            item_id=audiobook_id,
            audio_format=AudioFormat(
                content_type=ContentType.UNKNOWN,
            ),
            media_type=MediaType.AUDIOBOOK,
            stream_type=StreamType.HLS,
            path=<abs_url>/output.m3u8,
)
```
i.e. the path is an m3u8 playlist.

fmunkes:
> What player did you use to test it ?

I used a Chromecast, and tried just now via Airplay, same result. I can also try a player integrated via HomeAssistant, though that will be later today again.

marcelveldt:
> I guess, I'll return something like this then:

Exactly that should work out of the box! ffmpeg will then care care of stitching it together.

marcelveldt:
> I used a Chromecast, and tried just now via Airplay, same result. I can also try a player integrated via HomeAssistant, though that will be later today again.

Leave this with me, I will do some more testing and make the reporting more robust.
Also marking played/unplayed us currently broken in the frontend, I need to fix that.

marcelveldt:
Let's merge this one as-is now so people can start testing it in the beta cycle.
The multipart handling can then be added later in a small follow-up PR.

marcelveldt (approved):
Excellent work @fmunkes !

PR #1859, opened 2025-01-13
Opensubsonic tests for artist 'parsing'

Jc2k:
Same approach as Jellyfin - move the logic for parsing an opensubsonic type into a MA model into a standalone function so its easy and lightweight to test. Then throw syrupy at a directory full of fixtures.


marcelveldt (approved):
LGTM!

PR #1875, opened 2025-01-15
Feat: Add DSP pipeline details to stream information

maximmaxim345:
Depends on: https://github.com/music-assistant/models/pull/32
Companion Frontend PR: https://github.com/music-assistant/frontend/pull/822

# Overview

With this PR, details about the DSP are now attached to `streamdetails`, allowing us to display the DSP processing pipeline to the users.

# Implementation Details

To discern between a user-disabled DSP and a DSP disabled by Music Assistant, `DSPState` was introduced.

The DSP field of `StreamDetails` should always be up to date, since:
- Playback is restarted on changes to the `DSPConfig`
- `DSPDetails` are refreshed in case the number of players in the group changes

# DSPDetails

`DSPDetails` is essentially a `DSPConfig` with some adjustments to add dynamic information, enabling the frontend to display the DSP pipeline at any desired level of detail.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -57,7 +57,7 @@ from music_assistant.constants import (
     MASS_LOGO_ONLINE,
 )
 from music_assistant.helpers.api import api_command
-from music_assistant.helpers.audio import get_stream_details
+from music_assistant.helpers.audio import get_stream_details, get_stream_dsp_details
 from music_assistant.helpers.throttle_retry import BYPASS_THROTTLER
 from music_assistant.helpers.util import get_changed_keys
 from music_assistant.models.core_controller import CoreController
@@ -108,6 +108,7 @@ class CompareState(TypedDict):
     elapsed_time: int
     stream_title: str | None
     content_type: str | None
+    group_childs_count: int
 
 
 class PlayerQueuesController(CoreController):
@@ -962,6 +963,7 @@ class PlayerQueuesController(CoreController):
                 next_item_id=None,
                 elapsed_time=0,
                 stream_title=None,
+                group_childs_count=0,
             ),
         )
 
@@ -978,6 +980,7 @@ class PlayerQueuesController(CoreController):
             content_type=queue.current_item.streamdetails.audio_format.output_format_str
             if queue.current_item and queue.current_item.streamdetails
             else None,
+            group_childs_count=len(player.group_childs),
         )
         changed_keys = get_changed_keys(prev_state, new_state)
         # return early if nothing changed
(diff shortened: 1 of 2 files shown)


marcelveldt:
models package has been bumped to unblock this PR

music_assistant/controllers/player_queues.py
```diff
@@ -999,6 +1002,14 @@ def on_player_update(
...
             self._prev_states.pop(queue_id, None)
 
+        if "group_childs_count" in changed_keys:
+            # refresh DSP details since a player has been added/removed from the group
+            dsp = get_stream_dsp_details(self.mass, queue_id)
+            if queue.current_item and queue.current_item.streamdetails:
+                queue.current_item.streamdetails.dsp = dsp
+            if queue.next_item and queue.next_item.streamdetails:
+                queue.next_item.streamdetails.dsp = dsp
+
```
marcelveldt, line 1012: any particular reason you have this logic here instead of in the players controller ?
Or is it because dsp only impacts content sent by the queue and not any content playing from a 3rd party source ?

music_assistant/helpers/audio.py
```diff
@@ -183,6 +185,48 @@ async def strip_silence(
...
+def get_stream_dsp_details(
+    mass: MusicAssistant,
+    queue_id: str,
+) -> dict[str, DSPDetails]:
+    """Return DSP details of all players in the group, keyed by player_id."""
+    player = mass.players.get(queue_id)
+    dsp = {}
+
+    # player groups have no (explicit) leader since a player group has no audio output
+    if not player.provider.startswith("player_group"):
```
marcelveldt, line 218: the comment seems to imply the reverse of what is happening here?

maximmaxim345: I adjusted the comments to be not as confusing.

music_assistant/controllers/player_queues.py
maximmaxim345: > Or is it because dsp only impacts content sent by the queue and not any content playing from a 3rd party source ?

This was my thought.
And if  I remember correctly,  a queue is shared between multiple group members, while the `DSPDetails` dict also contains shared information.
What could be done is moving `dsp` from `streamdetails` to the `PlayerQueue` (since it's shared by all items, unlike the volume normalization).
But I kept it like this for now, to have `DSPDetails` together with all other `StreamDetails` in the frontend.
Let me know if should open a PR to move `DSPDetails`.

marcelveldt: Yeah all clear, I thought so but wanted to double check.

> Let me know if should open a PR to move DSPDetails.

No I think its fine this way as it will give us the flexibility of having a per stream decision on dsp.
Like you say, all stream-related info is now kept together. Let's keep it that way for a while and see where it goes.

marcelveldt (approved):
LGTM

PR #1876, opened 2025-01-15
chore: mypy for template player provider

Jc2k:


music_assistant/providers/_template_player_provider/__init__.py
```diff
@@ -111,7 +111,7 @@ def supported_features(self) -> set[ProviderFeature]:
         # you should return a tuple of provider-level features
```
marcelveldt:
```suggestion
        # you should return a set of provider-level features
```

marcelveldt (approved):
Thanks @Jc2k !

PR #1877, opened 2025-01-15
docker: use alpine3.21 with ffmpeg from mwader/static-ffmpeg

chatziko:
This is an experiment with improving the docker images. It does two main things:

1. Uses `python:3.12-alpine3.21` with static ffmpeg 7.1 builds from `mwader/static-ffmpeg`. These builds are created on alpine, but I think they overcome the musl issues by using glibc although they build on alpine.

2. Use a multi-stage build to decrease the image size. The python venv is build in a separate image and copied to the final one.

NOTE: I've only done some light testing of the resulting images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,64 @@
 # syntax=docker/dockerfile:1
 
-# FINAL docker image for music assistant server
-# This image is based on the base image and installs
-# the music assistant server from our built wheel on top.
-
+# Builder image. It builds the venv that will be copied to the final image
+# 
 ARG BASE_IMAGE_VERSION=latest
+FROM ghcr.io/music-assistant/base:$BASE_IMAGE_VERSION AS builder
 
-FROM ghcr.io/music-assistant/base:$BASE_IMAGE_VERSION
+# Fastest way to install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
 
-ARG MASS_VERSION
-ARG TARGETPLATFORM
+# create venv which will be copied to the final image
+ENV VIRTUAL_ENV=/app/venv
+RUN uv venv $VIRTUAL_ENV
+
 ADD dist dist
+COPY requirements_all.txt .
 
 # pre-install ALL requirements
 # comes at a cost of a slightly larger image size but is faster to start
 # because we do not have to install dependencies at runtime
-COPY requirements_all.txt .
 RUN uv pip install \
-    --no-cache \
+    --find-links "https://wheels.home-assistant.io/musllinux/" \
     -r requirements_all.txt
 
 # Install Music Assistant from prebuilt wheel
+ARG MASS_VERSION
 RUN uv pip install \
     --no-cache \
+    --find-links "https://wheels.home-assistant.io/musllinux/" \
     "music-assistant@dist/music_assistant-${MASS_VERSION}-py3-none-any.whl"
 
+# TODO: delete the unneeded architecture of librespot to decrease size!
+
+# we need to set (very permissive) permissions to the workdir
+# and /tmp to allow running the container as non-root
+# NOTE that home assistant add-ons always run as root (and use apparmor)
+# so we can't specify a user here
+#
+# IMPORTANT: chmod here, NOT on the final image, to avoid creating extra layers and increase size!
+#
+RUN chmod -R 777 /app
+
+##################################################################################################
+
+# FINAL docker image for music assistant server
+
+FROM ghcr.io/music-assistant/base:$BASE_IMAGE_VERSION
+
+ENV VIRTUAL_ENV=/app/venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# copy the already build /app dir
+COPY --from=builder /app /app
+
+# the /app contents have correct permissinos but for some reason /app itself does not.
+# so apply again, but ONLY to the dir (otherwise we increase the size)
+RUN chmod 777 /app
+
 # Set some labels
+ARG MASS_VERSION
+ARG TARGETPLATFORM
 LABEL \
     org.opencontainers.image.title="Music Assistant Server" \
     org.opencontainers.image.description="Music Assistant Server/Core" \
@@ -40,9 +73,9 @@ LABEL \
     io.hass.platform="${TARGETPLATFORM}" \
     io.hass.type="addon"
 
-RUN rm -rf dist
-
 VOLUME [ "/data" ]
 EXPOSE 8095
+
+WORKDIR $VIRTUAL_ENV
 
 ENTRYPOINT ["mass", "--config", "/data"]
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,64 +1,38 @@
 # syntax=docker/dockerfile:1
 
 # BASE docker image for music assistant container
-# Based on Debian Trixie (testing) because we need a newer version of ffmpeg (and snapcast)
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -1,31 +1,64 @@
...
     -r requirements_all.txt
 
 # Install Music Assistant from prebuilt wheel
+ARG MASS_VERSION
 RUN uv pip install \
     --no-cache \
+    --find-links "https://wheels.home-assistant.io/musllinux/" \
     "music-assistant@dist/music_assistant-${MASS_VERSION}-py3-none-any.whl"
 
+# TODO: delete the unneeded architecture of librespot to decrease size!
```
marcelveldt, line 32: what do you mean by this ?

Dockerfile.base
```diff
@@ -1,64 +1,38 @@
...
-        curl \
-        git \
-        wget \
+        jemalloc \
         tzdata \
-        python3 \
-        python3-venv \
-        python3-pip \
-        libsox-fmt-all \
-        libsox3 \
```
marcelveldt, line 23: libsox should be needed for ffmpeg or is it statically linked ?

Dockerfile.base
```diff
@@ -1,64 +1,38 @@
...
+FROM python:3.12-alpine3.21
 
 RUN set -x \
-    && apt-get update \
-    && apt-get install -y --no-install-recommends \
+    && apk add --no-cache \
         ca-certificates \
-        curl \
-        git \
-        wget \
```
marcelveldt, line 17: curl, wget, git are needed in the final image

Dockerfile.base
```diff
@@ -1,64 +1,38 @@
...
-        libjemalloc2 \
-        snapserver \
-    # cleanup
-    && rm -rf /tmp/* \
-    && rm -rf /var/lib/apt/lists/*
+        snapcast \
+        libnfs \
+        openssl-dev  # needed for airplay
 
+# Get static ffmpeg builds from https://hub.docker.com/r/mwader/static-ffmpeg/
```
marcelveldt, line 22: How good are these builds ? I mean, the core alpine team struggles with getting ffmpeg 7.1 to cooperate with musl so what makes this work magically ?

Dockerfile
chatziko: In `music_assistant/server/providers/spotify/bin` there are librespot binaries for different architectures, when building eg an amd64 image the arm ones can be deleted. With the multi-stage build, if we delete them from the `builder` image we'll save some space. Not a huge deal, that's why I left as TODO.

Dockerfile.base
```diff
@@ -1,64 +1,38 @@
...
-# TODO: Switch back to regular python stable debian image + manually build ffmpeg and snapcast ?
-FROM debian:trixie-slim
-
-ARG TARGETPLATFORM
+#
+# It should contain ONLY packages we want to deploy in the final image, not
+# packages needed to build. Build dependencies should be installed in the
+# "builder" image in the main Dockerfile.
 
+FROM python:3.12-alpine3.21
```
marcelveldt, line 9: I'd prefer switching to python 3.13 (but that can be step 2)

I was wondering in general if we should stick with alpine or bet on the safe side with debian.
Also, do you think those static ffmpeg builds work on debian maybe ?

Dockerfile.base
chatziko: If they were only meant for ffmpeg I guess they can be removed.

Dockerfile
marcelveldt: ah yeah that is what you mean, well those are max a few MB so maybe not that big of a deal

Dockerfile.base
marcelveldt: yes, we need libsox in ffmpeg but if we switch to a statically linked version, it should be included.

Dockerfile.base
chatziko: I had exactly the same question, I think the answer (judging from their [Dockerfile](https://github.com/wader/static-ffmpeg/blob/master/Dockerfile)) is that they actually use glibc (or at least partially) instead of musl, but I am not sure.

Now the quality of the builds is hard to assess, but the image is reasonably popular (500K pulls) which is a good sign. I'd say better than building ourselves in any case.

I guess what's important is to test whether the issues from music-assistant/support/issues/3257 are still present with these builds or not.

Dockerfile.base
chatziko: Yes the builds certainly work everywhere, this is the point of the static link. I've used them with debian in the past.

I don't know about alpine-vs-debian, I used alpine just to stay closer to the current stable image, since there were reported issues.

Size-wise there is a difference certainly, the current beta (debian) is 1.12G, for certain users it might be annoying (eg raspeberry pis with slow sdcards). But with the static ffmpeg we'll need way less debian packages, and together with the multi-stage savings the size might be much smaller. I could make a debian-based PR if you prefer sticking with it.

Dockerfile.base
chatziko: Oh, what for? (just curious)

marcelveldt: > Oh, what for? (just curious)

ffmpeg uses curl (or wget, I cant remember which) for streaming from http, at least that is what I remember.
git is used to install dependencies from a git source, something we still allow (for now)

Dockerfile.base
marcelveldt: I have a slight preference for alpine, since that allows us to leverage prebuilt binaries from the HA project but at the same time alpine has bit us in the back a couple of times.

Let's create a dev build from a branch so we can see for ourselves

marcelveldt:
merging into branch "alpine-ffmpeg" so we can create a testbuild and do some fine tweaking

marcelveldt:
thinking about this I realized we can make the image even much smaller if we do not install any of the extra requirements by default, just the core package. If we then have the venv folder as volume, it will persist restarts. So only first install takes a bit of time to install the package but any subsequent runs are fast.

Only problem this brings back is that you can no longer run the container as non root so maybe not worth the trouble

Dockerfile.base
marcelveldt: hmm, maybe it does no longer use curl or wget nowadays, we'll have to test

marcelveldt:
I think we can skip the builder entirely from the image or at least move it to the base image.
The base image should be the foundation and the normal image should only pick base image + MASS, nothing more.

The dev add-on is also based on only the base image:
https://github.com/music-assistant/home-assistant-addon/tree/main/music_assistant_dev

So what I was thinking is to simply test creating the VENV as volume and not pre-install any packages, except mass itself from the prebuilt wheel. No need for a builder layer unless we want to build something.

marcelveldt:
played a bit with the idea of having a dedicated (persistent) workdir/folder to store the additional packages but that was impossible to get right so the current solution of installing all dependencies into the image is still the best solution.

Anyways, made a few tweaks and now testing a dev image from this branch.
So far looks good, everything seems to work fine!

Dockerfile.base
wader: Hey lurking around a bit 👋 the builds only uses musl (maybe you confused glib and glibc?). The build do some patching (lib build fixes that i usually also upstream) and tweaking (static PIE, stack size etc) to work better with musl and behave more similar to glibc. I've use a similar build to this at work where is has decoded a lot of random media from the internet for years and have had very few problems.

If you want an even smaller ELF-binary i would suggest building ffmpeg yourself with less dependencies and features enabled.

PR #1879, opened 2025-01-16
Fix: DSP edge cases around groups

maximmaxim345:
This PR fixes various edge cases I missed with #1839.

# Reload Detection

The existing logic for issuing DSP reloads when grouping or ungrouping was flawed.
This reworks this section, handling all edge cases I can think of including:
- Don't restart playback if the leader already had DSP disabled
- Don't restart playback if a group shrank to a single device that has DSP disabled
- Proper handling of `PlayerGroup`s
-  `PlayerGroup` with only a single member

Additionally, the existing logic had another oversight where it would issue a reload command for every device in the group. This fixes that by only reloading the leader/`PlayerGroup`.

# `PlayerGroup`s with single members

Another previously missed edge case involves groups with single members that would mistakenly use the DSP Config of the `PlayerGroup` instead of the `Player`. This is now fixed.

This now allows all `PlayerGroup`s with single devices to correctly apply DSP.

### DIFF
--- a/music_assistant/controllers/players.py
+++ b/music_assistant/controllers/players.py
@@ -886,7 +886,7 @@ class PlayerController(CoreController):
         self._prev_states.pop(player_id, None)
         self.mass.signal_event(EventType.PLAYER_REMOVED, player_id)
 
-    def update(
+    def update(  # noqa: PLR0915
         self, player_id: str, skip_forward: bool = False, force_update: bool = False
     ) -> None:
         """Update player state."""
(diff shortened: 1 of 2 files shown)


marcelveldt:
The linter found something - maybe add an exclusion to that function for now, it complains about too many statements.
Something for later times is trying to split all that up into smaller parts but not a prio for now.

marcelveldt (approved):

PR #1883, opened 2025-01-17
FIX: Audiobookshelf - use non admin user endpoint + wrong login credentials

### LINKED ISSUE #3471: Audiobookshelf config error

2.4.0b16

n/a

- [x] Yes

### The problem

I am seeing two problems with configuration of Audiobookshelf (abs)

Problem 1:
The integration is accessing an ABS endpoint that requires admin access ([code](https://github.com/music-assistant/server/commit/fe63706bb60570c71b5f52a445bf5dc3f20ba399#diff-8baae458792e09c9839839cd39f5e1f9f09bf207dcdf3583e9c1009a9808c52dR128), [docs](https://api.audiobookshelf.org/#get-a-user)). When the user does not have admin access then configuration fails.

Image

I believe the integration could switch to a similar endpoint specific to the user logged in ([docs](https://api.audiobookshelf.org/#get-your-user)) that doesn't require admin access.

Problem 2:
After encountering the error the app is in bad state. I do not see an instance of abs listed in my music providers, which means I cannot edit the configuration or remove it. However I cannot add new instance as the app thinks it has one already, so I am stuck.

Image

Similar behaviour was highlighted in an issue for Sirius XM (https://github.com/music-assistant/support/issues/3127).

Exploring the api, I can see the attempted set for abs listed under music providers:

```
mass = MusicAssistantClient(mass_url, session)
print(mass.providers)

-> [..., ProviderInstance(type=<ProviderType.MUSIC: 'music'>, domain='audiobookshelf', name='Audiobookshelf', instance_id='audiobookshelf--nJqonrWt', lookup_key='audiobookshelf', supported_features={<ProviderFeature.BROWSE: 'browse'>, <ProviderFeature.LIBRARY_PODCASTS: 'library_podcasts'>, <ProviderFeature.LIBRARY_AUDIOBOOKS: 'library_audiobooks'>}, available=False, icon=None, is_streaming_provider=False)]
```

However I get errors when using the instance_id to inspect further.

```
abs = mass.providers[-1]
await mass.config.get_provider_config(abs.instance_id)

-> music_assistant_models.errors.MusicAssistantError: 'No config found for provider id audiobookshelf--nJqonrWt'
```

I was able to resolve this by downgrading to 2.3.6 then going back to 2.4.0b16

### How to reproduce

- run audiobookshelf (abs)
- add user to abs (non-admin)
- run music-assistant 2.4.0b16
- attempt to add abs provider with user created above

### Music Providers

Has the problem:
- audiobookshelf

Doesn't have the problem
- qobuz
- local filesystem

### Player Providers

n/a

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/18446595/music-assistant.log)

### Additional information

_No response_

n/a

Home Assistant OS

Alternative

fmunkes:
Fixes [#3471](https://github.com/music-assistant/support/issues/3471)

I just tested everything with non-admin users. The endpoint used appeared to be the only one where an admin user is strictly required. Will make sure in following PRs to also use a non-admin user for testing.

### DIFF
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -18,7 +18,7 @@ from music_assistant_models.enums import (
     ProviderFeature,
     StreamType,
 )
-from music_assistant_models.errors import MediaNotFoundError
+from music_assistant_models.errors import LoginFailed, MediaNotFoundError
 from music_assistant_models.media_items import (
     Audiobook,
     AudioFormat,
@@ -125,13 +125,18 @@ class Audiobookshelf(MusicProvider):
     async def handle_async_init(self) -> None:
         """Pass config values to client and initialize."""
         self._client = ABSClient()
-        await self._client.init(
-            session=self.mass.http_session,
-            base_url=str(self.config.get_value(CONF_URL)),
-            username=str(self.config.get_value(CONF_USERNAME)),
-            password=str(self.config.get_value(CONF_PASSWORD)),
-            check_ssl=bool(self.config.get_value(CONF_VERIFY_SSL)),
-        )
+        base_url = str(self.config.get_value(CONF_URL))
+        try:
+            await self._client.init(
+                session=self.mass.http_session,
+                base_url=base_url,
+                username=str(self.config.get_value(CONF_USERNAME)),
+                password=str(self.config.get_value(CONF_PASSWORD)),
+                check_ssl=bool(self.config.get_value(CONF_VERIFY_SSL)),
+            )
+        except RuntimeError:
+            # login details were not correct
+            raise LoginFailed(f"Login to abs instance at {base_url} failed.")
         await self._client.sync()
 
     async def unload(self, is_removed: bool = False) -> None:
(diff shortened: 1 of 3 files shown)


fmunkes:
Ok, so when a user uses credentials which are not ok, this is correctly reported. However, I nevertheless have
`
Provider audiobookshelf already loaded and only one instance allowed.
`
when I then correct them - after restarting Music Assistant everything is fine again.

-> this is somehow resolved by allowing multiple instances, which in the case of abs is fine. However, I would expect, that a failed setup would not 'count' as an instance. I will not add another commit besides [5515726](https://github.com/music-assistant/server/pull/1883/commits/5515726ecc788569eac7c1aa0cecedcc736544a0) to this PR, just let me know, if you think this is fine or not.

Jc2k:
While maybe niche I do think it makes sense for ABS to be multi instance.

marcelveldt:
I'll look into the single instance issue, that appears to be a bug in the config controller.
Nevertheless indeed makes sense to mark it as multi instance supported.

marcelveldt (approved):
LGTM

PR #1887, opened 2025-01-19
Feat: Ability to use Audiobookshelf Sessions/ HLS Stream

fmunkes:
Updated 01/19 14:20 documentation part

This PR introduces the ability to use an HLS stream for audiobooks ~~both audiobooks and podcasts~~. This allows to have an audiobook to be split among multiple files, as this is effectively hidden from Music Assistant.

# Details
- ABS:
    - We can acquire a [Session](https://api.audiobookshelf.org/#play-a-library-item-or-podcast-episode) for podcasts or audiobooks in ABS, as long as we provide a unique device id in our [Device Info Parameters](https://api.audiobookshelf.org/#device-info-parameters)
    - A transcoding session (that's what we use here) launches an ffmpeg process at ABS providing an HLS stream.
    - ffmpeg process only terminates, when we close the session.
- Implementation:
    - I used the device id approach of the jellyfin provider, and then added the id of the media item the session is for. This allows a session per media item.
    - ~~I added two config options - HLS for podcast, HLS for audiobooks~~
        - ~~For audiobooks default is true.~~
        - ~~For podcasts default is false.~~
    - If a session is need for audiobooks is determined automatically.
    - Schema has been updated for sessions.
    - Logging: I added a couple of debug logging information.
- Advice needed/ open questions:
    - I currently just collect the ids of open sessions in a `list[str]`.
    - I close all open sessions when we shutdown Music Assistant. I figured that's the only safe place to do so.
    - Reasoning:
        - If I were to close the session once on_played is called, a potentially running parallel stream of the same media item to another speaker would terminate as well.
        - Even if that's not the case, I think mass caches the stream details, i.e. `get_stream_details` would not be called again when the item is replayed.
    - Is the approach in place ok? Is there a better place to close open sessions?
    - Shall I limit the amount of sessions possible, and then maybe raise an error once we hit that limit? I could add another config option for the amount of possible open sessions.
- Possible improvements:
    - We could potentially update the time spent in that session, which would update the ABS statistics view. For me that's very low priority as I don't see an easy way of accessing the time spent (i.e. including possible seeking by user etc, not the media progress!) on that stream.
- Remarks:
    - on_played works nicely now, progress reporting to ABS seems functional.
      Thanks for that!

# Documentation update:
If that PR is merged, then:
## Known Issues/ Notes:
- Restarting Audiobookshelf terminates all open sessions. Music Assistant has no way to know this, so you must reload the provider. Otherwise the stream of your audiobook will be not available.

### DIFF
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -5,6 +5,7 @@ Audiobookshelf is abbreviated ABS here.
 
 from __future__ import annotations
 
+import hashlib
 from collections.abc import AsyncGenerator, Sequence
 from typing import TYPE_CHECKING
 
@@ -127,11 +128,12 @@ class Audiobookshelf(MusicProvider):
         """Pass config values to client and initialize."""
         self._client = ABSClient()
         base_url = str(self.config.get_value(CONF_URL))
+        username = str(self.config.get_value(CONF_USERNAME))
         try:
             await self._client.init(
                 session=self.mass.http_session,
                 base_url=base_url,
-                username=str(self.config.get_value(CONF_USERNAME)),
+                username=username,
                 password=str(self.config.get_value(CONF_PASSWORD)),
                 check_ssl=bool(self.config.get_value(CONF_VERIFY_SSL)),
             )
@@ -140,7 +142,21 @@ class Audiobookshelf(MusicProvider):
             raise LoginFailed(f"Login to abs instance at {base_url} failed.")
         await self._client.sync()
 
-        self.device_info: ABSDeviceInfo
+        # this is stolen from the Jellyfin provider, and gives an id surviving
+        # reboots/ provider removal+adds
+        device_id = hashlib.sha256(f"{self.mass.server_id}+{username}".encode()).hexdigest()
+
+        # this will be provided when creating sessions
+        self.device_info = ABSDeviceInfo(
+            device_id=device_id,
+            client_name="Music Assistant",
+            client_version=self.mass.version,
+            manufacturer="",
+            model=self.mass.server_id,
+        )
+
+        # cleanup old listening sessions
+        await self._client.close_all_playback_sessions_device(device_id=device_id)
 
     async def unload(self, is_removed: bool = False) -> None:
         """
@@ -375,6 +391,10 @@ class Audiobookshelf(MusicProvider):
         item_id = ""
         if session.media_type == "podcast":
             media_type = MediaType.PODCAST_EPISODE
+            podcast_id = session.library_item_id
+            session_id = session.id_
+            episode_id = session.episode_id
+            item_id = f"{podcast_id} {episode_id} {session_id}"
         else:
             media_type = MediaType.AUDIOBOOK
             audiobook_id = session.library_item_id
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -55,6 +55,8 @@ CONF_URL = "url"
 CONF_USERNAME = "username"
 CONF_PASSWORD = "password"
 CONF_VERIFY_SSL = "verify_ssl"
+CONF_USE_SESSION_AUDIOBOOK = "use_session_audiobook"
+CONF_USE_SESSION_PODCAST = "use_session_podcast"
 
 
 async def setup(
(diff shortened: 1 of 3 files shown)


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -98,6 +102,24 @@ async def get_config_entries(
             required=False,
             description="The password to authenticate to the remote server.",
         ),
+        ConfigEntry(
```
marcelveldt: Why make this configurable ?
Cant we just always use this mode when we detect multi-file audiobooks or podcast episodes ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -126,26 +148,44 @@ async def handle_async_init(self) -> None:
...
                 check_ssl=bool(self.config.get_value(CONF_VERIFY_SSL)),
             )
         except RuntimeError:
             # login details were not correct
             raise LoginFailed(f"Login to abs instance at {base_url} failed.")
         await self._client.sync()
 
+        # this is stolen from the Jellyfin provider, and gives an id surviving
+        # reboots/ provider removal+adds
+        self.device_id = hashlib.sha256(f"{self.mass.server_id}+{username}".encode()).hexdigest()
```
marcelveldt: why not use the instance id of your provider ?
so simply do self.instance_id

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -358,24 +398,82 @@ async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
...
+        )
+
     async def get_stream_details(
         self, item_id: str, media_type: MediaType = MediaType.TRACK
     ) -> StreamDetails:
         """Get stream of item."""
+        # self.logger.debug(f"Streamdetails: {item_id}")
         if media_type == MediaType.PODCAST_EPISODE:
-            return await self._get_stream_details_podcast_episode(item_id)
+            if not bool(self.config.get_value(CONF_USE_SESSION_PODCAST)):
```
marcelveldt: As said above, I would just use the playback session whenever we detect multi-file

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -422,11 +521,21 @@ async def _get_stream_details_podcast_episode(self, podcast_id: str) -> StreamDe
...
         self, media_type: MediaType, item_id: str, fully_played: bool, position: int
     ) -> None:
-        """Update progress in Audiobookshelf."""
+        """Update progress in Audiobookshelf.
+
+        In our case media_type may have 3 values:
+            - PODCAST
+            - PODCAST_EPISODE
+            - AUDIOBOOK
+        We ignore PODCAST (function is called on adding a podcast with position=None)
```
marcelveldt, line 493: Yes, the backend function just forwards all the internal playlog entries but it indeed doesnt make sense to act on the "parent type" - just only act on audiobook and podcast_episode indeed.

music_assistant/providers/audiobookshelf/abs_schema.py
```diff
@@ -314,3 +320,117 @@ class ABSLibrariesItemsResponse(BaseModel):
...
+@dataclass
+class ABSDeviceInfo(BaseModel):
+    """ABSDeviceInfo.
+
+    https://api.audiobookshelf.org/#device-info-parameters
+    https://api.audiobookshelf.org/#device-info
+    https://github.com/advplyr/audiobookshelf/blob/master/server/objects/DeviceInfo.js#L3
+    """
+
+    device_id: Annotated[str, Alias("deviceId")] = ""
```
marcelveldt, line 335: ha cool, you are using these new helpers from mushumaro to translate from/to json.
Nice!

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: You are completely right, just didn't think of it. Will change that.

fmunkes: Just noticed, that a podcast episode per ABS spec may only have a single track anyways, so I just removed that entirely, and have the session only for audiobooks now.

marcelveldt (approved):
Looks good, nice work!

OzGav:
@fmunkes I have added the note above but does this now mean the existing note about having to merge multi file audiobooks can be removed?

fmunkes:
> @fmunkes I have added the note above but does this now mean the existing note about having to merge multi file audiobooks can be removed?

Yes, that can be remove entirely, thanks!

PR #1888, opened 2025-01-19
Implemented paging in browse feature of music providers

robsonke:
- frontend pr is coming up
- let me know if this is an oke way to limit the results, especially in the specific providers overriding the browse method
- I've tested the radiobrowser option, I dont have local files or an audiobookshelf option


marcelveldt:
As discussed in discord, this is not going to work due to the fact you will completely loose sorting this way.
If you want to return multi-thousand items listings in browse, the only way is to split them up in sublevels.

PR #1890, opened 2025-01-20
fix: add missing AudioFormat metadata to Jellyfin provider

Jc2k:
This fixes 2 problems:

* There were 2 code paths that made an AudioFormat, and they produced different output (now unified into a single `audio_format(track)` helper)
* Some metadata was missing so some tracks would show incorrect metadata.


marcelveldt (approved):
LGTM

PR #1893, opened 2025-01-20
chore: ruff 0.9.1

Jc2k:
Updates ruff and runs ruff check / ruff format.

I have added A005 to the allow list:

```
music_assistant/helpers/datetime.py:1:1: A005 Module `datetime` shadows a Python standard-library module
music_assistant/helpers/json.py:1:1: A005 Module `json` shadows a Python standard-library module
music_assistant/helpers/logging.py:1:1: A005 Module `logging` shadows a Python standard-library module
```

 I lack the imagination to come up with names that I don't consider worse than shadowing stdlib.


marcelveldt (approved):
Nice, thanks @Jc2k !

PR #1900, opened 2025-01-21
Fix: Provide a proper track number for Deezer tracks if available

awhiemstra:
Use the track_position attribute if it's available as the track number. This ensures we get the right track number for things like multi-disc albums. Without this, tracks seem to follow the order in which they have been added, which does not match the actual track numbering.


marcelveldt (approved):
Thanks @awhiemstra !

PR #1901, opened 2025-01-21
fix: small refactor of tidal auth function

jozefKruszynski:


music_assistant/providers/tidal/__init__.py
```diff
@@ -131,18 +138,15 @@ async def setup(
 
 async def tidal_auth_url(auth_helper: AuthenticationHelper, quality: str) -> str:
     """Generate the Tidal authentication URL."""
+    config = TidalConfig(quality=quality, item_limit=10000, alac=False)
+    session = TidalSession(config=config)
+    url = session.pkce_login_url()
```
marcelveldt: Are you 100% sure there is no hidden I/O in here ?

jozefKruszynski: Hmmm, the wrapper function that I had previously was used to ensure that didn't cause issue, but then I started getting errors about the auth_helper.send_url not being allowed to be called in an async context without being run in an event loop.
I may have been a bit hasty so let's keep this open until I can test it further as I am definitely NOT 100% sure that there is no hidden I/O there.

jozefKruszynski: Latest version should be I/O safe

music_assistant/providers/tidal/__init__.py
```diff
@@ -131,16 +138,18 @@ async def setup(
...
         session = TidalSession(config=config)
         url = session.pkce_login_url()
-        auth_helper.send_url(url)
+        # Schedule auth_helper.send_url to run in event loop
+        auth_helper.mass.loop.call_soon_threadsafe(auth_helper.send_url, url)
         session_bytes = pickle.dumps(session)
         base64_bytes = base64.b64encode(session_bytes)
-        return base64_bytes.decode("utf-8")
+        future.set_result(base64_bytes.decode("utf-8"))
+        return future.result()
```
marcelveldt: Why the future ? You're running this entire function in a thread, you can just return the result

jozefKruszynski: I'm horrible at cleaning up my code after experiments. At least I finally learnt about `fixup` and `autosquash` with interactive rebase.

marcelveldt (approved):

PR #1903, opened 2025-01-22
No scrobble

khers:
Two unrelated commits here:
1. Remove scrobbling from the music provider so that we don't step on the toes of folks who want to use the scrobbling plugins instead.
2.  Emit a warning when a user tries to play an MPEG-4 file as these can fail with out streaming method if the moov atom is at the back of the file.


khers:
Gah, the scrobbling commit is not correct and mypy is mad about jellyfin when I try and commit so I will recreate when I can fix it.

PR #1904, opened 2025-01-22
Remove scrobble from Subsonic provider

khers:
The first commit disables scrobbling in the music provider so we don't step on the scrobble providers. The second adds a warning when a user plays an MPEG-4 file referencing our documentation on the possible problem.

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -764,7 +764,15 @@ class OpenSonicProvider(MusicProvider):
             raise UnsupportedFeaturedException(msg)
 
         mime_type = item.content_type
-        if mime_type.endswith("mpeg"):
+        # For mp4 or m4a files, better to let ffmpeg detect the codec in use so mark them unknown
+        if mime_type.endswith("mp4"):
+            self.logger.warning(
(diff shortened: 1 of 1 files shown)


music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -827,7 +829,7 @@ def _streamer() -> None:
                     timeOffset=seek_position,
                     estimateContentLength=True,
                 ) as stream:
-                    for chunk in stream.iter_content(chunk_size=40960):
+                    for chunk in stream.iter_content(chunk_size=4096):
```
marcelveldt: Any particular reason to choose such a small chunk size ?
I would rather raise it then lower it. Something like 32kb or even higher.

khers: Damn, that was not supposed to be left in (I was testing to see if different chunk sizes changed seek performance, they don't AFAICT). We currently use 40Kb as a chunk size and I'll leave that alone if you agree.

marcelveldt:
```suggestion
                    for chunk in stream.iter_content(chunk_size=40960):
```

PR #1905, opened 2025-01-23
Fix: Use Deezer record_type to set the album type of Deezer albums

awhiemstra:
The type property of an album in the Deezer api contains the resource type, not the album type. The album type is called record_type instead.

Make sure to explicitly convert between types, so that we can handle unrecognised album types as "unknown". Additionally, compilations are called "compile" in the Deezer API so we also need to handle that.

### DIFF
--- a/music_assistant/providers/deezer/__init__.py
+++ b/music_assistant/providers/deezer/__init__.py
@@ -688,7 +688,7 @@ class DeezerProvider(MusicProvider):
             return track.title_short
         return track.title
 
-    def get_album_type(self, album: deezer.Album):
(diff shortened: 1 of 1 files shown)


music_assistant/providers/deezer/__init__.py
```diff
@@ -688,6 +688,23 @@ def get_short_title(self, track: deezer.Track):
             return track.title_short
         return track.title
 
+    def get_album_type(self, album: deezer.Album):
```
marcelveldt:
```suggestion
    def get_album_type(self, album: deezer.Album) -> AlbumType:
```

marcelveldt (approved):
Thanks @awhiemstra !

PR #1912, opened 2025-01-26
Fix: Incorrect DSPState with synchronized playback

maximmaxim345:
Previously, the `DSPState` (in the `DSPDetails`) was incorrect in cases with synchronized playback with players that do not support multi device dsp.


maximmaxim345:
~This PR may have a bug with universal groups~
The bug is unrelated to DSPDetails.

marcelveldt (approved):

PR #1914, opened 2025-01-26
Feat: Audiobookshelf - extended browse + caching

fmunkes:
# Browse function
## Simple browse mode
Simple browse mode has the advantage, that a single sync call can be used, thus syncing is faster (previous implementation).
```
Audiobookshelf
    Library_Name_A (Audiobooks)
        Audiobook_1
        Audiobook_2
    Library_Name_B (Podcasts)
        Podcast_1
        Podcast_2
```
## Extended browse mode
Implements browsing by Author, Series, and Collections for Audiobook Libraries.
```
Audiobookshelf
    Library_Name_A (Audiobooks)
        Audiobooks
            Audiobook_1
            Audiobook_2
        Series
            Series_1
                Audiobook_1
                Audiobook_2
            Series_2
                Audiobook_3
                Audiobook_4
        Collections
            Collection_1
                Audiobook_1
                Audiobook_2
            Collection_2
                Audiobook_3
                Audiobook_4
        Authors
            Author_1
                Series_1
                Audiobook_1
                Audiobook_2
            Author_2
                Audiobook_3
    Library_Name_B (Podcasts)
        Podcast_1
        Podcast_2
```
The library sync here takes a little longer, as multiple API calls are required. I made it an option of the provider (up for discussion, I personally would always use extended browse, however, keeping the simple version around doesn't increase future maintenance really for me....). Browsing itself is always fast, as I only walk my data structure without any API calls.

## Caching
Simple data structure is filled with abs ids on sync, and cached using mass' cache functions. Thus browsing works also after restart without triggering a manual resync/ no API calls required.

# Documentation
If that PR were to be merged:
- not yet supported: only playlists
- If the user chooses to change browse mode of an already configured abs provider, the user should trigger a manual resync after adjusting.

### DIFF
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -706,8 +706,37 @@ class Audiobookshelf(MusicProvider):
                     )
                 )
             return items
-        elif isinstance(target, CacheAuthor | CacheCollection | CacheSeries):
+        elif isinstance(target, CacheCollection | CacheSeries):
             for book_id in target.audiobooks:
+                mass_item = await self.mass.music.get_library_item_by_prov_id(
+                    media_type=MediaType.AUDIOBOOK,
+                    item_id=book_id,
+                    provider_instance_id_or_domain=self.lookup_key,
+                )
+                if mass_item is not None:
+                    items.append(mass_item)
+        elif isinstance(target, CacheAuthor):
+            series = target.series
+            books = target.audiobooks
+            books_in_series: list[str] = []
+            for series_id in series:
+                _series = self._client.audiobook_libraries.series.get(series_id, None)
+                if _series is None:
+                    continue
+                books_in_series.extend(_series.audiobooks)
+                series_name = _series.name
+                _path_prefix, _path = full_path.split("://", 1)
+                _path = _path_prefix + "://" + "".join(_path.split("/")[:-2])  # remove authors/id
+                items.append(
+                    BrowseFolder(
+                        item_id=series_id,
+                        name=f"{series_name} (Series)",
+                        provider=self.lookup_key,
+                        path=f"{_path}/{BrowseExtendedKeys.SERIES.value.lower()}/{series_id}",
+                    )
+                )
+            books_without_series = set(books).difference(books_in_series)
+            for book_id in books_without_series:
                 mass_item = await self.mass.music.get_library_item_by_prov_id(
                     media_type=MediaType.AUDIOBOOK,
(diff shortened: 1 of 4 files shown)


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -103,6 +113,15 @@ async def get_config_entries(
...
             description="The password to authenticate to the remote server.",
         ),
+        ConfigEntry(
+            key=CONF_EXTENDED_BROWSE,
+            type=ConfigEntryType.BOOLEAN,
+            label="Extended browse",
+            required=False,
+            description="Browse by author, collection and series in the browse function"
+            " at the cost of a little longer sync times.",
+            default_value=False,
```
marcelveldt, line 123: I don't understand what browsing has to do with the sync times. These can be independent of eachother. 
Why not just do the standard stuff for the library sync and implement your own browse logic ? Imo this doe snot have to be configurable

music_assistant/providers/audiobookshelf/abs_client.py
```diff
@@ -92,6 +103,29 @@ async def init(
...
+        if self.mass is not None:
+            # try obtaining cached library items
+            data = await self.mass.cache.get(
+                key=CACHE_KEY_PODCAST_LIBRARIES,
+                base_key=self.cache_base_key,
+                category=CACHE_CATEGORY_ABS_CLIENT,
+                default=None,
+            )
+            if data is not None and isinstance(data, dict):
+                self.podcast_libraries = CachePodcastLibraries.from_dict(data)
```
marcelveldt, line 118: Not a fan of loading entire collections of data in memory with no real use. Whats in this CachePodcastLibraries ?

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Well, I wanted to have the information series, authors, collections readily available, so I put that into sync. As for the other comment, the only thing I store is the UUID of e.g. the collection, and the name. See file abs_cache_helpers Maybe my naming is not that great.

So I have
CacheLibrary
  Collections
     Collection A
       id: uuid
       name: <name>
       items: list[uuid s]
  Series
and so on. Alternatively I can put these API calls in the browse function, though I don't know how fast browsing then will be, as I have to return a list, and don't yield these items.

fmunkes:
After discussing on discord: Closing this in favor of moving the client into its own library with proper testing. When that is done and the client used within Music Assistant, I'll add the browsing feature.

PR #1920, opened 2025-01-30
feat: attach output format to the player

maximmaxim345:
Depends on https://github.com/music-assistant/models/pull/42

Attach the output format of all players in the group to `DSPDetails`.


music_assistant/helpers/audio.py
```diff
@@ -219,15 +221,26 @@ def get_stream_dsp_details(
...
     player = mass.players.get(queue_id)
-    dsp = {}
+    dsp: dict[str, DSPDetails] = {}
     group_preventing_dsp = is_grouping_preventing_dsp(player)
+    output_format = None
 
-    # We skip the PlayerGroups as they don't provide an audio output
-    # by themselves, but only sync other players.
-    if not player.provider.startswith("player_group"):
+    if player.provider.startswith("player_group"):
```
marcelveldt, line 227: At some point I think DSP should become its own subcontroller underneath the StreamsController instead of putting everything in the audio helpers (and make the audio helpers nothing more than basic helpers) but that is something for later, also applies to some of the other logic.

music_assistant/helpers/audio.py
```diff
@@ -219,15 +221,26 @@ def get_stream_dsp_details(
...
+        if group_preventing_dsp:
+            with contextlib.suppress(RuntimeError):
```
marcelveldt: Why this RuntimeError ? Is that raised in the playergroup ?

marcelveldt (approved):

music_assistant/helpers/audio.py
maximmaxim345: Added comments (and a warning) explaining that.

music_assistant/helpers/audio.py
maximmaxim345: I can look into that at at some later point in time. Refactoring it into a `DSPController` could also maybe help with this: https://github.com/orgs/music-assistant/discussions/3397.

marcelveldt: There is no rush but at some point it will make the code more maintainable in the log run but that does not only apply to DSP. There are more sections that can use some cleanup and refactoring. We can mark some things we notice on the backlog.

As for not requiring a stream restart on making changes...
I think that is already possible today if you really want, because you could consider running the dsp on each pcm chunk (of 1 second). In the end you get PCM chunks (of about 1 second) from the internal machinery and that gets fed into the final "per-player" ffmpeg process. 

Anyways, food for thought, for later, not now.

marcelveldt (approved):

PR #1928, opened 2025-02-01
fix: show chapters in ABS audiobook view

fmunkes:


marcelveldt (changes requested):
This should not be needed, as discussed in discord. 
Let's not do it this way but instead fix it at core level.

fmunkes:
I close this then, the chapter times appear fine

marcelveldt:
I was mistaken. For podcasts its fine to retrieve episodes later/lazy but for audiobooks they really need to be present.
If this causes excessive more requests on your end, please discuss on discord so we can change the model a bit at core level. TIP: just cache the episodes in your provider.

marcelveldt (approved):

PR #1933, opened 2025-02-03
Define track limit for get_playlist to avoid standard limit of 100 to…

bobiboy:
Currently when calling the get_playlist method of the YT Music API, the limit parameter is passed as "None" which causes the default limit of 100 tracks to apply. In order to load larger playlists, increasing the limit in line with other occurrences in the provider where 9999 is set as the limit.

### DIFF
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -52,7 +52,7 @@ async def get_playlist(
 
     def _get_playlist():
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        playlist = ytm.get_playlist(playlistId=prov_playlist_id, limit=None)
+        playlist = ytm.get_playlist(playlistId=prov_playlist_id, limit=9999)
         playlist["checksum"] = get_playlist_checksum(playlist)
         # Fix missing playlist id in some edge cases
         playlist["id"] = prov_playlist_id if not playlist.get("id") else playlist["id"]


MarvinSchenkel:
Are you sure this is required? If I look at the inline documentation [here](https://github.com/sigma67/ytmusicapi/blob/dfbcbcda7ca700c3bb8c34d2ff8e54263815e831/ytmusicapi/mixins/playlists.py#L21) it is mentioned that it will return all the songs in a playlist when `None` is provided. My playlists of > 100 tracks also load fine :
image

bobiboy:
Hi, yes you're probably right - I'm not too experienced with Python and assumed passing "None" would be the same as not passing it at all - but that doesn't seem logical based on how the method is set up. Something I didn't realize during testing is that the interface gives contradicting information - the header says there are 150 tracks (probably what I looked at while testing the change) while the actual list below is still only showing 100 tracks

image

marcelveldt:
That text "150 tracks" comes from the description - its basically freetext

marcelveldt:
So this PR can be closed ? Its already working as it should ?

MarvinSchenkel:
> Hi, yes you're probably right - I'm not too experienced with Python and assumed passing "None" would be the same as not passing it at all - but that doesn't seem logical based on how the method is set up. Something I didn't realize during testing is that the interface gives contradicting information - the header says there are 150 tracks (probably what I looked at while testing the change) while the actual list below is still only showing 100 tracks
> 
> image

Can you double check how many tracks this playlist has inside the YT Music web app?

bobiboy:
> > Hi, yes you're probably right - I'm not too experienced with Python and assumed passing "None" would be the same as not passing it at all - but that doesn't seem logical based on how the method is set up. Something I didn't realize during testing is that the interface gives contradicting information - the header says there are 150 tracks (probably what I looked at while testing the change) while the actual list below is still only showing 100 tracks
> > 
> > image
> 
> Can you double check how many tracks this playlist has inside the YT Music web app?

150 Tracks. I thought it might have been a matter of Auto playlists behaving differently (Super mix happens to have 100 tracks in the web app, but I was looking at My Liked music which has way more but was still only showing 100 tracks) so to confirm I created a new list with 150 tracks.

Overall I guess we can close the PR and create an issue instead for further investigation. Apologies for rushing it with the PR!

MarvinSchenkel:
> > > Hi, yes you're probably right - I'm not too experienced with Python and assumed passing "None" would be the same as not passing it at all - but that doesn't seem logical based on how the method is set up. Something I didn't realize during testing is that the interface gives contradicting information - the header says there are 150 tracks (probably what I looked at while testing the change) while the actual list below is still only showing 100 tracks
> > > image
> > 
> > 
> > Can you double check how many tracks this playlist has inside the YT Music web app?
> 
> 150 Tracks. I thought it might have been a matter of Auto playlists behaving differently (Super mix happens to have 100 tracks in the web app, but I was looking at My Liked music which has way more but was still only showing 100 tracks) so to confirm I created a new list with 150 tracks.
> 
> Overall I guess we can close the PR and create an issue instead for further investigation. Apologies for rushing it with the PR!

So then you confirm you can sync playlist with more than 100 tracks into MA? Then we're good to close this PR, otherwise we'll need to do some more digging :)

bobiboy:
> > > > Hi, yes you're probably right - I'm not too experienced with Python and assumed passing "None" would be the same as not passing it at all - but that doesn't seem logical based on how the method is set up. Something I didn't realize during testing is that the interface gives contradicting information - the header says there are 150 tracks (probably what I looked at while testing the change) while the actual list below is still only showing 100 tracks
> > > > image
> > > 
> > > 
> > > Can you double check how many tracks this playlist has inside the YT Music web app?
> > 
> > 
> > 150 Tracks. I thought it might have been a matter of Auto playlists behaving differently (Super mix happens to have 100 tracks in the web app, but I was looking at My Liked music which has way more but was still only showing 100 tracks) so to confirm I created a new list with 150 tracks.
> > Overall I guess we can close the PR and create an issue instead for further investigation. Apologies for rushing it with the PR!
> 
> So then you confirm you can sync playlist with more than 100 tracks into MA? Then we're good to close this PR, otherwise we'll need to do some more digging :)

Not sure what happened, but suddenly getting all 150 tracks in the list.. Apologies for the confusion, closed the PR!

PR #1947, opened 2025-02-07
Fix relative path not working with more than one level

### LINKED ISSUE #2291: Playlist: relative Path not working when going up two levels

2.0.3

_No response_

- [X] Yes

### The problem

relative Path in local Playlists not working

image

### How to reproduce

Use relative Path in Playlists

### Music Providers

local disc

### Player Providers

any

### Full log output

_No response_

### Additional information

_No response_

2024.5.3

Home Assistant Core

Raspberry Pi

OzGav:
Trying to fix https://github.com/music-assistant/support/issues/2291

Did some research. I can't test it locally but I did at https://www.online-python.com/ and I believe it works.

### DIFF
--- a/music_assistant/providers/filesystem_local/helpers.py
+++ b/music_assistant/providers/filesystem_local/helpers.py
@@ -201,9 +201,8 @@ def get_relative_path(base_path: str, path: str) -> str:
 
 def get_absolute_path(base_path: str, path: str) -> str:
     """Return the absolute path string for a path."""
-    if path.startswith(base_path):
-        return path
-    return os.path.join(base_path, path)
+    full_path = os.path.join(base_path, path)
+    return os.path.abspath(full_path)
 
 
 def sorted_scandir(base_path: str, sub_path: str, sort: bool = False) -> list[FileSystemItem]:


marcelveldt:
This is not the right way to fix that issue at all as the issue is for playlists. The helper you adjusted in the PR will break the entire file provider

OzGav:
Ok I thought that helper was used by the file system provider to parse the playlist…..

OzGav:
It is this line here I saw https://github.com/music-assistant/server/blob/114d42fee9edcd9ed4743e53247a381f4e04b9b1/music_assistant/providers/filesystem_local/__init__.py#L667

OzGav:
Actually probably this is where it needs to be fixed https://github.com/music-assistant/server/blob/114d42fee9edcd9ed4743e53247a381f4e04b9b1/music_assistant/providers/filesystem_local/__init__.py#L718

OzGav:
Could you use 

```
    async def _parse_playlist_line(self, line: str, playlist_path: str) -> Track | None:
        """Try to parse a track from a playlist line."""
        try:
            # if a relative path was given in an upper level from the playlist,
            # try to resolve it
            for parentpart in ("../", "..\\"):
                while line.startswith(parentpart):
                    if len(playlist_path) < 3:
                        break  # guard
                    playlist_path = os.path.dirname(playlist_path)
                    line = line[3:]
```

due to what I read here https://stackoverflow.com/questions/918154/relative-paths-in-python and https://www.geeksforgeeks.org/python-os-path-dirname-method/

So the second last line has been changed from `playlist_path = parentpart[:-3]`

PR #1948, opened 2025-02-08
ABS: Rewrite of provider, couple new features

fmunkes:
Added with this PR:

- Separation of API calls into separate [lib](https://github.com/music-assistant/aioaudiobookshelf/)
- Separation of functions, which translate abs -> mass into `parsers.py`, such that they can be tested easier in the future.
- Enhanced browse view (by series, collection, author)
- Caching of expanded abs lib items to reduce amount of API calls
- Additionally subscription to socketio API, gives immediate update of lib items of known libraries for podcasts and audiobooks
- No hls streams anymore by using @marcelveldt implementation in the filesystem provider for multi-file audiobooks

Future ideas:
- socketio callbacks for fully new libraries at abs (not lib items, that is included)
- once ui has the progress status for audiobooks, make sure they are shown
- inclusion of playlists at abs

@OzGav I will give proper documentation details once we are done reviewing this/ the lib.


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -138,32 +206,46 @@ def supported_features(self) -> set[ProviderFeature]:
...
-            client_version=self.mass.version,
-            manufacturer="",
-            model=self.mass.server_id,
+            await self._client_socket.init_client()
+        except AbsLoginError as exc:
+            raise LoginFailed(f"Login to abs instance at {base_url} failed.") from exc
+
+        self.cache_base_key = f"{CACHE_BASE_KEY_PREFIX}_{self.lookup_key}"
+
+        cached_libraries = await self.mass.cache.get(
```
marcelveldt, line 229: Am I reading this right that you are putting a large collection into the cache and load it all back into memory again as a class variable ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -182,139 +264,140 @@ def is_streaming_provider(self) -> bool:
...
+                    )
+                    if mass_item is None:
+                        self.logger.debug("Removing % , id: %s from cache.", media_type.value, id_)
+                        await self.mass.cache.delete(
+                            key=id_,
+                            base_key=self.cache_base_key,
+                            category=CACHE_CATEGORY_AUDIOBOOKS
+                            if media_type == MediaType.AUDIOBOOK
+                            else CACHE_CATEGORY_PODCASTS,
+                        )
```
marcelveldt: why is this needed ? I'm not following

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: It's the library helper structure, not the full library:
```python
@dataclass(kw_only=True)
class LibraryHelper(DataClassDictMixin):
    """Lib name + media items' uuids."""

    name: str
    item_ids: set[str] = field(default_factory=set)


@dataclass(kw_only=True)
class LibrariesHelper(DataClassDictMixin):
    """Helper class to store ABSLibrary name, id and the uuids of its media items.

    Dictionary is lib_id:AbsLibraryWithItemIDs.
    """

    audiobooks: dict[str, LibraryHelper] = field(default_factory=dict)
    podcasts: dict[str, LibraryHelper] = field(default_factory=dict)

```

so library name, uuid of that library, uuid of all items in that library (that's actually the same in the current version, it's just not cached.)

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: I wanted to remove cached versions of removed items on the abs side. THough I think they will expire anyways. Let me remove that code.

marcelveldt: Yes, they expire anyways.

fmunkes:
@OzGav 
Here is an update of the full documentation 
# Audiobookshelf
## Features
- Populates Audiobooks from all libraries accessible by the supplied user
- Populates Podcasts from all libraries accessible by the supplied user
- Browse Feature has the following structure:
```
        Library_Name_A (Audiobooks)
            Authors
                Author_1
                    Series_1
                    Audiobook_1
                    Audiobook_2
                Author_2
                    Audiobook_3
            Narrators
                Narrator_1
                    Audiobook_1
                    Audiobook_2
            Series
                Series_1
                    Audiobook_1
                    Audiobook_2
                Series_2
                    Audiobook_3
                    Audiobook_4
            Collections
                Collection_1
                    Audiobook_1
                    Audiobook_2
                Collection_2
                    Audiobook_3
                    Audiobook_4
            Audiobooks
                Audiobook_1
                Audiobook_2
        Library_Name_B (Podcasts)
            Podcast_1
            Podcast_2
```
Keep in mind that the Audiobook search function supports searching for Authors and Narrators as well
- Progress is synced in both ways and obtained just ahead of playing
- Event driven updates of podcast/ audiobook metadata in **known** libraries:
    - A change is immediately reflected to the mass database if mass is running
    - Newly added and just deleted items are immediately reflected as well
    - BUT: if mass was down while changes in the abs database occured, those will only be synced if a normal provider sync is triggered
- Support for single and multi-file audiobooks

## Configuration
The following is needed to setup this provider:
- the URL of an Audiobookshelf instance
- the username of an Audiobookshelf user
- the password of this user
- Optionally podcasts with no episodes yet downloaded may be skipped when syncing
- Optionally, SSL verification may be disabled

The user must be of type user, admin or root. Guest users are neither tested nor supported. 

## Known Issues / Notes
- multi-file audiobooks: the ui will show PCM as source file format (that is what is used internally) instead of the actual file format of the audiobook

Tested currently against ABS >= 2.19.0.
In the first instance of any problems ensure the server is running the latest version of the audiobookshelf software

## Not Yet Supported
- Playlists (not yet supported in core) 
- Edit provider feature
- Creation/deletion of a new library (I. E. not media items in a known library) is not reflected in an event driven way yet. Instead, use a normal sync.

fmunkes:
I only had very limited time for testing today, but since I didn't really change much, PR should be fine I guess.
Small side note - If this is known, I'm sorry, didn't see it somewhere - the 30s progress reporting occurs twice every 30s for a media player integrated via hass. Group count doesn't matter. With a chromecast directly integrated in mass it is called once every 30s. For play state changes reporting is as expected. Since progress is reported correctly nonetheless, not a big deal.

marcelveldt:
@fmunkes can you have a look at the merge conflict ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -537,104 +634,439 @@ async def on_played(
...
-        for library in library_list:
-            if library_id == library.id_:
+        return sorted(items, key=lambda x: x.name)
+
+    async def _browse_narrators(
+        self, current_path: str, library_id: str
+    ) -> Sequence[MediaItemType]:
+        # filter_str = get_library_filter_string(
+        #     filter_group=AbsFilterGroup.NARRATORS, filter_value=narrator.name
+        # )
```
marcelveldt: leftover test code ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -537,104 +634,439 @@ async def on_played(
...
+                await self.mass.music.audiobooks.add_item_to_library(
+                    parse_audiobook(
+                        abs_audiobook=abs_item,
+                        lookup_key=self.lookup_key,
+                        domain=self.domain,
+                        instance_id=self.instance_id,
+                        token=self._client.token,
+                        base_url=str(self.config.get_value(CONF_URL)).rstrip("/"),
+                    ),
+                    overwrite_existing=True,
```
marcelveldt, line 961: why overwrite the existing item ?

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Exactly, will be removed

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: I don't remember anymore, could be, that the metadata wasn't updated otherwise. I'll check in a bit

fmunkes: Yep, that is the case. I tried the title of  an audiobook - without the overwrite, the update does not work. With the overwrite, I'll have the media item updated immediately.

marcelveldt: interesting... so the title can even change ?

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+                provider_instance=instance_id,
+            )
+        },
+    )
+    mass_podcast.metadata.description = abs_podcast.media.metadata.description
+    if token is not None:
+        image_url = f"{base_url}/api/items/{abs_podcast.id_}/cover?token={token}"
+        mass_podcast.metadata.images = UniqueList(
+            [MediaItemImage(type=ImageType.THUMB, path=image_url, provider=lookup_key)]
+        )
```
marcelveldt, line 68: so without a token you wont have an image ?

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+        cover_url = f"{base_url}{api_url}"
+        mass_audiobook.metadata.images = UniqueList(
+            [MediaItemImage(type=ImageType.THUMB, path=cover_url, provider=lookup_key)]
+        )
+
+    # expanded version
+    if isinstance(abs_audiobook, AbsLibraryItemExpandedBook):
+        authors = UniqueList([x.name for x in abs_audiobook.media.metadata.authors])
+        narrators: UniqueList[str] = UniqueList(abs_audiobook.media.metadata.narrators)
+        mass_audiobook.authors = authors
```
marcelveldt:
```suggestion
        mass_audiobook.authors.set([x.name for x in abs_audiobook.media.metadata.authors])
```

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+    if token is not None:
+        api_url = f"/api/items/{abs_audiobook.id_}/cover?token={token}"
+        cover_url = f"{base_url}{api_url}"
+        mass_audiobook.metadata.images = UniqueList(
+            [MediaItemImage(type=ImageType.THUMB, path=cover_url, provider=lookup_key)]
+        )
+
+    # expanded version
+    if isinstance(abs_audiobook, AbsLibraryItemExpandedBook):
+        authors = UniqueList([x.name for x in abs_audiobook.media.metadata.authors])
```
marcelveldt:
```suggestion
```

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+        narrators: UniqueList[str] = UniqueList(abs_audiobook.media.metadata.narrators)
+        mass_audiobook.authors = authors
+        mass_audiobook.narrators = narrators
```
marcelveldt:
```suggestion
        mass_audiobook.narrators.set(abs_audiobook.media.metadata.narrators)
```

music_assistant/providers/audiobookshelf/parsers.py
fmunkes: Correct, but without a token you also won't have access. An authenticated client has a token.

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+        chapters = []
+        for idx, chapter in enumerate(abs_audiobook.media.chapters):
+            chapters.append(
+                MediaItemChapter(
+                    position=idx + 1,  # chapter starting at 1
+                    name=chapter.title,
+                    start=chapter.start,
+                    end=chapter.end,
+                )
+            )
```
marcelveldt:
```suggestion
        for idx, chapter in enumerate(abs_audiobook.media.chapters, 1):
            chapters.append(
                MediaItemChapter(
                    position=idx,
                    name=chapter.title,
                    start=chapter.start,
                    end=chapter.end,
                )
            )
```

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -0,0 +1,226 @@
...
+        mass_audiobook.metadata.chapters = chapters
+
+    elif isinstance(abs_audiobook, AbsLibraryItemMinifiedBook):
+        mass_audiobook.authors = UniqueList([abs_audiobook.media.metadata.author_name])
+        mass_audiobook.narrators = UniqueList([abs_audiobook.media.metadata.narrator_name])
```
marcelveldt:
```suggestion
        mass_audiobook.authors.set([abs_audiobook.media.metadata.author_name])
        mass_audiobook.narrators.set([abs_audiobook.media.metadata.narrator_name])
```

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -59,6 +62,75 @@
 # optionally hide podcasts with no episodes
 CONF_HIDE_EMPTY_PODCASTS = "hide_empty_podcasts"
 
+CACHE_BASE_KEY_PREFIX = "audiobookshelf"
```
marcelveldt: I suggest to just use the instance_id as base key

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -59,6 +62,75 @@
...
+
+
+ABSBROWSEITEMSTOPATH: dict[str, str] = {
+    AbsBrowseItemsBook.AUTHORS: AbsBrowsePaths.AUTHORS,
+    AbsBrowseItemsBook.NARRATORS: AbsBrowsePaths.NARRATORS,
+    AbsBrowseItemsBook.SERIES: AbsBrowsePaths.SERIES,
+    AbsBrowseItemsBook.COLLECTIONS: AbsBrowsePaths.COLLECTIONS,
+    AbsBrowseItemsBook.AUDIOBOOKS: AbsBrowsePaths.AUDIOBOOKS,
+}
+
```
marcelveldt: For a follow up iteration, maybe move this info a constants file

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: done

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Noted. Will do that.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -138,32 +210,46 @@ def supported_features(self) -> set[ProviderFeature]:
...
-            device_id=self.instance_id,
-            client_name="Music Assistant",
-            client_version=self.mass.version,
-            manufacturer="",
-            model=self.mass.server_id,
+            await self._client_socket.init_client()
+        except AbsLoginError as exc:
+            raise LoginFailed(f"Login to abs instance at {base_url} failed.") from exc
+
+        self.cache_base_key = f"{CACHE_BASE_KEY_PREFIX}_{self.lookup_key}"
```
marcelveldt: why not just use the lookup key because that is already unique and contains your domain as prefix

fmunkes: Did change that.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -182,139 +267,116 @@ def is_streaming_provider(self) -> bool:
         return False
 
     async def sync_library(self, media_types: tuple[MediaType, ...]) -> None:
-        """Run library sync for this provider."""
-        await self._client.sync()
+        """Obtain audiobook library ids and podcast library ids."""
+        libraries = await self._client.get_all_libraries()
+        # we can overwrite all libs with empty ids here, as they are directly
+        # filled afterwards again.
```
marcelveldt: There is a potential bug here: when only a specific media type is synced, the helper cache gets cleared here but not filled again...

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -182,139 +267,116 @@ def is_streaming_provider(self) -> bool:
...
+        for library in libraries:
+            if library.media_type == AbsLibraryMediaType.BOOK:
+                self.libraries.audiobooks[library.id_] = LibraryHelper(name=library.name)
+            elif library.media_type == AbsLibraryMediaType.PODCAST:
+                self.libraries.podcasts[library.id_] = LibraryHelper(name=library.name)
```
marcelveldt: check if the mediatype is requested for sync

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -138,32 +209,46 @@ def supported_features(self) -> set[ProviderFeature]:
...
+            self.libraries = LibrariesHelper.from_dict(cached_libraries)
 
-        self.logger.debug(f"Our playback session device_id is {self.instance_id}")
+        # set socket callbacks
+        self._client_socket.set_item_callbacks(
+            on_item_added=self._socket_abs_item_changed,
+            on_item_updated=self._socket_abs_item_changed,
+            on_item_removed=self._socket_abs_item_removed,
+            on_items_added=self._socket_abs_item_changed,
+            on_items_updated=self._socket_abs_item_changed,
```
marcelveldt, line 246: So this callback is the same for just a progress report and if the item actually changed ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -323,171 +385,154 @@ async def get_podcast_episodes(self, prov_podcast_id: str) -> list[PodcastEpisod
...
+            async for response in self._client.get_library_items(library_id=book_lib_id):
+                if not response.results:
+                    break
+                book_ids = [x.id_ for x in response.results]
+                # store uuids
+                self.libraries.audiobooks[book_lib_id].item_ids.update(book_ids)
+                # use expanded version for chapters/ caching.
+                books_expanded = await self._client.get_library_item_batch_book(item_ids=book_ids)
+                for book_expanded in books_expanded:
+                    await self._cache_set_audiobook(book_expanded)
```
marcelveldt: I was wondering if you really need all the caching because the sync will only be once in a while and then the user looks at the library items, next to that ABS is local so these calls are rather cheap. Or did you add the caching to speed up browsing ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -323,171 +385,154 @@ async def get_podcast_episodes(self, prov_podcast_id: str) -> list[PodcastEpisod
...
+                    (AudioFormat(content_type=content_type), stream_url, track.duration)
+                )
+
+            return StreamDetails(
+                provider=self.instance_id,
+                item_id=abs_audiobook.id_,
+                # for the concatanated stream, we need to use a pcm stream format
+                audio_format=AudioFormat(
+                    # values recommended by Marcel (some are default, but
+                    # nonetheless given here.
```
marcelveldt:
```suggestion
```

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -525,6 +573,54 @@ async def _get_stream_details_podcast_episode(self, podcast_id: str) -> StreamDe
...
 
+    async def get_audio_stream(
+        self, streamdetails: StreamDetails, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """
+        Return the (custom) audio stream for the provider item.
+
+        Only used for multi-file audiobooks, copy of Marcel's code
+        in the filesystem provider, only cosmetic (iter audioformat)
+        adjustments.
```
marcelveldt:
```suggestion
        Only used for multi-file audiobooks.
```

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: except for removed. In all other cases I always get the full media item (or a list of them).

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -537,104 +633,441 @@ async def on_played(
         We ignore PODCAST (function is called on adding a podcast with position=None)
 
         """
-        # self.logger.debug(f"on_played: {media_type=} {item_id=}, {fully_played=} {position=}")
         if media_type == MediaType.PODCAST_EPISODE:
```
marcelveldt, line 616: I'm missing a guard at the top of this function which checks when you last set the resume point to prevent a pingpong when the socket connections responds that an item was updated. Or did you guard that elsewhere ?

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Uh. Thanks!

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Hmm. I guess I can remove the caching. For browsing, only my library helpers are used - and these I'd like to keep in cache. But I can remove it for the media items.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Wanted to give you proper credit :)

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: There is no guard, because the progress callback is not in there. Will be another future PR

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: I removed it for audiobooks and podcasts, only helper libraries stay, otherwise I have no way to know the library names/ ids before a full sync took place in browsing. That however is small in size.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Ah, now I read it correctly, I'm sorry. The progress report callback is not part of the current code. It is a separate callback, which I will add to the provider with another PR. These callbacks are only used for item metadata changes.

marcelveldt (approved):
Excellent work!

PR #1949, opened 2025-02-08
Feat: Use the player prefered (lossless) format in universal groups

maximmaxim345:
With this PR, universal groups now no longer default to MP3.

If the enforce mp3 option is enabled on a player in a universal group, it will still use MP3 for just that player.

### DIFF
--- a/music_assistant/providers/player_group/__init__.py
+++ b/music_assistant/providers/player_group/__init__.py
@@ -501,7 +501,7 @@ class PlayerGroupProvider(PlayerProvider):
         self.ugp_streams[player_id] = UGPStream(
             audio_source=audio_source, audio_format=UGP_FORMAT, base_pcm_format=UGP_FORMAT
         )
-        base_url = f"{self.mass.streams.base_url}/ugp/{player_id}.mp3"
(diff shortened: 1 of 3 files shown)


music_assistant/providers/player_group/__init__.py
```diff
@@ -889,10 +895,20 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.Response:
...
         child_player_id = request.query.get("player_id")  # optional!
-
-        # Right now we default to MP3 output format, since it's the most compatible
-        # TODO: use the player's preferred output format
-        output_format = AudioFormat(content_type=ContentType.MP3)
+        output_format_str = request.path.rsplit(".")[-1]
+
+        if child_player_id and (child_player := self.mass.players.get(child_player_id)):
+            # Use the preferred output format of the child player
+            output_format = await self.mass.streams._get_output_format(
```
marcelveldt: you're accessing a private function here. Better promote that function to a public one

maximmaxim345: Done

marcelveldt (approved):

PR #1952, opened 2025-02-10
Fix: Improve parsing of local files playlist lines

OzGav:


marcelveldt:
I have edited your PR as it required a bunch more logic to handle all the (edge) cases. It should now be a lot more rigid and also be able to parse URL encoded lines

OzGav:
@marcelveldt as you have done quite a bit here will it skip any unavailable items and continue on to address this https://github.com/music-assistant/support/issues/3320

marcelveldt:
Yes that should also be fixed

PR #1959, opened 2025-02-13
Fix: Double grouped players with Universal Groups

maximmaxim345:
This PR addressed the bugs I found with https://github.com/music-assistant/server/pull/1912#issuecomment-2615244536.

Previously, it was possible to add already synchronized players to a universal group.

### DIFF
--- a/music_assistant/providers/player_group/__init__.py
+++ b/music_assistant/providers/player_group/__init__.py
@@ -627,6 +627,10 @@ class PlayerGroupProvider(PlayerProvider):
             raise InvalidDataError(
                 f"Player {child_player.display_name} already has another group active"
             )
+        if len(child_player.group_childs) > 0 or child_player.synced_to:
+            raise InvalidDataError(
+                f"Player {child_player.display_name} is already part of another group"
+            )
         group_player.group_childs.append(player_id)
 
         # handle resync/resume if group player was already playing
--- a/music_assistant/providers/player_group/__init__.py
+++ b/music_assistant/providers/player_group/__init__.py
@@ -551,6 +551,7 @@ class PlayerGroupProvider(PlayerProvider):
         """
         if group_player := self.mass.players.get(player_id):
             self._update_attributes(group_player)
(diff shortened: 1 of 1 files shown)


maximmaxim345:
I have marked this as a draft because it does not completely resolve the issue on startup.

In the case where a player is automatically assigned to synchronization after being powered on:

- The code checks if it is synchronized. Since it is powered off (or just powered on but not fully initialized), it thinks its not synchronized.
- The player is added to the universal group.
- Then, the player provider adds the underlying player to the group (not the universal group, a completely different player!)
- The universal group now also plays on a device not controlled by the group.

I tested this with Snapcast (2 snapweb players)

Do you have an Idea how to resolve this?

marcelveldt:
this check is good to have anyways - we could also consider to simply unjoin it first if we run into this

maximmaxim345:
Adjusted the PR to ungroup players before adding them, and periodically check if they are still ungrouped during polling.
This seams to fix grouping issues including those on with snapcast on startup.

music_assistant/providers/player_group/__init__.py
```diff
@@ -893,6 +897,34 @@ def _update_attributes(self, player: Player) -> None:
         player.can_group_with = can_group_with
         self.mass.players.update(player.player_id)
 
+    async def _ungroup_subgroups_if_found(self, player: Player) -> None:
+        """Verify that no player is part of a separate group."""
+        group_type = self.mass.config.get_raw_player_config_value(
+            player.player_id, CONF_ENTRY_GROUP_TYPE.key, CONF_ENTRY_GROUP_TYPE.default_value
+        )
+        if group_type == GROUP_TYPE_UNIVERSAL:
```
marcelveldt: you can change this into a guard clause by returning if the group type is not universal.
it saves some indentation and makes the code easier to read

maximmaxim345: Done

marcelveldt (approved):

PR #1968, opened 2025-02-20
fix(tidal): fix track playback with lyrics error suppression

jozefKruszynski:
raise original error for lyrics not existing, so that error suppression
works as expected and playback is unhindered


marcelveldt (approved):
next step is check where alle the calls to get the lyrics are coming from

PR #1969, opened 2025-02-20
Adding `number` as valid volume player control entities

alexbarcelo:
`input_number` entities are already accepted as volume player control entities. Adding `number` is natural as they behave almost identically.

[Discussion 3560](https://github.com/orgs/music-assistant/discussions/3560)

### DIFF
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -92,7 +92,7 @@ def filename_from_string(string: str) -> str:
 def try_parse_int(possible_int: Any, default: int | None = 0) -> int | None:
     """Try to parse an int."""
     try:
-        return int(possible_int)
+        return int(float(possible_int))
     except (TypeError, ValueError):
         return default
 
--- a/music_assistant/providers/hass/__init__.py
+++ b/music_assistant/providers/hass/__init__.py
@@ -454,14 +454,15 @@ class HomeAssistantProvider(PluginProvider):
                 service_data={"volume_level": volume_level / 100},
                 target={"entity_id": entity_id},
             )
-        else:
-            # The call to `set_value` works for both `number` and `input_number` entities
-            await self.hass.call_service(
-                domain=domain,
-                service="set_value",
-                target={"entity_id": entity_id},
-                service_data={"value": volume_level},
-            )
+            return
+
+        # At this point, `set_value` will work for both `number` or `input_number`
+        await self.hass.call_service(
+            domain=domain,
+            service="set_value",
+            target={"entity_id": entity_id},
+            service_data={"value": volume_level},
+        )
 
     def _update_control_from_state_msg(self, entity_id: str, state: CompressedState) -> None:
         """Update PlayerControl from state(update) message."""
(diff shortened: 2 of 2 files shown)


music_assistant/helpers/util.py
```diff
@@ -92,7 +92,7 @@ def filename_from_string(string: str) -> str:
 def try_parse_int(possible_int: Any, default: int | None = 0) -> int | None:
     """Try to parse an int."""
     try:
-        return int(possible_int)
+        return int(float(possible_int))
```
marcelveldt, line 95: why is this redundant conversion added here ?

music_assistant/providers/hass/__init__.py
```diff
@@ -445,16 +445,19 @@ async def _handle_player_control_mute_set(self, entity_id: str, muted: bool) ->
...
+
+        if domain == "media_player":
             await self.hass.call_service(
-                domain="media_player",
+                domain=domain,
                 service="volume_set",
                 service_data={"volume_level": volume_level / 100},
                 target={"entity_id": entity_id},
             )
         else:
```
marcelveldt: let's use a guard clause here. return from the media_player block and

music_assistant/helpers/util.py
alexbarcelo: `number` returns as float (e.g. "31.0") so it was not working. I don't know if it is impossible for other entities to return a float. The number is configured to be integer itself, but some place in the home assistant conversions is giving the state as a floating point.

music_assistant/providers/hass/__init__.py
alexbarcelo: Could you elaborate? I missed sth

music_assistant/helpers/util.py
marcelveldt: ah, its a float as string - got it. makes sense then

music_assistant/providers/hass/__init__.py
marcelveldt: You can prevent the else clause in a short function like this by returning early.
so you handle the top check and return afterwards. Its considered more pythonic and makes the code friendlier to read

music_assistant/providers/hass/__init__.py
```diff
@@ -445,16 +445,19 @@ async def _handle_player_control_mute_set(self, entity_id: str, muted: bool) ->
...
+        domain = entity_id.split(".", 1)[0]
+
+        if domain == "media_player":
             await self.hass.call_service(
-                domain="media_player",
+                domain=domain,
                 service="volume_set",
                 service_data={"volume_level": volume_level / 100},
                 target={"entity_id": entity_id},
             )
```
marcelveldt, line 456: so - do a return here and remove the else after

music_assistant/providers/hass/__init__.py
alexbarcelo: I was on the mindset of "minimizing changes to original codebase" and it didn't cross my mind.

Makes sense, fixed in the following commit.

marcelveldt (approved):
Nice, thanks!

PR #1970, opened 2025-02-20
A bit of cleanup in ABS Provider

fmunkes:
Small cleanup PR, no logical code changes.

PR #1971, opened 2025-02-20
Enhancement: Implement resume state syncing for ABS provider

fmunkes:
This mainly adds playlog syncing to abs which enables the continue listening items to be updated if a change outside of mass is happening.

Small side improvements:
- only required podcast item calls performed (oversight of me when I removed the caching)
- on_played: if marked unplayed, the progress in abs is removed now instead of setting it to zero.

Diff is reduced once #1970 is merged.


marcelveldt:
If you want this in before 2.4 is released to stable, please rebase and mark as ready for review

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -185,6 +190,39 @@ async def handle_async_init(self) -> None:
...
+        # obtain ids of items in progress
+        assert isinstance(self.mass.music.database, DatabaseConnection)
+        audiobook_playlog = await self.mass.music.database.get_rows(
+            DB_TABLE_PLAYLOG,
+            {
+                "media_type": MediaType.AUDIOBOOK.value,
+                "provider": self.lookup_key,
+            },
+        )
+        self.mass_playlog_inprogress_books = {x["item_id"] for x in audiobook_playlog}
```
marcelveldt: I'm not super fan of the fact that an individual provider is doing db queries on its own but I guess we already made some exceptions with file and builtin provider. Just wondering if we can apply one layer of abstraction with a helper function in the music controller.

Suggestion: extend the current "in progress items" with a provider filter ?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -551,6 +590,24 @@ async def on_played(
         """
         if media_type == MediaType.PODCAST_EPISODE:
             abs_podcast_id, abs_episode_id = item_id.split(" ")
+
+            # guard
```
marcelveldt: what's it guarding ? Maybe extend the comment for later readers

fmunkes: Added a couple of comments, and made the progress guard class' docstrings more verbose

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Agreed that this is a cleaner - I suggest we keep it for 2.4 as is, but then extend the function in the next beta cycle, avoiding to adjust a core function while in rc?
At the same time, the on_played callback could also be adjusted to include the MediaItemProgressReport.

marcelveldt: I will adjust the on_played callback, that is maybe the cleanest!

fmunkes: Sounds good, but that will still not allow me to get the media item IDs of items in progress if the provider is starting, if I'm not mistaken.
The MediaItemProgressReport though will make the on played a lot cleaner and more consistent, that's highly appreciated.

marcelveldt: Looked at that and that is actually not clean at all. MediaItemProgressReport is not provider related but media related and is most often the library representation - it is meant for external subscribers to scrobble/report the media playback. What info exactly do you miss now in the "on_played" callback ?

marcelveldt: > Sounds good, but that will still not allow me to get the media item IDs of items in progress if the provider is starting, if I'm not mistaken.

No, but tbh I'm not so sure why you want to do this instead of relying on the regular sync to happen.
You have the periodic sync and the realtime callbacks. That is good enough.

music_assistant/providers/audiobookshelf/__init__.py
marcelveldt: Yes, that looks good now - Somebody looking at this code in a couple of years will still understand it

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: > Looked at that and that is actually not clean at all. MediaItemProgressReport is not provider related but media related and is most often the library representation - it is meant for external subscribers to scrobble/report the media playback. What info exactly do you miss now in the "on_played" callback ?

I'm missing the duration.

fmunkes: > > Sounds good, but that will still not allow me to get the media item IDs of items in progress if the provider is starting, if I'm not mistaken.
> 
> No, but tbh I'm not so sure why you want to do this instead of relying on the regular sync to happen.
> You have the periodic sync and the realtime callbacks. That is good enough.

Problem is progress removal:

If abs removes a progress, it will be missing in the call result. But that means I'll have to mark unplayed. I didn't want to call that for every single item there is in the library, but just once for a single item, if I know that there is a progress on the mass side, which I can remove.

If I remove that part, then we have items in continue listening, although their progress is reset at abs.

That is essentially the first of the two callbacks.

fmunkes: Coming back to your initial comment, if I have the ability to filter for my provider in in the media in progress of the controller, I can remove the db stuff, and all is fine I think.

The duration in the on played would be convenient, as it saves acquiring a podcast episode just for that, but it's not strictly necessary.

marcelveldt: Updated the on_played callback with the extra metadata

marcelveldt: adding a provider filter to the "in_progress_items" is not useable as that will return library items (except for episodes) so that is kind of useless. Also, I still do not get why you want to retrieve it because its live pushed to you. You can end up in the situation where ABS played something without MA knowledge (which you correct in the sync) but you cant end up in the situation where MA played something without ABS knowledge.

fmunkes: > Updated the on_played callback with the extra metadata

First of all thanks for that! I'll update my on_played accordingly. And also thanks for being patient with me...

For your second comment, maybe there is a flaw in my logic. I agree, that everything is fine as long as the item
has some progress associated. Only in the case of progress removal I'll get a problem:
1. boundary conditions: I can always get a list of items in progress from abs 
2. The item was played, and has a progress - MA has an entry in it's playlog. All good.
3. The progress is discarded _at abs_ -> my list in 1 is missing an entry now. But how do I know which one is missing?
My solution to 3 was, that once at provider startup I read in all items ids in progress of the database. Afterward my provider takes care of adding/ removing items to this set. If I get my list in 1, I can just compare these two, and call mark_unplayed if something was removed at abs. The other solution would be to cache this sets between provider restarts, so I won't access the db.

We can also pass on progress removal completely - we would still have the in progress sync, and could just document that. Maybe that's actually for the best - at least from my listening behavior, if I remove a progress from an already played media item, I typically start listening to it again right after... It also reduces the logic to just one, and not two callbacks.

fmunkes: Ok, so i did exactly that. New `on_played` is used, and we only sync in progress items. If we remove an progress at abs, it will still be around in ma, but that's an edge case.

marcelveldt: You can also use the function in music controller to get all in progress items (just use a super high limit) but you must be aware that you get back library items for audiobooks so you'll have to translate them into a full item with provider mappings . But that is a lot of work for such an edge case. I would just do that later if this is really a problem.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -265,13 +281,9 @@ async def _get_abs_expanded_podcast(
...
 
-        Basis information,
-        abs_podcast = await self._client.get_library_item_podcast(
-            podcast_id=prov_podcast_id, expanded=False
-        ),
-        would be sufficient, but we rely on cache.
+        Basis information is sufficient.
         """
-        abs_podcast = await self._get_abs_expanded_podcast(prov_podcast_id=prov_podcast_id)
+        abs_podcast = await self._client.get_library_item_podcast(podcast_id=prov_podcast_id)
```
marcelveldt: get podcast must return a full podcast object, not some minified version

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Agreed, let's postpone that for the time being. It's probably not super important.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: That one is not minified, but I'll change it back anyways, as I forgot the type in the parsers.py, so it won't set the amount of episodes. Will be pretty late, but today.

marcelveldt:
I changed the signature of the on_played a tiny bit again to make it more clear that the provided item_id is already a provider item id

fmunkes:
> I changed the signature of the on_played a tiny bit again to make it more clear that the provided item_id is already a provider item id

adjusted accordingly

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: changed it to expanded. Also added the normal podcast version to the parsers for consistency, such that the abs_podcast to mass_podcast function yields the same podcast for all 3 versions

marcelveldt (approved):
Thanks @fmunkes !

PR #1976, opened 2025-02-22
fix: only MediaItemType has provider_mappings

fmunkes:
Pretty sure we need MediaItemType here, as provider_mappings are iterated.


marcelveldt (approved):
Thanks @fmunkes !

PR #1977, opened 2025-02-23
Add iTunes Podcast Search Provider

fmunkes:
Made this a draft due to rc of 2.4. but it does work as is.

This provider searches for podcasts via the [iTunes Search API](https://performance-partners.apple.com/search-api).

# Will need discussion:
- Did I understand the ThrottlerManager correctly?
- Would it make sense to cache the parsed podcast?
- podcastparser lib:
    - `podcastparser.parse` (l. 207) returns a dictionary, and I added `# type: ignore [no-any-return]`, as I don't know how to type hint this lib result.
    - `manifest.json`: since this lib is already part of podcastfeed, would I rather not add it there?
- icon: Not there yet, as I thought the iTunes icon promises way more than this provider provides. Maybe use the same/ a similar one as podcastfeed?
- The lint error is due to the `itunes_country_codes.json` - can we maybe add this to the ignore list of the spelling checker?

# Documentation
## Features
- Search for podcasts via iTunes API
## Configuration
- Configure your country.
- You may optionally disable explicit results
## Known issues/ notes
- The search results may contain video podcasts - I didn't come up with a reliable way to filter that, so I left them in.
- Maybe: If you'd like your podcast to be part of your library, you have to add it, or mark it as favorite. The provider is search only.


music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -0,0 +1,286 @@
...
+                "entity": "podcast",
+                "country": country,
+                "attribute": "titleTerm",
+                "explicit": explicit,
+                "limit": limit,
+                "term": search_query,
+            },
+            quote_via=urllib.parse.quote_plus,
+        )
+        url = f"https://itunes.apple.com/search?{params}"
```
marcelveldt: you can also pass these params to the aiohttp get call

fmunkes: done

music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -0,0 +1,279 @@
...
+
+    language_options = [ConfigValueOption(val, key.lower()) for key, val in country_codes.items()]
+    return (
+        ConfigEntry(
+            key=CONF_LOCALE,
+            type=ConfigEntryType.STRING,
+            label="Country",
+            required=True,
+            options=language_options,
+            default_value="de",
```
marcelveldt: any particular reason to default to de here and not US (or None)

music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -0,0 +1,279 @@
...
+                    podcast_cover=podcast_cover,
+                    domain=self.domain,
+                    lookup_key=self.lookup_key,
+                    instance_id=self.instance_id,
+                )
+
+        raise MediaNotFoundError("Episode not found")
+
+    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Get stream of item."""
```
marcelveldt:
```suggestion
        """Get streamdetails for item."""
```

music_assistant/providers/itunes_podcasts/schema.py
```diff
@@ -0,0 +1,50 @@
...
+        forbid_extra_keys = False
+        serialize_by_alias = True
+
+
+@dataclass(kw_only=True)
+class PodcastSearchResult(_BaseModel):
+    """PodcastSearchResult."""
+
+    kind: str | None = None
+    artist_name: Annotated[str | None, Alias("artistName")] = None
```
marcelveldt: I am pretty sure that mashumaro also has a Alias config with using the field, instead of going though Annotated.
https://github.com/Fatal1ty/mashumaro?tab=readme-ov-file#alias-option

music_assistant/providers/itunes_podcasts/parsers.py
```diff
@@ -0,0 +1,140 @@
...
+    episode_title = episode.get("title", "NO_EPISODE_TITLE")
+    episode_cover = episode.get("episode_art_url", podcast_cover)
+    episode_published = episode.get("published")
+    episode_enclosures = episode.get("enclosures", [])
+    if len(episode_enclosures) < 1:
+        raise RuntimeError
+    stream_url = episode_enclosures[0].get("url", None)
+    guid = episode.get("guid", stream_url)
+
+    episode_id = f"{prov_podcast_id} {stream_url} {guid}"
```
marcelveldt: is the stream url stable enough to use as episode id ?

fmunkes: Well, that is how it is done in gpodder for example, with the exception, if the episode has a guid associated. So what I could do is assign (only) the guid if it exists, and the url (or some mixture of title, podcast etc) otherwise - is that preferred? 
I'd then have to change the streamdetails as well, but that's fine.

music_assistant/providers/itunes_podcasts/__init__.py
fmunkes: I changed it to None.

music_assistant/providers/itunes_podcasts/schema.py
fmunkes: I did change that here now, but is there any other reason in this particular case than taste (and maybe later extensibility)? I used the `Annotated` everywhere in the audiobookshelf lib...

marcelveldt: Well mostly extensibility for when you want to use more/other mashumaro features

music_assistant/providers/itunes_podcasts/parsers.py
marcelveldt: I leave it up to you, what you prefer. I was just wondering if the url can change someday but I guess that wont matter that much (it will be discovered as a new item)

music_assistant/providers/itunes_podcasts/schema.py
fmunkes: Ok, then I leave the abs lib as is for the moment, but keep that in mind.

music_assistant/providers/itunes_podcasts/parsers.py
fmunkes: I changed it. We use guid if it is available, otherwise stream url. Most of the feeds appear not to have the guid, but I think it is preferred, if it is available.

fmunkes:
I also added caching of 1 day to the feed - I'm accessing the feed quite often, and I personally think it is worth to have caching in this case, as nothing is local. Also significantly improved the speed of rendering the episode list.

fmunkes:
Alright, so I would think, this is fine now - there is an alternative to using the feedparser, which would be to use the itunes collection id, and then acquire episodes via their api. However, we are then limited to 200 episodes max (sure, that's plenty) per podcast. On the other hand, I'd hope that the feed url is somewhat stable - in the end, also the itunes api yields the stream url from that feed, and the guid is also set to the stream url, so I think it's just two different ways of acquiring the same information.

marcelveldt (approved):
LGTM, nice work @fmunkes !

PR #1981, opened 2025-02-26
Feat: Move the output limiter option from DSP to Player Settings

maximmaxim345:
This PR moves the "output limiter" option from the DSP Settings to the regular Player Settings.

This has 3 advantages:
- Better discoverability.
- Decoupling from the DSP, since this option is more related with Volume Normalization.
- More control in grouped playback scenarios.

Migration is done automatically.

# Grouped Playback

The previous in the DSP integrated approach, the limiter could not be deactivated when synced with other players.
Now, the settings behaves the similarly to the Volume Normalization options:
- In permanent sync groups, the groups setting has priority.
- In regular sync groups, the leader has priority.

### DIFF
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -460,20 +460,7 @@ class ConfigController:
         In case the player does not have a DSP configuration, a default one is returned.
         """
         if raw_conf := self.get(f"{CONF_PLAYER_DSP}/{player_id}"):
-            config = DSPConfig.from_dict(raw_conf)
-            if config.enabled and not config.output_limiter:
-                # The DSP is enabled, and the user disabled the output limiter in a prior version
-                # Migrate the output limiter option to the player config
-                self.mass.config.set_raw_player_config_value(player_id, CONF_OUTPUT_LIMITER, False)
-                # The output_limiter option in the DSP config is now only used for knowing
-                # if the user disabled the limiter before. We therefore need to set it
-                # to the default value (so enabled), so this migration logic will never be called
-                # anymore for this player.
-                # TODO: remove this in a future release
-                config.output_limiter = True
-                self.set(f"{CONF_PLAYER_DSP}/{player_id}", config.to_dict())
-
-            return config
+            return DSPConfig.from_dict(raw_conf)
         else:
             # return default DSP config
             dsp_config = DSPConfig()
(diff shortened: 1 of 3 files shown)


music_assistant/helpers/audio.py
```diff
@@ -978,6 +980,23 @@ def is_grouping_preventing_dsp(player: Player) -> bool:
...
+    decides if the limiter should be turned on or not.
+    """
+    deciding_player_id = player.player_id
+    if player.active_group:
+        # Syncgroup, get from the group player
+        deciding_player_id = player.active_group
+    elif player.synced_to:
+        # Not in sync group, but synced, get from the leader
+        deciding_player_id = player.synced_to
+    return mass.config.get_raw_player_config_value(deciding_player_id, CONF_OUTPUT_LIMITER, True)
```
marcelveldt: There is one downside in using the default value like this on a raw value which is that if we may ever change the default value of the config entry, this will not change and is easily forgotten.

So maybe better to import the config entry from the constants and then use CONF_ENTRY_OUTPUT_LIMITER.key and CONF_ENTRY_OUTPUT_LIMITER.default_value

A bit more verbose but can safe some headache in the future

maximmaxim345: I agree. Changed it.

music_assistant/controllers/config.py
```diff
@@ -459,7 +460,16 @@ def get_player_dsp_config(self, player_id: str) -> DSPConfig:
...
         """
         if raw_conf := self.get(f"{CONF_PLAYER_DSP}/{player_id}"):
-            return DSPConfig.from_dict(raw_conf)
+            config = DSPConfig.from_dict(raw_conf)
+            if config.enabled and not config.output_limiter:
+                # TODO: remove this in a future release
+                self.mass.config.set_raw_player_config_value(
+                    player_id, CONF_OUTPUT_LIMITER, config.output_limiter
+                )
+                config.output_limiter = True
```
marcelveldt: I'm not sure I'm following here.
So if the output limiter was off, now its turned on ?

maximmaxim345: If we would keep the limiter off in the dsp settings, it would override the new limiter option every time.

maximmaxim345: Added a comment explaining that more in detail.

music_assistant/controllers/config.py
```diff
@@ -459,7 +460,20 @@ def get_player_dsp_config(self, player_id: str) -> DSPConfig:
...
+                self.mass.config.set_raw_player_config_value(player_id, CONF_OUTPUT_LIMITER, False)
+                # The output_limiter option in the DSP config is now only used for knowing
+                # if the user disabled the limiter before. We therefore need to set it
+                # to the default value (so enabled), so this migration logic will never be called
+                # anymore for this player.
+                # TODO: remove this in a future release
+                config.output_limiter = True
+                self.set(f"{CONF_PLAYER_DSP}/{player_id}", config.to_dict())
+
+            return config
```
marcelveldt: Makes more sense now indeed, thanks!
You could also have added the migration logic to the _migrate method in the config controller and just remove the key from config. Then its migrated for good.

marcelveldt: https://github.com/music-assistant/server/blob/dev/music_assistant/controllers/config.py#L798

maximmaxim345: Thanks for the hint.
Moved it to `_migrate`.

marcelveldt (approved):
Nice, thanks!

PR #1984, opened 2025-02-26
fix(tidal): refresh access token correctly

jozefKruszynski:


music_assistant/providers/tidal/__init__.py
```diff
@@ -661,8 +669,12 @@ async def _get_tidal_session(self) -> TidalSession:
...
             and self._tidal_session.access_token
-            and datetime.fromisoformat(str(self.config.get_value(CONF_EXPIRY_TIME)))
-            > (datetime.now() + timedelta(days=1))
+            and (
+                datetime.fromisoformat(str(self.config.get_value(CONF_EXPIRY_TIME))).replace(
+                    tzinfo=UTC
+                )
+                > (datetime.now(UTC) + timedelta(hours=1))
+            )
         ):
```
marcelveldt: this is horrible to read - better assign the expiry time into a local variable first

jozefKruszynski: You're absolutely right

marcelveldt (approved):

PR #1988, opened 2025-02-27
Add advanced networking (list of IPs) to Sonos S1 provider

mcblum:
I'll be the first to say I've never used Python really, so this is a bit of a WIP. I did notice there was one typing issue, but being new to Python I couldn't exactly figure out what it wanted from me :)

Line 382 reports `Item "float" of "float | int | str | list[float] | list[int] | list[str] | list[bool]" has no attribute "split"  [union-attr]` so I'll see if I can figure out how to type it. Functionally, the code does appear to work, and I'm able to use my manually-defined players with the S1 system.

For now, I've just disabled type checking for the assignment. If there's a better way, let me know and I'll update the PR!


mcblum:
@marcelveldt would you mind taking a look at this whenever you have a moment? Very much hoping to get this integrated into my HA setup!

music_assistant/providers/sonos_s1/__init__.py
```diff
@@ -120,6 +120,21 @@ async def get_config_entries(
...
+            "(e.g. '10.0.0.42, 10.0.0.45').\n"
+            "Invalid addresses may result in the Sonos provider "
+            "becoming unresponsive and server crashes.\n"
+            "Bidirectional unicast communication to and between all IPs is required.\n"
+            "NOT SUPPORTED, USE ON YOU'RE OWN RISK",
+            category="advanced",
+            default_value=None,
+            required=False,
+        ),
     )
```
marcelveldt: If we are duplicating this between providers we should turn this into a globally defined ConfigEntry.
Also I am pretty sure we support a better way to define multiple values instead of comma separating. 

I will have a look at this and update here.

music_assistant/providers/sonos_s1/__init__.py
```diff
@@ -362,6 +377,25 @@ async def poll_player(self, player_id: str) -> None:
...
+                    # player is offline
+                    self.logger.debug("Failed to add SonosPlayer %s: %s", player, err)
+                except Exception as err:
+                    self.logger.warning(
+                        "Failed to add SonosPlayer %s: %s",
+                        player,
+                        err,
+                        exc_info=err if self.logger.isEnabledFor(10) else None,
+                    )
+            return
```
marcelveldt: why the return ?
imo we should make the manual IP's option an ADDITIONAL discovery method, not completely disabling the primary discovery method.

music_assistant/providers/sonos_s1/__init__.py
```diff
@@ -362,6 +377,25 @@ async def poll_player(self, player_id: str) -> None:
 
     async def discover_players(self) -> None:
         """Discover Sonos players on the network."""
+        manual_ip_config: str | None
+        if (manual_ip_config := self.config.get_value(CONF_IPS)) is not None:  # type: ignore[assignment]
```
marcelveldt: let's not completely ignore typing here but help the type checker a bit by casting it.
Maybe even defining it above is enough:

manual_ip_config: str

marcelveldt:
Applied the suggested changes for you

music_assistant/providers/sonos_s1/__init__.py
mcblum: Smart! I have no idea why I didn't get an email notification with your comments, but I'm just seeing them now. Apologies for not responding.

mcblum:
Thank you, and sorry for missing these! I'll double check my notification settings.

mcblum:
@marcelveldt hey, question for you! I see this was merged to dev and I see it in my Home Assistant, but, for whatever reason, I don't see the option in the advanced config area. Am I doing something dumb?

marcelveldt:
It's only available in the beta. Are you on the beta?

mcblum:
Ah I'm not! That makes sense. I'll try to switch over.

mcblum:
@marcelveldt I think there's maybe a bug in this implementation. I'm not sure if I introduced it or not, and I'm going to try to reproduce it locally today, but here's the error I'm seeing:
```
2025-03-09 09:48:53.664 INFO (MainThread) [music_assistant] Loaded player provider SONOS S1
2025-03-09 09:48:53.666 ERROR (MainThread) [music_assistant] Error doing task: Task exception was never retrieved
  File "/app/venv/lib/python3.12/site-packages/soco/core.py", line 331, in __init__
    socket.inet_aton(ip_address)
OSError: illegal IP address string passed to inet_aton
  File "/app/venv/lib/python3.12/site-packages/music_assistant/providers/sonos_s1/__init__.py", line 375, in discover_players
    player = SoCo(ip_address)
             ^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.12/site-packages/soco/core.py", line 119, in __call__
    cls._instances[key][args] = super().__call__(*args, **kwargs)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.12/site-packages/soco/core.py", line 333, in __init__
    raise ValueError("Not a valid IP address string") from error
ValueError: Not a valid IP address string
During handling of the above exception, another exception occurred:
  File "/app/venv/lib/python3.12/site-packages/music_assistant/models/player_provider.py", line 47, in loaded_in_mass
    await self.discover_players()
  File "/app/venv/lib/python3.12/site-packages/music_assistant/providers/sonos_s1/__init__.py", line 383, in discover_players
    player,
    ^^^^^^
UnboundLocalError: cannot access local variable 'player' where it is not associated with a value
```

I'm seeing that the players that showed up before no longer do, though a few of them do load still!

mcblum:
Ah ok - I think I see what's happening. If there's either a bad IP or the Sonos device can't be reached, it seems to crash instead of continuing on. I'll do some learning about try/catch + graceful error handling in Python!

PR #1990, opened 2025-02-28
ABS: Login via API key

fmunkes:
Small enhancement to the login of the provider. Instead of providing username and password, one can use the token of a user instead. The token of each user is visible to admin user's in settings -> users [here](https://www.audiobookshelf.org/guides/users/).

Reasoning: abs supports authentication with OpenID - instead of implementing that as an additional flow, users' can just provide the token, and no additional larger logic is needed.

Lib update: added /api/authenticate endpoint for token users, which provides the same response as /login, as well as token convenience functions.

Discussions:
- [#3631](https://github.com/orgs/music-assistant/discussions/3631)
- [#3616](https://github.com/orgs/music-assistant/discussions/3616)
- [#4061 at abs](https://github.com/advplyr/audiobookshelf/discussions/4061)


muppie:
I was just trying to integrate ABS and MAS and it failed since I'm using OpenID - this is amazing!

marcelveldt (approved):
Nice, @fmunkes !

PR #1991, opened 2025-02-28
Fix Large PodcastEpisode listings not working

marcelveldt:
It turns out that some podcasts can contain many episodes, which could potentially result in a too large json message (especially with HA ingress). Adjust podcast episode listings to use asyncgenerator so it will be chunked sent over the api.


marcelveldt:
Fixes https://github.com/music-assistant/support/issues/3635

PR #1994, opened 2025-03-01
ABS: fix for another unexpected API response

fmunkes:
see [here](https://github.com/music-assistant/aioaudiobookshelf/issues/1), current_time can be None


marcelveldt (approved):
Thanks @fmunkes !

PR #1995, opened 2025-03-01
Adjust Tidal provider to a fully async implementation

jozefKruszynski:
Provider has been rewritten from scratch, however, backwards compatibility should be in place, so user should see no issue.


marcelveldt:
Looking good at first sight, so maybe merge and await beta feedback ?
What exactly is the breaking change ?
Please update the PR description with the info

jozefKruszynski:
Updated the description.

marcelveldt:
You can't add a bit of migration code to convert to the new implementation? Or make the configuration values compatible?

jozefKruszynski:
I'll see what I can do this evening. You're right that it makes sense to try to make it a smooth migration.

marcelveldt (approved):
LGTM, Nice work @jozefKruszynski !

MacgyverH:
After loading up Beta 7 I am getting the following:
```
[music_assistant.tidal] Failed to refresh token: {"status":400,"error":"invalid_request","sub_status":1002,"error_description":"invalid_request, Missing parameters: client_id"}
2025-03-06 04:58:21.627 WARNING (MainThread) [music_assistant] Error loading provider(instance) tidal--J8hgXhb5: Failed to authenticate with Tidal (will be retried later)
```

Manually reauthenticating fixes this.

Not sure if it's related to this change, but I also cannot search for new music from Tidal in 2.5.0b7.

PR #2000, opened 2025-03-04
iBroadcast: Correct disc numbers in case the album has a disc number provided

robsonke:
The disc numbers in iBroadcast are not always set in the track number (which felt like a workaround earlier) but often also in a disc number field at album level. Now supporting both flavours.

### DIFF
--- a/music_assistant/providers/ibroadcast/__init__.py
+++ b/music_assistant/providers/ibroadcast/__init__.py
@@ -372,14 +372,22 @@ class IBroadcastProvider(MusicProvider):
                 )
             },
         )
+        if track_obj["album_id"]:
+            album = await self._client.get_album(track_obj["album_id"])
+
         if "rating" in track_obj and track_obj["rating"] == 5:
             track.favorite = True
         if "length" in track_obj and str(track_obj["length"]).isdigit():
             track.duration = track_obj["length"]
-        # track number looks like 201, meaning, disc 2, track 1
-        if track_obj["track"] > 99:
+        # use the disc number if available
+        if album and album["disc"] > 0:
+            track.disc_number = album["disc"]
+            track.track_number = int(track_obj["track"])
+        # otherwise, track number might look like 201, meaning, disc 2, track 1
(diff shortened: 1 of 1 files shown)


music_assistant/providers/ibroadcast/__init__.py
```diff
@@ -372,14 +372,22 @@ async def _parse_track(self, track_obj: dict[str, Any]) -> Track:
                 )
             },
         )
+        if track_obj["album_id"]:
+            album = await self._client.get_album(track_obj["album_id"])
```
marcelveldt, line 376: Oof, this is heavy to do in the parse method of a track. Please make this optional because you're spamming api calls now.

robsonke: Yes, I know, but I need that anyway to set the album info a few lines below (431), or how is that supposed to happen?

marcelveldt: Only in the "get_track" method you should return a full track including album.
In regular (library) listings etc. you can simply return an ItemMapping as album (which is basically just a reference).

So. do you have something like the album id and the album name so you can create that itemmapping ?

Otherwise imagine a 2000 tracks listing, you wil be doing 2000 individual calls to get the album, which is inefficient.

robsonke: I have an album_id in the track info but that's it. I can enrich this in the library already but that only shifts the problem.
As you're aware, with the current ibroadcast API the library is already in memory, but that's not really an excuse. Still hoping for a better API in the near future.

marcelveldt: hmm ok, lets see how it goes then. Worst case you need to cache the album info a bit.

marcelveldt (approved):
Thanks @robsonke !

music_assistant/providers/ibroadcast/__init__.py
robsonke: thanks, just let me know, happy to help

PR #2008, opened 2025-03-08
Add ListenBrainz scrobbler provider

ijc:
Very heavily based on[^1] the last.fm scrobbler in #1850 but using [liblistenbrainz](https://github.com/metabrainz/liblistenbrainz) instead of pylast. Thanks to Willem-Jam and the reviewers there for figuring out all the hard parts...

I considered adding LB support to the existing plugin but the client library and auth flow were different enough that I thought a separate plugin was probably the way to go.

NB: I have tested this only in a backport to 2.4 branch.

For docs:

### Features

Connect to Listenbrainz and scrobble the songs you are playing. It also supports the now playing feature.

### Configuration

You will need your User Token from https://listenbrainz.org/settings/.

### Known issues / Notes

* Currently songs will only get scrobbled when they're fully played (90+%).
* Be careful for double scrobbling, which is possible if you use a music provider that also does scrobbling internally. A future improvement is to make it configurable which music providers will be scrobbled.

[^1]: i.e. a clone and hack of..

### DIFF
--- a/music_assistant/providers/listenbrainz_scrobble/__init__.py
+++ b/music_assistant/providers/listenbrainz_scrobble/__init__.py
@@ -1,0 +1,158 @@
+"""Allows scrobbling of tracks with the help of liblistenbrainz."""
+
+# icon.svg from https://github.com/metabrainz/design-system/tree/master/brand/logos
+# released under the Creative Commons Attribution-ShareAlike(BY-SA) 4.0 license.
+# https://creativecommons.org/licenses/by-sa/4.0/
+
+import asyncio
+import time
+from collections.abc import Callable
+from typing import Any
+
+from liblistenbrainz import Listen, ListenBrainz
+from music_assistant_models.config_entries import (
+    ConfigEntry,
+    ConfigValueType,
+    ProviderConfig,
+)
+from music_assistant_models.constants import SECURE_STRING_SUBSTITUTE
+from music_assistant_models.enums import ConfigEntryType, EventType
+from music_assistant_models.errors import SetupFailedError
+from music_assistant_models.event import MassEvent
+from music_assistant_models.playback_progress_report import MediaItemPlaybackProgressReport
+from music_assistant_models.provider import ProviderManifest
+
+from music_assistant.mass import MusicAssistant
+from music_assistant.models import ProviderInstanceType
+from music_assistant.models.plugin import PluginProvider
+
+CONF_USER_TOKEN = "_user_token"
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    token = config.get_value(CONF_USER_TOKEN)
+    if not token:
+        raise SetupFailedError("User token needs to be set")
+
+    assert token != SECURE_STRING_SUBSTITUTE
+
+    client = ListenBrainz()
+    client.set_auth_token(token)
+
+    return ListenBrainzScrobbleProvider(mass, manifest, config, client)
+
+
+class ListenBrainzScrobbleProvider(PluginProvider):
+    """Plugin provider to support scrobbling of tracks."""
+
+    _client: ListenBrainz = None
+    _currently_playing: str | None = None
+    _on_unload: list[Callable[[], None]] = []
+
+    def __init__(
+        self,
+        mass: MusicAssistant,
+        manifest: ProviderManifest,
+        config: ProviderConfig,
+        client: ListenBrainz,
+    ) -> None:
+        """Initialize MusicProvider."""
+        super().__init__(mass, manifest, config)
+        self._client = client
+
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await super().loaded_in_mass()
+
+        # subscribe to internal event
+        self._on_unload.append(
+            self.mass.subscribe(self._on_mass_media_item_played, EventType.MEDIA_ITEM_PLAYED)
+        )
+
+    async def unload(self, is_removed: bool = False) -> None:
+        """
+        Handle unload/close of the provider.
+
+        Called when provider is deregistered (e.g. MA exiting or config reloading).
+        """
+        for unload_cb in self._on_unload:
+            unload_cb()
+
+    async def _on_mass_media_item_played(self, event: MassEvent) -> None:
+        """Media item has finished playing, we'll scrobble the track."""
+        if self._client is None:
+            self.logger.error("no client available during _on_mass_media_item_played")
+            return
+
+        report = event.data
+
+        def make_listen(report: Any) -> Listen:
+            # album artist and track number are not available without an extra API call
+            # so they won't be scrobbled
+
+            # https://pylistenbrainz.readthedocs.io/en/latest/api_ref.html#class-listen
+            return Listen(
+                track_name=report.name,
+                artist_name=report.artist,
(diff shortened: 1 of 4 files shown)


music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+    """Initialize provider(instance) with given configuration."""
+    return ListenBrainzScrobbleProvider(mass, manifest, config)
+
+
+class ListenBrainzScrobbleProvider(PluginProvider):
+    """Plugin provider to support scrobbling of tracks."""
+
+    _client: ListenBrainz = None
+    _currently_playing: str | None = None
+    _on_unload: list[Callable[[], None]] = []
```
marcelveldt, line 53: If this provider is ever going to be multi instance, then be careful with class variables.

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _get_config(self) -> dict[str, ConfigValueType]:
+        return {
+            CONF_AUTH_TOKEN: self.config.get_value(CONF_AUTH_TOKEN),
+        }
```
marcelveldt: Lets not reinvent the wheel as this may cause confusion with MA's config.
I suggest you just feed the dict into the "_get_client" initialization

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await super().loaded_in_mass()
+
+        if not self.config.get_value(CONF_AUTH_TOKEN):
+            self.logger.info("No auth token available, don't forget to configure!")
+            return
+
+        self._client = _get_client(self._get_config())
```
marcelveldt: Is there a reason for doing the setup here instead of in setup ? This way the provider will never fail (and auto retry) on login error or server temporary outages etc.

marcelveldt: I suggest to do the initialization in the async_setup method

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+        """Media item has finished playing, we'll scrobble the track."""
+        if self._client is None:
+            self.logger.error("no client available during _on_mass_media_item_played")
+            return
+
+        report = event.data
+
+        def make_listen(report: Any) -> Listen:
+            # album artist and track number are not available without an extra API call
+            # so they won't be scrobbled
```
marcelveldt, line 94: If this is important for the scrobblers, we can add that data into the event data but I guess its not that important ?

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+            # reset currently playing to avoid it expiring when looping songs
+            self._currently_playing = None
+
+    def should_scrobble(self, report: MediaItemPlaybackProgressReport) -> bool:
+        """Determine if a track should be scrobbled, to be extended later."""
+        # ideally we want more precise control
+        # but because the event is triggered every 30s
+        # and we don't have full queue details to determine
+        # the exact context in which the event was fired
+        # we can only rely on fully_played for now
```
marcelveldt, line 139: you can also rely on the "is_playing" attribute to determine if the track stopped playing or is still ongoing.

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+        ConfigEntry(
+            key=CONF_AUTH_TOKEN,
+            type=ConfigEntryType.SECURE_STRING,
+            label="Auth Token",
+            required=True,
+            value=values.get(CONF_AUTH_TOKEN) if values else None,
+        ),
+    ]
+
+    return tuple(entries)
```
marcelveldt: why the intermediate list ?
you can simply directly return the tuple

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: This was inherited from the last.fm scrobbler.

I think it's unlikely that this one will be multi instance either.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: This was inherited from the last.fm scrobbler, not sure why it was done that way there but indeed I see no reason to keep it here.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: Inherited from last.fm scrobbler, where I guess it is part of the more complicated auth flow's requirements. Will simplify as you suggest.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: This comment came from the last.fm scrobbler.

Just supplying mbid seems to be enough for listenbrainz to figure things out (you can see my [feed](https://listenbrainz.org/user/ijc/), the entries from Saturday 8 March are from music assistant).

The [`Listen`class](https://pylistenbrainz.readthedocs.io/en/latest/api_ref.html#class-listen) accepts a bunch of other stuff relating to MB: artist_mbids, release_mbid, release_group_mbid, work_mbids, tracknumber, isrc and recording_msid, I guess they are nice to have if easily available but not critical if they are not.

While looking I also noticed some other interesting fields:
* `listening_from` -- I'll set this to `musicassistant`
* `spotify_id` -- I don't use spotify, but if this is easily available it'd be nice to add for those who do
* `tags`, `additional_info`-- One day I might add config support for these.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: This came from the last.fm scrobbler. I think they probably should both behave the same?

Maybe at some point it would be interesting to factor out the common bits of scrobbling into a base class or mixin or something.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: Another left over from the last.fm case, which is more complicated. I'll simplify.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: Did you mean `async def setup(self, config: CoreConfig) -> None:` rather than `async_setup`?

ijc: `async def setup` seems to work so I assume that's what you meant, change pushed.

marcelveldt (approved):
Nice work @ijc !

OzGav:
@ijc sorry I see the docs info at the top!

OzGav:
@ijc As a suggestion please keep an eye on issues related to lastfm as they might also affect yours but haven't been reported. Are you on Discord? There is currently a double scrobble report that is being investigated https://discord.com/channels/753947050995089438/1083643113484070953/1348582061148995624

ijc:
I have a discord account but I don't routinely go there. Thanks for the pointer.

PR #2009, opened 2025-03-08
Add Built-in (web) Player Provider

maximmaxim345:
Depends on: https://github.com/music-assistant/models/pull/58.
Frontend implementation: https://github.com/music-assistant/frontend/pull/902

# Overview

This PR adds a new, by default enabled, `BuiltinPlayer` provider for playing directly to the Music Assistant Frontend, enabling easy playback to the current device. 
It replaces the need for opening SnapWeb on a separate tab and allows playback from outside the network.
For now, it's still marked as experimental, but core functionality is implemented.

# Current Functionality

All core functionality should work with this set of PRs, specifically:
- Enabling and disabling the current browser as a target for playing music
- Play/Pause
- Volume control
- Persisting queue between sessions
- Timeout of disconnected sessions

# Future Enhancements

But there is still some work to do, like:
- General bug fixes
- Automatically enabling the web player when starting playback
- Replacing the "Experimental Web Player" with a more descriptive name like "Chrome on Android"
- Hiding the player from all other clients by default
- Deleting old queues after being disconnected for some time
- Lossless playback
- Support for selecting the streaming bitrate and container format
- Adding support for other protocols like SnapCast

# Screenshots

Before enabling:
Screenshot From 2025-03-08 19-01-36

After enabling:
Screenshot From 2025-03-08 18-57-45

### DIFF
--- a/music_assistant/helpers/webserver.py
+++ b/music_assistant/helpers/webserver.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Coroutine
-from typing import TYPE_CHECKING, Any, Final
+from typing import TYPE_CHECKING, Any, Final, Union
 
 from aiohttp import web
 
@@ -107,7 +107,9 @@ class Webserver:
     def register_dynamic_route(
         self,
         path: str,
-        handler: Callable[[web.Request], Coroutine[Any, Any, web.Response]],
+        handler: Callable[
+            [web.Request], Coroutine[Any, Any, Union[web.Response, web.StreamResponse]]
+        ],
         method: str = "*",
     ) -> Callable:
         """Register a dynamic route on the webserver, returns handler to unregister."""
--- a/music_assistant/helpers/webserver.py
+++ b/music_assistant/helpers/webserver.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Coroutine
-from typing import TYPE_CHECKING, Any, Final, Union
+from typing import TYPE_CHECKING, Any, Final
 
 from aiohttp import web
 
@@ -107,9 +107,7 @@ class Webserver:
     def register_dynamic_route(
         self,
         path: str,
-        handler: Callable[
-            [web.Request], Coroutine[Any, Any, Union[web.Response, web.StreamResponse]]
-        ],
+        handler: Callable[[web.Request], Coroutine[Any, Any, web.Response | web.StreamResponse]],
         method: str = "*",
     ) -> Callable:
         """Register a dynamic route on the webserver, returns handler to unregister."""
--- a/music_assistant/providers/builtin_player/__init__.py
+++ b/music_assistant/providers/builtin_player/__init__.py
@@ -1,0 +1,396 @@
+"""Built-in Player Provider for playing to the Music Assistant Web Interface."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from time import time
+from typing import TYPE_CHECKING, cast
+
+import shortuuid
+from aiohttp import web
+from music_assistant_models.config_entries import (
+    ConfigEntry,
+    ConfigValueType,
+    PlayerConfig,
+    ProviderConfig,
+)
+from music_assistant_models.constants import PLAYER_CONTROL_NATIVE, PLAYER_CONTROL_NONE
+from music_assistant_models.enums import (
+    BuiltinPlayerEventType,
+    ConfigEntryType,
+    ContentType,
+    EventType,
+    PlayerFeature,
+    PlayerState,
+    PlayerType,
+    ProviderFeature,
+)
+from music_assistant_models.errors import PlayerUnavailableError
+from music_assistant_models.media_items import AudioFormat
+from music_assistant_models.player import (
+    BuiltinPlayerEvent,
+    BuiltinPlayerState,
+    DeviceInfo,
+    Player,
+    PlayerMedia,
+)
+
+from music_assistant.constants import (
+    BASE_PLAYER_CONFIG_ENTRIES,
+    CONF_ENTRY_CROSSFADE,
+    CONF_ENTRY_CROSSFADE_DURATION,
+    CONF_ENTRY_FLOW_MODE_ENFORCED,
+    CONF_MUTE_CONTROL,
+    CONF_POWER_CONTROL,
+    CONF_VOLUME_CONTROL,
+    DEFAULT_PCM_FORMAT,
+)
+from music_assistant.helpers.audio import get_player_filter_params
+from music_assistant.helpers.ffmpeg import get_ffmpeg_stream
+from music_assistant.mass import MusicAssistant
+from music_assistant.models import ProviderInstanceType
+from music_assistant.models.player_provider import PlayerProvider
+
+if TYPE_CHECKING:
+    from music_assistant_models.provider import ProviderManifest
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    return BuiltinPlayerProvider(mass, manifest, config)
+
+
+async def get_config_entries(
+    mass: MusicAssistant,
+    instance_id: str | None = None,
+    action: str | None = None,
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """
+    Return Config entries to setup this provider.
+
+    instance_id: id of an existing provider instance (None if new instance setup).
+    action: [optional] action key called from config entries UI.
+    values: the (intermediate) raw values for config entries sent with the action.
+    """
+    # ruff: noqa: ARG001
+    return ()
+
+
+@dataclass
+class PlayerInstance:
+    """Dataclass for a connected instance."""
+
+    unregister_cbs: list[Callable[[], None]]
+    last_update: float
+
+
+class BuiltinPlayerProvider(PlayerProvider):
+    """Builtin Player Provider for playing to the Music Assistant Web Interface."""
+
+    _unregister_cbs: list[Callable[[], None]] = []
+    instances: dict[str, PlayerInstance] = {}
+
+    def __init__(
+        self, mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+    ) -> None:
+        """Initialize the provider."""
+        super().__init__(mass, manifest, config)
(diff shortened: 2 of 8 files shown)


music_assistant/helpers/webserver.py
```diff
@@ -107,7 +107,9 @@ def port(self):
     def register_dynamic_route(
         self,
         path: str,
-        handler: Callable[[web.Request], Coroutine[Any, Any, web.Response]],
+        handler: Callable[
+            [web.Request], Coroutine[Any, Any, Union[web.Response, web.StreamResponse]]
```
marcelveldt:
```suggestion
            [web.Request], Coroutine[Any, Any, web.Response | web.StreamResponse]
```

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+    _unregister_cbs: list[Callable[[], None]] = []
+    instances: dict[str, PlayerInstance] = {}
+
+    def __init__(
+        self, mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+    ) -> None:
+        """Initialize the provider."""
+        super().__init__(mass, manifest, config)
+        self._unregister_cbs = [
+            self.mass.register_api_command("bplayer/register", self.register_player),
```
marcelveldt: we could consider builtin_player/ as prefix or webplayer instead of "bplayer"

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
+"""Built-in Player Provider for playing to the Music Assistant Web Interface."""
```
marcelveldt: We could consider making the description a bit for universal as this is potentially not only suitable for the MA web interface but it creates a HTTP based generic player, which can be utilized by/in the MA frontend but potentially also be consumed by the user as an URL or even the HA mediabrowser player.

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+            for unregister_cb in instance.unregister_cbs:
+                unregister_cb()
+
+    async def get_player_config_entries(self, player_id: str) -> tuple[ConfigEntry, ...]:
+        """Return all (provider/player specific) Config Entries for the given player (if any)."""
+        # OPTIONAL
+        # this method is optional and should be implemented if you need player specific
+        # configuration entries. If you do not need player specific configuration entries,
+        # you can leave this method out completely to accept the default implementation.
+        # Please note that you need to call the super() method to get the default entries.
```
marcelveldt:
```suggestion
```

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+        return (
+            *BASE_PLAYER_CONFIG_ENTRIES,
```
marcelveldt: I assume you are deliberately grabbing the base entries here instead of just getting all base ones from super() ?

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+            ),
+            ConfigEntry(
+                key=CONF_MUTE_CONTROL,
+                type=ConfigEntryType.STRING,
+                label=CONF_MUTE_CONTROL,
+                default_value=PLAYER_CONTROL_NATIVE,
+                hidden=True,
+            ),
+            # For now only flow mode is supported
+            # TODO: also allow regular streams
```
marcelveldt, line 146: I don't think you can do this without flow mode ever. So I propose to just tie this on flow mode completely to not end up in a mess where you need to enqueue yourself

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+            CONF_ENTRY_FLOW_MODE_ENFORCED,
+            CONF_ENTRY_CROSSFADE,
+            CONF_ENTRY_CROSSFADE_DURATION,
+        )
+
+    async def on_player_config_change(self, config: PlayerConfig, changed_keys: set[str]) -> None:
+        """Call (by config manager) when the configuration of a player changes."""
```
marcelveldt:
```suggestion
```
Unneeded ?

music_assistant/providers/builtin_player/__init__.py
marcelveldt: Also, I'm working on a "local audio output" plugin which could easily benefit from this as well. A builtin player which can work out of the box (without the need for any 3rd party protocols)

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+            ),
+        )
+
+    async def play_media(
+        self,
+        player_id: str,
+        media: PlayerMedia,
+    ) -> None:
+        """Handle PLAY MEDIA on given player."""
+        url = f"bplayer/flow/{player_id}.mp3"
```
marcelveldt: maybe store the prefix in a constant and/or create a small helper method to generate the URL

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+
+    async def poll_player(self, player_id: str) -> None:
+        """Poll player for state updates.
+
+        This is called by the Player Manager;
+        if 'needs_poll' is set to True in the player object.
+        """
+        if instance := self.instances.get(player_id, None):
+            last_updated = time() - instance.last_update
+            if last_updated > 70:
```
marcelveldt: maybe store the 70 in a constant ?

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+    async def register_player(self, player_name: str, player_id: str | None) -> Player:
+        """Register a player.
+
+        Every player must first be registered through this `bplayer/register` API command
+        before any playback can occur.
+        Since players queues can time out, this command either will create a new player queue,
+        or restore it from the last session.
+
+        - player_name: Human readable name of the player, will we only used in case this call
+                       creates a new queue.
```
marcelveldt:
```suggestion
        - player_name: Human readable name of the player, will only be used in case this call
                       creates a new queue.
```

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+            last_update=time(),
+        )
+
+        player = Player(
+            player_id=player_id,
+            provider=self.instance_id,
+            type=PlayerType.PLAYER,
+            name=player_name,
+            available=True,
+            powered=True,
```
marcelveldt: maybe we shouldnt support power control at all for these builtin players ?
so just leave it out (which is None) and powercontrol set to "none"

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -0,0 +1,396 @@
...
+        queue = self.mass.player_queues.get(player_id)
+
+        if not (player := self.mass.players.get(player_id)):
+            raise web.HTTPNotFound(reason=f"Unknown player: {player_id}")
+
+        headers = {
+            "Content-Type": f"audio/{format_str}",
+            "Accept-Ranges": "none",
+            "Cache-Control": "no-cache",
+            "Connection": "close",
```
marcelveldt: I think you can also import the streaming headers as constant from audio helpers or streams controller.

music_assistant/helpers/webserver.py
maximmaxim345: Done

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Makes sense! Changed it.

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Expanded the description a bit.

My goal was to target the current MA frontend first, but allow HA and the future App to also use this provider.
But expanding it to allow more general use would also be nice.

But I don't know:
- How the mediabrowser in MA is implemented
- What adjustments would need to be made for your "local audio output"
- And how a user URL would be used

So let's discuss that on Discord!

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Done

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: No, that was just a leftover from early testing. Thanks for catching that!

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: I would like to keep this possibility open for now.
Seeking right now requires re-transmitting the streamed audio, potentially eating away precious mobile data.

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Right now yes. Removed it for now.

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Will do that in a future PR, since audio quality selection is not done yet.

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Done

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Done

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Good Idea, adjusted the code.

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: Tweaked that part to use the constant.

music_assistant/providers/builtin_player/__init__.py
marcelveldt: To support client-side seeking, you land into the minefield of having to support range support server-side which is nearly impossible because we often do not know upfront how big the data is going to be. That is why we now always let the server perform seeking (and simply restart the stream) but I'm open to suggestions to revisit that on a rainy day. I didn't consider the mobile client (data) aspect, that is a good one indeed.

marcelveldt (approved):
Nice work again @maximmaxim345 !

PR #2017, opened 2025-03-10
Ignore playlists which are stored with album tracks

### LINKED ISSUE #3206: Add to playlist dialog limited to 500 playlists

2.3.2

2024.11.4

- [X] Yes

### The problem

[Discord Discussion with Traveling_Gav about the issue here](https://discord.com/channels/753947050995089438/753947050995089442/1308725572628643941)

When trying to add a track to a playlist from the MA Frontend, it opens the dialog of playlists, but it only shows a subset of my playlists + options to add new playlist to MA or Spotify. 
A lot of my playlists from Spotify are prefixed with `-` and `<space>` making names like
```
- rock
- top40
- chill
- rap
```
All the playlists which follow this format are shown correctly, but _no other playlist_  is shown from any music providers. Adding a new playlist in Spotify with this prefix makes it show up, and same if I create a MA playlist with this prefix it show up, but all other playlists do not show up

- Adding a new playlist works, but it does *not* show up unless I give it a name with the same prefix
- The "Playlist" view in the frontend shows all the playlists correctly as expected

### How to reproduce

Create playlist named like `- <something>` and then try to add a track to a playlist in the frontend, and see what is being shown

1. Search for Track
2. Click on `...`
3. Click `Add to Playlist...`

Then only a subset of playlists are shown

### Music Providers

Spotify
Filesystem (Remote)

### Player Providers

N/A

### Full log output

_No response_

### Additional information

_No response_

2024.11.1

Home Assistant Container

Linux

marcelveldt:
Add setting to ignore playlist files which are in a nested level (e.g. next to the album tracks). This prevents that 1000s of (unpractical) albumtracks playlists end up in the library. The setting is enabled by default.

It still imports playlists that are in the music folder root or in a single sublevel (e.g. playlists).

Fixes https://github.com/music-assistant/support/issues/3206

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -75,6 +75,7 @@ from .constants import (
     AUDIOBOOK_EXTENSIONS,
     CONF_ENTRY_CONTENT_TYPE,
     CONF_ENTRY_CONTENT_TYPE_READ_ONLY,
+    CONF_ENTRY_IGNORE_ALBUM_PLAYLISTS,
     CONF_ENTRY_MISSING_ALBUM_ARTIST,
     CONF_ENTRY_PATH,
     IMAGE_EXTENSIONS,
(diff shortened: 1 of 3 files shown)


OzGav:
I think we need to be clear about what the setting will do. We have those other issues with people having their playlists in directories other than the root which could be down two levels off to the side. If I understand correctly the assumption is albums are in /ROOT/ARTIST/ALBUM  and so anything that is two levels down or below the root will be ignored?

marcelveldt:
> If I understand correctly the assumption is albums are in /ROOT/ARTIST/ALBUM and so anything that is two levels down or below the root will be ignored?

Correct. I think its clear enough.
I can not think of any valid reason to have your playlists in some deep folder. And if you do, then disable this setting. You cant have both.

OzGav:
Yep understood. I would make one suggestion to your label where it says "1 level deep in the folder structure" to "1 level deep anywhere in the folder structure" to try and avoid people thinking it only applies to albums as if MA has a way to work that out.

marcelveldt:
too late.. feel free to PR that later

PR #2028, opened 2025-03-11
[LastFM Scrobbler] Add some basic unit tests and fix a bug

wjzijderveld:
I extracted most of the logic to a dedicated event handler, so I didn't have to boot the entire provider in the tests. 

Are there guidelines/best-practices around naming of variables? Especially _ prefixes for "private/protected" variables? I'm quite new to Python so I'm just winging it right now.


marcelveldt:
Yeah, private vars are prefixed indeed. Have a look at the PEP8 style handbook here: https://peps.python.org/pep-0008/

marcelveldt (approved):

PR #2030, opened 2025-03-12
Match capitalisation

OzGav:
Fix the missing leading capital here

image

### DIFF
--- a/music_assistant/providers/squeezelite/__init__.py
+++ b/music_assistant/providers/squeezelite/__init__.py
@@ -303,7 +303,7 @@ class SlimprotoProvider(PlayerProvider):
                 label=f"Preset {index}",
                 description="Assign a playable item to the player's preset. "
                 "Only supported on real squeezebox hardware or jive(lite) based emulators.",
-                category="presets",
+                category="Presets",
                 required=False,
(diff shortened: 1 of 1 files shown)


marcelveldt (changes requested):
this is a key, not a label. If you want to prettify it, add it to the translation strings

OzGav:
I looked here https://github.com/music-assistant/frontend/blob/main/src/translations/en.json but it wasn't there so I looked elsewhere!

MacgyverH:
Translation is already being handled by the front end for the category names, right? Should just be a matter of adding a new entry into this section?
https://github.com/music-assistant/frontend/blob/6566bf1ef1d02dfaa0a045ce4d02fa5975b5eff0/src/translations/en.json#L291-L298

like this:
```
        "category": {
            "audio": "Audio",
            "advanced": "Advanced settings",
            "generic": "Generic settings",
            "announcements": "Announcements configuration",
            "airplay": "Airplay specific settings",
            "player_controls": "Player controls",
            "presets": "Presets"
        },
```

marcelveldt:
Exactly, a new category needs to be added there.

PR #2031, opened 2025-03-12
Feat: Mutichannel Parametric Equalizer

maximmaxim345:
Depends on https://github.com/music-assistant/models/pull/60
Companion PR for the frontend https://github.com/music-assistant/frontend/pull/917

# Overview
The Parametric Equalizer now natively support multiple channels, so you can individually correct the left and right speaker!
Also includes (partial) support for importing and exporting presets with multiple channels.

# Screenshot
## Editing a filter applied to all channels (the default)
image
## Editing a band only affecting the left channel
image
## Editing the preamp for the right channel
image

### DIFF
--- a/music_assistant/helpers/dsp.py
+++ b/music_assistant/helpers/dsp.py
@@ -33,7 +33,10 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
         # "volume" is handled for the whole audio stream only, so we'll use the pan filter instead
         elif has_per_channel_preamp:
             channel_config = []
-            for channel_id, gain_db in dsp_filter.per_channel_preamp.items():
+            all_channels = [AudioChannel.FL, AudioChannel.FR]
+            for channel_id in all_channels:
+                # Get gain for this channel, default to 0 if not specified
+                gain_db = dsp_filter.per_channel_preamp.get(channel_id, 0)
                 # Apply both the overall preamp and the per-channel preamp
                 total_gain_db = dsp_filter.preamp + gain_db
                 if total_gain_db != 0:
--- a/music_assistant/helpers/dsp.py
+++ b/music_assistant/helpers/dsp.py
@@ -24,11 +24,10 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
         List of FFmpeg filter parameter strings
     """
     filter_params = []
-    preamp = 0
 
     if isinstance(dsp_filter, ParametricEQFilter):
-        if dsp_filter.preamp:
-            preamp = dsp_filter.preamp
+        if dsp_filter.preamp != 0:
+            filter_params.append(f"volume={dsp_filter.preamp}dB")
         for b in dsp_filter.bands:
             if not b.enabled:
                 continue
@@ -111,8 +110,5 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
             filter_params.append(
                 f"equalizer=frequency=9000:width=18000:width_type=h:gain={dsp_filter.treble_level}"
             )
-
-    if preamp != 0:
-        filter_params.insert(0, f"volume={preamp}dB")
 
     return filter_params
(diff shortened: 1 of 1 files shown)


OzGav:
If you edit, say, the left channel and then go back to ALL and make a change there does that remove all settings from the left channel? Or does it add  the two together?

maximmaxim345:
Looking at it again, it was a bit confusing. I changed a couple of texts to be more clear.
I tried to make the associated channel options on bands and on the "Add band" button more explicit.

But I'm still not sure about "This band will be applied to - All".

> If you edit, say, the left channel and then go back to ALL and make a change there, does that remove all settings from the left channel? Or does it add the two together?

The channel option above the graph only changes two things:
- What channel/channels will be plotted in the graph
- To what channel a newly created band (with "Add band") will be bound by default.

Each band individually has a channel option, so nothing will be lost when switching back and forth.

There aren't many multi-channel Parametric EQs around as a reference (ROON, for example, doesn't support that). And those that I know of (Easy Effects and Neutron) are not that "graphical" and require converting a single channel EQ to a multichannel one (which copies all bands to all separate channels, and going back is not so straightforward).

MacgyverH:
Just a thought I had about this. but curious what everyone else thinks on this, and maybe I'm alone here. 

What is the possibility of adding the ability to "name" the bands as opposed to Just "Band 1", "Band 2"?

OzGav:
That makes sense to me now. It will get quite busy but I guess people that would do per channel EQ will know what they are doing. I wonder if you can save some real estate by slimming down the label to (L only) or (R only)?

Question for my education. Will it be clever if you import L& R presets and if the freq, gain and Q factor are the same for both channels will it will combine them into an ALL setting or is that undesirable.

marcelveldt:
models are bumped

maximmaxim345:
I actually tried a few prototypes before this one mainly:
1. Having a Channel option for the PEQ. So if a PEQ is set to left, it will only apply all filters to the left channel.
2. Showing only bands of the selected channel. So effectively each channel would have an separate list of bands.
3. Using opacity/css-styles to show the channel of a band.

But everything  has it's own drawback:
1. Then we could not support APO presets with multiple channels in one file
2. Filters applied to All channels become very confusing
3. We end up with 4 different styles for the band "pills"

> That makes sense to me now. It will get quite busy but I guess people that would do per channel EQ will know what they are doing. 

So, I think the best way is to try to keep regular All channel EQs as simple as possible and just accept that multichannel EQs can be complicated.

> I wonder if you can save some real estate by slimming down the label to (L only) or (R only)?

Like this?
image

> Question for my education. Will it be clever if you import L& R presets and if the freq, gain and Q factor are the same for both channels will it will combine them into an ALL setting or is that undesirable.

Interesting Idea.
I can only think of potential potential problems with surround sound, but that irrelevant since we don't support that (yet).
Will try that later with a future PR.


Also, what do you think of hiding all new options by default? Then it will behave exactly the same as before for users who don't need that:
image

maximmaxim345:
> Just a thought I had about this. but curious what everyone else thinks on this, and maybe I'm alone here.
> 
> What is the possibility of adding the ability to "name" the bands as opposed to Just "Band 1", "Band 2"?

I think this can quickly get very complicated.
Would having the option to select bands by clicking on the graph also solve your use case?

Please use the [Feature Requests & Ideas section](https://github.com/orgs/music-assistant/discussions/categories/feature-requests-and-ideas) for suggestions like this.

OzGav:
Yes to the mockup with “L only” I think that makes sense but up to you if you are happy with it.

If you can think of potential problems with surround then I say don’t do it as you don’t need to have to change later. I was just thinking out loud.

I like the idea of having the default simpler presentation and button to go multichannel. Looks neat and tidy!

maximmaxim345:
Ok, all multichannel options are now hidden by default and bands use a (translatable) short channel name.
Also adjusted the import logic a bit to allow for more use cases, and fixed a couple of bugs.

For the docs, we should probably have an section describing how to import different multi channel presets (since it depends on how it was exported), so write me on discord for that.

All in all, the PR should be ready now!

marcelveldt (approved):
Nice work!

PR #2034, opened 2025-03-13
[Chore] Move podcastparser parse functions from itunes search provider to helpers

fmunkes:
This PR moves the functions converting between a parsed podcast with episodes from podcastparser to MA objects into a helper file. The itunes search provider is adjusted accordingly.

I only added an optional different than default item id, such that the podcastfeed provider can be adjusted non-breaking.


marcelveldt (approved):
Nice!

PR #2035, opened 2025-03-13
Podcastfeed: Use helper functions

fmunkes:
This PR makes the podcastfeed provider use the helper functions from #2034.

I additionally added caching, such that you don't have the reload the full provider just for updating the episodes anymore. Instead, the parsed object is updated now on regular library syncs.

The provider is now mypy-ready.

The change is non-breaking, the original item ids are used.


music_assistant/providers/podcastfeed/__init__.py
```diff
@@ -161,111 +158,84 @@ async def get_podcast_episodes(
...
-                    remotely_accessible=True,
-                )
-            ]
-        episode.metadata.description = episode_obj["description"]
-        if "explicit" in episode_obj:
-            episode.metadata.explicit = episode_obj["explicit"]
 
-        return episode
+    async def _get_podcast(self) -> dict[str, Any]:
+        # see music-assistant/server@6aae82e
```
marcelveldt:
```suggestion
        # without user agent, some feeds can not be retrieved
        # https://github.com/music-assistant/support/issues/3596
```

marcelveldt (approved):
Nice work @fmunkes !

PR #2040, opened 2025-03-14
Add Recommendations to Tidal provider

jozefKruszynski:


music_assistant/controllers/music.py
```diff
@@ -495,7 +498,10 @@ async def recently_played(
         )
         db_rows = await self.mass.music.database.get_rows_from_query(query, limit=limit)
         result: list[ItemMapping] = []
-        available_providers = ("library", *get_global_cache_value("available_providers", []))
+        available_providers = (
+            "library",
+            *get_global_cache_value("available_providers", []),
+        )
```
marcelveldt: there are unrelated changes in the PR, maybe your lint setup is not what we've setup for the project ?

jozefKruszynski: Seems that I had Black installed and overriding Ruff in certain circumstances.

music_assistant/providers/tidal/__init__.py
```diff
@@ -761,17 +782,61 @@ async def get_track(self, prov_track_id: str) -> Track:
         except (ClientError, KeyError, ValueError) as err:
             raise MediaNotFoundError(f"Track {prov_track_id} not found") from err
 
-    async def get_playlist(self, prov_playlist_id: str) -> Playlist:
+    async def get_playlist(self, prov_playlist_id: str) -> TidalMix | Playlist:
```
marcelveldt: You're overriding the default implementation here which is not something we should allow in providers.
Can't you use any other way to detect the mix, like a private function which works this out so that the public method can stay untouched ?

Idea. Maybe have something in the item_id (like a prefix or something) ?

jozefKruszynski: I'll remove it

marcelveldt (approved):
Nice work @jozefKruszynski !

PR #2042, opened 2025-03-15
Refactor common code from listenbrainz and lastfm scrobblers

ijc:
Pull the common code from `LastFMEventHandler` into a common base class and use it for `ListenBrainzEventHandler` too.

This puts all the logic for deciding when to scrobble into a single place.

---

I wasn't really sure where the common code (or the tests which I moved over) belonged, happy to move them around to according to maintainer preferences though.

@wjzijderveld FYI

### DIFF
--- a/music_assistant/helpers/scrobbler.py
+++ b/music_assistant/helpers/scrobbler.py
@@ -1,0 +1,86 @@
+"""Helper class to aid scrobblers."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from music_assistant_models.event import MassEvent
+    from music_assistant_models.playback_progress_report import MediaItemPlaybackProgressReport
+
+
+class ScrobblerHelper:
+    """Base class to aid scrobbling tracks."""
+
+    logger: logging.Logger
+    currently_playing: str | None = None
+    last_scrobbled: str | None = None
+
+    def __init__(self, logger: logging.Logger) -> None:
+        """Initialize."""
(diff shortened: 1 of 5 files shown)


wjzijderveld:
Nice! I was planning to do something similar when I found some time. Ideally I'd like to make the tests a contract test, so actual implementations get tested instead of only the helper, but this is already amazing. 

The main reason I didn't do it in the first place, is because I also want to see if it's doable to make the abstraction on a level that makes it possible to just move all scrobblers into a single plugin. That sounds very doable for the scrobble part, but the dynamic configuration it requires might be a nightmare. Have you looked at that by any chance?

ijc:
The config/auth aspects of LastFM vs ListenBrainz were different enough that I didn't put much thought into combining them -- as you say that looks like it might be quite tricky.

music_assistant/helpers/scrobbler.py
```diff
@@ -0,0 +1,82 @@
+"""Helper class to aid scrobblers."""
+
+import asyncio
+import logging
+
+from music_assistant_models.event import MassEvent
+from music_assistant_models.playback_progress_report import MediaItemPlaybackProgressReport
```
marcelveldt: If these are only used for type checking, they should be moved into a "if TYPE_CHECKING" block
You can import the TYPE_CHECKING constant from the typing library

ijc: Done. Ruff disliked it until I added `from __future__ import annotations` too. Since I see that elsewhere in the code base I assume that's fine.

marcelveldt (approved):
Nice work!

PR #2043, opened 2025-03-16
feat: Built-in Web Player Part 2

maximmaxim345:
Depends on https://github.com/music-assistant/models/pull/61
Companion PR https://github.com/music-assistant/frontend/pull/918

Expanding on https://github.com/music-assistant/server/pull/2009 this PR better integrates the Web Player into Music Assistant.

# Main changes

- Power controls are now handled just like any other player in Music Assistant. So just pressing play should (hopefully) work now!
- The web player now is registered by default, so the experience is now more seamless.
- The player associated with the currently open browser now gets a screen icon
- Only one tab will be selected for doing the actual playback at a time. If that is closed, a different tab will be selected for playback.
- Players are now hidden from other users by default
- Players will soon also be hidden in Home Assistant by default
- The player is now named "This Device"

# Known Issues

There are still a few issues left. Some of those are:
- Since the player now is registered by default, it will now create (even more) "Experimental Web Players" in the Player Settings in Music Assistant
- There is no way to disable the Web Player for users that don't want to use that
- Don't yet know if the "only one tab plays music" logic is that reliable yet. But it seams to work for me now.

Solutions for all of those and more will come in future PRs.

### DIFF
--- a/music_assistant/providers/builtin_player/__init__.py
+++ b/music_assistant/providers/builtin_player/__init__.py
@@ -288,6 +288,8 @@ class BuiltinPlayerProvider(PlayerProvider):
             supported_features=player_features,
             needs_poll=True,
             poll_interval=POLL_INTERVAL,
+            hidden_by_default=True,
+            expose_to_ha_by_default=False,
         )
 
         await self.mass.players.register_or_update(player)
--- a/music_assistant/providers/builtin_player/__init__.py
+++ b/music_assistant/providers/builtin_player/__init__.py
@@ -214,6 +214,8 @@ class BuiltinPlayerProvider(PlayerProvider):
                 else BuiltinPlayerEventType.POWER_OFF
             ),
         )
+        if (not powered) and (player := self.mass.players.get(player_id)):
+            player.powered = False
 
     async def poll_player(self, player_id: str) -> None:
         """Poll player for state updates.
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):

marcelveldt:
> Since the player now is registered by default, it will now create (even more) "Experimental Web Players" in Music Assistant and Home Assistant.

Let's start by setting the "hidden_by_default" and "expose_to_ha_by_default" flags.

> There is no way to disable the Web Player for users that don't want to use that

Enabling the feature as a whole should be done in the UI config options.

music_assistant/providers/builtin_player/__init__.py
```diff
@@ -216,6 +235,11 @@ async def poll_player(self, player_id: str) -> None:
 
     async def remove_player(self, player_id: str) -> None:
         """Remove a player."""
+        self.mass.signal_event(
+            EventType.BUILTIN_PLAYER,
+            player_id,
+            BuiltinPlayerEvent(type=BuiltinPlayerEventType.TIMEOUT),
```
marcelveldt, line 241: Is timeout the only reason to remove a player ?

marcelveldt (approved):

music_assistant/providers/builtin_player/__init__.py
maximmaxim345: That's just a workaround for now so the frontend doesn't try to access the removed player.
Will figure out a better solution later.

PR #2048, opened 2025-03-18
Add artist and release mbids to listenbrainz scobbles

ijc:
https://github.com/music-assistant/models/pull/63 adds the relevant fields to `MediaItemPlaybackProgressReport`.

---

This will need a models bump to pickup the above PR before CI will pass.

[Here](https://listenbrainz.org/user/ijc/?max_ts=1741877325) are some listens without this change, listenbrainz has guessed at the release id based on the recording ids. Since the same recording can be in different versions of the release it doesn't always guess the same. The most visible impact is that the cover art varies, but also its not always the actual release I was listening to.

[Here](https://listenbrainz.org/user/ijc/?max_ts=1742313743) are some listens of the same album with this PR applied, since it has been told the release id it didn't need to guess so all the listens have the same release id and the cover art is consistent.


marcelveldt:
new version of the models package was released (1.1.41) - it just needs to be bumped in the pyproject and all_requirements.txt

ijc:
@marcelveldt Thanks! I've done the bump.

ijc:
Looks like the bump brings some new fields elsewhere (in `PodcastEpisode`) too. I can have a look this evening.

ijc:
I've set the new `PodcastEpisode.publish_date` field to `None` everywhere the linter complained about its omission. I suppose @khers has a plan to set it to something meaningful at some point.

khers:
> I've set the new `PodcastEpisode.publish_date` field to `None` everywhere the linter complained about its omission. I suppose @khers has a plan to set it to something meaningful at some point.

Thank you for taking on that work, I had meant to get to it today. The subsonic provider will be using this field shortly.

ijc:
Not sure what is wrong with the CI -- it looks like a problem with the snapcase dep installation which I don't think can be caused by the changes here.

marcelveldt:
CI can be a bit grumpy sometimes. I just triggered a re-run. Lets see if that helps

marcelveldt:
strictly taken the publish_date change should have been in a separate/dedicated PR (and maybe the model changed to have None as default) but lets ignore it for now.

marcelveldt (approved):

PR #2051, opened 2025-03-19
Add support for Spotify Connect volume control

Nezz:
This PR makes it possible to control the volume of the player using Spotify Connect's volume control:

https://github.com/user-attachments/assets/3fb8fefc-d295-4b25-a849-fcd6fe40ccb3

Caveat: Changing the volume from Music Assistant or Home Assistant will not change the volume in Spotify. Only the initial volume can be set in librespot.

### DIFF
--- a/music_assistant/providers/spotify_connect/__init__.py
+++ b/music_assistant/providers/spotify_connect/__init__.py
@@ -215,7 +215,7 @@ class SpotifyConnectProvider(PluginProvider):
                 "--volume-ctrl",
                 "fixed",
                 "--initial-volume",
-                f"{self.player.volume_level if self.player else 100}",
+                f"{self.player.volume_level if self.player and self.player.volume_level else 100}",
                 "--enable-volume-normalisation",
                 # forward events to the events script
                 "--onevent",
--- a/music_assistant/providers/spotify_connect/__init__.py
+++ b/music_assistant/providers/spotify_connect/__init__.py
@@ -25,6 +25,7 @@ from music_assistant_models.enums import (
     ProviderFeature,
     StreamType,
 )
+from music_assistant_models.errors import UnsupportedFeaturedException
 from music_assistant_models.media_items import AudioFormat
 from music_assistant_models.player import PlayerMedia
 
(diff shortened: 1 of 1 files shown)


music_assistant/providers/spotify_connect/__init__.py
```diff
@@ -213,7 +215,7 @@ async def _librespot_runner(self) -> None:
                 "--volume-ctrl",
                 "fixed",
                 "--initial-volume",
-                "100",
+                f"{self.player.volume_level if self.player else 100}",
```
marcelveldt: Note that volume_level can be None for players that do not support volume control so this should be a little more gracefully handled.

Nezz: Thanks for pointing this out. All my media players have volume, so I wasn't sure how that would behave. Updated👍

marcelveldt (approved):
Thanks @Nezz !

Algram:
@Nezz This does not work at all for me from the Spotify Android App. Is there anything I need to configure for it to work correctly?

Nezz:
It should work fine out of the box assuming that your player supports volume control.

xydix:
@Algram Its working for me on android but every step is small, very small. I guess it is 100 steps. Maybe this could be fixed @Nezz? In my case Spotify Android app on a Pixel 9 Pro. Tried with my Voice PE and Google cast. Same behavior in both cases.

Nezz:
The library used to connect with Spotify Connect is hadcoded to have 64 steps:
https://github.com/plietar/librespot/blob/8971d3aa6864469a79c2640bb2afaf41c3d3728d/src/spirc.rs#L94

A discussion I found indicates that they choose this because it's the same as the desktop Spotify app. At least that was the case 9 years ago.

Algram:
@Nezz So for me it does still not work at all from the official Spotify Android App, no matter how often I click the change volume button. Sometimes it jumps back to the volume set via the Music Assistant web interface, but it never increases or decreases the volume.

Nezz:
I recommend enabling the debug logs for the Spotify Connect provider. Look for messages starting with `Received metadata on webservice` when changing the volume.

PR #2052, opened 2025-03-21
Make airplay provider robust to empty dacp request

Dvad:
Some device (Phorus PS10) seems to send empty request
Maybe as a ack message?
Since there is nothing to do in that function with empty request we return early instead of failing later on the header parsing.


marcelveldt (approved):
Thanks!

music_assistant/providers/airplay/provider.py
```diff
@@ -548,7 +548,13 @@ async def _handle_dacp_request(  # noqa: PLR0915
                 raw_request += recv
                 if len(recv) < 1024:
                     break
-
+
```
marcelveldt:
```suggestion
```

PR #2036, opened 2025-03-13
Add gPodder podcast provider

fmunkes:
This PR adds support for clients compatible with the gPodder [Subscription](https://gpoddernet.readthedocs.io/en/latest/api/reference/subscriptions.html) and [EpisodeAction](https://gpoddernet.readthedocs.io/en/latest/api/reference/events.html) API. It has been tested with [opodsync](https://github.com/kd2org/opodsync) as an example for the classic gpodder api, as well as with the [Nextcloud GPodderSync](https://apps.nextcloud.com/apps/gpoddersync).

Functions of PR #2034 are used. I actively do not support the gpodder.net service, as it is frequently down, and very slow in response. We can of course discuss this. Docs:

# gPodder

## Features
- Progress reporting and acquiring
- Populates Libraries with Podcasts
- Updates playlog on regular provider syncs

### gpodder.net compatible webservice
A mygpo compatible web service is supported, and this provider is tested against
[opodsync](https://github.com/kd2org/opodsync)
For setup you need:
- the service url, e.g. http://192.168.1.20:14000 or https://sync.yourdomain.com 
- your username
- your password
- a device id
The device id can be any ascii string, but keep in mind, that this is used for syncing.
So another client must use the same client id.

Please note, that gpodder.net is deliberately _not_ supported. The provider relies on
frequent api calls, and the service hosted there is known to be either slow or fully
unresponsive, which will slow down MA itself. Consider using a locally hosted alternative.

### nextcloud-gpodder
The provider supports [nextcloud-gpodder/ gpoddersync](https://apps.nextcloud.com/apps/gpoddersync).
For setup, you need your nextcloud url, and then click the "authenticate with nextcloud" button to
start the authentication flow. Click save when you are done.

PR #2053, opened 2025-03-22
fix: gPodder not checking guid on playlog sync if it is present

fmunkes:
I forgot to check for the guid while the playlog is synced. This may cause a "MediaNotFoundException" on provider sync for nextcloud.
Fixed in this PR, I made it a little more verbose as well.

Issue was falsely report here #2036


marcelveldt (approved):
Thanks, @fmunkes !

PR #2060, opened 2025-03-24
Fix Audible provider crash during library sync

ztripez:
### Problem

The Audible provider was crashing during library sync due to a `TypeError` when attempting to iterate over a `None` value:

```
TypeError: 'NoneType' object is not iterable
```

This occurred when trying to process `platinum_keywords` in audiobook metadata.

### Fix

- Resolved the root issue using the `or []` pattern:

  ```python
  # Before:
  genre.replace("_", " ") for genre in audiobook_data.get("platinum_keywords", "")
  
  # After:
  genre.replace("_", " ") for genre in (audiobook_data.get("platinum_keywords") or [])
  ```

- Improved robustness by introducing better error handling:
  - Defensive checks for missing or malformed data
  - Safe access to nested structures using `.get()` with sensible defaults
  - Wrapped `int()` conversions in try/except blocks to prevent hard crashes
  - Ensured images are only added when valid paths exist
  - Added error handling for duration parsing to avoid propagation of bad data

### Result

The Audible provider can (should) now sync libraries without crashing, even when encountering books with incomplete or malformed metadata. One bad entry no longer halts the entire scan.

### DIFF
--- a/music_assistant/providers/audible/audible_helper.py
+++ b/music_assistant/providers/audible/audible_helper.py
@@ -69,6 +69,42 @@ class AudibleHelper:
         self.provider_instance = provider_instance
         self.logger = logger or logging.getLogger("audible_helper")
 
+    async def _process_audiobook_item(
+        self, audiobook_data: dict[str, Any], total_processed: int
+    ) -> tuple[Audiobook | None, int]:
+        """Process a single audiobook item from the library."""
+        content_type = audiobook_data.get("content_delivery_type", "")
+        if content_type in ("PodcastParent", "NonAudio"):
+            self.logger.debug(
+                "Skipping non-audiobook item: %s (%s)",
+                audiobook_data.get("title", "Unknown"),
+                content_type,
+            )
+            return None, total_processed + 1
+
+        # Ensure asin is a valid string
+        asin = str(audiobook_data.get("asin", ""))
+        cached_book = None
+        if asin:
+            cached_book = await self.mass.cache.get(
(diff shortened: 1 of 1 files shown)


music_assistant/providers/audible/audible_helper.py
```diff
@@ -449,54 +520,36 @@ async def _parse_audiobook(self, audiobook_data: dict[str, Any] | None) -> Audio
...
         reviews = audiobook_data.get("editorial_reviews", [])
-        if reviews:
-            book.metadata.review = _html_to_txt(reviews[0])
+        if reviews and reviews[0]:
+            book.metadata.review = _html_to_txt(str(reviews[0]))
+
+        # Set genres
         book.metadata.genres = {
-            genre.replace("_", " ") for genre in audiobook_data.get("platinum_keywords", "")
+            genre.replace("_", " ") for genre in (audiobook_data.get("platinum_keywords") or [])
```
marcelveldt, line 539:
```suggestion
            genre.replace("_", " ") for genre in (audiobook_data.get("platinum_keywords", []))
```

Why not assign a default value in the getter ?
Or is there a chance that the item does exist in the dict but with a null value ?

ztripez: Yes exactly, it seams like the item can exist in all states: missing, existing: null, existing: value.

marcelveldt: yikes, that is awful. All clear then.

marcelveldt (approved):

PR #2063, opened 2025-03-26
Soundcloud: Fallback to the track details for missing entries in playlists

robsonke:
Solves https://github.com/music-assistant/support/issues/3783


OzGav:
@robsonke mypy failing unfortunately

music_assistant/providers/soundcloud/__init__.py
```diff
@@ -264,11 +264,15 @@ async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> lis
...
-                # {'id': 123, 'kind': 'track', 'monetization_model': 'NOT_APPLICABLE',
-                # 'policy': 'ALLOW'}
+                # {'id': 123, 'kind': 'track', 'monetization_model': 'NOT_APPLICABLE'}
                 if "title" in item:
                     if track := await self._parse_track(item, index):
                         result.append(track)
+                # But also try to get the track details if the track is not in the playlist
+                else:
+                    track_details = await self._soundcloud.get_track_details(item["id"])
+                    if track := await self._parse_track(track_details[0], index):
```
marcelveldt, line 274: how often can this happen ? because this is a very expensive call to do. On large playlists you could end up doing many calls to the api. Or is this just incidental?

marcelveldt:
> @robsonke mypy failing unfortunately

That was because of the other issues we were talking about tonight. 
Just rebase your branch @robsonke and it should be cleaner

music_assistant/providers/soundcloud/__init__.py
robsonke: It's all playlists from track 6. See the screenshot in the linked issue. In the soundcloud lib I don't see any other calls. The API might have one but so far I know there's no docs of that either. Hence I was stuck :)

marcelveldt: Oof, what a mess. So on a 1000 tracks playlists, you are going to do 994 individual requests to "get track" ?
That is going to be slowwwwwww and most likely will get you end up in api blocking/throttlers.

Anyways, let's accept it for now as you cant fix a bad api but just that you are aware.

marcelveldt (approved):

music_assistant/providers/soundcloud/__init__.py
robsonke: Oh I am, it sucks. Those companies should stop being so closed, but wishful thinking..

PR #2067, opened 2025-03-27
Subsonic: Set metadata.release_date when available

khers:
We want to use this field for sorting episodes when displayed.


marcelveldt:
Do note that this is not yet used for sorting at all, we could add that later to the frontend.
For now, you should make sure that you give back a sorted listing with the position field.

marcelveldt (approved):

khers:
> Do note that this is not yet used for sorting at all, we could add that later to the frontend. For now, you should make sure that you give back a sorted listing with the position field.

Yup, I was thinking that the sorting should be done in the front end if we add that. I can do it in the provider, but that would be surprising to people who used other podcast providers.

PR #2074, opened 2025-03-29
ABS: Enhancement - support for recommendations

fmunkes:
This PR adds support for recommendations to ABS.

ABS has several categories in its [personalized view endpoint](https://api.audiobookshelf.org/#get-a-library-39-s-personalized-view). We are not adding recommendation rows per lib here, but collect them for all libs to avoid flooding the home tab. In my personal example, I have 7 recommendation categories I can pick from.

The lib has been adjusted to support the additional endpoint already.


marcelveldt (approved):
LGTM!

PR #2076, opened 2025-03-30
Add recommendations to itunes_podcasts

fmunkes:
This PR adds a single country-specific recommendation row with trending podcasts to the itunes podcasts provider.
I use the Apple API found [here](https://rss.marketingtools.apple.com/)

I also found a small bug - missing enumerate - in the podcast_parsers, and fixed that.


marcelveldt (approved):
Nice work!

PR #2086, opened 2025-03-31
fix: Recommendations for ABS and iTunes Podcasts

fmunkes:
ABS: Missing cached library ids were the trouble makers.

iTunes: More graceful handling of non-available items added.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ mass = "music_assistant.__main__:main"
 ignore-words-list = "provid,hass,followings,childs"
 skip = """*.js,*.svg,\
 music_assistant/providers/itunes_podcasts/itunes_country_codes.json,\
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -63,6 +63,7 @@ mass = "music_assistant.__main__:main"
 ignore-words-list = "provid,hass,followings,childs"
 skip = """*.js,*.svg,\
 music_assistant/providers/itunes_podcasts/itunes_country_codes.json,\
+music_assistant/providers/itunes_podcasts/__init__.py,\
```
marcelveldt: Is there a reason you're excluding a whole file from the codespell checker ?
why not just add a single word ?

fmunkes: I didn't think...

marcelveldt (approved):
Thanks @fmunkes !

PR #2097, opened 2025-04-02
Chore: Add mypy for helpers

fmunkes:
- In line 1483 I changed from `player` to `player.player_id` - is that correct?
- In line 96 the asyncio.Task needs a type hint I don't know yet

Besides:
It is a lot of "assert foo is not None" etc - is that ok?
I'm not able to check for MusicProvider without the `if TYPE_CHECKING` due to a circular import otherwise. Could we make that cleaner?


music_assistant/helpers/audio.py
```diff
@@ -167,27 +200,37 @@ async def _stream_from_cache() -> AsyncGenerator[bytes, None]:
...
+        assert self._cache_file is not None
         CACHE_FILES_IN_USE.add(self._cache_file)
         self._first_part_received.clear()
         self._all_data_written.clear()
         extra_input_args = self.org_extra_input_args or []
+        audio_source: AsyncGenerator[bytes, None] | str | int
         if self.org_stream_type == StreamType.CUSTOM:
-            audio_source = self.mass.get_provider(self.streamdetails.provider).get_audio_stream(
+            provider = self.mass.get_provider(self.streamdetails.provider)
+            assert provider is not None
```
marcelveldt: this assert is redundant with the check below

music_assistant/helpers/audio.py
```diff
@@ -167,27 +200,37 @@ async def _stream_from_cache() -> AsyncGenerator[bytes, None]:
...
+                assert isinstance(provider, MusicProvider)
+            audio_source = provider.get_audio_stream(
                 self.streamdetails,
             )
         elif self.org_stream_type == StreamType.ICY:
             raise NotImplementedError("Caching of this streamtype is not supported!")
         elif self.org_stream_type == StreamType.HLS:
             if self.streamdetails.media_type == MediaType.RADIO:
                 raise NotImplementedError("Caching of this streamtype is not supported!")
+            assert self.org_path is not None
```
marcelveldt: the asserts are unavoidable to help the type checker but i prefer to add a comment next to it that the assert is for type checking so somebody reading it later will have a clue why all the asserts are there.
Another approach could be to simply put all asserts behind a "if TYPE_CHECKING" block but that is even more verbose

marcelveldt:
```suggestion
            assert self.org_path is not None  # for type checking
```

marcelveldt:
> I'm not able to check for MusicProvider without the if TYPE_CHECKING due to a circular import otherwise. Could we make that cleaner?

Yeah, i have been staring at that a couple of times as well and i couldn't come up with a clean solution.
We can make it easier if we let the "get_provider" simply always return a provider and raise otherwise.

Maybe @Jc2k has some ideas ?

Jc2k:
If it's abnormal for get_provider to fail, let's make it raise. Otherwise if it's reasonable to fail, leaving it as optional means eventually mypy will force all get_provider users to consider that possibility (there isn't a way to signal "you must handle this exception").

Not a fan of if TYPE_CHECKING but can't think of a way around it.

music_assistant/helpers/audio.py
fmunkes: I added the comment to all assertions, and will do so for the next files as well. Indeed, let's not make it even more verbose.

fmunkes:
I left the "if TYPE_CHECKING" in then for this specific case. Even if the function raises, I still need to check the provider type - or would you introduce a function like "get_music_provider"?

fmunkes:
I went through all helper files now. I'm having a little trouble type hinting the dynamic routes (and maybe double check the static routes) in webserver.py, and would ask for help there.
```
music_assistant/helpers/webserver.py:36: error: Missing type parameters for generic type "Callable"  [type-arg]
music_assistant/helpers/webserver.py:114: error: Missing type parameters for generic type "Callable"  [type-arg]
music_assistant/helpers/webserver.py:125: error: Function is missing a return type annotation  [no-untyped-def]
music_assistant/helpers/webserver.py:150: error: Returning Any from function declared to return "Response"  [no-any-return]
```
Other than that I think it's not looking too bad.

fmunkes:
I ignored the webserver.py again, such that this PR at least passes all tests. Also merged the recent changes here again.

marcelveldt (approved):
Amazing work @fmunkes !

PR #2100, opened 2025-04-03
Make Zeroconf only bind to the default interface

altserg:
Aligning Music Assistant and Home Assistant behavior. This patch forces Zeroconf to bind only to the default interface, that actually already includes all possible system interfaces

Currently Zoeroconf in Mass is listening on all interfaces, inclusing the default interface 0.0.0.0. This causes errors in setups when the system re-connects to different WiFi networks and systems dunning Docker containers.

Similar issue was fixes in Hass in 2020, issue link: https://github.com/home-assistant/core/issues/35031

### DIFF
--- a/music_assistant/mass.py
+++ b/music_assistant/mass.py
@@ -19,7 +19,13 @@ from music_assistant_models.errors import MusicAssistantError, SetupFailedError
 from music_assistant_models.event import MassEvent
 from music_assistant_models.helpers import set_global_cache_values
 from music_assistant_models.provider import ProviderManifest
-from zeroconf import InterfaceChoice, IPVersion, NonUniqueNameException, ServiceStateChange, Zeroconf
+from zeroconf import (
+    InterfaceChoice,
+    IPVersion,
+    NonUniqueNameException,
+    ServiceStateChange,
+    Zeroconf,
+)
 from zeroconf.asyncio import AsyncServiceBrowser, AsyncServiceInfo, AsyncZeroconf
 
 from music_assistant.constants import (
@@ -129,7 +135,7 @@ class MusicAssistant:
         self.version = await get_package_version("music_assistant") or "0.0.0"
         # create shared zeroconf instance
         # TODO: enumerate interfaces and enable IPv6 support
-        self.aiozc = AsyncZeroconf(ip_version=IPVersion.V4Onlyy, interfaces=InterfaceChoice.Default)
+        self.aiozc = AsyncZeroconf(ip_version=IPVersion.V4Only, interfaces=InterfaceChoice.Default)
         # create shared aiohttp ClientSession
         self.http_session = ClientSession(
             loop=self.loop,
--- a/music_assistant/mass.py
+++ b/music_assistant/mass.py
@@ -19,7 +19,7 @@ from music_assistant_models.errors import MusicAssistantError, SetupFailedError
 from music_assistant_models.event import MassEvent
 from music_assistant_models.helpers import set_global_cache_values
 from music_assistant_models.provider import ProviderManifest
-from zeroconf import IPVersion, NonUniqueNameException, ServiceStateChange, Zeroconf
+from zeroconf import InterfaceChoice, IPVersion, NonUniqueNameException, ServiceStateChange, Zeroconf
 from zeroconf.asyncio import AsyncServiceBrowser, AsyncServiceInfo, AsyncZeroconf
 
 from music_assistant.constants import (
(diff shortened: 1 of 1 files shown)


Jc2k:
Won't this break setups where they do want zeroconf on multiple interfaces?

Doesn't HA have a UI for configuring this? Default interface is the default setting for that UI I think.

marcelveldt:
We could tie it to the interface that was chosen in the stream server settings ?

altserg:
Currently Zeroconf in MASS binds to 0.0.0.0, 127.0.0.1, eth0, wlan0, etc, all interfaces. This is doubling for all interfaces, as 0.0.0.0 actually means all interfaces.
Hass and avahi only bind to 0.0.0.0. 
As I indicated, this pull requests aligns mass code base with hass, where similar issue was fixed 4 years ago

Jc2k:
> As I indicated, this pull requests aligns mass code base with hass, where similar issue was fixed 4 years ago

It's what HASS does *in some cases*. But HA has configuration for the use cases this change on its own breaks.

Jc2k:
> We could tie it to the interface that was chosen in the stream server settings ?

I think that would be a better option.

altserg:
The server normally binds to 0.0.0.0, so functionally it will be the same.
However, in case you change the server bind address to a particular interface/ip address, like eth0, then it won't bind to others, like wlan0. This might be undesired behavior for some cases, but probably will be ok for most of regular use cases. But I can't judge if this will also be ok for containers, as I'm not an expert there.

marcelveldt:
Let's merge this as-is. At some point I'd like to split off the mdns discovery logic into its own core module and then also add some config entries to make it configurable. I will add a task for that on the backlog.

marcelveldt:
Please have a look at the linter issues found in the CI check

marcelveldt:
Fixed the lint issues for you

altserg:
Thank you. I was out, couldn't fix the linting

jjhelin:
> > We could tie it to the interface that was chosen in the stream server settings ?
> 
> I think that would be a better option.

Definitely it would be a better option, as for people with multiple vlans the streaming interface/vlan is the subnet shared with players. In my case music assistant inhabits two vlans, with the default being used to publish the web server, and the secondary for communicating with players (that reside in a restricted vlan/subnet).

Binding zeroconf to default broke player discovery going from 2.5.0 to 2.5.1 due to this commit. Until this issue is fixed, myself and others with a similar setup are prevented from upgrading. I personally can not circumvent this issue easily by redefining the default interface, as music assistant shares an LXC container with other components that rely on the current networking setup.

@marcelveldt can this issue be reopened, or should I open a new bug/improvement ticket?

Jc2k:
You can't really reopen a PR that's merged, so would need to be an issue. Or a PR with a revert or one of maintainer suggestions implemented.

Can you confirm how you've isolated it to this change - have you run 2.5.1 with just this change reverted?

jjhelin:
> You can't really reopen a PR that's merged, so would need to be an issue. Or a PR with a revert or one of maintainer suggestions implemented.

Alright, sounds reasonable.

> Can you confirm how you've isolated it to this change - have you run 2.5.1 with just this change reverted?

I have not modified the 2.5.1 by producing a custom build with this reverted. I'm running music-assistant in docker with host mode (for mdns/zeroconf), and going from 2.5.0 to 2.5.1 resulted in my AirPlay and ChromeCast speakers no longer being available. So I took a look at the changelog and found this issue. The discussion and code change strongly suggested this as the culprit, but I have not done actual proof via elimination.

For what it's worth, intuitively binding the socket to only the default interface (0.0.0.0) alone would seem sufficient, as the socket probably receives mdns packets arriving on all interfaces. However, my speculation is that sending is the issue. I have no experience with mdns/zeroconf, but I would imagine it now only sends broadcasts on one interface. And with a system with multiple interfaces, it selects the one with the default route, which in my case is not on the subnet with the speakers.

I'm not advocating for merely reverting the change in this PR, after all it tackles a real issue. My preference would be to make the mdns/zeroconf socket bind interface configurable as you suggested, e.g. using the streaming server interface as @marcelveldt marcelveldt speculated. Actually, now rereading this thread, it seems he may have already ticketed this: "I will add a task for that on the backlog." If so, there is no need for me to open a duplicate.

Sorry for the handwavy stuff, really appreciate the work you guys have put in this.

:edit: netstat -a |grep mdns with 2.5.0 (the first is my speaker vlan)

```
udp        0      0 10.10.40.17:mdns        0.0.0.0:*
udp        0      0 172.19.0.1:mdns         0.0.0.0:*
udp        0      0 172.17.0.1:mdns         0.0.0.0:*
udp        0      0 homeassistant.local.domain.net:mdns 0.0.0.0:*
udp        0      0 172.20.0.1:mdns         0.0.0.0:*
udp        0      0 172.18.0.1:mdns         0.0.0.0:*
udp        0      0 localhost:mdns          0.0.0.0:*
udp        0      0 10.10.30.17:mdns        0.0.0.0:*
udp        0      0 0.0.0.0:mdns            0.0.0.0:*
```
and with 2.5.1:

```
udp        0      0 0.0.0.0:mdns            0.0.0.0:*
```

PR #2101, opened 2025-04-04
always stop cliraop on pause

### LINKED ISSUE #3819: airplay: no audio output after pausing playback

2.4.4

- [x] Yes

### The problem

Hi there,

My daughter enjoys hearing audible audio books with music assistant using the airplay provider (shaiport-sync). After a while she takes a break and pauses playback. But often when she comes back 
and tries to resume playback there is no audio output at all.

I noticed that the **cliraop** instance  for her airplay output was still running, so I simply killed it.
It was really nice to see that after that a new cliraop process was started and the audio playback continued as expected.

The airplay provider is `shairport-sync` and I have [**allow_session_interruption** enabled in`shairport-sync`](https://github.com/mikebrady/shairport-sync/blob/a2326f32313b1e3fc4b2bade28e40425bbd240c6/scripts/shairport-sync.conf#L112)
Maybe cliraop does not handle session interruption / idle timeouts properly.

### How to reproduce

Use airplay output with shairport-sync instance. Pause playback. Resume playback after 30min. 

### Music Providers

audible

### Player Providers

airplay

### Full log output

-

### Additional information

-

2025.3.0

Home Assistant Container

Linux

- [x] Yes

r10r:
Simply stop `cliraop` instance on pause to avoid disconnected player on resume when playback was paused.
This is reproducible with `shairport-sync` when either session interruption or a session timeout is active.

Fixes https://github.com/music-assistant/support/issues/3819

### DIFF
--- a/music_assistant/providers/airplay/provider.py
+++ b/music_assistant/providers/airplay/provider.py
@@ -275,13 +275,7 @@ class AirplayProvider(PlayerProvider):
         player = self.mass.players.get(player_id)
         if not player:
             return
-        if player.group_childs:
-            # pause is not supported while synced, use stop instead
-            self.logger.debug("Player is synced, using STOP instead of PAUSE")
-            await self.cmd_stop(player_id)
-            return
-        airplay_player = self._players[player_id]
-        await airplay_player.cmd_pause()
+        await self.cmd_stop(player_id)
 
     @lock
     async def play_media(


marcelveldt:
This should not be needed. MA will send a STOP anyways after 30 seconds of pause. 
Shairport sync seems to be the only endpoint having this issue. I don't think this is a solution this way, we need to dig deeper why this happens. Downside of sending stop is that the entire session is disconnected, making a resume slower.

r10r:
Hi @marcelveldt  thanks for your response.

>  This should not be needed. MA will send a STOP anyways after 30 seconds of pause.

Hmm, this is strange. So the question is why is the cliraop process still running then ?

> I don't think this is a solution this way, we need to dig deeper why this happens.

You're right.

> Downside of sending stop is that the entire session is disconnected, making a resume slower.

But the resume is a really fast and does not bother me. But maybe this is in issue with lower end devices. 
Anyway I consider it  a solution until there is a better way to fix this since my daughter is really happy with it and so am I. 

Curiously iOS airplay clients I don't have any issues. When the (`shairplay-sync`) session timeouts or is interrupted in they simply get disconnected and playback stops. This is ways nicer in music-assistant because it pauses playback instead.

marcelveldt:
I just tested this for a while and I can confirm that the change in this PR is not needed.

- After being paused for 30 seconds, MA sends the stop and the session is disconnected.
- After sending pause+play within 30 seconds, resume works correctly
- After sending play after the player was stopped, resume works as expected.

PR #2103, opened 2025-04-05
Snapcast fix external server stream

SantiagoSotoC:


music_assistant/providers/snapcast/__init__.py
```diff
@@ -595,6 +595,7 @@ async def _streamer() -> None:
                 stream_path = stream.path
             if not stream.path:
                 stream_path = "tcp://" + stream._stream["uri"]["host"]
+            stream_path = stream_path.replace("0.0.0.0", self._snapcast_server_host)
```
konikvranik, line 598: Isn't it better to construct the stream path right with the Snapcast server host in the line above? Or is the stream path sent from Snapcast server this wrong way?

marcelveldt, line 598: Please make this more readable. So something like this:

```
if not self._use_builtin_server:
    port = stream._stream["uri"]["port"]
    stream_path = f"tcp://{self._snapcast_server_host}:{port}"
elif stream.path:
    stream_path = stream.path
else:
    stream_path = "tcp://" + stream._stream["uri"]["host"]
```

SantiagoSotoC: There is no port attribute, I would have to use a split, I think it is less readable. 

  Besides leaving the ip 0.0.0.0.0 for the local server is also a problem.

marcelveldt: How comes the host is 0.0.0.0 ?
Why is this snapcast library such a mess ?

marcelveldt (approved):
I'll merge this for now but please try to make this code a bit friendlier to read. Maybe add some comments.

PR #2104, opened 2025-04-05
Audible: fixes type error of release_date metadata

ztripez:
Small fix to address provider crash if the audiobook returns release_date as a string


ztripez:
Ah you have already  addressed this.

marcelveldt:
The linter found an issue. I think you are both importing the datetime module and datetime from datetime module
EDIT: fixed it for you

marcelveldt (approved):

PR #2107, opened 2025-04-05
Fix usage of album loudness during playback

cpaulik:
This fixes usage of album loudness tags as discussed in https://github.com/orgs/music-assistant/discussions/3822

### DIFF
--- a/music_assistant/controllers/streams.py
+++ b/music_assistant/controllers/streams.py
@@ -14,6 +14,7 @@ import shutil
 import urllib.parse
 from collections.abc import AsyncGenerator
 from dataclasses import dataclass, field
+from math import inf
 from typing import TYPE_CHECKING
 
 from aiofiles.os import wrap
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/streams.py
```diff
@@ -1021,7 +1022,13 @@ async def get_queue_item_stream(
         elif streamdetails.volume_normalization_mode == VolumeNormalizationMode.MEASUREMENT_ONLY:
             # volume normalization with known loudness measurement
             # apply volume/gain correction
-            gain_correct = streamdetails.target_loudness - streamdetails.loudness
+            if streamdetails.prefer_album_loudness and streamdetails.loudness_album not in [
+                inf,
+                -inf,
+            ]:
```
marcelveldt:
```suggestion
            if streamdetails.prefer_album_loudness and streamdetails.loudness_album not in (
                inf,
                -inf,
            ):
```
let's use a tuple here. Also I wonder I you got an invalid measurement in the db.

cpaulik: > Also I wonder I you got an invalid measurement in the db.

How so. What looks suspicious to you?

marcelveldt:
Good catch btw, somehow this code fell off the grid. It was there at some point.

marcelveldt (approved):
Thanks, @cpaulik !

PR #2108, opened 2025-04-05
Also use previous track to work out if an album is playing

cpaulik:
In the previous implementation the last track in an album used the track replay gain instead of the album replay gain

Related to https://github.com/orgs/music-assistant/discussions/3822

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1041,9 +1041,9 @@ class PlayerQueuesController(CoreController):
                 and queue_item.media_item.album.item_id == next_item.media_item.album.item_id
             )
         )
+        current_index = self.index_by_id(queue_id, queue_item.queue_item_id)
         previous_track_from_same_album = (
-            queue.current_index is not None
-            and (previous_index := max(queue.current_index - 1, 0))
+            (previous_index := max(current_index - 1, 0))
             and (previous_index > 0)
             and (previous_item := self.get_item(queue_id, previous_index))
             and (
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1026,8 +1026,9 @@ class PlayerQueuesController(CoreController):
             queue.display_name,
         )
 
-        # work out if we are playing an album and if we should prefer album loudness
-        playing_album_tracks = (
+        # work out if we are playing an album and if we should prefer album
+        # loudness
+        next_track_from_same_album = (
             next_index is not None
             and (next_item := self.get_item(queue_id, next_index))
             and (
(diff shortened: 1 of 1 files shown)


cpaulik:
I'm also happy to refactor this logic into a separate function and add tests if the general idea is accepted.

music_assistant/controllers/player_queues.py
```diff
@@ -1040,6 +1041,23 @@ async def _load_item(
                 and queue_item.media_item.album.item_id == next_item.media_item.album.item_id
             )
         )
+        current_index = self._queues[queue_id].current_index
```
marcelveldt: This is a bit dangerous because maybe some new playback started and the queue was first cleared ?
Also this assumes that current_index is already updated (this relies on the player).
I would suggest to just do prev_index = next_index-1 if next_index else None and then ofc guard on the None

music_assistant/controllers/player_queues.py
```diff
@@ -1040,6 +1041,23 @@ async def _load_item(
...
+        previous_track_from_same_album = (
+            current_index is not None
+            and (previous_index := max(current_index - 1, 0))
```
marcelveldt: ah, this is exactly what I meant above in my comment. Just delete the line 1044 then

music_assistant/controllers/player_queues.py
cpaulik: If I understood correctly then `next_index` is `None` for the last item in the queue so if I get `prev_index` from that then this won't work in this edge case.

I've now made a small change to re-use the `queue` object from line 1017. Would that be safer?

music_assistant/controllers/player_queues.py
```diff
@@ -1040,6 +1041,22 @@ async def _load_item(
                 and queue_item.media_item.album.item_id == next_item.media_item.album.item_id
             )
         )
+        previous_track_from_same_album = (
+            queue.current_index is not None
```
marcelveldt: queue.current_index is the CURRENT index. That is potentially not the index before queue_item which is loaded here because this is all async operated. You run into a minefield of race conditions.

Use the helper functions "get_next_item" and "get_index" giving the index of the queue_item to correctly lookup the item before and after the item that is being loaded. That is safer than relying on queue.current_index or queue.current_item because that is only telling us what the player is actually now playing, not what is being loaded.

music_assistant/controllers/player_queues.py
```diff
@@ -1026,8 +1026,9 @@ async def _load_item(
             queue.display_name,
         )
```
marcelveldt, line 1028: index = self.index_by_id(queue_id, queue_item.queue_item_id)
prev_index = max(0, index-1)

music_assistant/controllers/player_queues.py
```diff
@@ -1040,6 +1041,22 @@ async def _load_item(
                 and queue_item.media_item.album.item_id == next_item.media_item.album.item_id
             )
         )
+        previous_track_from_same_album = (
+            queue.current_index is not None
+            and (previous_index := max(queue.current_index - 1, 0))
```
```suggestion
        current_index = self.index_by_id(queue_id, queue_item.queue_item_id)
        previous_track_from_same_album = (
            (previous_index := max(current_index - 1, 0))
```

This should do it

music_assistant/controllers/player_queues.py
marcelveldt: It's more that you should not rely on queue.current_item in this part of the code (where its async loading some item into the buffer of a player). Maybe the player is buffering ahead or the item load happens before the metadata is updated. That is why we pass along the item in these functions, because it is the item which is actually being loaded/requested and not what the metadata report is saying we're playing.

cpaulik: Thanks for the explanation.

cpaulik:
Thanks. I applied your suggestion and it works as expected in my local environment.

marcelveldt (approved):
Thanks, @cpaulik !

PR #2115, opened 2025-04-07
Enhancement: gPodder - optional handling of delete action

fmunkes:
This PR introduces the optional handling of a [delete episode action](https://gpoddernet.readthedocs.io/en/latest/api/reference/events.html). User options:

- no handling - as before
- mark played - these episodes will be marked as finished, and still appear in the UI
- hide - these episodes will be hidden from the episode listing

Given the API docs, this Action type is rather for clients which can download episodes, so I thought it should be configurable.
See music-assistant/support#3838.


music_assistant/providers/gpodder/__init__.py
```diff
@@ -48,7 +54,12 @@
     parse_podcast_episode,
 )
 from music_assistant.models.music_provider import MusicProvider
-from music_assistant.providers.gpodder.client import EpisodeActionNew, GPodderClient
+from music_assistant.providers.gpodder.client import (
```
marcelveldt:
```suggestion
from .client import (
```

music_assistant/providers/gpodder/__init__.py
```diff
@@ -219,6 +244,20 @@ async def get_config_entries(
...
+        ConfigEntry(
+            key=CONF_DELETED_EPISODES,
+            type=ConfigEntryType.STRING,
+            label="Handling of deleted episodes",
+            description="",
+            multi_value=False,
+            options=[
+                ConfigValueOption("Hide", DeletedEpisodesHandling.HIDE.value),
+                ConfigValueOption("Mark played", DeletedEpisodesHandling.MARK_PLAYED.value),
+                ConfigValueOption("No handling", DeletedEpisodesHandling.NONE.value),
```
marcelveldt: So, what would be the purpose of keeping a deleted episode ?
Isnt it true that deleted episodes should also dissapear from MA ? Or am I missing some point here ?

fmunkes: API docs state, this is used to signal other clients, that a previously downloaded episode has been removed from local storage.
I didn't see, why this should mean, that you should not be able to play that episode on MA anymore. But I can follow your suggestion, I'm undecided here.

marcelveldt: I would not even make this configurable at all, that only adds complexity and maintenance.
Instead simply mark these episodes with available set to False

fmunkes: Yeah, that's much better, I adjusted it.

music_assistant/providers/gpodder/__init__.py
```diff
@@ -387,17 +399,22 @@ async def get_library_podcasts(self) -> AsyncGenerator[Podcast, None]:
...
+                        case EpisodeActionPlay():
+                            await self.mass.music.mark_item_played(
+                                mass_episode,
+                                fully_played=_action.position >= _action.total,
+                                seconds_played=_action.position,
+                            )
+                        # case EpisodeActionDelete():
+                        #     # mark unplayed, so it disappears from playlog
+                        #     # it will be marked unavailable when yielding episodes
+                        #     await self.mass.music.mark_item_unplayed(mass_episode)
```
marcelveldt: leftover ?

fmunkes: yes!

marcelveldt (approved):
LGTM, thanks @fmunkes !

PR #2116, opened 2025-04-07
Various small bugfixes and enhancements

marcelveldt:
- Prevent lingering ffmpeg proc in airplay playback
- Prevent queue resume position counting when playback pause
- Fix sonos pause support when airplay mode active
- Use safe default of 16 bits for sonos

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1022,7 +1022,7 @@ class PlayerQueuesController(CoreController):
         BYPASS_THROTTLER.set(True)
 
         self.logger.debug(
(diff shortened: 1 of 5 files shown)


tgozed:
I've been having stuttering issues when listening to the radio on my Sonos play:1 on every version after 2.3.6

It got a LOT better with this release, I'm guessing because of the safe default change, but it's still skipping every minute or so. There's nothing in the logs that corresponds with the skips. Should I raise a bug or is there something I can do differently?

tgozed:
Just read a few different threads. Tried chunking the playback but it was still skipping but switched the codec from flat to MP3 and that's fixed it.

PR #2118, opened 2025-04-08
Tidal: Various playlist and mix changes

jozefKruszynski:


music_assistant/providers/tidal/__init__.py
```diff
@@ -466,35 +466,57 @@ async def _get_data(
             return await self._handle_response(response, return_etag)
 
     @prepare_api_request
-    async def _post_data(
+    async def _post_or_put_data(
```
marcelveldt: this looks a bit wird with all the IF's underneath. I would suggest a dedicated PUT and POST method and helpers for shared logic

jozefKruszynski: Fair enough, I had it separate first and then changed to this and I couldn't decide which way made more sense in the end.

music_assistant/providers/tidal/__init__.py
```diff
@@ -495,6 +496,40 @@ async def _post_data(
...
+                    "dict[str, Any]",
+                    await self._handle_response(response, return_etag=False),
+                )
+        else:
+            # Use json parameter for JSON data (default)
+            async with self.mass.http_session.put(url, json=data, **kwargs) as response:
+                return cast(
+                    "dict[str, Any]",
+                    await self._handle_response(response, return_etag=False),
+                )
```
marcelveldt:
```suggestion
        # Use json parameter for JSON data (default)
        async with self.mass.http_session.put(url, json=data, **kwargs) as response:
            return cast(
                "dict[str, Any]",
                await self._handle_response(response, return_etag=False),
            )
```
There is a return above so no need for the else

music_assistant/providers/tidal/__init__.py
```diff
@@ -495,6 +496,40 @@ async def _post_data(
                     await self._handle_response(response, return_etag=False),
                 )
```
marcelveldt: same comment here for this else block

marcelveldt (approved):
Thanks, @jozefKruszynski !

PR #2122, opened 2025-04-09
Round loudness measurements to two digits

cpaulik:
This avoids very long floating point nunbers showing up in the frontend.

Before:

image

After:

image

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -842,10 +842,11 @@ class MusicController(CoreController):
             },
         )
         if db_row and db_row["loudness"] != inf and db_row["loudness"] != -inf:
-            loudness = db_row["loudness"]
+            loudness = round(db_row["loudness"], 2)
             loudness_album = db_row["loudness_album"]
-            if loudness_album in (inf, -inf):
-                loudness_album = None
+            loudness_album = (
+                None if loudness_album in (None, inf, -inf) else round(loudness_album, 2)
+            )
             return (loudness, loudness_album)
 
         return None


music_assistant/controllers/music.py
```diff
@@ -842,10 +842,9 @@ async def get_loudness(
             },
         )
         if db_row and db_row["loudness"] != inf and db_row["loudness"] != -inf:
-            loudness = db_row["loudness"]
+            loudness = round(db_row["loudness"], 2)
             loudness_album = db_row["loudness_album"]
-            if loudness_album in (inf, -inf):
-                loudness_album = None
+            loudness_album = None if loudness_album in (inf, -inf) else round(loudness_album, 2)
```
```suggestion
            loudness_album = None if loudness_album in (None, inf, -inf) else round(loudness_album, 2)
```
You need to guard this because loudness_album is an optional value

cpaulik: Thanks. Updated. I wrongly assumed it could only be `(inf, -inf)` based on the previous code but of course the previous code only re-set it to `None` anyway.

music_assistant/controllers/music.py
```diff
@@ -842,10 +842,9 @@ async def get_loudness(
             },
         )
         if db_row and db_row["loudness"] != inf and db_row["loudness"] != -inf:
-            loudness = db_row["loudness"]
+            loudness = round(db_row["loudness"], 2)
             loudness_album = db_row["loudness_album"]
-            if loudness_album in (inf, -inf):
-                loudness_album = None
+            loudness_album = None if loudness_album in (None, inf, -inf) else round(loudness_album, 2)
```
marcelveldt: this doesn't fit a one-liner anymore. See the linter warnings from CI.

TIP: setup the dev environment with the setup script so pre-commit gets installed

cpaulik: Sorry, I just accepted your suggestion in Github. I've now re-done this locally. Hope it works now.

marcelveldt (approved):
Thanks, @cpaulik !

PR #2124, opened 2025-04-12
fix: library items removed by music provider not removed in MA library

fmunkes:
The `sync_library` function of the music provider model missed adding db ids to the currently available item id set.


marcelveldt (approved):
Thabks, @fmunkes !

PR #2126, opened 2025-04-12
Fix all lowercase display names for AirPlay players

lassebm:
The Zeroconf cache keys are all lowercase, use service name from AsyncServiceInfo.


marcelveldt (approved):
Thanks @lassebm !
As a follow-up PR we should maybe just drop the name argument from the callback as it exists in the info object anyways.

PR #2127, opened 2025-04-13
Fix AirPlay branding

lassebm:
The correct Apple branding is "AirPlay". Also fixed "macOS" a few places.


marcelveldt (approved):
Thanks @lassebm !

PR #2142, opened 2025-04-22
Add Yamaha MusicCast provider

fmunkes:
Adds support for Yamaha MusicCast. The provider uses the aiomusiccast library. There is some revisiting to be done in the "sync to" player section, see comments in the code.

Docs:
# MusicCast

- supported devices are those by [aiomusiccast](https://github.com/vigonotion/aiomusiccast), i.e. devices since 2015
- if the device is not supported in HA, MA won't support it either, as we use the same library

## Multi-zone devices
If a device has multiple zones, i.e. main + at max 3. others, only one of these zones can use netusb input. This means you can only stream to one of these zones. However, if you are streaming to main, the other zones will still be available for grouping to main, as the provider then uses "main_sync" for these sources.

If a non-main zone is either group member or group server, the main zone cannot join this group.

## Features
- grouping/ ungrouping
- perfect sync
- multi-zone devices
  - specify a non-net zone when leaving the group in player config to switch to
- auto-discovery
- enqueuing
- gapless playback
- metadata sent alongside
- source selection of MusicCast device
- playback information for non-MA network content
- HA integration and MA integration can be used alongside


OzGav:
Is the second remark related to the first? Im not sure what that means for the user?

Could you also check this table and advise what is supported from sync correction down? https://www.music-assistant.io/player-support/

Also a bunch of people will be very happy! This is no. 11 on the most wanted list :-)

edit: where you have this, is that true for all groups types, temporary, sync and universal?

> If a non-main zone is either group member or group server, the main zone cannot join this group.

fmunkes:
> Is the second remark related to the first? Im not sure what that means for the user?

Yeah, it's the same. It means, that the receiver display or app will show as media information "Music Assistant", and that doesn't change, no matter what is playing.

> Could you also check this table and advise what is supported from sync correction down? https://www.music-assistant.io/player-support/

Same as blue sound, though the stereo pair can be set up within the app, I never used that, I'll try to verify. If the table entry means, that you can select left and right within MA, then no for stereo pair.

> > If a non-main zone is either group member or group server, the main zone cannot join this group.

An AV receiver only has the ability to join the network with a single zone only (either main, or zone2, zone3, zone4). Now, if you use a zone (2,3,4), main or another zone cannot join. However, if you stream to main, the other zone can sync their input via "main_sync". That's technically not an MC link, but the UI will show this as a group member (which is quite nice...).

The revisiting part I am referring too is exactly that, and it's only a UI thing, the functionality is fine. So the UI offers you that main joins a synced zone, although it's not possible, it simply won't work.

OzGav:
Some SVG icons for you. If you can't easily download them I'll PR them in after the merge

icon
icon_monochrome

Working on the docs now. Standby for a draft for comment. On the grouping I believe I understand what you are saying but just wanted to confirm the behaviour is the same for all group types. I wondered if universal might be different...?

OzGav:
OK looking at this more. So are you saying it isn't possible to send metadata to the player? 

I believe yes the stereo pair support means you can send an individual channel to the player and then group them for stereo pair. I'll mark that as a X on the table

Draft: https://github.com/music-assistant/beta.music-assistant.io/pull/30

fmunkes:
> OK looking at this more. So are you saying it isn't possible to send metadata to the player?

~~It is possible to send metadata per file play request. But the players don't support enqueuing. Without queue flow mode I end up with an UnsupportedFeatureException, so apparently either you are able to enqueue, or you play in queue flow mode. I added the code for metadata, but the provider would need some kind of notification (e.g. another play_media call once a track is finished on a provider not supporting enqueuing) for this too work - let's see what Marcel thinks. I personally don't mind having Music Assistant on my display...~~
Forget this. It is possible. Enqueuing works. Metadata too. Some tweaks to be done, I'll do that tomorrow.

> I believe yes the stereo pair support means you can send an individual channel to the player and then group them for stereo pair. I'll mark that as a X on the table

Sounds good!

> Draft: [music-assistant/beta.music-assistant.io#30](https://github.com/music-assistant/beta.music-assistant.io/pull/30)

Thanks!

fmunkes:
I added enqueuing + metadata information
@OzGav -> I updated the docs above, 2 remarks gone and features enhanced with enqueuing, gapless playback and metadata.

I think the PR is ready now.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+
+    last_poll: float  # time.time
+    controlled_by_mass: bool
+    current_queue_item_id: str | None
+
+
+class MusicCast(PlayerProvider):
+    """MusicCast."""
+
+    musiccast_players: dict[str, MusicCastPlayer] = {}
```
marcelveldt, line 160: Be aware that you now have made these class variables so you can never make this provider multi instance without refactoring this. Just a heads up to keep in mind.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+            await avt_set_url(
+                self.mass.http_session, zone_player.physical_device, player_media=media
+            )
+            await avt_play(self.mass.http_session, zone_player.physical_device)
+
+            if ma_player := self._get_ma_player(player_id):
+                ma_player.current_media = media
+                await self.mass.players.register_or_update(ma_player)
+
+            # await self._cmd_run(player_id, zone_player.play_url, media.uri)
```
marcelveldt: leftover ?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+            case MusicCastPlayerState.IDLE | MusicCastPlayerState.OFF:
+                player.state = PlayerState.IDLE
+
+        # SOURCES
+        player.source_list = UniqueList([])
+        for source_id, source_name in device.source_mapping.items():
+            control = source_id in MC_CONTROL_SOURCE_IDS
+            passive = source_id in MC_PASSIVE_SOURCE_IDS
+            player.source_list.append(
+                # UI bug? I can't control my sources...
```
marcelveldt: wdym here ?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+
+        # GROUPING
+        # A zone cannot be synced to another zone or main of the same device.
+        # Additionally, a zone can only be synced, if main is currently not using any netusb
+        # function.
+        # For a Zone which will be synced to main, grouping emits a "main_sync" instead
+        # of a mc link.
+
+        # TODO: Revisit this - it produces the expected output (?), but the ui
+        # does not update correctly?
```
marcelveldt: Please discuss

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+            if _main_zone is None:
+                return
+            _other_zones = set(_main_zone.other_zones)
+
+            if device == _main_zone:
+                # a main zone cannot join another zone, but a zone can join
+                # main, see below.
+                can_group_with.difference_update(_other_zones)
+            elif device in _other_zones:
+                # can_group_with.difference_update(_other_zones.difference({device}))
```
marcelveldt: leftover?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+                # enforce a zone to be either the server, or sync to main of
+                # same device. makes life much easier.
+                can_group_with = {device, _main_zone}
+
+        # player.can_group_with = {
+        #     x.ma_player_id for x in can_group_with if x.ma_player_id is not None
+        # }
```
marcelveldt: leftover?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+
+        if len(device.musiccast_group) == 1:
+            if device.musiccast_group[0] == device:
+                # we are in a group with ourselves.
+                player.group_childs = UniqueList([])
+                player.synced_to = None
+                player.active_group = None
+
+        elif not device.is_client and not device.is_server:
+            player.group_childs = UniqueList([])
```
marcelveldt, line 758: I'm pretty sure we have a clear method on the list so you dont have to assign yourself

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+
+        # player.can_group_with = {
+        #     x.ma_player_id for x in can_group_with if x.ma_player_id is not None
+        # }
+        player.can_group_with = {self.instance_id}
+
+        if len(device.musiccast_group) == 1:
+            if device.musiccast_group[0] == device:
+                # we are in a group with ourselves.
+                player.group_childs = UniqueList([])
```
marcelveldt:
```suggestion
                player.group_childs.clear()
```

music_assistant/providers/musiccast/__init__.py
marcelveldt:
```suggestion
            player.group_childs.clear()
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+                player.active_group = None
+
+        elif not device.is_client and not device.is_server:
+            player.group_childs = UniqueList([])
+            player.synced_to = None
+            player.active_group = None
+
+        elif device.is_client:
+            _synced_to_id = self._get_player_id_from_mc_zone_player(device.group_server)
+            player.group_childs = UniqueList([])
```
marcelveldt:
```suggestion
            player.group_childs.clear()
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,790 @@
...
+            player.synced_to = _synced_to_id
+            player.active_group = _synced_to_id
+
+        elif device.is_server:
+            player.group_childs = UniqueList(
+                [self._get_player_id_from_mc_zone_player(x) for x in device.musiccast_group]
+            )
```
marcelveldt:
```suggestion
            player.group_childs.set(
                [self._get_player_id_from_mc_zone_player(x) for x in device.musiccast_group]
            )
```

music_assistant/providers/musiccast/avt_helpers.py
```diff
@@ -0,0 +1,223 @@
+"""Helpers to make an UPnP request.
+
+didl_lite helpers of MA didn't help, unfortunately...
+"""
```
marcelveldt: This has a lot of overlap with the existing DIDL metadata helpers. Lets try to unifi/merge here!

music_assistant/providers/musiccast/constants.py
```diff
@@ -0,0 +1,80 @@
+"""Constants."""
```
marcelveldt:
```suggestion
"""Constants for the MusicCast provider."""
```

music_assistant/providers/musiccast/constants.py
```diff
@@ -0,0 +1,80 @@
...
+
+from music_assistant.constants import (
+    CONF_ENTRY_ENABLE_ICY_METADATA,
+    CONF_ENTRY_FLOW_MODE,
```
marcelveldt: Do you want to even offer flow mode ? You can also choose to completely hide the option (so default disabled) if you think that all MusicCast players handle the enqueueing just fine

OzGav:
@fmunkes I have made changes to the docs that I think capture everything. I will add any player specific settings once I see it in action as I have a Yamaha amp

music_assistant/providers/musiccast/__init__.py
fmunkes: Thanks!

music_assistant/providers/musiccast/avt_helpers.py
fmunkes: Actually the comment I put there is obsolete, as I do use the didl_helper function from musicassistant/helpers now. At some point I didn't, but it turned out, that wasn't the cause of a problem I had while developing.

However, there is quite some overlap with the overall xml creation found in the sonos provider, so what I could do is move my helpers into the a upnp helper file (or better, rename the didl file to upnp_helper, and add these function, and then in a follow up PR other providers could be adjusted? Only issue is, I don't have sonos speaker, so I can't test that.

music_assistant/providers/musiccast/constants.py
fmunkes: Let's remove if for the time being - I did not encounter any issues with my speaker models. If user reports differ, we can still re-add it.

music_assistant/providers/musiccast/avt_helpers.py
marcelveldt: Sounds good. I can do the testing, no problem.

music_assistant/providers/musiccast/__init__.py
fmunkes: Sorry. the comment is outdated. It works as it should here. I'll remove the comment.

music_assistant/providers/musiccast/__init__.py
fmunkes: What I wanted to achieve, is, that MA knows about the limitation on zones given. But I somehow never got that right. I.e. if a zone plays, I'd like the UI to not show the main zone of the same device as sync possible. But somehow I failed. Maybe there is a bug in the code, but I don't see it. When I log the output of this, it correctly shows, that main cannot group with its zones, but a zone can group with main.

music_assistant/providers/musiccast/__init__.py
fmunkes: related to above.

music_assistant/providers/musiccast/__init__.py
fmunkes: related to above.

music_assistant/providers/musiccast/avt_helpers.py
fmunkes: Thanks, excellent.

fmunkes:
I added the suggestions. I renamed didl_lite.py to upnp.py (these are the dlna + sonos file changes, nothing else changed there). I moved my helpers into that file. Adjusting the other providers will be a separate PR.

Open is the comment on the 'can_group_with' - we can try figuring this out in this PR, or I will try adding it in another.

One additional note - in the requirements there are setuptools without version pinning. It is a dependency missing for aiomusiccast (which is pinned). Is that fine in this case?

marcelveldt:
> Open is the comment on the 'can_group_with'

Simply set it to your entire instance_id if all players can group with eachother. If that is not the case, you need to ensure it contains a list of player id's to which the player can group.

>  in the requirements there are setuptools without version pinning. It is a dependency missing for aiomusiccast (which is pinned). Is that fine in this case?

That is strange, why is that needed ?

fmunkes:
> Simply set it to your entire instance_id if all players can group with eachother. If that is not the case, you need to ensure it contains a list of player id's to which the player can group.

That's what I did, but it doesn't work as I'd like it too. But now I had the thought, it might be me misinterpreting. Is the "can_group_with" in both directions? Cause I am using it as "A cannot join B", but "B can join A". But it might be either both can join each other, or none

> 
> That is strange, why is that needed ?

Because of this... https://github.com/vigonotion/aiomusiccast/blob/main/aiomusiccast/__init__.py

marcelveldt:
> Is the "can_group_with" in both directions?

yes

> Because of this... https://github.com/vigonotion/aiomusiccast/blob/main/aiomusiccast/__init__.py

Ugh, they should add it as requirement then to their pyproject.
Anyways, fine to add it to your manifest for now until resolved but it probably wont accept a non versioned entry there so you may want to consider pinning a very low minimal version such as blahdependency>=1.0.0

fmunkes:
> > Is the "can_group_with" in both directions?
> 
> yes

Not a big deal, I removed the code then, and added some warning - users will have to refer to the docs then.

> Ugh, they should add it as requirement then to their pyproject. Anyways, fine to add it to your manifest for now until resolved but it probably wont accept a non versioned entry there so you may want to consider pinning a very low minimal version such as blahdependency>=1.0.0

I added that.
I guess all suggestions/ remarks are included now.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,776 @@
...
+        # aiomusiccast logs all fetch requests after udp message as debug.
+        # same approach as in upnp
+        if self.logger.isEnabledFor(VERBOSE_LOG_LEVEL):
+            logging.getLogger("aiomusiccast").setLevel(logging.DEBUG)
+        else:
+            logging.getLogger("aiomusiccast").setLevel(self.logger.level + 10)
+
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await super().loaded_in_mass()
```
marcelveldt:
```suggestion
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,776 @@
...
+        for mc_player in self.musiccast_players.values():
+            mc_player.physical_device.remove()
+
+    async def get_player_config_entries(
+        self,
+        player_id: str,
+    ) -> tuple[ConfigEntry, ...]:
+        """Return all (provider/player specific) Config Entries for the given player (if any)."""
+        base_entries = await super().get_player_config_entries(player_id)
+
```
marcelveldt:
```suggestion
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,776 @@
...
+        mc_player = self.musiccast_players.get(device_id)
+        if mc_player is None:
+            return None
+        return mc_player.physical_device.zone_devices.get(zone)
+
+    def _get_ma_player(self, player_id: str) -> Player | None:
+        for player in self.players:
+            if player.player_id == player_id:
+                return player
+        return None
```
marcelveldt: why not just do self.mass.players.get(player_id) here ?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,776 @@
...
+    async def _cmd_run(self, player_id: str, fun: Callable[..., Any], *args: Any) -> None:
+        """Help function for all player cmds."""
+        try:
+            await fun(*args)
+        except MusicCastConnectionException:
+            await self._set_player_unavailable(player_id)
+            self.logger.debug("Player became unavailable.")
+        except MusicCastGroupException:
+            # can happen, user shall try again.
+            ...
```
marcelveldt: You could consider turning this into a decorator instead. then just decorate the functions where you want to catch this

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,776 @@
...
+        mc_player = self.musiccast_players.get(device_id)
+        if mc_player is None:
+            return
+        mc_player.physical_device.remove()
+        for player in mc_player.get_all_players():
+            # disable zones as well.
+            player.available = False
+            await self.mass.players.register_or_update(player)
+
+    async def _cmd_run(self, player_id: str, fun: Callable[..., Any], *args: Any) -> None:
```
marcelveldt: I think the type hint for "fun" is wrong ? Shouldn't this be a coroutine function ?

music_assistant/providers/musiccast/__init__.py
fmunkes: changed. sometimes the obvious is still far away ;)

music_assistant/providers/musiccast/__init__.py
fmunkes: I (hopefully) fixed it - could you please double check.

music_assistant/providers/musiccast/__init__.py
fmunkes: Can I do this in a follow up PR? - for some functions the decorator would work right away, but for some others, the player id is not right away around, so I will have to refactor them a little.

fmunkes:
Oh no... - I just restarted my HA installation after updating, and had to realize, that the MA integration in HA doesn't load any more.

Logs have a lot of times `mashumaro.exceptions.InvalidFieldValue: Field "current_media" of type Optional[PlayerMedia] in Player has invalid value`, with the cause `Field "uri" of type str is missing in PlayerMedia instance`.

Now, in the model for the Player, comment and type hint do differ:
```
# current_media: return current active/loaded item on the player
# this may be a MA queue item, url, uri or some provider specific string
# includes metadata if supported by the provider/player
current_media: PlayerMedia | None = None
```
I was following the comment here, and assign queue items to current media in my update attributes. My question would be then, what the correct way is - on the MA side, there are no issues (at least, I didn't see any), but in the HA integration there are now

music_assistant/providers/musiccast/__init__.py
marcelveldt: yes, ofcourse. It was just a tip

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,767 @@
...
+        queue = self.mass.player_queues.get_active_queue(player.player_id)
+        if (
+            queue is not None
+            and queue.session_id is not None
+            and update_helper.current_queue_item_id is not None
+            and update_helper.controlled_by_mass
+        ):
+            if queue.current_item is not None:
+                if update_helper.current_queue_item_id == queue.current_item.queue_item_id:
+                    player.current_media = queue.current_item  # type: ignore[assignment]
```
marcelveldt: please dont do this

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,767 @@
...
+            if queue.next_item is not None:
+                if update_helper.current_queue_item_id == queue.next_item.queue_item_id:
+                    player.current_media = queue.next_item  # type: ignore[assignment]
```
marcelveldt: please dont do this

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,767 @@
...
+        else:
+            player.current_media = PlayerMedia(
+                uri=f"{player.player_id}_{device.source_id}",
+                title=device.media_title,
+                artist=device.media_artist,
+                album=device.media_album_name,
+                image_url=device.media_image_url,
+            )
```
marcelveldt: tip: use the helper instead to set player_media

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,767 @@
...
+
+        # SOURCE
+        if device.source_id == "server" and update_helper.controlled_by_mass:
+            player.active_source = queue.queue_id if queue is not None else None
```
marcelveldt: this is not good. what if the player is playing a group ? Then the active queue is not the player itself but that of the group player

music_assistant/providers/musiccast/__init__.py
marcelveldt: You shouldnt hav eto interact with the queue here directly as long as you can set the current queue item id on the playermedia object.

So something like:

```
mass_player.set_current_media(uri=update_helper.url, queue_item_id=update_helper.current_queue_item_id)
```

music_assistant/providers/musiccast/__init__.py
marcelveldt: queue item != PlayerMedia so dont assign like this. 
Use the set_current_media helper to just update the current url and the queue_item_id and you're all set, nothing else needed

music_assistant/providers/musiccast/__init__.py
fmunkes: Fixed. Thanks. I use the helper now, which I missed embarrassingly.

music_assistant/providers/musiccast/__init__.py
fmunkes: Fixed, like above.

music_assistant/providers/musiccast/__init__.py
fmunkes: Done.

music_assistant/providers/musiccast/__init__.py
fmunkes: While I fixed your comment, I'm still trying to figure out a reliable way if a player is currently controlled by MA or not.

fmunkes: Ok. Found a solution which I would argue works in 99% of all use cases. I commented in the code, the edge case is the internal dlna function of the device - if that is used, checking the source id is not enough, and I have to fallback to the media title.

I also took the opportunity to check in the player commands, if we are MA controlled, and then use the UPnP function, and the device function otherwise. This avoids starting the internal server item (not the MA item), which might still be device cached, as I have learned.

Thank you for being patient with me (again...). Your review brought up many pitfalls I didn't see by myself at all.

music_assistant/providers/musiccast/musiccast.py
```diff
@@ -0,0 +1,602 @@
...
+        for inp in self.zone_data.input_list:
+            label = self.device.data.input_names.get(inp, "")
+            if inp != label and (
+                label in self.zone_data.input_list
+                or list(self.device.data.input_names.values()).count(label) > 1
+            ):
+                label += f" ({inp})"
+            if label == "":
+                label = inp
+            ret[inp] = label
```
marcelveldt: Just a tip for later. In python we actually prefer descriptive variable names.
So instead of "inp" better use "input".

fmunkes: Ah, that's one of the copied over functions. If you give me a minute, I'll change that right away.

fmunkes: done.

fmunkes:
One additional question:
Is there any way in my code how the `queue.current_item` can be none if I start playing an item? Cause sometimes (I thought it's a lock issue, but it's not helping) queue.current_item is None, despite the media playing, sometimes not - and then my "controlled_by_mass" logic doesn't work. Seems to happen more frequently with audiobooks, than with tracks.

```python
# play_media function
# async with lock didn't get me further, though it's probably a good idea nevertheless.
# interesting bit of the function:

await avt_play(self.mass.http_session, zone_player.physical_device)
if ma_player := self.mass.players.get(player_id):
    ma_player.current_media = media
    await self.mass.players.register_or_update(ma_player)
```

marcelveldt:
> Is there any way in my code how the queue.current_item can be none if I start playing an item?

That is a chicken and the egg situation. You get a call to play_media which initiates the playback and then you report what is playing on the "current_media" so the queue knows that as well and then sets the "current item".
So its the other way around: YOU (as player provider) need to provide the info what is active, not the other way around.

A common way to detect what is playing, is by looking at the uri/url that is loaded in the player.

music_assistant/providers/musiccast/constants.py
```diff
@@ -0,0 +1,80 @@
...
+    CONF_ENTRY_OUTPUT_CODEC,
+    create_sample_rates_config_entry,
+)
+
+# Constants for players
+# both the http profile and icy didn't matter for me testing it.
+PLAYER_CONFIG_ENTRIES = (
+    CONF_ENTRY_OUTPUT_CODEC,
+    CONF_ENTRY_HTTP_PROFILE_DEFAULT_2,
+    CONF_ENTRY_ENABLE_ICY_METADATA,
```
marcelveldt: If the MusicCast devices do not support ICY metadata at all (only used with flow mode) then maybe we need to create a hidden variant of that setting, just like we do for http profile ? That is better than offering settings that do not make any impact

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+
+        player.volume_level = int(
+            zone_data.current_volume / (zone_data.max_volume - zone_data.min_volume) * 100
+        )
+        player.volume_muted = zone_data.mute
+
+        # STATE
+        # we can only use one zone at a time for server playback
+        player.elapsed_time = None
+        player.elapsed_time_last_updated = None
```
marcelveldt: To prevent weird race conditions, better store this in a temp variable and then after all the checks below, assign the final value to the player.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                player.state = PlayerState.PAUSED
+            case MusicCastPlayerState.PLAYING:
+                player.state = PlayerState.PLAYING
+                player.elapsed_time = device.media_position
+                player.elapsed_time_last_updated = device.media_position_updated_at
+            case MusicCastPlayerState.IDLE | MusicCastPlayerState.OFF:
+                player.state = PlayerState.IDLE
+
+        # SOURCES
+        player.source_list = UniqueList([])
```
marcelveldt:
```suggestion
```
this is already a uniquelist by default, no need to reassign a new list

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+        for source_id, source_name in device.source_mapping.items():
```
marcelveldt, line 678:
```suggestion
        source_list: list[PlayerSource] = []
        for source_id, source_name in device.source_mapping.items():
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+            control = source_id in MC_CONTROL_SOURCE_IDS
+            passive = source_id in MC_PASSIVE_SOURCE_IDS
+            player.source_list.append(
```
marcelveldt:
```suggestion
           source_list.append(
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                PlayerSource(
+                    id=source_id,
+                    name=source_name,
+                    passive=passive,
+                    can_play_pause=control,
+                    can_seek=False,
+                    can_next_previous=control,
+                )
+            )
```
marcelveldt, line 690:
```suggestion
            )
        player.source_list.set(source_list)
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+            # i.e. we are not a group member.
+            # the device's source id is server, if controlled by upnp, but also, if the internal
+            # dlna function of the device are used. As a fallback, we then
+            # use the item's title. This can only fail, if our current and next item
+            # has the same name as the external.
+            controlled_by_mass = False
+            if (
+                queue is not None
+                and queue.current_item is not None
+                and queue.current_item.media_item is not None
```
marcelveldt: You need to come up with a better check, this will lead to issues.
Cant you just check the media that is loaded ?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                    can_play_pause=control,
+                    can_seek=False,
+                    can_next_previous=control,
+                )
+            )
+
+        # UPDATE UPNP HELPER
+        update_helper = self.upnp_update_helper.get(player.player_id)
+        now = time.time()
+        queue = self.mass.player_queues.get_active_queue(player.player_id)
```
marcelveldt:
```suggestion
```
please don't interact with the queue yourself - that is a bad pattern.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                    _item_ids.append(queue.next_item.queue_item_id)
+                    _names.append(queue.next_item.media_item.name)
+                controlled_by_mass = (
+                    player.player_id in _player_current_url
+                    and _queue_item_id in _item_ids
+                    and device.source_id == "server"
+                    # this last one might fail in the rare case, that the external controller
+                    # is using the same title and its part of current/ next of our queue
+                    and device.media_title in _names
+                )
```
marcelveldt:
```suggestion
           controlled_by_mass = player.player_id in _player_current_url and self.mass.streams.base_url in _player_current_url
```

all this needs to be removed, you need to tell the queue controller what is playing - not the other way around!

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+            self.upnp_update_helper[player.player_id] = update_helper
+
+        # UPDATE PLAYBACK INFORMATION
+        if (
+            update_helper.current_queue_item_id is not None
+            and update_helper.current_uri is not None
+            and update_helper.controlled_by_mass
+        ):
+            player.set_current_media(
+                uri=update_helper.current_uri, queue_item_id=update_helper.current_queue_item_id
```
marcelveldt: If you don't have access to current_queue_item_id from the player's own metadata, then just only set the current URL is sufficient. The queue will figure it out on its own in that case based on the url.

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                and _server_update_helper.current_uri is not None
+                and _server_update_helper.controlled_by_mass
+            ):
+                player.set_current_media(
+                    uri=_server_update_helper.current_uri,
+                    queue_item_id=_server_update_helper.current_queue_item_id,
+                )
+            else:
+                player.set_current_media(
+                    uri=f"{_server_id}_{_server.source_id}",
```
marcelveldt, line 751: is this correct ? what if e.g. Spotify is playing , will this be set to the spotify URI ?
Aren't you mixing up player source and media ?

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+                image_url=device.media_image_url,
+            )
+
+        # SOURCE
+        player.active_source = device.source_id  # fallback
+        if update_helper.controlled_by_mass and queue is not None:
+            # controlled by ma means directly controlled, not group member
+            player.active_source = queue.queue_id
+        elif device.is_client:
+            _server_id = self._get_player_id_from_mc_zone_player(device.group_server)
```
marcelveldt:
```suggestion
        
        if device.is_client and not update_helper.controlled_by_mass:
            player.active_source = device.source_id  # fallback
            _server_id = self._get_player_id_from_mc_zone_player(device.group_server)
```

music_assistant/providers/musiccast/__init__.py
```diff
@@ -0,0 +1,840 @@
...
+            _server_update_helper = self.upnp_update_helper.get(_server_id)
+            _server_queue = self.mass.player_queues.get_active_queue(_server_id)
+            if (
+                _server_update_helper is not None
+                and _server_queue is not None
+                and _server_update_helper.controlled_by_mass
+            ):
+                player.active_source = _server_queue.queue_id
```
marcelveldt:
```suggestion
```
You can leave all this out, active_source will be set for you. If you can not figure it out, just set it to None

music_assistant/providers/musiccast/__init__.py
fmunkes: Understood. What I am doing now is retrieving the upnp xml title from the device. If the device is controlled by its own upnp server access function, that xml is not updated, so the device media title and this title differ. There is the edge case of both titles being the same, but that's something I could live with.
On top, the source must be server.

music_assistant/providers/musiccast/constants.py
fmunkes: I didn't see that being supported on my devices, and flow mode is disabled anyways, so I'll create such a config value.

fmunkes:
> > Is there any way in my code how the queue.current_item can be none if I start playing an item?
> 
> That is a chicken and the egg situation. You get a call to play_media which initiates the playback and then you report what is playing on the "current_media" so the queue knows that as well and then sets the "current item". So its the other way around: YOU (as player provider) need to provide the info what is active, not the other way around.
> 
> A common way to detect what is playing, is by looking at the uri/url that is loaded in the player.

I understand now, where all my trouble did come from. Thank you. Took me a while!

music_assistant/providers/musiccast/__init__.py
fmunkes: I use both, thanks to the didl data.

music_assistant/providers/musiccast/__init__.py
fmunkes: since the device.media_position and media_position_update_at can be both be non, I just directly assign them now. I hope that's fine.

music_assistant/providers/musiccast/__init__.py
fmunkes: Hmm, this is only done, if the player is externally controlled. I don't see a way, how I can access what is then playing in terms of a uri, but I have title etc information. I.e. I'm using it to show playback information (e.g. a net radio station), even though it is not MA controlling the player - I wanted to use a uri which is definitely not part of the MA lib.

music_assistant/providers/musiccast/__init__.py
fmunkes: Ok. The title one is not good enough, maybe because escaping and not exactly same chars.. Let's keep source is server - if the user uses the device itself to access a dlna server, then the xml will not be updated, and if ma was the last upnp player, the progress bar is advanced. That is an edge case, but I think it's acceptable, nothing happens to the queue...

music_assistant/providers/musiccast/__init__.py
fmunkes: And remove the queue item id too now. Only uri left - it worked for all media items i tested this against, except radio.

music_assistant/providers/musiccast/__init__.py
marcelveldt: Ah OK, in some other providers (such as sonos or cast) the actual URI is also provided, such as the spotify uri.
But if that info is absent then so be it. At least you have some info like a title and image and such.

marcelveldt:
One final comment, you note "HA integration and MA integration can be used alongside" but that should actually be discouraged as it is another connection to the device and it will cause duplicate entities. So while it does work, we should maybe inform that there is no need to do that.

Just a nitpick.

marcelveldt (approved):
Looks good, amazing work again @fmunkes !

darthdurden:
This is great! I have been waiting for this to start using Music Assistant! Set it up this morning and it works great! I did notice 1 bug that is giving me issues:

When I set up a playlist, the track listing on the UI never changes past the original song during playback (unless I hit the "next" button. And sometimes if I pause and play it again, it'll start back at that old song. It is like it is not recognizing when a song ends and just thinks you're play the same song again even when it does actually continue to the next song. If I play things to the local device (a tablet in this case) it works fine, so I think this is likely the new provider.

plimptm:
cheers @fmunkes and thanks for this contribution! Was just combing through the release notes for 2.6.0 to see if anything related to MusicCast was included...this is a really big improvement.

PR #2144, opened 2025-04-24
tidal: allow audio quality switching after authentication

jozefKruszynski:
Allows a user to switch between Lossless and Hi Res audio qualities after authentication. Also made the drop down options a bit prettier

### DIFF
--- a/music_assistant/providers/tidal/__init__.py
+++ b/music_assistant/providers/tidal/__init__.py
@@ -8,7 +8,6 @@ import json
 from collections.abc import Awaitable, Callable
 from contextlib import suppress
 from datetime import datetime
-from enum import StrEnum
 from typing import TYPE_CHECKING, Any, TypeVar, cast
 
 from aiohttp import ClientConnectionError, ClientResponse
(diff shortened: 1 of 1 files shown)


music_assistant/providers/tidal/__init__.py
```diff
@@ -180,13 +188,13 @@ async def get_config_entries(
...
                 type=ConfigEntryType.STRING,
-                label=CONF_QUALITY,
-                required=True,
-                hidden=True,
-                value=cast("str", values.get(CONF_QUALITY) or TidalQualityEnum.HI_RES.value),
-                default_value=cast(
-                    "str", values.get(CONF_QUALITY) or TidalQualityEnum.HI_RES.value
-                ),
+                label="Quality setting for Tidal:",
+                description="HIGH_LOSSLESS = 16bit 44.1kHz, HI_RES = Up to 24bit 192kHz",
```
marcelveldt: I don't like the fact that you are using a technical term for the option title. Please make that a user friendly name.

jozefKruszynski: You mean in the description?

jozefKruszynski: I'll switch out to what Tidal use, as that makes more sense, I think I haven't changed them since the early days, and never thought much of it.

jozefKruszynski: How's this?

image

image

jozefKruszynski: Removed the unnecessary enum to simplify

marcelveldt (approved):

PR #2145, opened 2025-04-24
Several improvements to announcements

chatziko:
The generic announcement implementation is particularly glitchy for snapcast players cause pause/resume is far from smooth (this is known and mentioned in the doc). This PR adds a native announcement implementation which provides a much better experience. The idea is pretty simple:
- Use a separete snapcast stream for announcements
- Switch the snapcast group to it while the announcement is playing
- Swtch back afterwards

This is as close to "ducking" as we can easily get, the announcement plays over the music (unfortunately __not__ mixed) and the playback is not interrupted at all.

#### Stream names

The current implementation uses stream names that are derived from the __queue__ name. This has the following issues:

- It is buggy when two players use the same queue simultaneously (eg in universal groups). See https://github.com/music-assistant/support/issues/3922
- It makes it harder to find which stream correspond to which player, if we need to delete a stream while it is not active.

So I opted for a cleaner solution:
- each player has at most two streams active: one for music one for announcements
- the stream names are based on the player name (not the queue name)

It's a bit less pretty, but the player and queue names very often coincide, and stream names are only visible in snapweb anyway, nobody really sees them.

#### Volume level

Unfortunately snapcast sets a volume per client (not per stream), so we need a way to set the announcement volume without affecting the music volume. I opted for the simplest solution: save the previous volume, change it, restore later.

This has the downside that the temporary change will be visible in the UI. We could hide it (eg by hiding snapcast volume updates from the player controller), but it's probably not worth the extra complexity.

#### Other notes

- The changes are quite simple actually, mainly adding a `play_announcement` method, and refactor some code that is common with `play_media`.

- Interacting with the player (pause, play, etc) during announcement is possible. Eg: an announcement starts when idle then
  the user presses play. Playback will start, and the snapcast group will switch to the music steam when the announcement is over.

- The control script (`control.py`) is not used at all for announcement streams.

- I deleted a couple of lines that seemed obsolete to me, in separate commits for easier review

### DIFF
--- a/music_assistant/controllers/streams.py
+++ b/music_assistant/controllers/streams.py
@@ -919,25 +919,39 @@ class StreamsController(CoreController):
         use_pre_announce: bool = False,
     ) -> AsyncGenerator[bytes, None]:
         """Get the special announcement stream."""
+        filter_params = ["loudnorm=I=-10:LRA=11:TP=-2"]
+
+        if use_pre_announce:
+            # Note: TTS URLs might take a while to load cause the actual data are often generated
+            # asynchronously by the TTS provider. If we ask ffmpeg to mix the pre-announce, it will
+            # wait until it reads the TTS data, so the whole stream will be delayed. It is much
+            # faster to first play the pre-announce using a separate ffmpeg stream, and only
+            # afterwards play the TTS itself.
+            #
+            # For this to be effective the player itself needs to be able to start playback fast.
+            # If the returned stream is used as input to ffmpeg we should pass -probesize 8096.
+            #
+            # Finally we also need to make sure we don't make other blocking requests to the TTS
+            # data, eg to get the duration (async_parse_tags).
+            #
+            async for chunk in get_ffmpeg_stream(
+                audio_input=ANNOUNCE_ALERT_FILE,
+                input_format=AudioFormat(content_type=ContentType.try_parse(ANNOUNCE_ALERT_FILE)),
+                output_format=output_format,
+                filter_params=filter_params,
+            ):
+                yield chunk
+
         # work out output format/details
         fmt = announcement_url.rsplit(".")[-1]
         audio_format = AudioFormat(content_type=ContentType.try_parse(fmt))
-        extra_args = []
-        filter_params = ["loudnorm=I=-10:LRA=11:TP=-2"]
-        if use_pre_announce:
-            extra_args += [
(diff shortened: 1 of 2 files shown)


marcelveldt (changes requested):
While I do (highly) appreciate the code quality improvements of this PR I do not get the implementation of switching to a separate stream for the announcement. Switching a stream should be something like a second or so, versus missing an entire block of music because its playing in the void. Maybe something else is going on on a higher level why this doesn't work as you want/expect ?

With only little effort it would be possible to mix the announcement into the pcm stream sent to the snapcast TCP port, that would be a much better solution. Setting up a second stream is NOT a solution imo.

So, let's first have a discussion about this on discord and discuss the issue you are trying to solve here.

chatziko:
Ok, just opened a post in discord.

music_assistant/controllers/streams.py
```diff
@@ -919,25 +919,39 @@ async def get_announcement_stream(
...
+            # data, eg to get the duration (async_parse_tags).
+            #
+            async for chunk in get_ffmpeg_stream(
+                audio_input=ANNOUNCE_ALERT_FILE,
+                input_format=AudioFormat(content_type=ContentType.try_parse(ANNOUNCE_ALERT_FILE)),
+                output_format=output_format,
+                filter_params=filter_params,
+            ):
+                yield chunk
+
```
marcelveldt, line 944: This will not work in all cases because a player might detect EOF and stop playing. Did you try this with a couple of different players ? Especially players that get the url themselves ?

marcelveldt:
As discussed on discord, we are going to merge this so at least we have an improvement over the current implementation and then iterate from there. Doing the announcement streaming mixing is not as straight forward as we think and need more discussion.

music_assistant/controllers/streams.py
chatziko: Not really, let me test before merging and/or think alternative solutions.

I tried to do it within ffmpeg, that is convince it to start merging A and B before B arrives, but it might be impossible before reading even a single byte of B.

marcelveldt: I did some base tests with a few players and it seems to work just fine. So let's keep it as-is and change only if we get issue reports

marcelveldt (approved):
Nice work @chatziko !

PR #2149, opened 2025-04-25
Add debug message about concatenating announcement stream

chatziko:
Add a debug message that might help some poor developer in the future understand why TTS does not work for their player.

PS. the import reorganization was made by pre-commit ruff (nothing to do with the actual change)

### DIFF
--- a/music_assistant/controllers/streams.py
+++ b/music_assistant/controllers/streams.py
@@ -925,9 +925,9 @@ class StreamsController(CoreController):
             # For this to be effective the player itself needs to be able to start playback fast.
             # If the returned stream is used as input to ffmpeg we should pass -probesize 8096.
             #
-            # Finally we also need to make sure we don't make other blocking requests to the TTS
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/streams.py
```diff
@@ -928,6 +928,12 @@ async def get_announcement_stream(
             # Finally we also need to make sure we don't make other blocking requests to the TTS
             # data, eg to get the duration (async_parse_tags).
             #
+            if not output_format.codec_type.is_pcm():
+                self.logger.debug(
+                    "Concatenating non-pcm TTS and pre-announce streams, so far players seem to "
+                    "tolerate this, but it might break some player in the future"
+                )
+
```
marcelveldt: I would just put this as comment in the code and not debug message

chatziko: Done.

marcelveldt (approved):
Thanks, @chatziko !

PR #2150, opened 2025-04-26
Support for Apple Music authentication in web UI

maxlyth:
This is my first OSS PR so not sure of the process but this is my first draft updates to the Apple Music Provider that delivers a UI to retrieve the music_user_token. I can't confirm it is fully working as I am missing the widevine binaries for my dev env which are required to complete the provider setup.

### DIFF
--- a/music_assistant/providers/apple_music/__init__.py
+++ b/music_assistant/providers/apple_music/__init__.py
@@ -23,6 +23,7 @@ from music_assistant_models.enums import (
     StreamType,
 )
 from music_assistant_models.errors import (
+    LoginFailed,
     MediaNotFoundError,
     MusicAssistantError,
     ResourceTemporarilyUnavailable,
@@ -113,45 +114,53 @@ async def get_config_entries(
     # Check for valid app token (1st with regex and then API check) otherwise display a config field
     default_app_token_valid = False
     async with (
-        mass.http_session.get("https://api.music.apple.com/v1/test",
+        mass.http_session.get(
+            "https://api.music.apple.com/v1/test",
             headers={"Authorization": f"Bearer {MUSIC_APP_TOKEN}"},
-            ssl=True, timeout=10) as response,
+            ssl=True,
+            timeout=10,
+        ) as response,
     ):
         if response.status == 200:
             values[CONF_MUSIC_APP_TOKEN] = f"{MUSIC_APP_TOKEN}"
             default_app_token_valid = True
 
-
     # Action is to launch MusicKit flow
     if action == "CONF_ACTION_AUTH":
         # TODO: check the developer token is valid otherwise user is going to have bad experience
         async with AuthenticationHelper(mass, values["session_id"]) as auth_helper:
-            flow_base_url = f"/apple_music_auth/{values["session_id"]}/"
+            flow_base_url = f"/apple_music_auth/{values['session_id']}/"
             flow_timeout = 600
             parent_file_path = pathlib.Path(__file__).parent.resolve()
+
             async def serve_mk_auth_page(request: web.Request) -> web.Response:
                 auth_html_path = parent_file_path.joinpath("musickit_auth/musickit_wrapper.html")
                 return web.FileResponse(auth_html_path, headers={"content-type": "text/html"})
+
             async def serve_mk_auth_css(request: web.Request) -> web.Response:
                 auth_css_path = parent_file_path.joinpath("musickit_auth/musickit_wrapper.css")
                 return web.FileResponse(auth_css_path, headers={"content-type": "text/css"})
+
             async def serve_mk_glue(request: web.Request) -> web.Response:
                 return_html = f"const app_token='{values[CONF_MUSIC_APP_TOKEN]}';"
                 return_html += f"const user_token='{values[CONF_MUSIC_USER_TOKEN]}';"
                 return_html += f"const return_url='{auth_helper.callback_url}';"
                 return_html += f"const flow_timeout={flow_timeout - 10};"
-                return_html += "const mass_buid='2025.1.1';"
+                return_html += f"const mass_buid='{mass.version}';"
                 return web.Response(body=return_html, headers={"content-type": "text/javascript"})
 
             mass.webserver.register_dynamic_route(flow_base_url + "index.html", serve_mk_auth_page)
             mass.webserver.register_dynamic_route(flow_base_url + "index.css", serve_mk_auth_css)
             mass.webserver.register_dynamic_route(flow_base_url + "index.js", serve_mk_glue)
             try:
-                values[CONF_MUSIC_USER_TOKEN] = (await auth_helper.authenticate(flow_base_url
-                    + "index.html", flow_timeout))["music-user-token"]
+                values[CONF_MUSIC_USER_TOKEN] = (
+                    await auth_helper.authenticate(flow_base_url + "index.html", flow_timeout)
+                )["music-user-token"]
+            except KeyError:
+                # no music-user-token URL param was found so user probably cancelled the auth
+                pass
             except Exception as error:
-                # TODO: No logger instance available here
-                mass.logger.error("Authentication Helper failed: %s", error)
+                raise LoginFailed(f"Failed to authenticate with Apple '{error}'.")
             finally:
                 mass.webserver.unregister_dynamic_route(flow_base_url + "index.html")
                 mass.webserver.unregister_dynamic_route(flow_base_url + "index.css")
@@ -178,8 +187,6 @@ async def get_config_entries(
             value=values.get(CONF_MUSIC_USER_TOKEN) if values else None,
         ),
     )
-
-
 
 
 class AppleMusicProvider(MusicProvider):
--- a/music_assistant/providers/apple_music/__init__.py
+++ b/music_assistant/providers/apple_music/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import json
 import os
+import re
 from typing import TYPE_CHECKING, Any
 
 import aiofiles
@@ -104,6 +105,12 @@ async def get_config_entries(
     values: the (intermediate) raw values for config entries sent with the action.
     """
 
+    def validate_token(token):
+        if not isinstance(token, str):
+            return False
+        valid = re.findall(r'[a-zA-Z0-9=/+]{32,}==$', token)
+        return bool(valid)
+
     # Check if we have a developer token otherwise display a config field for an override
     default_app_token_valid = False
     async with (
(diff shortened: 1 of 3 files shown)


maxlyth:
This is my first OSS PR so not sure of the process but this is my first draft updates to the Apple Music Provider that delivers a UI to retreive the music_user_token. I can't confirm it is fully working as I am missing the widevine binaries which are required to complete the provider setup

maxlyth:
I'm not loving the icon in the Access Request screen but I believe that is linked to Bundle ID in the Apple developer account that owns the developer token. I don't want to touch anything related to that token while it is used for production code. 
Screenshot 2025-04-26 at 06 47 36
I wrote this code while the developer token was broken, hence the code to test its validity and optionally show a config field to enter a local developer token.

There is nothing secret in these tokens other than the connection to an issuing entity so the app_var obfuscation is overkill. I still think these should be hosted rather than baked into code with a shorter expiry. That would actually improve secuirty and allow them to be rolled if abused.

maxlyth:
I know I will forget this edge case so I want to mention that maybe the route `./callback/*` should always serve a page with `window.close()` even if the specific authentication helper has expired. 

The service redirecting to the helper url cannot know it has become invalid. I realise this change that could impact other consumers of auth helper but could be made optionally safe with a url param such as `&autoclose`

This PR is for AM Provider only but as that workflow requires a new browser window this suggestion for the auth helper would help clean up error conditions

music_assistant/providers/apple_music/__init__.py
```diff
@@ -97,21 +103,143 @@ async def get_config_entries(
...
+        # TODO: check the developer token is valid othwerwise user is going to have bad experience
+        async with AuthenticationHelper(mass, values["session_id"]) as auth_helper:
+            mass.webserver.register_dynamic_route(f"/apple_music_auth/", serve_musickit_auth_page)
+            url = f"/apple_music_auth/?redirect={auth_helper.callback_url}&token={values[CONF_DEVELOPER_TOKEN]}"
+            try:
+                values[CONF_MUSIC_USER_TOKEN] = (await auth_helper.authenticate(url))["music-user-token"]
+            except Exception as error:
+                # TODO: No logger instance available here
+                ("Authentication Helper failed: %s", error)
+            mass.webserver.unregister_dynamic_route(f"/apple_music_auth/")
```
marcelveldt:
```suggestion
            finally:
                mass.webserver.unregister_dynamic_route(f"/apple_music_auth/")
```

music_assistant/providers/apple_music/__init__.py
```diff
@@ -97,21 +103,143 @@ async def get_config_entries(
...
+            action="CONF_ACTION_AUTH",
+            description="You need to authenticate on Apple Music.",
+            action_label="Authenticate with Apple Music",
+            value=values.get(CONF_MUSIC_USER_TOKEN) if values else None,
         ),
     )
 
 
+async def serve_musickit_auth_page(request: web.Request) -> web.Response:
+    # params = dict(request.query)
```
marcelveldt: I would simply place an html file within your provider code and then return the web filehandler here.

It should be something like 
```
auth_html_path = pathlib.Path(__file__).parent.resolve().joinpath("auth.html")
return web.Fileresponse(auth_html_path, content_type="text/html")
```

or something along those lines

maxlyth:
I too did not like mixing the HTML/JS in with Python but this is a draft PR to check I had not breached any major foundations. I have a new version where the HTML is more dynamic so as not to pollute the browser history and also ensure the MusicKit tokens are only held in ephemeral code rather than passed as persisted URL params.

This version is also missing the ability to sign-out of MusicKit. So far I have deduced that the current AppleAccount is persisted by multiple keys in browser localStorage but there is no documentation. I need some more time for code decompiling and trial and error.

music_assistant/providers/apple_music/__init__.py
maxlyth: I too did not like mixing the HTML/JS in with Python but this is a draft PR to check I had not breached any major foundations. I have a new version where the HTML is more dynamic so as not to pollute the browser history and also ensure the MusicKit tokens are only held in ephemeral code rather than passed as persisted URL params.

This version is also missing the ability to sign-out of MusicKit. So far I have deduced that the current AppleAccount is persisted by multiple keys in browser localStorage but there is no documentation. I need some more time for code decompiling and trial and error.

maxlyth:
The MusicKit workflow still needs an interstitial window for return redirect but now that window has a fully functioning UI to refresh token, switch accounts and clean-up helper data from browser local storage. 

You will need to delete your current token in provider to see anything and it is likely that you also have browser cached creds so also will need to choose option to sign out from Apple Music to see the full workflow.

Given the app token is semi-private and shared for all MA users I now pass this to the helper without a url param so it does not get recorded in browser history. All UI is 100% local resources with the exception of MusicKit from Apple CDN 

The user token refresh should be working but is hard to test as token is proprietary. Also refresh requires manual intervention.

maxlyth:
> I'm not loving the icon in the Access Request screen but I believe that is linked to Bundle ID in the Apple developer account that owns the developer token.

I think I have found a way to fix the ugly generic icon in the second stage of the auth flow. However even for first stage icon I could not find a canonical MA icon in either the frontend or server repos and ended lifting one from the music-assistant.io code base. Another disconnected repo so I base64 encoded the image in my HTML as I could not be certain the URL would not change.

The `mass.webserver.register_dynamic_route()` feature is cool on a Provider level but could there not be a static assets route for shared resources. Obviously canonical MA artwork but maybe also shared css for authentication helpers so there is stylistic convergence.

An augument could be made to do this in the front end repo but that needs a double PR for an provider auth UI update. My opinion is Provider auth code should stay in the server repo. eg the Spotifiy helper should be in mass-server, not the documentation repo.

Personally I only use the AppleMusic Provider so don't know what lurks beneath in the other providers but I might be semi-volunteering to refactor any that use external resources so they run 100% local if anyone wants me to do that and points me to the dragons.

maxlyth:
I realise there are `# TODO:` items in code base around error handling I can't force and edge cases. I am also missing translations and documentation. However I am hoping this initial release acts as a proof of concept and provides a reliable solution for 90%+ of users.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -97,21 +103,90 @@ async def get_config_entries(
...
+                # TODO: No logger instance available here
+                mass.logger.error("Authentication Helper failed: %s", error)
+            finally:
+                mass.webserver.unregister_dynamic_route(flow_base_url + "index.html")
+                mass.webserver.unregister_dynamic_route(flow_base_url + "index.css")
+                mass.webserver.unregister_dynamic_route(flow_base_url + "index.js")
+
     # ruff: noqa: ARG001
     return (
         ConfigEntry(
```
MarvinSchenkel, line 171: Why do we need a separate user config for the MUSIC_APP_TOKEN? This is just the developer token, right? Can't we just inject that into the html directly based on the value coming from the app_var?

maxlyth: Marcel wanted me to separate the HTML from the Python code and serve it as a static file. This made sense as VSCode could lint out any typos but did mean I could no longer use Python string substitutions to make dynamic HTML. 

I did not want to define some secret substitution format so I could parse and the HTML and inject the tokens as it is too easy to break in the future. I also did not want to pass token as URL params as that records them in client browser history.

Instead I write params into a super simple .js file that defines them as variables and make the core  HTML simply includes that js file thereby importing the dynamic data that needs to be passed from Python. I think this is secure and far more durable.

marcelveldt: 2 different discussions going on here. I believe @MarvinSchenkel just asks why it needs to be user configurable ? Do we want the app token to even be configurable or just always use the MA one ?

maxlyth: If you are questioning why the App token is configurable then I could say why not? The user should never see this config entry as it is only revealed if the App Token (sorry renamed that from dev token to reflect its true low status) fails the API Bearer token API test. 

However if the centrally issued token is revoked or the user has not (or cannot) updated MA for 6 moonths to get a fresh token then they are presented the option to enter a token as a config entry. I needed this code and tested that it worked pre 2.5.2 so I could continue development with a seperately issued token. I don't think this is a bad option to have and could save peoples bacon at edge cases.

maxlyth: Personally I think MA should issue App tokens with a shorter expiry and regenerate them nightly with a GitAction cron automation. The MA Apple Music Provider can then periodically fetch a fresh one. This would allow MA to revoke a token if is was ever abused without risking a strike from Apple.

maxlyth:
I think I have a fix to stop lint from rejecting commit. At the same time I also wrote a fairly large chunk of code to handle a corner case where the user closes the interstitial window but the auth help panel is left open with no way to clean up the mess without waiting for long timeout or refresh page.

Some more control over the auth helper would be great but for now a fix is to intercept the interstitial page unload and then make am empty GET to the helper as the interstitial is torn down by the browser. It works in Safari/Chrome but more control over the auth helper would negate the need for this hack. Is there a code owner for the helper?

music_assistant/providers/apple_music/__init__.py
marcelveldt: Yes, that is an excellent idea but difficult in a Open Source scenario. Are we going to rely on cloud then to provide the token ? How would that work ? Now we have the freedom to create a new apptoken for each release and we have no dependency on a cloud service. I'm open to suggestions here but this is a bit difficult

marcelveldt:
the helper doesnt have a specific code owner, its a team effort and we improve it when needed

marcelveldt:
Just gave it a quick test and it works awesome, well done!

marcelveldt (approved):
Nice work @maxlyth !

marcelveldt:
Actually... its not working behing a reverse proxy (such as Home Assistant Ingress) - it only works when directly connecting to the (insecure) webserver at port 8095.

EDIT: fixed... it was just a leading slash issue. All good now

maxlyth:
This is broken for me using the nightly HA Dev Add-on. There are a lot of browser redirects in this integration so likely something to do with HA ingress and host resolution that did not show up in non-proxied dev testing.
For me it fails at the hand back to the auth helper so the music-user-token is not populated and there is no way for the user to manually provide this config. This is my first OSS PR so not sure what the process is to re-open/rollback this PR

MarvinSchenkel:
> This is broken for me using the nightly HA Dev Add-on. There are a lot of browser redirects in this integration so likely something to do with HA ingress and host resolution that did not show up in non-proxied dev testing. For me it fails at the hand back to the auth helper so the music-user-token is not populated and there is no way for the user to manually provide this config. This is my first OSS PR so not sure what the process is to re-open/rollback this PR

We have not released a new beta, so this PR has not reached the end users yet. If you can, you can just fix it in a follow up PR. You can use the dev add-on with your feature branch in order to test your changes in a HA environment.

marcelveldt:
Please don't reply to closed PR's, you are lucky that an issue reporter pointed at this but in general nobody reads replies to closed PR's/issues, certainly not me (I already cant keep up with 100s of regular notifications per day, let alone closed ones).

Looks like it does not work from HA ingress. It does work when connecting to the MA url directly, making this most probably a browser security issue.

I'll have a quick check and otherwise the PR needs to be reverted for now.

PR #2151, opened 2025-04-26
Volume fixes for native announcements

chatziko:
Two volume-related fixes for native announcements:

- Sonos & Snapcast: don't call `get_announcement_volume` in `play_announcement`.
  The `volume_level` argument we get was already adjusted by `get_announcement_volume` (see `PlayerController::play_announcement`), we shouldn't adjust it twice (although in the current implementation of `get_announcement_volume` the second call makes no difference).

- Snapcast: don't adjust the volume when `volume_level is None`


marcelveldt (approved):
Thanks, @chatziko !

PR #2152, opened 2025-04-27
fix: more graceful episode handling for podcast parsers

fmunkes:
See music-assistant/support#3919

If a single episode within the feed lacks the media information, the whole feed is unusable. This is fixed in this PR, and the providers, gpodder, itunes_podcast, podcastfeed, are adjusted accordingly.


marcelveldt (approved):
LGTM, thanks @fmunkes !

PR #2154, opened 2025-04-28
[Airplay] Handle mDNS remove messages without service info

poupounetjoyeux:
When using some mDNS backend like avahi on shairport-sync, mDNS remove message never contains service information but contains the correct name

Currently the player is correctly discovered but if you down it, its state is never updated to 'unavailable' This is causing an issue if we try to play a music on it when it's down because of music assistant just making the Airplay provider unusable and a restart of the server is needed (for all other Airplay player too)

Since service info are needed only to register the player, this change provides a fallback for this type of mDNS services that will just try to parse the name in case it's a remove state change and no info are provided

### DIFF
--- a/music_assistant/providers/airplay/provider.py
+++ b/music_assistant/providers/airplay/provider.py
@@ -187,14 +187,18 @@ class AirPlayProvider(PlayerProvider):
     ) -> None:
         """Handle MDNS service state callback."""
         if not info:
-            return
-        if "@" in info.name:
-            raw_id, display_name = info.name.split(".")[0].split("@", 1)
-        elif deviceid := info.decoded_properties.get("deviceid"):
-            raw_id = deviceid.replace(":", "")
-            display_name = info.name.split(".")[0]
+            if state_change != ServiceStateChange.Removed:
+                return
+            if "@" in name:
+                raw_id, display_name = name.split(".")[0].split("@", 1)
         else:
-            return
+            if "@" in info.name:
+                raw_id, display_name = info.name.split(".")[0].split("@", 1)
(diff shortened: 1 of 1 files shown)


music_assistant/providers/airplay/provider.py
```diff
@@ -187,14 +187,18 @@ async def on_mdns_service_state_change(
...
         """Handle MDNS service state callback."""
         if not info:
-            return
-        if "@" in info.name:
-            raw_id, display_name = info.name.split(".")[0].split("@", 1)
-        elif deviceid := info.decoded_properties.get("deviceid"):
-            raw_id = deviceid.replace(":", "")
-            display_name = info.name.split(".")[0]
+            if state_change != ServiceStateChange.Removed:
+                return
```
marcelveldt: maybe adjust the top call to "if not info and state_change != ServiceStateChange.Removed" to make it easier to read. Also add a comment why we're returning early

poupounetjoyeux: Hello and thanks for the quick review! 😄 

Regarding the comment, I added it

Regarding the condition on top call, if I move it, it means I will need to add an elif case "elif not info" because of the name parsing not using the same parameter (name in one branch vs. info.name in the other one)
Is it what you mean ?
However I inverted the condition logic to make it simpler to read

marcelveldt (approved):
Thanks @poupounetjoyeux !

poupounetjoyeux:
I fixed the lint issue but don't think I'm able to add labels over the PR (or maybe I didn't find the way to make it from my phone 😅)

PR #2160, opened 2025-04-30
Subsonic: Move to new api library

khers:
I refactored the open subsonic api library to better match the specification documentation and to ensure that the objects we return meet that documentation. The new libary also adds typing hints and removed a bunch of hand rolled parsing in favor of mashumaro classes.

This is a big change and I believe I have tested every possible interaction with the provider (and updated the tests) so that it should all be working.


khers:
looks like I used a python 3.13 feature in the library, I will update that and then update this commit

music_assistant/providers/opensubsonic/parsers.py
```diff
@@ -48,11 +45,11 @@ def parse_artist(
     )
 
     artist.metadata.images = UniqueList()
-    if sonic_artist.cover_id:
+    if sonic_artist.coverArt:
```
marcelveldt: is the non-pythonic camelcase name intended ?

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -198,7 +200,7 @@ def _parse_track(
             name=sonic_song.title,
             album=album,
             duration=sonic_song.duration if sonic_song.duration is not None else 0,
-            disc_number=sonic_song.disc_number or 0,
+            disc_number=sonic_song.discNumber or 0,
```
marcelveldt: You're replacing pythonic snake_case with non-pythonic camelCase here.

music_assistant/providers/opensubsonic/parsers.py
khers: Yes, both of these py-opensonic objects changed to camelCase, this is an unfortunate side effect of the move to dataclasses. The Open Subsonic spec inherited the camel casing from the original spec and they chose to maintain backwards compatibility, even when the original spec was broken. See: https://opensubsonic.netlify.app/docs/responses/playlistwithsongs/ as an example of the camel casing (not the broken). I could rename all these fields, but (happy to be wrong here) that would require me doing hand rolled parsing and the point of moving to dataclasses was to drop that.

khers:
I'd like to add the parser move and tests here as well, but happy to split that into its own PR if you prefer.

music_assistant/providers/opensubsonic/parsers.py
marcelveldt: Well, then you can use aliases to always do the conversion for you. Look at the library @fmunkes just created for ABS for example

khers: I'll be back, good thing there are only 300 or so fields that need annotating.

Me editorializing: this snake vs camel case is the first time I have seen a 'pythonic' thing that is genuinely silly as opposed to rooted in a real concern. Don't get me wrong, I am here for a 0 consequence religious war (vi for life) but to make it a part of the 'one true way' for the language is a bit extra.

marcelveldt: it is just a style standard and we (like many other projects) follow the PEP styles - it is ok to follow an existing library which doesn't have all the styles but if you want to setup a library from scratch today, snake case it is.

marcelveldt:
> I'd like to add the parser move and tests here as well, but happy to split that into its own PR if you prefer.

What you prefer. It can also be in a follow up. You name it

music_assistant/providers/opensubsonic/parsers.py
khers: Okay, so the newly pushed commits should do it. The entire library is now PEP style compliant.

marcelveldt (approved):
Looks good, nice job @khers !

octaviz:
with this version will it be compatible with Airsonic Advanced?

khers:
> with this version will it be compatible with Airsonic Advanced?

Only if Airsonic Advance is now using the Open Subsonic API spec defined here: https://opensubsonic.netlify.app/docs 
To the best of my knowledge they are not and have no plans to do so.

octaviz:
Thanks for the response, and why do other projects like feishin support it? airsonic Advanced itself works perfectly. Airsonic is supposed to be more current than Subsonic, which is a dead project, there are other apps like Sonixd that also work with Airsonic. I think there are many more users using Airsonic than Subsonic. Thanks

https://github.com/jeffvli/feishin

khers:
To be clear, this provider does not work with the original Subsonic either. It only works with servers that implement the API specification I linked to earlier. A few options there are Gonic, Navidrome, Lightweight Music Server (LMS), Ampache, or NextCloud Music Plugin. The original Subsonic API was never officially licensed and was not particularly well thought out. The Open specification has a license (Apache 2.0) and is attempting to correct the shortcomings of the original while maintaining backwards compatibility _for applications_. This is why the application you linked works with both the original API and the new Open standard, it was written to only use the original API.

While Airsonic-Advanced is newer than Subsonic, it is not actively maintained (last commit was more than a year ago). I understand there is a user base there, I worry about leaving an Airsonic-Advanced server exposed to the internet given that the maintainer is no longer working on it.

The Open Subsonic provider is written to make use of several of the new return types with richer data and the new endpoints described in that API. Servers implementing the old API, including the subsonic 5.3 forks that are Airsonic and Airsonic-Advanced, do not have the new API and will not work with the provider.

I had hoped to avoid some of this confusion by calling the provider *Open Subsonic* rather than just Subsonic.

octaviz:
Thank you very much for your response, I didn't really know how this API thing worked, that's why I didn't understand why the one I used did work, it's all very complex to be honest, I have some knowledge but not that much, every day you learn something.

PR #2163, opened 2025-05-01
Add extra artist splitters in tag parser

### LINKED ISSUE #3926: Issue with multiple tracks with multiple contributing artists

2.5.1

- [x] Yes

### The problem

There's an issue with recordings that have multiple artists. I have a database of 20.000+ tracks and a query on track_artists only gives me 24 tracks with multiple artists (should be maaaany more).
A random check on one of these shows that the track 'It takes two' shows 'Tina Turner / Rod Stewart & Tina Turner' as the artist.

After investigating a few of the tracks, it shows that this issue occurs whenever two versions of the same track are present in the library, but with two different artists.

One of the tracks had artist Tina Turner
The other 'Rod Stewart & Tina Turner'.
This becomes 'Tina Turner & Rod Stewart & Tina Turner'
In addition, the name of the artist 'Rod Stewart' becomes 'Tina Turner & Rod Stewart' (but the description of the artist refers to Rod Stewart). 
MA thinks this name in the Artist tag is correct for the first-mentioned Musicbrainz Artist Id of this track (which is Rod Stewart).

### How to reproduce

Add the same song to the library twice
Make sure they both contain both Musicbrainz Artist IDs of contributing artists
Make sure they both have different Artist tags (e.g. change the order of both)
Possibly change the order of the Musicbrainz Artist IDs in the file, could be that this contributes to the issue
Expected bug is to find that at least one of the Artists is being stored incorrectly AND that both Artist tags are shown for the both tracks

### Music Providers

Local disk. All music tagged with MusicBrainz Picard.

### Player Providers

N/A

### Full log output

No reference to the issue in the log file.

### Additional information

None

2025.3.3

Home Assistant Container

Linux

- [x] Yes

OzGav:
This might fix https://github.com/music-assistant/support/issues/3926

Musicbrainz shows this 
image

I considered that perhaps an artist might have "with" in their name but it appears not according to ChatGPT and the MB Artist IDs should clarify anyway?

image

### DIFF
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -66,7 +66,7 @@ def split_artists(
     # string. Try to parse the featuring artists and separate them.
     splitters = (
         "featuring", " feat. ", " feat ", "feat.",
-        "duet with", "with", " with ", " duet with "
+        "duet with", "with", " duet with ", " with "
     )
     if allow_ampersand:
         splitters = (*splitters, " & ")
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -64,7 +64,7 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.", "with", " with ", 
+    splitters = ("featuring", " feat. ", " feat ", "feat.", "with", " with ",
                  " duet with ", "duet with")
     if allow_ampersand:
         splitters = (*splitters, " & ")
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -71,6 +71,7 @@ def split_artists(
         " duet with ",
         " with ",
         " ft. ",
+        " vs. ",
         ", ",
         " + ",
     )
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -71,7 +71,7 @@ def split_artists(
         " duet with ",
         " with ",
         " ft. ",
-        " Ft. "
+        " Ft. ",
     )
     if allow_ampersand:
         splitters = (*splitters, " & ")
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -64,7 +64,7 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.")
+    splitters = ("featuring", " feat. ", " feat ", "feat.", "with", " with ")
     if allow_ampersand:
         splitters = (*splitters, " & ")
     artists = split_items(org_artists, allow_unsafe_splitters=False)
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -70,6 +70,8 @@ def split_artists(
         " feat ",
         " duet with ",
         " with ",
+        " ft. ",
+        " Ft. "
     )
     if allow_ampersand:
         splitters = (*splitters, " & ")
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -68,8 +68,6 @@ def split_artists(
         " featuring ",
         " feat. ",
         " feat ",
-        "duet with",
-        "with",
         " duet with ",
         " with ",
     )
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -68,7 +68,6 @@ def split_artists(
         "featuring",
         " feat. ",
         " feat ",
-        "feat.",
         "duet with",
         "with",
         " duet with ",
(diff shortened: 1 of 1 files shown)


music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,8 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.")
+    splitters = ("featuring", " feat. ", " feat ", "feat.", "with", " with ",
+                 " duet with ", "duet with")
```
marcelveldt: put " duet with" before just "with" to prefer splitting on that first

OzGav: done

music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,10 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.")
+    splitters = (
+        "featuring", " feat. ", " feat ", "feat.",
+        "duet with", "with", " with ", " duet with "
```
OzGav:
```suggestion
        "duet with", "with", " duet with ", " with "
```

OzGav:
I don't know why ruff is failing?

marcelveldt:
The formatting is not correct. Are you editing the files from a local dev environment ?
If you set it up according to the dev docs then ruff will run locally to do the formatting for you

OzGav:
I use a Windows PC so haven't been able to fully setup the dev environment as yet. I am using VS code and have installed the ruff extension but it wasn't working for some reason. ChatGPT has saved the day!

music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,16 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.")
+    splitters = (
+        "featuring",
```
marcelveldt:
```suggestion
        " featuring ",
```

music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,16 @@ def split_artists(
...
+        " feat. ",
+        " feat ",
+        "feat.",
```
marcelveldt:
```suggestion
```

music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,16 @@ def split_artists(
...
+        "duet with",
+        "with",
```
marcelveldt:
```suggestion
```

OzGav:
That seems to be all the obvious ones. The comma is the remaining question

music_assistant/helpers/tags.py
```diff
@@ -64,7 +64,15 @@ def split_artists(
     # when not using the multi artist tag, the artist string may contain
     # multiple artists in freeform, even featuring artists may be included in this
     # string. Try to parse the featuring artists and separate them.
-    splitters = ("featuring", " feat. ", " feat ", "feat.")
+    splitters = (
+        " featuring ",
```
marcelveldt, line 68: do we need a title case version of all of them ?
so both " featuring " and "Featuring " - I can also simply do this programaticlaly to make the list easier

OzGav: Yeah maybe programmatically to save this list getting too long?

OzGav: I have removed the Ft. and added some more punctuation based joiners

OzGav:
What do you think about adding " / ", as I just found this https://musicbrainz.org/release/1a35b63f-a10e-420f-af2d-12d21632e5a1

It wouldn't catch AC/DC as that has no spaces?

OzGav:
Here is a possible solution? Help from ChatGPT and Copilot with the regex parts!

```
import re

def split_artists(
    org_artists: str | tuple[str, ...], allow_ampersand: bool = False
) -> tuple[str, ...]:
    """Parse all artists from a string."""
    final_artists: list[str] = []

    splitters = [
        "featuring",
        "feat.",
        "feat",
        "duet with",
        "with",
        "ft.",
        "vs.",
        ",",
        "+",
        "/",  # handled carefully in regex
    ]
    if allow_ampersand:
        splitters.append("&")

    # Build regex pattern:
    # - For /, +, &, require spaces around to avoid splitting in names like AC/DC
    # - For others, allow optional spaces
    special_case_splitters = {"/", "+", "&"}
    pattern = r"|".join(
        rf"\s+{re.escape(s)}\s+" if s in special_case_splitters else rf"\s*{re.escape(s)}\s*"
        for s in splitters
    )
    splitter_re = re.compile(pattern, flags=re.IGNORECASE)

    artists = split_items(org_artists, allow_unsafe_splitters=False)
    for item in artists:
        subitems = splitter_re.split(item)
        for subitem in subitems:
            clean_item = subitem.strip()
            if clean_item and clean_item not in final_artists:
                final_artists.append(clean_item)

    if not final_artists:
        return artists
    return tuple(final_artists)
```

marcelveldt (commented):
Thanks @OzGav !

PR #2165, opened 2025-05-03
fix: musiccast - volume level handling + device discovery + current track

fmunkes:
Fixes for 2 bugs reported in music-assistant/support#3953

1. volume attributes have no type hints in aiomusiccast, and can be None - this is properly handled now
2. device discovery did not cover all possible connection errors - this should be fixed as well

Update May, 4th:
I added an additional commit - the current song/ queue progress didn't update correctly, but playing itself was without problems. Fixed by using "clear_all" within the current_media helper function.


marcelveldt (approved):
LGTM
Thanks, @fmunkes

PR #2166, opened 2025-05-03
Add local lrc lyrics parsing to filesystem provider

fmunkes:
Adds reading a local lrc file to the filesystem provider as discussed on github (1st commit). 

The 2. commit is and idea how this could be used in the lrclib provider - should the local lrc file have been added later than the library sync, then we could still receive it. However, **I didn't know how to properly test the lrc provider part.**

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -971,6 +971,16 @@ class LocalFileSystemProvider(MusicProvider):
                     tags.track_album_loudness,
                 )
             )
+
+        # possible lrclib metadata
+        # synced lyrics are saved as "filename.lrc" by lrcget alongside
+        # the actual file location - just change the file extension
+        assert file_item.ext is not None  # for type checking
+        lrc_path = f"{file_item.absolute_path.removesuffix(file_item.ext)}lrc"
+        if await self.exists(lrc_path):
+            async with aiofiles.open(lrc_path) as lrc_file:
+                track.metadata.lrc_lyrics = await lrc_file.read()
+
         return track
 
     async def _parse_artist(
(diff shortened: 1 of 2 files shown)


music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1495,6 +1500,21 @@ async def _get_local_images(
         await self.cache.set(folder, images, base_key=cache_base_key, expiration=120)
         return images
 
+    async def _get_lrc_data(self, track: Track) -> str | None:
+        # possible lrclib metadata
+        # synced lyrics are saved as "filename.lrc" by lrcget alongside
+        # the actual file location - just change the file extension
+        file_item = await self.resolve(track.item_id)
```
marcelveldt: why don't you just pass the file item from the track parser ?

music_assistant/providers/lrclib/__init__.py
```diff
@@ -110,6 +112,16 @@ async def get_track_metadata(self, track: Track) -> MediaItemMetadata | None:
...
+        # check, if the track comes from any filesystem provider, and might have stored
+        # local lyrics
+        if (filesystem_provider := self.mass.get_provider(track.provider)) and isinstance(
+            filesystem_provider, LocalFileSystemProvider | SMBFileSystemProvider
+        ):
+            if lrc_lyrics := await filesystem_provider._get_lrc_data(track):
+                metadata = MediaItemMetadata()
+                metadata.lrc_lyrics = lrc_lyrics
+                return metadata
+
```
marcelveldt: No, this is not allowed this way as it mixes up responsibilities.
Parsing the metadata from the filesystem is contained in the filesystem, not here.

fmunkes: Alright, I'll revert that, this will then also fulfill your first comment

music_assistant/providers/filesystem_local/__init__.py
fmunkes: That function is gone now, see below, it'll just see if that file exists within the _parse_track

marcelveldt (approved):
Thanks, @fmunkes !

PR #2167, opened 2025-05-03
Subsonic: Fix: Only lookup album for track when an id exists

khers:
We are unconditionally looking up an album regardless of if one exists. Only do that work when we have an ID to lookup.


marcelveldt (approved):
lgtm

PR #2168, opened 2025-05-04
Add subsonic scrobbler plugin

Clusters:
This feature adds a plugin which back-reports the media playback to Subsonic media servers using the scrobble REST endpoint.
It should also work with multiple Subsonic music providers.

Note that, while the Now Playing event is reported to the Subsonic server very accurately, the playback finish event only seems to be reported some seconds before the song ends. This works for me and will also work for most users, but isn't particular in line with the Subsonic definition, where it should be reported after reaching 50% playback progress of a song.

Also note, that I wasn't able to test Audiobooks and Podcast episodes, as I don't have any in my library.

Here are two possible improvements:

~~- On server start, the plugin fails to load because of a race condition with the Subsonic music provider. Because of the async nature, the setup will be triggered at a similar time as the Subsonic music provider. This will heal itself, because MASS retries later again.
Possible solution to that would be, to remove that check (it should make no harm to load the plugin without a Subsonic music provider). Or to wait until the plugin is loaded, either via a timeout, or if we can load the plugin setup after Subsonic music provider has loaded completely anyhow.~~ fixed with depends_on

- The _conn member is used, even though it is protected.
I went with the code in line with the other two scrobblers. Both also use protected members. We could leave it like that or make the extra mile and remove the protection.

~~This PR requires changes in the models repository as well. See https://github.com/music-assistant/models/pull/77~~


OzGav:
Thanks. If you can review the documentation for the other plugins and provide text for each heading for this one that would be great.

music_assistant/providers/subsonic_scrobble/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+  "type": "plugin",
+  "domain": "subsonic_scrobble",
+  "name": "Subsonic Scrobbler",
+  "description": "Report your music playback back to Subsonic server",
+  "codeowners": ["@clusters"],
+  "documentation": "",
+  "multi_instance": false,
+  "builtin": false,
+  "requirements": [
+    "py-opensonic==5.3.1"
```
khers: If this goes in before https://github.com/music-assistant/server/pull/2160, I will update that pull request to pull this plugin forward as well.

Clusters: I could change it to `py-opensonic>=5.3.1,<7.0.0`, I don't see any risk in doing so.

Maybe it would be wise even to set it to `py-opensonic>=5.3.1`, as it just requires one type of that dependency and it would minimize the risk of a dependency conflict with the music provider :)

music_assistant/controllers/player_queues.py
```diff
@@ -2084,6 +2084,7 @@ def _handle_playback_progress_report(
             EventType.MEDIA_ITEM_PLAYED,
             object_id=item_to_report.media_item.uri,
             data=MediaItemPlaybackProgressReport(
+                item_id=item_to_report.media_item.item_id,
```
marcelveldt: why is this needed ? There is alrerady an uri reported

music_assistant/providers/subsonic_scrobble/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
```
marcelveldt, line 1: why not add "depends_on": "subsonic" to the manifest ?

music_assistant/controllers/player_queues.py
marcelveldt, line 1: please limit a PR to a single provider. Any changes to core modules belong to a dedicated PR

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        if report.media_type not in (
+            MediaType.TRACK,
+            MediaType.AUDIOBOOK,
+            MediaType.PODCAST_EPISODE,
+        ):
+            return None
+        prov = self._mass.get_provider(report.provider)
+        if not isinstance(prov, OpenSonicProvider):
+            return None
```
marcelveldt: This is not the way to do this, sorry.
You will receive an uri within the MediaItemPlaybackProgressReport, which could be from SubSonic provider but it can also be a library item. In that case you need to enumerate the provider mappings of the media item to find out if one of them is from subsonic provider.

```
from music_assistant.helpers.uri import parse_uri
    # unwrap library item to check if we have a subsonic mapping...
    library_item = await self.mass.music.get_library_item_by_prov_id(media_type, item_id, provider_instance_id_or_domain)
    if library_item is None:
        return None
    for mapping in library_item.provider_mapings:
        ...
elif not provider_instance_id_or_domain.startswith("subsonic"):
    # not an item from subsonic provider, ignore...
    return None
```

marcelveldt: A way to make this a bit less complicated, would be if we pass along the original mediaitem from the queue item onto the MediaItemPlaybackProgressReport. That way you can simply always enumerate the provider_mappings on the media item.

music_assistant/providers/subsonic_scrobble/manifest.json
khers: The scrobble method didn't change in the churn, but in my conversion to be 'more pythonic' the library is fundamentally not backwards compatible. I think the solution @marcelveldt suggested is the correct one. This plugin should depend on the Open Subsonic music provider, ~even better if we refuse to configure this one without having the provider configured~. EDIT: nevermind, the code already checks this one.

Clusters:
Thanks so far. The week started quite busy for me, I'll tend to all the comments at the weekend latest. Sorry!

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+            unload_cb()
+
+
+class SubsonicScrobbleEventHandler(ScrobblerHelper):
+    """Handles the scrobbling event handling."""
+
+    def __init__(self, mass: MusicAssistant, logger: logging.Logger) -> None:
+        """Initialize."""
+        super().__init__(logger)
+        self._mass = mass
```
marcelveldt: mass can be a public variable

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+            Tuple[OpenSonicProvider | None, str]: The provider or None, and the Subsonic item_id.
+        """
+        if provider_instance_id_or_domain == "library":
+            # unwrap library item to check if we have a subsonic mapping...
+            library_item = await self._mass.music.get_library_item_by_prov_id(
+                media_type, item_id, provider_instance_id_or_domain
+            )
+            if library_item is None:
+                return None, item_id
+            assert isinstance(library_item, (Track, Audiobook, PodcastEpisode))
```
marcelveldt:
```suggestion
            assert isinstance(library_item, Track | Audiobook | PodcastEpisode)
```

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+            return prov, item_id
+        # not an item from subsonic provider, ignore...
+        return None, item_id
+
+    async def _update_now_playing(self, report: MediaItemPlaybackProgressReport) -> None:
+        def handler(prov: OpenSonicProvider, item_id: str, uri: str) -> None:
+            try:
+                self.logger.info("scrobble play now event")
+                prov._conn.scrobble(item_id, submission=False)
+                self.logger.debug(f"track {uri} marked as 'now playing'")
```
marcelveldt:
```suggestion
                self.logger.debug("track %s marked as 'now playing'", uri)
```

prevent using f-strings in log statements

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # the opensubsonic library is not async friendly,
+        # so we need to run it in a executor thread
+        await asyncio.to_thread(handler, prov, item_id, report.uri)
+
+    async def _scrobble(self, report: MediaItemPlaybackProgressReport) -> None:
+        def handler(prov: OpenSonicProvider, item_id: str, uri: str) -> None:
+            try:
+                prov._conn.scrobble(item_id, submission=True, listen_time=int(time.time()))
+                self.logger.debug(f"track {uri} marked as 'played'")
```
marcelveldt: prevent using f-strings in log statements

music_assistant/providers/subsonic_scrobble/manifest.json
Clusters: was changed to depends_on. This also fixed the issue with the race condition on startup 👍

music_assistant/controllers/player_queues.py
Clusters: I've removed the changes in player_queues, parsing the provided path now.

music_assistant/providers/subsonic_scrobble/manifest.json
Clusters: Wasn't aware of that. Makes much more sense 👍

music_assistant/controllers/player_queues.py
Clusters: Not longer relevant, as the changes are only in the provider scope now, after parsing the path.

music_assistant/providers/subsonic_scrobble/__init__.py
Clusters: I absolutely wasn't aware of the library items. Thanks for the hint!
I went with the above suggestions, solely because using the mediaitem insiide the MediaItemPlaybackProgressReport would have lead to a larger refactor.
This can be done in a separate PR I guess. But I don't see any hurry in that. At first, I wasn't sure of the computational expense of such a parsing. But I did some testing, and it seems to be very low.

music_assistant/providers/subsonic_scrobble/__init__.py
Clusters: Changed to public.

music_assistant/providers/subsonic_scrobble/__init__.py
Clusters: Changed to more modern typing.

music_assistant/providers/subsonic_scrobble/__init__.py
Clusters: Changed to string formatting.

music_assistant/providers/subsonic_scrobble/__init__.py
Clusters: Changed to string formatting.

marcelveldt (approved):
Looks good!
Thanks @Clusters

marcelveldt:
> The _conn member is used, even though it is protected.

Let's change that in a follow-up PR if needed. It's indeed a bit strange.

PR #2169, opened 2025-05-04
Always choose mp3 streaming format for the Soundcloud provider

robsonke:
This should solve issue [#3815](https://github.com/music-assistant/support/issues/3815)

Which addresses two issues (both resulting in not start playing):

1. Not every url in the transcoding list has a stream url
2. In some tracks the first transcoding entries contain non-supported formats (or aren't autodetected by ffmpeg, I'm not sure)

Disclaimer: I tested this with the use cases provided in the issue, also with a bunch of tracks in my own library but there might be tracks without mp3 streams too. To quickly solve that in future, I've added a preset list of formats you would like to have as a result.


marcelveldt (approved):
LGTM

PR #2171, opened 2025-05-05
Grammar and additional text suggestion

OzGav:


music_assistant/providers/musiccast/__init__.py
```diff
@@ -202,17 +202,21 @@ async def get_player_config_entries(
...
                         type=ConfigEntryType.STRING,
-                        label="Switch to this non-net source on group leave.",
+                        label="Switch to this non-net source when leaving a group.",
                         default_value=PLAYER_MAP_ZONE_SWITCH[zone_player.zone_name],
-                        description="Switch to this non-net source on group leave. "
-                        " This must be the source_id.",
+                        description="The zone will switch to this source when "
+                        " leaving a group. The source must be one which does not require "
+                        " a network connection. "
+                        " The available sources can be seen when selecting 'Select source' "
```
fmunkes: The available source ids are given in the log at the start of MA - it specifically is not what is shown in the UI, as that is the friendly name, which could have been potentially adjusted by the user.

fmunkes: Adding here. My text is actually misleading for one case, and I had to look through my source again, why I added this to the main zone as well. Two cases:
1. Any zone, including main, has net input, but user wants to stream to another zone. This will be the source the device switches too, if another zone takes net input.
2. Here text is fine: A non-main zone is grouped with main, but then leaves the group - the device will switch this zone to this source. This case does not apply to a main zone.

OzGav: I have slimmed this right down post our discussions and will add some more to the docs to elaborate.

fmunkes (approved):

PR #2172, opened 2025-05-05
Fix: Improvements to MusicKit auth workflow

maxlyth:
The previous version of this PR has issues with HA ingress as tokens were passed back to auth helper via url params that were truncated by an intermediate process causing workflow to fail.

The MA auth helper has since been modified to fix truncation and work on the primary ingress route but this PR includes further mods to support submission of auth data via POST.

This version also tracks time stamp for last confirmed refresh on user token and prompts user to re-authenticate if it is likely for it to expire within 30 days.


maxlyth:
I have four other development streams for this music provider that are not at PR quality yet. I have documented these as TODO items in the main branch code to alert other devs where I have existing research and domain knowledge that is not publicly documented.

maxlyth:
I have set up a reverse proxy to simulate HA ingress for my dev setup and I know this PR is still broken hence switching it to draft. The auth helper returns a callback url on a different host to the ingress url. This is likely to create CORS issues and requires your MA webserver to be configured as exposed. 

I don't understand why the auth helper specifies a protocol:host:port combo as the browser starting the process does not need this now the callback is on the same router. Is the auth helper also used server side without a browser for other workflows? 

I can fix this for HA ingress by subtracting the host part of the callback url to calculate an absolute ingress url but this will then break with more complex reverse proxies with url rewriting so I need the local part of callback url clearly defined.

If the current absolute url pattern is required in order not to break other consumers then I can add another optional parameter but that is just more entropy.

marcelveldt:
> The auth helper returns a callback url on a different host to the ingress url

No, it does not. HA Ingress sets some HTTP headers to inform our webserver about the proxied path.
For regular reverse proxies, you can simply set the correct URL in the settings --> core modules --> web server. That is the path which will be used in the auth callback.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -1,4 +1,17 @@
-"""Apple Music musicprovider support for MusicAssistant."""
+"""
+Apple Music musicprovider support for MusicAssistant.
+
+TODO MUSIC_APP_TOKEN expires after 6 months so should have a distribution mechanism outside
+  compulsory application updates. It is only a semi-private key in JWT format so code be refreshed
+  daily by a GitHub action and downloaded by the provider each initialise.
+TODO Widevine keys can be obtained dynamically from Apple Music API rather than copied into Docker
+  build. This is undocumented but @maxlyth has a working example.
```
marcelveldt, line 8: I doubt that is allowed

music_assistant/providers/apple_music/__init__.py
```diff
@@ -1,4 +1,17 @@
-"""Apple Music musicprovider support for MusicAssistant."""
+"""
+Apple Music musicprovider support for MusicAssistant.
+
+TODO MUSIC_APP_TOKEN expires after 6 months so should have a distribution mechanism outside
+  compulsory application updates. It is only a semi-private key in JWT format so code be refreshed
```
marcelveldt, line 5: Please explain that to me in a PM so we can check if that is allowed, safe and not depending on cloud infrastructure. Note that Github does not allow it to be used as a file host!

marcelveldt, line 5: Maybe better just create these TODOs on the backlog. I'll add you to the provider devs team so you have the access rights

maxlyth:
> > The auth helper returns a callback url on a different host to the ingress url
> 
> No, it does not. HA Ingress sets some HTTP headers to inform our webserver about the proxied path. For regular reverse proxies, you can simply set the correct URL in the settings --> core modules --> web server. That is the path which will be used in the auth callback.

I had not realised that HA Ingress was a special case and set custom http_request.headers that are consumed by MA webserver. I have figured that out now and have a better understanding about how MA server and front-end router interact but still don't have a good way to simulate HA ingress in a dev environment. 

I am confident that I can still get a general solution but doing auth via a browser with MusicKit.js sucks and the result is overly complex. I know that MusicKit uses OAuth under the hood so it should be possible to unwrap it and obtain a token directly but I don't see any prior art on how to do that. Longer term a better solution is required.

Now I have a better understanding of HA ingress I will build a better simulator and update the PR once I am confident I can test it.

music_assistant/providers/apple_music/__init__.py
maxlyth: I don't like the TODOs in the source but they mostly rely on research that is not publicly documented and I don't want others to waste time working on stuff I have already figured out. I have not seen the backlog and it is likely someone else looking into these capabilities will not either

maxlyth: Other commercial Apple Music API providers use a CDN to distribute regularly updated App Tokens but most of the work for this capability for MA would be in devops/infrastructure. This is not a high priority today but will be if an underlying certificate gets revoked. I will diarise to PM you on Discord.

music_assistant/providers/apple_music/__init__.py
maxlyth: There is an undocumented Apple Music API endpoint at https://play.itunes.apple.com/WebObjects/MZPlay.woa/wa/acquireWebPlaybackLicense that MusicKit uses to obtain streaming keys in the embedded browser player. This looks like it might also work for Fairplay and PlayReady content but more research is required. The endpoint is public, just not documented, and using it is probably more conformant than shared widevine keys.

marcelveldt:
If you have Home Assistant running, you can also use the dev/nightly add-on to test things, you can also point it at a specific branch (or even PR)

music_assistant/providers/apple_music/__init__.py
marcelveldt: The backlog is publicaly available and communicated at multiple places in the docs:

https://github.com/orgs/music-assistant/projects/2/views/1

MarvinSchenkel:
Closing this PR in favour of [this one](https://github.com/music-assistant/server/pull/2230) since I could not make changes to this branch

PR #2178, opened 2025-05-06
chore: MusicCast player config improvements + friendly names

fmunkes:
@OzGav and I had a little discussion on provider and documentation

2 adjustments code-wise:

1. For multi-zone players, the player config shows a drop down list of available sources (friendly names) for group/ network leaving. Consequently, I removed the log line printing the available source_ids. This change is non-breaking, internally still the source id is stored, but it should be less error prone, than typing the source_id manually.
2. The log output prints the player's friendly name now, if it is available.

There is also a pure text-change PR here music-assistant/server#2171  - may I merge it?


marcelveldt (approved):
Thanks, @fmunkes !

PR #2179, opened 2025-05-06
chore: sonos - use upnp helpers

fmunkes:
as discussed - **This PR needs device testing, I do not own Sonos speakers**

I believe this is actually the only place, where we have redundant upnp code.


marcelveldt (approved):
Thanks, @fmunkes !

PR #2182, opened 2025-05-10
Fix: Prevent NoneType error whenever the track or album name is empty in a Plex playlist

cociweb:
This pull request resolves cases when a Plex playlist has one (or more) track where the track name and/or album are not defined. In this case, a NoneType error exception was raised, and loading the playlist was totally blocked, as described in https://github.com/music-assistant/support/issues/2711#issuecomment-2746682724.

Additional info log message added for the user to be notified.


marcelveldt (approved):
Thanks, @cociweb !

PR #2188, opened 2025-05-12
Feature: Subsonic: Add bookmark tracking for podcast episodes

khers:
The open subsonic spec supports maintaining a play position mid-episode but we haven't ever wired it up. With the support that has been added to the provider base class, we now have a simple way to make use of this functionality. We will create/update a bookmark for a podcast episode every 30 seconds and when it is stopped. We will delete this bookmark when an episode has beem fully played.

When an episode is started, we check to see if we have a bookmark and resume from that point.

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
 
     from libopensonic.media import Album as SonicAlbum
     from libopensonic.media import Artist as SonicArtist
+    from libopensonic.media import Bookmark as SonicBookmark
     from libopensonic.media import Child as SonicSong
     from libopensonic.media import OpenSubsonicExtension
     from libopensonic.media import Playlist as SonicPlaylist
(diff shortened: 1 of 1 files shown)


music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -750,8 +751,57 @@ async def on_played(
...
+            try:
+                await self._run_async(self._conn.delete_bookmark, id=ep_id)
+            except DataNotFoundError:
+                # We probably raced with something else deleting this bookmark, not really a problem
+                return
+
+        # Otherwise, create a new bookmark for this item or update the existing one
+        # MA provides a position in seconds but expects it back in milliseconds, while
+        # the Open Subsonic spec expects a position in milliseconds but returns it in
+        # seconds, go figure.
```
marcelveldt, line 772: LOL
someday we should maybe refactor this and either use miliseconds for both or use (fractional) seconds everywhere as float

marcelveldt (approved):
LGTM, nice work @khers

PR #2191, opened 2025-05-15
fix: unbound library_items

### LINKED ISSUE #3979: Browsing Tracks in Tidal/Spotify gives error: "cannot access local variable 'library_items' where it is not associated with a value"

2.5.2

- [x] Yes

### The problem

When clicking the "Tracks" folder under browse, Tidal (for example), no items are shown, and an error pops up in the snackbar:
`cannot access local variable 'library_items' where it is not associated with a value`

### How to reproduce

Browse > [Provider] > Tracks

### Music Providers

Tidal, Spotify

Note - I have two accounts logged into Spotify, and only one of them shows this error. The other simply shows no tracks at all (only the 'Go Up' item), but no error.

### Player Providers

N/A, issue is not affecting playback.

### Full log output

Note: It is the second Spotify account (`Alice アリス`) that does not show an error, if at all helpful.
In these logs, Tidal and the Web Server have been changed to `debug` logging.

[music-assistant.log](https://github.com/user-attachments/files/20124723/music-assistant.log)

### Additional information

- It doesn't matter which music provider is currently in use, the behavior does not change.
- It doesn't matter which player provider is currently in use, the behavior does not change.
- No grouping.
- Playback is being done through the UI.
- Container, using the `docker-compose` configuration provided by MA, with host networking mode.
- Browsing tracks appears to work fine for local files and Music Assistant.
- Clearing my browser cache didn't change anything, nor did trying a different browser.

N/A

Home Assistant Container

Linux

- [x] Yes

fmunkes:
Could potentially fix music-assistant/support#3979

Not 100% sure though, just seen it while starting to type hint the models.

### DIFF
--- a/music_assistant/models/music_provider.py
+++ b/music_assistant/models/music_provider.py
@@ -463,7 +463,7 @@ class MusicProvider(Provider):
                 return [x async for x in self.get_library_tracks()]
             library_item_ids = cast("list[int]", library_item_ids)
             query = "tracks.item_id in :ids"
-            query_params = {"ids": library_items}
+            query_params = {"ids": library_item_ids}
             return await self.mass.music.tracks.library_items(
                 extra_query=query, extra_query_params=query_params
             )


marcelveldt (approved):
Good catch @fmunkes !

PR #2192, opened 2025-05-15
Fix: Subsonic: Expose the conn object for use outside of Provider

khers:
The subsonic scrobbling plugin makes use of this member as well, so let's not leave it protected.


khers:
@Clusters Does making this member public work for you?

Clusters:
@khers yes, I also think we can give up the protection here. Thanks for the PR!

Looks good to me

marcelveldt (approved):

PR #2194, opened 2025-05-18
Chore: Separate BrowseFolder from MediaItemType

fmunkes:
Separate BrowseFolder from MediaItemType Alias in models. This gives some more clarity where provider mappings are available as discussed, and should make the type hinting of models/ controller more readable.

I did not have to change much, the majority of functions either does not allow a BrowseFolder anyways, or it doesn't make sense. Went through all occurences of MediaItemType/ MediaItemTypeOrItemMapping

Depends on: music-assistant/models#80


music_assistant/controllers/player_queues.py
```diff
@@ -1590,7 +1590,7 @@ async def _preload_streamdetails(item_id_in_buffer: str) -> None:
         )
 
     async def _resolve_media_items(
-        self, media_item: MediaItemTypeOrItemMapping, start_item: str | None = None
+        self, media_item: MediaItemTypeOrItemMapping | BrowseFolder, start_item: str | None = None
```
marcelveldt: Maybe we should also separate the ItemMapping here again to make this more readable and simple ?
So this will become "MediaItemType | ItemMapping | BrowseFolder" ?

fmunkes: So do you want to remove the MediaItemTypeOrItemMapping Alias completely, or just where it goes along with a BrowseFolder?

marcelveldt: Don't know, what do you think ?
We could maybe leave the alias intact for now and then slowly phase it out ?

fmunkes: well, I do like the shorthand of using the MediaItemTypeOrItemMapping Alias, but I think, I'd rather separate it everywhere - for me it emphasizes, that these are separate things, despite being closely related. First commit will be just your suggestion, second one will change it everywhere. Should you disagree, I just revert that one :-)

fmunkes: I adjusted the model PR as well.

marcelveldt (approved):
Nice work @fmunkes !

PR #2195, opened 2025-05-18
Chore: Mypy for models

fmunkes:
Needs: #2194

There is the following problem left with the music_provider model (as an example):
```
music_assistant/models/music_provider.py:677: error: Argument 1 to "add_item_to_library" of "MediaControllerBase" has incompatible type "Artist | Album | Track | Radio | Playlist | Audiobook | Podcast | PodcastEpisode"; expected "Artist"  [arg-type]
```
All in sync_library, 4 occassions - and it is driving me nuts ;) - there appears to be some subtlety I simply don't see, I tried to replicate this in some dummy code outside of mass, and wasn't able too, mypy never complained.
The controller uses a TypeVar(bound="MediaItemType"), the argument is of type "MediaItemType" - doesn't "Artist | Album | Track | ..." mean I get one of these? So isn't that the same as this (which works fine with mypy):

```python
from typing import TypeVar

class A: ...
class B: ...
class C(A): ...

DummyAlias = A | B
T = TypeVar("T", bound="DummyAlias")

def fun_type_var(var: T) -> T:
    return var

def fun_alias(var: DummyAlias) -> DummyAlias:
    return fun_type_var(var)

fun_alias(A())
fun_alias(B())
fun_alias(C())
```


marcelveldt:
> There is the following problem left with the music_provider model 

Yeah, its indeed strange... maybe because "MediaItemType" accepts more types than the actual media controllers it calls there ? So it could be as simply as creating a type alias BaseMediaType = Artist | Album | Track | Radio | Playlist | Audiobook | Podcast and then do a cast ?

There is no media controller for e.g. an episode so that might be what it is complaining about, hence the cast would be needed

fmunkes:
Latest commit passes mypy (server needs a new model version, to make the lint happy). However, what I added feels like "coding around" a problem, I still don't get. Specifically,
```python
if isinstance(prov_item, Artist):
    assert isinstance(controller, ArtistsController)
    library_item = await controller.update_item_in_library(item_id, prov_item)
elif isinstance(prov_item, Album):
    assert isinstance(controller, AlbumsController)
    library_item = await controller.update_item_in_library(item_id, prov_item)
elif isinstance(prov_item, Track):
    assert isinstance(controller, TracksController)
    library_item = await controller.update_item_in_library(item_id, prov_item)
# and so on
```
works, but I am not allowed to separate these, and move it to a more sensible place, i.e.
```python
if isinstance(prov_item, Artist):
    assert isinstance(controller, ArtistsController)
elif isinstance(prov_item, Album):
    assert isinstance(controller, AlbumsController)
elif isinstance(prov_item, Track):
    assert isinstance(controller, TracksController)
# and so on

# later
library_item = await controller.update_item_in_library(item_id, prov_item)
```
fails mypy (and also pyright, if that matters)

marcelveldt (approved):
Nice work @fmunkes !

PR #2203, opened 2025-05-23
Fix: Subsonic Scrobble: Split the internal id for Podcast Episode

khers:
The Subsonic API only recently added an optional endpoint for retrieving the metadata for an individual podcast episode but most servers do not yet implement it. The only way to get an individual episode is through its channel. This means that we have to include both IDs in the MA provider ID for an episode.

This causes a problem when scrobbling because we are presented with an ID that is not actually valid for the subsonic server. In the case where the string EP_CHAN_SEP exists in the provider id, split it and only call scrobble on the episode portion.


khers:
@Clusters Does this look okay to you?

Clusters:
@khers looks good to me. But can't test this really with Navidrome currently.

khers:
I have tested it with Gonic and it works

marcelveldt (approved):
Thanks, @khers !

PR #2216, opened 2025-06-05
Add Opus to audiobook extensions

jmberg:
Audiobooks can be encoded in opus format just like music can.


marcelveldt (approved):
Thanks @jmberg !

jmberg:
Thanks. I do wonder if it's really even worth distinguishing, vs. just setting AUDIOBOOK_EXTENSIONS = TRACK_EXTENSIONS ?

PR #2217, opened 2025-06-08
Draft: Bluesound sync

Cyanogenbot:
Adding support for Bluesound sync, and multi zone systems. Right now this branch is broken as it seems that the, unsync command doesn't seem to run.

### DIFF
--- a/music_assistant/controllers/players.py
+++ b/music_assistant/controllers/players.py
@@ -187,6 +187,13 @@ class PlayerController(CoreController):
         """Return Player by name or None if no match is found."""
         return next((x for x in self._players.values() if x.name == name), None)
 
+    @api_command("players/get_by_ip")
+    def get_by_ip(self, ip_address: str) -> Player | None:
+        """Return Player by name or None if no match is found."""
+        return next(
+            (x for x in self._players.values() if x.device_info.ip_address == ip_address), None
+        )
+
     # Player commands
 
     @api_command("players/cmd/stop")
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/players.py
```diff
@@ -187,6 +187,13 @@ def get_by_name(self, name: str) -> Player | None:
         """Return Player by name or None if no match is found."""
         return next((x for x in self._players.values() if x.name == name), None)
 
+    @api_command("players/get_by_ip")
+    def get_by_ip(self, ip_address: str) -> Player | None:
+        """Return Player by name or None if no match is found."""
+        return next(
+            (x for x in self._players.values() if x.device_info.ip_address == ip_address), None
+        )
+
```
marcelveldt, line 196: 1) Lets do not a cor change within a provider feature PR
2) Why is this needed ? What is the purpose of having a "get by ip" added to the api ?

Cyanogenbot: Hi Marcel, on discord I mentioned that I probably should change this into a local helper anyway. Regardless, I added this as the bluesound api doesnt use report names, or any other identifier, so when checking what devices are connected to a speaker (for syncing and unsyncing) it only mentions the ip address. A "get by ip" command makes it easier for me to retrieve and send the unsync command to the right device. (as you need to send the command not to the speaker that needs to unsync, but the speaker it is synced to)

marcelveldt: I get that, but leave it as helper in your provider or add a helper in the utils or something. It does not need to be added to the players core controller

OzGav:
Hi @Cyanogenbot Just checking in to make sure this still should be marked as draft?

marcelveldt:
@Cyanogenbot we have refactored the players controller and player model. Your bluesound provider has also been adjusted (but the main part has been done by AI) so please check and rebase your PR. There is a chance that we fixed your issue in between

OzGav:
@Cyanogenbot has this been overtaken by https://github.com/music-assistant/server/pull/2359?

OzGav:
If you are happy it has been then this can probably be closed?

PR #2218, opened 2025-06-08
Fix spaces in GUID for podcasts in iTunes and Gpodder

LinuxSBC:
Currently, if the RSS feed GUID has a space in it, such as in the [This American Life](https://www.thisamericanlife.org/podcast/rss.xml) podcast, both iTunes Podcast Search and Gpodder break due to splitting on every space. This PR fixes that issue by forcing them to only split on the first space.

I don't really like that this requires combining two strings and then splitting them apart later, and it would break if the podcast ID had a space in it, but this works well enough for now while I'm attempting a refactor.

### DIFF
--- a/music_assistant/providers/gpodder/__init__.py
+++ b/music_assistant/providers/gpodder/__init__.py
@@ -510,9 +510,9 @@ class GPodder(MusicProvider):
         self, prov_episode_id: str, add_progress: bool = True
     ) -> PodcastEpisode:
         """Get Podcast Episode. Add progress information."""
-        podcast_id, guid_or_stream_url = prov_episode_id.split(" ")
+        podcast_id, guid_or_stream_url = prov_episode_id.split(" ", 1)
         async for mass_episode in self.get_podcast_episodes(podcast_id, add_progress=add_progress):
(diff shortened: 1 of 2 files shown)


fmunkes (commented):
My problem with this is, that IMHO a guid should not be allowed to contain spaces. For reference, your podcast's episode id is set to '45896 at https://www.thisamericanlife.org' as an example. So what I would favor as a solution is simply ignoring the guid for podcast where this occurs.

fmunkes:
> I don't really like that this requires combining two strings and then splitting them apart later, and it would break if the podcast ID had a space in it, but this works well enough for now while I'm attempting a refactor.

Yeah, maybe a splitter with more characters would have been better - however, changing this now will break backwards compatibility, and I don't see that justified. I think, simply adjusting the parsers in `helpers/podcast_parsers.py` is sufficient to make them ignore a guid with spaces, and fall back to the episode's stream url.

marcelveldt:
Is this PR still relevant now that #2238 is merged ?

fmunkes:
> Is this PR still relevant now that #2238 is merged ?

No, #2238 fixes the issue.
@LinuxSBC thanks for both reporting a bug and providing a possible solution.

PR #2225, opened 2025-06-10
Chore: Subsonic: Clean up last of the typing errors and add tests

khers:
Take the time to clean up the last of the strict type checker errors. While in here, let's add test cases that exercise some of the work arounds we have for various subsonic servers.


marcelveldt (approved):
LGTM!

PR #2226, opened 2025-06-11
Feat: Subsonic: Add configurable recommendations

khers:
The provider has the ability to ask for things like newest albums, most played albums, and anything that is starred by a user. These could be used to fill out the recommendation rows in the home page if the user wants. Add configuration for what kinds (if any) of these recommendations are desired and the size of the list.

As requested in: https://github.com/orgs/music-assistant/discussions/3797

### DIFF
--- a/music_assistant/providers/opensubsonic/__init__.py
+++ b/music_assistant/providers/opensubsonic/__init__.py
@@ -13,7 +13,11 @@ from .sonic_provider import (
     CONF_BASE_URL,
     CONF_ENABLE_LEGACY_AUTH,
     CONF_ENABLE_PODCASTS,
+    CONF_NEW_ALBUMS,
     CONF_OVERRIDE_OFFSET,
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Looks good to me. Any reasons though why the recommendations are turned off by default?

khers:
I don't have strong feelings about off or on by default. This is the least surprising to existing users, but new users might prefer them enabled. Happy to change the default.

MarvinSchenkel:
I would say let's turn it on. It's the best way to get new features to our users. @OzGav thoughts?

OzGav:
We don’t have a standard, discovery is something we want to do more of so I think yes to on.

khers:
The latest version defaults all the displays to True

MarvinSchenkel (approved):

MarvinSchenkel:
Awesome job!

PR #2231, opened 2025-06-14
tidal: modify pagination function

### LINKED ISSUE #4045: Tidal provider issue

2.5.4

- [x] Yes

### The problem

Since yesterday I am having trouble syncing my Tidal library in music assistant.

It only syncs 49 songs out of more than 500 that I have stored in my account.

I have it installed through home assistant, I checked logs and there are no related errors.

I've tried restarting the add-on, logging back into the Tidal integration but it hasn't worked. any ideas what might be happening?

I also tried deploying music assistant outside of home assistant and it went from syncing 49 songs to 59 songs.

### How to reproduce

1. Deploy music assistant with home assistant or docker
2. Enable tidal provider
3. Sync music
4. Evaluate the difference in tracks between music assistant and Tidal

### Music Providers

I only use tidal provider

### Player Providers

Squeezelite

### Full log output

[logs.txt](https://github.com/user-attachments/files/20647252/logs.txt)

### Additional information

Initially I had problems with my local dns, but after restarting the integration and my dns, the error disappeared from the logs but the synchronization problem persists.

2025.5.3

Home Assistant OS

- [x] Yes

jozefKruszynski:
Closes [#4045](https://github.com/music-assistant/support/issues/4045)

Depending on beta cycle, should also be backported to stable

### DIFF
--- a/music_assistant/providers/tidal/__init__.py
+++ b/music_assistant/providers/tidal/__init__.py
@@ -596,13 +596,22 @@ class TidalProvider(MusicProvider):
         item_key: str = "items",
         nested_key: str | None = None,
         limit: int = DEFAULT_LIMIT,
+        cursor_based: bool = False,
         **kwargs: Any,
     ) -> AsyncGenerator[Any, None]:
         """Paginate through all items from a Tidal API endpoint."""
         offset = 0
+        cursor = None
+
         while True:
             # Get a batch of items
-            params = {"limit": limit, "offset": offset}
+            params = {"limit": limit}
+            if cursor_based:
+                if cursor:
+                    params["cursor"] = cursor  # Add cursor if available
+            else:
+                params["offset"] = offset  # Use offset for offset-based pagination
+
             if "params" in kwargs:
                 params.update(kwargs.pop("params"))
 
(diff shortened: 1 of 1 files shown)


jozefKruszynski:
~~Seems this change breaks mix pagination termination 😞~~ 

Synch works correctly now across all Tidal media types

music_assistant/providers/tidal/__init__.py
```diff
@@ -625,8 +625,9 @@ async def _paginate_api(
             offset += len(items)
 
             # Stop if we've received fewer items than the limit
-            if len(items) < limit:
-                break
```
marcelveldt, line 629: Isnt it a bit risky to remove this ? Or is the brak on empty result enough ?

music_assistant/providers/tidal/__init__.py
```diff
@@ -625,8 +625,9 @@ async def _paginate_api(
...
+            if endpoint == "favorites/mixes":
+                if len(items) < limit:
```
marcelveldt: try to combine these 2 ifs

music_assistant/providers/tidal/__init__.py
jozefKruszynski: It's unreliable in it's current state as often the first page of results is not the same size as the limit, could be because of tracks in favourites being removed, but still present as a favourite.

I think I need to look further at this, the empty items break seems to break out correctly at the moment, but this could be luck.

music_assistant/providers/tidal/__init__.py
jozefKruszynski: Gonna refactor this further, I discovered that the V2 api uses cursor pagination rather than paged offsets, so as more stuff moves to the V2 api, I need to reliably page these. Therefore I'll create a second v2_pagination function to handle this.

marcelveldt:
marked as draft - please use the "ready for review" button once you're ready for another/final review

marcelveldt (approved):
LGTM
Thanks, @jozefKruszynski !

PR #2238, opened 2025-06-17
Ignore a guid containing spaces in podcasts parsing

fmunkes:
small fix to ignore a badly formed guid.

This fixes the issue referenced in #2218.


music_assistant/helpers/podcast_parsers.py
```diff
@@ -120,7 +125,8 @@ def parse_podcast_episode(
     except ValueError:
         # we are missing the episode enclosure or stream information
         return None
-    guid_or_stream_url = guid if guid is not None else stream_url
+    # We treat a guid as invalid if it should contain a space.
```
marcelveldt:
```suggestion
    # We treat a guid as invalid if contains a space.
```

music_assistant/helpers/podcast_parsers.py
```diff
@@ -83,6 +83,11 @@ def get_stream_url_and_guid_from_episode(*, episode: dict[str, Any]) -> tuple[st
         raise ValueError("Episode enclosure is missing")
     if stream_url := episode_enclosures[0].get("url"):
         guid = episode.get("guid")
+        if guid is not None:
+            # The media's item_id is {prov_podcast_id} {guid_or_stream_url}
+            # see parse_podcast_episode.
+            # However, the guid must not contain a space, otherwise it is invalid.
+            guid = None if len(guid.split(" ")) > 1 else guid
```
marcelveldt: just a thought but what about parsing the string into an actual GUID to check if its valid ?

so something like this:

```
try:
    guid = uuid.UUID4(possible_uuid_str)
except ValueError:
    guid = None
```

fmunkes: That would be much preferred indeed, yet unfortunately many podcasts do not put a proper uid in there. I've seen pure integers, or something like this `gid://art19-episode-locator/V0/hvosPPcp7VLqHg1P2XRpZI66F2ERb1xcBUfPKMW2tyI` ([here](https://rss.art19.com/was-bisher-geschah)) in one of the podcasts I listen too. I enhanced the comment a bit.

marcelveldt (approved):

PR #2240, opened 2025-06-19
2.5.5 backport

khers:
These are the fix commits I can find that are not already in stable. Let me know if I missed any.


khers:
It looks like the tests are hosed in stable prior to this back port. I suspect the move to pull the subsonic library forward jumped an update to tests, maybe? Not sure if the tests should also be cherry picked to fix

marcelveldt (approved):
Thanks @khers !

PR #2242, opened 2025-06-21
Subsonic: Display the newest podcast episodes as front page recommendation

khers:
The front page recommendation area makes for a great place to display the newest episode from your subscribed feeds. If podcasts are enabled in the provider, add a row for this.


marcelveldt (approved):
LGTM, nice work @khers

PR #2249, opened 2025-06-24
Player controller (and model) refactor

marcelveldt:
Implements https://github.com/music-assistant/backlog/issues/2

Depends on a models bump: https://github.com/music-assistant/models/pull/91

Long overdue refactor of the players controller to make the code more maintainable in the longer term and ease the development and maintenance of (new) player providers as well as prepare for future changes.

- Differentiate between a Player class used internally in the server and a PlayerState which is the (snapshotted) state of a player at a certain point in time and what is shared over the API with clients.

- Adjust Player model in shared models package to only include shared attributes (final state) - within the server code this model is referred to as PlayerState.

- Created a serverside model for a Player class with actual functions that can be subclassed by player providers.


MarvinSchenkel (approved):
Let's go!

PR #2254, opened 2025-06-27
Parallelize Recommendation creation in Subsonic provider

khers:
When all the recommendations are enabled the main page is rather slow to load. Split each kind into a helper function and invoke each helper that is enabled in parallel.

The newest podcast episode recommendation was the worst offender due to the number of API calls for getting podcast channels. This was broken into an initial pass over episodes to collect a set of channel ids, then a parallel retrieval of the unique channel ids.

If there is a cleaner/more pythonic way to do this, I am happy to rework.


marcelveldt:
I would involve the cache here. You can use mass.cache.get and mass.cache.set with your own unique keys. 
Even a short TTL on the cache can already greatly reduce loading costs. For an example of hot to leverage the cache, see some of the other providers like AudioBookShelf or Tidal.

Otherwise I think code looks fine to solve this now

khers:
Use the cache instead of constructing my own dict? Sounds good, I will introduce here but it sounds like this would be a good change for most of the podcast accesses so I will have a look at expanding in another PR.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -776,6 +776,72 @@ def _streamer() -> None:
 
         self.logger.debug("Done streaming %s", streamdetails.item_id)
 
+    async def _get_podcast_channel_async(self, chan_id: str) -> None:
+        chan = await self._run_async(self.conn.get_podcasts, inc_episodes=True, pid=chan_id)
+        await self.mass.cache.set(key=chan_id, data=chan, expiration=3600)
```
marcelveldt: the channel id is maybe a bit too generic as cache key.
I would recommend also defining a cache category in the mass global constants (around line 700) and assign that category to all of your cache set/get actions. It is also faster because it's indexed so the cache lookup can first do an indexed lookup on the category aimed at subsonic and only has to do the text search

marcelveldt: It also reminds me that we should maybe turn the cache category into an IntEnum at some point but that is for later

khers: I'd like to get fast recos in before the next beta, but I will take a look at converting the cache categories to an IntEnum this weekend.

marcelveldt (approved):
LGTM

PR #2265, opened 2025-07-03
Move Cache Categories to an Enum Class

khers:
The more Pythonic way to hold this collection of integer constants is to use the IntEnum class.


music_assistant/constants.py
```diff
@@ -697,20 +698,23 @@ def create_sample_rates_config_entry(
 
 
 # CACHE categories
+class CacheCategory(IntEnum):
```
marcelveldt: Would it be an idea to have this enum defined in the cache controller module itself ?
For visibility/discoverability I mean

khers:
Looks like that broke the import, this still needs fixing

OzGav:
@khers Have you had a chance to progress this?

khers:
I have not. Untangling this will require a pretty invasive change to pull apart the dependencies. I think it should wait until the player refactor is done because that one is also wide spread and invasive.

OzGav:
I think the player refactor is done? https://github.com/music-assistant/server/pull/2249

khers:
I hadn't seen that, I will try and get back to this one then.

marcelveldt:
superseded by https://github.com/music-assistant/server/pull/2430

PR #2255, opened 2025-06-30
Fix library query with random order

maximmaxim345:
My attempt to fix random ordering with search and favorite filters not working correctly due to filters being applied after random selection instead of before.
Tested with a couple dozens of query's through Home Assistant, seams to work correctly now.  

Supersedes #2206

Fixes music-assistant/support#3943 and music-assistant/backlog#10

PR #2270, opened 2025-07-06
Fix library query with random order and improve maintainability

### LINKED ISSUE #3943: Get Library search unreliable

2.6.0b5 & 2.5.2

- [x] Yes

### The problem

With this action:

```
action: music_assistant.get_library
data:
  config_entry_id: 01JS2BHC9PB00EWJZ7C9D42MMH
  media_type: album
  search: Kinderboeken
  limit: 1
  offset: 0
  order_by: random
response_variable: random_album
```

I sometimes correctly get a response like this:

```
items:
  - media_type: album
    uri: library://album/1
    name: Abeltje
    version: ""
    image: null
    artists:
      - media_type: artist
        uri: library://artist/317
        name: Kinderboeken
        version: ""
        image: null
limit: 1
offset: 0
order_by: random
media_type: album
```

But very often (I'd say it's about fifty-fifty) I get this:
```
items: []
limit: 1
offset: 0
order_by: random
media_type: album
```

When I leave the search part out, it never returns an empty response.
Could you please confirm and solve this bug? Thanks in advance!

### How to reproduce

See above.

### Music Providers

Filesystem localhost

### Player Providers

Doesn't matter.

### Full log output

I don't think this is necessary here.

### Additional information

None.

2025.4.2

Home Assistant Container

Windows

- [x] Yes

maximmaxim345:
This PR fixes a couple of issues I missed in #2255 and refactors the query building logic into smaller, maintainable helper methods.

## Fixes:

Random Ordering with Filters: Fixed random ordering not working correctly when search/favorite filters were applied. Filters were being applied after random selection instead of before. Now all filters are applied within the random subquery itself.

Album Search Limits: Fixed album searches returning more results than the specified limit due to duplicate prevention logic. Now properly calculates remaining limit and stops when the original limit is reached.

## Testing:

Verified with a couple of search queries through Home Assistant and Music Assistant with various combinations of random ordering, search filters, and favorite filters.

For my dev setup, these queries previously worked incorrectly:
```yaml
action: music_assistant.get_library
data:
  config_entry_id: 01JZAM4T977MRTC2XFKXC925PA
  media_type: album
  offset: 0
  order_by: random
  limit: 1
  favorite: true
  search: ghost
```
and:
```yaml
action: music_assistant.get_library
data:
  config_entry_id: 01JZAM4T977MRTC2XFKXC925PA
  media_type: album
  offset: 0
  order_by: random
  favorite: true
  album_type:
    - single
  search: ad
  limit: 1
```

## Limitations

In case a search term does not yield enough results, it additionally searches results with the Album Artists and appends them to the search results. But both results are randomized independently.
This means the search results aren't fully randomized together - album name matches appear first in random order, then artist name matches appear in their own random order.

Fixes: music-assistant/support#3943, music-assistant/backlog#10

### DIFF
--- a/music_assistant/controllers/media/albums.py
+++ b/music_assistant/controllers/media/albums.py
@@ -171,7 +171,7 @@ class AlbumsController(MediaControllerBase[Album]):
             # Calculate how many more items we need to reach the original limit
             remaining_limit = limit - len(result)
 
-            for _album in await self._get_library_items_by_query(
+            for album in await self._get_library_items_by_query(
                 favorite=favorite,
                 search=None,
                 limit=remaining_limit,
@@ -182,8 +182,8 @@ class AlbumsController(MediaControllerBase[Album]):
                 extra_join_parts=extra_join_parts,
             ):
                 # prevent duplicates (when artist is also in the title)
-                if _album.uri not in existing_uris:
-                    result.append(_album)
+                if album.uri not in existing_uris:
+                    result.append(album)
                     # Stop if we've reached the original limit
                     if len(result) >= limit:
                         break
--- a/music_assistant/controllers/media/albums.py
+++ b/music_assistant/controllers/media/albums.py
@@ -167,10 +167,14 @@ class AlbumsController(MediaControllerBase[Album]):
             )
             extra_query_params["search_artist"] = f"%{search}%"
             existing_uris = {item.uri for item in result}
+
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):

marcelveldt:
At some point we should also consider splitting up the music controller in more parts, for example a dedicated database helper or child class

PR #2271, opened 2025-07-06
Attempt to fix "This Device" disappearing and reappearing on Firefox

maximmaxim345:
Previously the Builtin Web Player still timed out sometimes when using Firefox.
This PR increases the previous headroom of 30s to 60s, which should be enough for Firefox throttling JavaScript in the background.

But I couldn't verify this fix myself since it already worked before on all devices I have.


marcelveldt:
Note that we will not accept PR's to player providers until the big refactor PR of the player controller and model is finished:
https://github.com/music-assistant/server/pull/2249

marcelveldt:
@maximmaxim345 now that the whole player controller/model refactor is done, this can be unblocked

marcelveldt (approved):

PR #2275, opened 2025-07-07
Added recommendations to the SoundCloud provider

robsonke:
This PR adds a new provider feature for recommendations. I'm using the generated playlists from SoundCloud to populate MA.

An example, SoundCloud:
image

And in Music Assistant:
image

Note: please let me know if the debug loggers are acceptable. It gives me useful info but it's also a bit of pollution.

### DIFF
--- a/music_assistant/providers/soundcloud/__init__.py
+++ b/music_assistant/providers/soundcloud/__init__.py
@@ -424,6 +424,9 @@ class SoundcloudMusicProvider(MusicProvider):
     async def _parse_playlist(self, playlist_obj: dict[str, Any]) -> Playlist:
         """Parse a Soundcloud Playlist response to a Playlist object."""
         playlist_id = str(playlist_obj["id"])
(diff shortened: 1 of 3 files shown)


OzGav:
Any chance of removing the `Related tracks:` prefix so you can see more of the title?

robsonke:
> Any chance of removing the `Related tracks:` prefix so you can see more of the title?

I can, but it's the name of the playlist. That it's built up as related tracks to track x. Just showing track x as name, feels empty/unclear.

OzGav:
I see that “Related Tracks” is on every picture as well though?

robsonke:
> I see that “Related Tracks” is on every picture as well though?

Good point, that's clear enough. Let me do some dirty string cutting.
Edit: better
image

MarvinSchenkel (approved):
Looks good to me, thanks @robsonke

PR #2283, opened 2025-07-19
Add bbc sounds provider

kieranhogg:
This is my initial version of a [BBC Sounds](https://www.bbc.co.uk/sounds) music provider for a review and feedback.

The two main areas I would like some feedback/eyes on are: 
- the 'now playing' metadata updating functionality - this works great but I'm unsure if there is a neater mechanism
- it seems to have resolved itself now, but frequently on first play of a station the player would just display e.g. "bbc_station". Hitting play again would always fix the issue, so I was unsure whether it was a stream handling issue, a bug in my code (maybe I am missing an event in the listener?), or something to do with caching


MarvinSchenkel:
What you can do for the for populating the media title is keep a reference to your `StreamDetails` inside your class and just update that. Have a look at the [SiriusXM provider](https://github.com/music-assistant/server/blob/dba36517ab8eb4e6a6f42d5090b88896831474df/music_assistant/providers/siriusxm/__init__.py#L254) for an example of this. That way, you can remove all the player/queue logic. 

Other than that, looks like another great addition to the music providers, nice job! :-)

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+
+    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Get streamdetails for a track/radio."""
+        station = await self.client.stations.get_station(item_id, include_stream=True)
+        if not station or not station.stream:
+            raise MusicAssistantError(f"No stream found for {item_id}")
+
+        return StreamDetails(
+            stream_title=station.description,
```
MarvinSchenkel: My guess is that this is the reason the initial play displays 'bbc_station' on the initial start. I think you want to use the logic from `_display_station` here to properly calculate the stream title on the first playback.

kieranhogg:
> What you can do for the for populating the media title is keep a reference to your `StreamDetails` inside your class and just update that. Have a look at the [SiriusXM provider](https://github.com/music-assistant/server/blob/dba36517ab8eb4e6a6f42d5090b88896831474df/music_assistant/providers/siriusxm/__init__.py#L254) for an example of this. That way, you can remove all the player/queue logic. 

Thanks for the review! On the StreamDetails part, this was what I originally wrote, but when I was playing different stations to multiple players it broke down as it only pointed to a reference of whichever stream had changed last. I had looked at the Sirius implementation but couldn't work out how, if it does, deal with that situation.

MarvinSchenkel:
> > What you can do for the for populating the media title is keep a reference to your `StreamDetails` inside your class and just update that. Have a look at the [SiriusXM provider](https://github.com/music-assistant/server/blob/dba36517ab8eb4e6a6f42d5090b88896831474df/music_assistant/providers/siriusxm/__init__.py#L254) for an example of this. That way, you can remove all the player/queue logic.
> 
> Thanks for the review! On the StreamDetails part, this was what I originally wrote, but when I was playing different stations to multiple players it broke down as it only pointed to a reference of whichever stream had changed last. I had looked at the Sirius implementation but couldn't work out how, if it does, deal with that situation.

@marcelveldt should we provide some generic functionality for this use-case? I can image a music provider with an endless stream might want to somehow notify the queue in case when the metadata has changed.

OzGav:
@kieranhogg once you get close to merging can you provide some text for the docs? We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+        # TODO: detect blank song images
+        queue = self.mass.player_queues.get(queue_id=queue_id)
+        if not queue:
+            return
+        station = await self.client.stations.get_station(station_id=station_id)
+        if not station:
+            return
+        now_playing = await self.client.schedules.currently_playing_song(station, image_size=1280)
+
+        # TODO: check if there is a neater way to do this
```
marcelveldt: The better way to do this is from "get_streamdetails". So you hand out the streamdetails and then you keep updating the stream metadata. You may not interact with the queue or queue item directly.

Just launch a task from the "get_streamdetails" that keeps updating the stream metadata.
And then the best way to break out of it is when the streamdetails have a "seconds_streamed" filled filled because that means the stream was done.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+                    MediaItemImage(
+                        type=img_type,
+                        path=station.logo_url,
+                        remotely_accessible=True,
+                        provider=self.domain,
+                    )
+                    for img_type in image_types
+                ]
+            )
+            queue.current_item.media_item.name = f"BBC {station.name}"
```
marcelveldt: this is not allowed

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+            queue.current_item.streamdetails.stream_title = station.description
+            queue.current_item.media_item.metadata.images = images
```
marcelveldt: this is not allowed

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+            if queue.current_item.streamdetails.provider == self.domain:
+                station_id = queue.current_item.streamdetails.item_id
+                self.logger.debug("Found a bbc_sounds queue")
+                if queue.state == PlayerState.PLAYING and queue.current_item is not None:
+                    self.logger.debug("Queue is playing, updating its now playing metadata")
+                    await self._update_player_now_playing(queue.queue_id, station_id)
+
+                elif queue.state in [PlayerState.PAUSED, PlayerState.IDLE]:
+                    self.logger.debug("Queue is paused, reverting to station information")
+                    await self._display_station(queue.queue_id, station_id)
```
marcelveldt: Please drop this as its not the way to do this

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,405 @@
...
+            },
+        )
+
+    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Get streamdetails for a track/radio."""
+        station = await self.client.stations.get_station(item_id, include_stream=True)
+        if not station or not station.stream:
+            raise MusicAssistantError(f"No stream found for {item_id}")
+
+        return StreamDetails(
```
marcelveldt: So, when returning the streamdetails, you also launch a background task that keeps updating the stream's metadata until the "seconds_streamed" property is set. Then you can break out of that task.

marcelveldt:
So, what kind of metadata is provided ? Only a single title or artist-title ? Or even more ?
We have played with the thought of extending stream_title to stream_metadata so you can provide even more details.

kieranhogg:
> So, what kind of metadata is provided ? Only a single title or artist-title ? Or even more ? We have played with the thought of extending stream_title to stream_metadata so you can provide even more details.

Thanks for the review comments, I suspected as much on the queue meddling so I will get those resolved before I move out of draft.

On the metadata discussion, because of the functionality of the Sounds API it really does fall inbetween the current models of streaming with low metadata and playing music sources with high metadata. 

For pure live streaming, I am aiming to replicate as close as I can to the functionality of the official app. I was able to do this with the queue meddling, so obviously a supported method would be useful but I appreciate it's quite a niche usecase at the moment. A few screenshots below to illustrate, but the short version is the API returns a currently playing song's titles (almost exclusively track artist and song title) and an appropriate image. So no album metadata.

IMG_0952
*iOS home screen when radio programme is playing*

IMG_0953
*iOS home screen when station plays a song*

MA

*Similar functionality I had achieved with existing code*

Within the Sounds API (and therefore my provider) the lines between live stations, previous programmes and podcasts do get a little blurred. In this context, the lack of similar metadata for the live stream does stand out a little. E.g. an available podcast series below.

Podcast

### More Context
The sounds API returns details on various containers and objects which I convert into dataclasses in the library. This isn't completely finalised yet, but here is the current mapping to the respective MA dataclasses which might be useful to see:

```
self.type_mapping = {
    Category: Playlist,
    Collection: Playlist,
    Station: Radio,
    Stream: StreamDetails,
    MenuItem: BrowseFolder,
    RadioSeries: MAPodcast,
    RadioShow: MAPodcastEpisode,
    Podcast: MAPodcast,
    PodcastEpisode: MAPodcastEpisode,
    ScheduleItem: Radio,
}
```
Of these, a **RadioShow** and a **Stream** (perhaps **PodcastEpisode** but haven't confirmed) both have the ability to reference what song is playing at a given time through the API. So in an ideal world, I would like to show the currently playing metadata for those two items.

The other useful functionality is that a current radio stream has a definite length available through the API. Seeking backwards within a live stream to the beginning of the current show is supported, as is seeking forwards to the current live edge. As it stands, I don't think this is possible without making it a **PodcastEpisode** or a **Track**, but that feels very hacky, as well as needing more logic when the current programme ends, even though the stream continues playing fine.

marcelveldt:
So, I added "stream_metadata" to the streamdetails where you can dynamically update the metadata of the currently playing track/item. 

https://github.com/music-assistant/models/commit/2eaa5ca9846850fb994a1e3bf6a9d2b052f68cd1

I will now update the server to bump the models so this is available. Just make sure to rebase your branch and then re-run the setup script.

Then, you can add some logic to keep updating the streamdetails.stream_metadata while its playing

OzGav:
@kieranhogg just checking in how this is going? It is often requested so would be good to get it merged.

kieranhogg:
> @kieranhogg just checking in how this is going? It is often requested so would be good to get it merged.

Yeah I keep meaning to get this pushed any day now. It's 99% done including the suggestions in this PR, but there's a few rough edges I'm trying to sort out — as I'm shoehorning the Sounds catalogue into Podcast, PodcastEpisode and Track which works until it doesn't.

OzGav:
@kieranhogg We hope to do a beta release soon so just checking whether this might be ready in a couple of days. No probs if not.

PR #2293, opened 2025-07-29
Improve lyrics availability

OzGav:
This PR does two things:

1. It scans for lyrics via the lrclib provider when a track is added to the library
2.  It makes some changes to the lrclib provider to skip the album name if it is not available as this was causing the lookup to fail unnecessarily. And it also makes plain lyrics available, which is supported by the frontend.


OzGav:
Lint is failing on files not part of this PR?

music_assistant/controllers/metadata.py
```diff
@@ -591,12 +591,23 @@ async def _update_track_metadata(self, track: Track, force_refresh: bool = False
...
+        # if the track needs a refresh (based on REFRESH_INTERVAL_TRACKS) AND
+        # online metadata is enabled.
+        if (force_refresh or needs_refresh) and self.config.get_value(CONF_ENABLE_ONLINE_METADATA):
             for provider in self.providers:
                 if ProviderFeature.TRACK_METADATA not in provider.supported_features:
                     continue
+                    
+                # Stop unnecessary calls to lrclib
+                if provider.lookup_key == "lrclib":
+                    if track.metadata and (track.metadata.lyrics or track.metadata.lrc_lyrics):
```
marcelveldt: please combine this if statements into one

OzGav: Fixed

OzGav:
I cant find any whitespace in the file and nor could copilot?

marcelveldt:
> I cant find any whitespace in the file and nor could copilot?

You have a blank line with a whitespace.
Just run the dev environment per the setup instructions which also automatically runs pre-commit to catch these errors.
Now, from within the venv, run "pre-commit run --all-files" to auto correct it.

OzGav:
found it

PR #2304, opened 2025-08-06
Split up data and cache with XDG directories

jamiethecat:
To not break existing setups, we use the old data directory if it exists.

I'd also like to port all of the path logic to the more modern `pathlib` in the future :)

### DIFF
--- a/music_assistant/__main__.py
+++ b/music_assistant/__main__.py
@@ -35,14 +35,32 @@ def get_arguments() -> argparse.Namespace:
     """Arguments handling."""
     parser = argparse.ArgumentParser(description="MusicAssistant")
 
-    default_data_dir = os.path.join(os.path.expanduser("~"), ".musicassistant")
+    if os.path.isdir(old_data_dir := os.path.join(os.path.expanduser("~"), ".musicassistant")):
+        data_dir = old_data_dir
+        cache_dir = os.path.join(data_dir, ".cache")
+    else:
+        data_dir = os.path.join(
+            os.getenv("XDG_DATA_HOME", os.path.join(os.path.expanduser("~"), ".local", "share")),
+            "music-assistant",
+        )
+        cache_dir = os.path.join(
+            os.getenv("XDG_CACHE_HOME", os.path.join(os.path.expanduser("~"), ".cache")),
+            "music-assistant",
+        )
 
     parser.add_argument(
+        "--data-dir",
         "-c",
         "--config",
-        metavar="path_to_config_dir",
-        default=default_data_dir,
-        help="Directory that contains the MusicAssistant configuration",
+        metavar="path_to_data_dir",
+        default=data_dir,
+        help="Directory that contains MusicAssistant data",
+    )
+    parser.add_argument(
+        "--cache-dir",
+        metavar="path_to_cache_dir",
+        default=cache_dir,
+        help="Directory that contains MusicAssistant cache",
     )
     parser.add_argument(
         "--log-level",
@@ -56,6 +74,7 @@ def get_arguments() -> argparse.Namespace:
         action=argparse.BooleanOptionalAction,
         help="Start in safe mode (core controllers only, no providers)",
     )
+
     return parser.parse_args()
 
 
@@ -171,9 +190,11 @@ def main() -> None:
     """Start MusicAssistant."""
     # parse arguments
     args = get_arguments()
-    data_dir = args.config
-    if not os.path.isdir(data_dir):
-        os.makedirs(data_dir)
+
+    data_dir = args.data_dir
+    cache_dir = args.cache_dir
+
+    os.makedirs(data_dir, exist_ok=True)
 
     # TEMP: override options though hass config file
     hass_options_file = os.path.join(data_dir, "options.json")
@@ -192,7 +213,7 @@ def main() -> None:
 
     # setup logger
     logger = setup_logger(data_dir, log_level)
-    mass = MusicAssistant(data_dir, safe_mode)
+    mass = MusicAssistant(data_dir, cache_dir, safe_mode)
 
     # enable alpine subprocess workaround
     _enable_posix_spawn()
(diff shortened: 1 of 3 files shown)


marcelveldt:
I don't understand why we need to account for XDG_CONFIG_HOME as we're targeting only on a dockerized setup based on Alpine. I dont think I'm a fan of this change unless you can argue why this is needed

music_assistant/__main__.py
```diff
@@ -173,19 +160,53 @@ def _global_loop_exception_handler(_: Any, context: dict[str, Any]) -> None:
...
+            os.path.join(os.path.expanduser("~"), ".config", "music-assistant"),
+        )
+        cache_dir = os.getenv(
+            "XDG_CACHE_HOME",
+            os.path.join(os.path.expanduser("~"), ".cache", "music-assistant"),
+        )
+        state_dir = os.getenv(
+            "XDG_STATE_HOME",
+            os.path.join(os.path.expanduser("~"), ".local", "state", "music-assistant"),
+        )
```
marcelveldt: I have no clue why all this split up is needed ?

music_assistant/mass.py
```diff
@@ -112,10 +112,19 @@ class MusicAssistant:
     streams: StreamsController
     _aiobrowser: AsyncServiceBrowser
 
-    def __init__(self, storage_path: str, safe_mode: bool = False) -> None:
+    def __init__(
+        self,
+        data_dir: str,
+        config_dir: str,
+        cache_dir: str,
+        state_dir: str,
```
marcelveldt: If you want to split up persistent vs non-persistent data, I think it will be enough to have a data and cache dir only. This is way to complex.

jamiethecat:
The XDG directories are the locations where users expect specific files to be, since it's encouraged and commonplace for software to put their data there.

Some users choose to transfer their `.config` directory across installations, but not their data, and state is for non-persistent, non-cache data.

I think it would still be an improvement to just put persistent data and logs in `XDG_DATA_HOME` and cache in `XDG_CACHE_HOME`, if you find that to be better though :)

marcelveldt:
@jamiethecat this software is not supposed to be run standalone but only/always in a containerized manner. So then again I dont really understand why this change is needed. Running it without the container is not even supported by us in any way.

What I do support is splitting up data and cache to make a difference between them.
For the HA add-on we simply exclude the cache subfolder from the backup to partly cover this usecase but better would be to simply have a /data volume and a /cache volume

jamiethecat:
Fair, one of the reasons for the PR was to split these up.

It does makes sense to use the standard directories for these if it was split up anyway, no? Two birds with one stone. And I can keep only data and cache for simplicity.

This also allows setting these directories with the standard environment variables instead of the `--config` flag, which is a plus in my opinion.

I'm not too familiar with Docker volumes, so feel free to close this PR if that's desired instead in any case :)

marcelveldt:
I think you have a fair point with splitting up data vs cache but we should aim at the intended way to run the server, which is containerized, so we have a controlled OS/base-layer with predictable OS-level dependencies/binaries and such.

Running the server within your own linux distro is not supported anyways, as you we have no way to guarantee that it will run, considering we depend on several binaries. The only way to run the server standalone is as development environment, in a python venv.

So, moving forward I think its OK to adjust the main entrypoint logic a bit to use the XDG environmental variables but it will not make much sense for the production setup as that will always be run as container, where the data and cache volumes will be mounted.

Your PR can then be simplified a bit with just differentiating between data and cache

jamiethecat:
Alright, updated the PR to only handle `XDG_DATA_HOME` and `XDG_CACHE_HOME`.

music_assistant/__main__.py
```diff
@@ -174,12 +161,31 @@ def main() -> None:
...
-    if not os.path.isdir(data_dir):
-        os.makedirs(data_dir)
+
+    if os.name == "nt":
+        app_data = os.getenv("APPDATA", r"C:\Users\Default\AppData\Roaming")
+        local_app_data = os.getenv("LOCALAPPDATA", r"C:\Users\Default\AppData\Local")
+
+        old_storage_path = os.path.join(app_data, ".musicassistant")
+        storage_path = os.path.join(local_app_data, "MusicAssistant")
+        cache_path = os.path.join(storage_path, "cache")
```
marcelveldt: windows support can be skipped entirely - it wont run at all

music_assistant/__main__.py
```diff
@@ -34,19 +34,6 @@
...
-        parser.error("Unable to find default data dir")
-    default_data_dir = os.path.join(default_data_dir, ".musicassistant")
-
-    parser.add_argument(
-        "-c",
-        "--config",
-        metavar="path_to_config_dir",
-        default=default_data_dir,
-        help="Directory that contains the MusicAssistant configuration",
-    )
```
marcelveldt: the commandline argument must be kept. Just add a second one for the cache dir.
Just set the default to the XDG_DATA_HOME env var.

I think the nicest way to handle all this is by specifying a data-dir and cache-dir argument and add config as alias (for backwards compatibility).

music_assistant/__main__.py
```diff
@@ -35,17 +35,32 @@ def get_arguments() -> argparse.Namespace:
...
     parser.add_argument(
+        "--data-dir",
         "-c",
         "--config",
-        metavar="path_to_config_dir",
-        default=default_data_dir,
-        help="Directory that contains the MusicAssistant configuration",
+        metavar="path_to_data_dir",
+        default=data_dir,
+        help="Directory that contains MusicAssistant data",
```
marcelveldt:
```suggestion
        help="Directory that contains MusicAssistant persistent data",
```

music_assistant/__main__.py
```diff
@@ -35,17 +35,32 @@ def get_arguments() -> argparse.Namespace:
...
+    )
+    parser.add_argument(
+        "--cache-dir",
+        metavar="path_to_cache_dir",
+        default=cache_dir,
+        help="Directory that contains MusicAssistant cache",
```
marcelveldt:
```suggestion
        help="Directory that contains MusicAssistant cache data",
```

marcelveldt (approved):
Thanks, @jamiethecat !

PR #2306, opened 2025-08-08
Update librespot binaries

OzGav:


ansballard:
👋🏻 In case it helps to have someone testing, I pulled these into my current music assistant instance and they threw `RuntimeError: Unable to locate Librespot for linux/x86_64`. So I ran down where those binaries might've come from and found https://github.com/music-assistant/librespot/actions/runs/16800333322#artifacts. The non-musl artifact failed the same way, but the musl version seems to be working just fine! I pulled the x86 binary, updated the permissions to `chown root:root` and `chmod 777`, and mapped it to the existing binary name with volumes so I can swap out later when it's updated upstream.

```yaml
# docker-compose.yml
# ...
volumes:
  - ./music-assistant/librespot-linux-musl-x86_64:/app/venv/lib64/python3.12/site-packages/music_assistant/providers/spotify/bin/librespot-linux-x86_64
  - ./music-assistant/librespot-linux-musl-x86_64:/app/venv/lib64/python3.12/site-packages/music_assistant/providers/spotify/bin/librespot-linux-x86_64
```

marcelveldt (approved):

marcelveldt:
Please test the dev addon if he binaries are working correctly

PR #2311, opened 2025-08-10
YTMusic: Auto update dependencies for frequently breaking packages

MarvinSchenkel:
This PR adds some logic to automatically update packages that frequently break when Google changes things on their end. Installing these packages to their latest version on provider.init() ensures that we implement these fixes as soon as they are available, thus not requiring us to push out a patch release every time YT Music is broken.


marcelveldt (approved):
Fine for now, but we must also create some automated test somehow/somewhere to do a nightly check if the dependencies did not make any backward incompatible changes.

PR #2315, opened 2025-08-10
Add podcast functionality to Spotify

OzGav:
Has been tested to work locally. 

Only obvious issue is when adding a podcast to the library in MA it doesn't get added to the Spotify library. I am assuming there is some call elsewhere in the MA codebase that isn't getting called for podcasts? All other Spotify media types get added and removed from Spotfy without issue.

There is still some debug code in there as there were problems getting the streams to start but that has settled down so maybe it was a Spotify issue. So let me know if that should be removed.

Otherwise, it is working well. Caching is used to speed up the UI episode lists. Played status is pulled from Spotify.


music_assistant/providers/spotify/__init__.py
```diff
@@ -1,1030 +1,888 @@
...
+    MEDIA_TYPE_TO_SEARCH,
+    PODCAST_FEATURES,
+)
+from .helpers import get_config_entries, get_librespot_binary
+from .parsers import parse_album, parse_artist, parse_playlist, parse_podcast, parse_track
+from .podcast_helpers import PodcastManager
+from .streaming import LibrespotStreamer
+
+# Expose functions that Music Assistant expects at module level
+__all__ = ["get_config_entries", "setup"]
```
marcelveldt: can we keep the "get_config_entries" and "setup" here  in the init file to keep consistent with all providers ?

music_assistant/providers/spotify/__init__.py
```diff
@@ -1,1030 +1,888 @@
...
+    config: ProviderConfig,
+) -> SpotifyProvider:
+    """Initialize provider(instance) with given configuration."""
+    if config.get_value(CONF_REFRESH_TOKEN) in (None, ""):
+        msg = "Re-Authentication required"
+        raise SetupFailedError(msg)
+    return SpotifyProvider(mass, manifest, config)
+
+
+class SpotifyProvider(MusicProvider):
```
marcelveldt: If you want to split things up, I would move this into a provider.py instead.

music_assistant/providers/spotify/__init__.py
```diff
@@ -1,1030 +1,888 @@
...
+    """Implementation of a Spotify MusicProvider with Podcast support - Modular Architecture."""
```
marcelveldt:
```suggestion
    """Implementation of a Spotify MusicProvider."""

```

music_assistant/providers/spotify/streaming.py
```diff
@@ -0,0 +1,357 @@
...
+                        self.logger.warning(f"{err_msg} - will retry once")
+                        continue
+
+    async def _stream_episode_via_librespot(
+        self, streamdetails: StreamDetails, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """Stream episode using librespot with retry logic."""
+        self.logger.info(f"Episode {streamdetails.item_id}")
+        spotify_uri = f"spotify:episode:{streamdetails.item_id}"
+
```
marcelveldt: This is a lot of duplicated code. Other than the media type (and thus the uri) there is no difference between streaming a track and an episode. So this should be combined into one function

music_assistant/providers/spotify/streaming.py
```diff
@@ -0,0 +1,357 @@
...
+        except asyncio.CancelledError:
+            pass
+        except Exception as e:
+            self.logger.debug(f"Error monitoring stderr: {e}")
+
+        return {
+            "audio_key_errors": audio_key_errors,
+            "decoder_errors": decoder_errors,
+            "cdn_issues": cdn_issues,
+        }
```
marcelveldt: let's get rid of all of this

music_assistant/providers/spotify/helpers.py
```diff
@@ -1,32 +1,200 @@
...
+                return librespot_path
+            return None
+        except OSError:
+            return None
+
+    base_path = os.path.join(os.path.dirname(__file__), "bin")
+    system = platform.system().lower().replace("darwin", "macos")
+    architecture = platform.machine().lower()
+
+    if bridge_binary := await check_librespot(
```
marcelveldt: bridge_binary ?

music_assistant/providers/spotify/helpers.py
```diff
@@ -1,32 +1,200 @@
...
+            "so make sure to disable any popup blockers.\n\n"
+            "Also make sure to perform this action from your local network"
+        )
+    elif action == CONF_ACTION_AUTH:
+        return "Authenticated to Spotify. Press save to complete setup."
+    else:
+        return "Authenticated to Spotify. No further action required."
+
+
+async def get_config_entries(
```
marcelveldt: this one should be moved back to the __init__.py

music_assistant/providers/spotify/helpers.py
OzGav: That is the variable name in the existing code?

OzGav: I changed it to librespot_binary

OzGav:
I believe all the review comments have been addressed. I also noticed a few minor issues that have been fixed

OzGav:
Closing in favour of this one that is passing lint. All comments from above I believe are incorporated.

https://github.com/music-assistant/server/pull/2321

PR #2319, opened 2025-08-11
Fix playback with the Web Player for iOS devices 

### LINKED ISSUE #4146: Local playback on iOS devices no longer possible with current beta.

2.6.0b14

Home Assistant OS Addon

- [x] Yes

### The problem

I was hoping that I can test the latest beta which should solve this issue: https://github.com/music-assistant/support/issues/3858

I have installed the normal MA and the beta MA, of course only one of the two is running.
When I run the latest normal MA, I can playback on the iOS device using "This device" (but it repeats after a minute or so, which is the issue that should be fixed with the latest beta).
However when I try to do the same on the latest beta MA, there is simply no playback. Playback just does not start.

When on beta MA and trying to playback on a Windows PC using Firefox with "This device", there is no problem, playback is possible. Issue seems to be on iOS devices only.

### How to reproduce

Install the latest beta and try to play music on the local iOS device.

### Music Providers

Local Music on my NAS

### Player Providers

"This Device" on iOS

### Full log output

[d5369777_music_assistant_beta_2025-07-30T19-00-44.271Z.log](https://github.com/user-attachments/files/21516344/d5369777_music_assistant_beta_2025-07-30T19-00-44.271Z.log)

### Additional information

Nothing to add, thank you

2025.7.4

Home Assistant OS

- [x] Yes

maximmaxim345:
This PR removes special handling of probe requests since they only worked on docker installations of Music Assistant.
With this PR the Web Player now should also work with Home Assistant OS installations of Music Assistant.

Closes https://github.com/music-assistant/support/issues/4146

### DIFF
--- a/music_assistant/providers/builtin_player/player.py
+++ b/music_assistant/providers/builtin_player/player.py
@@ -244,6 +244,12 @@ class BuiltinPlayer(Player):
         if request.method != "GET":
             return resp
 
+        # Check for a client probe request (from an iPhone/iPad)
+        if (range_header := request.headers.get("Range")) and range_header == "bytes=0-1":
+            self.logger.debug("Client is probing the stream.")
+            # We don't early exit here since playback would otherwise never start
+            # when using iOS with Home Assistant OS
+
         media = player.current_media
         if queue is None or media is None:
             raise web.HTTPNotFound(reason="No active queue or media found!")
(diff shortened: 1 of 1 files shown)


marcelveldt:
removed the backport to stable label because this cant be backported since this is touching the refactored code

marcelveldt (approved):

PR #2321, opened 2025-08-11
Add Podcasts to Spotify

OzGav:


music_assistant/providers/spotify/constants.py
```diff
@@ -0,0 +1,110 @@
...
+    args: list[str]
+
+
+# Librespot configuration profiles
+LIBRESPOT_PROFILES: dict[str, LibrespotProfile] = {
+    "track": {"bitrate": "320", "args": ["--disable-audio-cache", "--dither", "none"]},
+    "episode_1": {"bitrate": "160", "args": ["--dither", "none"]},
+    "episode_2": {"bitrate": "160", "args": ["--disable-audio-cache"]},
+    "episode_3": {"bitrate": "96", "args": ["--disable-audio-cache", "--dither", "none"]},
+}
```
marcelveldt, line 92: I dont think this is needed at all

music_assistant/providers/spotify/podcast_helpers.py
```diff
@@ -0,0 +1,485 @@
...
+        page_size = 50
+        offset = 0
+
+        while True:
+            episodes_data = await self.provider._get_data(
+                f"shows/{prov_podcast_id}/episodes",
+                limit=page_size,
+                offset=offset,
+                market="from_token",  # Critical for getting user's resume_point data
+            )
```
marcelveldt, line 157: why not use the "get_all_items" helper ?

OzGav: The drawbacks are that get_all_items isn't currently setup to support caching like this one is. Also from the perspective of memory efficiency, get_all_items would require all episodes to be fetched into memory first which isn't ideal with a podcasts which can have hundreds of episodes.

music_assistant/providers/spotify/constants.py
OzGav: This came about because in my initial testing I had a lot of stream failures. Even now it is often that the first attempt to stream fails. I would like to keep this in for now and I can revisit during the more extensive testing that I will do when adding audiobooks?

marcelveldt: This PR has become a beast to review because it contains too many changes at once. Its both a refactor into smaller modules and also adding podcasts support. That should have been at least 2 smaller PR's. Also it adds caching for no particular reason as core can already handle caching. 

So better would be to first cover al basic functionality by implementing the core methods that are needed for podcasts (get_podcasts, get_podcast, get_podcast_epiodes etc.)

That way I can also help with testing bit by bit. 
In any case this logic here has nothing to do with any stream errors, we would have to dig into that first step by step

music_assistant/providers/spotify/podcast_helpers.py
marcelveldt: No,  "_get_all_items" is a generator so instead of loading all items into memory it just yields them one by one for further processing.

music_assistant/providers/spotify/podcast_helpers.py
```diff
@@ -0,0 +1,485 @@
...
+                            # Mark with resume position but not fully played
+                            await self.mass.music.mark_item_played(
+                                episode,
+                                fully_played=False,
+                                seconds_played=episode.resume_position_ms // 1000,
+                            )
+                    except Exception as e:
+                        self.logger.warning(
+                            f"Failed to sync played status to MA for episode {episode.item_id}: {e}"
+                        )
```
marcelveldt, line 207: This is an unexpected side effect that really may not be existing here

music_assistant/providers/spotify/podcast_helpers.py
```diff
@@ -0,0 +1,485 @@
...
+                    ):
+                        stats["synced_positions"] += 1
+
+                except Exception as e:
+                    self.logger.warning(f"Error syncing episode {episode.item_id}: {e}")
+                    stats["errors"] += 1
+
+        except Exception as e:
+            self.logger.error(f"Error during bulk sync for podcast {podcast_id}: {e}")
+            stats["errors"] += 1
```
marcelveldt, line 344: the whole super complicated caching stuff must be dropped in the initial PR

marcelveldt (changes requested):
This PR needs to be split up into smaller digestable parts:

1) refactor/cleanup into smaller parts
2) implement the podcast support in a minimum form 
3) if needed, follow-up with added caching etc.

PR #2328, opened 2025-08-12
Improve Lyrics Availability

OzGav:
Same changes as here https://github.com/music-assistant/server/pull/2293 which I will close plus mypy compatability on metadata.py


music_assistant/controllers/metadata.py
```diff
@@ -207,7 +208,8 @@ def providers(self) -> list[MetadataProvider]:
         """Return all loaded/running MetadataProviders."""
         if TYPE_CHECKING:
             return cast("list[MetadataProvider]", self.mass.get_providers(ProviderType.METADATA))
-        return self.mass.get_providers(ProviderType.METADATA)
+        else:
+            return self.mass.get_providers(ProviderType.METADATA)
```
marcelveldt: this else doesnt make any sense

music_assistant/controllers/metadata.py
```diff
@@ -261,21 +264,38 @@ async def update_metadata(
     ) -> MediaItemType:
         """Get/update extra/enhanced metadata for/on given MediaItem."""
         if isinstance(item, str):
-            item = await self.mass.music.get_item_by_uri(item)
+            retrieved_item = await self.mass.music.get_item_by_uri(item)
+            if not hasattr(retrieved_item, "provider") or not hasattr(retrieved_item, "media_type"):
+                raise RuntimeError("Retrieved item is not a valid MediaItemType")
```
marcelveldt: this is not even possible - I would leave this out

music_assistant/controllers/metadata.py
```diff
@@ -261,21 +264,38 @@ async def update_metadata(
...
+            item = cast("MediaItemType", retrieved_item)
+
+        # Type narrowing to ensure item is MediaItemType
+        if not hasattr(item, "provider") or not hasattr(item, "media_type"):
+            raise RuntimeError("Invalid item type for metadata update")
```
marcelveldt: what else can it be ?

music_assistant/controllers/metadata.py
```diff
@@ -298,19 +318,30 @@ async def get_image_data_for_item(
         )
         if not img_path:
             return None
-        return await self.get_thumbnail(img_path, size)
+        thumbnail = await self.get_thumbnail(img_path, provider="builtin", size=size)
+        # Ensure we return bytes, not str
+        if isinstance(thumbnail, str):
+            return None  # or handle the base64 string case appropriately
```
marcelveldt: ?

music_assistant/controllers/metadata.py
```diff
@@ -298,19 +318,30 @@ async def get_image_data_for_item(
...
         img_type: ImageType = ImageType.THUMB,
         resolve: bool = True,
     ) -> str | None:
         """Get url to image for given media media_item."""
         if not media_item:
             return None
         if isinstance(media_item, ItemMapping):
-            media_item = await self.mass.music.get_item_by_uri(media_item.uri)
+            if not media_item.uri:
+                return None
```
marcelveldt: this is weird and not possible

music_assistant/controllers/metadata.py
```diff
@@ -298,19 +318,30 @@ async def get_image_data_for_item(
...
+            retrieved_item = await self.mass.music.get_item_by_uri(media_item.uri)
```
marcelveldt: oof, lets try to avoid this. retrieving the full mediaitem is a heavy call because it potentially needs to query a provider api.

You can simply return early here if the itemmapping.image has a value

music_assistant/controllers/metadata.py
OzGav: You are right I am still learning about how all this type checking stuff works

music_assistant/controllers/metadata.py
OzGav: Apparently it could be a BrowseFolder so I have made some changes which I think is neater.

music_assistant/controllers/metadata.py
OzGav: Yep bad. Changed to check for the BrowseFolder as before.

music_assistant/controllers/metadata.py
OzGav: OK will return if image is already available now.

music_assistant/controllers/metadata.py
```diff
@@ -298,19 +309,39 @@ async def get_image_data_for_item(
         )
         if not img_path:
             return None
-        return await self.get_thumbnail(img_path, size)
+        thumbnail = await self.get_thumbnail(img_path, provider="builtin", size=size)
+        # Ensure we return bytes, not str
+        if isinstance(thumbnail, str):
+            return None  # or handle the base64 string case appropriately
```
marcelveldt: my previous comment is not resolved. this check here is weird

marcelveldt: I think you simply need to help the type checker a bit because its guaranteed that the output is not str (because the base64 flag is set to false) - this reminds me that we may want to drop that base64 option completely, as far as I know its not used anywhere. If we want we can then add an extra helper method to specifically return a base64 encoded image string

OzGav: I’m away from home for a week so can’t make any changes that need testing. If you have a suggestion to keep this moving along that would be great.

OzGav: What do you think of this solution?

music_assistant/controllers/metadata.py
```diff
@@ -591,12 +620,26 @@ async def _update_track_metadata(self, track: Track, force_refresh: bool = False
...
+                if (
+                    provider.lookup_key == "lrclib"
+                    and track.metadata
+                    and (track.metadata.lyrics or track.metadata.lrc_lyrics)
+                ):
+                    self.logger.debug(
+                        "Lyrics already exist for %s, skipping LRCLIB lookup for this track.",
+                        track.name,
+                    )
+                    continue
```
marcelveldt: Let's put this check in the lrclib provider itself

OzGav: Done

OzGav:
Restarted here after getting into trouble when rebasing https://github.com/music-assistant/server/pull/2357

PR #2332, opened 2025-08-15
Add (Roku) Media Assistant provider

MedievalApple:
This pull request adds support for the [Media Assistant](https://github.com/MedievalApple/Media-Assistant) app on Roku devices. The Media Assistant app was originaly made to fix media playback from Home Assistant in the Roku intergration, after Roku discontinued their built-in playonroku app. I have updated the app to add support for the additional features to make this player provider possible.

Device Support: 
- This would support any Roku on Roku OS V6.0 or higher that has Media Assistant Installed from the channel store or sideloaded.

Note:
- Media Assistant needs to be on version V1.2 or higher (current is V1.2)

Player/Provider Features:
- Automatic discovery of Roku players
- Flow Mode
- Enqueue
- Displaying media metadata
- FLAC support


music_assistant/providers/roku_media_assistant/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    from music_assistant.models import ProviderInstanceType
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    # setup is called when the user wants to setup a new provider instance.
+    # you are free to do any preflight checks here and but you must return
+    # an instance of your provider.
```
marcelveldt: you can ditch the comments

music_assistant/providers/roku_media_assistant/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    action: [optional] action key called from config entries UI.
+    values: the (intermediate) raw values for config entries sent with the action.
+    """
+    # ruff: noqa: ARG001
+    # Config Entries are used to configure the Player Provider if needed.
+    # See the models of ConfigEntry and ConfigValueType for more information what is supported.
+    # The ConfigEntry is a dataclass that represents a single configuration entry.
+    # The ConfigValueType is an Enum that represents the type of value that
+    # can be stored in a ConfigEntry.
+    # If your provider does not need any configuration, you can return an empty tuple.
```
marcelveldt:
```suggestion
```

music_assistant/providers/roku_media_assistant/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+            category="advanced",
+        ),
+        ConfigEntry(
+            key=CONF_AUTO_DISCOVER,
+            type=ConfigEntryType.BOOLEAN,
+            label="Allow automatic Roku discovery",
+            default_value=True,
+            description="Enable automatic discovery of Roku players.",
+            category="advanced",
+        ),
```
marcelveldt, line 64: why would you not allow this ? I mean, with this setting disabled, its a useless provider unles you add some way to manually enter player IP's or something ?

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,480 @@
...
+        # SnapCast and Fully Kiosk.
+
+        # Examples of player providers that optionally use 'flow mode' are Google Cast and
+        # Home Assistant. They provide a config entry to enable flow mode playback.
+
+        # Examples of player providers that natively support enqueuing of items are Sonos,
+        # Slimproto and Google Cast.
+
+        if not (queue := self.mass.player_queues.get_active_queue(self.player_id)):
+            return
```
marcelveldt: why is this ?
This limits playback to queue media only and so blocking any external sources such as those from plugin sources.
You should be able to start playback entirely from the provider media object here

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,480 @@
...
+                app_running = (
+                    device_info.app.app_id == self.provider.config.get_value(CONF_ROKU_APP_ID)
+                    if not device_info.app.screensaver
+                    else False
+                )
+
+            current_duration = 0
+
+            if queue.current_item is not None and queue.current_item.media_item is not None:
+                current_duration = cast("int", queue.current_item.media_item.duration)
```
marcelveldt: you should have to interact with the queue directly.

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,480 @@
...
+                if media.media_type == MediaType.RADIO
+                else media.artist,
+                "albumArt": media.image_url,
+                "songFormat": "flac",
+                "duration": "" if media.duration is None else current_duration,
+                "timeOffset": "" if media.duration is None else (current_duration - media.duration),
+                "isLive": "true" if media.media_type == MediaType.RADIO else "",
+            }
+
+            if queue.flow_mode and queue.current_item:
```
marcelveldt: In flow mode, you will only receive a single audio stream.
unless you have some way to dynamically overwrite the currently playing media or your device supports ICY metadata, there is no way to get the media details.

This implementation here just sets the media info from the first track playing and then it will be wrong when the next track plays.

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,480 @@
...
+                        "albumArt": media.image_url,
+                        "songFormat": "flac",
+                        "duration": media.duration,
+                        "enqueue": "true",
+                    },
+                )
+                self.queued = media
+        except Exception:
+            self.logger.error("Failed to Enqueue Media on: %s", self.name)
+            return
```
marcelveldt, line 249: So, your mediaplayer actually does support enqueuing ? Then you dont need flow mode at all.

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,480 @@
...
+                            int(media_state["position"].split(" ", 1)[0]) / 1000
+                        )
+                        self._attr_elapsed_time_last_updated = time.time()
+                    except Exception:
+                        self.logger.info(
+                            "Playback Position received from %s Was Invalid", self.name
+                        )
+
+                if not (queue := self.mass.player_queues.get_active_queue(self.player_id)):
+                    return
```
marcelveldt: please dont interact with the queue directly from a player implementation.

music_assistant/providers/roku_media_assistant/__init__.py
MedievalApple: I have removed all the template comments.

music_assistant/providers/roku_media_assistant/__init__.py
MedievalApple: I wanted to include the ability to add players by IP. In the `__init__` I return the `CONF_ENTRY_MANUAL_DISCOVERY_IPS` config which allows the entry of manual IP's and then these manual players are setup in the `provider -> loaded_in_mass`. Is this the correct way to do this or is there a better way?

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: I was using the queue to get the true duration of the media track, since the returned duration from the `play_media` function changes when you seek in the song. I did this since the device supports showing the duration and correcting the displayed playback position separate from the stream. For example for showing accurate song duration and playback position and dynamically changing it during a continuous audio stream. Is there a better way to get the true media duration?

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: I left flow mode in even though the device does fully support enqueuing, since I thought it would be a nice option if you want cross-fading since the device has no native support for it. I can remove flow mode if needed.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: I was using the queue to get the media metadata to dynamically update the devices metadata when in flow mode. I wasn't sure if this was the proper way to access this data outside the `play_media` function?

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: The device does support updating the media metadata dynamically without affecting playback. This is done in the player poll.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: That was an accident 🤣, I will get that fixed.

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: Does the device support gapless playback ? In that case add the gapless feature flag and crossfade will work even without flow mode

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: OK, let's first see if the device supports gapless - in that case the flow mode code can be left out.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: The device does not support gapless playback there is a slight buffering delay between songs.

music_assistant/providers/roku_media_assistant/__init__.py
marcelveldt: That is correct - check for an example the sonos provider

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: OK all clear! In that case you indeed need to offer the flow mode as an alternative option (or just enable it by default) to add gapless (and crossfade) support

music_assistant/providers/roku_media_assistant/__init__.py
MedievalApple: Adding players manually by IP is now added.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: This is now fixed.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: I changed this so it no longer interacts with the queue here, I was able to make it work without this. Though I am still curious if there is a way to get the full song duration within the `play_media` function like the one you can get with `queue.current_item.media_item.duration`?

music_assistant/providers/roku_media_assistant/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+  "type": "player",
+  "domain": "roku_media_assistant",
+  "name": "Media Assistant (Roku)",
+  "description": "Support for Roku's running Media Assistant.",
+  "codeowners": ["@medievalapple"],
+  "requirements": ["async-upnp-client==0.45.0","rokuecp==0.19.5"],
```
marcelveldt, line 7: OK, this is interesting as we have the dlna provider with the same requirement. Lets keep an eye on it. Worst case scenario we need to abstract some of the dlna logic out into a helper module

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,371 @@
...
+        self._attr_type = PlayerType.PLAYER
+        self._attr_supported_features = {
+            PlayerFeature.POWER,  # if the player can be turned on/off
+            PlayerFeature.PAUSE,
+            PlayerFeature.VOLUME_MUTE,
+            PlayerFeature.ENQUEUE,
+        }
+        self._attr_volume_muted = False
+        self._attr_volume_level = 100
+
```
marcelveldt:
```suggestion
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,371 @@
...
+            CONF_ENTRY_HTTP_PROFILE,
+        ]
+
+    async def power(self, powered: bool) -> None:
+        """Handle POWER command on the player."""
+        try:
+            device_info = await self.roku.update()
+
+            app_running = False
+
```
marcelveldt:
```suggestion
            app_running = False
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,371 @@
...
+                self._attr_active_source = None
+                if app_running:
+                    await self.roku.remote("home")
+                    await self.roku.remote("power")
+
+            logger = self.provider.logger.getChild(self.player_id)
+            logger.info("Received POWER command on player %s", self.display_name)
+            # update the player state in the player manager
+            self.update_state()
+        except Exception:
```
marcelveldt, line 98: In a follow-up PR, try to narrow down this exception instead of broadly catching everything

music_assistant/providers/roku_media_assistant/manifest.json
MedievalApple: Ok, I based the automatic player discovery on the one from the dlna provider since dlna devices have a similar discovery method to Roku's, both of them use upnp for discovery. Though the upnp lib is only used for automatic discovery of Roku's the rest of the communications is handled through rokuecp.

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: Since the device needs flow mode as an option, is this way of accessing the current media metadata ok or should it be rewritten / removed?

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: Yeah, I'm thinking about this a lot. This is about updating metadata on the player itself when flow mode is enabled. Normally a player does then not support metadata updating but in your case (and the cast provider) it does.

Then there is no other way then to just query the queue state yourself but this only catches the fact that a MA queue is actually playing. If a pluginsource is playing, that will not be catched...

Anyways, I will play with this idea a bit in the cast provider maybe and then you can copy it over to your provider.
For now its fine to just query the queue yourself (but add a check to only do that if the queue is actually active).

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: Yeah, this is all clear now. I missed the fact that the player supports both enqueuing and flow mode (including metadata updates)

music_assistant/providers/roku_media_assistant/player.py
marcelveldt: As explained above, fine to leave for now.

music_assistant/providers/roku_media_assistant/manifest.json
marcelveldt: Yeah, all clear. At some point (if we add more providers that depend on upnp for discovery) we might need to add a global upnp discovery, just like we did for mdns.

For now this is ok

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,368 @@
...
+                "t": "a",
+                "albumName": media.album if media.album is not None else "",
+                "songName": media.title,
+                "artistName": "Music Assistant Radio"
+                if media.media_type == MediaType.RADIO
+                else media.artist
+                if media.artist is not None
+                else "Flow Mode"
+                if self.flow_mode
+                else "Music Assistant",
```
marcelveldt: Add some brackets here to make it more readable

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                app_running = (
+                    device_info.app.app_id == self.provider.config.get_value(CONF_ROKU_APP_ID)
+                    if not device_info.app.screensaver
+                    else False
+                )
+
+            f_media = {
+                "u": media.uri,
+                "t": "a",
+                "albumName": media.album if media.album is not None else "",
```
marcelveldt:
```suggestion
                "albumName": media.album or "",
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                "songName": media.title,
+                "artistName": (
+                    "Music Assistant Radio"
+                    if media.media_type == MediaType.RADIO
+                    else media.artist
+                    if media.artist is not None
+                    else ("Flow Mode" if self.flow_mode else "Music Assistant")
+                ),
+                "albumArt": (
+                    "" if self.flow_mode else media.image_url if media.image_url is not None else ""
```
marcelveldt:
```suggestion
                    "" if self.flow_mode else media.image_url or ""
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                ),
+                "songFormat": "flac",
+                "duration": "" if media.duration is None else media.duration,
```
marcelveldt:
```suggestion
                "duration": media.duration or "",
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                "isLive": (
+                    "true"
+                    if media.media_type == MediaType.RADIO
+                    or media.duration is None
+                    or self.flow_mode
+                    else ""
+                ),
+            }
+
+            if queue := self.mass.player_queues.get_active_queue(self.player_id):
```
marcelveldt: this is a tiny bit dangerous because the playback request could have been originated from a group player or even a different source. So I would specifically check the queue_id within the PlayerMedia.

NOTE: That we've just bumped the models on dev where queue_id is replaced by source_id on the PlayerMedia object

marcelveldt:
```suggestion
            if queue := self.mass.player_queues.get_active_queue(media.source_id):
```

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                        current_duration = cast("int", queue.current_item.media_item.duration)
+
+                    f_media.update(
+                        {
+                            "timeOffset": ""
+                            if media.duration is None
+                            else (current_duration - media.duration),
+                            "duration": "" if media.duration is None else current_duration,
+                        }
+                    )
```
marcelveldt: I'm not understanding why this is needed

music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -0,0 +1,386 @@
...
+                        self._attr_elapsed_time = (
+                            int(media_state["position"].split(" ", 1)[0]) / 1000
+                        )
+                        self._attr_elapsed_time_last_updated = time.time()
+                    except Exception:
+                        self.logger.info(
+                            "Playback Position received from %s Was Invalid", self.name
+                        )
+
+                if not (queue := self.mass.player_queues.get_active_queue(self.player_id)):
```
marcelveldt:
```suggestion
                if not (queue := self.mass.player_queues.get_active_queue(self.current_media.source_id)):
```
See my comment above about the source id on the media

music_assistant/providers/roku_media_assistant/player.py
MedievalApple: Part of the dynamic metadata update the player is able to update the displayed duration and position. and this was to fix an issue with the displayed playback time but I realized the issue was on the Roku side and not the data I was sending from the provider. I fixed it on the Roku app side, So I have removed this section.

marcelveldt:
@MedievalApple can you have a look at the 2 final remaining lint issues ? Then we're good to go

OzGav:
@MedievalApple can you provide some text for the docs? We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

Please provide some good detail on the device support requirements. 

Also does this satisfy these? 
https://github.com/orgs/music-assistant/discussions/1038
https://github.com/orgs/music-assistant/discussions/430

marcelveldt (approved):
Nice work @MedievalApple !

MedievalApple:
@OzGav Here is some text for the docs

I would say this satisfy those two discussions. Since the official playonroku app that allowed you to cast media was discontinued in OS V11.5. And this provides a updated/expanded method for playback on Roku devices similar to the original playonroku.

FEATURES
- Roku's are auto-detected by Music Assistant
- Can play/pause using physical buttons
- Audio quality is lossless 48 kHz/16-bit

CONFIGURATION

- Your Roku must be on Roku OS V9.1 or higher to install the Media Assistant App (Note: The app has only been tested on a minimum of OS V13.0).

1. Install the Media Assistant app from the Roku Channel Store or sideload it on your Roku.

- Roku Channel Store Link (https://channelstore.roku.com/details/625f8ef7740dff93df7d85fc510303b4/media-assistant)

- Sideload Link (https://github.com/MedievalApple/Media-Assistant)

2. If you sideloaded the app, you will need to change the Player Provider Setting in Advanced -> **"App ID of Media Assistant"** to **"dev"**.

3. On newer Roku OS versions, in order for Music Assistant to communicate with the Roku. You must ensure **mobile app control is enabled**. To check this go to your Roku's settings and navigate to **(Settings -> System -> Advanced system settings -> Control by mobile apps -> Network access)** and check to see if **Network access** is set to Enabled.


KNOWN ISSUES/ NOTES
- Physical buttons for skipping do not work.
- Roku's do not support crossfading of audio. If you want crossfade and/or full gapless support, enable the "flow mode" in the player's advanced settings. Enabling flow mode may solve playback issues however you will lose the displayed time remaining.

NOT YET SUPPORTED
- Volume currently has to be controlled on the device.

OzGav:
@MedievalApple Thanks. Can you please confirm this is correct
image

edit: And the docs PR is here if you would like to review it https://github.com/music-assistant/beta.music-assistant.io/pull/43

MedievalApple:
@OzGav Yeah, That all looks correct👍

OzGav:
Hi again @MedievalApple . I am updating the beta docs in preparation for the 2.8 release. As we now have the merged player protocol functionality I need to add a SETTINGS section to each player provider but I dont have any Roku devices. Refer to the Cast provider https://beta.music-assistant.io/player-support/google-cast/ Can you PR the beta branch or give me the text that I can paste in? 

Looking at the current docs page those two settings will go into a provider settings section and thus I just need to know what settings are in the output protocol section. Thanks!

MedievalApple:
Hi @OzGav, sorry for the delay response here are the additional/updated settings for the Roku devices.

Provider Advanced Settings:

- Allow automatic Roku discovery is on by default and enables the automatic discovery of Roku devices.

- Manual IP addresses for discovery is used to add device addresses if automatic discovery is off.

- App ID of Media Assistant is only used if the Media Assistant app was sideloaded onto the device.

Device specific settings:

- Enable queue flow mode. Enabling this option will send all tracks as a continuous audio stream. This allows for support of gapless or crossfading. This can also help if your Roku is having have difficulty transitioning between tracks. This does have the side effect of losing some displayed metadata.

- Output channel mode. The default is Stereo but other options are Left channel only, Right channel only or Mono (both channels)

- Sample rates supported by this player. This setting defaults to Roku's stated max sample rates of 44.1kHz / 16 bits and 48kHz / 16 bits but the sample rates and bit depths supported by the Roku can be manually set. Unsupported sample rates may work depending on the Roku device.

- Output codec to use for streaming audio to the player. The default is FLAC but other options are MP3, AAC or WAV. Some codecs may load faster than others depending on the Roku device.

OzGav:
@MedievalApple All good. Thanks!

PR #2338, opened 2025-08-17
Fixes for the squeezelite provider after refactor

marcelveldt:


OzGav:
Looks like this PR may fix https://github.com/music-assistant/support/issues/3704

OzGav:
There is also the muting problem but maybe that is in aioslimproto? https://github.com/music-assistant/support/issues?q=is%3Aissue%20state%3Aopen%20label%3ASqueezelite

PR #2340, opened 2025-08-17
Fix recursion issue after MusicCast Player Refactor

fmunkes:
Remove code which overwrites a property. Causes recursion errors.

### DIFF
--- a/music_assistant/providers/musiccast/player.py
+++ b/music_assistant/providers/musiccast/player.py
@@ -290,6 +290,19 @@ class MusicCastPlayer(Player):
 
         self.update_state()
 
+    @property
+    def synced_to(self) -> str | None:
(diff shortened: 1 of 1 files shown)


marcelveldt:
why does this cause a recursion error ? That shouldt happen. Also, the implementation currently in there seems to be correct and better than the default implementation

fmunkes:
This is not working yet, recursion error is back. I verified, that this line of code
```python
return self._get_player_id_from_zone_device(self.zone_device.group_server)
```
works without issues - it does return a proper player id. I don't get the error at all currently. I'll try again late tomorrow.

fmunkes:
Found it.

marcelveldt (approved):

PR #2341, opened 2025-08-18
Fix missing file breaks queue

OzGav:
Hopefully fixes https://github.com/orgs/music-assistant/discussions/3486

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -1546,8 +1546,8 @@ class LocalFileSystemProvider(MusicProvider):
             except (FileNotFoundError, OSError) as err:
                 # Log the missing file and raise MediaNotFoundError so the queue can skip it
                 self.logger.warning(
-                    "File not found during resolve: %s - Error: %s", 
-                    absolute_path, 
+                    "File not found during resolve: %s - Error: %s",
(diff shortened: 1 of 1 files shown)


marcelveldt:
This is the wrong location to fix this. You should handle exceptions where you need to handle it - not this low level

OzGav:
Should it be in the get_stream_details method for the file system provider then?

marcelveldt:
> Should it be in the get_stream_details method for the file system provider then?

Yes, that is the correct way to fix it!
Thanks!

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -841,11 +841,20 @@ async def create_playlist(self, name: str) -> Playlist:
...
+                return await self._get_stream_details_for_audiobook(item_id)
+            if media_type == MediaType.PODCAST_EPISODE:
+                return await self._get_stream_details_for_podcast_episode(item_id)
+            return await self._get_stream_details_for_track(item_id)
+        except FileNotFoundError as err:
+            self.logger.warning(
+                "File not found for media item %s: %s",
+                item_id,
+                str(err),
+            )
```
```suggestion
        except FileNotFoundError:
            self.logger.warning(
                "File not found for media item %s",
                item_id
            )
```

marcelveldt (approved):
Thanks, @OzGav !

PR #2342, opened 2025-08-18
audiobookshelf: Implement more efficient multi-file seeking

tsipinakis:
The previous seeking implementation relied on downloading all audiobook files and concatenating them together (and then doing it again every time the user seeked). With a large book, this very quickly became slow and a resource hog.

A more efficient way is to use the audio track duration information to determine which track to start the playback at.

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -1287,8 +1287,13 @@ async def get_file_stream(
 async def get_multi_file_stream(
     mass: MusicAssistant,  # noqa: ARG001
     streamdetails: StreamDetails,
+    seek_position: int = 0,
 ) -> AsyncGenerator[bytes, None]:
-    """Return audio stream for a concatenation of multiple files."""
+    """Return audio stream for a concatenation of multiple files.
+
+    Arguments:
+    seek_position: The position to seek to in seconds
+    """
     files_list: list[str] = streamdetails.data
     # concat input files
     temp_file = f"/tmp/{shortuuid.random(20)}.txt"  # noqa: S108
@@ -1306,7 +1311,16 @@ async def get_multi_file_stream(
                 bit_depth=streamdetails.audio_format.bit_depth,
                 channels=streamdetails.audio_format.channels,
             ),
-            extra_input_args=["-safe", "0", "-f", "concat", "-i", temp_file],
+            extra_input_args=[
+                "-safe",
+                "0",
+                "-f",
+                "concat",
+                "-i",
+                temp_file,
+                "-ss",
+                str(seek_position),
+            ],
         ):
             yield chunk
     finally:
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -58,6 +58,7 @@ from music_assistant_models.media_items import (
 from music_assistant_models.media_items.media_item import RecommendationFolder
 from music_assistant_models.streamdetails import StreamDetails
 
+from music_assistant.helpers.audio import get_multi_file_stream
 from music_assistant.models.music_provider import MusicProvider
 from music_assistant.providers.audiobookshelf.parsers import (
     parse_audiobook,
@@ -83,6 +84,7 @@ from .constants import (
 from .helpers import LibrariesHelper, LibraryHelper, ProgressGuard
 
 if TYPE_CHECKING:
+    from aioaudiobookshelf.schema.audio import AudioTrack as AbsAudioTrack
     from aioaudiobookshelf.schema.events_socket import LibraryItemRemoved
     from aioaudiobookshelf.schema.media_progress import MediaProgress
     from aioaudiobookshelf.schema.user import User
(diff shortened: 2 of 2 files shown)


OzGav:
@fmunkes FYI

tsipinakis:
(There's no mark as draft button on GH?)

Anyway, reviews are welcome. I'm currently using this at home to chase out any remaining bugs. I'll let you know when I consider this bug-free.

If you see anything obvious in the code feel free to comment :)

fmunkes:
> (There's no mark as draft button on GH?)
> 
> Anyway, reviews are welcome. I'm currently using this at home to chase out any remaining bugs. I'll let you know when I consider this bug-free.
> 
> If you see anything obvious in the code feel free to comment :)

Take your time, no need to hurry.

tsipinakis:
Ok I think this is ready to review now - managed to find some time to field test it today :)
Listened to quite a few chapters & seeked around, no issues found.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -526,6 +525,66 @@ async def _get_stream_details_audiobook(
             allow_seek=True,
         )
 
+    def _get_track_from_position(
+        self, tracks: list[AbsAudioTrack], seek_position: int
+    ) -> tuple[int, int]:
+        for i, track in enumerate(tracks):
+            self.logger.debug(f"Seeking to position {seek_position}")
```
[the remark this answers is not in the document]
tsipinakis: This is useful for comparing with the requested position. Up to you what level of debug logging you want to keep.

music_assistant/helpers/audio.py
```diff
@@ -77,6 +77,8 @@
 CACHE_BASE_KEY: Final[str] = "audio_cache_path"
 CACHE_FILES_IN_USE: set[str] = set()
 
+__all__ = ["get_multi_file_stream"]
+
```
fmunkes: Why do we need the `__all__` here ? The shouldn't be a `from ... import *` anywhere.

music_assistant/helpers/audio.py
```diff
@@ -1287,6 +1289,7 @@ async def get_file_stream(
 async def get_multi_file_stream(
     mass: MusicAssistant,  # noqa: ARG001
     streamdetails: StreamDetails,
+    seek_position: int = 0,
```
fmunkes, line 1290: Please add a unit comment here.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -526,6 +527,72 @@ async def _get_stream_details_audiobook(
             allow_seek=True,
         )
 
+    def _get_track_from_position(
+        self, tracks: list[AbsAudioTrack], seek_position: int
```
fmunkes, line 529: Another unit for me please - or rename it to something like "seek_position_s" to include the unit.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -526,6 +527,72 @@ async def _get_stream_details_audiobook(
...
+    ) -> tuple[int, int]:
+        for i, track in enumerate(tracks):
+            self.logger.debug(f"Seeking to position {seek_position}")
```
fmunkes: Maybe we can shorten the debug log to something like "skipped x tracks to seek to position xy." Otherwise possible error reports of users might be hard to read.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -526,6 +527,72 @@ async def _get_stream_details_audiobook(
...
+                )
+                continue
+
+            position = max(position, 0)
+
+            self.logger.debug(
+                f"Found seeked track {track.title}, offset {offset}, position {position}"
+            )
+            return i, position
+        return 0, 0
```
fmunkes: I think the index is hard to read - what about return the remaining list of tracks directly, and there position?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -526,6 +527,72 @@ async def _get_stream_details_audiobook(
...
+
+    async def get_audio_stream(
+        self, streamdetails: StreamDetails, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """Retrieve the audio track at the requested position."""
+        i, position = self._get_track_from_position(streamdetails.data, seek_position)
```
fmunkes: See above. If you'd like to keep the index, then please name it track_index or similar.

fmunkes (changes requested):
That works nicely! Just added a couple of comments, which should ease future maintainability I think.

marcelveldt:
> The previous seeking implementation relied on downloading all audiobook files and concatenating them together (and then doing it again every time the user seeked).

That is not exactly true as it would download and concatenate the files into a cache for fast local seeking. 
So maybe that in your case there is not enough diskspace to do this local caching.

In any case, a more efficient seeking within the provider itself is always preferred over having to download and concatenate all parts

music_assistant/helpers/audio.py
tsipinakis: Indeed I'm not sure why that's there, removed.

music_assistant/helpers/audio.py
tsipinakis: Done.

music_assistant/providers/audiobookshelf/__init__.py
tsipinakis: Done.

music_assistant/providers/audiobookshelf/__init__.py
tsipinakis: Done

music_assistant/providers/audiobookshelf/__init__.py
tsipinakis: Done, now returns a track list and position

music_assistant/providers/audiobookshelf/__init__.py
tsipinakis: Fixed

OzGav:
@tsipinakis you can mark the PR comments as resolved where you indicate you have fixed the comment. You can decide whether the AI suggestion is valid and resolve that as well. Once all suggestions are resolved you can request another review from Fabian.

edit: also update the branch to make sure there are no introduced problems there.

fmunkes (approved):

fmunkes:
Excellent! Thanks @tsipinakis !

PR #2348, opened 2025-08-20
Add Resonate Provider

maximmaxim345:
# Add Resonate Support to Music Assistant

This PR adds a player provider for the [Resonate Protocol](https://github.com/Resonate-Protocol/spec) built on top of the [aioresonate](https://github.com/Resonate-Protocol/aioresonate) library.

While `aioresonate` and this provider are not feature-complete yet, most of the essential features are working, including:
- Adheres mostly to the current spec (https://github.com/Resonate-Protocol/aioresonate/pull/33)
- Playing music on single and grouped speakers
- Sending metadata, like the current track's name, album, and the album cover image
- Automatically reconnecting to players on restart
- Starting and stopping playback through the player
- Support for uncompressed PCM (used by default for now), Opus, and FLAC audio, selectable through the player's settings.

This provider can already be tested on a Home Assistant Voice Preview Edition running the [WIP ESPHome implementation](https://github.com/esphome/esphome/pull/10183) of the Resonate Protocol.

## Issues

There are still a couple of issues left to solve in future PRs, including but not limited to:
- Volume/Mute support: Currently not fully supported by the spec
- DeviceInfo is empty: Currently not supported by the spec
- The Resonate Specification will still change with minor breaking changes
- `MULTI_DEVICE_DSP` is not yet supported
- Logging is still very verbose, especially in `aioresonate`
- Only 48kHz/16bit is supported for now
- The Audio Pipeline doesn't show the encoding used by the player yet
- Output codec settings need to be redone: can only be set per group (by sync leader), ignores the preferred codec reported by the player
- Uncaught exceptions on client disconnects


music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,152 @@
...
+
+    player: PlayerInstance
+    unsub_event_cb: Callable[[], None]
+
+    def __init__(self, provider: ResonateProvider, player_id: str) -> None:
+        """Initialize the Player."""
+        super().__init__(provider, player_id)
+        player = provider.server.get_player(player_id)
+        assert player is not None
+        self.player = player
```
marcelveldt: Not a fan of this player.player - I assume this is the Resonate player object ?
So should we maybe call it "api" or "resonate" or "resonate_player" here ?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,152 @@
...
+        self.logger.debug("Received PlayerEvent: %s", event)
+        match event:
+            case VolumeChangedEvent(volume, muted):
+                self._attr_volume_level = volume
+                self._attr_volume_muted = muted
+                self.update_state()
+            case _:
+                self.logger.error("Unknown resonate player event: %s", event)
+
+    @override
```
marcelveldt: this override should be dropped as its an implementation of an ABC method

music_assistant/providers/resonate/player.py
maximmaxim345: Good idea, changed it to `api`, and also renamed the `server` to `server_api`

music_assistant/providers/resonate/player.py
maximmaxim345: But aren't we still overriding a method here?
My LSP gives this warning when removing that override:
```
Method "volume_set" is not marked as override but is overriding a method in class "Player"
```

marcelveldt: Yeah, that is just an error in your editor then because the base class is an ABC

maximmaxim345: Good to know! Removed the overrides.

music_assistant/providers/resonate/provider.py
```diff
@@ -0,0 +1,123 @@
...
+    @override
+    def supported_features(self) -> set[ProviderFeature]:
+        """Return the features supported by this Provider."""
+        return {
+            ProviderFeature.SYNC_PLAYERS,
+            ProviderFeature.CREATE_GROUP_PLAYER,
+            ProviderFeature.REMOVE_GROUP_PLAYER,
+        }
+
+    @override
```
marcelveldt: These override statements should really not be needed. You are implementing methods of an ABC class.
Maybe some weirdness in your IDE ?

maximmaxim345: Removed them.
Hmm, what LSP do you use? I'm now on VSCode with the default Pyright/Pylance and get [these warnings](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportImplicitOverride.md).
I'll create a PR with better options for Pyright to disable this and some other incorrect warnings for everyone.

marcelveldt: pyright is ignoring the fact that the base class is marked as an ABC class and then its the other way around. You specify only which methods may NOT be overrided.

music_assistant/providers/resonate/icon.svg
marcelveldt, line 1: we should remember to update this later

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        resonate_client = provider.server_api.get_client(player_id)
+        assert resonate_client is not None
+        self.api = resonate_client
+        self.api.disconnect_behaviour = DisconnectBehaviour.STOP
+        self.unsub_event_cb = resonate_client.add_event_listener(self.event_cb)
+        self.unsub_group_event_cb = resonate_client.group.add_event_listener(self.group_event_cb)
+
+        self.logger = self.provider.logger.getChild(player_id)
+        # init some static variables
+        self._attr_name = resonate_client.name
```
marcelveldt, line 83: is the name always static ? The client cant ever change name ?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        self._attr_type = PlayerType.PLAYER
+        self._attr_supported_features = {
+            PlayerFeature.SET_MEMBERS,
+        }
+        self._attr_can_group_with = {provider.lookup_key}
+        self._attr_power_control = PLAYER_CONTROL_NONE
+        self._attr_device_info = DeviceInfo()
```
marcelveldt, line 90: future improvement is sending device info from the client

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        self._attr_volume_level = resonate_client.volume
+        self._attr_volume_muted = resonate_client.muted
```
marcelveldt, line 92: ~~These are dynamic (just like maybe the name as stated above) - I would move this into some method that updates the dynamic attributes and you can call it on some places~~

NVM - I spotted a dedicated event for volume updates

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        self._attr_available = True
+        self._attr_poll_interval = 5  # For metadata updates
+        self._attr_needs_poll = False
```
marcelveldt: this contradicts. needs poll is false while you set a poll interval?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        self.logger.info("Received STOP command on player %s", self.display_name)
+        # We don't care if we stopped the stream or it was already stopped
+        _ = self.api.group.stop()
+
+    async def play_media(self, media: PlayerMedia) -> None:
+        """Play media command."""
+        self.logger.info(
+            "Received PLAY_MEDIA command on player %s with uri %s", self.display_name, media.uri
+        )
+
```
marcelveldt:
```suggestion
        self.logger.debug(
            "Received PLAY_MEDIA command on player %s with uri %s", self.display_name, media.uri
        )
```

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        if muted:
+            self.api.mute()
+        else:
+            self.api.unmute()
+
+    async def stop(self) -> None:
+        """Stop command."""
+        self.logger.info("Received STOP command on player %s", self.display_name)
+        # We don't care if we stopped the stream or it was already stopped
+        _ = self.api.group.stop()
```
marcelveldt:
```suggestion
      self.api.group.stop()
```

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        self._attr_elapsed_time_last_updated = time.time()
+        self._attr_active_source = media.source_id
+        # playback_state will be set by the group state change event
+
+        pcm_format = AudioFormat(
+            content_type=ContentType.PCM_S16LE,
+            sample_rate=48000,
+            bit_depth=16,
+            channels=2,
+        )
```
marcelveldt, line 199: I assume this is hardcoded for now and we will dynamically determine this later based on client capabilities ?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+            )
+
+        output_codec = cast("str", self.config.get_value(CONF_OUTPUT_CODEC, "pcm"))
+
+        # Convert string codec to AudioCodec enum
+        codec_mapping = {
+            "pcm": AudioCodec.PCM,
+            "flac": AudioCodec.FLAC,
+            "opus": AudioCodec.OPUS,
+        }
```
marcelveldt: why not simply do AudioCodec(conf_value_output_codec) ?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+            self.api.group.add_client(player.api)
+            self._attr_group_members.append(player_id)
+        self.update_state()
+
+    async def poll(self) -> None:
+        """Poll player for metadata updates and send to resonate players."""
+        if self.playback_state != PlaybackState.PLAYING:
+            self._attr_needs_poll = False
+            return
+        await self.update_metadata()
```
marcelveldt: not sure why this is needed - arent we doing everything based on eventing ?

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+
+    def _update_media_art(self, image_data: bytes) -> None:
+        image = Image.open(BytesIO(image_data))
+        self.api.group.set_media_art(image)
```
marcelveldt, line 275: do check if this is considered blocking IO, otherwise you might need to wrap it into an executor

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        # Send metadata to the group
+        self.api.group.set_metadata(metadata)
+
+    async def get_config_entries(self) -> list[ConfigEntry]:
+        """Return all (provider/player specific) Config Entries for the player."""
+        # OPTIONAL
+        # this method is optional and should be implemented if you need player specific
+        # configuration entries. If you do not need player specific configuration entries,
+        # you can leave this method out completely to accept the default implementation.
+        # Please note that you need to call the super() method to get the default entries.
```
marcelveldt:
```suggestion
```

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+        default_entries = await super().get_config_entries()
+
```
marcelveldt:
```suggestion
```

music_assistant/providers/resonate/player.py
```diff
@@ -0,0 +1,394 @@
...
+                media.source_id, media.queue_item_id
+            )
+            assert queue is not None  # for type checking
+            assert start_queue_item is not None  # for type checking
+            audio_source = self.mass.streams.get_queue_flow_stream(
+                queue=queue, start_queue_item=start_queue_item, pcm_format=pcm_format
+            )
+        else:
+            # assume url or some other direct path
+            audio_source = get_ffmpeg_stream(
```
marcelveldt, line 232: So now we have a single audio-source (pcm stream) into the Resonate group.
But what if we want to do some DSP processing or whatnot ? Are we going to do that within the resonate client lib (and we pass some settings/config per player) or are we going to request a pcm stream per player from MA ?

music_assistant/providers/resonate/player.py
maximmaxim345: At least with the current spec, yes.
I guess if the client wants to change its name, it needs to reconnect (to send a new `client/hello` message).
But this also creates a new `ResonatePlayer` object in Music Assistant.

music_assistant/providers/resonate/player.py
maximmaxim345: Will add it once it gets merged into the spec (https://github.com/Resonate-Protocol/spec/pull/12)

music_assistant/providers/resonate/player.py
maximmaxim345: Correct, but am already working on `aioresonate` to allow asking for the most optimal sample rate is.
16 bit is currently the only sample rate supported by `aioresonate`. Will change later.

music_assistant/providers/resonate/player.py
maximmaxim345: Passing a single stream to all group members was easier to prototype with, and makes `aioresonate` easier to use for applications outside of Music Assistant.
Am working on a major refactor of `aioresonate` right now that will allow Music Assistant to override the audio stream of individual speakers in the group,
while also keeping a single unaltered stream for all other speakers with DSP disabled (for efficiency).

music_assistant/providers/resonate/player.py
maximmaxim345: Do we have another way except polling to get the current metadata with queue flow streams?

music_assistant/providers/resonate/player.py
maximmaxim345: I am calling this in a `asyncio.to_thread` already.

music_assistant/providers/resonate/icon.svg
maximmaxim345: Added a reminder to the Resonate project board.

music_assistant/providers/resonate/player.py
maximmaxim345: Added comments, to explain why.
Done this for quick metadata updates, but polling is only started once we actually play media.

music_assistant/providers/resonate/player.py
maximmaxim345: Done

music_assistant/providers/resonate/player.py
maximmaxim345: Done

music_assistant/providers/resonate/player.py
marcelveldt: yeah, just subscribe to the queue_updated event

music_assistant/providers/resonate/player.py
marcelveldt: OK, sorry I missed that

music_assistant/providers/resonate/player.py
maximmaxim345: Polling is no longer needed.

marcelveldt (approved):
Let's go!
Nice work @maximmaxim345

PR #2349, opened 2025-08-21
Add podcast support to Spotify provider

OzGav:
Things still to do either now or post merging this.

* Use the MA caching system to reduce API calls and speed up the frontend
* Pull down the played status of the episodes during retrieval and sync them with the MA database


music_assistant/providers/spotify/__init__.py
```diff
@@ -127,6 +130,34 @@ async def get_config_entries(
...
         ),
+        ConfigEntry(
+            key=CONF_ENABLE_PODCASTS,
+            type=ConfigEntryType.BOOLEAN,
+            label="Enable Podcast Support",
+            description="Enable support for Spotify podcasts and episodes. "
+            "This will include podcasts in search results and library.",
+            default_value=True,
+            value=values.get(CONF_ENABLE_PODCASTS, True) if values else True,
+        ),
```
marcelveldt: lets not make this configurable - awaiting our new options to control the library import per media type

music_assistant/providers/spotify/__init__.py
```diff
@@ -127,6 +130,34 @@ async def get_config_entries(
...
+            key=CONF_PLAYED_THRESHOLD,
+            type=ConfigEntryType.INTEGER,
+            label="Played Threshold (%)",
+            description="Percentage of episode completion to consider it 'played' "
+            "when not explicitly marked by Spotify (50 = 50%, 90 = 90%).",
+            default_value=90,
+            value=values.get(CONF_PLAYED_THRESHOLD, 90) if values else 90,
+            range=(1, 100),
+        ),
         ConfigEntry(
```
marcelveldt: why make this configurable ?

music_assistant/providers/spotify/constants.py
```diff
@@ -57,3 +60,9 @@
     ProviderFeature.ARTIST_TOPTRACKS,
     ProviderFeature.SIMILAR_TRACKS,
 }
+
+# Podcast Features
+PODCAST_FEATURES = {
+    ProviderFeature.LIBRARY_PODCASTS,
+    ProviderFeature.LIBRARY_PODCASTS_EDIT,
+}
```
marcelveldt: no need for this - the feature is simply always present

music_assistant/providers/spotify/__init__.py
```diff
@@ -127,6 +130,34 @@ async def get_config_entries(
...
+            "This will include podcasts in search results and library.",
+            default_value=True,
+            value=values.get(CONF_ENABLE_PODCASTS, True) if values else True,
+        ),
+        ConfigEntry(
+            key=CONF_SYNC_PLAYED_STATUS,
+            type=ConfigEntryType.BOOLEAN,
+            label="Sync Played Status from Spotify",
+            description="Automatically sync episode played status from Spotify to Music Assistant. "
+            "Episodes marked as played in Spotify will be marked as played in MA.",
```
marcelveldt: disable this option if you never use the spotify app directly to play podcasts

music_assistant/providers/spotify/streaming.py
```diff
@@ -28,18 +29,33 @@ async def get_audio_stream(
...
+
+        # Use explicit media type from streamdetails
+        if hasattr(streamdetails, "media_type") and streamdetails.media_type:
+            is_episode = streamdetails.media_type == MediaType.PODCAST_EPISODE
+        else:
+            # Fallback to bitrate detection if media_type not available
+            is_episode = streamdetails.audio_format.bit_rate == 160
+
+        media_type = "episode" if is_episode else "track"
+        bitrate = "160" if is_episode else "320"
```
marcelveldt: why is this needed ? I'm assuming that librespot is smart enough to pick a lower bitrate if the requested one is not available

music_assistant/providers/spotify/streaming.py
```diff
@@ -28,18 +29,33 @@ async def get_audio_stream(
         self, streamdetails: StreamDetails, seek_position: int = 0
     ) -> AsyncGenerator[bytes, None]:
         """Return the audio stream for the provider item."""
-        spotify_uri = f"spotify://track:{streamdetails.item_id}"
+        # Ensure librespot binary is available
+        if not self.provider._librespot_bin:
+            raise AudioError(f"Librespot binary not available for {streamdetails.item_id}")
```
marcelveldt: this is checked in the init code, you only need to assert it here for the type checker

music_assistant/providers/spotify/provider.py
```diff
@@ -338,6 +445,10 @@ async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool
             await self._delete_data("me/tracks", {"ids": [prov_item_id]})
         elif media_type == MediaType.PLAYLIST:
             await self._delete_data(f"playlists/{prov_item_id}/followers")
+        elif media_type == MediaType.PODCAST:
+            if not self.podcasts_enabled:
+                raise ValueError("Podcast support is disabled")
```
marcelveldt:
```suggestion
```

music_assistant/providers/spotify/provider.py
```diff
@@ -267,6 +336,40 @@ async def get_playlist(self, prov_playlist_id: str) -> Playlist:
         playlist_obj = await self._get_data(f"playlists/{prov_playlist_id}")
         return parse_playlist(playlist_obj, self)
 
+    # Podcast methods (delegate to podcast manager)
```
marcelveldt: not sure I see the real benefit of the podcasts helper here

music_assistant/providers/spotify/provider.py
```diff
@@ -203,6 +263,15 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
             if item and item["track"]["id"]:
                 yield parse_track(item["track"], self)
 
+    async def get_library_podcasts(self) -> AsyncGenerator[Podcast, None]:
+        """Retrieve library podcasts from spotify."""
+        if not self.podcasts_enabled:
+            return
```
marcelveldt:
```suggestion
```

music_assistant/providers/spotify/provider.py
```diff
@@ -166,6 +219,13 @@ async def search(
                     if (item and item["id"])
                 ]
                 items_received += len(api_result["playlists"]["items"])
+            if self.podcasts_enabled and "shows" in api_result:
```
marcelveldt: please ditch all the checks of podcasts are enabled unless there is a very specific reason to NOT include podcasts support

music_assistant/providers/spotify/podcast_helpers.py
```diff
@@ -0,0 +1,149 @@
...
+        episode_obj = await self.provider._get_data(
+            f"episodes/{prov_episode_id}", market="from_token"
+        )
+        if not episode_obj:
+            raise ValueError(f"No episode data returned from API for ID: {prov_episode_id}")
+        return parse_podcast_episode(episode_obj, self.provider)
+
+    async def get_resume_position(self, item_id: str, media_type: MediaType) -> tuple[bool, int]:
+        """
+        Get resume position for episode from Spotify.
```
marcelveldt: So, there is one downside about always overwriting the resume state from spotify which is that you can easily resume if you only play from MA so maybe we should only overwrite from spotify if MA does not have a playstate registered

music_assistant/providers/spotify/__init__.py
OzGav: I have seen a couple of people asking for a similar thing with the scrobbling providers so I thought maybe people will want this?

music_assistant/providers/spotify/__init__.py
OzGav: Yeah I will remove it and to be honest I was probably getting confused with Audiobooks. When doing my research I saw that they are only available in certain countries so I think it would make sense for that type to be able to disable it.

music_assistant/providers/spotify/streaming.py
OzGav: When I was testing this I had so many problems getting a stream to consistently play. 160 was the most likely working configuration although I note it doesn't align with what they say here https://support.spotify.com/au/article/audio-quality. Having said all of that, I was working on this during the Spotify outage so all might be well now. Let me do some more testing using 320.

music_assistant/providers/spotify/provider.py
OzGav: Yes have removed all of that functionality.

music_assistant/providers/spotify/podcast_helpers.py
OzGav: OK so this relates to the earlier suggestion to have an option for those that only listen to podcasts through MA?

music_assistant/providers/spotify/constants.py
OzGav: Removed

music_assistant/providers/spotify/streaming.py
OzGav: fixed

music_assistant/providers/spotify/provider.py
OzGav: Removed

music_assistant/providers/spotify/provider.py
OzGav: I'll obviously defer to you but podcast helpers does have a reasonable amount of podcast specific code that cleans up the main provider file?

music_assistant/providers/spotify/constants.py
OzGav: But I do note that all of the other providers have this?

music_assistant/providers/spotify/__init__.py
OzGav: So are you saying add an option to indicate if you solely use MA for podcasts? Then use that to disable irrelvant functionality like syncing podcast resume points?

music_assistant/providers/spotify/__init__.py
marcelveldt: Sure but next time do this like this in a follow-up PR (if there is user demand) so you can keep the initial implementation small and easier to review

music_assistant/providers/spotify/constants.py
marcelveldt: I meant, no need to put it in a separate constant. You can simply put it in the supported features

OzGav: Ok got it. Added back in.

music_assistant/providers/spotify/streaming.py
OzGav: OK I have tested again and using the same bitrates for tracks and podcasts seems to be working fine now so all that logic has been removed across this PR

music_assistant/providers/spotify/provider.py
OzGav: I did find two methods which weren't value adding at all and removed them.

OzGav: Upon further consideration I have removed the class

music_assistant/providers/spotify/__init__.py
OzGav: Actually I think you meant disable this by default. I have done that and expanded the description.

music_assistant/providers/spotify/podcast_helpers.py
OzGav: I have adjusted the logic here as I am unsure of how to obtain the MA playback state. Can you tell me how to go about that if it is possible?

music_assistant/providers/spotify/__init__.py
```diff
@@ -127,6 +133,27 @@ async def get_config_entries(
...
             hidden=not auth_required,
         ),
+        ConfigEntry(
+            key=CONF_SYNC_PLAYED_STATUS,
+            type=ConfigEntryType.BOOLEAN,
+            label="Sync Played Status from Spotify",
+            description="Automatically sync episode played status from Spotify to Music Assistant. "
+            "Episodes marked as played in Spotify will be marked as played in MA. "
+            "Only enable this if you use both the Spotify app, and Music Assistant "
+            "for podcast playback.",
```
marcelveldt:
```suggestion
            description="Always overwrite episode played status from Spotify to Music Assistant. "
            "Episodes marked as played in Spotify will be marked as played in MA, and the play "
            "state in Music Assistant will always be overwritten /n/n"
            "Only enable this if you (primarily) use the Spotify app, next to Music Assistant "
            "for podcast playback.",
```

music_assistant/providers/spotify/__init__.py
```diff
@@ -127,6 +133,27 @@ async def get_config_entries(
...
+        ),
+        ConfigEntry(
+            key=CONF_PLAYED_THRESHOLD,
+            type=ConfigEntryType.INTEGER,
+            label="Played Threshold (%)",
+            description="Percentage of episode completion to consider it 'played' "
+            "when not explicitly marked by Spotify (50 = 50%, 90 = 90%).",
+            default_value=90,
+            value=values.get(CONF_PLAYED_THRESHOLD, 90) if values else 90,
+            range=(1, 100),
```
marcelveldt, line 155:
```suggestion
            range=(1, 100),
            depends_on= CONF_SYNC_PLAYED_STATUS,
```

music_assistant/providers/spotify/provider.py
```diff
@@ -76,29 +101,39 @@ async def handle_async_init(self) -> None:
...
-        }
+        base = SUPPORTED_FEATURES.copy()
         if not self.custom_client_id_active:
             # Spotify has killed the similar tracks api for developers
             # https://developer.spotify.com/blog/2024-11-27-changes-to-the-web-api
             base.add(ProviderFeature.SIMILAR_TRACKS)
+
+        # Always include podcast features
+        # base.update(PODCAST_FEATURES)
+
```
marcelveldt:
```suggestion
```

music_assistant/providers/spotify/provider.py
```diff
@@ -139,33 +178,45 @@ async def search(
                 "search", q=search_query, type=searchtype, limit=page_limit, offset=offset
             )
             if "artists" in api_result:
-                searchresult.artists += [
+                new_artists = [
```
marcelveldt: whats up with the new_ prefix ?

music_assistant/providers/spotify/provider.py
```diff
@@ -267,6 +330,107 @@ async def get_playlist(self, prov_playlist_id: str) -> Playlist:
         playlist_obj = await self._get_data(f"playlists/{prov_playlist_id}")
         return parse_playlist(playlist_obj, self)
 
+    # Podcast methods (integrated from podcast_helpers.py)
```
marcelveldt:
```suggestion
```

music_assistant/providers/spotify/provider.py
```diff
@@ -267,6 +330,107 @@ async def get_playlist(self, prov_playlist_id: str) -> Playlist:
...
+        """Get full podcast episode details by id."""
+        episode_obj = await self._get_data(f"episodes/{prov_episode_id}", market="from_token")
+        if not episode_obj:
+            raise MediaNotFoundError(f"Episode not found: {prov_episode_id}")
+        return parse_podcast_episode(episode_obj, self)
+
+    async def get_resume_position(self, item_id: str, media_type: MediaType) -> tuple[bool, int]:
+        """Get resume position for episode from Spotify."""
+        if media_type != MediaType.PODCAST_EPISODE:
+            raise NotImplementedError("Resume position only supported for podcast episodes")
```
marcelveldt, line 362: can we at least throw an exception here based on MusicAssistantError ?
Maybe an InvalidDataError ?

music_assistant/providers/spotify/provider.py
```diff
@@ -267,6 +330,107 @@ async def get_playlist(self, prov_playlist_id: str) -> Playlist:
...
+            if not fully_played and episode_obj.get("duration_ms", 0) > 0:
+                completion_ratio = position_ms / episode_obj["duration_ms"]
+                if completion_ratio >= self.played_threshold:
+                    fully_played = True
+
+            return fully_played, position_ms
+
+        except Exception as e:
+            self.logger.debug(f"Failed to get resume position from Spotify for {item_id}: {e}")
+            raise NotImplementedError("Failed to get resume position from Spotify")
```
marcelveldt: can we ditch the outer try...except block ?

music_assistant/providers/spotify/provider.py
```diff
@@ -267,6 +330,107 @@ async def get_playlist(self, prov_playlist_id: str) -> Playlist:
...
+        safe_position = position or 0
+        if media_item.duration > 0:
+            completion_percentage = (safe_position / media_item.duration) * 100
+        else:
+            completion_percentage = 0
+
+        self.logger.debug(
+            f"Episode played in MA: {prov_item_id} "
+            f"({completion_percentage:.1f}%, fully_played: {fully_played}) "
+            f"- Cannot sync back to Spotify due to API limitations"
```
marcelveldt, line 422: we should keep an eye on librespot because I think they are working on fixing the playback reporting

music_assistant/providers/spotify/provider.py
```diff
@@ -505,7 +683,11 @@ async def _get_data(self, endpoint: str, **kwargs: Any) -> dict[str, Any]:
         headers["Accept-Language"] = f"{locale}, {language};q=0.9, *;q=0.5"
         async with (
             self.mass.http_session.get(
-                url, headers=headers, params=kwargs, ssl=True, timeout=120
+                url,
+                headers=headers,
+                params=kwargs,
+                ssl=True,
```
marcelveldt:
```suggestion
```
The ssl arg can be dropped as it has already been set on the http_session itself

music_assistant/providers/spotify/provider.py
OzGav: Can't remember. Must have been a leftover from earlier in the development... Removed.

music_assistant/providers/spotify/provider.py
OzGav: method reworked generally to provide better exception handling

music_assistant/providers/spotify/provider.py
OzGav: fixed

music_assistant/providers/spotify/provider.py
OzGav: As far as I know there just isn't an API call to sync positions back???

marcelveldt: No, but the folks from librespot are working on that - librespot will then sync the playstate to spotify (and we receive it again from the api)

music_assistant/providers/spotify/podcast_helpers.py
marcelveldt: Let's leave it as-is for now and see what is being reported by users. We can always adjust later.

marcelveldt (approved):
Nice work, @OzGav !

PR #2351, opened 2025-08-23
Allow announcement chime to be customised

OzGav:
Satisfies this feature request https://github.com/orgs/music-assistant/discussions/896

streams.py includes some mypy fixes

### DIFF
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import asyncio
 import base64
 import logging
 import os
@@ -11,7 +10,6 @@ from urllib.parse import urlparse
 from uuid import uuid4
 
 import aiofiles
-import aiohttp
 import shortuuid
 from aiofiles.os import wrap
 from cryptography.fernet import Fernet, InvalidToken
@@ -97,44 +95,6 @@ def validate_announcement_chime_url(url: str) -> None:
 
     except Exception as e:
         raise InvalidDataError(f"Invalid URL format: {e!s}") from e
-
-
-async def test_announcement_chime_url(url: str) -> None:
-    """Test if the chime URL is accessible."""
-    if not url or not url.strip():
-        return  # Empty URL is valid
-
-    url = url.strip()
-
-    try:
-        timeout = aiohttp.ClientTimeout(total=10)  # 10 second timeout
-        async with aiohttp.ClientSession(timeout=timeout) as session:
-            try:
-                # Try HEAD first (more efficient)
-                async with session.head(url) as response:
-                    if response.status == 404:
-                        raise InvalidDataError("The file at that URL doesn't seem to exist")
-                    if response.status != 200:
-                        raise InvalidDataError(f"URL returned status code {response.status}")
-            except aiohttp.ClientResponseError as e:
-                if e.status == 405:  # Method not allowed, try GET
-                    async with session.get(url) as response:
-                        if response.status == 404:
-                            raise InvalidDataError("The file at that URL doesn't seem to exist")
-                        if response.status != 200:
-                            raise InvalidDataError(f"URL returned status code {response.status}")
-                if e.status == 404:
-                    raise InvalidDataError("The file at that URL doesn't seem to exist")
-                raise InvalidDataError(f"URL not accessible: HTTP {e.status}")
-
-    except aiohttp.ClientError as e:
-        raise InvalidDataError(f"URL not accessible: {e!s}")
-    except TimeoutError:
-        raise InvalidDataError("URL connection timeout - may be unreachable")
-    except InvalidDataError:
-        raise  # Re-raise InvalidDataError as-is
-    except Exception as e:
-        raise InvalidDataError(f"Error accessing URL: {e!s}")
 
 
 class ConfigController:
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/config.py
```diff
@@ -66,6 +69,76 @@
...
+        audio_extensions = (".mp3", ".wav", ".flac", ".ogg", ".m4a", ".aac")
+        if not any(path_lower.endswith(ext) for ext in audio_extensions):
+            raise InvalidDataError(
+                f"URL must point to an audio file ({', '.join(audio_extensions)})"
+            )
+
+    except InvalidDataError:
+        raise  # Re-raise InvalidDataError as-is
+    except Exception as e:
+        raise InvalidDataError(f"Invalid URL format: {e!s}")
```
marcelveldt:
```suggestion
    except Exception as e:
        raise InvalidDataError(f"Invalid URL format: {e!s}") from e
```

music_assistant/controllers/config.py
```diff
@@ -66,6 +69,76 @@
...
+
+async def test_announcement_chime_url(url: str) -> None:
+    """Test if the chime URL is accessible."""
+    if not url or not url.strip():
+        return  # Empty URL is valid
+
+    url = url.strip()
+
+    try:
+        timeout = aiohttp.ClientTimeout(total=10)  # 10 second timeout
```
marcelveldt: never initiate a new aiohttp client session. Always use the global one that lives in the mass class.
That said, I would not do this skip because it will delay the playback. You will know soon enough if it doesnt work

music_assistant/controllers/config.py
```diff
@@ -700,6 +773,26 @@ async def save_core_config(
...
+            else:
+                # Handle other types by converting to string
+                url = str(url_value)
+
+            # First validate URL format
+            validate_announcement_chime_url(url)
+
+            # Then test if URL is accessible (only if not empty)
+            if url and url.strip():
+                await test_announcement_chime_url(url)
```
marcelveldt: not a fan of this hack

music_assistant/controllers/streams.py
```diff
@@ -250,6 +250,19 @@ async def get_config_entries(
                 category="advanced",
                 required=True,
             ),
+            ConfigEntry(
```
marcelveldt, line 253: One thing to consider, instead of defining it in the core controller is to have it optionally defined in the "play_announcement" call itself. That is also similar to how you can do it in Home Assistant to play an announcement.

Just change "pre_announce" to be None | str | bool where a string can be an url to an pre announce sound or a True boolean to have the default sound.

music_assistant/controllers/config.py
OzGav: test_announcement_chime_url removed

music_assistant/controllers/config.py
OzGav: code simplified

music_assistant/controllers/streams.py
OzGav: I guess I am just answering the current feature request which did not request a per announcement option. There is an argument for simpliciity here I think, in that, this is a simple once off change that someone can make versus having to add the URL to each action. Making the change to the play_announcement call might be breaking? My gut feel is to implement this and see how many people ask for a per announcement option?

marcelveldt: I want to bring the announcement feature to the same level/functionality as the one that is builtin HA for Assist devices. So I do think it makes sense. But it can be a 2 step approach. Global config in Players controller and next is per-player override in the announce call.

But the implementation here is not good. Config needs to be transparant. If we want to do validation, we should extend the ConfigEntry model with an optional validate callback, not hack in the config controller

OzGav: This is starting to sound like this is beyond my current capabilities?

marcelveldt: Yeah, well, I think so... I can help setting this up maybe and you can finish it ?
We will need to start with a change to the models to add an optional verification callback

OzGav: Ok. If the models are updated with the callback then I can see what I am working with and try and go from there?

marcelveldt: This is step 1:
https://github.com/music-assistant/models/pull/109

Then we need to add your url verification method to the helpers and we can assign it to the callback of the config entry on the players controller. Then last step is the play_announcement method itself

OzGav: Ok so then in maybe `util.py ` helper

```
def validate_announcement_chime_url(url: ConfigValueType) -> bool:
    """Validate announcement chime URL format."""
    if not url or url == "":
        return True  # Empty URL is valid
    
    if not isinstance(url, str):
        return False
        
    try:
        parsed = urlparse(url.strip())
        
        if parsed.scheme not in ("http", "https"):
            return False
        
        if not parsed.netloc:
            return False
        
        path_lower = parsed.path.lower()
        audio_extensions = (".mp3", ".wav", ".flac", ".ogg", ".m4a", ".aac")
        return any(path_lower.endswith(ext) for ext in audio_extensions)
        
    except Exception:
        return False
```

???

marcelveldt: yes, but the url would be a string in the typing

OzGav: @marcelveldt Just want to confirm where you said " Global config in Players controller " you meant create a get_config_entries method as there isn't one at the moment?

marcelveldt:
Superseded by https://github.com/music-assistant/server/pull/2403

PR #2354, opened 2025-08-25
Fix erroneous stripping of text from album title

OzGav:
Found an issue where "Mamma Mia! The Movie Soundtrack" was being changed to :Mamma Mia! The Movie" with Soundtrack in the version.


OzGav:
Pre-commit isn't finding anything with the lint so I'm leaving that for now...

PR #2357, opened 2025-08-27
Improve Lyrics Availability

OzGav:
I ended up in some rebasing mypy pain but all fixed now (I think)

The changes here are metadata.py and __init__.py for lrclib. The checking of existing lyrics is now done solely in the lrclib provider as requested.

All the other changes are to get mypy across the line. If there is anything weird in there that won't be surprising....

### DIFF
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -339,7 +339,6 @@ class MetaDataController(CoreController):
                 return None  # can not happen, but guard for type checker
             media_item = cast("MediaItemType", retrieved_item)
 
-        # The rest of the original code, which is correct
         if media_item and media_item.metadata.images:
             for img in media_item.metadata.images:
                 if img.type != img_type:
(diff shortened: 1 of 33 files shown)


music_assistant/controllers/metadata.py
```diff
@@ -206,9 +207,7 @@ async def close(self) -> None:
     @property
     def providers(self) -> list[MetadataProvider]:
         """Return all loaded/running MetadataProviders."""
-        if TYPE_CHECKING:
-            return cast("list[MetadataProvider]", self.mass.get_providers(ProviderType.METADATA))
-        return self.mass.get_providers(ProviderType.METADATA)
+        return cast("list[MetadataProvider]", self.mass.get_providers(ProviderType.METADATA))
```
marcelveldt, line 210: strictly speaking, doing a cast adds a bit of runtime overhead so we try to avoid it as much as we can or hide it between a "if TYPE_CHECKING" check

marcelveldt: In this case, this will not be called that often so its fine to leave it but consider it as an FYI

music_assistant/controllers/metadata.py
```diff
@@ -314,27 +326,39 @@ async def get_image_url_for_item(
...
 
+        if media_item and media_item.metadata.images:
+            for img in media_item.metadata.images:
+                if img.type != img_type:
+                    continue
+                if img.remotely_accessible and not resolve:
+                    continue
+                if img.remotely_accessible and resolve:
+                    return self.get_image_url(img)
+                return img.path
```
marcelveldt:
```suggestion
                if not img.remotely_accessible and not resolve:
                    # ignore image if its not remotely accessible and we don't allow resolving
                    continue
                return self.get_image_url(img, prefer_proxy=not img.remotely_accessible)
```

music_assistant/controllers/metadata.py
OzGav: Thanks. This is new to me but I’ll take it onboard to avoid this in future.

marcelveldt (approved):
Thanks, @OzGav !

PR #2359, opened 2025-08-27
Bluesound native grouping and control of external sources

JoProbst:
Implemented native grouping for Bluesound devices, as well as access/display of local player sources.
Those at work for the Powernode N330 and the Pulse Mini P200.
I could not test anything against other devices and mostly test Tidal, Radio Paradise, HDMI and Bluetooth playback.
I set the default poll interval to 10 seconds to faster reflect changes that are made over the BluOS app, hope that's okay.

For the player sources I'm combining the native inputs with the custom presets that can be set by the user.

### DIFF
--- a/music_assistant/providers/bluesound/const.py
+++ b/music_assistant/providers/bluesound/const.py
@@ -1,0 +1,113 @@
+"""Constants for the Bluesound provider."""
+
+from __future__ import annotations
+
+from music_assistant_models.enums import PlaybackState, PlayerFeature
+
+from music_assistant.models.player import PlayerSource
+
+DEFAULT_POLL_INTERVAL = 10
+
+PLAYER_FEATURES_BASE = {
+    PlayerFeature.SET_MEMBERS,
+    PlayerFeature.VOLUME_MUTE,
+    PlayerFeature.PAUSE,
+    PlayerFeature.SELECT_SOURCE,
+    PlayerFeature.NEXT_PREVIOUS,
+    PlayerFeature.SEEK,
+}
+
+PLAYBACK_STATE_MAP = {
+    "play": PlaybackState.PLAYING,
+    "stream": PlaybackState.PLAYING,
+    "stop": PlaybackState.IDLE,
+    "pause": PlaybackState.PAUSED,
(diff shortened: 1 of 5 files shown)


music_assistant/providers/bluesound/const.py
```diff
@@ -0,0 +1,113 @@
+"""Constants for the Bluesound provider."""
+
+from __future__ import annotations
+
+from music_assistant_models.enums import PlaybackState, PlayerFeature
+
+from music_assistant.models.player import PlayerSource
+
+DEFAULT_POLL_INTERVAL = 10
```
marcelveldt: Let's not hammer these speakers with useless requests every 10 seconds. Their hardware is (very) limited and their webserver is easily overloaded. Next to the fact it causes the speakers to consume more energy.

Stick with a dynamic approach where e.g. you poll them once every minute or 30 seconds when the speaker is off/idle and when playing increase the interval

music_assistant/providers/bluesound/provider.py
```diff
@@ -47,42 +51,32 @@ async def on_mdns_service_state_change(
         self, name: str, state_change: ServiceStateChange, info: AsyncServiceInfo | None
     ) -> None:
         """Handle MDNS service state callback for BluOS."""
+        if state_change == ServiceStateChange.Removed:
+            # Wait for connection to fail, same as sonos.
+            return
```
marcelveldt, line 56: So, and where is this connection timeout handled then ?

music_assistant/providers/bluesound/const.py
JoProbst: Sure, sounds good. I've updated it to have two different polling intervals, one for playback and one for idle. I feel like the faster polling while playing will improve the UX while using external sources a bit.

music_assistant/providers/bluesound/provider.py
JoProbst: Sorry about that. I'm now checking if the status update goes through, if not, the player is set unavailable. Definitely still room for improvement though.

marcelveldt (approved):
LGTM,

Thanks, @JoProbst !

PR #2360, opened 2025-08-28
Switch to static logo images in Radio Paradise provider

OzGav:
While we have the stream metadata it is currently unavailable in the frontend. Also when viewing the radio view or browse it was a bit weird having album images. So hopefully this will now have the static mages there and once the frontend updates are done there will be changing album art

### DIFF
--- a/music_assistant/providers/radioparadise/__init__.py
+++ b/music_assistant/providers/radioparadise/__init__.py
@@ -43,7 +43,9 @@ if TYPE_CHECKING:
     from music_assistant.models import ProviderInstanceType
 
 # Base URL for station icons
-STATION_ICONS_BASE_URL = "https://raw.githubusercontent.com/music-assistant/music-assistant.io/main/docs/assets/icons"
+STATION_ICONS_BASE_URL = (
+    "https://raw.githubusercontent.com/music-assistant/music-assistant.io/main/docs/assets/icons"
+)
 
 # Radio Paradise channel configurations with hardcoded channels
 RADIO_PARADISE_CHANNELS: dict[str, dict[str, Any]] = {
@@ -168,7 +170,6 @@ class RadioParadiseProvider(MusicProvider):
 
         # If API call fails or no current song, static station icon from _parse_radio() is used
         return radio
-
 
     async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
         """Get streamdetails for a radio station."""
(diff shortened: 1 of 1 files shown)


music_assistant/providers/radioparadise/__init__.py
```diff
@@ -238,6 +251,22 @@ def _parse_radio(self, channel_id: str) -> Radio:
...
+                    provider=self.lookup_key,
+                    type=ImageType.THUMB,
+                    path=icon_url,
+                    remotely_accessible=True,
+                )
+            ]
+            radio.metadata.images = UniqueList(images)
+
+        return radio
+
```
marcelveldt: This doesn't make any sense at all

music_assistant/providers/radioparadise/__init__.py
```diff
@@ -145,6 +156,7 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
             current_song = metadata["current"]
             cover_path = current_song.get("cover")
             if cover_path:
+                # Override static station icon with current song cover art
```
marcelveldt: No, you shouldnt override the station icon but you should provide current song art in the streamdetails.stream_metadata

OzGav: Fixed

music_assistant/providers/radioparadise/__init__.py
OzGav: I'm just setting the station icon which I have created. The provider doesn't supply station icons. Is that what you mean?

music_assistant/providers/radioparadise/__init__.py
```diff
@@ -238,6 +229,22 @@ def _parse_radio(self, channel_id: str) -> Radio:
...
+            icon_url = f"{STATION_ICONS_BASE_URL}/{station_icon}"
+            images = [
+                MediaItemImage(
+                    provider=self.lookup_key,
+                    type=ImageType.THUMB,
+                    path=icon_url,
+                    remotely_accessible=True,
+                )
+            ]
+            radio.metadata.images = UniqueList(images)
```
marcelveldt:
```suggestion
            radio.metadata.add_image(
                MediaItemImage(
                    provider=self.lookup_key,
                    type=ImageType.THUMB,
                    path=icon_url,
                    remotely_accessible=True,
                )
            )
```

OzGav: Ok now I see. I saw that MediaItemMetadata had an images field and assumed direct assignment was the way to handle it. I should have looked more carefully at the class methods to see if there were purpose-built functions. Thanks for the solution.

marcelveldt (approved):
Thanks, @OzGav !

PR #2364, opened 2025-09-01
Fix group volume and player state after Players controller refactor

maximmaxim345:
This PR fixes 2 grouping related issues after the player refactor:
1. `group_volume` is now correctly updated after a group members volume was changed
2. Make players removed from a group available for playback (by immediately updating their state that includes `synced_to`)

### DIFF
--- a/music_assistant/controllers/players.py
+++ b/music_assistant/controllers/players.py
@@ -1354,6 +1354,16 @@ class PlayerController(CoreController):
                     # - the leader has DSP enabled
                     self.mass.create_task(self.mass.players.on_player_dsp_change(player_id))
 
+        if ATTR_GROUP_MEMBERS in changed_values:
+            # Removed group members also need to be updated since they are no longer part
+            # of this group and are available for playback again
+            prev_group_members = changed_values[ATTR_GROUP_MEMBERS][0] or []
+            new_group_members = changed_values[ATTR_GROUP_MEMBERS][1] or []
+            removed_members = set(prev_group_members) - set(new_group_members)
+            for player_id in removed_members:
+                if removed_player := self.get(player_id):
+                    self.mass.loop.call_soon(removed_player.update_state, True)
+
         # signal player update on the eventbus
         self.mass.signal_event(EventType.PLAYER_UPDATED, object_id=player_id, data=player)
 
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/players.py
```diff
@@ -1509,7 +1509,11 @@ def iter_group_members(
         exclude_self: bool = True,
     ) -> Iterator[Player]:
         """Get (child) players attached to a group player or syncgroup."""
-        for child_id in list(group_player.group_members):
+        group_members = group_player.group_members.copy()
+        if not exclude_self and group_player.player_id not in group_members:
+            # Always include the group player itself if not excluded
```
marcelveldt: this is weird... so a group player is suddenly going to be included here ?
Let's not do that.

music_assistant/models/player.py
```diff
@@ -997,6 +997,20 @@ def update_state(self, force_update: bool = False) -> None:
         # signal the state update to the PlayerController
         self.mass.players.signal_player_state_update(self, changed_values)
 
+        if group_members_value := changed_values.get("group_members"):
+            # Removed group members also need to be updated since they are no longer part
+            # of this group and are available again for playback
+            prev_group_members, new_group_members = group_members_value
+
```
marcelveldt:
```suggestion
```

music_assistant/models/player.py
```diff
@@ -997,6 +997,20 @@ def update_state(self, force_update: bool = False) -> None:
...
+            removed_members = set(prev_group_members) - set(new_group_members)
+            for member in removed_members:
+                if player := self.mass.players.get(member):
+                    player.update_state()
+        if changed_values.get("volume_level") and self.synced_to:
+            if player := self.mass.players.get(self.synced_to):
+                # We need to update group_volume
+                player.update_state()
```
marcelveldt: this logic is already handled in the player controller.

music_assistant/models/player.py
```diff
@@ -997,6 +997,20 @@ def update_state(self, force_update: bool = False) -> None:
...
 
+        if group_members_value := changed_values.get("group_members"):
+            # Removed group members also need to be updated since they are no longer part
+            # of this group and are available again for playback
+            prev_group_members, new_group_members = group_members_value
+
+            removed_members = set(prev_group_members) - set(new_group_members)
+            for member in removed_members:
+                if player := self.mass.players.get(member):
+                    player.update_state()
```
marcelveldt: this logic should not be here but in the players controller

music_assistant/models/player.py
```diff
@@ -289,7 +289,13 @@ def group_members(self) -> list[str]:
 
         If there are currently no group members, this should return an empty list.
         """
-        return self._attr_group_members
+        if len(self._attr_group_members) >= 1 and self.player_id not in self._attr_group_members:
+            # always ensure the player_id is in the group_members list
+            return [self.player_id, *self._attr_group_members]
```
marcelveldt: I'm missing a check for playertype here. This may only be executed this way if the playertype is player, so not a not group

maximmaxim345: Missed that (again), should be correct now!

marcelveldt (approved):
Thanks, @maximmaxim345

PR #2369, opened 2025-09-02
Strip off Radio Paradise enhanced title text

OzGav:
The enhanced title with upcoming tracks and artists needs to be stripped off for the add to favourite button to work.

### DIFF
--- a/music_assistant/controllers/players.py
+++ b/music_assistant/controllers/players.py
@@ -1113,8 +1113,8 @@ class PlayerController(CoreController):
                     raise PlayerCommandFailed("No current item to add to favorites")
                 # send the streamtitle into a global search query
                 search_artist, search_title_title = stream_title.split(" - ", 1)
-                # strip off the enhanced text added by the Radio Paradise provider
-                search_title_title = search_title_title.split("|")[0].strip()
+                # strip off any additional comments in the title (such as from Radio Paradise)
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/players.py
```diff
@@ -1113,6 +1113,8 @@ async def add_currently_playing_to_favorites(self, player_id: str) -> None:
                     raise PlayerCommandFailed("No current item to add to favorites")
                 # send the streamtitle into a global search query
                 search_artist, search_title_title = stream_title.split(" - ", 1)
+                # strip off the enhanced text added by the Radio Paradise provider
+                search_title_title = search_title_title.split("|")[0].strip()
```
marcelveldt: We should maybe update this cod enow that we have stream_metadata as replacement for the old stream_title.
I dont like having provider specific code in core controllers but this can be generalized a bit

marcelveldt:
```suggestion
                # strip off any additional comments in the title (such as from Radio Paradise)
                search_title_title = search_title_title.split(" | ")[0].strip()
```

OzGav: Yes. Along the same lines I wondered if the radio browser and tune in providers should be updated to use the stream_metadata as well?

marcelveldt (approved):
Thanks, @OzGav

PR #2372, opened 2025-09-03
Fix snapcast grouping after the player model refactor

maximmaxim345:
After the refactor, the Snapcast player provider had several issues related to grouping:

* The `set_members` logic was unable to recover if the state in Music Assistant didn't match the actual Snapcast state.
* Players removed via `set_members` continued to play.
* Stream names were incorrectly identified by the player's name instead of its ID.
* Canceled `_stream_task`s were not being reset to `None`.
* Group updates weren't passed to the player, so it never handled updates like volume when grouped.

Additionally, this PR removes the `ungroup` implementation because it is already automatically derived from `set_members`.

### DIFF
--- a/music_assistant/providers/snapcast/player.py
+++ b/music_assistant/providers/snapcast/player.py
@@ -119,6 +119,7 @@ class SnapCastPlayer(Player):
                 self._stream_task.cancel()
                 with suppress(asyncio.CancelledError):
                     await self._stream_task
+            self._stream_task = None
 
     async def volume_mute(self, muted: bool) -> None:
         """Send MUTE command to given player."""
@@ -169,6 +170,7 @@ class SnapCastPlayer(Player):
                 self._stream_task.cancel()
                 with suppress(asyncio.CancelledError):
                     await self._stream_task
+            self._stream_task = None
 
         # get stream or create new one
         stream_name = self._get_stream_name(SnapCastStreamType.MUSIC)
--- a/music_assistant/providers/snapcast/player.py
+++ b/music_assistant/providers/snapcast/player.py
@@ -137,16 +137,23 @@ class SnapCastPlayer(Player):
         assert group is not None  # for type checking
         # handle client additions
         for player_id in player_ids_to_add or []:
-            if player_id not in group.clients:
-                snapcast_id = self.provider._get_snapclient_id(player_id)
+            snapcast_id = self.provider._get_snapclient_id(player_id)
+            if snapcast_id not in group.clients:
                 await group.add_client(snapcast_id)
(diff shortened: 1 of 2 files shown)


music_assistant/providers/snapcast/player.py
```diff
@@ -391,7 +379,7 @@ def _get_stream_name(self, stream_type: SnapCastStreamType) -> str:
         2. Easily identify which stream belongs to which player, for instance to be able to
            delete a music stream even when it is not active due to an announcement.
         """
-        safe_name = create_safe_string(self.display_name, replace_space=True)
+        safe_name = create_safe_string(self.player_id, replace_space=True)
```
marcelveldt, line 382: sure about this ? this will result in a rather ugly stream name in snapweb

music_assistant/providers/snapcast/provider.py
```diff
@@ -277,7 +276,7 @@ def _handle_group_update(self, snap_group: Snapgroup) -> None:
         for snap_client in self._snapserver.clients:
             if ma_player := self.mass.players.get(self._get_ma_id(snap_client.identifier)):
                 assert isinstance(ma_player, SnapCastPlayer)  # for type checking
-                snap_client.set_callback(ma_player._handle_player_update)
+                ma_player._handle_player_update(snap_client)
```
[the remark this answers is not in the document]
marcelveldt: I think CoPilot is actually right here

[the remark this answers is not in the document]
maximmaxim345: `ma_player._handle_player_update(snap_client)` matches the previous behavior before the refactor.
The callback is already registered in in [the regular handler](https://github.com/music-assistant/server/pull/2372/files/ea51427983269bb228e556c4f4711b610e71721d#diff-21321791b47154d213b69941090081686c81b6c3f4bb3ed2392bac7af1dbf80fL271).
With just the set callback volume updates when grouped never get handled/saved in music assistant.

music_assistant/providers/snapcast/player.py
maximmaxim345: It was set to `self.player_id` before the refactor.
I was first thinking the snapcast provider was completely broken, but turns out it works if all players have unique names.
But if you have multiple snapweb clients (that have all the same name by default), playback gets completely mixed up between players.

But we could also set it to display_name + player_id or display_name + hash(player_id) or so=

marcelveldt: ah ok, like that. In that case leave it like this then

marcelveldt (approved):
Thanks, @maximmaxim345

PR #2378, opened 2025-09-05
Modernise setup script to use uv consistently

Crooked-Krokr:
# Modernise setup script to use uv consistently

## Summary
Modernise setup script to use uv for faster, more reliable dependency management

## Changes
- Add upfront check for uv availability with installation guidance
- Replace mixed pip/uv usage with pure uv throughout
- Addition of pre-commit missing warning
- Improve status reporting to show both Python and uv versions
- Add reminder about venv activation

## Benefits
- **Faster installs**: uv is significantly quicker than pip for large dependency trees
- **Better reliability**: uv handles dependency conflicts more intelligently  
- **Improved developer experience**: Clear error messages and automatic tooling setup
- **Modern tooling**: Aligns with current Python ecosystem best practices

## Testing
- Tested on Ubuntu 22.04 with WSL
- Verified behaviour with and without existing venv
- Confirmed all dependencies install correctly
- pre-commit passed across the board of checks

```bash
(.venv) crooks@LAPTOP-QJ718NJD:/mnt/x/git/server-1$ pre-commit run --all-files
🐶 Ruff Linter...........................................................Passed
🐶 Ruff Formatter........................................................Passed
🐍 Check Python AST......................................................Passed
🔠 Check for case conflicts..............................................Passed
ℹ️  Check docstring is first.............................................Passed
🧐 Check that executables have shebangs..................................Passed
｛ Check JSON files......................................................Passed
💥 Check for merge conflicts.............................................Passed
🔗 Check for broken symlinks.........................(no files to check)Skipped
✅ Check TOML files......................................................Passed
✅ Check code for common misspellings....................................Passed
🕵️  Detect Private Keys..................................................Passed
⮐  Fix End of Files......................................................Passed
🛑 Don't commit to stable branch.........................................Passed
✄  Trim Trailing Whitespace..............................................Passed
mypy.....................................................................Passed
gen_requirements_all.....................................................Passed
```


OzGav:
Simple lint issue to fix

marcelveldt:
While I appreciate the suggestion, I think this is a massive complication of the setup script for a simple issue.
Would it make sense to look into uv as that has native support to specify a python version for the environment and will handle all of this out of the box.

Crooked-Krokr:
> While I appreciate the suggestion, I think this is a massive complication of the setup script for a simple issue. Would it make sense to look into uv as that has native support to specify a python version for the environment and will handle all of this out of the box.

@marcelveldt You're absolutely right about uv being the more elegant solution. I appreciate the feedback and completely agree that my approach added unnecessary complexity.
I'd be happy to rework this using uv's native Python version management, which would be much cleaner and more maintainable. Alternatively, if documentation updates would be more valuable to the project, I could focus there instead.

What would be most useful for the codebase?

marcelveldt:
Well, this PR in it's current shape cant be accepted. Its like shooting a mosquito with a rocket.
If the issue about the wrong versioned python venv is real, we could also consider starting by a warning in the docs and/or setup script log.

But a more elegant way to handle this nowadays is to let uv deal with setting up the env as it can do it all at once.

Crooked-Krokr:
> But a more elegant way to handle this nowadays is to let uv deal with setting up the venv as it can do it all at once.

I see that now, I can't believe how much smoother that was with uv; I couldn't work out how I'd missed it until reading its only 8 months old...  😅
Changes are committed but I've kicked this PR back to draft so I can update the description above tomorrow for review.

scripts/setup.sh
```diff
@@ -1,28 +1,41 @@
...
-uv pip install -r requirements_all.txt
-pre-commit install
+[[ -f requirements_all.txt ]] && uv pip install -r requirements_all.txt
+
+
+# Install pre-commit hooks if pre-commit is available
+if command -v pre-commit &>/dev/null; then
+  pre-commit install
+else
+  echo "⚠️  pre-commit not available. Install with: uv pip install pre-commit"
```
marcelveldt, line 37: pre-commit will be installed with the test requirements isnt it ?

marcelveldt:
This looks much cleaner now indeed.

scripts/setup.sh
Crooked-Krokr: My concern was if the dependency or project file was changed in future; this would flag it. Likely not required though I agree as the pre-commit install would fail regardless.
I can remove this back to just pre-commit install if you would prefer?

Crooked-Krokr:
> This looks much cleaner now indeed.

Thanks, I think the pythonic rust is flaking off now. 😅

marcelveldt (approved):
Thanks, @Crooked-Krokr !

PR #2379, opened 2025-09-07
Audiobookshelf: Implement new JWT authorization

fmunkes:
ABS introduced the JWT authentication system in [v2.26](https://github.com/advplyr/audiobookshelf/discussions/4460), and will remove support for the old authentication system based on permanent tokens not before September, 30th.
This PR updates the lib and the provider to use the new authentication system.

We have two different kind of users, group A uses username and password, group B uses a token (and probably OIDC within ABS).
For group A, presumably the majority, this update is *non-breaking*.

For group B the provider logs a big warning should it detect an old token in combination with a newer version. These users will have to create an API key within ABS themselves. It will be breaking once the old system is removed from ABS.
There are in principle new API endpoints for the API keys, so for group B there might be a non-breaking way, but:
- this would only work for a subset, as the user has to be an admin
- the endpoints are not yet properly documented and might change (I only looked at the source)
- I consider a token user advanced and not afraid of looking at the log/ release notes. MA trying to create tokens in abs might be the opposite of what these users expect.

### Normal API requests
The library automatically detects on login if the new system is used. If so, should the access token expire on a request, it will automatically renew said token with the refresh token. Should the refresh token expire, the lib will raise an RefreshTokenExpiredError.
The provider handles this exception and reauthenticates to renew both tokens if needed.
Of course, for token users, this is all not needed, as they simply provide a long-lasting token.

### FFmpeg
The access token might expire during a stream. Should the user after expiration try to seek, or the next file of a multi-file audiobook is queued, this breaks the stream. I allowed the core ffmpeg helpers to optionally raise an exception, such that this can be handled. A small pause between two streams should remain unnoticed - we are having podcasts and audiobooks.
Also: The default access token expiry is 12 hrs, and refresh token 7 days. Restarting the whole ffmpeg process shouldn't happen to often.

This however forced the use of custom streams, which was added in #2342 - can we backport both PRs?

## Docs
Please update the OIDC section:
If OIDC is configured:
From version 2.26 and above abs uses the JWT token system internally. It is possible to create permanent API keys for an external application. Please follow the ABS docs at https://www.audiobookshelf.org/guides/api-keys/ to create such an API key.

Note that should you insert an old legacy token, your provider will not work anymore once these are removed from ABS.


OzGav:
@fmunkes In regards to the docs change. Do we keep the existing text for users pre-2.26? For example,

```
- If [OIDC](https://www.audiobookshelf.org/guides/oidc_authentication/) is configured:
    - Pre version 2.26, <b>Token instead of user/password.</b> Add the token in the specified field. This token can be obtained by an admin user for any user within the ABS settings -> users
    - From version 2.26 audiobookshelf uses the JWT token system internally. It is possible to create permanent API keys for an external application. Please follow the audiobookshelf docs at https://www.audiobookshelf.org/guides/api-keys/ to create such an API key
    
    
!!! note
    Should you insert an old legacy token, your provider will not work anymore once these are removed from ABS.
```

fmunkes:
Yeah, sounds good let's keep the back reference.

~~Also, the non-passing tests are the results of me accidentally editing the models locally, so I try to create another PR for the FfmpegError later.~~

I realized there is an AudioError already, so I am using that for the Exception now.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -1,17 +1,23 @@
 """Audiobookshelf (abs) provider for Music Assistant."""
+# pyright does not understand the decorator, but mypy runs just fine
+# pyright: reportIncompatibleMethodOverride=false
```
marcelveldt: we don't use (nor support) pyright so please remove this comment

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -190,12 +226,14 @@ def supported_features(self) -> set[ProviderFeature]:
             ProviderFeature.RECOMMENDATIONS,
         }
 
+    @handle_refresh_token
```
marcelveldt: It feels weird to me that the init method is marked with this decorator while the logic for login is setup below in the function itself ?!

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -480,47 +560,27 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
...
-
-        self.logger.debug(
-            f'Using direct playback for audiobook "{abs_audiobook.media.metadata.title}".'
-        )
-        media_url = abs_audiobook.media.tracks[0].content_url
-        stream_url = f"{base_url}{media_url}?token={token}"
-
         return StreamDetails(
-            provider=self.lookup_key,
+            provider=self.instance_id,
```
marcelveldt: why did you change this ? lookup_key is just fine in this case.
This will be automatically set to the instance_id for unique providers

music_assistant/providers/audiobookshelf/manifest.json
```diff
@@ -4,8 +4,12 @@
...
   "name": "Audiobookshelf",
   "description": "Audiobookshelf (audiobookshelf.org) as audiobook and podcast provider",
-  "codeowners": ["@fmunkes"],
-  "requirements": ["aioaudiobookshelf==0.1.7"],
+  "codeowners": [
+    "@fmunkes"
+  ],
+  "requirements": [
+    "aioaudiobookshelf==0.1.8"
+  ],
```
marcelveldt: this formatting doesnt match. Are you using the recommended IDE and linter settings ?

marcelveldt:
Marked as draft, awaiting the review comments to be addressed.
Once these are resolved, please mark the PR ready for review again, thanks!

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Done.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: removed

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: That was a copy/ paste accident.

music_assistant/providers/audiobookshelf/manifest.json
fmunkes: fixed as well I think. I never check the json, that was badly formatted a long time then, didn't change my config. Thanks for pointing that out.

marcelveldt (approved):
Thanks, @fmunkes !

PR #2383, opened 2025-09-08
Fix saving of player settings on powered off players

maximmaxim345:
After the player model refactor, player settings could only be saved after starting playback since `player.active_source` wasn't set yet.

### DIFF
--- a/music_assistant/controllers/players.py
+++ b/music_assistant/controllers/players.py
@@ -1598,8 +1598,9 @@ class PlayerController(CoreController):
             return  # guard against player not being registered (yet)
         player.set_config(config)
         player.update_state()
-        assert player.active_source is not None  # for type checking
-        resume_queue: PlayerQueue | None = self.mass.player_queues.get(player.active_source)
+        resume_queue: PlayerQueue | None = (
+            self.mass.player_queues.get(player.active_source) if player.active_source else None
+        )
         if player_disabled:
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/players.py
```diff
@@ -1598,8 +1598,9 @@ async def on_player_config_change(self, config: PlayerConfig, changed_keys: set[
             return  # guard against player not being registered (yet)
         player.set_config(config)
         player.update_state()
-        assert player.active_source is not None  # for type checking
-        resume_queue: PlayerQueue | None = self.mass.player_queues.get(player.active_source)
+        resume_queue: PlayerQueue | None = (
+            self.mass.player_queues.get(player.active_source) if player.active_source else None
```
marcelveldt, line 1602: Uhm, wouldn't this both result None ?

So, "self.mass.player_queues.get(None)" would also return None

maximmaxim345: Technically yes, but then we have another mypy error when we want to enable it in this part of the repo someday.
But I can change it to `self.mass.player_queues.get(cast("str", player.active_source))` if you like.

marcelveldt: ah yeah, true. Leave it like this.

marcelveldt (approved):
Thanks, @maximmaxim345

PR #2391, opened 2025-09-10
Fix groups after the player model refactor

maximmaxim345:
Fixes around player group support after the large player model refactor.
For a list of all changes and their reasons see the commits.

Frontend fixes are in:
- https://github.com/music-assistant/frontend/pull/1123

Issues left:
- [x] Leaders in `SyncGroupPlayer`s still have the groups queue after powering it off
- [x] `SyncGroupPlayer.supported_features` is not set based on the group members capabilities.
- [x] Switch leader when it gets ungrouped
- [x] Support empty sync groups
- [x] Multiple groups can be activated simultaneously
- [x] `update_state` that doesn't update `synced_to` of other players in some cases
- [x] Groups don't always get powered off it conflicts with another group
- [x] There still a couple of delays waiting for state updates, but they are pretty short now.
- [x] Backup/restore logic on restart of Music Assistant 

Future work (for future PRs):
- Properly support all `supported_features` of the underlying player within `SyncGroupPlayer`.
- State updates sometimes happen multiple times per player (in `signal_player_state_update`).
- Rename `active_group` to `grouped_to` or consolidate `synced_to` and `active_group` to just `grouped_to`
- add a `static_members` to show static group members in the frontend


music_assistant/models/player.py
```diff
@@ -633,13 +643,17 @@ def synced_to(self) -> str | None:
 
         If this player is not synced to another player (or is the sync leader itself),
         this should return None.
+        If this player is a member of a active permanent group, this should return the id of the
+        group.
```
marcelveldt: No, that is not true. This can only return the ID of a"sync leader" - a group is not that

music_assistant/models/player.py
```diff
@@ -633,13 +643,17 @@ def synced_to(self) -> str | None:
...
         """
         # default implementation: feel free to override
         for player in self.mass.players.all():
             if player.player_id == self.player_id:
                 # skip self
                 continue
-            if self.player_id in player.group_members:
+            if self.player_id in player.group_members and (
+                player.powered or player.active_group == PlaybackState.PLAYING
+            ):
```
marcelveldt: why this extra check ?

music_assistant/models/player.py
maximmaxim345: Ok, I'll adjust the default implementation to ignore group players then.

music_assistant/models/player.py
maximmaxim345: Since the refactor the default implementation of `synced_to` always pointed to a group player if there was one.
So the frontend hid all players that were part of the group, no matter if it was powered on or not.
But that won't be needed when I fix `synced_to`.

Wait..., `player.active_group == PlaybackState.PLAYING` makes no sense...

music_assistant/models/player.py
```diff
@@ -1332,10 +1346,17 @@ def __init__(
...
+        # Copy the list so not every added player becomes a static member
         self._attr_device_info = DeviceInfo(model="Sync Group", manufacturer=provider.name)
         self._set_attributes()
 
+    async def on_registered(self) -> None:
+        """Complete the initialization once the player was registered."""
+        # Config is only available after the player was registered
+        self._attr_group_members = list(
+            cast("list[str]", self.config.get_value(CONF_GROUP_MEMBERS, []))
+        )
```
marcelveldt: I'm not fully understanding why this is different then setting it in the init of the player ?

music_assistant/providers/universal_group/player.py
```diff
@@ -162,7 +171,7 @@ async def power(self, powered: bool) -> None:
                         await self.mass.players.cmd_power(member.active_group, False)
                         await asyncio.sleep(1)
                     await asyncio.sleep(1)
-                if member.synced_to:
+                if member.synced_to and member.synced_to != self.player_id:
```
marcelveldt: This can not ever be true so this check is wrong

music_assistant/providers/universal_group/player.py
```diff
@@ -241,6 +251,7 @@ async def play_media(self, media: PlayerMedia) -> None:
         self._attr_elapsed_time = 0
         self._attr_elapsed_time_last_updated = time() - 1
         self._attr_playback_state = PlaybackState.PLAYING
+        self._attr_active_source = media.queue_id
```
marcelveldt, line 255: Reading this I think we should have named "queue_id" in the PlayerMedia "source_id" as this is now super confusing. Let's adjust that in the models and add a fallback property for backwards compatibility

music_assistant/models/player.py
marcelveldt: You should just override it in the group player.

marcelveldt: synced_to (as the name implies!) can only return an id of a player to which a player is synced to, so not an id of a player the player is part of the group. We have "active_group" for that

music_assistant/models/player.py
marcelveldt: yeah, you can leave this out

music_assistant/models/player.py
maximmaxim345: `self.config` is empty at `__init__`. The full config is only loaded after it was registered [here](https://github.com/music-assistant/server/blob/087e36d55e1c8d9c054547ae1e48dc936821cf2b/music_assistant/controllers/players.py#L1186)
Before that `self.config` is pretty much empty, so `group_members` is always empty. See [here](https://github.com/music-assistant/server/blob/087e36d55e1c8d9c054547ae1e48dc936821cf2b/music_assistant/models/player.py#L153C9-L153C92) and [here](https://github.com/music-assistant/server/blob/ebb843a45329fe3455161702987c8e8798ca60e1/music_assistant/controllers/config.py#L389C1-L401C48).

music_assistant/models/player.py
```diff
@@ -1409,60 +1512,126 @@ async def pause(self) -> None:
...
+            if other_group := self.mass.players.get(group):
+                try:
+                    other_group.check_feature(PlayerFeature.SET_MEMBERS)
+                    await other_group.set_members(player_ids_to_remove=[member.player_id])
+                except UnsupportedFeaturedException:
+                    # if the other group does not support SET_MEMBERS or it is a static
+                    # member, we need to power it off to leave the group
+                    await other_group.power(False)
+                # Wait a bit so the state can settle
+                await asyncio.sleep(0.1)
```
maximmaxim345: is it fine to have short sleeps here?
The reason why this is required here is that:
- After the refactor properties like `synced_to` and `active_group`  are computed based on other players
- calling `update_state` doesn't call `update_state` of other players directly, but only through `self.mass.loop.call_soon`
- so if the player is updated, those properties are not yet correct for all players

Or should I change the logic to be able to handle incorrect states?
Or is it better to make `update_state` update all states immediately instead?

music_assistant/models/player.py
```diff
@@ -1409,60 +1512,126 @@ async def pause(self) -> None:
...
+    def _backup_leader_queue(self) -> None:
+        if leader := self.sync_leader:
+            self._leader_backup_active_source = leader.active_source
+            self._leader_backup_current_media = leader.current_media
+        else:
+            self._leader_backup_active_source = None
+            self._leader_backup_current_media = None
+
+    def _restore_leader_queue(self) -> None:
+        if leader := self.sync_leader:
```
maximmaxim345, line 1593: I admit, this is pretty hacky. Do you have an idea how to better solve/refactor this?
I think this workaround is also not working across restarts of Music Assistant.

music_assistant/models/player.py
```diff
@@ -1409,60 +1512,126 @@ async def pause(self) -> None:
...
+            leader.active_source = self._leader_backup_active_source
+            leader.current_media = self._leader_backup_current_media
+        self._leader_backup_active_source = None
+        self._leader_backup_current_media = None
```
maximmaxim345: I admit, this is pretty hacky. Do you have an idea how to better solve/refactor this?
I think this workaround is also not working across restarts of Music Assistant.

maximmaxim345:
After the refactor static member are now members that are always member to that group, as opposed to members that are by default part of that group (like it was before).
One advantage right now is that regular group members and static group members are not differentiated in the frontend UI. 
Meaning: static members are still shown with an checkbox, but clicking it just throws an error message.

music_assistant/models/player.py
```diff
@@ -1409,60 +1512,126 @@ async def pause(self) -> None:
...
+                    other_group.check_feature(PlayerFeature.SET_MEMBERS)
+                    await other_group.set_members(player_ids_to_remove=[member.player_id])
+                except UnsupportedFeaturedException:
+                    # if the other group does not support SET_MEMBERS or it is a static
+                    # member, we need to power it off to leave the group
+                    await other_group.power(False)
+                # Wait a bit so the state can settle
+                await asyncio.sleep(0.1)
+        if (
+            member.synced_to is not None
```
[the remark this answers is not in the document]
maximmaxim345: The code is only called when `sync_leader` is set (and not None): If we have group members that we need to handle, we also have a sync leader.

music_assistant/models/player.py
marcelveldt: why would you need this ?

music_assistant/models/player.py
marcelveldt: I think its better to simply call update_state directly on the child players to prevent race conditions like this. In the end a sleep is also just a bit of guessing

music_assistant/models/player.py
marcelveldt: imo you can simply restore active source and current media to None and the provider implementation will override it if it knows better.

maximmaxim345: ~If you turn off a group (or ungroup the leader), the leader still thinks it is playing the groups stream. The frontend for example then thinks the powered off group is playing for some reason.
This restores whatever was playing before that, so turning on and off a group won't null out what was playing before.~

Edit: Will try that out

maximmaxim345: Works!

music_assistant/models/player.py
maximmaxim345: Done

music_assistant/controllers/players.py
```diff
@@ -1385,13 +1423,18 @@ def signal_player_state_update(
 
         # update/signal group player(s) child's when group updates
         for child_player in self.iter_group_members(player, exclude_self=True):
-            self.mass.loop.call_soon(child_player.update_state, True)
+            child_player.update_state(True)
```
marcelveldt: why force_update set to True ? That will create a redundant event

marcelveldt:
```suggestion
            child_player.update_state()
```

music_assistant/controllers/players.py
```diff
@@ -1385,13 +1423,18 @@ def signal_player_state_update(
...
         # update/signal group player(s) when child updates
         for group_player in self._get_player_groups(player, powered_only=False):
-            self.mass.loop.call_soon(group_player.update_state, True)
+            group_player.update_state(True)
```
marcelveldt:
```suggestion
            group_player.update_state()
```

music_assistant/controllers/players.py
```diff
@@ -1385,13 +1423,18 @@ def signal_player_state_update(
...
         # update/signal manually synced to player when child updates
         if (synced_to := player.synced_to) and (synced_to_player := self.get(synced_to)):
-            self.mass.loop.call_soon(synced_to_player.update_state, True)
+            synced_to_player.update_state(True)
```
marcelveldt:
```suggestion
            synced_to_player.update_state()
```

music_assistant/controllers/players.py
```diff
@@ -1385,13 +1423,18 @@ def signal_player_state_update(
...
+        # update/signal active groups when a group member updates
+        if (active_group := player.active_group) and (
+            active_group_player := self.get(active_group)
+        ):
+            active_group_player.update_state(True)
```
marcelveldt:
```suggestion
            active_group_player.update_state()
```

music_assistant/controllers/players.py
```diff
@@ -1375,7 +1413,7 @@ def signal_player_state_update(
             removed_members = set(prev_group_members) - set(new_group_members)
             for player_id in removed_members:
                 if removed_player := self.get(player_id):
-                    self.mass.loop.call_soon(removed_player.update_state, True)
+                    removed_player.update_state(True)
```
marcelveldt:
```suggestion
                    removed_player.update_state()
```

music_assistant/providers/universal_group/player.py
marcelveldt: I have made this adjustment in the models - will merge after this PR

marcelveldt (approved):
Looks good! Thanks @maximmaxim345 !

PR #2392, opened 2025-09-11
Add caching and fix audiobook podcast mixup

OzGav:
Was adding a first round of caching and during testing realised audiobooks were being returned by Spotify with the podcast API call. These are now filtered out and later a new PR will be done to add audiobook support by using the get audiobook endpoint.

I want to cache the list of episodes as well but that needs more refactoring so will do that in a subsequent PR

### DIFF
--- a/music_assistant/providers/spotify/parsers.py
+++ b/music_assistant/providers/spotify/parsers.py
@@ -210,13 +210,8 @@ def parse_playlist(playlist_obj: dict[str, Any], provider: SpotifyProvider) -> P
     return playlist
 
 
-def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Podcast | None:
+def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Podcast:
     """Parse spotify podcast (show) object to generic layout."""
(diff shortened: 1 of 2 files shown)


music_assistant/providers/spotify/parsers.py
```diff
@@ -210,8 +210,13 @@ def parse_playlist(playlist_obj: dict[str, Any], provider: SpotifyProvider) -> P
     return playlist
 
 
-def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Podcast:
+def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Podcast | None:
     """Parse spotify podcast (show) object to generic layout."""
+    # Filter out audiobooks - they have different characteristics
+    description = podcast_obj.get("description", "")
+    if description.startswith("Author(s):") and "Narrator(s):" in description:
+        return None
```
marcelveldt: I dont particular like this side effect - I think this method should be about parsing. So have the check earlier

OzGav: Logic moved earlier

marcelveldt (approved):
Thanks, @OzGav

PR #2396, opened 2025-09-14
Add command line server startup instructions to the development docs

tavva:


marcelveldt:
May I ask about the purpose of adding this to the dev docs ? 
Debugging is generally done using the (vscode) debugger and not by running the server

tavva:
For those who don't use VS Code.

marcelveldt (approved):

PR #2331, opened 2025-08-14
Add option to control favourite status on streaming provider import

OzGav:
Fixes https://github.com/music-assistant/support/issues/3556

PR #2405, opened 2025-09-16
Add config options to control how library items are synced to MA

marcelveldt:
Add provider configuration options to control (per mediatype) if/how items should be imported into the MA library.

Replaces #2331 

preview:
Scherm­afbeelding 2025-09-21 om 00 40 56


marcelveldt:
@OzGav FYI - maybe you can test this one out and let me know if more options are needed.

OzGav:
I have thought about it and searched the discussions and the only thing I thought of was control of radio station import?

This seems to be working as advertised but I noticed one thing. I was testing with Deezer. There is a track favourited there. This track gets pulled into the database fine but the album associated with it also get added to the database but is not added as a favourite even if the "import and mark as favourite option" is selected. This is possibly desirable behaviour as the album is not in the Deezer "favourites". 

Indeed, the same sort of behaviour occurs if you select do not import everything except for albums. The albums are favourited but the associated artists which are created (not imported) are not. Also artists which are created from playlist contents are not favourited even though "Import and favourite" may have been selected for artists.

I think this makes sense though. WIll add notes to the docs about it if this is how it will work.

Last thing is it desired behaviour to have these options hidden if the provider is disabled? I would probably have them showing as all other items are?

marcelveldt:
> I have thought about it and searched the discussions and the only thing I thought of was control of radio station import?

What do you mean by that ? There is an option added to control that, same as the other media types.

> This track gets pulled into the database fine but the album associated with it also get added to the database but is not added as a favourite even if the "import and mark as favourite option" is selected. 

Yeah, and I think that is exactly what you would want, isn't it ? In the end it's the track that you added to the library and not the track's artist and/or album. What we could do is add some more options to control this behavior but rule of thumb is that we can't NOT add the track artist(s) or album to the library without big refactor due to the relations between objects. If that is the desired behavior we can ofc take that challenge but it will be a bit daunting.

> Last thing is it desired behaviour to have these options hidden if the provider is disabled? I would probably have them showing as all other items are?

These are inserted at runtime based on provider features so the settings can only be visible if the provider is running

OzGav:
I’m not seeing any options on radio providers. This is tunein but radio browser is the same
image

everything else is understood and yes don’t change the behaviour I will just explain it in the docs as there will invariably be questions about it.

music_assistant/controllers/config.py
```diff
@@ -35,6 +35,12 @@
     CONF_DEPRECATED_EQ_BASS,
     CONF_DEPRECATED_EQ_MID,
     CONF_DEPRECATED_EQ_TREBLE,
+    CONF_ENTRY_LIBRARY_IMPORT_ALBUMS,
+    CONF_ENTRY_LIBRARY_IMPORT_ARTISTS,
+    CONF_ENTRY_LIBRARY_IMPORT_AUDIOBOOKS,
+    CONF_ENTRY_LIBRARY_IMPORT_PLAYLISTS,
+    CONF_ENTRY_LIBRARY_IMPORT_PODCASTS,
+    CONF_ENTRY_LIBRARY_IMPORT_TRACKS,
     CONF_ONBOARD_DONE,
```
OzGav, line 56:
```suggestion
    CONF_ENTRY_LIBRARY_IMPORT_RADIOS,
    CONF_ONBOARD_DONE,
```

music_assistant/controllers/config.py
```diff
@@ -241,22 +247,42 @@ async def get_provider_config_entries(
...
+                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_ALBUMS)
+            if provider.supports_feature(ProviderFeature.LIBRARY_TRACKS):
+                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_TRACKS)
+            if provider.supports_feature(ProviderFeature.LIBRARY_PLAYLISTS):
+                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_PLAYLISTS)
+            if provider.supports_feature(ProviderFeature.LIBRARY_AUDIOBOOKS):
+                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_AUDIOBOOKS)
+            if provider.supports_feature(ProviderFeature.LIBRARY_PODCASTS):
+                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_PODCASTS)
+
```
OzGav, line 330:
```suggestion
            if provider.supports_feature(ProviderFeature.LIBRARY_RADIOS):
                extra_entries.append(CONF_ENTRY_LIBRARY_IMPORT_RADIOS)

```

music_assistant/constants.py
```diff
@@ -634,6 +634,254 @@ def create_output_codec_config_entry(
...
+    key="library_import_album_tracks",
+    type=ConfigEntryType.BOOLEAN,
+    label="Import album tracks",
+    description="By default, when importing albums into the library, "
+    "only the Album itself will be imported into the Music Assistant Library, "
+    "allowing you to manually browse and select which tracks you want to import. \n\n"
+    "If you want to override this default behavior, "
+    "you can use this configuration option.\n\n"
+    "Please note that some streaming providers may already define this behavior unsolicited, "
+    "by automatically marking all tracks from the album in their library/favorites.",
```
OzGav:
```suggestion
    "by automatically adding all tracks from the album to their library/favorites.",
```

PR #2406, opened 2025-09-17
Add live and soundtrack album types

OzGav:
This is working well locally. If the approach is sound I will add the inference logic to the other providers.

I did a reasonable amount of research on the inference logic and I believe the number of false positives will be very low. In my testing I got none. Even if there is one it just means one album appears in the filtered list which I think is acceptable given the features overall usefulness.

Frontend changes https://github.com/music-assistant/frontend/pull/1138

### DIFF
--- a/music_assistant/providers/deezer/__init__.py
+++ b/music_assistant/providers/deezer/__init__.py
@@ -6,7 +6,6 @@ from asyncio import TaskGroup
 from collections.abc import AsyncGenerator
 from dataclasses import dataclass
 from math import ceil
-from typing import Any, Literal
 
 import deezer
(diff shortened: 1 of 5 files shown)


music_assistant/providers/deezer/__init__.py
marcelveldt, line 1: please isolate these changes in a dedicated PR

OzGav: I didn’t think it would pass mypy though?

OzGav: done!

marcelveldt (approved):
Thanks, @OzGav !

PR #2407, opened 2025-09-18
Add Smart fades feature

MarvinSchenkel:
This PR adds a first iteration of Smart Fades 🎉 . 

Smart fades aims to improve upon our existing crossfade functionality. This has now been refactored into Smart Fades. We can currently do the following:
- Analyze bpm, beats and downbeats using librosa on preload
- Detect optimal smart fade duration (1, 2, 4 or 8 bars)
- Time stretch the outgoing track to match the incoming track's bpm, if bpms are close enough
- Perform beat alignment by aligning the song's downbeats
- Mix songs using EQ filters: low pass / high pass filters using exponential curves
- Fallback to default crossfade should something go wrong

Current limitations:
- No phrase/phase detection, matching downbeats is not enough for perfect alignment
- Time signature is currently hardcoded at 4/4
- ffmpeg does not natively support frequency sweeps for the EQ filters. This has been approximated by playing with volume

Currently creating this as a draft PR since I expect a fair bit of feedback, but I would love to start the conversation here. Also added a couple of question in-line that I need some help with.

The dev addon is fully up to date and should be a quick way to take this for a spin.

TODO: 
- Lokalize after we agree on the config entries setup

Depends on https://github.com/music-assistant/models/pull/108


MarvinSchenkel:
I have also been thinking about how we can write this in a more OO-way as an `AudioPipeline` with `Filters`. This might make sense when we add more Smart Fades modes later. Something like 
```python
pipeline = AudioPipeline()
pipeline.add(TimeStretch())
pipeline.add(BeatAlignment())
pipeline.add(DJClassicEQFilters())
ffmpeg_args = pipeline.get_ffmpeg_args()
```

Might be a bit overkill for this first version, but keen to hear your opinion on this as well.

music_assistant/constants.py
```diff
@@ -281,17 +284,18 @@
...
-CONF_ENTRY_CROSSFADE = ConfigEntry(
-    key=CONF_CROSSFADE,
+CONF_ENTRY_DEPRECATED_CROSSFADE = ConfigEntry(
+    key=CONF_DEPRECATED_CROSSFADE,
     type=ConfigEntryType.BOOLEAN,
     label="Enable crossfade",
     default_value=False,
     description="Enable a crossfade transition between (queue) tracks.",
     category="audio",
+    hidden=True,  # Hidden, use Smart Fades instead
```
marcelveldt, line 293: why not deprecate and migrate it altogether ?
simply add a bit of migration logic to the config controller to transform crossfade=True into a default option for smart fades ? Then you can simply drop this entire setting

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
     depends_on=CONF_FLOW_MODE,
 )
 
+CONF_ENTRY_SMART_FADES_MODE = ConfigEntry(
+    key=CONF_SMART_FADES_MODE,
+    type=ConfigEntryType.STRING,
+    label="Smart Fades Mode",
```
marcelveldt:
```suggestion
    label="Enable Smart Fades",
```

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
...
+    options=[
+        # TODO: Question: Should we move the enum to models and import it from there ?
```
marcelveldt: you could potentially use the enum values here yeah

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
...
+        ConfigValueOption("Disabled", "disabled"),
+        ConfigValueOption("Smart Fades", "smart_fades"),
+        ConfigValueOption("Default Crossfade", "default_crossfade"),
+    ],
+    default_value="disabled",
+    description="Select the crossfade mode to use when transitioning between tracks.\n\n"
+    "- 'Smart Fades': Uses beat matching and DJ-like EQ filters to create smooth transitions"
+    " between tracks.\n"
+    "- 'Default Crossfade': Regular crossfade that crossfades the last/first x-seconds of a track.",
+    category="smart_fades",
```
marcelveldt: does this need its own category or are we going to stick with the audio category here ?

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
...
+        # TODO: Question: Should we move the enum to models and import it from there ?
+        ConfigValueOption("Disabled", "disabled"),
+        ConfigValueOption("Smart Fades", "smart_fades"),
+        ConfigValueOption("Default Crossfade", "default_crossfade"),
+    ],
+    default_value="disabled",
+    description="Select the crossfade mode to use when transitioning between tracks.\n\n"
+    "- 'Smart Fades': Uses beat matching and DJ-like EQ filters to create smooth transitions"
+    " between tracks.\n"
+    "- 'Default Crossfade': Regular crossfade that crossfades the last/first x-seconds of a track.",
```
marcelveldt:
```suggestion
    "- 'Standard Crossfade': Regular crossfade that crossfades the last/first x-seconds of a track.",
```

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
...
 
+CONF_ENTRY_SMART_FADES_MODE = ConfigEntry(
+    key=CONF_SMART_FADES_MODE,
+    type=ConfigEntryType.STRING,
+    label="Smart Fades Mode",
+    options=[
+        # TODO: Question: Should we move the enum to models and import it from there ?
+        ConfigValueOption("Disabled", "disabled"),
+        ConfigValueOption("Smart Fades", "smart_fades"),
+        ConfigValueOption("Default Crossfade", "default_crossfade"),
```
marcelveldt:
```suggestion
        ConfigValueOption("Standard Crossfade", "default_crossfade"),
```

i would use "regular" or "standard" here instead of the word default

music_assistant/constants.py
```diff
@@ -301,15 +305,57 @@
...
     type=ConfigEntryType.INTEGER,
     range=(1, 15),
     default_value=8,
-    label="Crossfade duration",
-    description="Duration in seconds of the crossfade between tracks (if enabled)",
-    depends_on=CONF_CROSSFADE,
-    category="advanced",
+    label="Fallback crossfade duration",
+    description="Duration in seconds of the fallback crossfade between tracks in case"
+    " Smart Fade Mode = 'Defaul Crossfade' or when a Smart Fade fails",
```
marcelveldt: some typos and adjust the wording a bit

music_assistant/controllers/player_queues.py
```diff
@@ -1147,7 +1150,12 @@ async def _load_item(
         )
         # allow stripping silence from the begin/end of the track if crossfade is enabled
         # this will allow for (much) smoother crossfades
-        if await self.mass.config.get_player_config_value(queue_id, CONF_CROSSFADE):
+        if self.mass.config.get_raw_player_config_value(
+            queue_id, CONF_DEPRECATED_CROSSFADE, False
+        ) or await self.mass.config.get_player_config_value(queue_id, CONF_SMART_FADES_MODE) in (
```
marcelveldt: this can also be simplified if you drop the legacy crossfade setting

music_assistant/constants.py
MarvinSchenkel: That would mean moving it to the models package or defining the enum in constants.py to avoid circular imports. I guess constants.py would be the easiest in this case?

music_assistant/constants.py
MarvinSchenkel: Is there an example of what this might look like? Would save some hassle in the stream controller as well.

music_assistant/constants.py
MarvinSchenkel: Well that's the question. I think it depends on how we want to position this feature. If we plan to expand on this even more (e.g. adding 1 or 2 more options) it might be worth putting them in their own category. If this is it and we don't want to 'market' this feature too much, then I guess the audio category is fine too. Thoughts?

music_assistant/constants.py
MarvinSchenkel: Doesn't this suggest the setting is boolean (on/off)? We can actually pick multiple modes here.

music_assistant/constants.py
marcelveldt: in the config controller there is a migrate method with some existing migration logic

music_assistant/constants.py
marcelveldt: In a autistic world yes but most users will simply read this as "hey, I can enable something here" and then they can select the mode how to enable it.

music_assistant/constants.py
marcelveldt: yep

music_assistant/constants.py
marcelveldt: Let's say that 3 or more options needs its own category (or if they dont belong to an existing category at all) so in this case for discoverability I would say to keep them in the audio category still

music_assistant/helpers/smart_fades.py
```diff
@@ -0,0 +1,1041 @@
...
+
+if TYPE_CHECKING:
+    from music_assistant_models.streamdetails import StreamDetails
+
+    from music_assistant.mass import MusicAssistant
+
+MAX_SMART_CROSSFADE_DURATION = 45
+ANALYSIS_FPS = 100
+ANALYSIS_PCM_FORMAT = AudioFormat(
+    content_type=ContentType.PCM_F32LE, sample_rate=44100, bit_depth=32, channels=2
```
maximmaxim345: Why not directly ask for mono audio here?
I mean, only `_prepare_audio_for_librosa` uses the audio data right now, but it just converts it into mono.
Or are you planning to also do stereo analysis later?

music_assistant/helpers/smart_fades.py
```diff
@@ -0,0 +1,1041 @@
...
+    DISABLED = "disabled"  # No frequency filtering, volume crossfade only
+
+
+@dataclass
+class SmartFadesAnalysis(DataClassDictMixin):
+    """Beat tracking analysis data for BPM matching crossfade."""
+
+    bpm: float
+    beats: np.ndarray[Any, np.dtype[np.float64]]  # Beat positions
+    downbeats: np.ndarray[Any, np.dtype[np.float64]]  # Downbeat positions
```
maximmaxim345: I think you can also use `NDArray[np.float64]` from `numpy.typing`. Than you won't need to use `Any`

music_assistant/helpers/smart_fades.py
```diff
@@ -0,0 +1,1041 @@
...
+    async def _get_audio_bytes_from_stream_details(self, streamdetails: StreamDetails) -> bytes:
+        """Retrieve bytes from the audio stream."""
+        audio_data = b""
+        async for chunk in get_media_stream(
+            self.mass,
+            streamdetails=streamdetails,
+            pcm_format=ANALYSIS_PCM_FORMAT,
+            filter_params=[],
+        ):
+            audio_data += chunk
```
maximmaxim345: You should use `bytearray` instead. Using `bytes` here has a time complexity of O(n^2).

music_assistant/providers/builtin_player/player.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         self._attr_needs_poll = True
         self._attr_poll_interval = POLL_INTERVAL
         self._attr_hidden_by_default = True
-        self._attr_expose_to_ha_by_default = False
+        self._attr_expose_to_ha_by_default = True
```
marcelveldt:
```suggestion
        self._attr_expose_to_ha_by_default = False
```

music_assistant/models/player.py
```diff
@@ -1502,11 +1502,12 @@ async def get_config_entries(self) -> list[ConfigEntry]:
             allowed_conf_entries = (
                 CONF_HTTP_PROFILE,
                 CONF_ENABLE_ICY_METADATA,
-                CONF_CROSSFADE,
+                CONF_DEPRECATED_CROSSFADE,
                 CONF_CROSSFADE_DURATION,
                 CONF_OUTPUT_CODEC,
                 CONF_FLOW_MODE,
                 CONF_SAMPLE_RATES,
+                # TODO: Should we allow the Smart Fades Config entry too ?
```
marcelveldt: yes, otherwise a group playback doesnt have smart fades

music_assistant/helpers/smart_fades.py
```diff
@@ -0,0 +1,879 @@
+"""Smart Fades - Object-oriented implementation with intelligent fades and adaptive filtering."""
+
+# TODO: Figure out if we can achieve shared buffer with StreamController on full
+# current and next track for more EQ options.
```
marcelveldt, line 4: My plan is to rework the whole StreamDetails caching and just have the buffer on the streamdetails object itself of the raw pcm instead of relying on the filesystem.

music_assistant/helpers/smart_fades.py
```diff
@@ -0,0 +1,879 @@
...
+# TODO: Refactor the Analyzer into a metadata controller after we have split the controllers
+# TODO: Refactor the Mixer into a stream controller after we have split the controllers
```
marcelveldt, line 6: Let's make sure we come up with a plan soon. This is a core feature so we either bake it into the core controller (as submodule) or we simply invent something like providertype.core_module so we can isolate everything in provider-like compartments

music_assistant/helpers/smart_fades.py
MarvinSchenkel: Yeah would be great if we can have a single shared buffer for both the stream controller (for playback) and other features like this one. Would be nice if we also have that available for the next song

music_assistant/providers/builtin_player/player.py
MarvinSchenkel: ~~Didn't we just set this to True to make sure the HA entities keep showing up?~~
Nvm, this is the built in player

music_assistant/helpers/smart_fades.py
MarvinSchenkel: Let's discuss soon 👌

marcelveldt (approved):
Looks good, let's merge and then we can always iterate on it later to further enhance it.
Well done, @MarvinSchenkel !

PR #2410, opened 2025-09-21
Add Audiobook Support to Spotify provider

OzGav:


music_assistant/controllers/media/audiobooks.py
```diff
@@ -184,7 +184,7 @@ async def _update_library_item(
                 "narrators": serialize_to_json(
                     update.narrators if overwrite else cur_item.narrators or update.narrators
                 ),
-                "duration": update.duration or update.duration,
+                "duration": update.duration if overwrite else cur_item.duration or update.duration,
```
marcelveldt, line 183: Normally you need to put this in a dedicated bugfix PR - then it can also be backported to stable if needed.
~~For now lets allow it~~

Let's move the changes to the core controllers to a dedicated PR please

OzGav: ok thanks. Will know for next time.

music_assistant/controllers/music.py
```diff
@@ -1053,28 +1053,45 @@ async def get_resume_position(self, media_item: Audiobook | PodcastEpisode) -> t
...
-            {
-                "media_type": media_item.media_type.value,
-                "item_id": media_item.item_id,
-                "provider": media_item.provider,
-            },
-        ):
-            resume_position_ms = (
-                db_entry["seconds_played"] * 1000 if db_entry["seconds_played"] else 0
+            # Type guard: ensure this is a MusicProvider with get_resume_position method
+            if hasattr(provider, "get_resume_position") and callable(provider.get_resume_position):
```
marcelveldt: why not simply assert isinstance MusicProvider here ?

OzGav: Simply because I am new at the typing thing so not always sure about the best way

music_assistant/controllers/music.py
```diff
@@ -1053,28 +1053,45 @@ async def get_resume_position(self, media_item: Audiobook | PodcastEpisode) -> t
...
+                with suppress(NotImplementedError):
```
marcelveldt, line 1088: Maybe we should simply specify a provider feature for this

music_assistant/controllers/music.py
marcelveldt, line 1: this, together with the change to the audiobooks core controller should be moved to a dedicated bugfix PR

OzGav: OK I will do that now.

music_assistant/controllers/music.py
```diff
@@ -1053,28 +1053,45 @@ async def get_resume_position(self, media_item: Audiobook | PodcastEpisode) -> t
...
+                        provider_position_ms,
+                    ) = await provider.get_resume_position(
+                        prov_mapping.item_id, media_item.media_type
+                    )
+                    break  # Use first provider that returns data
+
+        # Get MA's internal position from playlog
+        ma_fully_played = False
+        ma_position_ms = 0
+        if self.database and (
```
marcelveldt: self.database always has a value - this check does not make sense.

music_assistant/controllers/music.py
```diff
@@ -1053,28 +1053,45 @@ async def get_resume_position(self, media_item: Audiobook | PodcastEpisode) -> t
...
+                    "provider": media_item.provider,
+                },
             )
-            return (db_entry["fully_played"], resume_position_ms)
+        ):
+            ma_position_ms = db_entry["seconds_played"] * 1000 if db_entry["seconds_played"] else 0
+            ma_fully_played = db_entry["fully_played"]
 
-        return (False, 0)
+        # Return the higher position to ensure users never lose progress
```
marcelveldt, line 1111: I'm not sure about this.
Its either the provider supporting the accurate progress OR the MA lib as fallback. 
For example, what if you simply re-started playing ? We'll have to carefully consider the possibilities here and I think combined with some PlayerFeature that defines what the provider actually supports

music_assistant/providers/spotify/parsers.py
```diff
@@ -330,3 +331,76 @@ def parse_podcast_episode(
         episode.metadata.preview = episode_obj["audio_preview_url"]
 
     return episode
+
+
+def parse_audiobook(audiobook_obj: dict[str, Any], provider: SpotifyProvider) -> Audiobook:
+    """Parse spotify audiobook object to generic layout."""
+    provider.logger.info(f"DEBUG: parse_audiobook called for {audiobook_obj.get('id', 'unknown')}")
```
marcelveldt:
```suggestion
```

music_assistant/providers/spotify/parsers.py
```diff
@@ -330,3 +331,76 @@ def parse_podcast_episode(
         episode.metadata.preview = episode_obj["audio_preview_url"]
 
     return episode
+
+
+def parse_audiobook(audiobook_obj: dict[str, Any], provider: SpotifyProvider) -> Audiobook:
```
marcelveldt, line 341: As a future improvement to the Spotify provider, we can create TypedDict models for the data coming from the data. Or switch to the spotifyaio library (from Joost) to do the intermediate part

music_assistant/providers/spotify/streaming.py
```diff
@@ -29,15 +29,35 @@ async def get_audio_stream(
...
 
+        async for chunk in self._stream_spotify_uri(spotify_uri, seek_position):
+            yield chunk
+
+    async def get_audio_stream_by_uri(
+        self, spotify_uri: str, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """Return the audio stream for the provider item using a direct Spotify URI."""
+        async for chunk in self._stream_spotify_uri(spotify_uri, seek_position):
+            yield chunk
```
marcelveldt: I'm missing the point of this method completely

music_assistant/controllers/music.py
OzGav: Fixed in new PR

music_assistant/controllers/music.py
OzGav: Ill leave that to your direction!

music_assistant/controllers/music.py
OzGav: Fixed in new PR

music_assistant/controllers/music.py
OzGav: If you were restarting playing then I expect the MA value would be higher?

music_assistant/providers/spotify/streaming.py
OzGav: Yeah me too. A lot of iterations with getting the multipart audiobook streaming to work. This has been removed.

music_assistant/providers/spotify/parsers.py
OzGav: Definately can look at that in the future

music_assistant/controllers/music.py
marcelveldt: yeah, it can be done later

OzGav: Changes are now in the other PR

music_assistant/controllers/music.py
marcelveldt: Lets see where it goes.

music_assistant/providers/spotify/__init__.py
```diff
@@ -171,27 +191,36 @@ async def get_config_entries(
             hidden=auth_required,
         ),
         ConfigEntry(
-            key=CONF_SYNC_PLAYED_STATUS,
+            key=CONF_ENABLE_AUDIOBOOKS,
             type=ConfigEntryType.BOOLEAN,
-            label="Sync Played Status from Spotify",
+            label="Enable Audiobook Support",
```
marcelveldt: why does this need a toggle ? Cant we simply test this

OzGav: Because not all countries support Audiobooks so it is disabled by default. If you live in a supported country then you can enable it. ~~Hmm I just noticed my handy URL to the list of supported countries is gone so I will add that back in... https://support.spotify.com/us/authors/article/audiobooks-availability/~~ Was looking at the wrong entry it is there

OzGav: Did you mean maybe try pinging the API to see if it works and enable if so?

marcelveldt: yes, in the provider init simply try to fetch the audiobooks

marcelveldt: or otherwise simply test the country (from userprofile) to a list of supported countries but then we need to keep updating that list

OzGav: Yeah exactly so I’m not excited about that option. I’ll try and implement the test fetch.

OzGav: OK this is done. It detects that I am in a supported country but I wasn't able to test for someone that isn't.

music_assistant/providers/spotify/__init__.py
```diff
@@ -171,27 +171,27 @@ async def get_config_entries(
...
-            key=CONF_PLAYED_THRESHOLD,
-            type=ConfigEntryType.INTEGER,
-            label="Played Threshold (%)",
-            description="Percentage of episode completion to consider it 'played' "
-            "when not explicitly marked by Spotify (50 = 50%, 90 = 90%).",
-            default_value=90,
-            value=values.get(CONF_PLAYED_THRESHOLD, 90) if values else 90,
-            range=(1, 100),
-            depends_on=CONF_SYNC_PLAYED_STATUS,
+            key=CONF_SYNC_AUDIOBOOK_PROGRESS,
```
marcelveldt, line 191: I would only return this config entry if "provider.audiobooks_supported" is true

music_assistant/providers/spotify/provider.py
```diff
@@ -86,36 +91,50 @@ async def handle_async_init(self) -> None:
         self._librespot_bin = await get_librespot_binary()
         # try login which will raise if it fails
         await self.login()
+        self.audiobooks_supported = await self._test_audiobook_support()
+        if not self.audiobooks_supported:
+            self.logger.info(
+                "Audiobooks are not available in your region. "
```
marcelveldt:
```suggestion
                "Audiobook support disabled: Audiobooks are not available in your region. "
```

music_assistant/providers/spotify/provider.py
```diff
@@ -86,36 +91,50 @@ async def handle_async_init(self) -> None:
...
+            )
+
+    async def _test_audiobook_support(self) -> bool:
+        """Test if audiobooks are supported in user's region."""
+        try:
+            # Use your existing _get_data method instead
+            await self._get_data("me/audiobooks", limit=1)
+            self.logger.debug("Audiobook support detected")
+            return True
+        except Exception as e:
```
marcelveldt: try to narrow down this exception. My best educated guess is that we will get a aiohttp.ClientError with status code 403

OzGav: done

music_assistant/providers/spotify/__init__.py
OzGav: That makes sense but could you give me a pointer as to the best way to do that as I thought that config generation would happen before the API check I am doing in andle_async_init? Some ways I thought of are:
1. Check audiobook availability during config generation using another API test
2. Show configs but disable them if audiobooks aren't available

Im sure there is a better way that isnt apparent to me!

music_assistant/providers/spotify/__init__.py
```diff
@@ -171,27 +171,27 @@ async def get_config_entries(
...
             description="Automatically sync episode played status from Spotify to Music Assistant. "
             "Episodes marked as played in Spotify will be marked as played in MA."
             "Only enable this if you use both the Spotify app and Music Assistant "
             "for podcast playback.",
             default_value=False,
-            value=values.get(CONF_SYNC_PLAYED_STATUS, True) if values else True,
+            value=values.get(CONF_SYNC_PODCAST_PROGRESS, True) if values else True,
             category="sync_options",
         ),
         ConfigEntry(
```
marcelveldt, line 190: do something like this above in the method

audiobooks_supported = instance_id and (prov_instance := mass.get_provider(instance_id)) and prov_instance.audiobooks_supported

then in the configentry you can add hidden=not audiobooks_supported

OzGav: OK that was tricky but I believe I have implemented your idea correctly.

music_assistant/providers/spotify/__init__.py
OzGav: done

marcelveldt (dismissed):
Nice work, @OzGav !

marcelveldt (approved):

PR #2413, opened 2025-09-22
Several small (typing) fixes for the Deezer provider

OzGav:
This started in a different PR but the MyPy changes were a distraction. MyPy is the majority of the PR however there is one bug fix in there and that is in three places similar to this
`url=getattr(artist, "link", None),  # Sometimes the API doesn't return a link`


marcelveldt (approved):
LGTM

PR #2419, opened 2025-09-22
Add DI.fm radio network provider

benklop:
This pull request adds a new AudioAddict provider.


benklop:
this was pretty much entirely AI generated given a similar plugin for kodi as an example, so several grains of salt are likely pretty healthy. It does seem to work for me, for what that's worth.

music_assistant/providers/audioaddict/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+        ):
+            msg = f"{self.domain}: Invalid listen key provided"
+            raise ProviderUnavailableError(msg)
+
+        # Test API connectivity by trying to get channels from first enabled network
+        try:
+            first_network = enabled_networks[0]
+            await self._get_channels(first_network)
+            self.logger.info("%s: Successfully connected to AudioAddict API", self.domain)
+        except Exception as err:
```
OzGav: You should be catching the specific errors you are interested in here and using the MA errors and not these broad catch-alls.

music_assistant/providers/audioaddict/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+        "description": "Classical music radio stations",
+    },
+    "zenradio": {
+        "domain": "zenradio.com",
+        "display_name": "ZenRadio",
+        "description": "Ambient and relaxation radio stations",
+    },
+}
+
+QUALITY_SETTINGS = {
```
OzGav: Just use the highest quality

music_assistant/providers/audioaddict/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+                for channel_data in channels:
+                    channel_name = str(channel_data.get("name", "")).lower()
+                    if search_query_lower in channel_name:
+                        radio = self._channel_to_radio(channel_data, network_key)
+                        radios.append(radio)
+
+                        if len(radios) >= limit:
+                            break
+
+            except Exception as err:
```
OzGav: Too broad

music_assistant/providers/audioaddict/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+    async def get_library_radios(self) -> AsyncGenerator[Radio, None]:
+        """Retrieve all radio stations from active networks."""
+        for network_key in self._get_active_networks():
+            try:
+                channels = await self._get_channels(network_key)
+
+                for channel_data in channels:
+                    yield self._channel_to_radio(channel_data, network_key)
+
+            except Exception as err:
```
OzGav: Too broad

music_assistant/providers/audioaddict/__init__.py
```diff
@@ -0,0 +1,637 @@
...
+    async def _api_request(
+        self,
+        network_key: str,
+        endpoint: str,
+        use_https: bool = True,
+        **params: Any,
+    ) -> Any:
+        """Make a generic API request to AudioAddict."""
+        # Network validation happens in _validate_network_key
+        self._validate_network_key(network_key)
```
OzGav: This is the second time this is validated

OzGav:
Once you get close to merging can you provide some text for the docs? We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

music_assistant/providers/audioaddict/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "music",
+  "domain": "audioaddict",
+  "stage": "beta",
+  "name": "AudioAddict",
+  "description": "AudioAddict Network Radio provider for Music Assistant. Supports DI.FM, RadioTunes, RockRadio, JazzRadio, ClassicalRadio, and ZenRadio streaming services.",
+  "codeowners": ["@benklop"],
+  "requirements": [],
+  "documentation": "https://audioaddict.com/",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/music-providers/difm-network/",
```

OzGav:
The other thing looking into this is that the term audioaddict as a concept seems to no longer exist. audioaddict.com redirects to jazzradio.com. So I’m not sure this is the best name for the provider although I don’t have a better one right now.

OzGav:
Also I am not seeing any use of [stream metadata](https://github.com/music-assistant/models/blob/4078f29b61059fc8447ef53a53a80a6f62d16f3e/music_assistant_models/streamdetails.py#L17)

benklop:
Thanks for all the feedback! Apologies for the rough state. I'll address all this probably this weekend.

benklop:
> The other thing looking into this is that the term audioaddict as a concept seems to no longer exist. audioaddict.com redirects to jazzradio.com. So I’m not sure this is the best name for the provider although I don’t have a better one right now.

I was also looking for a better name, and just wasn't finding anything. Maybe "DI.fm Network" would work, as I'm pretty sure di.fm is probably the most well known and oldest.

marcelveldt:
Marked as draft, awaiting the review feedback to be addressed. Please hit the "Ready for review" button once you are ready for another review. Thanks!

OzGav:
> I was also looking for a better name, and just wasn't finding anything. Maybe "DI.fm Network" would work, as I'm pretty sure di.fm is probably the most well known and oldest.

I think that would work. I just looked at their app store and all of the apps (for the six networks) are published by "Digitally Incorporated Inc"

music_assistant/providers/audioaddict/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "music",
+  "domain": "audioaddict",
+  "stage": "beta",
+  "name": "AudioAddict",
+  "description": "AudioAddict Network Radio provider for Music Assistant. Supports DI.FM, RadioTunes, RockRadio, JazzRadio, ClassicalRadio, and ZenRadio streaming services.",
```
OzGav:
```suggestion
  "description": "Stream curated internet radio from DI.FM, RadioTunes, RockRadio, JazzRadio, ClassicalRadio, and ZenRadio.",
```

OzGav:
Resolve the conversations with a comment when they are done. Does this provider not provide any artist or track information as part of their API or in the stream?

music_assistant/providers/audioaddict/manifest.json
benklop: renamed.

music_assistant/providers/audioaddict/__init__.py
benklop: corrected.

music_assistant/providers/audioaddict/__init__.py
benklop: corrected.

music_assistant/providers/audioaddict/__init__.py
benklop: corrected.

music_assistant/providers/audioaddict/__init__.py
benklop: done.

music_assistant/providers/audioaddict/__init__.py
benklop: corrected.

benklop:
> Resolve the conversations with a comment when they are done. Does this provider not provide any artist or track information as part of their API or in the stream?

I believe it's there in the stream - I'm still working on how to retrieve that. Album art appears to be present in the web interface, but I'm not sure if that's accurate for the ICY stream and I'm not sure if it's in the API.

benklop:
> > Resolve the conversations with a comment when they are done. Does this provider not provide any artist or track information as part of their API or in the stream?
> 
> I believe it's there in the stream - I'm still working on how to retrieve that. Album art appears to be present in the web interface, but I'm not sure if that's accurate for the ICY stream and I'm not sure if it's in the API.

I'm further confused because in my browser, it appears the player _has_ the metadata shown. The album art there is the generic channel art and not anything specific for what is playing, but I'm not sure the ICY metadata provides that.
image

Is there in fact more that I need to do to provide the metadata correctly?

OzGav:
No if its showing then the core is sorting that for you. Unless you can see a problem then its probably fine.

benklop:
I just located much more API documentation - it may be possible to implement a few other interesting features . https://web.archive.org/web/20140426192326/http://tobiass.eu/api-doc.html#trackinfo

music_assistant/providers/digitally_incorporated/__init__.py
```diff
@@ -0,0 +1,625 @@
...
+
+        if not valid_networks:
+            self.logger.warning(
+                "%s: No valid networks enabled, defaulting to all networks", self.domain
+            )
+            return list(NETWORKS.keys())
+
+        return valid_networks
+
+    async def _api_request(
```
marcelveldt, line 456: If you want to limit the amount of (concurrent) calls to the api and/or the service has some fair use / rate limiting policy, consider implementing a throttler. We do have a default one available that you can simply use with a decorator.

benklop: Thanks, I will do that. I'm not actually aware of any policy or limits, but no reason to ddos them

marcelveldt (approved):
Looks good, nice work @benklop !

benklop:
@marcelveldt sorry about this, I just opened https://github.com/music-assistant/server/pull/2472 with a bugfix and a handful of enhancements I had been working on but forgot to push.

OzGav:
@benklop I also need some text for the docs as well We have the following headings. You don’t need text for all of them. Please see the other providers for ideas of what to add. Just the text is fine I will do the actual PR for the docs as a number of things need to be done.

FEATURES
CONFIGURATION
KNOWN ISSUES/ NOTES
NOT YET SUPPORTED

benklop:
# FEATURES:

- retrieves and plays hundreds of curated radio streams from di.fm,
jazzradio.com, rockradio.com, zenradio.com, classicalradio.com, and
radiotunes.com.
- channel artwork
- genre information
- detailed channel descriptions


# CONFIGURATION:

requires a premium subscription and a "listen key", retrievable from the
settings panel on any one of the sites in the network. The key from all
sites is identical and subscribing to one gets you access to all.


# KNOWN ISSUES/NOTES:

only one steam at a time from the network is officially supported. In
practice it seems like one search from each site in the network may be
allowed, but it is unclear and subject to change. If you try to pay a
second stream and it stops playing after 30 seconds, that indicates another
conflicting stream of already playing from the network. Currently playing
streaming clients can be listed in settings page of any member in the
network.

# NOT YET SUPPORTED:

more metadata is available in the API but not yet exposed, including:
- per track album art
- up votes and down votes per track
- exact length and elapsed time of the current track
- start time off the current track
- all information for up to 9 previous tracks
- list of other channels similar to each channel in the network
- the name of the curator of each channel
- significant artists featured on each channel
- broad categories in addition to narrow genres
- upcoming live events on some channels

OzGav:
I assume no stations are added by default and you go to browse and then add to the library from there?

benklop:
There is a selector in the configuration that is used to determine whether
they are added to the radio list or not. By default they all are - perhaps
that should be changed so the default only makes them browsable.

OzGav:
Yeah I guess we have the import options now. I might just highlight that leaving the default will result in hundreds of stations automatically added

OzGav:
You can review the page here https://music-assistant.github.io/music-providers/digitally-incorporated/

OzGav:
@benklop  Hi Ben. We are looking to standardise the behaviour of all the radio providers. I have a question about DI.FM. You have added `ProviderFeature.LIBRARY_RADIOS` but this is actually for when a provider has its own library and you want to sync that to MA. For example, TUNEIN you can "favourite" stations on their website and this then appears on a page called LIBRARY. So my question is, does DI.FM have a library concept? If not, I will remove that provider feature and the `get_library_radios` method.

benklop:
> @benklop  Hi Ben. We are looking to standardise the behaviour of all the radio providers. I have a question about DI.FM. You have added `ProviderFeature.LIBRARY_RADIOS` but this is actually for when a provider has its own library and you want to sync that to MA. For example, TUNEIN you can "favourite" stations on their website and this then appears on a page called LIBRARY. So my question is, does DI.FM have a library concept? If not, I will remove that provider feature and the `get_library_radios` method.  

There are favorites, ~~but I'm not sure if I can get to those in the API or not. I'll have to take a look. For now, it might make sense to remove that feature.~~ , they can be accessed using ListFavorites(ctx) I think. I'll see about a PR for that.

OzGav:
Thanks Ben. For your info if you implement that then those “favourite” stations should be automatically added to the library via the get_library_radios method. All other stations should be found via search or browse.

benklop:
@OzGav https://github.com/music-assistant/server/pull/3458 this seems to work for me. favorites have to be set on the web - theres no way for us to edit them without requesting more credentials from the user. maybe a reasonable thing to do - but more work than i want to do right now

PR #2421, opened 2025-09-23
Add Phishin provider

OzGav:
Satisifies https://github.com/orgs/music-assistant/discussions/3258

From testing the only thing that isn't working but I would like it to is "Appears On" sections are empty: Track detail pages don't show which albums/shows contain each song

### Phish.in Provider Features & Limitations
**Core Features**
**Music Library Access**

Stream all Phish concerts from 1983-present with available audio
Browse 1,800+ shows organized as albums in Music Assistant
Access 50,000+ individual track performances
Artist view shows Phish with external IDs (MusicBrainz, Discogs, TADB)
Some early shows have incomplete or missing audio
1,000 item limits on library views for UI responsiveness

**Search Functionality**

Multi-venue search: Find shows by venue name (e.g., "Madison" returns MSG shows)
Track search: Find specific song performances across different shows
Playlist search: Access user-created playlists from Phish.in community
Title-only matching (no fuzzy search or advanced queries)

**Browse Organization**

By Year/Era: Browse shows chronologically by year or multi-year periods
By Venue: All venues with available audio, sorted by show count
By Tag: Browse by musical tags (jams, guest appearances, etc.)
This Day in History: Shows that happened on current date in past years
Top Content: Most liked shows and tracks from community voting
User Playlists: Community-curated track collections
Random Shows: Discover random performances


music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1055 @@
...
+        """Return the features supported by this provider."""
+        return {
+            ProviderFeature.BROWSE,
+            ProviderFeature.SEARCH,
+            ProviderFeature.LIBRARY_ARTISTS,
+            ProviderFeature.LIBRARY_ALBUMS,
+            ProviderFeature.LIBRARY_TRACKS,
+            ProviderFeature.LIBRARY_PLAYLISTS,
+            ProviderFeature.ARTIST_ALBUMS,
+            ProviderFeature.ARTIST_TOPTRACKS,
```
marcelveldt: please move this to a constant in the __init__ file and drop the property (that is only for runtime overrides)

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1055 @@
...
+                artists=artists[:limit] if MediaType.ARTIST in media_types else [],
+                albums=albums[:limit] if MediaType.ALBUM in media_types else [],
+                tracks=tracks[:limit] if MediaType.TRACK in media_types else [],
+                playlists=playlists[:limit] if MediaType.PLAYLIST in media_types else [],
+            )
+        except (MediaNotFoundError, ProviderUnavailableError):
+            raise
+        except Exception as err:
+            self.logger.error("Search failed for query '%s': %s", search_query, err)
+            raise ProviderUnavailableError(f"Search error: {err}") from err
```
marcelveldt, line 144:
```suggestion
            raise MediaNotFoundError(f"Search error: {err}") from err
```

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1055 @@
...
+            raise
+        except Exception as err:
+            self.logger.error("Failed to get library albums: %s", err)
+            raise ProviderUnavailableError(f"Library albums error: {err}") from err
+
+    async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
+        """Retrieve library tracks from the provider."""
+        page = 1
+        per_page = 50
+        max_pages = 20  # 1000 tracks max for UI performance
```
marcelveldt: why ?

OzGav: Why 1000 limit or why 50 at a time? I was trying to balance UI speed vs practicality of scrolling through so many tracks? The actual API limit is 1000 per page?

music_assistant/providers/phishin/provider.py
OzGav: done

music_assistant/providers/phishin/provider.py
marcelveldt: the core logic should take care of paging once its in the library so you should have to worry that much about it

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1061 @@
...
+
+                page += 1
+
+        except (MediaNotFoundError, ProviderUnavailableError):
+            raise
+        except Exception as err:
+            self.logger.error("Failed to get library albums: %s", err)
+            raise ProviderUnavailableError(f"Library albums error: {err}") from err
+
+    async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
```
marcelveldt: So, what I'm wondering is if we should return all these items as library items ?
The definition of that is that the user added them to the library/favorites of the streaming provider but that is not the case here. You are just returning the whole catalog as library which I think is maybe not the right way to do this ?

Should we perhaps treat this the same as Radio Browser where you can simply Browse all available items and then manually add something you like to your MA library ?

music_assistant/providers/phishin/provider.py
OzGav: I have increased the per page amount for the API calls.

music_assistant/providers/phishin/provider.py
OzGav: Yes that sounds like a better plan. This has been changed.

music_assistant/providers/phishin/helpers.py
```diff
@@ -0,0 +1,463 @@
...
+    PHISH_MUSICBRAINZ_ID,
+    PHISH_TADB_ID,
+    REQUEST_TIMEOUT,
+)
+
+if TYPE_CHECKING:
+    from music_assistant.models.music_provider import MusicProvider
+
+
+@use_cache(expiration=3600)  # 1 hour
```
marcelveldt: so, instead of defining the cache for all requests, maybe consider setting it on each method seperate. That way you can play with the expiration. For example the details of one artist or album can have an expiration of something like 60 days, they wont likely change, while the expiration of a playlist-tracks listingneeds to be short lived as it can change dynamically

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1015 @@
...
+    def __init__(
+        self,
+        mass: MusicAssistant,
+        manifest: ProviderManifest,
+        config: ProviderConfig,
+        supported_features: set[ProviderFeature],
+    ) -> None:
+        """Initialize the provider."""
+        super().__init__(mass, manifest, config)
+        self._supported_features = supported_features
```
marcelveldt:
```suggestion
```
You are not overriding the method from the base class so you can drop this

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1015 @@
...
+
+    @property
+    def supported_features(self) -> set[ProviderFeature]:
+        """Return the features supported by this provider."""
+        return self._supported_features
```
marcelveldt:
```suggestion
```
You are not overriding the method from the base class so you can drop this

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1015 @@
...
+
+                            if current_page >= total_pages or len(venue_shows) >= limit:
+                                break
+
+                            page += 1
+
+                    except ProviderUnavailableError:
+                        self.logger.warning(
+                            "Failed to fetch shows for venue %s: API unavailable", venue_slug
+                        )
```
marcelveldt: why not simply let the exception bubble up instead of swallowing it here silently ?
if you let the exception raise, the api handler will catch it and forward and error to the user that the action failed

marcelveldt: hmm I see you kind of do this but then wrap everything in an outer exception, i think that can be deduplicated by just only have the outer exception handler which then wraps it into ProviderUnavailableError or MediaNotFoundError with a sensible error message.

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1015 @@
...
+
+            return SearchResults(
+                artists=artists[:limit] if MediaType.ARTIST in media_types else [],
+                albums=albums[:limit] if MediaType.ALBUM in media_types else [],
+                tracks=tracks[:limit] if MediaType.TRACK in media_types else [],
+                playlists=playlists[:limit] if MediaType.PLAYLIST in media_types else [],
+            )
+        except (MediaNotFoundError, ProviderUnavailableError):
+            raise
+        except Exception as err:
```
marcelveldt:
```suggestion
        except  (MediaNotFoundError, ProviderUnavailableError, aiohttp.ClientError) as err:
```
gave the aiottp.ClientError as example - just try to catch specific exceptions that can actually occur

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,1015 @@
...
+
+            return albums
+
+        except (MediaNotFoundError, ProviderUnavailableError):
+            raise
+        except Exception as err:
+            self.logger.error("Failed to get shows for venue %s: %s", venue_slug, err)
+            raise ProviderUnavailableError(f"Venue shows error: {err}") from err
+
+    @use_cache(expiration=86400)  # 24 hours - historical tag data is stable
```
marcelveldt: why not do 30 days if its this stable data ?

OzGav: All cache durations reviewed and proper justifcation added to each.

music_assistant/providers/phishin/provider.py
OzGav: Fixed

music_assistant/providers/phishin/provider.py
OzGav: removed

music_assistant/providers/phishin/provider.py
OzGav: removed. I wasnt fully across the changes you made to the supported features

music_assistant/providers/phishin/helpers.py
OzGav: This decorator removed. All methods now have individual cache durations with a justification comment

OzGav:
Just confirming this is only issue I know of now. Ideas welcome!

> From testing the only thing that isn't working but I would like it to is "Appears On" sections are empty: Track detail pages don't show which albums/shows contain a track (even after I click through the album's track list to get to the tracks details)

marcelveldt:
It means the search is not returning any tracks with the album set

OzGav:
Still need help. I have cleaned up the track titles so they will match. I have passed the show data so the album info should be added to the track in track_to_ma_track. I have added a get_similar_tracks method but that is never called ? get_track_albums is never called. So I am missing some architecural understanding about what should happen to populate that view. I do see Spotify matches but I cant see what is different between the providers...

marcelveldt:
get_similar_tracks is only called for radio mode (did you add that feature ?)
if get_track_albums is never called it means that you dont have the album set correctly on the tracks

OzGav:
Thanks for the pointers. This is now ready to go.

music_assistant/providers/phishin/provider.py
```diff
@@ -0,0 +1,963 @@
...
+                return []
+
+            playlist_data = await api_request(
+                self, ENDPOINTS["playlist_by_slug"].format(slug=playlist_slug)
+            )
+
+            # Simple pagination - fetch all tracks and slice
+            page_size = 50
+            start_idx = page * page_size
+            end_idx = start_idx + page_size
```
marcelveldt: This is not needed. If you dont support paging, just return all results at once (if page == 0) but make sure that you return no results on page > 0

OzGav: Fixed

marcelveldt (approved):
Nice work @OzGav !

PR #2423, opened 2025-09-24
Fix pulling ARD Audiothek subscriptions and ordering of podcasts with many episodes

jfeil:
This fix removes the unsupported subscription filtering.
Further it fixes the missing offset to the episode idx and corrects the ordering of displayed episodes.


marcelveldt (approved):
Thanks, @jfeil !

PR #2425, opened 2025-09-24
gpodder/ itunes/ rssfeed: fix bad header for certain podcasts

fmunkes:
A while ago we added a ```{"User-Agent": "Mozilla/5.0"}``` when requesting a podcast feed, as reports showed failure when retrieving otherwise, see music-assistant/support#3596 . A user on discord now provided me with podcast feeds, where the opposite is true, the header being added prevents retrieval.

I combined the individual functions of gpodder/ itunes_podcast/ podcastfeed into a single helper function, which tries retrieval including header first, and tries again without header.

Fixes the issue for me.


music_assistant/helpers/podcast_parsers.py
```diff
@@ -16,6 +19,28 @@
...
+    response: aiohttp.ClientResponse | None = None
+    # without user agent, some feeds can not be retrieved
+    # https://github.com/music-assistant/support/issues/3596
+    # but, reports on discord show, that also the opposite may be true
+    for headers in [{"User-Agent": "Mozilla/5.0"}, {}]:
+        response = await session.get(feed_url, headers=headers)
+        if response.status == 200:
+            break
+    if response is None or response.status != 200:
+        raise RuntimeError
```
marcelveldt:
```suggestion
        response = await session.get(feed_url, headers=headers, raise_for_status=True)
```

So you can simply add raise_for_status as argument to do the status check for you

marcelveldt (approved):
Thanks, @fmunkes !

PR #2427, opened 2025-09-26
Fix lint errors in other PRs

OzGav:


fmunkes:
That is the correct spelling of the config option, see [here](https://github.com/release-drafter/release-drafter). So, you can add an exception to the pyproject.toml for code spell.

OzGav:
Ok closing this. I am away so will need someone else to do the exception. Thanks Fabian.

PR #2429, opened 2025-09-26
More grouping fixes after the player refactor

maximmaxim345:
Another round of grouping related fixes after the player refactor, specifically:
- replace player model hook `on_registered` with `on_config_updated` (refactors squeezelite to use that instead of overriding the `@final` `set_config`).
- Adds support for  `static_group_members`: https://github.com/music-assistant/models/pull/118
- Static group members are now checked in the `PlayerController` instead of each `Player`
- Unregistered players now get automatically ungrouped
- Players that become disabled or unavailable now also get automatically ungrouped
- Update `SyncGroupPlayer` and `UniversalGroupPlayer`:
  - `player.set_members` will allow static group members to be removed
  - groups only start with available players when powered on.

Depends on: https://github.com/music-assistant/models/pull/118
Required by: https://github.com/music-assistant/frontend/pull/1153


marcelveldt (approved):
LGTM, Thanks!

PR #2432, opened 2025-09-27
Adding alexa language config value

ChrisBkt:
- Adding alexa language config value
- Adding german language model

Needed for non-english countries, because alexa does not recognize command to play audio.
With language config integration, languages can be added to Language array, so users can set their language by their own.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,11 +112,6 @@ jobs:
           echo "patch=${patch}" >> $GITHUB_OUTPUT
           echo "minor=${patch%.*}" >> $GITHUB_OUTPUT
           echo "major=${patch%.*.*}" >> $GITHUB_OUTPUT
-      - name: Generate lower case repository owner
-        id: lower_owner
-        uses: ASzc/change-string-case-action@v6
-        with:
-          string: ${{ github.repository_owner }}
       - name: Build and Push release
         uses: docker/build-push-action@v6.18.0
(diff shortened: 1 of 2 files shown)


OzGav:
@alams154

alams154:
Looks good, thanks!

alams154 (approved):

OzGav:
@ChrisBkt please review the lint problems

ChrisBkt:
@OzGav Fixed linting i think.
I'm not sure where the remaining linting error "unsupported imports" is from.
I did not added any import.

Is there maybe a ConfigEntry type to have a dropdown menu with suggestions, so you don't need to type in en-US or de-DE for example, because currently if you type in a language shortcut that is not implemented, it will auto select english.

OzGav:
Have a look here https://github.com/music-assistant/models/blob/57d968a2e3fc2406d8d5b469d3ad43ca3aa0afb4/music_assistant_models/config_entries.py#L86

.github/workflows/release.yml
marcelveldt, line 1: please put this in a dedicated PR (if needed) - it at least doenst belong here in this PR for alexa provider

ChrisBkt: Sorry i'm not sure why this landed in this branch.
Should stay in my own testing branch.
Will remove it soon when i add the language suggestions.

marcelveldt (approved):
Thanks, @ChrisBkt !

PR #2435, opened 2025-09-28
Podcastfeed Handle images no longer at URL

OzGav:
https://github.com/orgs/music-assistant/discussions/4247#discussioncomment-14452634

Per the discussion above this RSS feed provider can allow very old podcasts that have broken image links. This will allow fallback to be used and shown in the episode list of the podcast.

Having them show on the Now Playing view is yet to be resolved

### DIFF
--- a/music_assistant/providers/podcastfeed/__init__.py
+++ b/music_assistant/providers/podcastfeed/__init__.py
@@ -235,3 +235,29 @@ class PodcastMusicprovider(MusicProvider):
             data=self.parsed_podcast,
             expiration=60 * 60 * 24,  # 1 day
         )
+
+    async def resolve_image(self, path: str) -> str | bytes:
+        """Resolve image for RSS provider with fallback to podcast cover."""
+        if not path.startswith("http"):
+            return path
+
+        try:
+            async with self.mass.http_session.get(
+                path, raise_for_status=True, allow_redirects=False
+            ) as response:
+                # Check for redirect status codes (3xx)
+                if 300 <= response.status < 400:
+                    raise ClientError(f"Redirect status: {response.status}")
+
+                return await response.read()
+
+        except (ClientError, Exception):
+            # Try podcast cover fallback
+            podcast_cover = self.parsed_podcast.get("cover_url")
+            if podcast_cover and isinstance(podcast_cover, str) and podcast_cover != path:
+                async with self.mass.http_session.get(
+                    podcast_cover, raise_for_status=True
+                ) as response:
+                    return await response.read()
+
+            raise MediaNotFoundError(f"Episode image not found: {path}")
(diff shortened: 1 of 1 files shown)


music_assistant/providers/podcastfeed/__init__.py
```diff
@@ -235,3 +235,29 @@ async def _cache_set_podcast(self) -> None:
...
         )
+
+    async def resolve_image(self, path: str) -> str | bytes:
+        """Resolve image for RSS provider with fallback to podcast cover."""
+        if not path.startswith("http"):
+            return path
+
+        try:
+            async with self.mass.http_session.get(
+                path, raise_for_status=True, allow_redirects=False
```
marcelveldt: why disallow redirects ?

music_assistant/providers/podcastfeed/__init__.py
```diff
@@ -235,3 +235,29 @@ async def _cache_set_podcast(self) -> None:
...
+            ) as response:
+                # Check for redirect status codes (3xx)
+                if 300 <= response.status < 400:
+                    raise ClientError(f"Redirect status: {response.status}")
```
marcelveldt: you can simply do response.raise_for_status()

music_assistant/providers/podcastfeed/__init__.py
```diff
@@ -235,3 +235,29 @@ async def _cache_set_podcast(self) -> None:
             data=self.parsed_podcast,
             expiration=60 * 60 * 24,  # 1 day
         )
+
+    async def resolve_image(self, path: str) -> str | bytes:
```
marcelveldt, line 260: If you implement this, you will also have to mark all images as not externally accessible

music_assistant/providers/podcastfeed/__init__.py
OzGav: Because in my testing the situation was a 301 for port 80 then a 301 for port 443 then a 404 which needed to be caught but after thinking about your comment there might be legitmate redirects so this has been reworked.

music_assistant/providers/podcastfeed/__init__.py
OzGav: This has already been changed due to your previous comment about allowing redirects - now validates content-type instead

music_assistant/providers/podcastfeed/__init__.py
OzGav: OK have a look at the revised code as I wasn't initially sure I could that for this provider only but indeed I can. This solves the whole problem the reporter has and if you agree that it is OK for this fix for this provider (where old RSS feeds may have broken image URLs) then this will fix the issue completely.

marcelveldt (approved):

PR #2437, opened 2025-09-28
mypy fixes for webserver.py

OzGav:


music_assistant/controllers/webserver.py
```diff
@@ -269,16 +272,16 @@ def __init__(self, webserver: WebserverController, request: web.Request) -> None
...
-        self._to_write: asyncio.Queue = asyncio.Queue(maxsize=MAX_PENDING_MSG)
-        self._handle_task: asyncio.Task | None = None
-        self._writer_task: asyncio.Task | None = None
+        self._to_write: asyncio.Queue[str | None] = asyncio.Queue(maxsize=MAX_PENDING_MSG)
+        self._handle_task: asyncio.Task[Any] | None = None
+        self._writer_task: asyncio.Task[None] | None = None
         self._logger = webserver.logger
         # try to dynamically detect the base_url of a client if proxied or behind Ingress
-        self.base_url: str | None = None
+        self.client_base_url: str | None = None
```
marcelveldt: why did the name of this variable change ?

OzGav: I believe it was conflicting with the WebserverController's base_url property?

OzGav: The original error was `Property "base_url" defined in "WebserverController" is read-only Mypymisc`

music_assistant/controllers/webserver.py
```diff
@@ -53,8 +52,9 @@ class WebserverController(CoreController):
     """Core Controller that manages the builtin webserver that hosts the api and frontend."""
 
     domain: str = "webserver"
+    _server: Webserver
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
maximmaxim345: Why not just change it to `__init__(self, mass: MusicAssistant)` and pass that to `super().__init__()`?

music_assistant/controllers/webserver.py
```diff
@@ -164,9 +164,13 @@ async def setup(self, config: CoreConfig) -> None:
         else:
             ingress_tcp_site_params = None
         base_url = str(config.get_value(CONF_BASE_URL))
-        self.publish_port = int(config.get_value(CONF_BIND_PORT))
+        port_value = config.get_value(CONF_BIND_PORT)
+        self.publish_port = (
+            int(port_value) if isinstance(port_value, (int, float, str)) else DEFAULT_SERVER_PORT
+        )
```
maximmaxim345: I think `port_value` should always be a int why not assert it then, so we know if something breaks?

music_assistant/controllers/webserver.py
maximmaxim345: Strange, mypy seams to work for me when keeping the old `base_url` name

music_assistant/controllers/webserver.py
OzGav: Because Im a noob at Python typing! Happy for you to make any adjustments as you see fit!

music_assistant/controllers/webserver.py
OzGav: Maybe there was a mypy difference with my dev environment but I am running the setup script regularly. Again happy to go with what you guys determine is optimal.

music_assistant/controllers/webserver.py
OzGav: Again good question. I am not fully across the best technqiue to use in which circumstance.

music_assistant/controllers/webserver.py
maximmaxim345: Looks like github actions also doesn't complain about `base_url`. Interesting

music_assistant/controllers/webserver.py
maximmaxim345: No worries! Love having your help!

maximmaxim345 (approved):
LGTM, Thanks @OzGav !

PR #2438, opened 2025-09-29
mypy fixes for Deezer

OzGav:
Also need to remove the Deezer line from the excludes in pyproject.toml


marcelveldt:
> Also need to remove the Deezer line from the excludes in pyproject.toml

So, why did you not yet do that ?

OzGav:
I wasn’t sure if it was ok to change that but I will do that soon

marcelveldt (approved):
Thanks, @OzGav !

PR #2456, opened 2025-09-30
Remove enhanced title for RadioParadise

OzGav:


music_assistant/providers/radioparadise/parsers.py
```diff
@@ -88,6 +88,7 @@ def build_stream_metadata(current_song: dict[str, Any], metadata: dict[str, Any]
     next_song = metadata.get("next")
     block_data = metadata.get("block_data")
     enhanced_title = enhance_title_with_upcoming(title, current_song, next_song, block_data)
+    enhanced_title = title  # TODO remove after frontend update
 
     return StreamMetadata(
         title=enhanced_title,
```
marcelveldt:
```suggestion
# TODO: Find a way to forward the next_song data to the frontend in the stream metadata
# enhanced_title = enhance_title_with_upcoming(title, current_song, next_song, block_data)
# enhanced_title = title  # TODO remove after frontend update

return StreamMetadata(
title=title,
```

marcelveldt: Let's leave the title as the title and think later of better ways to do this

marcelveldt (dismissed):

marcelveldt (approved):

PR #2465, opened 2025-10-02
More SyncGroupPlayer fixes

maximmaxim345:
Specifically issues I found when using the Sync groups created with the Resonate Provider.
- Set `active_source` to the group instead of the internally selected sync leader. This fixes a bug where some players disappear from the sync list, but only when playing music.


marcelveldt:
fixed in the follow-up-fixes branch, lets continue there

PR #2472, opened 2025-10-03
Various improvements to the DI Provider

benklop:
This pull request introduces significant improvements to the `digitally_incorporated` provider.

Get metadata from a more verbose endpoint
Set genre information
Set rich descriptions
Get channel image URLs correctly


music_assistant/providers/digitally_incorporated/__init__.py
```diff
@@ -607,19 +719,38 @@ def _channel_to_radio(self, channel_data: dict[str, Any], network_key: str) -> R
...
         )
 
+    def _is_disabled_channel(self, channel_data: dict[str, Any]) -> bool:
+        """Check if a channel is disabled based on its name pattern."""
+        name = channel_data.get("name")
+        if not name or not isinstance(name, str) or len(name) < 2:
+            return False
+
+        # Disabled channels have names starting with 'X' followed by a (capitalized) channel name.
+        return bool(name[0] == "X" and name[1].isupper())
```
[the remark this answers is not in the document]
benklop: this simply is the pattern used by the provider. this already handles edge cases by not filtering out channels that start with X, as they would not have the second letter as a capital letter.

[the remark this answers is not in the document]
benklop: Note that the previous /listen/channels API endpoint didnt include the disabled channels but also included very little useful metadata. Having to filter the disabled channels is the downside of getting more metadata. If preferred, we could possibly get the other API endpoint as well and use the inteersection of the two to filter the channels instead. I don't know if this is completely equivalent, nor if it would functionally be less fragile.

music_assistant/providers/digitally_incorporated/__init__.py
```diff
@@ -247,7 +247,7 @@ async def search(
                 for channel_data in channels:
                     channel_name = str(channel_data.get("name", "")).lower()
                     if search_query_lower in channel_name:
-                        radio = self._channel_to_radio(channel_data, network_key)
+                        radio = await self._channel_to_radio(channel_data, network_key)
```
marcelveldt: So, one thing to consider here is tha added overhead of requiring a coroutine for all these parsers.
Parsers should normally be lightweight, IO-free methods that simple take the raw data and spit out the MA model equivalent. Within such parser make sure to not do (or have the need to) make any API calls.

If your api returns incomplete data in the main listings, that is perfectly fine. If you then return the full item details within the "get_artist", "get_track" etc. methods then it can be retrieved later when its needed.

If the data is really minimal, you can also consider returning it as ItemMapping objects. Then the internal machinery will know that it needs to fetch the full object before e.g. playing or storing it.

music_assistant/providers/digitally_incorporated/__init__.py
```diff
@@ -567,17 +661,33 @@ def _channel_to_radio(self, channel_data: dict[str, Any], network_key: str) -> R
...
+        # Create metadata with optional image and genres
         metadata = MediaItemMetadata(
-            description=f"{network_info['description']} - {channel_name}",
+            description=self._get_description(channel_data),
             explicit=False,
         )
 
+        # Add genre information
+        try:
+            genres = await self._get_channel_genres(channel_data, network_key)
```
marcelveldt: so, is this avoidable to do for EVERY station ?

benklop: We should be able to grab it once for each site in the network, and use the returned information to determine the genre info for all stations in that site. It ships be reasonable for us to do this during provider scans and just cache the results.

music_assistant/providers/digitally_incorporated/__init__.py
benklop: Sorry, I'm pretty new to python, particularly things like async methods. I'll see what I can figure out, but I'm very open to any suggestions about more efficient ways to restructure the code.

benklop:
I'll come up with a better solution

music_assistant/providers/digitally_incorporated/__init__.py
marcelveldt: Yeah, that will be better. Otherwise you are going to a extra call for every station in a listing.

marcelveldt:
Marked as draft while you work on the feedback. Please mark it as ready for review once ready. Thanks!

marcelveldt (approved):
LGTM!

PR #2477, opened 2025-10-03
Fix: Smart fades beat alignment

MarvinSchenkel:
Some Smart Fades enhancements:
- We now align the start of the crossfade exactly with a downbeat of the outgoing track to ensure proper beat alignment
- We can extrapolate downbeats in case of an outro without detected downbeats
- Use rubberband for a higher quality time stretch (atempo is lossy)
- Apply a uniform tempo as gradual tempo ramp causes audible artifacts at the segment boundaries. Downside is that we have to reduce the stretch % a little as the difference is way too noticeable >5%.


marcelveldt (approved):
LGTM!

PR #2478, opened 2025-10-03
ARD Audiothek: Introduce filter of not published episodes to reduce loading times

jfeil:
This change adds a filter to the GraphQL request to filter not yet or no longer published episodes. Some live channels contain more than 1000 depublished episodes and this reduces the loading times by a significant amount


marcelveldt (approved):
LGTM, Thanks @jfeil !

PR #2484, opened 2025-10-05
Add WebDAV provider

OzGav:
This provider extends the Local File System Provider to support music libraries hosted on WebDAV servers. It overrides the storage access methods (exists, resolve, _scandir, _read_file) to use WebDAV PROPFIND and HTTP instead of local filesystem calls. All parsing, syncing, and streaming logic is inherited from the parent provider.

It supports tracks, albums, artists, playlists, single-file and multi-file audiobooks, and podcasts with full metadata parsing via ffprobe. Authentication is handled via Basic Auth with URL-encoded credentials embedded in URLs for ffprobe/ffmpeg access.

The parent filesystem provider has been refactored to use overridable methods (_scandir, _read_file, _get_chapter_path) for storage operations, allowing WebDAV to substitute HTTP access without duplicating any logic.


music_assistant/providers/webdav/helpers.py
```diff
@@ -0,0 +1,226 @@
...
+        SetupFailedError: Server configuration issues
+    """
+    auth = None
+    if username:
+        auth = aiohttp.BasicAuth(username, password or "")
+
+    connector = aiohttp.TCPConnector(ssl=verify_ssl)
+
+    try:
+        async with aiohttp.ClientSession(
```
marcelveldt: don't create new aiohttp.ClientSessions, always use the existing ones from mass

music_assistant/providers/webdav/parsers.py
marcelveldt, line 1: All this feels completely redundant as you could simply subclass the existing Filesystem provider.

marcelveldt (changes requested):
- multi part streaming has changed in core
- 80% of the code is redundant with filesystem provider

music_assistant/providers/webdav/parsers.py
OzGav: OK I am onto it. I think I might need the models bumped so I can use MultiPartPath ?

music_assistant/providers/webdav/helpers.py
OzGav: fixed

music_assistant/providers/webdav/parsers.py
OzGav: So far I have addressed redundancy by:

Deleting parsers.py
Only overriding file access methods - exists(), resolve(), _scandir() (WebDAV PROPFIND vs local filesystem)
Minimal custom builders for audiobooks/podcasts - parent's _parse_audiobook() and _parse_podcast_episode() call sorted_scandir() which all expect local filesystem paths, not WebDAV URLs - would require major refactor of parent to support both
WebDAV-specific code limited to: authenticated URL building and network requests with BasicAuth

Core parsing logic (tags → media objects) all comes from parent now. Custom code only where network access fundamentally differs from filesystem access.

OzGav: I see now that the models have been bumped I am just fighting with vscode to actually recognise that even though I have run the setup script and  models 1.1.61 is installed

music_assistant/providers/webdav/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+            label="Content type",
+            options=[
+                ConfigValueOption("Music", "music"),
+                ConfigValueOption("Audiobooks", "audiobooks"),
+                ConfigValueOption("Podcasts", "podcasts"),
+            ],
+            default_value="music",
+            description="The type of content stored on this WebDAV server",
+            hidden=instance_id is not None,
+        ),
```
marcelveldt, line 64: Cant you import this from filesystem prov constants ?

music_assistant/providers/webdav/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+            type=ConfigEntryType.STRING,
+            label="Action when album artist tag is missing",
+            options=[
+                ConfigValueOption("Use track artist(s)", "track_artist"),
+                ConfigValueOption("Use folder name", "folder_name"),
+                ConfigValueOption("Use 'Various Artists'", "various_artists"),
+            ],
+            default_value="various_artists",
+            description="What to do when a track is missing the album artist tag",
+        ),
```
marcelveldt: Cant you import this from filesystem prov constants ?

music_assistant/providers/webdav/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        ConfigEntry(
+            key="ignore_album_playlists",
+            type=ConfigEntryType.BOOLEAN,
+            label="Ignore playlists in album folders",
+            default_value=True,
+            description="Ignore playlist files found in album subdirectories",
+        ),
```
marcelveldt: Cant you import this from filesystem prov constants ?

music_assistant/providers/webdav/constants.py
```diff
@@ -0,0 +1,54 @@
...
+
+IMAGE_EXTENSIONS: Final[tuple[str, ...]] = ("jpg", "jpeg", "png", "gif", "bmp", "webp", "tiff")
+
+SUPPORTED_EXTENSIONS: Final[tuple[str, ...]] = (
+    *TRACK_EXTENSIONS,
+    *PLAYLIST_EXTENSIONS,
+    *AUDIOBOOK_EXTENSIONS,
+    *IMAGE_EXTENSIONS,
+)
+
```
marcelveldt: why is this all duplicated from filesystem provider ?!

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+
+    @property
+    def instance_name_postfix(self) -> str | None:
+        """Return a (default) instance name postfix for this provider instance."""
+        try:
+            parsed = urlparse(self.base_url)
+            if parsed.path and parsed.path != "/":
+                return PurePosixPath(parsed.path).name
+            return parsed.netloc
+        except Exception:
```
marcelveldt: no broad exceptions please

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                relative_path=file_path,
+                absolute_path=webdav_url,
+                is_dir=webdav_item.is_dir,
+                checksum=webdav_item.last_modified or "unknown",
+                file_size=webdav_item.size,
+            )
+
+        except MediaNotFoundError:
+            raise
+        except Exception as err:
```
marcelveldt: Can you narrow down the exception a bit maybe ?

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+            raise MediaNotFoundError(f"Failed to resolve WebDAV path {file_path}: {err}") from err
+
+    def get_absolute_path(self, file_path: str) -> str:
+        """Return WebDAV URL for given file path."""
+        return build_webdav_url(self.base_url, file_path)
+
+    async def browse(self, path: str) -> Sequence[MediaItemType | ItemMapping | BrowseFolder]:
```
marcelveldt: I dont see why this needs to be overwritten from the base implementation ?

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                    )
+        except Exception as err:
+            self.logger.error(f"Failed to browse WebDAV path {item_path}: {err}")
+
+        return items
+
+    async def get_track(self, prov_track_id: str) -> Track:
+        """Get full track details by id."""
+        file_item = await self.resolve(prov_track_id)
+        auth_url = self._build_authenticated_url(prov_track_id)
```
marcelveldt: Cant you build the authenticated url in the resolve method ?

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        tags = await async_parse_tags(auth_url, file_item.file_size)
+        return await self._parse_track(file_item, tags, full_album_metadata=True)
+
+    async def get_album(self, prov_album_id: str) -> Album:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                            provider=self.instance_id,
+                            name=item.filename,
+                        )
+                    )
+        except Exception as err:
+            self.logger.error(f"Failed to browse WebDAV path {item_path}: {err}")
+
+        return items
+
+    async def get_track(self, prov_track_id: str) -> Track:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                if track.album and isinstance(track.album, Album):
+                    album = track.album
+                    # Scan for folder images on every get_album call
+                    folder_images = await self._get_local_images(prov_album_id)
+                    if folder_images:
+                        album.metadata.images = folder_images
+                    return album
+        raise MediaNotFoundError(f"No tracks found in album path: {prov_album_id}")
+
+    async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        auth_url = self._build_authenticated_url(file_path)
+        tags = await async_parse_tags(auth_url, file_item.file_size)
+        audiobook = self._build_audiobook_from_tags(file_item, tags)  # Use custom builder
+
+        if file_item.relative_path != prov_audiobook_id:
+            audiobook.item_id = prov_audiobook_id
+
+        return audiobook
+
+    async def get_podcast_episodes(
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+
+        for episode_file in sorted(episode_files, key=lambda x: x.filename):
+            auth_url = self._build_authenticated_url(episode_file.relative_path)
+            tags = await async_parse_tags(auth_url, episode_file.file_size)
+            episode = self._build_podcast_episode_from_tags(
+                episode_file, tags
+            )  # Use custom builder
+            yield episode
+
+    async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        podcast_folder = str(PurePosixPath(prov_episode_id).parent)
+        folder_images = await self._get_local_images(podcast_folder)
+
+        assert isinstance(episode.podcast, Podcast)
+        if folder_images:
+            episode.podcast.metadata.images = folder_images
+
+        return episode
+
+    async def sync_library(self, media_type: MediaType, import_as_favorite: bool = False) -> None:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+            self.logger.error(f"WebDAV library sync failed due to connection error: {err}")
+            raise ProviderUnavailableError(f"WebDAV server connection failed: {err}") from err
+        except Exception as err:
+            self.logger.error(f"WebDAV library sync failed with unexpected error: {err}")
+            raise SetupFailedError(f"WebDAV library sync failed: {err}") from err
+        finally:
+            self.sync_running = False
+            self.logger.info(f"Completed library sync for WebDAV provider {self.name}")
+
+    async def _get_audiobook_chapters_from_cache(
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                    self.logger.warning(f"Error during scan: {result}")
+
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
+            raise
+        except aiohttp.ClientError as err:
+            self.logger.warning(f"WebDAV client error scanning path {path}: {err}")
+        except Exception as err:
+            self.logger.warning(f"Failed to scan WebDAV path {path}: {err}")
+
+    async def _process_webdav_item(
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                return True
+
+        except Exception as err:
+            self.logger.error(
+                f"Error processing WebDAV item {item.relative_path}: {err}",
+                exc_info=self.logger.isEnabledFor(10),
+            )
+        return False
+
+    async def _process_track(self, item: FileSystemItem, prev_checksum: str | None) -> None:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
+            raise
+        except aiohttp.ClientError as err:
+            self.logger.error(
+                f"Connection error processing WebDAV track {item.relative_path}: {err}"
+            )
+        except Exception as err:
+            self.logger.error(f"Failed to process WebDAV track {item.relative_path}: {err}")
+
+    async def _process_audiobook(self, item: FileSystemItem, prev_checksum: str | None) -> None:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
+            raise
+        except aiohttp.ClientError as err:
+            self.logger.error(
+                f"Connection error processing WebDAV audiobook {item.relative_path}: {err}"
+            )
+        except Exception as err:
+            self.logger.error(f"Failed to process WebDAV audiobook {item.relative_path}: {err}")
+
+    async def _process_multifile_audiobook(
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
+            raise
+        except aiohttp.ClientError as err:
+            self.logger.error(
+                f"Connection error processing multi-file audiobook {folder_path}: {err}"
+            )
+        except Exception as err:
+            self.logger.error(f"Failed to process multi-file audiobook {folder_path}: {err}")
+
+    async def _process_podcast(self, item: FileSystemItem, prev_checksum: str | None) -> None:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
+            raise
+        except aiohttp.ClientError as err:
+            self.logger.error(
+                f"Connection error processing WebDAV podcast {item.relative_path}: {err}"
+            )
+        except Exception as err:
+            self.logger.error(f"Failed to process WebDAV podcast {item.relative_path}: {err}")
+
+    async def _get_local_images(
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+                                remotely_accessible=False,
+                            )
+                        )
+
+            return UniqueList(images)
+        except Exception as err:
+            self.logger.debug(f"Failed to get images from {item_path}: {err}")
+            return UniqueList()
+
+    async def resolve_image(self, path: str) -> str | bytes:
```
marcelveldt, line 235: if you fix the resolve and exists base methods correctly with the webdav approach, you should have to override all the functions like this

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+        """Resolve image path to actual image data or URL."""
+        webdav_url = build_webdav_url(self.base_url, path)
+        session = self.mass.http_session if self.verify_ssl else self.mass.http_session_no_ssl
+
+        async with session.get(webdav_url, auth=self._auth) as resp:
+            if resp.status != 200:
+                raise MediaNotFoundError(f"Image not found: {path}")
+            return await resp.read()
+
+    async def _get_stream_details_for_track(self, item_id: str) -> StreamDetails:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+            media_type=MediaType.TRACK,
+            stream_type=StreamType.HTTP,  # CRITICAL: HTTP not LOCAL_FILE
+            duration=library_item.duration,
+            size=file_item.file_size,
+            path=auth_url,  # CRITICAL: authenticated URL
+            can_seek=True,
+            allow_seek=True,
+        )
+
+    async def _get_stream_details_for_podcast_episode(self, item_id: str) -> StreamDetails:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+            media_type=MediaType.PODCAST_EPISODE,
+            stream_type=StreamType.HTTP,
+            duration=int(tags.duration or 0),
+            size=file_item.file_size,
+            path=auth_url,
+            allow_seek=True,
+            can_seek=True,
+        )
+
+    async def _get_stream_details_for_audiobook(self, item_id: str) -> StreamDetails:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,1040 @@
...
+            media_type=MediaType.AUDIOBOOK,
+            stream_type=StreamType.HTTP,
+            duration=library_item.duration,
+            size=file_item.file_size,
+            path=auth_url,
+            can_seek=True,
+            allow_seek=True,
+        )
+
+    def _build_audiobook_from_tags(self, file_item: FileSystemItem, tags: AudioTags) -> Audiobook:
```
marcelveldt: this does not have to be overwritten from the base implementation imo

marcelveldt (changes requested):
Sorry, I cant accept this PR like this. 90% of the code is just a copy from the filesystem provider. If its written a bit smarter, then you can simply reuse most of the base implementation. 

Because streaming and tags reading need access to the file anyways, I would suggest returning the (athenticated) webdav URL in the "resolve" function.

music_assistant/providers/webdav/constants.py
OzGav: Just because I am a slow learner! Fixed!

music_assistant/providers/webdav/__init__.py
OzGav: fixed

music_assistant/providers/webdav/__init__.py
OzGav: fixed

music_assistant/providers/webdav/__init__.py
OzGav: fixed

music_assistant/providers/webdav/provider.py
OzGav: fixed

music_assistant/providers/webdav/provider.py
OzGav: For each of the following Ill explain why I think it has to be overwritten which principally is because of the need to use PROPFIND but Im sure you will let me know how this could be otherwise solved!

The base calls os.scandir(abs_path) which expects a filesystem path, not a URL. This override uses _scandir which makes WebDAV PROPFIND requests.

music_assistant/providers/webdav/provider.py
OzGav: fixed

music_assistant/providers/webdav/provider.py
OzGav: yes. Fixed

music_assistant/providers/webdav/provider.py
OzGav: Yes upon review the image retrieval is happening elsewhere so this has been removed

music_assistant/providers/webdav/provider.py
OzGav: After implementing your earlier suggestion which removed the auth_url line then indeed this can be removed

music_assistant/providers/webdav/provider.py
OzGav: Actually I didnt need this at all as its only for providers which support ProviderFeature.LIBRARY_PODCASTS

music_assistant/providers/webdav/provider.py
OzGav: I think this one is required as it calles scandir

If I dont have this then I get this error
```
2025-10-06 23:03:54.080 ERROR (MainThread) [music_assistant.webserver] Error handling message: CommandMessage(message_id=623, command='music/podcasts/podcast_episodes', args={'item_id': '1', 'provider_instance_id_or_domain': 'library'})
  File "/home/ma/server/music_assistant/controllers/webserver.py", line 392, in _run_handler
    async for item in iterator:
  File "/home/ma/server/music_assistant/controllers/media/podcasts.py", line 102, in episodes
    async for episode in self._get_provider_podcast_episodes(
  File "/home/ma/server/music_assistant/controllers/media/podcasts.py", line 235, in _get_provider_podcast_episodes
    async for item in prov.get_podcast_episodes(item_id):
  File "/home/ma/server/music_assistant/providers/filesystem_local/__init__.py", line 760, in get_podcast_episodes
    for item in await asyncio.to_thread(sorted_scandir, self.base_path, prov_podcast_id):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ma/server/music_assistant/providers/filesystem_local/helpers.py", line 224, in sorted_scandir
    for x in os.scandir(sub_path)
             ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'http://192.168.1.250:5005/Podcast-Test/A History of Rock Music in 500 Songs'
```

music_assistant/providers/webdav/provider.py
OzGav: I think it does as the base uses os.scandir, but this provider uses _scan_recursive and _process_webdav_item which do the network operations.

OzGav:
Currently refactoring. Problems appear to be arising due to the base class using os.scandir

marcelveldt:
> Currently refactoring. Problems appear to be arising due to the base class using os.scandir

Well, if needed we can make some changes to the filesystem provider to make it more flexible ?

marcelveldt:
BTW: It already looks loads better now! Only thing maybe is that also the sync and browse methods can be deleted and just use the base implementation, unless I'm missing something

OzGav:
edit: OK I have a possible way ahead. I will look at submitting a completely revised PR in the next few days. Thanks as usual for your patience and pointers as I learn!

OzGav:
Please have a look at the direction I am going. However, I may have hit a wall with podcasts and audiobooks but I havent had time to fully look at it. I would appreciate your thoughts on Claude's assessment
===========================================
You're right - it's an infinite loop. Look at the pattern:

Scans "Moby Dick" folder → finds 45 MP3 files
Processes first file → triggers _get_chapters_for_audiobook
_get_chapters_for_audiobook scans "Moby Dick" folder again → finds 45 files again
Processes each file → each one triggers _get_chapters_for_audiobook again
Infinite loop

The problem: The parent's _get_chapters_for_audiobook is designed for LOCAL filesystems where it can quickly re-scan a folder. For WebDAV, every scan is a network call, and it's processing ALL 45 files every time, each triggering another scan.
The solution: WebDAV needs to override _get_chapters_for_audiobook to prevent this. But this defeats the purpose of the refactor.
Actually, looking closer: The parent has logic to handle this with the IsChapterFile exception (when tags.track > 1). But it's not being raised properly because the files don't have proper track numbers in their tags.
This is why your old custom code worked - you handled multi-file audiobooks differently during the sync phase, caching chapters upfront and marking the folder as processed.
This refactor won't work cleanly for audiobooks/podcasts without significant additional complexity. The parent's design assumes local filesystem efficiency that doesn't translate well to network storage.
Recommendation: Keep the refactor for music only, but keep your custom audiobook/podcast processing from the original code.

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -354,9 +353,32 @@ async def sync_library(self, media_type: MediaType) -> None:
...
 
+        # Run the sync implementation (can be overridden by subclasses)
+        cur_filenames = await self._sync_library_impl(file_checksums)
+
+        end_time = time.time()
+        self.logger.info(
+            "Library sync for %s completed in %.2f seconds",
+            self.name,
+            end_time - start_time,
+        )
```
marcelveldt: does this still make sense now that we have the tasks controller which already provides and logs this info ?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -762,7 +773,7 @@ async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> lis
         await self.mass.cache.set(
             key=cache_key,
             data=result,
-            expiration=3600 * 24,  # Cache for 24 hours
+            expiration=3600 * 24 * 365,  # Cache for 1 year
```
marcelveldt: I would extend the comment why you cache this long: because we rely on the file timestamp as checksum

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1241,7 +1252,7 @@ async def _parse_audiobook(self, file_item: FileSystemItem, tags: AudioTags) ->
 
         # parse other info
         audio_book.authors.set(tags.writers or tags.album_artists or tags.artists)
-        audio_book.metadata.genres = set(tags.genres)
+        audio_book.metadata.genres = set(tags.genres) if tags.genres else {"Spoken Word"}
```
marcelveldt: is "Spoken Word" a valid top level genre we provide ?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1299,7 +1309,9 @@ async def _parse_podcast_episode(
         """Parse full PodcastEpisode details from file tags."""
         # ruff: noqa: PLR0915
         podcast_name = tags.album or file_item.parent_name
-        podcast_path = get_relative_path(self.base_path, file_item.parent_path)
+        # Use relative_parent_path for WebDAV compatibility (parent_path uses os.path.dirname
+        # which doesn't work correctly for URLs)
```
```suggestion
```
I think this comment can be dropped

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1329,6 +1341,7 @@ async def _parse_podcast_episode(
                 name=podcast_name,
                 sort_name=tags.album_sort,
                 publisher=tags.tags.get("publisher"),
+                total_episodes=0,
```
marcelveldt: why is this ?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1353,7 +1366,7 @@ async def _parse_podcast_episode(
                 )
             )
         # parse other info
-        episode.metadata.genres = set(tags.genres)
+        episode.metadata.genres = set(tags.genres) if tags.genres else {"Spoken Word"}
```
marcelveldt: should this default/fallback genre be a constant ?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1606,8 +1622,7 @@ async def _get_local_images(
         if extra_thumb_names is None:
             extra_thumb_names = ()
         images: UniqueList[MediaItemImage] = UniqueList()
-        abs_path = self.get_absolute_path(folder)
-        folder_files = await asyncio.to_thread(sorted_scandir, self.base_path, abs_path, sort=False)
+        folder_files = await self._scandir_impl(folder)
```
marcelveldt: should we rename this to just _scandir or "_traverse_directory" as method name ?
I find the _scandir_impl name a bit weird tbh

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1664,40 +1679,13 @@ async def check_write_access(self) -> None:
         except Exception as err:
             self.logger.debug("Write access disabled: %s", str(err))
 
-    async def resolve(
-        self,
-        file_path: str,
-    ) -> FileSystemItem:
+    async def resolve(self, file_path: str) -> FileSystemItem:
```
marcelveldt, line 1659: why aren't you just overriding this method in webdav provider ?
I dont get all this _impl adjustments

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1664,40 +1679,13 @@ async def check_write_access(self) -> None:
...
-        return await asyncio.to_thread(_create_item)
+        return await self._resolve_impl(file_path)
 
     async def exists(self, file_path: str) -> bool:
         """Return bool is this FileSystem musicprovider has given file/dir."""
-        if not file_path:
-            return False  # guard
-        abs_path = self.get_absolute_path(file_path)
-        return bool(await exists(abs_path))
+        return await self._exists_impl(file_path)
```
marcelveldt: same comment, why not just override the base methods

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1799,7 +1787,11 @@ async def _get_stream_details_for_audiobook(self, item_id: str) -> StreamDetails
                 stream_type=StreamType.LOCAL_FILE,
                 duration=duration,
                 path=[
-                    MultiPartPath(path=self.get_absolute_path(path), duration=duration)
+                    MultiPartPath(
+                        # For WebDAV, paths are already full URLs; for local files, join with base
+                        path=path if path.startswith("http") else self.get_absolute_path(path),
```
marcelveldt: weird - you should instead just override here

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1951,3 +1947,39 @@ async def _get_podcast_metadata(self, podcast_folder: str) -> dict[str, Any]:
...
+                relative_path=get_relative_path(self.base_path, file_path),
+                absolute_path=absolute_path,
+                is_dir=False,
+                checksum=str(int(stat_info.st_mtime)),
+                file_size=stat_info.st_size,
+            )
+
+        return await asyncio.to_thread(_create_item)
+
+    async def _scandir_impl(self, path: str) -> list[FileSystemItem]:
```
marcelveldt: I'd like you to drop all the "_imp" methods and just override the base class in your webdav provider

music_assistant/providers/filesystem_local/__init__.py
OzGav: Agreed. sync related info log messages removed

music_assistant/providers/filesystem_local/__init__.py
OzGav: Comment updated

music_assistant/providers/filesystem_local/__init__.py
OzGav: Yes it is
image

music_assistant/providers/filesystem_local/__init__.py
OzGav: Removed

music_assistant/providers/filesystem_local/__init__.py
OzGav: Yes good idea. Will make maintenance easier

music_assistant/providers/filesystem_local/__init__.py
OzGav: The _impl pattern separated "interface logic" from "storage logic" so the override point was clean but looking back at it it was a minor thing really so will adjust these.

music_assistant/providers/filesystem_local/__init__.py
OzGav: _scandir it is

music_assistant/providers/filesystem_local/__init__.py
OzGav: fixed

music_assistant/providers/filesystem_local/__init__.py
OzGav: done

music_assistant/providers/filesystem_local/__init__.py
OzGav: Done

music_assistant/providers/filesystem_local/__init__.py
OzGav: Total episode count isn't known when parsing individual episode files. Didn't realise the field was optional — removed.

OzGav:
Refactored the parent filesystem provider to expose small overridable methods `(_scandir`, `_read_file`, `_get_chapter_path`) instead of using direct `aiofiles.open` and `os.scandir` calls throughout. This lets the WebDAV provider override just the storage access layer without duplicating any parsing or business logic. 

Also I noticed several places in the local file system provider where `aiofiles.open` was used directly such as for reading NFO files, playlist files, lyrics files, and audiobook descriptions — all now go through `_read_file`

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1402,6 +1388,9 @@ async def _parse_podcast_episode(
             episode.podcast.metadata.add_image(episode.image)
         elif not episode.image and episode.podcast.image:
             episode.metadata.add_image(episode.podcast.image)
+        # ensure podcast has a default genre if none set
+        if not episode.podcast.metadata.genres:
+            episode.podcast.metadata.genres = {"Spoken Word"}
```
```suggestion
            episode.podcast.metadata.genres = {DEFAULT_AUDIOBOOK_PODCAST_GENRE}
```

music_assistant/providers/webdav/__init__.py
marcelveldt: no not fixed as they are still here, duplicated

music_assistant/providers/webdav/__init__.py
marcelveldt: no not fixed as they are still here, duplicated

music_assistant/providers/webdav/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+async def webdav_propfind(
+    session: aiohttp.ClientSession,
+    url: str,
+    depth: int = 1,
+    timeout: int = 30,
+    auth: aiohttp.BasicAuth | None = None,
+) -> list[WebDAVItem]:
+    """Execute a PROPFIND request on a WebDAV resource.
```
marcelveldt:
```suggestion
    """
    Execute a PROPFIND request on a WebDAV resource.
```

music_assistant/providers/webdav/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+async def webdav_test_connection(
+    base_url: str,
+    username: str | None,
+    password: str | None,
+    verify_ssl: bool,
+    timeout: int = 10,
+) -> None:
+    """Test WebDAV connection and authentication.
```
marcelveldt:
```suggestion
    """
    Test WebDAV connection and authentication.
```

music_assistant/providers/webdav/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+    :param password: Optional password.
+    :param verify_ssl: Whether to verify SSL certificates.
+    :param timeout: Connection timeout in seconds.
+    :raises LoginFailed: Authentication failed.
+    :raises SetupFailedError: Connection or configuration error.
+    """
+    auth = aiohttp.BasicAuth(username, password) if username and password else None
+    connector = aiohttp.TCPConnector(ssl=verify_ssl)
+
+    async with aiohttp.ClientSession(connector=connector) as session:
```
marcelveldt: why is this not using the MA global client session ?

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,328 @@
...
+            return parsed.path[len(base_parsed.path) :].strip("/")
+        return path
+
+    async def exists(self, file_path: str) -> bool:
+        """Check if WebDAV resource exists."""
+        if not file_path:
+            return False
+        file_path = self._normalize_path(file_path)
+        webdav_url = build_webdav_url(self.base_url, file_path)
+        session = self.mass.http_session if self.verify_ssl else self.mass.http_session_no_ssl
```
marcelveldt: why dont you have a property that has this logic ? now you're repeating this over and over

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,328 @@
...
+
+    async def _scandir(self, path: str) -> list[FileSystemItem]:
+        """List WebDAV directory contents with caching."""
+        cache_key = f"scandir_{path}"
+        if cached := await self.cache.get(
+            key=cache_key,
+            provider=self.instance_id,
+            category=0,
+        ):
+            return cast("list[FileSystemItem]", cached)
```
marcelveldt: either provide the base class to the cache.get method or instantiate FileSystemItem from dict here.

music_assistant/providers/webdav/provider.py
```diff
@@ -0,0 +1,328 @@
...
+        path = self._normalize_path(path)
+        webdav_url = build_webdav_url(self.base_url, path)
+        session = self.mass.http_session if self.verify_ssl else self.mass.http_session_no_ssl
+
+        webdav_items = await webdav_propfind(session, webdav_url, depth=1, auth=self._auth)
+        filesystem_items = self._convert_webdav_items(webdav_items, webdav_url, path)
+
+        await self.cache.set(
+            key=cache_key,
+            data=filesystem_items,
```
marcelveldt: this should be serializable objects

music_assistant/providers/webdav/helpers.py
OzGav: No good reason upon review so fixed.

music_assistant/providers/webdav/provider.py
OzGav: Of course. Fixed.

music_assistant/providers/webdav/provider.py
OzGav: Done

music_assistant/providers/webdav/provider.py
OzGav: Fixed

music_assistant/providers/webdav/__init__.py
OzGav: Somehow they have crept back in after the last many months. Definitely fixed now!

music_assistant/providers/webdav/__init__.py
OzGav: Same again

marcelveldt (approved):
Let's go, nice work @OzGav !
Just do me one favor and properly test the fileprovider for any regressions

PR #2485, opened 2025-10-05
Various small fixes for the Sonos provider

marcelveldt:
- Fix for Airplay mode
- Fix for radio playback

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1951,11 +1951,10 @@ class PlayerQueuesController(CoreController):
         if player.current_media.source_id == queue_id and player.current_media.queue_item_id:
             return player.current_media.queue_item_id
         # special case for sonos players
-        if (
-            player.current_media.uri == f"mass:queue:{queue_id}"
-            and player.current_media.queue_item_id
-        ):
-            return player.current_media.queue_item_id
+        if player.current_media.uri.startswith(f"mass:{queue_id}"):
+            if player.current_media.queue_item_id:
+                return player.current_media.queue_item_id
+            return player.current_media.uri.split(":")[-1]
         # try to extract the item id from a mass stream url
         if (
             player.current_media.uri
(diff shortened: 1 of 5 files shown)


Tsacha100:
📝 Title:

Error setting up Sonos S1 player: 'model_name' when HA Sonos integration is active on the same host

Description

Hello,
I’m experiencing an issue with the Sonos S1 provider in Music Assistant.
When I start Music Assistant, I get the following error in the logs:

2025-10-11 20:08:07.090 ERROR (MainThread) [music_assistant.sonos_s1]
Error setting up Sonos player RINCON_000E58C7032E01400: 'model_name'
name'


After this message, the affected Sonos player does not appear in Music Assistant and cannot be used.

Environment

Home Assistant version: (please fill in, e.g. 2025.10.1)

Music Assistant version: (please fill in, e.g. 2025.10.0)

Install type: (e.g. HA Add-on / Docker / Core)

Sonos firmware: S1

Network setup: All devices on the same local network (same subnet)

Host setup: Home Assistant + Music Assistant + HA Sonos integration running on the same host

Details

According to the documentation, it is not recommended to use both the Home Assistant Sonos integration and the Music Assistant Sonos S1 provider on the same host.
However, even when both are active, I would not expect the setup to fail completely with a 'model_name' KeyError.

It seems the MA Sonos S1 provider is attempting to access a property that is missing or unavailable during discovery when both integrations are running simultaneously, possibly due to conflicting UPnP requests or incomplete responses from the speaker.

What I tried

Disabling the HA Sonos integration resolves the issue — MA then discovers all Sonos S1 speakers correctly.

Keeping both active (on the same host) reliably triggers the 'model_name' error.

I also tried specifying manual IP addresses for discovery, but that did not change the result.

Expected behavior

Music Assistant should handle the missing model_name attribute gracefully or warn the user that both integrations cannot coexist on the same host, instead of crashing the setup for that device.

Logs
2025-10-11 20:08:07.090 ERROR (MainThread) [music_assistant.sonos_s1]
Error setting up Sonos player RINCON_000E58C7032E01400: 'model_name'

Suggestions

PR #2491, opened 2025-10-07
YTMusic: Fix disc/track number + favourite status

MarvinSchenkel:
Investigated [this](https://discord.com/channels/753947050995089438/1420743795737231401) issue where YTM tracks would lose their tracknumber in the UI.

- YTM is very inconsistent with `videoId` that is being returned for albums. Had to do some black magic to match them to the audio playlist
- Made sure the AlbumController now saves `album_tracks` when needed
- Added `favorite` to the MediaItems in preparation for our library import logic


marcelveldt:
Can't we just later inject it in the library if we fetch the album tracks ?

MarvinSchenkel:
> Can't we just later inject it in the library if we fetch the album tracks ?

Well I was thinking about that, but that means we would need to change the `AlbumController` to also set `album_tracks` when an Album is updated / inserted. This could work, but this means we would also need to fetch Album tracks when an Album since Tracks are not part of an Album.

Is that really worth the hassle? Note that the information will still be partly incomplete because apparently this information is _not always_ returned from YTM :-)

marcelveldt:
Well, the problem with this is, is that you loose the correct order when playing back albums. So we will have to find a better way somehow. For album tracks we can always consider just returning the index of the items when returned from the api

MarvinSchenkel:
> Well, the problem with this is, is that you loose the correct order when playing back albums. So we will have to find a better way somehow. For album tracks we can always consider just returning the index of the items when returned from the api

Fair enough. In that case I will just assume the order returned from the api and use that. From what I have seen, the tracks are returned in the proper order anyway in the case that we have this information available.

music_assistant/controllers/media/albums.py
```diff
@@ -445,6 +449,36 @@ async def _set_album_artist(
         )
         return ItemMapping.from_item(db_artist)
 
+    async def _set_album_track(self, db_id: int, track: Track) -> None:
+        """Store Album Track info."""
+        db_track: Track | None = None
+        if track.provider == "library":
+            return
+        if existing := await self.mass.music.tracks.get_library_item_by_prov_id(
```
marcelveldt: this is now doing a db call for each track while you already collected them all at line 234 ? Maybe just lookup the existing item to first check if the item already has a track number ?

MarvinSchenkel: Fixed

music_assistant/controllers/media/albums.py
```diff
@@ -249,6 +249,18 @@ async def tracks(
                 provider_mapping.item_id, provider_mapping.provider_instance
             )
             for provider_track in provider_tracks:
+                # In some cases (looking at you YTM) the disc/track number is not obtained from
+                # library_tracks. Ensure to update the disc/track number when interacting with
+                # album tracks
+                db_track = next(
+                    (x for x in db_items if x.sort_name == provider_track.sort_name), None
+                )
+                if db_track and db_track.track_number != provider_track.track_number:
```
marcelveldt:
```suggestion
                if db_track and db_track.track_number == 0 and db_track.track_number != provider_track.track_number:
```

marcelveldt: Let's only accept a result from the provider if the db value is currently broken

music_assistant/controllers/media/albums.py
```diff
@@ -249,6 +249,18 @@ async def tracks(
                 provider_mapping.item_id, provider_mapping.provider_instance
             )
             for provider_track in provider_tracks:
+                # In some cases (looking at you YTM) the disc/track number is not obtained from
+                # library_tracks. Ensure to update the disc/track number when interacting with
+                # album tracks
+                db_track = next(
+                    (x for x in db_items if x.sort_name == provider_track.sort_name), None
+                )
+                if db_track and db_track.track_number == 0 and db_track.track_number != provider_track.track_number:
```
OzGav:
```suggestion
                if (db_track and
                      db_track.track_number == 0 and
                      db_track.track_number != provider_track.track_number):
```

music_assistant/controllers/media/albums.py
```diff
@@ -249,6 +249,22 @@ async def tracks(
                 provider_mapping.item_id, provider_mapping.provider_instance
             )
             for provider_track in provider_tracks:
+                # In some cases (looking at you YTM) the disc/track number is not obtained from
+                # library_tracks. Ensure to update the disc/track number when interacting with
+                # album tracks
+                db_track = next(
+                    (x for x in db_items if x.sort_name == provider_track.sort_name), None
```
marcelveldt: not sure how stable the sort_name is to use - maybe we should also add version here. Anyways, we can adjust it if we may run into issues.

MarvinSchenkel: Fixed!

marcelveldt (approved):

PR #2495, opened 2025-10-08
audiobookshelf: Fix random chapter skipping

tsipinakis:
In case a chapter further than the starting chapter has a duration which makes skipped_duration + duration < seek_positon that chapter would be silently skipped. Lets go back to using a separate function to find the position to fix that.

Additionally, bring back the guard to prevent landing in the final couple of seconds of a track for a smoother chapter seek

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -1287,8 +1287,13 @@ async def get_file_stream(
 async def get_multi_file_stream(
     mass: MusicAssistant,  # noqa: ARG001
     streamdetails: StreamDetails,
+    seek_position: int = 0,
 ) -> AsyncGenerator[bytes, None]:
-    """Return audio stream for a concatenation of multiple files."""
+    """Return audio stream for a concatenation of multiple files.
+
+    Arguments:
+    seek_position: The position to seek to in seconds
+    """
     files_list: list[str] = streamdetails.data
     # concat input files
     temp_file = f"/tmp/{shortuuid.random(20)}.txt"  # noqa: S108
@@ -1306,7 +1311,16 @@ async def get_multi_file_stream(
                 bit_depth=streamdetails.audio_format.bit_depth,
                 channels=streamdetails.audio_format.channels,
             ),
-            extra_input_args=["-safe", "0", "-f", "concat", "-i", temp_file],
+            extra_input_args=[
+                "-safe",
+                "0",
+                "-f",
+                "concat",
+                "-i",
+                temp_file,
+                "-ss",
+                str(seek_position),
+            ],
         ):
             yield chunk
     finally:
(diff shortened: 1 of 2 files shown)


music_assistant/helpers/audio.py
```diff
@@ -944,6 +944,46 @@ async def get_file_stream(
...
+        # the previous track. If we're within 2 second of the end, skip the current track
+        if position + 2 >= part.duration:
+            LOGGER.debug(
+                f"Skipping to the next part due to seek position being at the end: {position}"
+            )
+            return parts[i + 1 :], 0
+
+        return parts[i:], int(position)
+
+    return None, 0
```
marcelveldt: Maybe just raise some IndexError here instead of returning invalid data ?

music_assistant/helpers/audio.py
```diff
@@ -955,19 +995,11 @@ async def get_multi_file_stream(
...
-        if not isinstance(part, MultiPartPath):
-            raise InvalidDataError("Multi-file streamdetails requires a list of MultiPartPath")
-        if seek_position and part.duration and (skipped_duration + part.duration) < seek_position:
-            skipped_duration += part.duration
-            continue
-        files_list.append(part.path)
-    if seek_position:
-        seek_position -= int(skipped_duration)
+    parts, seek_position = _get_part_from_position(streamdetails.path, seek_position)
+    assert parts is not None
```
marcelveldt: I would ditch this assert and just raise an exception in the _get_part_from_position if a invalid combination of params is given

music_assistant/helpers/audio.py
```diff
@@ -944,6 +944,46 @@ async def get_file_stream(
...
+            raise InvalidDataError("Track has None duration, cannot find position")
+        if skipped_duration + part.duration < seek_position:
+            skipped_duration += part.duration
+            continue
+
+        position = seek_position - skipped_duration
+
+        # Seeking in some tracks is inaccurate, making the seek to a chapter land on the end of
+        # the previous track. If we're within 2 second of the end, skip the current track
+        if position + 2 >= part.duration:
```
marcelveldt: Maybe this can even be 5 seconds ?
I don't know, I don't have/use audiobooks to test this but it feels unlikely you will be seeking in the last X seconds of a chapter

music_assistant/helpers/audio.py
```diff
@@ -944,6 +944,46 @@ async def get_file_stream(
...
+            raise InvalidDataError("Multi-file streamdetails requires a list of MultiPartPath")
+        if part.duration is None:
+            raise InvalidDataError("Track has None duration, cannot find position")
+        if skipped_duration + part.duration < seek_position:
+            skipped_duration += part.duration
+            continue
+
+        position = seek_position - skipped_duration
+
+        # Seeking in some tracks is inaccurate, making the seek to a chapter land on the end of
```
marcelveldt: should we replace the word track here with chapter ?

fmunkes: Please leave this as track, or even file. A chapter might span multiple files.

marcelveldt: then maybe 'part' is the right word here ?

music_assistant/helpers/audio.py
tsipinakis: According to my tests 2 seconds is good enough - the thing to avoid is to have a few "unintelligible" seconds before the next chapter starts. 2 seconds is long enough for that.

PR #2496, opened 2025-10-08
Support multiple accounts in the home screen recommendations for Tidal

### LINKED ISSUE #4295: Multiple Tidal Accounts break Home Screen

2.6.0

Docker Container

### Carefully read the Troubleshooting FAQ and confirm that

- [x] I have fixed any errors in the logs that relate to tags
- [x] I have disabled tool such as AdGuard, Pi-hole or pfSense and retested

### Have you reviewed the [Open](https://github.com/music-assistant/support/issues) and [Closed](https://github.com/music-assistant/support/issues?q=is%3Aissue%20state%3Aclosed) Issues and [Discussions](https://github.com/orgs/music-assistant/discussions) to resolve this yourself?

- [x] Yes

### The problem

It's great that Tidal "Custom Mixes" and "Favorite Radio Stations" now show up on the Home Screen.
However, I have two Tidal accounts connected to MA, and the following happens:

 - Every section (e.g. Custom Mixes) shows up twice, but the content is identical and seems to come only from one Tidal account
 - Disabling one of the two sections disables both
 - Editing the Home Screen becomes quite quirky: Clicking the move buttons on one of the two entries results in things moving in unexpected ways.

Would it be possible to:
 - Make the two sections display the content from both accounts?
 - Ideally (much less important) also allow disabling/moving them separately?

Thank you!

### How to reproduce

Connect two Tidal accounts to a MA instance and visit the Home screen.

### Music Providers

This is Tidal-specific.

### Player Providers

I'm using Airplay players, but I do not think this is relevant.

### Additional information

<none>

2025.9.4

Home Assistant Container

- [x] Yes

jozefKruszynski:
Add user ids to home rows and cache keys to ensure multiple tidal accounts are usable.

Fixes: https://github.com/music-assistant/support/issues/4295

### DIFF
--- a/music_assistant/providers/tidal/__init__.py
+++ b/music_assistant/providers/tidal/__init__.py
@@ -61,7 +61,6 @@ from .constants import (
     BROWSE_URL,
     CACHE_CATEGORY_ISRC_MAP,
     CACHE_CATEGORY_RECOMMENDATIONS,
-    CACHE_KEY_RECOMMENDATIONS_ALL,
     CONF_ACTION_CLEAR_AUTH,
     CONF_ACTION_COMPLETE_PKCE_LOGIN,
     CONF_ACTION_START_PKCE_LOGIN,
@@ -1011,19 +1010,9 @@ class TidalProvider(MusicProvider):
         except (ClientError, KeyError, ValueError) as err:
             raise MediaNotFoundError(f"Playlist {prov_playlist_id} not found") from err
 
+    @use_cache(expiration=3600, category=1)
     async def recommendations(self) -> list[RecommendationFolder]:
         """Get this provider's recommendations organized into folders."""
-        # Check cache first
-        cached_recommendations: list[RecommendationFolder] = await self.mass.cache.get(
-            CACHE_KEY_RECOMMENDATIONS_ALL,
-            provider=self.instance_id,
-            category=CACHE_CATEGORY_RECOMMENDATIONS,
-        )
-
-        if cached_recommendations:
-            self.logger.debug("Returning cached recommendations (TTL: 1 hour)")
-            return cached_recommendations
-
         results: list[RecommendationFolder] = []
 
         # Pages to fetch
@@ -1062,15 +1051,6 @@ class TidalProvider(MusicProvider):
             )
 
             self.logger.debug("Created %d recommendation folders from Tidal", len(results))
-
-            # Cache the results for 1 hour (3600 seconds)
-            await self.mass.cache.set(
-                key=CACHE_KEY_RECOMMENDATIONS_ALL,
-                data=results,
-                provider=self.instance_id,
-                category=CACHE_CATEGORY_RECOMMENDATIONS,
-                expiration=3600,
-            )
 
         except (ClientError, ResourceTemporarilyUnavailable) as err:
             # Network-related errors
--- a/music_assistant/providers/tidal/__init__.py
+++ b/music_assistant/providers/tidal/__init__.py
@@ -1041,20 +1041,24 @@ class TidalProvider(MusicProvider):
         module_page_names: dict[str, str] = {}
 
         try:
-            # Get the number of configured tidal instances once for this operation
-            tidal_instance_count = await self.mass.config.get_provider_instance_count(self.domain)
+            # Get all Tidal provider instances - await the coroutine
+            all_tidal_configs = await self.mass.config.get_provider_configs(ProviderType.MUSIC)
+            # Filter to only Tidal configs
+            tidal_configs = [config for config in all_tidal_configs if config.domain == self.domain]
+            # Sort by instance_id to get a consistent "first" instance
+            sorted_instances = sorted(tidal_configs, key=lambda x: x.instance_id)
             # Process pages and collect modules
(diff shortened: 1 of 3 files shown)


music_assistant/providers/tidal/__init__.py
```diff
@@ -1403,7 +1428,7 @@ async def get_page_content(self, page_path: str = "pages/home") -> TidalPagePars
                 "parsed_at": parser._parsed_at,
             }
             await self.mass.cache.set(
-                key=page_path,
+                key=f"{self.auth.user_id}_{page_path}",
                 data=cache_data,
                 provider=self.instance_id,
```
marcelveldt, line 1454: So... as you can see here, the instance_id is already being provided as unique cache key.
So you dont have to adjust the key above

music_assistant/providers/tidal/__init__.py
```diff
@@ -1158,25 +1161,47 @@ def get_icon_for_type(media_type: MediaType) -> str:
...
+            user_identifier = None
+            if self.auth.user and self.auth.user.profile_name:
+                user_identifier = self.auth.user.profile_name
+            elif self.auth.user and self.auth.user.user_name:
+                user_identifier = self.auth.user.user_name
+            else:
+                user_identifier = str(self.auth.user_id)
+
+            # Create folder name with user identifier
+            folder_name = f"{module_title} ({user_identifier})"
```
marcelveldt: the downside of this is that the name is now also being appended even if you only have only single Tidal account. Is that a problem ? Otherwise we can consider adding some helper to our config controller to find the number of instances ?

jozefKruszynski: I also thought about this, and was unsure about it. I personally do not mind seeing my username on the tidal rows, but some people might be annoyed by it.

Adding an instance count helper sounds sensible, should I do so and add it to this PR, or add a separate PR just for the instance count helper?

I assume the second option is favourable as it would be a general helper and not just for tidal.

music_assistant/providers/tidal/__init__.py
jozefKruszynski: That makes sense. I'll remove it.

music_assistant/providers/tidal/__init__.py
marcelveldt: Feel free to add it in this PR, its just a very small, isolated helper function

music_assistant/providers/tidal/__init__.py
marcelveldt: you can also replace all of this by a single use_cache() decorator on the function. It accepts the base_key and category as parameters

music_assistant/providers/tidal/__init__.py
```diff
@@ -1078,13 +1091,37 @@ async def _process_recommendation_pages(
...
+            # Check if we should skip this page for this instance
+            if page_path == "pages/home" and show_user_identifier:
+                # Get all Tidal provider instances - await the coroutine
+                all_tidal_configs = await self.mass.config.get_provider_configs(ProviderType.MUSIC)
+                # Filter to only Tidal configs
+                tidal_configs = [
+                    config for config in all_tidal_configs if config.domain == self.domain
+                ]
+                # Sort by instance_id to get a consistent "first" instance
+                sorted_instances = sorted(tidal_configs, key=lambda x: x.instance_id)
```
marcelveldt: so... actually because of this you already had a overview of the instances... you could have simply counted the configs here

jozefKruszynski: A question then, do you think it makes sense to change the helper function in the config controller to return not just a count but the instance ids so that then, in case other providers need similar functionality, it is already there?

Or, KISS and remove the helper function until such time that another provider actually does need similar functionality?

I suppose option 2 is the more sensible approach.

marcelveldt: yeah, option 2.

music_assistant/providers/tidal/__init__.py
```diff
@@ -1010,51 +1010,48 @@ async def _get_mix_playlist_tracks(
         except (ClientError, KeyError, ValueError) as err:
             raise MediaNotFoundError(f"Playlist {prov_playlist_id} not found") from err
 
+    @use_cache(expiration=3600, category=1)
```
marcelveldt:
```suggestion
    @use_cache(expiration=3600, category=CACHE_KEY_RECOMMENDATIONS_ALL)
```

music_assistant/providers/tidal/__init__.py
```diff
@@ -60,7 +61,6 @@
     BROWSE_URL,
     CACHE_CATEGORY_ISRC_MAP,
     CACHE_CATEGORY_RECOMMENDATIONS,
-    CACHE_KEY_RECOMMENDATIONS_ALL,
     CONF_ACTION_CLEAR_AUTH,
```
marcelveldt, line 64:
```suggestion
    CACHE_CATEGORY_RECOMMENDATIONS,
    CACHE_KEY_RECOMMENDATIONS_ALL,
    CONF_ACTION_CLEAR_AUTH,
```

marcelveldt (approved):
LGTM!

PR #2497, opened 2025-10-09
fix: audiobookshelf: authenticate not necessary in streamdetails

fmunkes:
This line is not necessary, we handle a bad token via ffmpeg failing, and it blocks streaming for token only users.


OzGav:
backport to stable label?

fmunkes:
> backport to stable label?

No, we did not include the JWT authorization in stable, and this ultimately introduced the bug.

Once abs drops the old system completely,  I'll come back to think about backporting/ including in a point release.

MarvinSchenkel (approved):

PR #2498, opened 2025-10-10
Add VBAN Receiver plugin provider

sprocket-9:
This PR is a plugin provider to allow MA to act as a VBAN (PCM over UDP) protocol receiver: 
https://vb-audio.com/Voicemeeter/vban.htm
https://vb-audio.com/Voicemeeter/VBANProtocol_Specifications.pdf

## Use case
Stream whole system or per-app audio into MA as a kind of network AUX input.

## Requirements
A VBAN sender.

Their website has some very advanced apps for routing Windows system audio over VBAN, which I still need to try dabbling with.  In the VBAN Receiver plugin settings, match the stream configuration settings.

Pipewire-based systems have vban_receive/vban_send modules https://docs.pipewire.org/page_module_vban_send.html. 

## Testing
I've been using pipewire which is working well, so testing is focused on this method.  A couple of pipewire config files are required, one to set up the vban sender and the other to prevent the output sink for squeezelite changing on pipewire events.

~/.config/pipewire/pipewire.conf.d/01-vban.conf

```
{
        name = libpipewire-module-vban-send
        args = {
            source.ip = 0.0.0.0
            destination.ip = 127.0.0.1 # MA Dev server IP
            destination.port = 6980
            sess.min-ptime = 1
            sess.latency.msec = 2
            sess.name = "Network AUX"
            audio.format = "S16LE"
            audio.rate = 44100
            stream.props = {
                media.class = "Audio/Sink"
                node.name = "vban-sender-aiovban"
                node.description = "VBAN aiovban sender"
            }
        }
    }
```

~/.config/pipewire/pipewire-pulse.conf.d/squeezelite-static-target-object.conf

```
# When AVR power status changes, it was causing squeelite clients
# to switch playback to the default audio/sink HDMI device
# This rule prevents this happening
pulse.rules = [
    {
        matches = [
            {
                # all keys must match the value. ! negates. ~ starts regex.
                application.process.binary = "squeezelite-pulse"
            }
        ]
        actions = {
            update-props = {
                node.dont-reconnect = true
            }
        }
    }
]
```

The vban_send module creates an audio sink, available as an audio output in the desktop enviroment sound manager for routing whole system or per-app audio to. 

Alternatively, start an instance of squeezelite, setting -s option to the dev server:

`/usr/bin/squeezelite-pulse -o vban-sender-aiovban -n VBANsender -s localhost -m 01:00:00:00:00:01`

This squeezelite MA player will now function as a VBAN sender for plugin testing - whatever is played is converted to VBAN packets and loopbacked to localhost:6980 for reception by this PR's plugin and out to any standard MA player.

Use a standard MA player to listen:  player's UI burger menu entry: Select Source -> VBAN Receiver.  Due to MA buffering there will be a good few seconds of delay before audio is heard.

Of course to really test it out, best try starting the sender on another device and check performance over the LAN (change destination.ip in 01-vban.conf and -s option in squeezlite).  I started a sender on my prod MA server and used the VBANsender player to stream VBAN to the VBAN Receiver plugin on my dev server - takes a bit of UI juggling and can get a little confusing sometimes when debugging issues :smiley: 
In my testing it's working well over wired and wifi, but obviously every network is different so YMMV.     

qpwgraph is a GUI pipewire graph manager which is very useful for visualising and patching pipewire nodes, as is 'wpctl status/inspect' cli command.

## Issues
1. The plugin is multi instance, but due to how the underlying aiovban module currently works with MA, each instance must have a different Sender IP address and UDP port configured in the plugin settings, so no using multiple VBAN stream names from the same sender IP.  Working on a fix in aiovban.

2. At least with squeezelite players, selecting the plugin as a player's source will play the audio stream, but the frontend UI will not reflect this or show the plugin as the active source.  The WS msg `player_updated` shows the player's active_source is still the player_id mac,  not the plugin's source_id.  In 2.7.0b3 the spotify connect plugin also has this issue now, so maybe a regression from recent refactoring?

3. If a plugin is in use by player1, MA UI will show the plugin as selectable for player2, but MA server checks and  prevents this.  I think for this kind of plugin it would be nice to allow player 2 to become the active listener and kick player1 out.

4. Audio delay between initiating playback and hearing the audio.  I read the Discord comments re: possibility of reducing delays in MA's PCM handling for the local audio plugin proposal, so that would be very welcome and appreciated here too!  
I've used VBAN to watch a video and stream the app audio out ceiling speakers with imperceptible lip sync issues, which was pretty nice.  This was using pipewire to link the output of the VBAN receiver source node direct to the HW sink for minimal delay though.  Instead of fiddling on the cli it would be great to use the MA UI to select the plugin source for a player and have MA switch to a low delay "raw" pcm mode.  Probably have to mess with buffers on the squeezelite side too.  Anyway, the video application is a niche use case within an already niche plugin :relaxed: , any improvements on lowering MA's PCM processing audio delay would be sweet!

6.  The large list of supported VBAN PCM sample rates in the plugin configuration UI is grabbed from the aiovban source - not sure if these should be paired down to a subset compatible with MA?

## Todo
All done.


MarvinSchenkel:
Sending some 🎁 os your way! Your PR looks good to me, just waiting for you to address the todo's before merging.

marcelveldt:
Marked as draft as there are still some TODO's left - please hit the "ready for review" button once you are ready with the PR.
As for optimizing the backend for lower latency: I am looking into that, independent of this PR. Its tricky though because some players simply require a bit of buffer when streaming from HTTP.

marcelveldt:
FYI: The optimizations to handle PlugInSource in MA are now merged in dev. 
This fixes your bullet 2 and 4.

For point 3 I'll see what we can do.

sprocket-9:
Thanks for the heads-up, been trying out the latest changes.

2.7.0b5
MA source: radioparadise station
Plugin source: spotify connect
Player: squeezelite

On 2: First transition from MA -> plugin source, it works ok, audio plays first time and plugin is shown as active source in UI.  After that:  
* Switching back to play MA radio, the plugin source is still shown as active source in UI.
* MA -> Plugin or plugin -> MA transitions always takes two presses of the play button in either MA or spotify app side.  First press stops the current source, second press plays the new source. 

Can supply logs if required but I think squeezelite is your test player anyway so should be reproducible.

On 3:  :+1:

On 4: Due to the UI bug with 2. I haven't been able to test the new changes fully yet with my plugin.

sprocket-9:
Tried things out with the vban plugin this time.

vban sender/receiver/PluginSource.audio_format all set at 44100/16.

Don't know if the intent is for MA to output at the same rate as the source to save conversions, but it's using the highest supported sample rate as set in the player's settings.

When MA's final output format is using a higher sample rate than the source, the audio plays slowed down.
Forcing MA to use 44100/16 by only having this rate enabled in the squeezelite player's settings and the audio plays at the correct speed.

sprocket-9:
# VBAN Receiver Plugin Provider
This plugin allows Music Assistant to act as a VBAN protocol (PCM audio over UDP) receiver. 

Think of it as a kind of network AUX input for MA, where it receives a remote VBAN audio source and makes it available to MA as a streaming audio source for a Player.

Examples of remote sources which can be sent over VBAN:
* Whole system audio
* Application audio such as a web browser or music app.
* Soundcard input such as Mic or Line-in

# Configuration
The plugin is multi-instance so ensure each instance of the plugin is listening on its own unique UDP port and configured to match the stream parameters as set by the VBAN Sender:

* UDP port
* Stream name
* Sample rate
* Bit depth/audio format
* Channels

The advanced settings VBAN queue size/strategy should normally not need to be changed.   

To listen to the plugin audio: Player's UI burger menu -> Select Source -> VBAN Receiver. 

## VBAN Senders

### VB-Audio Windows Apps
Most of the [Voicemeeter apps](https://vb-audio.com/index.htm) (from the creator of the VBAN protocol) have VBAN Sender functionality, the "simplest" one being Voicemeeter Banana.  These audio mixer apps are full-featured, fairly complex and configuration highly system dependent.  The documentation and Youtube videos far better sources of information on how to configure them than can be supplied here. The basic idea is the apps will install many Voicemeeter system audio devices, select the Voicemeeter Input device as the default output device to route whole system audio over VBAN, or select it as the output for an application in the Volume Mixer.

Basic tips:
* Make sure to run the install program as Administrator.

* In the Hardware Out section, configure the A1 Output Device as a MME device.

* For some, disabling power saving on network cards may be necessary.

Press the VBAN button to configure an Outgoing Stream:
* IP Address to:  MA Server IP
* Stream name, sample rate and audio format - mirror these in the VBAN Receiver plugin settings
* UDP Port - Help documentation shows the app used to allow setting the VBAN Receiver's UDP port here, but that's no longer the case and defaults to port 6980 so use this port in the VBAN Receiver plugin settings.
* Press "On" to start the stream.

If error numbers/red lights appear next to the outgoing VBAN stream in the Voicemeeter VBAN section this is a **sender-side issue, NOT receiver/MA side**.  Something in the mixer configuration is likely wrong.

# Pipewire
Pipewire systems have the [vban-send](https://docs.pipewire.org/page_module_vban_send.html) module which creates a system audio sink.  Any audio sent to the sink is converted to VBAN packets and sent to the VBAN Receiver at the destination ip:port which needs set to the IP address of the MA server and listener port in the VBAN Receiver plugin.

`~/.config/pipewire/pipewire.conf.d/01-vban.conf`

```
{
	name = libpipewire-module-vban-send
	args = {
		source.ip = 0.0.0.0
		destination.ip = 127.0.0.1 # MA Server IP
		destination.port = 6980 # VBAN Receiver plugin listener port
		sess.name = "Network AUX" # Match in VBAN Receiver plugin
		audio.format = "S16LE" # Match in VBAN Receiver plugin
		audio.rate = 44100 # Match in VBAN Receiver plugin
		audio.channels = 2 # Match in VBAN Receiver plugin
		stream.props = {
			media.class = "Audio/Sink"
			node.name = "vban-sender-ma"
			node.description = "VBAN sender for MA"
		}
	}
}
```

In a desktop environment, the sink is available as an audio output in the sound manager UI for routing whole system or per-app audio to.

In a headless environment the [pw-link](https://docs.pipewire.org/page_man_pw-link_1.html) command will link audio output ports to the vban-sender-ma input ports by port ID.  The links are only temporary and need recreated on reboot or pipewire restart, where the port IDs will likely have changed.  Dynamic linking using node properties rather than port IDs is achievable using the Wireplumber session manager and a lua script:

https://bennett.dev/auto-link-pipewire-ports-wireplumber/
https://github.com/bennetthardwick/dotfiles/blob/master/.config/wireplumber/scripts/auto-connect-ports.lua
https://franks-reich.net/posts/creating_pipewire_links_with_wireplumber/

## Known Issues / Notes
While VBAN is designed for real-time audio,  this is not the ultimate goal of this plugin - getting remote system audio into MA is.  That being said, the plugin simply acts as a middle-man and sends the incoming packets from the Sender straight to MA for processing as quickly as it arrives.  MA is designed to handle the audio from plugins as quickly as possible, so the processing delay should be pretty low, however the audio buffers used by the various Players supported by MA is also a factor, so there will be at least some delay in the audio output.

The audio is sent over connectionless UDP packets,  so the quality of the network is a factor in the end results.  Wired network vs WiFi, dropped packets, high latency, jitter etc all play their part.  UDP does not retransmit lost packets.

A VBAN Sender which is a low power device or device under high utilization which struggles to send VBAN packets in a timely fashion may result in choppy audio.

This plugin only supports the VBAN AUDIO sub protocol type, no other sub protocols are supported.

sprocket-9:
@marcelveldt ready for review thanks.  As this was marked as a draft not sure if you saw my comments above, but there's still a couple of issues with the PluginSource changes - UI not tracking active source correctly and audio playing at the incorrect speed when the player sample rate does not match the source sample rate.

@OzGav I've added some docs in the comment above for you to have a look at, thanks.

sprocket-9:
Hmm noticing high cpu use with the plugin enabled.  Will change back to a draft while I look into it.

marcelveldt:
In the meanwhile I fixed the issues you reported with the pluginsources so would be nice if you can retest that as well

sprocket-9:
Done a little more testing.

Tested at: 6cff699d924e549f693809898dc750bf4da382ac
MA source: radioparadise station
Plugin source: vban receiver
Player: squeezelite

1. Possible timing issue with player plugin source takeover?  P1 playing plugin source and P2 selects plugin source in UI.  The logging order suggests there's a small window between stopping P1 and starting P2 when two plugin get_audio_streams() could be running and reading packets from the plugin queue.

```
DEBUG (MainThread) [music_assistant.players] Plugin source VBAN Receiver: Test is already in use by player P1, stopping playback there first.
DEBUG (MainThread) [music_assistant.players] Handling command cmd_stop for player P1
DEBUG (MainThread) [music_assistant.players] Handling command play_media for player P2
DEBUG (MainThread) [music_assistant.streams] Start streaming PluginSource vban_receiver--7F6uGHkz to P2
DEBUG (MainThread) [music_assistant.vban_receiver.7F6uGHkz] Getting VBAN PCM audio stream for Player: P2
DEBUG (MainThread) [music_assistant.streams] Finished streaming PluginSource vban_receiver--7F6uGHkz to P1
```

If P1 doesn't stop cleanly for whatever reason it could get messy and I did experience this sometimes while writing the plugin.  My solution was using a per plugin id asyncio.Lock and trying a pluginlock.acquire() with timeout in streams.get_plugin_source_stream to ensure P1 has definitely stopped.  Or can this be better handled somewhere in the plugin itself?


2. Plugin -> MA Queue transitions requires doing UI -> Select Source -> "MA Queue" twice.
MA Queue -> Plugin transition works first time now.
active_source is being tracked correctly in both directions now so UI is showing correct source info. 

3. Sameple rate conversion for squeezelite players - when MA's final output format is using a higher sample rate than the source, the audio plays slowed down.  Forcing MA to use 44100/16 by only having this rate enabled in the squeezelite player's settings and the audio plays at the correct speed.

4. Exception when doing UI -> Select source -> Plugin
```
2025-11-11 22:27:41.678 ERROR (MainThread) [music_assistant.webserver] Error handling message: CommandMessage(message_id='18bf852b-642b-41d5-8c26-0310259128e6', command='players/cmd/select_source', args={'player_id': '01:00:00:00:00:01', 'source': 'vban_receiver--7F6uGHkz'})
  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/players/player_controller.py", line 128, in wrapper
    await func(self, *args, **kwargs)
  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/players/player_controller.py", line 1004, in play_media
    await player.play_media(media)
  File "/opt/git/ma/server/vban-udpchange/music_assistant/providers/squeezelite/player.py", line 229, in play_media
    await self._handle_play_url_for_slimplayer(
    ...<5 lines>...
    )
  File "/opt/git/ma/server/vban-udpchange/music_assistant/providers/squeezelite/player.py", line 426, in _handle_play_url_for_slimplayer
    if queue and queue.repeat_mode == RepeatMode.ONE:
       ^^^^^
UnboundLocalError: cannot access local variable 'queue' where it is not associated with a value

  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/webserver.py", line 479, in _run_handler
    result = await result
             ^^^^^^^^^^^^
  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/players/player_controller.py", line 1032, in select_source
    await self._handle_select_plugin_source(player, cast("PluginProvider", plugin_prov))
  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/players/player_controller.py", line 2161, in _handle_select_plugin_source
    await self.play_media(
    ...<12 lines>...
    )
  File "/opt/git/ma/server/vban-udpchange/music_assistant/controllers/players/player_controller.py", line 130, in wrapper
    raise PlayerCommandFailed(str(err)) from err
music_assistant_models.errors.PlayerCommandFailed: cannot access local variable 'queue' where it is not associated with a value
```

sprocket-9:
Made sure the plugin only processes incoming packets when in_use_by is set, which takes care of the cpu use I noticed.

This is ready for review now.  The mypy linting errors I also got locally and I think it was due to an out of date models package? Or possibly a mypy cache thing.

OzGav:
You need to resolve the mypy errors.

sprocket-9:
Ack something isn't right here.  The force push I did yesterday doesn't have any of the changes I made locally.  Will have a look tomorrow.

sprocket-9:
That's looking better!  Adventures in git worktrees and mismatched local and remote branch names were my undoing.

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+            options=[ConfigValueOption(x, x) for x in _get_vban_sample_rates()],
+            label="PCM sample rate",
+            description="The VBAN PCM sample rate to expect from the remote VBAN sender. "
+            "This MUST match what the remote VBAN sender has set otherwise audio streaming "
+            "will not work.",
+            required=True,
+        ),
+        ConfigEntry(
+            key=CONF_AUDIO_CHANNELS,
+            type=ConfigEntryType.INTEGER,
```
OzGav, line 166: In the docs you gave the example of "mono or stereo". If they are the only two then consider changing this to a dropdown and map it

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+            options=[ConfigValueOption(x, x) for x in _get_supported_pcm_formats()],
+            label="PCM audio format",
+            description="The VBAN PCM audio format to expect from the remote VBAN sender. "
+            "This MUST match what the remote VBAN sender has set otherwise audio streaming "
+            "will not work.",
+            required=True,
+        ),
+        ConfigEntry(
+            key=CONF_PCM_SAMPLE_RATE,
+            type=ConfigEntryType.STRING,
```
OzGav: This could be an integer to protect against mistakes in input or even better a dropdown with the typical options of 44100. 48000, 88200, 96000, 176400, 192000.

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+            category="advanced",
+            required=True,
+        ),
+        ConfigEntry(
+            key=CONF_VBAN_QUEUE_SIZE,
+            type=ConfigEntryType.INTEGER,
+            default_value=AsyncVBANClientMod.default_queue_size,
+            label="Receiver: VBAN packets queue size",
+            description="Possibly increase if MA is running on a very low power device, "
+            "otherwise should not need changed.",
```
OzGav:
```suggestion
            "however, this should not normally need to be changed.",
```

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+            description="What should happen if the receiving queue fills up?",
+            category="advanced",
+            required=True,
+        ),
+        ConfigEntry(
+            key=CONF_VBAN_QUEUE_SIZE,
+            type=ConfigEntryType.INTEGER,
+            default_value=AsyncVBANClientMod.default_queue_size,
+            label="Receiver: VBAN packets queue size",
+            description="Possibly increase if MA is running on a very low power device, "
```
OzGav:
```suggestion
            description="This can be increased if MA is running on a very low power device, "
```

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+CONF_SENDER_HOST = "sender_host"
+CONF_PCM_AUDIO_FORMAT = "audio_format"
+CONF_PCM_SAMPLE_RATE = "sample_rate"
+CONF_AUDIO_CHANNELS = "audio_channels"
+CONF_VBAN_QUEUE_STRATEGY = "vban_queue_strategy"
+CONF_VBAN_QUEUE_SIZE = "vban_queue_size"
+
+VBAN_QUEUE_STRATEGIES = {
+    "Clear entire queue": BackPressureStrategy.DROP,
+    "Clear the oldest half of queue": BackPressureStrategy.DRAIN_OLDEST,
```
OzGav:
```suggestion
    "Clear the oldest half of the queue": BackPressureStrategy.DRAIN_OLDEST,
```

OzGav:
@sprocket-9 Here is the draft docs page. Let me know of any mistakes. https://github.com/music-assistant/beta.music-assistant.io/pull/50

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+            ),
+            metadata=StreamMetadata(
+                title=self._vban_stream_name,
+                artist=self._sender_host,
+            ),
+            stream_type=StreamType.CUSTOM,
+        )
+
+    @property
+    def supported_features(self) -> set[ProviderFeature]:
```
OzGav: You have passed the supported features to the parent class on line 215 so this is not required.

OzGav:
> When MA's final output format is using a higher sample rate than the source, the audio plays slowed down.
> Forcing MA to use 44100/16 by only having this rate enabled in the squeezelite player's settings and the audio plays at the correct speed.

This might be a problem for groups where the output format is fixed?

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: The Voicemeeter apps support 1-8 channels so I've made a dropdown for this range.

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: Converted to integer.  It's already a dropdown using the range of sample rates supported by aiovban library with a sensible default of 44100.

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: Fixed.

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: Fixed.

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: Fixed.

music_assistant/providers/vban_receiver/__init__.py
sprocket-9: I've removed this @property.

sprocket-9:
> This might be a problem for groups where the output format is fixed?

Hmm so I originally thought the bug was MA wasn't resampling the plugin source audio to match the higher output sample rate of squeezelite players, in which case the answer to your question would have been no, once the bug was fixed.

However, I think the bug has been in my plugin as I was passing a str instead of an int for AudioFormat.sample_rate .  Since changing CONF_PCM_SAMPLE_RATE to an integer, the new behaviour is MA doesn't resample, the output sample rate matches the plugin source sample rate.  So if this rate and the groups fixed output format differ, then it will be a problem.

Would be good to get some clarification on what the intended behaviour is though.

sprocket-9:
Tried a few more tests with different source and player sample rates.

If the squeezelite player supports the sample rate of the plugin source, MA will output at that rate, even if the player supports a higher rate. 

If the player doesn't support the sample rate of the plugin source, then MA chooses the highest sample rate the player supports for output, but it's not resampling the audio.

marcelveldt:
MA will only resample if needed. So if your plugin e.g.outputs at 22khz and the player supports that, it should not resample.
But according to your description there might be a bug somewhere I need to look into

marcelveldt:
This might be an issue with the squeezelite provider itself - let's track that in a dedicated issue report.
I propose to merge your PR now to get some beta testers on board and look at the resampling issue in parallel.

music_assistant/providers/vban_receiver/__init__.py
```diff
@@ -0,0 +1,324 @@
...
+
+class VBANReceiverProvider(PluginProvider):
+    """Implementation of a VBAN protocol receiver plugin."""
+
+    def __init__(
+        self, mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+    ) -> None:
+        """Initialize MusicProvider."""
+        super().__init__(mass, manifest, config, SUPPORTED_FEATURES)
+        self.logger = self.logger.getChild(self.instance_id.split("--")[1])
```
marcelveldt:
```suggestion
```
let's not override the default logger behavior

marcelveldt (approved):
Nice work, @sprocket-9 !

OzGav:
@sprocket-9 are you on our discord server? A user has reported a problem with the vban plugin

sprocket-9:
Yep problem is with the release of 0.8.0 of aiovban dependency which I'll need to look at.  Marvin has merged a pr to pin it to 0.6.3.

PR #2500, opened 2025-10-10
Subsonic: Add debug statement for resolving cover art

khers:
I am seeing failures in the request claiming there is no cover art path provided. I want to see this value when the failures occur.

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -176,6 +176,7 @@ class OpenSonicProvider(MusicProvider):
 
     async def resolve_image(self, path: str) -> bytes | Any:
         """Return the image."""
+        self.logger.debug("Requesting cover art for '%s'", path)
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Just wondering, do we really need to merge this or is this something you could run on a dev addon in order to resolve the issue? You can point the MA dev addon to a feature branch.

khers:
I'm sorry, I should have been more clear in the patch leader. The Open Subsonic provider is fairly unique in Music providers for MA in that there are at least 6 server implementations that follow the Open Subsonic spec and use this provider (there are far more server implementations). If I were diagnosing problems with the interactions between MA and Gonic (the implementation I use) then I would just run it locally. However, I have seen logs with failures in this method when working with other implementations (LMS, NextCloud Music, and Navidrome come to mind) and I would like to know what cover art ID we are sending that is causing the server to complain.

MarvinSchenkel (approved):

PR #2501, opened 2025-10-11
Add Dlna server Plugin

OzGav:
My first attempt at a plugin!

This creates a DLNA server which allows playback of local file system files.

In my testing I tried:
Android BubbleUPNP: Worked very well allowing queueing of all artist tracks, or all album tracks or individual tracks
VLC on WIndows: Allowed playback of individual tracks
nPlayer Lite on iOS: Played back individual tracks and albums.

Overall, the plugin seemed to be working and variations in functionality I assess are due to the respective clients. BubbleUPNP for example provides a rich experience with full artwork and full capabilities.

Streaming content is also served 

Hopefully satisfies https://github.com/orgs/music-assistant/discussions/388


music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1045 @@
...
+            # Serve the file
+            return cast(
+                "web.Response", web.FileResponse(path=file_path, headers={"Accept-Ranges": "bytes"})
+            )
+
+        except MediaNotFoundError:
+            return web.Response(status=404, text="Track not found")
+        except Exception as err:
+            self.logger.exception("Error streaming track")
+            return web.Response(status=500, text=str(err))
```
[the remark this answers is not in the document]
OzGav: Fixed

music_assistant/providers/dlna_server/__init__.py
```diff
@@ -0,0 +1,1045 @@
...
+                    <errorCode>{error_code}</errorCode>
+                    <errorDescription>{error_description}</errorDescription>
+                </UPnPError>
+            </detail>
+        </s:Fault>
+    </s:Body>
+</s:Envelope>"""
+
+        return web.Response(
+            text=error_xml,
```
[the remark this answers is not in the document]
OzGav: Fixed

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1044 @@
...
+            self.logger.info(
+                "DLNA Server started successfully - "
+                "Music Assistant is now discoverable as a DLNA Media Server"
+            )
+        except OSError as err:  # Socket/network errors
+            self.logger.exception("Failed to start DLNA server due to network error")
+            raise SetupFailedError(f"Failed to start DLNA server: {err}") from err
+        except Exception as err:  # Unexpected errors
+            self.logger.exception("Failed to start DLNA server")
+            raise SetupFailedError("Failed to start DLNA server") from err
```
marcelveldt, line 113:
```suggestion
```

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1044 @@
...
+
+    async def unload(self, is_removed: bool = False) -> None:
+        """Handle unload/close of the provider."""
+        # Stop SSDP server
+        if self._ssdp_server:
+            try:
+                await self._ssdp_server.stop()
+            except Exception as err:
```
marcelveldt: can this exception be narrowed down ?

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1044 @@
...
+        if self._routes_registered:
+            try:
+                self.mass.streams.unregister_dynamic_route("/dlna/description.xml")
+                self.mass.streams.unregister_dynamic_route("/dlna/ContentDirectory.xml")
+                self.mass.streams.unregister_dynamic_route("/dlna/ConnectionManager.xml")
+                self.mass.streams.unregister_dynamic_route("/dlna/control/ContentDirectory")
+                self.mass.streams.unregister_dynamic_route("/dlna/control/ConnectionManager")
+                self.mass.streams.unregister_dynamic_route("/dlna/track/*")
+                # self.mass.streams.unregister_dynamic_route("/dlna/event/ContentDirectory")
+            except Exception as err:
```
marcelveldt: remove the try...catch

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1044 @@
...
+    async def unload(self, is_removed: bool = False) -> None:
+        """Handle unload/close of the provider."""
+        # Stop SSDP server
+        if self._ssdp_server:
+            try:
+                await self._ssdp_server.stop()
+            except Exception as err:
+                self.logger.warning("Error stopping SSDP server: %s", err)
+
+        # Unregister HTTP routes
```
marcelveldt: do this above stopping the server

MarvinSchenkel:
Just had a little play with this. Nice job!

A few things I noticed:
- Works really well playing local/smb file sources.
- Also works in the dev addon via HA ingress!

Now, it also seems to index media from **streaming** providers. When opening the 'Artist' folder, it seems to call `get_streamdetails()` for every track in there, causes a long delay for YTM for example, since resolving streamurls takes 1-2sec per Track. Playing any of the Tracks coming from a streaming provider results in an error.

I think the goal here is to only serve the content from the local/smb file providers? If so, can we configure the plugin to only advertise media items that are provided by local/smb?

music_assistant/providers/dlna-server/__init__.py
OzGav: Fixed

music_assistant/providers/dlna-server/__init__.py
OzGav: fixed

OzGav:
@MarvinSchenkel I think I have to do something. What do you think of only scanning at the track level so you would see the artist then album but inside would be empty if there were no local tracks? Otherwise I think it could be very slow if the provider had to scan every track of an artist just in case one had a local mapping?

MarvinSchenkel:
> @MarvinSchenkel I think I have to do something. What do you think of only scanning at the track level so you would see the artist then album but inside would be empty if there were no local tracks? Otherwise I think it could be very slow if the provider had to scan every track of an artist just in case one had a local mapping?

The library functions should have a `provider` parameter that allows you to filter on provider. Something like this should work:

Add a constant on the top:
`SUPPORTED_PROVIDERS = ("filesystem", "filesystem_smb")`

Then inside `_get_children()`, this should work for artists
```
artists = []
for provider in SUPPORTED_PROVIDERS:
  artists.extend(await self.mass.music.artists.library_items(
      limit=limit, offset=offset, order_by="sort_name",  provider=provider
  ))
artist_items = [self._create_artist_container(artist) for artist in artists]
total = len(artists)
```

Of course we will also need this for albums, tracks, album_tracks...

OzGav:
@MarvinSchenkel Yes I was off on a tangent for a second as I'm doing a million things here. I havent had a chance to test it but I pushed a fix (hopefully!)

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1094 @@
...
+            # Serve the file
+            return cast(
+                "web.Response", web.FileResponse(path=file_path, headers={"Accept-Ranges": "bytes"})
+            )
+
+        except MediaNotFoundError:
+            return web.Response(status=404, text="Track not found")
+        except Exception:
+            self.logger.exception("Error streaming track")
+            return web.Response(status=500, text="Internal server error")
```
marcelveldt: we have a streams controller to handle this. This may not be handled this way directly from a provider

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1094 @@
...
+            album_id = parent_id[6:]
+            tracks = await self._get_filesystem_tracks_for_album(album_id)
+            paginated_tracks = (
+                list(tracks)[offset : offset + limit] if limit > 0 else list(tracks)[offset:]
+            )
+            track_items = [await self._create_track_item(track) for track in paginated_tracks]
+            didl_xml = self._wrap_didl_items(track_items)
+            return didl_xml, len(track_items), len(tracks)
+
+        return self._create_empty_didl(), 0, 0
```
marcelveldt, line 786: why are we not traversing the browse endpoint here or the libraries ?
This is making some very strange assumptions for local filesystems etc.

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1094 @@
...
+            albums, total = await self._get_filesystem_albums(limit, offset)
+            album_items = [
+                self._create_album_container(album, ALBUMS_CONTAINER_ID) for album in albums
+            ]
+            didl_xml = self._wrap_didl_items(album_items)
+            return didl_xml, len(album_items), total
+
+        if parent_id == TRACKS_CONTAINER_ID:
+            tracks, total = await self._get_filesystem_tracks(limit, offset)
+            track_items = [await self._create_track_item(track) for track in tracks]
```
marcelveldt: this is going to be horrible slow and its completely unneeded to do this this way

music_assistant/providers/dlna-server/__init__.py
```diff
@@ -0,0 +1,1094 @@
...
+        prov = self.mass.get_provider(provider_instance)
+
+        # Default values
+        file_ext = "mp3"
+        mime_type = "audio/mpeg"
+        file_size = 0
+
+        if prov and isinstance(prov, MusicProvider):
+            try:
+                streamdetails = await prov.get_stream_details(prov_item_id, MediaType.TRACK)
```
marcelveldt: You are fetching the streamdetails for each track in a listing!
That is going to cause MASSIVE issues. This should really be dropped

marcelveldt:
This PR has major architectural challenges for which I think we currently do not have the capacity at all. We have about 80 open user issues reports which all have a higher priority than adding this niche feature to our codebase.

Marked as draft and we will revisit this when capacity is in a better place.
For now I think the vibe coding energy can better be spent at solving small user requests and bugfixes.

music_assistant/providers/dlna-server/__init__.py
OzGav: I think this has been addressed now. The streaming code no longer accesses files directly - instead it uses get_media_stream and get_ffmpeg_stream from the audio helpers which I believe goes through the proper streams controller pipeline. Let me know if I've still got this wrong!

music_assistant/providers/dlna-server/__init__.py
OzGav: The old filesystem-specific methods are gone and it now uses the library APIs instead - things like `iter_library_items()` for artists/albums/tracks and `self.mass.music.artists.albums()` for getting an artist's albums. Appears to work with any provider now not just local files.

music_assistant/providers/dlna_server/__init__.py
```diff
@@ -0,0 +1,1165 @@
...
+        await self._ssdp_server.start()
+        self.logger.info("SSDP advertisement started - devices can now discover this server")
+
+    # ==================== HTTP Handlers ====================
+
+    async def _handle_device_description(self, request: web.Request) -> web.Response:
+        """Handle device description request."""
+        base_url = self.mass.streams.base_url
+
+        device_xml = f"""<?xml version="1.0"?>
```
marcelveldt, line 211: should this come from a helper ?

music_assistant/providers/dlna_server/__init__.py
```diff
@@ -0,0 +1,1165 @@
...
+    </s:Body>
+</s:Envelope>"""
+
+        return web.Response(
+            text=response_xml,
+            content_type="text/xml",
+            charset="utf-8",
+        )
+
+    async def _handle_track_stream(self, request: web.Request) -> web.StreamResponse:
```
marcelveldt, line 578: I don't like this because it opens the door for piracy.
we need at least a concept of sessions and attaching these requests to a player(queue) and limit concurrent streams / detect rapid downloads etc.

Also its duplicating logic that we already have in the streams controller.

music_assistant/providers/dlna_server/ssdp.py
```diff
@@ -0,0 +1,358 @@
...
+SUPPORTED_SEARCH_TARGETS = {
+    "ssdp:all",
+    "upnp:rootdevice",
+    "urn:schemas-upnp-org:device:MediaServer:1",
+    "urn:schemas-upnp-org:service:ContentDirectory:1",
+    "urn:schemas-upnp-org:service:ConnectionManager:1",
+}
+
+
+class SSDPServer:
```
marcelveldt, line 29: look what you can unify from the new discovery controller instead of reinventing the wheel

marcelveldt:
DLNA is an outdated, very insecure protocol. I'm not so sure if we really want to add this support to our codebase.
Do we have some numbers about user demand ?

OzGav:
I agree with you and the user experience varies considerably by the client software. However, this was actually your feature request 😀 Currently, it is at number 14 of the unimplemented requests with 26 votes https://github.com/orgs/music-assistant/discussions?discussions_q=is%3Aopen+sort%3Atop+-label%3Aimplemented

marcelveldt:
> I agree with you and the user experience varies considerably by the client software. However, this was actually your feature request 😀 Currently, it is at number 14 of the unimplemented requests with 26 votes https://github.com/orgs/music-assistant/discussions?discussions_q=is%3Aopen+sort%3Atop+-label%3Aimplemented

Yeah, I don't think 26 votes satisfies introducing a security risk. Let's focus on high demand requests. 
Everything we add also needs maintenance long term. 

This PR can be accepted if we find a way to fix the security risks and the open door for piracy - otherwise I'm afraid it has been a waste of your time, sorry!

MarvinSchenkel:
As discussed, let's close this since we wont be implementing it in the foreseeable future

PR #2503, opened 2025-10-11
Add Pandora provider

OzGav:
This PR introduces a new Pandora Music Provider focused exclusively on Radio Station playback.

Key Features
Radio Integration: Allows users to access and play their saved Pandora radio stations within Music Assistant.

Authentication: Uses the web-based Pandora REST API (username/password login) to retrieve the user's station list.

Streaming: Streams are served by fetching multiple playlist fragments from Pandora, assembling a multi-part path for continuous playback. 

🛑 Scope Limitations (API Constraint)
This provider is limited to Radio Stations Only and does not support search or station creation.This constraint is due to Pandora's API structure:

The REST API used here only provides access to login, station lists, and playback fragments.

Features like Search and Station Creation are restricted to Pandora's separate GraphQL API, which requires a completely different OAuth authentication flow and developer credentials.

Thus users must continue to create stations on the official Pandora app/website, and those stations will then automatically appear in MA.

⏱️ Streaming Time Limit
The provider build a dummy list of 1000 URLs which should ensure playback lasts for more than 2 days.


TermeHansen:
Interesting work.

I have used pianobar for many years and it works very well. Have you looked to that code for how to use the api?

https://github.com/PromyLOPh/pianobar

OzGav:
Thanks for the suggestion! I’ve reviewed pianobar’s implementation, but there’s a key concern: pianobar uses extracted partner keys from official Pandora clients and encryption meant for specific devices. Pandora has explicitly stated that pianobar “is an unauthorized third-party application, and using it is a violation of our Terms of Use.” While they’ve tolerated it for years, it’s technically unauthorized.
This provider uses Pandora’s modern REST API (the same endpoints their website uses) with standard username/password authentication—no extracted keys or reverse-engineered encryption. The tradeoff is we can’t implement search/station creation yet, so users need to manage stations via Pandora’s official apps for now.

music_assistant/helpers/audio.py
```diff
@@ -379,10 +379,13 @@ async def get_stream_details(
             streamdetails.stream_type in (StreamType.ICY, StreamType.HLS, StreamType.HTTP)
             and streamdetails.media_type == MediaType.RADIO
         ):
-            assert isinstance(streamdetails.path, str)  # for type checking
-            resolved_url, stream_type = await resolve_radio_stream(mass, streamdetails.path)
-            streamdetails.path = resolved_url
-            streamdetails.stream_type = stream_type
+            # Only resolve if path is a string (traditional radio URL)
+            # Skip resolution for multi-file radio streams (list of URLs)
+            if isinstance(streamdetails.path, str):
```
MarvinSchenkel: What is the exact use of this `if` statement? The old code already has an `assert` in there, basically ensuring that the `streamdetails.path` is a valid `str`?

music_assistant/providers/pandora/constants.py
```diff
@@ -0,0 +1,47 @@
+"""Constants for the Pandora provider."""
+
+from __future__ import annotations
+
+# Configuration Keys
+CONF_USERNAME = "username"
```
MarvinSchenkel: We already have constants for username and password over [here](https://github.com/music-assistant/server/blob/03d575bf0fc5d2502aa792a05c573500bb0620e3/music_assistant/constants.py#L50)

OzGav: rookie mistake. Fixed

music_assistant/helpers/audio.py
OzGav: The change adds a type check to only resolve single-URL radio streams, skipping multi-file radio streams like in this provider.  Without this check, MA would try to call resolve_radio_stream() on the list of URLs (MultiPartPath[]), which would fail because resolve_radio_stream() expects a single string URL.

MarvinSchenkel: In that case, just add it to the `if` statement above:

```
if (
    streamdetails.stream_type in (StreamType.ICY, StreamType.HLS, StreamType.HTTP)
    and streamdetails.media_type == MediaType.RADIO
    and isinstance(streamdetails.path, str)
):
  ...
```
Then we can also get rid of the `assert isinstance(streamdetails.path, str)  # for type checking`

OzGav: fixed

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,400 @@
...
+                if len(results) >= limit:
+                    break
+
+        return SearchResults(radio=results)
+
+    async def browse(self, path: str) -> list[Radio]:
+        """Browse radio stations."""
+        # For now, just return all stations like get_library_radios
+        # Could be enhanced with categories/genres in the future
+        return [station async for station in self.get_library_radios()]
```
marcelveldt: this can be left out if you dont support this yet. So the default implementation can be used.

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,400 @@
...
+        if media_type != MediaType.RADIO:
+            raise MediaNotFoundError(f"Unsupported media type: {media_type}")
+
+        # Create 1000 dummy tracks pointing to our dynamic handler
+        # This provides roughly two days of radio streaming assuming 3 mins per track
+        parts = [
+            MultiPartPath(
+                path=f"{self.mass.streams.base_url}/{self.instance_id}_stream?"
+                f"station_id={item_id}&track_num={i}",
+                duration=180,  # Placeholder duration
```
marcelveldt: duration can be left out, its an optional argument

marcelveldt:
```suggestion
```

OzGav: Thanks!

OzGav: Fixed below

music_assistant/providers/pandora/provider.py
OzGav: I tried removing that but then there is nothing seen when browsing?

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,421 @@
...
+                f"station_id={item_id}&track_num={i}",
+            )
+            for i in range(1000)
+        ]
+        # Is this what is envisaged to pass the static variables?
+        #    MultiPartPath(
+        #        path=f"{self.mass.streams.base_url}/{self.instance_id}_stream?"
+        #        f"station_id={item_id}&track_num={i}&queue_id={queue_id}"
+        #        f"&queue_item_id={queue_item_id}",
+        #    )
```
marcelveldt: yeah, here you can pass whatever you want

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,421 @@
...
+                return web.Response(status=404, text="No audio URL available")
+
+            # Update metadata if we have queue context
+            if queue_id and queue_item_id:
+                queue_item = self.mass.player_queues.get_item(queue_id, queue_item_id)
+                if queue_item and queue_item.streamdetails:
+                    queue_item.streamdetails.stream_metadata = StreamMetadata(
+                        title=track.get("songTitle", "Unknown Song"),
+                        artist=track.get("artistName", "Unknown Artist"),
+                    )
```
marcelveldt: note that I have added some more fields to the StreamMetadata so if you have more info, you can add it if you want.

music_assistant/providers/pandora/provider.py
marcelveldt: yeah, it will only work after the first sync was completed.

OzGav: After the sync I see the stations in the radio view but there is nothing when I browse to Pandora? This is after I see this in the log
2025-10-28 11:40:25.015 INFO (MainThread) [music_assistant.music] Sync task for Pandora/radios completed

music_assistant/providers/pandora/provider.py
OzGav: Thanks I have done that.

music_assistant/providers/pandora/provider.py
OzGav: Cool, so I understand where we are at is these two questions:
How to pass queue_id/queue_item_id properly?
How to sync metadata updates with actual playback timing?

marcelveldt: yeah, just pass queue_id and queue_item_id to that path.
then in the function where you resolve the dynamic url, get the queue_item from the queue controller.
then you can set the stream_metadata of the queueitem.streamdetails

This way no polling logic is needed, you just set the metadata as soon as ffmpeg requests the next url, which is about 2 seconds in advance.

music_assistant/providers/pandora/provider.py
marcelveldt: but you did implement the get_library_radios ?

OzGav: Yes on line 196?

music_assistant/providers/pandora/provider.py
OzGav: OK this is somewhat working. However, the next track is getting fetched 75secs (roughly) before the end of the current, not 2 secs :-( 

I also assume I need to change the definition of get_stream_details in music_provider.py and all providers that this to
```
    async def get_stream_details(
        self,
        item_id: str,
        media_type: MediaType,
        queue_id: str | None = None,
        queue_item_id: str | None = None,
    ) -> StreamDetails:
```

marcelveldt: is the mediatype set to radio ?
what player type is this ?

OzGav: Yes MediaType is RADIO. See line 289.

I think the first test might have been with THIS DEVICE but now I am using Sonos and the metadata is changing about 45secs before the track actually starts.

I also have some odd behaviour in the frontend but I don't want to muddy the waters with that right now...

marcelveldt: that is weird and should not happen. a radio stream only has a 2 seconds buffer and not 45 or worse.
So I wonder where that 45s is coming from. Did you rebase the PR to contain all changes we recently made to the streams controller ?

marcelveldt: So, the Sonos should fetch the "serve_queue_item_stream" from streams.py where the logic there decides its a radio mediaitem and so it will fetch the audio data from "get_queue_item_stream" and apply the readrate limiter.

OzGav: I am not sure where to look at the moment. Here is a timeline from a test I just ran:
Track 0 started playing at 10:35:09

Track 1 (Curator Message):

URL requested: 10:38:47 (timestamp 1762821527)
Metadata updated: 10:38:49 (timestamp 1762821529)
Track 1 actually played ~30 seconds later

Track 2 (Sounds of Silence):

URL requested: 10:39:19 (timestamp 1762821559)
Metadata updated: 10:39:21 (timestamp 1762821561)
Track actually started playing: 10:40:06
45 seconds early! 

Could the Sonos player be buffering?

marcelveldt: Yeah, this matches the Sonos buffer indeed. But that should not be possible if you mark the item as Radio (as we dont allow too much buffer for radio streams)

music_assistant/helpers/audio.py
```diff
@@ -283,7 +283,10 @@ async def get_stream_details(
             try:
                 BYPASS_THROTTLER.set(True)
                 streamdetails = await music_prov.get_stream_details(
-                    prov_media.item_id, media_item.media_type
+                    prov_media.item_id,
+                    media_item.media_type,
+                    queue_id=queue_item.queue_id,
+                    queue_item_id=queue_item.queue_item_id,
```
marcelveldt: please remove this leftover code, its touching a core controller in a provider PR

music_assistant/helpers/audio.py
```diff
@@ -299,8 +302,8 @@ async def get_stream_details(
         if (
             streamdetails.stream_type in (StreamType.ICY, StreamType.HLS, StreamType.HTTP)
             and streamdetails.media_type == MediaType.RADIO
+            and isinstance(streamdetails.path, str)
```
marcelveldt, line 320: please remove this leftover code, its touching a core controller in a provider PR

music_assistant/models/music_provider.py
```diff
@@ -403,7 +403,13 @@ async def get_resume_position(self, item_id: str, media_type: MediaType) -> tupl
...
         raise NotImplementedError
 
-    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+    async def get_stream_details(
+        self,
+        item_id: str,
+        media_type: MediaType,
+        queue_id: str | None = None,
+        queue_item_id: str | None = None,
+    ) -> StreamDetails:
```
marcelveldt: please revert this code, its touching a core controller in a provider PR

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,454 @@
...
+                )
+            },
+        )
+
+    async def get_stream_details(
+        self,
+        item_id: str,
+        media_type: MediaType,
+        queue_id: str | None = None,
+        queue_item_id: str | None = None,
```
marcelveldt: I don't like this change - let's revert it.

music_assistant/providers/pandora/provider.py
marcelveldt: still you need to drop this. it should not be needed

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,454 @@
...
+                if len(results) >= limit:
+                    break
+
+        return SearchResults(radio=results)
+
+    async def browse(self, path: str) -> list[Radio]:
+        """Browse radio stations."""
+        # For now, just return all stations like get_library_radios
+        # Could be enhanced with categories/genres in the future
+        return [station async for station in self.get_library_radios()]
```
marcelveldt:
```suggestion
```

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,454 @@
...
+        # This provides roughly two days of radio streaming assuming 3 mins per track
+        parts = [
+            MultiPartPath(
+                path=f"{self.mass.streams.base_url}/{self.instance_id}_stream?"
+                f"station_id={item_id}&track_num={i}&queue_id={queue_id}"
+                f"&queue_item_id={queue_item_id}",
+            )
+            for i in range(1000)
+        ]
+        return StreamDetails(
```
marcelveldt, line 248: So thinking about this more and more I think we need a more structural change for this. 
My proposal would be to add a update handler to the streamdetails itself to update the metadata.

So an optional callback on streamdetails would be by proposal.

music_assistant/models/music_provider.py
OzGav: I was just trying to get what I thought your proposed fix was working since it was in draft. Removed.

music_assistant/helpers/audio.py
OzGav: Removed

music_assistant/providers/pandora/provider.py
OzGav: Removed

music_assistant/providers/pandora/provider.py
OzGav: I am hearing you and I don't know why but if I remove that and then remove and re-add the provider then the browse folder is empty. I'm sure I'm doing something wrong but I don't know what!

Image

and then with that function put back

Image

music_assistant/helpers/audio.py
OzGav: On this one this is actually required to make the provider work because the current code assumes all HTTP radio streams are single URLs, but this provider uses multi-part paths for sequential track streaming.

OzGav: Or should I just put it through a quick dedicated PR?

music_assistant/providers/pandora/provider.py
OzGav: This sounds like a good idea. Having exact timing for the metadata update would be ideal.

OzGav: OK dynamic metadata code is commented out, to be revisited in the future.

music_assistant/providers/pandora/provider.py
OzGav: Dyanmic metadata to be revisited in the future

music_assistant/helpers/audio.py
marcelveldt: OK, it's fine for this time

music_assistant/providers/pandora/provider.py
marcelveldt: It will only show something when the sync was completed ?

OzGav: If I freshly add the provider and then do a sync nothing is shown in the browse view

```
2025-11-27 09:58:33.163 DEBUG (MainThread) [music_assistant.webserver] Handling command config/providers/save
2025-11-27 09:58:33.167 DEBUG (MainThread) [music_assistant] Loading provider pandora
2025-11-27 09:58:33.168 DEBUG (MainThread) [music_assistant.pandora] Log level configured to GLOBAL
2025-11-27 09:58:34.540 INFO (MainThread) [music_assistant.pandora] Successfully authenticated with Pandora
2025-11-27 09:58:34.541 INFO (MainThread) [music_assistant] Loaded music provider Pandora
2025-11-27 09:58:34.598 DEBUG (MainThread) [music_assistant.webserver] Handling command config/providers
2025-11-27 09:58:34.603 DEBUG (MainThread) [music_assistant.webserver] Handling command config/providers
2025-11-27 09:58:39.598 DEBUG (MainThread) [music_assistant.controllers.config] Saved data to persistent storage
2025-11-27 09:58:44.550 DEBUG (MainThread) [music_assistant.pandora] Start sync of Radios to Music Assistant library.
2025-11-27 09:58:44.550 DEBUG (MainThread) [music_assistant.pandora] Fetching Pandora stations
2025-11-27 09:58:44.759 DEBUG (MainThread) [music_assistant.pandora] Found 3 Pandora stations
2025-11-27 09:58:44.768 DEBUG (MainThread) [music_assistant.music.radio] added Dire Straits Radio to database (id: 34)
2025-11-27 09:58:44.779 DEBUG (MainThread) [music_assistant.music.radio] added Elton John Radio to database (id: 35)
2025-11-27 09:58:44.787 DEBUG (MainThread) [music_assistant.music.radio] added Fleetwood Mac Radio to database (id: 36)
2025-11-27 09:58:44.810 INFO (MainThread) [music_assistant.music] Sync task for Pandora/radios completed
2025-11-27 09:58:44.813 DEBUG (MainThread) [music_assistant.music] Performing database cleanup...
2025-11-27 09:58:44.856 DEBUG (MainThread) [music_assistant.music] Database cleanup done
2025-11-27 09:58:49.936 DEBUG (MainThread) [music_assistant.webserver] Handling command music/browse
2025-11-27 09:58:53.049 DEBUG (MainThread) [music_assistant.webserver] Handling command music/browse
2025-11-27 09:58:56.443 DEBUG (MainThread) [aiohttp.server] Ignored premature client disconnection.
2025-11-27 09:58:56.975 DEBUG (MainThread) [music_assistant.webserver] Handling command players/all
2025-11-27 09:58:56.981 DEBUG (MainThread) [music_assistant.webserver] Handling command music/artists/count
2025-11-27 09:59:01.074 DEBUG (MainThread) [music_assistant.webserver] Handling command music/browse
```

image

I dont know if it is something with my dev environment. Let me try in the dev server

OzGav: TODO comment added to remove the browse method when the core bug is fixed

music_assistant/providers/pandora/provider.py
OzGav: TODO added to remove this when core bug fixed

music_assistant/providers/pandora/provider.py
marcelveldt: this is now added to the core codebase, see my comments on the BBC provider how it works (or see the already opened PR for BBC provider which implements it already)

music_assistant/providers/pandora/provider.py
marcelveldt: I'd rather not have in the code and fix the core issue immediately

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,432 @@
...
+    mass: MusicAssistant,  # noqa: ARG001
+    instance_id: str | None = None,  # noqa: ARG001
+    action: str | None = None,  # noqa: ARG001
+    values: dict[str, ConfigValueType] | None = None,  # noqa: ARG001
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    return (
+        ConfigEntry(
+            key=CONF_USERNAME,
+            type=ConfigEntryType.SECURE_STRING,
```
marcelveldt: note that by using a secure string, it will also no longer be visible in the frontend. Is that accepted ?

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,432 @@
...
+            stream_type=StreamType.HTTP,
+            path=parts,
+            can_seek=False,
+            allow_seek=False,
+            stream_metadata=StreamMetadata(
+                title="Pandora",
+            ),
+        )
+
+    # @use_cache(300)  # Cache fragments for 5 minutes
```
marcelveldt: leftover comment?

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,432 @@
...
+            #                           (art["url"] for art in album_art if art.get("size") == 500),
+            #                           album_art[-1]["url"] if album_art else None,
+            #                        )
+            #                    queue_item.streamdetails.stream_metadata = StreamMetadata(
+            #                        title=track.get("songTitle", "Unknown Song"),
+            #                        artist=track.get("artistName", "Unknown Artist"),
+            #                        album=track.get("albumTitle"),
+            #                        image_url=album_art_url,
+            #                        duration=track.get("trackLength"),
+            #                    )
```
marcelveldt:
```suggestion
```

music_assistant/providers/pandora/provider.py
```diff
@@ -0,0 +1,432 @@
...
+                    break
+
+        return SearchResults(radio=results)
+
+    # TODO: Remove this when default implementation is fixed
+    async def browse(self, path: str) -> list[Radio]:
+        """Browse radio stations."""
+        # For now, just return all stations like get_library_radios
+        # Could be enhanced with categories/genres in the future
+        return [station async for station in self.get_library_radios()]
```
marcelveldt:
```suggestion
```

marcelveldt: lets fix core instead!

music_assistant/providers/pandora/provider.py
OzGav: Fixed

music_assistant/providers/pandora/provider.py
OzGav: I'm not sure yet what needs to change. The only thing I have noticed so far from looking is this line

https://github.com/music-assistant/server/blob/a478337872908123863b6f6b370972451dac0177/music_assistant/models/music_provider.py#L563

should it be `provider=self.instance_id,` like all the others? Not a factor here but I wanted to mention it before I forgot. Beyond that it isn't clear to me why it is failing to render the list?

music_assistant/providers/pandora/provider.py
OzGav: Fixed

music_assistant/providers/pandora/provider.py
OzGav: Not necessary. Changed to string.

music_assistant/providers/pandora/provider.py
marcelveldt: yeah that should be instance_id for sure

marcelveldt: But what i meant is that we need to fix core - not perse in this PR.
Just remove the workaround so we can merge it and then focus on the issue in core

OzGav: Function removed

OzGav:
OK this is ready again. I had to implement the skipping of the curator messages as the API response for their durations was grossly wrong (as in thousands of seconds instead of 15-30) and this meant it was impossible to time the metadata update with the track playback. Seems to be working well locally within the constraint of just playing the "radio stations" in the users "Collections"

marcelveldt (approved):
Looks good, @OzGav !

PR #2506, opened 2025-10-13
Fix random chapter skipping

tsipinakis:
In case a chapter further than the starting chapter has a duration which makes skipped_duration + duration < seek_positon that chapter would be silently skipped. Lets go back to using a separate function to find the position to fix that.

Additionally, bring back the guard to prevent landing in the final couple of seconds of a track for more smoother chapter seek

Reopening of #2495 , I managed to mess up the rebase which confused github.

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -944,6 +944,48 @@ async def get_file_stream(
             yield data
 
 
+def _get_parts_from_position(
+    parts: list[MultiPartPath], seek_position: int
+) -> tuple[list[MultiPartPath], int]:
+    """Get the remaining parts list from a timestamp.
+
+    Arguments:
+    parts: The list of  parts
+    seek_position: The seeking position in seconds of the tracklist
+
+    Returns:
+        In a tuple, A list of  parts, starting with the one at the requested
+        seek position and the position in seconds to seek to in the first
+        track.
+    """
+    skipped_duration = 0.0
+    for i, part in enumerate(parts):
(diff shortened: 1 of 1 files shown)


music_assistant/helpers/audio.py
```diff
@@ -944,6 +944,48 @@ async def get_file_stream(
             yield data
 
 
+def _get_part_from_position(
```
marcelveldt:
```suggestion
def _get_parts_from_position(
```

music_assistant/helpers/audio.py
```diff
@@ -944,6 +944,48 @@ async def get_file_stream(
...
+        In a tuple, A list of  parts, starting with the one at the requested
+        seek position and the position in seconds to seek to in the first
+        track.
+    """
+    skipped_duration = 0.0
+    for i, part in enumerate(parts):
+        if not isinstance(part, MultiPartPath):
+            raise InvalidDataError("Multi-file streamdetails requires a list of MultiPartPath")
+        if part.duration is None:
+            raise InvalidDataError("Track has None duration, cannot find position")
```
marcelveldt: duration is optional. In case its absent (in any of the parts) we need to return the full partslist and let ffmpeg deal with seeking in the files. That will be suboptimal ofc but better than nothing. In the end we aim that providers should simply provide this info in the partslist

marcelveldt:
```suggestion
            return parts, seek_position
```

marcelveldt (approved):
Thanks, @tsipinakis !

PR #2507, opened 2025-10-13
Simplify library sync/import settings

marcelveldt:
As discussed on discord, a simplification of the library/sync settings.

- Whether to enable sync the sync of a certain mediatype into the library is now just a boolean toggle.
- Items will be imported without the favorite flag, unless the provider actually supports a dedicated favorite flag.
- Sync back (2-way sync) is now also just a simple toggle.
- Provider features have been added to allow providers to sync back the favorite status if they really support that.


khers:
I think this is what I described I wanted but I'd like to restate what I think the code is doing. A per provider config value is added to control if items are always imported as favorites or if that is conditional. If that is correct then this looks good to me.

marcelveldt:
> I think this is what I described I wanted but I'd like to restate what I think the code is doing. A per provider config value is added to control if items are always imported as favorites or if that is conditional. If that is correct then this looks good to me.

No, the favorite importing is controlled by the provider itself. The toggle that is added in config is only about syncing the library items to MA

MarvinSchenkel:
For example, YT Music and Apple Music make a distinction between 'in library' and 'favourite'. So I can have an album in my library, but only 'favourited' two songs. With this PR, MA will sync the entire album but only flag those 2 tracks as favourites, as per the provider.

Great addition @marcelveldt 🎉

MarvinSchenkel (approved):

PR #2508, opened 2025-10-15
Mark Jellyfin as unmaintained

OzGav:


marcelveldt (approved):
Unfortunate but it is what it is. At least this sets te right expectations if people run into issues.

PR #2279, opened 2025-07-10
Add plex connect provider

chicco-carone:
This pull request introduces enhancements to the Plex integration in Music Assistant, focusing on improving stream data handling and adding a new plugin provider for reporting playback timeline events. The most important changes include restructuring stream data, updating playback tracking logic, and introducing the `PlexConnect` plugin provider.

### Enhancements to Plex Stream Data Handling:
* Refactored the `get_stream_details` method to use a structured dictionary (`plex_data`) for stream metadata, replacing the direct use of `plex_track` in the `data` field of `StreamDetails`. This improves clarity and maintainability and is used by the plex connect plugin to report back to the server the data.
* Updated the `on_streamed` method to use the new `plex_data` structure for marking tracks as played, replacing direct attribute access with dictionary keys.

### Introduction of the `PlexConnect` Plugin Provider:
* Added a new `PlexConnect` plugin provider to monitor and report Plex playback timeline events. This plugin adds support to synchronize track playback back to the plex server.
* Created a corresponding `manifest.json` file for the `PlexConnect` plugin, specifying its metadata, requirements, and documentation link.

This pr contains the remain of the code from #2228

PR #2512, opened 2025-10-17
Plex: Add real-time playback state reporting and timeline updates

anatosun:
- Add instant timeline updates when playback state changes
- Add periodic 5-second position updates during active playback
- Improve scrobbling with better error handling and logging
- Add client identification headers for Music Assistant
- Subscribe to MEDIA_ITEM_PLAYED events for real-time sync

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -1209,7 +1209,7 @@ class PlexProvider(MusicProvider):
 
                     # Parse URI to check if it's a Plex track
                     try:
-                        media_type, provider_instance_or_domain, item_id = await parse_uri(
+                        media_type, provider_instance_or_domain, item_id = parse_uri(
                             current_item.uri
                         )
                     except Exception as err:
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -1173,7 +1173,7 @@ class PlexProvider(MusicProvider):
                         "identifier": "com.plexapp.plugins.library",
                     }
                     self.logger.debug("Unscrobbling track in Plex: %s", unscrobble_params)
-                    self._plex_server.query("/:/unscrobble", unscrobble_params)
+                    self._plex_server.query("/:/unscrobble", params=unscrobble_params)
                     self.logger.info("Track %s marked as unplayed in Plex", prov_item_id)
 
             except Exception as err:
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -22,7 +22,6 @@ from music_assistant_models.enums import (
     ContentType,
     ImageType,
     MediaType,
-    PlaybackState,
     ProviderFeature,
     StreamType,
 )
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -1122,7 +1122,7 @@ class PlexProvider(MusicProvider):
                         "identifier": "com.plexapp.plugins.library",
                     }
                     self.logger.debug("Scrobbling track to Plex: %s", scrobble_params)
-                    self._plex_server.query("/:/scrobble", scrobble_params)
+                    self._plex_server.query("/:/scrobble", params=scrobble_params)
                     self.logger.info("Track %s marked as played in Plex", prov_item_id)
 
                 # If position is 0 and not playing, mark as unplayed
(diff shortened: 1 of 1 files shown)


marcelveldt:
This appears to be a duplicate of the already open (stale) PR here:
https://github.com/music-assistant/server/pull/2279

That approach separated the scrobbling from the music provider.
If we think the scrobbling is what more or less every plex user would expect, then I would indeed keep it in the existing plex music provider. If its more of a niche feature it would be better to have it in a dedicated plugin maybe.

I don't think it matters that much but in terms of code maaintainability/readability I would consider moving the scrobling features into its own module/class and then have a scrobble instance within the provider.

music_assistant/providers/plex/__init__.py
```diff
@@ -979,15 +1031,425 @@ async def on_streamed(
...
+            except Exception as err:
+                self.logger.exception(
+                    "Failed to update Plex timeline for track %s: %s",
+                    prov_item_id,
+                    err,
+                )
+
+        await asyncio.to_thread(update_timeline)
+
+    async def _periodic_timeline_update(self) -> None:  # noqa: PLR0915
```
marcelveldt: I don't fully follow why this is needed ? We already have the "on_played" callback which will hit every 30 seconds while an item is playing. Is that not enough ? This adds a lot of overhead

music_assistant/providers/plex/__init__.py
```diff
@@ -979,15 +1031,425 @@ async def on_streamed(
...
+                    del self._last_reported_state[stopped_key]
+
+            except asyncio.CancelledError:
+                self.logger.debug("Periodic timeline update task cancelled")
+                raise
+            except Exception as err:
+                self.logger.exception("Error in periodic timeline update: %s", err)
+                await asyncio.sleep(5)
+
+    async def _on_media_item_played_event(self, event: MassEvent) -> None:
```
marcelveldt: This is redundant with the on_played callback. You would only implement this if you were a plugin actively listening for all playback events to scrobble it to e.g. listenbrainz or whatever.
In this case you are the plex provider, listening for any playback of plex items.

anatosun:
I wanted to implement more aggresive updates but I realize now that it's sort of unnecessary. Thank you for reviewing the PR!

music_assistant/providers/plex/__init__.py
marcelveldt: this comment is not resolved. Can this entire handler be removed from the PR ?

music_assistant/providers/plex/__init__.py
```diff
@@ -393,6 +408,23 @@ def connect() -> PlexServer:
...
+    async def unload(self, is_removed: bool = False) -> None:
+        """Handle unload/close of the provider."""
+        # Cancel periodic timeline update task
+        if self._timeline_update_task and not self._timeline_update_task.done():
+            self._timeline_update_task.cancel()
+            with suppress(asyncio.CancelledError):
+                await self._timeline_update_task
+
+        await super().unload(is_removed)
+
```
marcelveldt:
```suggestion
```
seems to be dead code

music_assistant/providers/plex/__init__.py
```diff
@@ -334,6 +336,10 @@ class PlexProvider(MusicProvider):
     _plex_library: PlexMusicSection = None
     _myplex_account: MyPlexAccount = None
     _baseurl: str
+    _timeline_update_task: Task[None] | None = None
+    _last_reported_state: dict[
+        str, tuple[str, int, PlaybackState]
+    ]  # queue_id_itemid -> (track_id, position, state)
```
marcelveldt:
```suggestion
```
leftover dead code

music_assistant/providers/plex/__init__.py
```diff
@@ -979,15 +1011,160 @@ async def on_streamed(
...
+                self._plex_server.query("/:/timeline", params=params)
+
+                # If fully played, also scrobble
+                if fully_played:
+                    scrobble_params = {
+                        "key": str(plex_track.ratingKey),
+                        "identifier": "com.plexapp.plugins.library",
+                    }
+                    self.logger.debug("Scrobbling track to Plex: %s", scrobble_params)
+                    self._plex_server.query("/:/scrobble", scrobble_params)
```
marcelveldt:
```suggestion
                    self._plex_server.query("/:/scrobble", parms=scrobble_params)
```

anatosun: It should be `params` here, right? Could you modify the suggestion so I can apply it?

marcelveldt: sorry for the typo. I see you already corrected it

marcelveldt (approved):
Nice work, @anatosun !

PR #2513, opened 2025-10-17
Add dynamic chapter url retrieval solution to ABS

marcelveldt:
While streaming an audiobook from ABS, the token may expire halfway a chapter, so as soon as playback starts of a new chapter, the token is outdated and playback will fail.

This workarounds this by ensuring we always fetch the token between chapters.
This is done by sending an intermediate dynamic url to the streamdetails which we resolve in the provider just in time to the authenticated url.

Untested because I don't have a ABS account/setup but @fmunkes should be able to test if this approach works.
Marked as draft so first can be confirmed if it actually works.


fmunkes (approved):
Thank you very much!

PR #2514, opened 2025-10-18
Add (initial) support for AirPlay 2 to the AirPlay player provider

bradkeifer:
Initial work to support a seperate cli binary for AirPlay2 devices.


music_assistant/providers/airplay/player.py
```diff
@@ -271,25 +291,33 @@ async def play_media(self, media: PlayerMedia) -> None:
...
                 return
 
-        # setup RaopStreamSession for player (and its sync childs if any)
+        # setup StreamSession for player (and its sync childs if any)
         sync_clients = self._get_sync_clients()
         provider = cast("AirPlayProvider", self.provider)
-        raop_stream_session = RaopStreamSession(provider, sync_clients, input_format, audio_source)
-        await raop_stream_session.start()
+        if self.mass.config.get_raw_player_config_value(self.player_id, CONF_AIRPLAY_VERSION) == 2:
+            ap2_stream_session = AirPlay2StreamSession(
```
marcelveldt: In a follow-up PR we should aim on using a single session regardless the ap version

music_assistant/providers/airplay/helpers.py
```diff
@@ -89,29 +91,59 @@ def is_broken_raop_model(manufacturer: str, model: str) -> bool:
...
-            if returncode == 0 and output.strip().decode() == "cliraop check":
-                return cliraop_path
+            if version == 1:
+                args = [
+                    cli_path,
+                    "-check",
+                ]
+                passing_output = "cliraop check"
+            else:
+                config_file = os.path.join(os.path.dirname(__file__), "bin", "cliap2.conf")
```
marcelveldt, line 106: Is the .conf a typo ?

music_assistant/providers/airplay/provider.py
```diff
@@ -144,6 +148,37 @@ async def _setup_player(
...
+            player_id, CONF_AIRPLAY_VERSION, None
+        )
+        if conf_airplay_version is None:
+            # There is currently no configured AirPlay version, therefore we will
+            # save the discovered airplay version to the provider config.
+            # This is needed for the player to know which protocol to use.
+            self.logger.debug(f"Saving AirPlay version {disc_airplay_version} for {display_name}")
+            await self.mass.config.save_player_config(
+                player_id, {CONF_AIRPLAY_VERSION: disc_airplay_version}
+            )
```
marcelveldt: this is not allowed, you should instead simply provide a default value in the config entry, not save it yourself from the provider to the config controller.

music_assistant/providers/airplay/player.py
bradkeifer: ok

music_assistant/providers/airplay/helpers.py
bradkeifer: No, it isn't. This is a construct from owntones that I don't think we can easily overcome. There is a configuration file that is built as part of the build processing of owntones. As we make progress on the CI building, I am pretty sure that we will be able to get the binary and config file built and everything installed in the appropriate directories such that we don't need to specify it as a command line argument

music_assistant/providers/airplay/provider.py
bradkeifer: ok. Can you please help me with how to do that? I have created a helper function to identify devices which should default to ap2 `is_airplay2_model()`. It takes the same approach as the `broken_raop_model()` helper. I suspect that ultimately, we should use one of the `txt` properties from the `_airplay._tcp` discovery info to determine whether to use raop or ap2.

music_assistant/providers/airplay/airplay2.py
```diff
@@ -0,0 +1,518 @@
...
+    async def start(self) -> None:
+        """Initialize AirPlay2StreamSession."""
+        # initialize airplay stream for all players
+
+        # get current ntp and start AirPlay2Stream per player
+        assert self.prov.cli_bin
+        args = [
+            self.prov.cli_bin,
+            "--config",
+            os.path.join(os.path.dirname(__file__), "bin", "cliap2.conf"),
```
marcelveldt: why is the binary called cliap2.conf ?

music_assistant/providers/airplay/provider.py
marcelveldt: Once the missing arguments are in place so that sync works, I would simply default to airplay 2 if the airplay mdns record is found

bradkeifer: Actioned. We now default to airplay 2 if the mdns record is found and the cliap2 binary is in place and the config entry is set to airplay2. The config entry default is currently 1, but we can change that to 2 once we have a working and battle tested cliap2.

music_assistant/providers/airplay/player.py
```diff
@@ -171,6 +185,22 @@ async def get_config_entries(self) -> list[ConfigEntry]:
...
                 category="airplay",
             ),
+            ConfigEntry(
+                key=CONF_AIRPLAY_VERSION,
+                type=ConfigEntryType.INTEGER,
+                default_value=1,
+                required=True,
+                label="AirPlay version to use for streaming",
+                description="AirPlay version 1 protocol uses RAOP.\n"
+                "AirPlay version 2 is an exenstion of RAOP.\n"
```
OzGav:
```suggestion
                "AirPlay version 2 is an extention of RAOP.\n"
```

marcelveldt:
Marked as draft because this is still a wip and not yet functional

music_assistant/providers/airplay/airplay2.py
```diff
@@ -0,0 +1,511 @@
...
+                self.session.input_format,
+                AIRPLAY_PCM_FORMAT,
+            ),
+        )
+        self._stream_bytes_sent = 0
+        await self._ffmpeg_proc.start()
+        chunksize = get_chunksize(AIRPLAY_PCM_FORMAT)
+        # wait for cliap2 to be ready
+        await asyncio.wait_for(self._started.wait(), 20)
+        async for chunk in self._ffmpeg_proc.iter_chunked(chunksize):
```
bradkeifer: Is this an efficient method? I suspect it is reopening the named pipe for each chunk written, and given this is a fairly IO intensive process, it might not be so efficient. Perhaps we should open the named pipe just before starting cliap2 and then have a member function to write to the named pipe using the file descriptor for the open pipe, and then only try to reopen the pipe upon a `BrokenPipeError`?
I tried using the elegant method of using extra args to the `ffmpeg` process as suggested by @marcelveldt, but this did not work. The reason was that the `chunk` was not sent to `mass` (for it to then forward to cliap2) and as a result, `mass` couldn't keep track of the `_stream_bytes_sent` and `_total_bytes_sent` values.

music_assistant/providers/airplay/constants.py
marcelveldt, line 44: we should maybe combine this list with the airplay2 models so we by default enforce them to AP2

music_assistant/providers/airplay/player.py
```diff
@@ -202,16 +232,69 @@ async def pause(self) -> None:
...
+                self._airplay_version = configured_version
+            else:
+                # This will happen when the device does not support the configured version
+                # in which case, we just stick with the discovered version
+                # Should we force an update of the config entry to match the discovered version??
+                self.logger.error(
+                    f"{self.name}:failed to refresh discovery info for AirPlay version "
+                    f"{configured_version}. "
+                    f"Continuing with previously detected version {self._airplay_version}."
+                )
```
bradkeifer: I am not sure what we should do in this instance. I don't think it is possible for there to be zero discovery information, so this case should never happen. I would appreciate advice on what to do please.

music_assistant/providers/airplay/constants.py
marcelveldt: at some point we can rename this constant to "ENFORCE_AIRPLAY_2"

kenyonj, line 45: Not sure if it’s appropriate now, but “Juke Audio” amplifiers only work with Airplay 2 and not RAOP so perhaps we should add them here as well?
```suggestion
    ("Juke Audio", "*"),
    # A recent fw update of newer gen Sonos speakers block RAOP (airplay 1) support,
```

bradkeifer:
Superseded by #2571

PR #2519, opened 2025-10-21
Plex: Add configurable bidirectional favorite sync with ratings

anatosun:
This PR implements bidirectional synchronization between Music Assistant favorites and Plex user ratings, allowing users to like/unlike tracks and albums in Music Assistant with ratings automatically syncing to Plex.

The rating mapping is configurable in the UI.

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -6,7 +6,6 @@ import asyncio
 import logging
 from asyncio import Task, TaskGroup
 from collections.abc import Awaitable
-from contextlib import suppress
 from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar, cast
 
 import plexapi.exceptions
@@ -89,6 +88,9 @@ CONF_LOCAL_SERVER_SSL = "local_server_ssl"
 CONF_LOCAL_SERVER_VERIFY_CERT = "local_server_verify_cert"
 CONF_IMPORT_COLLECTIONS = "import_collections"
 CONF_COLLECTION_PREFIX = "collection_prefix"
+CONF_PLEX_LIKE_RATING = "plex_like_rating"
+CONF_PLEX_FAVORITE_THRESHOLD = "plex_favorite_threshold"
+CONF_PLEX_UNLIKE_RATING = "plex_unlike_rating"
 
 FAKE_ARTIST_PREFIX = "_fake://"
 
@@ -99,6 +101,8 @@ SUPPORTED_FEATURES = {
     ProviderFeature.LIBRARY_ALBUMS,
     ProviderFeature.LIBRARY_TRACKS,
     ProviderFeature.LIBRARY_PLAYLISTS,
+    ProviderFeature.FAVORITE_ALBUMS_EDIT,
+    ProviderFeature.FAVORITE_TRACKS_EDIT,
     ProviderFeature.BROWSE,
     ProviderFeature.SEARCH,
     ProviderFeature.ARTIST_ALBUMS,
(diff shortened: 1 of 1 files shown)


music_assistant/providers/plex/__init__.py
```diff
@@ -1001,3 +1048,55 @@ def _refresh_plex_token() -> MyPlexAccount:
             return self._myplex_account
 
         return await asyncio.to_thread(_refresh_plex_token)
+
+    async def set_favorite(self, prov_item_id: str, media_type: MediaType, favorite: bool) -> None:
+        """Set favorite status by setting rating in Plex."""
+        if favorite:
+            # Set like rating
+            rating = cast("float", self.config.get_value(CONF_PLEX_LIKE_RATING))
```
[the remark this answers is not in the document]
anatosun: This should be rather obvious from the configuration IMHO.

music_assistant/providers/plex/__init__.py
```diff
@@ -1001,3 +1048,55 @@ def _refresh_plex_token() -> MyPlexAccount:
...
+                "Set Plex rating to %s for album with ID %s (ratingKey: %s)",
+                rating,
+                prov_item_id,
+                plex_album.ratingKey,
+            )
+
+    async def library_add(self, item: MediaItemType) -> bool:
+        """Add item to library by setting the configured like rating in Plex."""
+        if item.media_type in (MediaType.TRACK, MediaType.ALBUM):
+            await self.set_favorite(item.item_id, item.media_type, True)
```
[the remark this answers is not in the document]
anatosun: The controller already sets the correct provider-specific item_id before calling library_add() (lines 625-626
  and 714-715). The controller does:  `prov_item.item_id = prov_mapping.item_id`. This means the item.item_id passed to library_add() is already the Plex provider ID. Am I getting this right, or is there indeed an unintended behavior that could arise?

music_assistant/providers/plex/__init__.py
```diff
@@ -1001,3 +1048,55 @@ def _refresh_plex_token() -> MyPlexAccount:
...
+        """Remove item from library by setting the configured unlike rating in Plex."""
+        if media_type in (MediaType.TRACK, MediaType.ALBUM):
+            await self.set_favorite(prov_item_id, media_type, False)
+            return True
+        self.logger.info(
+            "Provider %s does not support library_remove for media type %s",
+            self.name,
+            media_type,
+        )
+        return False
```
marcelveldt:
```suggestion
```

as far as I'm aware you cant edit the plex library - unless you want to actually delete items from the plex library ?

You need to implement the ProviderFeature.FAVORITE_*_EDIT feature and only the "set_favorite" method.
Only implement library_add and library_remove if you actually want to modify the plex library which I think is not your intention here, you just want to act on (un)favoriting.

music_assistant/providers/plex/__init__.py
```diff
@@ -1001,3 +1048,55 @@ def _refresh_plex_token() -> MyPlexAccount:
...
+        if item.media_type in (MediaType.TRACK, MediaType.ALBUM):
+            await self.set_favorite(item.item_id, item.media_type, True)
+            return True
+        self.logger.info(
+            "Provider %s does not support library_add for media type %s",
+            self.name,
+            item.media_type,
+        )
+        return False
+
```
marcelveldt:
```suggestion
```

see my comment on library_remove - you don't need to implement this, you only need the set_favorite method

music_assistant/providers/plex/__init__.py
```diff
@@ -95,6 +99,10 @@
     ProviderFeature.LIBRARY_ALBUMS,
     ProviderFeature.LIBRARY_TRACKS,
     ProviderFeature.LIBRARY_PLAYLISTS,
+    ProviderFeature.LIBRARY_ALBUMS_EDIT,
+    ProviderFeature.LIBRARY_TRACKS_EDIT,
+    ProviderFeature.FAVORITE_ALBUMS_EDIT,
+    ProviderFeature.FAVORITE_TRACKS_EDIT,
```
marcelveldt:
```suggestion
    ProviderFeature.FAVORITE_ALBUMS_EDIT,
    ProviderFeature.FAVORITE_TRACKS_EDIT,
```

marcelveldt: Items are in the Plex library by default - the plex library is not editable (unless you consider removing files as an option here, which I dont think so). You only need the favorite edit features and the set_favorite method.

However, I just realized a mistake in the core controller where it first checks library edit before considering favorite edit, I will fix that

music_assistant/providers/plex/__init__.py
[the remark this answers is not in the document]
marcelveldt: your assumption is correct

marcelveldt:
Marked as draft until the (final) review comments are addressed. Press the "ready for review" button once you're ready.

music_assistant/providers/plex/__init__.py
marcelveldt: issue in core is now fixed (if you rebase your branch)

anatosun:
Thanks a lot for the review and the fix. Since I'm new to the codebase, I had assumed that it was the proper way to implement the rating feature.

marcelveldt (approved):
Very nice feature @anatosun !
Reminder to self: We should also implement it this way in the local files provider some day.

PR #2253, opened 2025-06-25
Implement pagination for plex provider

chicco-carone:
This pull request refactors the `get_library_tracks` method in the `music_assistant/providers/plex/__init__.py` file to improve the handling of large datasets by implementing pagination.

### Refactoring for pagination:

* [`music_assistant/providers/plex/__init__.py`](diffhunk://#diff-44fe5cf92cc42bee8cefcfc3e56f544728a8a954cf8c944ecd1c095b1ef31d87L830-R846): Updated the `get_library_tracks` method to use a paginated approach with a page size of 500 and an offset mechanism. This replaces the previous implementation that fetched all tracks at once, enhancing scalability and performance for large libraries. 

This is the first pr for #2228 containing only the fixes to the plex provider. After this is merged i'll proceed to create the other one for the plex connect plugin

PR #2520, opened 2025-10-21
Plex: Use container_start and container_size for pagination

ferges123:
Fixes "Unknown filter field" error when syncing Plex library by using the correct pagination parameters (container_start/container_size) expected by plexapi.

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -832,8 +832,8 @@ class PlexProvider(MusicProvider):
                 await self._run_async(
                     self._plex_library.searchTracks,
                     title=None,
-                    limit=page_size,
-                    offset=offset,
+                    container_size=page_size,
+                    container_start=offset,
                 ),
             )
             if not batch:


marcelveldt:
I assume that Plex change the api syntax ? Do we need to check for a certain minimal version now or something ?

marcelveldt:
Marked as draft until the (question is answered. I'm wondering if this api change is now relevant to all plex users or only the ones that updated to the latest and greatest.

Press the "ready for review" button once you're ready.

ferges123:
It appears the offset parameter was used by mistake in the https://github.com/music-assistant/server/pull/2253.
The python-plexapi library has never used offset for search pagination (but only for playback control e.g., _client.seekTo(offset=...)_) The correct method for paginating search results has always been to use the container_start and container_size parameters.

Here is the official documentation confirming this:
- _search()_ function docs: Lists container_start and container_size as the correct arguments for pagination. https://python-plexapi.readthedocs.io/en/latest/modules/library.html#plexapi.library.LibrarySection.search
- _fetchItems()_ function docs: This is the underlying function called by search. It also confirms the parameters and explicitly states: container_start (None, int) – offset to get a subset of the data. https://python-plexapi.readthedocs.io/en/latest/modules/base.html#plexapi.base.PlexObject.fetchItems

marcelveldt (approved):
Thanks, @ferges123 !

PR #2521, opened 2025-10-21
Plex: Add support for importing collections as playlists

anatosun:
- Add configurable collection import toggle (disabled by default)
- Add configurable collection prefix (default: "Collection: ")
- Collections are imported as read-only playlists
- Support for collections containing tracks and albums
- Include collection poster/thumbnail images


marcelveldt (approved):
LGTM! Nice work @anatosun 

BTW: Feel free to add/replace your name as maintainer in the manifest for Plex, it seems that you are fixing a lot of stuff in the provider after it has been neglected for a while by the original author

PR #2526, opened 2025-10-21
Plugin fixes

sprocket-9:
A few fixes for state updates when a player's source is a plugin.

#### player.active_source setter
player.state.active_source = plugin_source_id gets overwritten with player_id at next state update as the getter player.active_source always returns player_id unless _attr_active_source is set.

#### deepcopy(source.metadata)
Without the copy, spotify-connect plugin updates its metadata object on track change but  prev_state = deepcopy(self._state) already has the changed metadata so no differences are detected when compared against new state.

### polling updates for player with plugin source
Think this is required as I can't see anywhere else an update_state will occur for a player with a plugin source.

Only the spotify-connect plugin currently updates its PluginSource.metadata on track changes and requires a player to update_state() in order to send the changes to frontend. Not all plugins will require this as they're just an audio stream with no changes in metadata, so for a little bit more efficiency could add a boolean attribute in models.plugin.PluginSource to set whether state updates are required.

### DIFF
--- a/music_assistant/controllers/players/player_controller.py
+++ b/music_assistant/controllers/players/player_controller.py
@@ -2019,6 +2019,9 @@ class PlayerController(CoreController):
                 # to ensure the queue has accurate details
                 player_playing = player.playback_state == PlaybackState.PLAYING
                 if player_playing:
+                    if player.active_source and self.get_plugin_source(player.active_source):
+                        self.trigger_player_update(player.player_id)
+                        continue
                     self.mass.loop.call_soon(
                         self.mass.player_queues.on_player_update,
                         player,
--- a/music_assistant/controllers/streams.py
+++ b/music_assistant/controllers/streams.py
@@ -1009,7 +1009,7 @@ class StreamsController(CoreController):
             if plugin_source.stream_type == StreamType.CUSTOM
             else plugin_source.path
         )
-        player.state.active_source = plugin_source_id
+        player.active_source = plugin_source_id
         plugin_source.in_use_by = player_id
         try:
             async for chunk in get_ffmpeg_stream(
@@ -1026,7 +1026,7 @@ class StreamsController(CoreController):
                 "Finished streaming PluginSource %s to %s", plugin_source_id, player_id
             )
             await asyncio.sleep(0.5)
-            player.state.active_source = player.player_id
+            player.active_source = player.player_id
             plugin_source.in_use_by = None
 
     async def get_queue_item_stream(
--- a/music_assistant/models/player.py
+++ b/music_assistant/models/player.py
@@ -814,7 +814,6 @@ class Player(ABC):
         return None
 
     @property
-    @final
     def active_source(self) -> str | None:
         """
         Return the FINAL active source of the player.
@@ -828,6 +827,12 @@ class Player(ABC):
                 return parent_player.active_source
         # in case player's source is None, return the player_id (to indicate MA is active source)
         return self._active_source or self.player_id
+
+    @active_source.setter
+    @final
+    def active_source(self, source: str) -> None:
+        """Set the active source of the player."""
+        self._attr_active_source = source
 
     @cached_property
     @final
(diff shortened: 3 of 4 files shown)


music_assistant/controllers/players/player_controller.py
```diff
@@ -2019,6 +2019,9 @@ async def _poll_players(self) -> None:
                 # to ensure the queue has accurate details
                 player_playing = player.playback_state == PlaybackState.PLAYING
                 if player_playing:
+                    if player.active_source and self.get_plugin_source(player.active_source):
+                        self.trigger_player_update(player.player_id)
+                        continue
```
marcelveldt, line 2024: lets not do this here vut at the place where it belongs

music_assistant/controllers/streams.py
```diff
@@ -1009,7 +1009,7 @@ async def get_plugin_source_stream(
             if plugin_source.stream_type == StreamType.CUSTOM
             else plugin_source.path
         )
-        player.state.active_source = plugin_source_id
+        player.active_source = plugin_source_id
```
marcelveldt, line 1012: lets not do this this way

music_assistant/models/player.py
```diff
@@ -829,6 +828,12 @@ def active_source(self) -> str | None:
         # in case player's source is None, return the player_id (to indicate MA is active source)
         return self._active_source or self.player_id
 
+    @active_source.setter
+    @final
+    def active_source(self, source: str) -> None:
+        """Set the active source of the player."""
+        self._attr_active_source = source
+
```
marcelveldt, line 836: We cant assume each player implementation uses the properties this way. So this setter is not accepted

marcelveldt:
I agree that the plugin source is a bit borked atm and I was already looking into it. 
If you dont mind I will look into this myself as it needs a bit more changes to handle it all properly, sorry about that

music_assistant/models/player.py
sprocket-9: Oh of course, for players with internal sources like AVRs etc.  I can get a bit myopic when thinking from a squeezelite perspective!

sprocket-9:
Yeah no worries,  I can close this and leave you to the changes.  

I did have a follow up PR planned that's plugin related.  I want to allow a plugin source to switch player i.e player2 selects a plugin source already in use by player1 and have the transition handled cleanly, with a lock in get_plugin_source_stream to prevent concurrent player access to a plugin's get_audio_stream. 

If you're open to this feature I can knock up a PR and submit it so you can see at least see what I'm proposing and maybe factor it into your changes?

marcelveldt:
I'm actually chewing a bit on the whole concept of plugin providers as there are a few edge cases now not covered. If you give me a few days I'm pretty sure I can work out the quirks and give you a good foundation to continue your provider

sprocket-9:
Sounds good!  Take your time, I'll give things a try out when you merge.

marcelveldt:
> Sounds good! Take your time, I'll give things a try out when you merge.

One question: Do you think it would make sense to allow one plugin provider actually providing multiple sources ?
Or does the current model of 1 plugin = 1 source good enough ?
You can ofc provide multiple sources today by a config entry and allowing multiple instances but maybe there are situations where one plugin can provide a few sources at once - not that I can currently think of but you never know.

sprocket-9:
> One question: Do you think it would make sense to allow one plugin provider actually providing multiple sources ? Or does the current model of 1 plugin = 1 source good enough ? You can ofc provide multiple sources today by a config entry and allowing multiple instances but maybe there are situations where one plugin can provide a few sources at once - not that I can currently think of but you never know.

I can see a case for this if the plugin needs sole control of a base resource like a socket or a hardware device that would cause a clash if multiple instances of the plugin were installed.

With the vban protocol it's possible for a receiver to listen on a UDP socket and receive multiple named streams from a sender, with each stream having different pcm audio formats.  To implement this would require one plugin with multiple PluginSources.  Now I'm definitely NOT saying I want or need this!  It's just an example. For this case it's just as easy to install a second instance of the plugin and listen on another UDP port.

Another plugin idea I had is using a rtl-sdr stick to provide a list of DAB radio stations to MA.  That would be one resource providing multiple sources, Once the card is tuned to a DAB frequency, I *think* you can listen simultaneously to all the stations provided on that multiplex, which would map to multiple PluginSources.  However, I've not really put any thought into how I'd implement this or if it will ever happen! :smile:

PR #2530, opened 2025-10-21
Fix recursion loop in fake volume mute control

OzGav:
When using fake mute control, unmuting causes infinite recursion because:

`cmd_volume_mute(False)` sets internal flags and calls `cmd_volume_set()` to restore volume
`cmd_volume_set()` checks if player is muted, and if so, calls `cmd_volume_mute(False)`
This creates a loop because `player.update_state()` wasn't called before step 2

Fix: Call `player.update_state()` before `cmd_volume_set()` to ensure the player's muted state is updated and visible to the volume set check.


marcelveldt (approved):
Thanks, @OzGav !

PR #2532, opened 2025-10-22
Plex: Add anatosun as codeowner and change stage to alpha

anatosun:
I'd like to take over as maintainer of the Plex provider and move it from unmaintained to alpha stage. Some features still need more testing by fellow Plex users, which is why this is marked as alpha for now. Now that the Plex API has become public, we might need to rejuvenate the code and explore new features.

Question: How can I discuss additional features with other contributors? Is there a discord I can join?

### DIFF
--- a/music_assistant/providers/plex/manifest.json
+++ b/music_assistant/providers/plex/manifest.json
@@ -1,10 +1,10 @@
 {
   "type": "music",
   "domain": "plex",
-  "stage": "unmaintained",
+  "stage": "alpha",
   "name": "Plex Media Server Library",
   "description": "Stream your personal music, podcasts, and radio via your Plex media server.",
-  "codeowners": ["@micha91"],
+  "codeowners": ["@micha91", "@anatosun"],
   "requirements": ["plexapi==4.17.1"],
   "documentation": "https://music-assistant.io/music-providers/plex/",
   "multi_instance": true


marcelveldt:
This is very much appreciated! The Plex provider was in an unmaintained state for a while now because the original maintainer is busy with other stuff and cant devote any free time to OSS anymore.

Yes! We have a discord where we have user support channels but also dev channels to discuss development, features and such. https://discord.gg/UwdRHNW5

music_assistant/providers/plex/manifest.json
```diff
@@ -1,10 +1,10 @@
 {
   "type": "music",
   "domain": "plex",
-  "stage": "unmaintained",
+  "stage": "alpha",
   "name": "Plex Media Server Library",
   "description": "Stream your personal music, podcasts, and radio via your Plex media server.",
-  "codeowners": ["@micha91"],
+  "codeowners": ["@micha91", "@anatosun"],
```
marcelveldt, line 7: if micha91 doesn't return in another haf year or so and you keep chanhing the provider this much, feel free to just only put your name here and remove micha91. This field is about noticing the maintainer of the provider. For credits we can leverage a docstring/comment in the code itself.

marcelveldt (approved):

OzGav:
A little heads up about this issue https://github.com/music-assistant/support/issues/2308 which would love some attention!

PR #2537, opened 2025-10-23
Squeezelite: Remove volume mute PlayerFeature

OzGav:
Looking at https://github.com/music-assistant/support/issues/3903 and https://github.com/music-assistant/support/issues/4023

After investigation I think the issue lies in the squeezelite software and I have proposed a change to address it here https://github.com/ralph-irving/squeezelite/pull/253

While waiting for that this PR switches squeezelite to the fake mute control

P.S. And I'm not sure if this fixes it but this seems to be solved as well https://github.com/music-assistant/support/issues/3944 so if this is accepted I will close that as well.

### DIFF
--- a/music_assistant/providers/squeezelite/player.py
+++ b/music_assistant/providers/squeezelite/player.py
@@ -15,11 +15,6 @@ from aioslimproto.models import Preset as SlimPreset
 from aioslimproto.models import SlimEvent
 from aioslimproto.models import VisualisationType as SlimVisualisationType
 from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption
-from music_assistant_models.constants import (
-    PLAYER_CONTROL_FAKE,
-    PLAYER_CONTROL_NATIVE,
-    PLAYER_CONTROL_NONE,
-)
 from music_assistant_models.enums import (
     ConfigEntryType,
     ContentType,
@@ -39,7 +34,6 @@ from music_assistant.constants import (
     CONF_ENTRY_HTTP_PROFILE_FORCED_2,
     CONF_ENTRY_OUTPUT_CODEC,
     CONF_ENTRY_SYNC_ADJUST,
-    CONF_MUTE_CONTROL,
     DEFAULT_PCM_FORMAT,
     VERBOSE_LOG_LEVEL,
     create_sample_rates_config_entry,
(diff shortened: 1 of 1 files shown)


marcelveldt:
This may not be backported to stable as its incompatible. No changes to players/playerproviders may be bacported

music_assistant/providers/squeezelite/player.py
```diff
@@ -148,8 +154,28 @@ async def get_config_entries(self) -> list[ConfigEntry]:
...
+                ConfigValueOption(title="None", value=PLAYER_CONTROL_NONE),
+                ConfigValueOption(title="Fake mute control", value=PLAYER_CONTROL_FAKE),
+                ConfigValueOption(title="Native mute control", value=PLAYER_CONTROL_NATIVE),
+            ],
+            category="player_controls",
+        )
+
+        # Filter out the base mute control entry and add our custom one
+        filtered_entries = [e for e in base_entries if e.key != CONF_MUTE_CONTROL]
+
```
marcelveldt: why so complicated ? you can simply remove the Playerfeature.VOLUME_MUTE from the supported features of the player

OzGav: Ok i wasn’t sure how this would work as I couldn’t find an example. So if I remove that the fake option automatically becomes the default?
edit: OK I tried it and I see you get the NONE and FAKE options but I would like FAKE to be the default so it works without having to change anything?

marcelveldt: I don't think its a good idea to set fake mute as default

OzGav: Ok I will swap this around and add a note in the docs.

OzGav: Fixed

OzGav:
Talking with the squeezelite people it seems there is no native mute. If they use their remotes to mute it just sets the volume to 0 and then back to the original setting on unmute. So I think removing the mute feature will be the permanent solution. If we could come up with an opton to enable the fake option at the provider level rather than the individual player level I think that would be good as it is tedious having to go to each player to make the same change.

marcelveldt:
> Talking with the squeezelite people it seems there is no native mute. If they use their remotes to mute it just sets the volume to 0 and then back to the original setting on unmute. So I think removing the mute feature will be the permanent solution. If we could come up with an opton to enable the fake option at the provider level rather than the individual player level I think that would be good as it is tedious having to go to each player to make the same change.

The player does not support mute. So I don't think we should offer even more workarounds than the one we already have to add a fake mute on player level.

marcelveldt (approved):
Thanks, @OzGav !

PR #2540, opened 2025-10-23
Typing fixes for the stream controller

OzGav:


music_assistant/controllers/streams.py
```diff
@@ -248,9 +252,16 @@ async def setup(self, config: CoreConfig) -> None:
             self.publish_ip,
             self.publish_port,
         )
+        bind_port_value = self.publish_port
+        if isinstance(bind_port_value, int):
+            bind_port = bind_port_value
+        elif isinstance(bind_port_value, (float, str)):
+            bind_port = int(bind_port_value)
+        else:
+            bind_port = 8097
```
marcelveldt:
```suggestion
        bind_port = int(self.publish_port) if self.publish_port else 8097
```

music_assistant/controllers/streams.py
```diff
@@ -349,9 +362,12 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
                     "Failed to get streamdetails for QueueItem %s: %s", queue_item_id, e
                 )
                 queue_item.available = False
-                raise web.HTTPNotFound(reason=f"No streamdetails for Queue item: {queue_item_id}")
+                raise InvalidDataError(f"No streamdetails for Queue item: {queue_item_id}")
```
marcelveldt: sure about this? shouldnt we just raise an http error here ?

music_assistant/controllers/streams.py
```diff
@@ -395,12 +411,21 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
...
         else:
-            smart_fades_mode = await self.mass.config.get_player_config_value(
-                queue.queue_id, CONF_SMART_FADES_MODE
+            smart_fades_mode = SmartFadesMode(
+                str(
+                    await self.mass.config.get_player_config_value(
+                        queue.queue_id, CONF_SMART_FADES_MODE
+                    )
+                    or SmartFadesMode.DISABLED
+                )
```
marcelveldt:
```suggestion
                str(
                    await self.mass.config.get_player_config_value(
                        queue.queue_id, CONF_SMART_FADES_MODE
                    )
                )
```

The configEntry has a default value it returns, it can not be None at all.
If the entry does not exist, it will raise a Keyerror

music_assistant/controllers/streams.py
```diff
@@ -395,12 +411,21 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
...
             )
-            standard_crossfade_duration = self.mass.config.get_raw_player_config_value(
+            standard_crossfade_duration_val = self.mass.config.get_raw_player_config_value(
                 queue.queue_id, CONF_CROSSFADE_DURATION, 10
             )
+            if isinstance(standard_crossfade_duration_val, (int, float)):
+                standard_crossfade_duration = int(standard_crossfade_duration_val)
+            else:
+                standard_crossfade_duration = 10
```
marcelveldt:
```suggestion
            standard_crossfade_duration_val = int(self.mass.config.get_raw_player_config_value(
queue.queue_id, CONF_CROSSFADE_DURATION, 10
))
```

music_assistant/controllers/streams.py
```diff
@@ -788,12 +811,19 @@ async def get_queue_flow_stream(
             smart_fades_mode = SmartFadesMode.DISABLED
             standard_crossfade_duration = 0
         else:
-            smart_fades_mode = await self.mass.config.get_player_config_value(
+            smart_fades_mode_val = await self.mass.config.get_player_config_value(
```
marcelveldt: some comment as above, it cant be none

music_assistant/controllers/streams.py
```diff
@@ -788,12 +811,19 @@ async def get_queue_flow_stream(
...
                 queue.queue_id, CONF_SMART_FADES_MODE
             )
-            standard_crossfade_duration = self.mass.config.get_raw_player_config_value(
+            smart_fades_mode = SmartFadesMode(
```
marcelveldt: some comment as above there is already a default returned

music_assistant/controllers/streams.py
```diff
@@ -788,12 +811,19 @@ async def get_queue_flow_stream(
...
+                str(smart_fades_mode_val) if smart_fades_mode_val else SmartFadesMode.DISABLED.value
+            )
+            standard_crossfade_duration_val = self.mass.config.get_raw_player_config_value(
                 queue.queue_id, CONF_CROSSFADE_DURATION, 10
             )
+            if isinstance(standard_crossfade_duration_val, (int, float)):
```
marcelveldt: some comment as above there is already a default returned

music_assistant/controllers/streams.py
```diff
@@ -1054,22 +1104,30 @@ async def get_queue_item_stream(
             filter_rule += ":print_format=json"
             filter_params.append(filter_rule)
         elif streamdetails.volume_normalization_mode == VolumeNormalizationMode.FIXED_GAIN:
-            # apply used defined fixed volume/gain correction
-            gain_correct: float = await self.mass.config.get_core_config_value(
+            # apply user defined fixed volume/gain correction
+            gain_value_raw = await self.mass.config.get_core_config_value(
```
marcelveldt: some comment as above there is already a default returned

music_assistant/controllers/streams.py
OzGav: It was a line ball call when I did a pros and cons for the change. In the end I thought “No streamdetails” is more of a data problem (data exists but is invalid)? Thoughts?

marcelveldt: yeah but this is a http request method. There you should return it as http error so the player will back off

OzGav: Understood. I’ll fix all this up later today.

OzGav: fixed

music_assistant/controllers/streams.py
OzGav: Fixed

music_assistant/controllers/streams.py
OzGav: mypy wasn't happy with your suggested change. Please review again where I have tried to keep your suggestion but used cast to keep mypy happy?

music_assistant/controllers/streams.py
OzGav: I have changed this to use cast() to silence the type checker, or should the return type of get_raw_player_config_value be fixed to be more specific when a default is provided?"

music_assistant/controllers/streams.py
OzGav: see above

music_assistant/controllers/streams.py
OzGav: Claude tends to want to be very defensive so I need to work out how best to avoid that. I did simplify a number of these sorts of things during my review but I will look to improve further...

music_assistant/controllers/streams.py
OzGav: fixed

music_assistant/controllers/streams.py
OzGav: fixed

music_assistant/controllers/streams.py
```diff
@@ -118,6 +123,11 @@ def parse_pcm_info(content_type: str) -> tuple[int, int, int]:
     return (sample_rate, sample_size, channels)
 
 
+async def _bytes_to_generator(data: bytes) -> AsyncGenerator[bytes, None]:
+    """Convert bytes to async generator for ffmpeg input."""
+    yield data
+
+
```
marcelveldt: this is not a typing change

music_assistant/controllers/streams.py
```diff
@@ -293,9 +303,16 @@ async def setup(self, config: CoreConfig) -> None:
             self.publish_ip,
             self.publish_port,
         )
+        bind_port_value = self.publish_port
```
marcelveldt: bind port != publish port

music_assistant/controllers/streams.py
```diff
@@ -293,9 +303,16 @@ async def setup(self, config: CoreConfig) -> None:
...
+        if isinstance(bind_port_value, int):
+            bind_port = bind_port_value
+        elif isinstance(bind_port_value, (float, str)):
+            bind_port = int(bind_port_value)
+        else:
+            bind_port = 8097
         await self._server.setup(
             bind_ip=bind_ip,
-            bind_port=self.publish_port,
+            bind_port=bind_port,  # Use validated bind_port, not self.publish_port
```
marcelveldt: a bind port is something different than a publish port

music_assistant/controllers/streams.py
```diff
@@ -363,13 +380,19 @@ async def get_plugin_source_url(
...
     ) -> str:
         """Get the url for the Plugin Source stream/proxy."""
-        if plugin_source.audio_format.content_type.is_pcm():
-            fmt = ContentType.WAV.value
-        else:
-            fmt = plugin_source.audio_format.content_type.value
+        # Use player-specific output codec if available, fallback to flac
+        output_codec_val = await self.mass.config.get_player_config_value(
+            player_id, CONF_OUTPUT_CODEC
+        )
```
marcelveldt: where is this coming from ?

music_assistant/controllers/streams.py
```diff
@@ -441,12 +467,21 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
             # no crossfade on non-tracks
             smart_fades_mode = SmartFadesMode.DISABLED
         else:
-            smart_fades_mode = await self.mass.config.get_player_config_value(
-                queue.queue_id, CONF_SMART_FADES_MODE
+            smart_fades_mode = SmartFadesMode(
```
marcelveldt: this whole block reads weird.

music_assistant/controllers/streams.py
```diff
@@ -441,12 +467,21 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
...
+                )
             )
-            standard_crossfade_duration = self.mass.config.get_raw_player_config_value(
+            standard_crossfade_duration_val = self.mass.config.get_raw_player_config_value(
                 queue.queue_id, CONF_CROSSFADE_DURATION, 10
             )
+            if isinstance(standard_crossfade_duration_val, (int, float)):
+                standard_crossfade_duration = int(standard_crossfade_duration_val)
+            else:
+                standard_crossfade_duration = 10
```
marcelveldt: why is this not just an int conversion

music_assistant/controllers/streams.py
```diff
@@ -1074,11 +1117,11 @@ async def get_queue_flow_stream(
                     # we need to correct the bytes_written accordingly so the duration
                     # calculations at the end of the track are correct
                     crossfade_part_len = len(crossfade_part)
-                    bytes_written += crossfade_part_len / 2
-                    if last_play_log_entry:
+                    bytes_written += crossfade_part_len // 2
+                    if last_play_log_entry and last_play_log_entry.seconds_streamed is not None:
```
marcelveldt: this is not a typing change, its changing logic again

music_assistant/controllers/streams.py
```diff
@@ -1153,8 +1196,13 @@ async def get_queue_flow_stream(
...
             last_part_seconds = len(last_fadeout_part) / pcm_sample_size
-            queue_track.streamdetails.seconds_streamed += last_part_seconds
-            queue_track.streamdetails.duration += last_part_seconds
+            if queue_track.streamdetails:
+                if queue_track.streamdetails.seconds_streamed is not None:
+                    queue_track.streamdetails.seconds_streamed += last_part_seconds
+                if queue_track.streamdetails.duration is not None:
+                    queue_track.streamdetails.duration = int(
+                        queue_track.streamdetails.duration + last_part_seconds
+                    )
```
marcelveldt: this is awful

music_assistant/controllers/streams.py
```diff
@@ -1220,13 +1271,18 @@ async def get_plugin_source_stream(
         # this should already be set by the player controller, but just to be sure
         plugin_source.in_use_by = player_id
 
+        # Determine audio input and validate it's not None
+        audio_input = (
+            plugin_prov.get_audio_stream(player_id)
+            if plugin_source.stream_type == StreamType.CUSTOM
+            else plugin_source.path
+        )
+        if audio_input is None:
```
marcelveldt: I'm not liking the fact that we now need 1000s extra lines of codes for something that can not happen

music_assistant/controllers/streams.py
```diff
@@ -1268,22 +1324,31 @@ async def get_queue_item_stream(
             filter_rule += ":print_format=json"
             filter_params.append(filter_rule)
         elif streamdetails.volume_normalization_mode == VolumeNormalizationMode.FIXED_GAIN:
-            # apply used defined fixed volume/gain correction
-            gain_correct: float = await self.mass.config.get_core_config_value(
+            # apply user defined fixed volume/gain correction
+            gain_value_raw = await self.mass.config.get_core_config_value(
```
marcelveldt: all this config value handling is getting really annoying. We should fix this at the roots by providing a type hint to the get_*_config_value functions, now we are repeating the same checks 100000 times.

I think it can be as simple as as adding an extra parameter to the functions what the expected output type should be and we can then use that as return type (by using a typing generic).

music_assistant/controllers/streams.py
OzGav: The penny has now dropped that the config entry only will allow an INT so this is actually simple.

music_assistant/controllers/streams.py
OzGav: Looking back this was a mistake way back. Reverted.

music_assistant/controllers/streams.py
OzGav: simplified

music_assistant/controllers/streams.py
OzGav: Simplify this as well

music_assistant/controllers/streams.py
OzGav: True...switched to an assert and added the int() cast

music_assistant/controllers/streams.py
OzGav: We need a whole number of bytes in the end so switchd to int() ?

music_assistant/controllers/streams.py
OzGav: switched to cast

music_assistant/controllers/streams.py
OzGav: So for this PR and all these config value type checks - should I:

Try fixing the config system itself to handle types properly (bigger change)
Just cast everything
Keep the isinstance checks but add a TODO comment

Leaning towards #2 since I think fixing it at the roots would be a logic change?

music_assistant/controllers/streams.py
OzGav: Being overly defensive again solving a typing problem. Switched to cast and removed this.

music_assistant/controllers/streams.py
OzGav: now fixed

music_assistant/controllers/streams.py
```diff
@@ -419,13 +427,13 @@ async def serve_queue_item_stream(self, request: web.Request) -> web.Response:
             headers=headers,
         )
         resp.content_type = f"audio/{output_format.output_format_str}"
-        http_profile: str = await self.mass.config.get_player_config_value(
-            queue_id, CONF_HTTP_PROFILE
+        http_profile = str(
+            await self.mass.config.get_player_config_value(queue_id, CONF_HTTP_PROFILE)
```
marcelveldt: use the return_type argument to get_player_config_value

music_assistant/controllers/streams.py
```diff
@@ -609,10 +617,12 @@ async def serve_queue_flow_stream(self, request: web.Request) -> web.Response:
             reason="OK",
             headers=headers,
         )
-        http_profile_value = await self.mass.config.get_player_config_value(
-            queue_id, CONF_HTTP_PROFILE
+        http_profile = (
+            await self.mass.config.get_player_config_value(
+                queue_id, CONF_HTTP_PROFILE, return_type=str
+            )
+            or "default"
```
marcelveldt: we should maybe also add a default/fallback to the get_player_config_value to make statements like these easier

music_assistant/controllers/streams.py
marcelveldt: that is not what I'm talking about - there is now a check added: and last_play_log_entry.seconds_streamed is not None

music_assistant/controllers/streams.py
```diff
@@ -1275,22 +1298,33 @@ async def get_queue_item_stream(
...
+            gain_correct = round(
+                await self.mass.config.get_core_config_value(
+                    self.domain,
+                    CONF_VOLUME_NORMALIZATION_FIXED_GAIN_TRACKS
+                    if streamdetails.media_type == MediaType.TRACK
+                    else CONF_VOLUME_NORMALIZATION_FIXED_GAIN_RADIO,
+                    return_type=float,
+                )
+                or 0.0,
+                2,
```
marcelveldt: this is horrible to read

music_assistant/controllers/streams.py
OzGav: yep sorry missed this one.

music_assistant/controllers/streams.py
OzGav: I'll leave that to you to decide...! but yes that does seem like a good idea.

music_assistant/controllers/streams.py
OzGav: broken into 3 steps

music_assistant/controllers/streams.py
OzGav: OK that is removed and an assert used instead.

music_assistant/controllers/streams.py
marcelveldt: Yeah, let's add that - it will help making these statements easier to read.

marcelveldt: Added that - if you update this branch you can use it

OzGav: default parameter now used throughout

MarvinSchenkel (approved):
Nice work Gav!

PR #2544, opened 2025-10-24
Add extra buffering to queue stream to create backpressure

marcelveldt:
With crossfade/smart fades enabled, make sure we have enough headroom to do the (JIT) processing. Do this by placing a buffer in front of the queue flow stream.

This will fix short hiccups in the audio streaming when teh next track is being loaded, as has been reported for snapcast and some airplay players


music_assistant/controllers/streams.py
```diff
@@ -761,6 +762,7 @@ def get_announcement_url(
         # like https hosts and it also offers the pre-announce 'bell'
         return f"{self.base_url}/announcement/{player_id}.{content_type.value}"
 
+    @use_buffer(30, 5)
```
MarvinSchenkel: Do we maybe want to make this configureable per player / playerprovider with these values as defaults?

marcelveldt: no, there is zero reason for that

MarvinSchenkel (approved):

PR #2545, opened 2025-10-25
MusicCast: Fix ungroup & discovery

fmunkes:
2 fixes in this PR:
* music-assistant/support#4335
  aiohttp's json parsing is part of exception handling now
* Ungrouping not working:
  Ungrouping calls player.set_members, and the player remove was not implemented in the rewrite. As a side note: If I shut down Music Assistant, player.ungroup is called directly, i.e. then the players are ungroup even without the fix. Is this expected?

### DIFF
--- a/music_assistant/providers/musiccast/player.py
+++ b/music_assistant/providers/musiccast/player.py
@@ -527,9 +527,18 @@ class MusicCastPlayer(Player):
     ) -> None:
         """Set multiple members.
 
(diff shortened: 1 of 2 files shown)


marcelveldt:
> If I shut down Music Assistant, player.ungroup is called directly

Maybe at shutdown all players are being powered off and ungroup is part of power off ?

marcelveldt (approved):

PR #2546, opened 2025-10-25
Player controller mypy fixes

OzGav:


khers (commented):
Looks good to me.

Thanks for taking this on!

music_assistant/controllers/players/player_controller.py
```diff
@@ -102,23 +103,24 @@ def handle_player_command[PlayerControllerT: "PlayerController", **P, R](
     @functools.wraps(func)
     async def wrapper(self: PlayerControllerT, *args: P.args, **kwargs: P.kwargs) -> R | None:
         """Log and handle_player_command commands to players."""
-        player_id = kwargs["player_id"] if "player_id" in kwargs else args[0]
+        player_id = kwargs.get("player_id") or args[0]
+        assert isinstance(player_id, str)
```
marcelveldt:
```suggestion
        assert isinstance(player_id, str). # for type checking
```

music_assistant/controllers/players/player_controller.py
```diff
@@ -102,23 +103,24 @@ def handle_player_command[PlayerControllerT: "PlayerController", **P, R](
     @functools.wraps(func)
     async def wrapper(self: PlayerControllerT, *args: P.args, **kwargs: P.kwargs) -> R | None:
```
marcelveldt:
```suggestion
    async def wrapper(self: PlayerControllerT, *args: P.args, **kwargs: P.kwargs) -> None:
```
The return type can actually be None because all of the player "cmd_" methods are not returning anything

music_assistant/controllers/players/player_controller.py
```diff
@@ -102,23 +103,24 @@ def handle_player_command[PlayerControllerT: "PlayerController", **P, R](
...
+        assert isinstance(player_id, str)
         if (player := self._players.get(player_id)) is None or not player.available:
             # player not existent
             self.logger.warning(
                 "Ignoring command %s for unavailable player %s",
                 func.__name__,
                 player_id,
             )
-            return
+            return None
```
marcelveldt:
```suggestion
            return
```

music_assistant/controllers/players/player_controller.py
```diff
@@ -102,23 +103,24 @@ def handle_player_command[PlayerControllerT: "PlayerController", **P, R](
...
 
         self.logger.debug(
             "Handling command %s for player %s",
             func.__name__,
             player.display_name,
         )
         try:
-            await func(self, *args, **kwargs)
+            return await func(self, *args, **kwargs)
```
marcelveldt:
```suggestion
           await func(self, *args, **kwargs)
```

music_assistant/controllers/players/player_controller.py
```diff
@@ -941,7 +943,7 @@ async def play_announcement(
                     ),
                     media_type=MediaType.ANNOUNCEMENT,
                     title="Announcement",
-                    custom_data=announce_data,
+                    custom_data=dict(announce_data),
```
marcelveldt, line 946: its weird that this one is needed because AnnounceData is a typed dict.

music_assistant/controllers/players/player_controller.py
OzGav: Ruff doesn't like it but apparently mypy does!

music_assistant/controllers/players/player_controller.py
OzGav: Not sure but this is the error

`music_assistant/controllers/players/player_controller.py:912: error: Argument "custom_data" to "PlayerMedia" has incompatible type "AnnounceData"; expected "dict[str, Any] | None"  [arg-type]`

marcelveldt: Yeah, which I think is lame haha. You kind of expect mypy to understand that the TypedDict is exactly that.

marcelveldt (approved):
Thanks, @OzGav !

PR #2552, opened 2025-10-26
Fix filesystem SMB provider

marcelveldt:
After our change from alpine to debian, the filesystem SMB provider broke.
This fixed it and also implements a few more config options (instead of having to manually fiddle with the options string)

### DIFF
--- a/music_assistant/providers/filesystem_smb/__init__.py
+++ b/music_assistant/providers/filesystem_smb/__init__.py
@@ -5,8 +5,9 @@ from __future__ import annotations
 import os
 import platform
 from typing import TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


kraphtuos:
Hey. Not sure if this is the best way to report this. I believe this refactoring introduced some issues.

In `_build_linux_mount_cmd`, password is no longer passed separately through an environment variable. This creates two problems:
1. As the previous comment explained, the password can now leak through system log (at level debug)
2. If the password contains certain special characters, they are not properly escaped and can cause problems.

I’m happy to make a PR but I don’t know how to test a change like this. Wondering if you could help with this?
Thanks!

PR #2562, opened 2025-10-27
Bluos default config tweaks

Cyanogenbot:
Fixed replay behavior from music-assistant/support#4318
Added default support for hi-rez modes, as all bluos devices support 24bit 192khz


marcelveldt (approved):
Thanks, @Cyanogenbot !

marcelveldt:
> @marcelveldt is putting safe_max_sample_rate at 192000 fine by you or should we just default 44000? Otherwise these tweaks are tested and good to go

As far as I'm aware all bluos devices support it so then its safe.

PR #2565, opened 2025-10-28
Fix Squeezelite sample rate for multi client streams

OzGav:
Previously, squeezelite players in sync groups were limited to 
44.1kHz/16-bit output even when configured to support higher 
sample rates and bit depths (e.g., 96kHz/24-bit). Individual 
players worked correctly, but grouping them caused quality 
degradation.

The issue was in _serve_multi_client_stream() which created 
AudioFormat with only the content type, causing sample rate 
and bit depth to default to minimum values.

This fix:
- Uses streams controller's get_output_format() to properly 
  determine output format based on player capabilities
- Preserves original content format through MultiClientStream
- Passes content format to get_output_format() for proper 
  quality negotiation

Grouped players now output at min(content_format, player_capability),
matching the behavior of individual players.

Fixes: [music-assistant/server#<issue_number>](https://github.com/music-assistant/support/issues/4300)


music_assistant/providers/squeezelite/multi_client_stream.py
```diff
@@ -20,12 +20,14 @@ def __init__(
         self,
         audio_source: AsyncGenerator[bytes, None],
         audio_format: AudioFormat,
+        content_format: AudioFormat,
```
marcelveldt: why is this ? shouldn't be needed. multiclientstream is alway just pcm

music_assistant/providers/squeezelite/player.py
```diff
@@ -235,11 +235,14 @@ async def play_media(self, media: PlayerMedia) -> None:
             return
 
         # this is a syncgroup, we need to handle this with a multi client stream
+        flow_format = await self.mass.streams._select_flow_format(self)
```
marcelveldt: eeks what is this

music_assistant/providers/squeezelite/player.py
```diff
@@ -265,24 +268,28 @@ async def play_media(self, media: PlayerMedia) -> None:
             audio_source = ugp_stream.get_stream(master_audio_format, filter_params=None)
         elif media.source_id and media.queue_item_id:
             # regular queue stream request
+            queue_item = self.mass.player_queues.get_item(media.source_id, media.queue_item_id)
```
marcelveldt: I dont really see the benefit of adding an extra variable here

music_assistant/providers/squeezelite/player.py
```diff
@@ -265,24 +268,28 @@ async def play_media(self, media: PlayerMedia) -> None:
...
             audio_source = self.mass.streams.get_queue_flow_stream(
                 queue=self.mass.player_queues.get(media.source_id),
-                start_queue_item=self.mass.player_queues.get_item(
-                    media.source_id, media.queue_item_id
-                ),
+                start_queue_item=queue_item,
                 pcm_format=master_audio_format,
             )
+            content_format = queue_item.streamdetails.audio_format
```
marcelveldt: ???

music_assistant/providers/squeezelite/player.py
```diff
@@ -265,24 +268,28 @@ async def play_media(self, media: PlayerMedia) -> None:
...
+                    content_type=ContentType.try_parse(media.uri) or ContentType.UNKNOWN
+                ),
                 output_format=master_audio_format,
             )
         # start the stream task
         self.multi_client_stream = stream = MultiClientStream(
-            audio_source=audio_source, audio_format=master_audio_format
+            audio_source=audio_source,
+            audio_format=master_audio_format,
+            content_format=content_format,
```
marcelveldt:
```suggestion
```

music_assistant/providers/squeezelite/provider.py
```diff
@@ -194,19 +198,23 @@ async def _serve_multi_client_stream(self, request: web.Request) -> web.Response
             "Start serving multi-client flow audio stream to %s",
             child_player.display_name,
         )
-        output_format = AudioFormat(content_type=ContentType.try_parse(fmt))
+
+        output_format = await self.mass.streams.get_output_format(
+            output_format_str=fmt,
+            player=child_player,
+            content_sample_rate=stream.content_format.sample_rate,  # ← CHANGE
+            content_bit_depth=stream.content_format.bit_depth,
```
marcelveldt: this needs to be the flow pcm format!

music_assistant/providers/squeezelite/multi_client_stream.py
OzGav: Fixed - reverted this change. The issue was in how the flow sample rate was determined, not in MultiClientStream.

music_assistant/providers/squeezelite/player.py
OzGav: From what I understand, previously the flow always used INTERNAL_PCM_FORMAT.sample_rate (a fixed value). Now it calls _select_flow_format() to get the player's maximum supported sample rate, then caps it to the content sample rate. This ensures the flow runs at min(content_sample_rate, player_max_sample_rate) to match single player behavior. Thats my plan anyway?

music_assistant/providers/squeezelite/player.py
OzGav: I think its needed to access queue_item.streamdetails.audio_format.sample_rate so we can set the flow format to min(content_sample_rate, player_max_sample_rate). This ensures the flow runs at the appropriate sample rate for the content rather than always at player maximum??

music_assistant/providers/squeezelite/player.py
OzGav: Removed

music_assistant/providers/squeezelite/provider.py
OzGav: Fixed

music_assistant/providers/squeezelite/player.py
OzGav: Reverted after review

music_assistant/providers/squeezelite/player.py
OzGav: Actually I see that it is a private method so if that is no good then I think this should work instead?

```
# Get player's maximum supported sample rate
supported_rates_conf = await self.mass.config.get_player_config_value(
    self.player_id, CONF_SAMPLE_RATES, unpack_splitted_values=True
)
supported_sample_rates: tuple[int] = tuple(int(x[0]) for x in supported_rates_conf)
player_max_sample_rate = max(supported_sample_rates) if supported_sample_rates else INTERNAL_PCM_FORMAT.sample_rate

# For queue streams, cap to content sample rate
content_sample_rate = player_max_sample_rate  # Default to player max
if media.source_id and media.queue_item_id:
    queue_item = self.mass.player_queues.get_item(media.source_id, media.queue_item_id)
    content_sample_rate = min(
        queue_item.streamdetails.audio_format.sample_rate,
        player_max_sample_rate
    )
```

OzGav: Also I guess the question might be why do this at all? I was thinking that the flow sample rate optimization might reduce unnecessary CPU usage on resource-constrained devices (eg. RPi)?

marcelveldt: Yeah, its a bit pointless. That is why I settle on 48/24 default for flow streams. Anything higher is ridiculous.

music_assistant/providers/squeezelite/player.py
marcelveldt: Yeah, but you could have left it in the block below. You only use it once.

music_assistant/providers/squeezelite/player.py
marcelveldt: i was talking about this :
await self.mass.streams._select_flow_format(self)

Calling that method with self

music_assistant/providers/squeezelite/player.py
OzGav: Fixed

music_assistant/providers/squeezelite/player.py
OzGav: Regarding the sample rate: The original issue was that grouped players were limited to 44.1/16 when the track was 96/24 and in my testing individual players don't downsample. As single players don't seem to have a limitation, why do groups behave differently?
Also, regarding _select_flow_format(self) - I think I see the issue - passing the player object (self) to the streams controller method breaks separation of concerns? Assuming you are supportive of streaming at the max the player can handle, what's the proper way to get the player's maximum supported sample rate for the flow? Should I get the player's max sample rate directly from config as I showed above?

marcelveldt: You would have to iterate all players and then select the maximum supported sample rate and use that as base for the flow stream. That or just settle the flow stream to something like 96khz (always float32)

OzGav: OK I have done the iteration for all players in the group. Seems to work well in my testing.

music_assistant/providers/squeezelite/multi_client_stream.py
```diff
@@ -7,7 +7,7 @@
 
 from music_assistant_models.media_items import AudioFormat
 
-from music_assistant.helpers.audio import get_ffmpeg_stream
+from music_assistant.helpers.audio import get_ffmpeg_stream  # type: ignore[attr-defined]
```
marcelveldt: why is this type ignore needed ?

music_assistant/providers/squeezelite/player.py
```diff
@@ -681,12 +711,16 @@ def _get_sync_clients(self) -> Iterator[SlimClient]:
         """Get all sync clients for a player."""
         yield self.client
         for member_id in self.group_members:
-            if slimplayer := self.provider.slimproto.get_player(member_id):
+            if member_id == self.player_id:  # ← Skip if it's the master itself
```
marcelveldt:
```suggestion
            if member_id == self.player_id:  # ← Skip if it's the leader itself
```

music_assistant/providers/squeezelite/multi_client_stream.py
OzGav:
```suggestion
from music_assistant.helpers.audio import get_ffmpeg_stream
```

OzGav: Good question. I just checked and its not needed now so removed.

marcelveldt (approved):
Thanks, @OzGav !

PR #2567, opened 2025-10-28
Add bbc sounds provider

kieranhogg:
Creating a new MR for this as I seem to have failed to rebase properly and it was showing more than the expected amount of commits in the previous one (#2283).

I've addressed the original comments raised by updating the queue details to use the new method. In hindsight I should have just fixed that and merged as the rest of the functionality took a disproportionate amount of time!

## Overview
This provider now implements almost all of the full BBC Sounds functionality/catalog. The logic of whether to show the UK version or the slimmed down international version is down to the Sounds API itself. I've made this a manual toggle in the settings just until it's been a bit more road-tested.

### Existing issues and queries

- Using the new StreamMetadata works fine, the image_url doesn't seem to be used, however. I'm not sure if the display image is meant to use this yet, or not?
- I currently use some folder images from a remote URL, if there is a way to include these locally that feels a neater solution

### Future discussions
With maybe a few final tweaks, this is good enough as a starting point. Looking at another iteration, these are areas I need some guidance on whether it is worth pursuing a discussion on discord/issues to improve or adapt existing functionality, or whether they are not going to be likely or possible:

- Currently streams are set to be not seekable, ideally I'd like to look to bring pausing and seeking functionality to the stations, whether that is with the current functionality or with a custom stream implementation
- Mapping all the objects to Track, Podcast and PodcastEpisode has ended up with some odd things. So at a minimum, expanding the Radio object to include something like RadioEpisode, RadioSeries. E.g. while it is functional, browsing the API can give a list of mixed PodcastEpisodes and not being able to show which Podcast they belong to
- Linked to this, the listing in the frontend are a little limiting when trying to show more information about an item
- Logged in users are able to subscribe to radio shows and podcasts which then appear in their "My Sounds" area. At one staged I had these appear in the Library podcasts section, which worked functionally, but running the provider as a half-streaming, half-not streaming felt a bit odd. The main issue was that it surfaces the fact that radio shows are classed as podcasts so they appeared under Podcasts but I would assume users would expect them to be in Radio?

Any guidance on how welcome a discussion or plan in these areas is welcome.


OzGav:
So you weren't able to map all pre-recorded episodes to Podcast for example?

kieranhogg:
> So you weren't able to map all pre-recorded episodes to Podcast for example?

The vast majority of playable items have a container which does map to a Podcast or a RadioSeries (which then also maps to a Podcast eventually). There are a handful of items that are just mapped to Tracks as they don't have a parent container.

This largely works well with the content retrieved. In the below screenshot, the first row are actual podcasts, the bottom row are treated as podcats, but arguably some are more closer to a radio series, or a collection, but it still all works fine.

image

At its least polished part, when you navigate to the "Latest news" menu item in the built-in API menu, you are presented with a collection of items which are mapped to Track and PodcastEpisode, but seeing them all in a jumble together highlights how much I'm bending MA's existing models and functionality to allow it to work.

image

There's also few clunky edge cases where I can probably improve things without any extra MA changes, I just need to think of some tidier solutions. E.g. browsing radio shows from yesterday, looks mostly fine:

image

But when played the metadata is not the neatest!

image

OzGav:
Yeah its tough when they don't have a coherent structure that you can work from. My Internet Archive provider would be way worse than this as it is just a bunch of people adding "stuff" in a variety of ways. At the end of the day if people can find in MA what they can find in the native app then that would seem to be a good outcome?

OzGav:
In regards to the logo if you can create a larger one 512x512 that would be great (wikimedia has one you could crop) and make a monochrome one as well where the three rectangles are just plain white.

music_assistant/providers/bbc_sounds/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "music",
+  "domain": "bbc_sounds",
+  "stage": "beta",
+  "name": "BBC Sounds",
+  "description": "Listen to your favourite BBC radio stations in Music Assistant",
```
OzGav:
```suggestion
  "description": "Stream live BBC radio shows, podcast series and on-demand audio.",
```

kieranhogg:
# Docs

**FEATURES**
- Stream live BBC radio shows, podcast series and on-demand audio
- Shows the currently playing song when streaming radio stations
- Access your personalised BBC Sounds content*
- Keeps your My Sounds listen history up to date with your Music Assistant activity*


_*Requires a logged in BBC account_

**CONFIGURATION**
The provider works without a login, but you will be restricted to the limited International version of the Sounds menu

**KNOWN ISSUES/ NOTES**
The full functionality is enabled when you are a UK-based listener, and signed in to your BBC account. The BBC has slowly been moving towards excluding non-UK listeners from using its own Sounds platform, more details [here](https://www.bbc.co.uk/sounds/help/questions/listening-outside-the-uk/outside-uk-changes). This provider currently allows access to live and catch up radio to International listeners via the Sounds API, but this could be restricted in the future.

**NOT YET SUPPORTED**
- Pausing and seeking live radio
- The international menu should show avaiable podcasts, this will be supported in a future release, but is not currently
- Showing the currently playing song on catch-up
- For signed-in users, accessing your subscribed content through the library is not yet supported, but is planned

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+            description="BBC Sounds has several recommendation categories, configure if "
+            "and where these are shown",
+            type=ConfigEntryType.STRING,
+            options=[
+                ConfigValueOption(
+                    "Show recommendations on the home page",
+                    _Constants.CONF_RECOMMENDATIONS_HOMEPAGE,
+                ),
+                ConfigValueOption(
+                    "Show recommendations in folders in the browse page",
```
OzGav:
```suggestion
                    "Show recommendations in folders on the browse page",
```

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+            label="Enable UK Sounds content (beta)",
+            default_value=False,
+            description="Enabling this setting unlocks the full content catalog if you are a UK"
+            "listener. As the API returns a wide range of media items under the same type, this "
+            "is turned off by default until it more widely tested for stability.",
+        ),
+        ConfigEntry(
+            key=_Constants.CONF_INTRO,
+            type=ConfigEntryType.LABEL,
+            label="A BBC Sounds account is optional, but some UK-only content may not work without",
```
OzGav:
```suggestion
            label="A BBC Sounds account is optional, but some UK-only content may not work without it",
```

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+    """
+    # ruff: noqa: ARG001
+
+    return (
+        ConfigEntry(
+            key=_Constants.CONF_ENABLE_UK_CONTENT,
+            type=ConfigEntryType.BOOLEAN,
+            label="Enable UK Sounds content (beta)",
+            default_value=False,
+            description="Enabling this setting unlocks the full content catalog if you are a UK"
```
OzGav:
```suggestion
            description="Enabling this setting unlocks the full content catalog if you are a UK "
```

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+            "listener. As the API returns a wide range of media items under the same type, this "
```
OzGav: Im not sure the users need to know why this is off by default? Seems like a TODO comment in the code?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+    action: [optional] action key called from config entries UI.
+    values: the (intermediate) raw values for config entries sent with the action.
+    """
+    # ruff: noqa: ARG001
+
+    return (
+        ConfigEntry(
+            key=_Constants.CONF_ENABLE_UK_CONTENT,
+            type=ConfigEntryType.BOOLEAN,
+            label="Enable UK Sounds content (beta)",
```
OzGav:
```suggestion
            label="Enable UK Sounds content",
```

The whole provider is beta :-)

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        return True
+
+    async def get_track(self, prov_track_id: str) -> Track:
+        """Get full track details by id.
+
+        Only called if provider supports ProviderFeature.LIBRARY_TRACKS.
```
OzGav: You dont have this feature enabled so I don't think this function should be here?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+            provider_domain=self.domain,
+            provider_instance=self.instance_id,
+        )
+
+    def _stream_error(self, item_id: str, media_type: MediaType) -> MusicAssistantError:
+        return MusicAssistantError(f"Couldn't get stream details for {item_id} ({media_type})")
+
+    @property
+    def supported_features(self) -> set[ProviderFeature]:
+        """Return the features supported by this Provider."""
```
OzGav: Have a look at the demo provider as how the supported features is passed has changed https://github.com/music-assistant/server/blob/b67c6f882af763a7c30ace9a1de2266dd89a99c9/music_assistant/providers/_demo_music_provider/__init__.py#L100

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+        return SUPPORTED_FEATURES
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
```
OzGav:
```suggestion
        """Return True as the provider is a streaming provider."""
```

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+        track = await self.adaptor.new_object(episode_info, force_type=Track)
+        if not isinstance(track, Track):
+            raise MusicAssistantError(f"Incorrect track returned for {prov_track_id}")
+        return track
+
+    async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
+        # If we are requesting a previously-aired radio show, we lose access to the
+        # schedule time. The best we can find out from the API is original release
+        # date, so the stream title loses access to the air date
+        """Get full podcast epsisode details by id."""
```
OzGav:
```suggestion
        """Get full podcast episode details by id."""
```

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+        ):
+            podcasts = [await self.adaptor.new_object(show) for show in search_result.shows]
+            results.podcasts = [podcast for podcast in podcasts if isinstance(podcast, Podcast)]
+
+        return results
+
+    async def get_podcast(self, prov_podcast_id: str) -> Podcast:
+        """Get full podcast details by id.
+
+        Only called if provider supports ProviderFeature.LIBRARY_PODCASTS.
```
OzGav: You should remove these comments from the model. edit: Actually if you don't have these features declared as supported then these are not required?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+        self,
+        media_type: MediaType,
+        prov_item_id: str,
+        fully_played: bool,
+        position: int,
+        media_item: MediaItemType,
+        is_playing: bool = False,
+    ) -> None:
+        """Handle callback when a (playable) media item has been played."""
+        # This is an OPTIONAL callback that is called when an item has been streamed.
```
OzGav: You should remove these comments

OzGav:
@kieranhogg You can review the docs here https://github.com/music-assistant/beta.music-assistant.io/pull/49

Let me know if any changes are required!

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: Thought I was going mad and imagining it, but yes this does seem to be called even if ProviderFeature.LIBRARY_TRACKS isn't supported.

image

OzGav: I'm not the expert here but I understand the LIBRARY_* features are for when the provider has a library of media items that you are syncing to MA. So if you have "tracks" in your BBC Sounds account then it makes it possible to pull those tracks across. So either it is a bug or the comment in the code is outdated...? Marcel will know!

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: I was just brave and removed the toggle, the provider and API should handle it without

kieranhogg:
> In regards to the logo if you can create a larger one 512x512 that would be great (wikimedia has one you could crop) and make a monochrome one as well where the three rectangles are just plain white.

sounds-512
sounds-mono-white
sounds-mono

music_assistant/providers/bbc_sounds/__init__.py
marcelveldt: that docstring is simply not right - you are probably returning a track somewhere and when those details are viewed, the core wants to fetch the full details through this method. If that is something you wont be able to handle, we'll have to look for a workaround.

kieranhogg: It's the one from the demo provider so I need to delete it away. It works completely fine as is, I return Tracks which are streaming items. get_track() and get_stream_details() both work as I need them to, so unless it's problematic, no changes needed.

kieranhogg:
I've just pushed a couple of extra tweaks for this PR. If there are no issues with using remote URLs for the images and StreamMetadata.image_url doesn't actually get used for the player image then this one is complete from my side, any further review comments aside.

marcelveldt:
StreamMetadata.image_url is not yet being used - the frontend needs to be adjusted still.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+                    _Constants.CONF_STREAM_FORMAT_HLS,
+                ),
+                ConfigValueOption(
+                    "MPEG-DASH",
+                    _Constants.CONF_STREAM_FORMAT_DASH,
+                ),
+            ],
+            default_value=_Constants.CONF_STREAM_FORMAT_HLS,
+            required=True,
+        ),
```
marcelveldt, line 145: why would you make this an option and not just pick one that works best ?
If you want to keep it, I would at least set the category to "advanced"

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+                    "Show recommendations in folders on the browse page",
+                    _Constants.CONF_RECOMMENDATIONS_BROWSE,
+                ),
+                ConfigValueOption(
+                    "Disable recommendations", _Constants.CONF_RECOMMENDATIONS_DISABLE
+                ),
+            ],
+            default_value=_Constants.CONF_RECOMMENDATIONS_HOMEPAGE,
+            required=True,
+        ),
```
marcelveldt: Why make this configurable ? Just return recommendations and if the user doesnt want them, they can hide the rows on the homescreen

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+            key=_Constants.CONF_UPDATE_INTERVAL,
+            category="advanced",
+            type=ConfigEntryType.INTEGER,
+            label="Player update interval (seconds)",
+            description="How often to check for now playing updates",
+            default_value=5,
+            range=(1, 60),
+            depends_on="now_playing",
+            depends_on_value=True,
+        ),
```
marcelveldt: why would you make this an option and not just pick a value that works best ?
If you want to keep it, I would at least set the category to "advanced"

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+        ConfigEntry(
+            key=_Constants.CONF_NOW_PLAYING,
+            category="advanced",
+            type=ConfigEntryType.BOOLEAN,
+            label="Show 'now playing' details?",
+            description=(
+                "Show details of the currently playing track instead of the station details"
+            ),
+            default_value=True,
+        ),
```
marcelveldt: why would you make this an option ?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+
+    def _stream_error(self, item_id: str, media_type: MediaType) -> MusicAssistantError:
+        return MusicAssistantError(f"Couldn't get stream details for {item_id} ({media_type})")
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True as the provider is a streaming provider."""
+        return True
+
+    async def get_track(self, prov_track_id: str) -> Track:
```
marcelveldt, line 239: this doesnt need any caching ?

note that you can use the @use_cache decorator

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+        """Get full track details by id."""
+        episode_info = await self.client.streaming.get_by_pid(
+            pid=prov_track_id, stream_format=self.stream_format
+        )
+        track = await self.adaptor.new_object(episode_info, force_type=Track)
+        if not isinstance(track, Track):
+            raise MusicAssistantError(f"Incorrect track returned for {prov_track_id}")
+        return track
+
+    async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
```
marcelveldt, line 250: same comment about caching

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -0,0 +1,860 @@
...
+                    prev = segment
+                    segment = next(segments_iter)
+                self.logger.warning("Starting with first segment")
+                if prev and seconds > prev.offset["start"] and seconds < prev.offset["end"]:
+                    if stream_details.stream_metadata:
+                        stream_details.stream_metadata.artist = prev.titles["primary"]
+                        stream_details.stream_metadata.title = prev.titles["secondary"]
+                        if prev.image_url:
+                            stream_details.stream_metadata.image_url = prev.image_url
+            while True:
```
marcelveldt, line 337: maybe you should break out this loop if "seconds_treamed" is set on the streamdetails ?
or do you have another way of aborting this task ?

kieranhogg: You suggested this previously but I really struggled with using seconds_streamed to break out, although it was a while now so I can't remember exactly why. 

I cancel the tasks in on_played() which seems to work. Although as I write this out, I do wonder if this handles the scenario of multiple streams to multiple players correctly. I'll need to test.

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: This is a holdover from when I used the incorrect method of swapping the stream details over — it appeared to be playing the song and I thought some might prefer just to see the station name. It's redundant now so I'll remove, thanks.

marcelveldt:
Converted to draft so you can work out the latest small feedback comments. Push the "ready for review" button as soon as you're ready for the final check and merge!

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: I was sadly correct that my `on_played()` doesn't quite work fully. As `get_stream_details()` doesn't have any knowledge of the queue, there's no neat way of keeping track which task needs to be aborted, the current code just stores the last started stream which isn't correct.

So coming back to the original `seconds_streamed` issue in the hope of breaking out there, this is what I observe when I start, pause, then stop a stream. From your suggestion, I understand this not to be the expected behaviour?

```
2025-11-08 13:43:14.814 INFO (MainThread) [music_assistant.bbc_sounds] Seconds streamed for bbc_6music: None
2025-11-08 13:43:15.858 INFO (MainThread) [music_assistant.bbc_sounds] Seconds streamed for bbc_6music: None
2025-11-08 13:43:16.900 INFO (MainThread) [music_assistant.bbc_sounds] Seconds streamed for bbc_6music: None
2025-11-08 13:43:16.982 VERBOSE (MainThread) [music_assistant.webserver] Received: {"command":"players/cmd/power","message_id":22,"args":{"player_id":"demo_0","powered":false}}
2025-11-08 13:43:16.982 DEBUG (MainThread) [music_assistant.webserver] Handling command players/cmd/power
2025-11-08 13:43:16.985 DEBUG (MainThread) [music_assistant.players] Handling command cmd_power for player Demo Player demo_0
2025-11-08 13:43:16.986 DEBUG (MainThread) [music_assistant.players] Handling command cmd_ungroup for player Demo Player demo_0
2025-11-08 13:43:16.987 DEBUG (MainThread) [music_assistant.players] Handling command cmd_stop for player Demo Player demo_0
2025-11-08 13:43:16.988 INFO (MainThread) [music_assistant._demo_player_provider.demo_0] Received STOP command on player Demo Player demo_0
2025-11-08 13:43:17.197 INFO (MainThread) [music_assistant._demo_player_provider.demo_0] Received POWER OFF command on player Demo Player demo_0
2025-11-08 13:43:17.942 INFO (MainThread) [music_assistant.bbc_sounds] Seconds streamed for bbc_6music: None
2025-11-08 13:43:19.046 INFO (MainThread) [music_assistant.bbc_sounds] Seconds streamed for bbc_6music: None

Comment (kieranhogg) [reply music_assistant/providers/bbc_sounds/__init__.py:337]: I could also drop a task_id or something similar in StreamDetails.data to keep track of if that's easier than debugging the above...

Comment (marcelveldt) [reply music_assistant/providers/bbc_sounds/__init__.py:337]: Yeah that would be a neater way to do this, start the updater task in the get_streamdetails and then store it in the data. Then within that task, simply keep updating until seconds streamed is set. Just with a bit of debounce if playback is restarted within the first 15min.

Comment (kieranhogg) [reply music_assistant/providers/bbc_sounds/__init__.py:337]: See above though, seconds_streamed stays unset when I test?

File: music_assistant/providers/bbc_sounds/__init__.py [hunk 20]
```diff
@@ -0,0 +1,800 @@
...
+            episodes = [await self.adaptor.new_object(track) for track in search_result.episodes]
+            results.tracks = [track for track in episodes if type(track) is Track]
+
+        if media_types is None or MediaType.PODCAST in media_types:
+            podcasts = [await self.adaptor.new_object(show) for show in search_result.shows]
+            results.podcasts = [podcast for podcast in podcasts if isinstance(podcast, Podcast)]
+
+        return results
+
+    async def get_podcast(self, prov_podcast_id: str) -> Podcast:
```
MarvinSchenkel, line 713: We can also use caching here

music_assistant/providers/bbc_sounds/__init__.py
MarvinSchenkel: @marcelveldt Can you help out here? I think this is the last pending thing before we can merge this.

MarvinSchenkel:
In general, I think we can use some more caching and longer cache periods. Have a look at the other providers, e.g. [Qobuz](https://github.com/music-assistant/server/blob/dev/music_assistant/providers/qobuz/__init__.py) and [YTM](https://github.com/music-assistant/server/blob/dev/music_assistant/providers/ytmusic/__init__.py). Other than that, I think this is good to be merged.

music_assistant/providers/bbc_sounds/__init__.py
MarvinSchenkel: @kieranhogg I had a chat to Marcel. He will need to add something to the core code for this. For now, we can merge your PR and add a fix for the metadata in an upcoming PR. Just checking in with you if you need to disable some code before we merge this PR while we wait for the core code to be ready?

kieranhogg: The three different versions I've had for this are:

1. Disable the now playing functionality until a future update
2. Keep track of the last created task and terminate that one only, this works fine if you're streaming one station (current code)
3. Don't keep track of any tasks, they would keep running until their stream is replaced by something else (or if there is some MA garbage collection for tasks?)

Or something else temporary like keeping track of each `prov_item_id` and their task and killing them in `on_played()`, but could potentially overstep if multiple streams are using the same station?

Happy to take advice on the least bad option

MarvinSchenkel: I would disable it for now and just display something static, e.g. ""BBC Sounds" or something similar. We can then fix it as soon as the core code is ready. I would like to merge this so people can start testing as we are about to start  the beta round for 2.7 ;-)

MarvinSchenkel (approved):
Thanks @kieranhogg for your hard work 🎉 !

music_assistant/providers/bbc_sounds/__init__.py
marcelveldt: @kieranhogg would you be fine with a callback that gets called every X seconds in your code ?
So the idea I'm currently think of is to have an optional update_metadata callback on the streamdetails object.
You can then set that to a (async) method of your choice and it will be called periodically as long as playback is still ongoing. Then within your callback you can do sanity checks if updating is needed, set the fields on the streamdetails.stream_metadata if needed.

Would that work for you ?
we would then pass in the streamdetails to the callback to you can still track it back

kieranhogg: @marcelveldt Thanks for taking some time to think it over; that feels like a neat solution. I looked at various different ways to clean up the tasks, e.g. `queue_id` from `streamdetails`, but then `on_played()` only references `prov_item_id` which isn't unique so there was always a risk when cleaning up tasks itself.

Thinking of possible edge cases, there seems to be sufficient data in `streamdetails` if in future a stream could be playing 5 minutes behind live by using `created_at + seek_position + seconds_streamed` (once I get my head around seconds_streamed...).

Non-live media metadata didn't make it into this version, but that should be even easier.

OzGav:
@kieranhogg Just checking the provider against the docs and I note there are a couple of changes to the settings. One question, what is the "Show local radio stations" toggle for?

music_assistant/providers/bbc_sounds/__init__.py
marcelveldt: @kieranhogg 

I have just added stream_metadata_update_callback to the streamdetails (models bump upcoming in the server)
you can register any async (instance) method there and it will be called at each X interval with the streamdetails object at first parameter and the played seconds as second parameter

the interval can be controlled with stream_metadata_update_interval

```
StreamDetails(
    blah=blah,
    ....
    stream_metadata=StreamMetadata(initial metadata...)
    stream_metadata_update_callback = self._update_stream_metadata,
    stream_metadata_update_interval = 5  # =5 seconds
)
```

and the actual callback:
```
async def _update_stream_metadata(self, streamdetails: StreamDetails, elapsed_time: int) -> None:
    # do check if metadata needs to be updated
    # if needed, simply update the metadata fields here
    streamdetails.stream_metadata.title = blah etc
```

elapsed_time in this case will be the actual elapsed time and not the "seconds streamed" which can be a bit ahead due to buffering.

Hope this is useful this way!

@OzGav also FYI as you wanted to do something similar for the Pandora provider I believe ?

PR #2569, opened 2025-10-29
Typing fixes for the Cache controller

OzGav:


music_assistant/controllers/cache.py
```diff
@@ -105,12 +105,11 @@ async def get(
                     cache object matches the checksum provided
         - default: value to return if no cache object is found
         """
+        assert self.database is not None
         assert key, "No key provided"
         if allow_bypass and BYPASS_CACHE.get():
             return default
         cur_time = int(time.time())
-        if checksum is not None and not isinstance(checksum, str):
-            checksum = str(checksum)
```
marcelveldt, line 113: this may not be removed - this ensures that if somehow non string data is sent for the checksum (e.g. a int) it will be converted to str

OzGav: When I asked Claude about that when I was going through this I got this which made sense to me?

> Ah, looking at the type hints, checksum is declared as str | None.
> So when you check if checksum is not None, mypy knows it must be a str at that point (since those are the only two options). The isinstance(checksum, str) check is always True, making not isinstance(checksum, str) always False - hence unreachable.
> You can just remove those lines since checksum is already typed as str | None and can't be anything else. If you want to ensure it's a string when not None, the type system already guarantees that.
> If the function might actually receive non-string values at runtime (despite the type hint), you'd need to change the type hint to str | None | Any or similar, but that's usually not the right approach.

What approach do you think is best in this situation?

marcelveldt: then either hunt the entire code base where this is being called with an integer (e.g. a timestamp) or change the type hint to str | int | None

This is about guarding for bad shit, not to be be typing purists.

OzGav: This whole typing thing weirds me out as an old C programmer :-) Python seems easier because it doesn't force you to think about types, but without mypy you're just deferring the problem. In C, the compiler made you deal with it immediately. In Python, you deal with it when suddenly there is a crash.

marcelveldt: haha, yeah sure but in this case I'm pretty sure some of the callers of this method are sending an int as checksum.
so we should either fix that or allow it. removing the logic from the code simply breaks it.

music_assistant/controllers/cache.py
```diff
@@ -162,10 +163,9 @@ async def set(
         - checksum: optional argument to store with the cache object
         - persistent: if True the cache object will not be deleted when clearing the cache
         """
+        assert self.database is not None
         if not key:
             return
-        if checksum is not None and not isinstance(checksum, str):
-            checksum = str(checksum)
```
marcelveldt, line 168: please leave this in but always convert the field to str

OzGav: I think I have got this right...

music_assistant/controllers/cache.py
```diff
@@ -148,7 +149,7 @@ async def set(
         expiration: int = DEFAULT_CACHE_EXPIRATION,
         provider: str = "default",
         category: int = 0,
-        checksum: str | None = None,
+        checksum: str | int | None = None,
```
marcelveldt:
```suggestion
        checksum: str | None = None,
```

checksum must always be str (the discussion was just about other places in the code where we might be sending this as an int)

music_assistant/controllers/cache.py
```diff
@@ -162,9 +163,10 @@ async def set(
         - checksum: optional argument to store with the cache object
         - persistent: if True the cache object will not be deleted when clearing the cache
         """
+        assert self.database is not None
         if not key:
             return
-        if checksum is not None and not isinstance(checksum, str):
+        if checksum is not None:
             checksum = str(checksum)
```
marcelveldt, line 170: yeah, this is fine like this - if some other (not yet fully typed) piece of the existing code now calls this with a non string value, we will simply convert it.

marcelveldt (approved):

PR #2570, opened 2025-10-29
Typing fixes for the config controller

OzGav:


music_assistant/controllers/config.py
```diff
@@ -362,7 +369,11 @@ async def save_provider_config(
         if instance_id is not None:
             config = await self._update_provider_config(instance_id, values)
         else:
-            config = await self._add_provider_config(provider_domain, values)
+            result = await self._add_provider_config(provider_domain, values)
+            if isinstance(result, list):
+                msg = "Unexpected return type from _add_provider_config"
+                raise TypeError(msg)
```
marcelveldt: ???

music_assistant/controllers/config.py
```diff
@@ -480,7 +491,7 @@ async def get_player_config_value(
         player_id: str,
         key: str,
         unpack_splitted_values: bool = False,
-    ) -> ConfigValueType:
+    ) -> ConfigValueType | tuple[str, ...] | list[tuple[str, ...]]:
```
marcelveldt, line 490: I don't think this return type is correct

music_assistant/controllers/config.py
OzGav: Yep sorry I ended up changing the return type for add_provider_config and forgot to come back and revert this.

music_assistant/controllers/config.py
OzGav: If unpack_splitted_values is False the function returns 
`conf.values[key].value if conf.values[key].value is not None else conf.values[key].default_value`

The value field is typed as ConfigValueType.

However, if unpack_splitted_values is True then the function returns:
`conf.values[key].get_splitted_values()`
This method returns
`def get_splitted_values(self) -> tuple[str, ...] | list[tuple[str, ...]]:`

marcelveldt: hmmm, yeah, its because of those spllitted values... hmm.

marcelveldt (approved):
Thanks, @OzGav !

PR #2574, opened 2025-10-29
Typing fixes for the Radio media controller

OzGav:
This is a small change for radio.py but mainly prepares pyproject.toml so the media controllers can be done one by one...


music_assistant/controllers/media/radio.py
```diff
@@ -52,10 +56,11 @@ async def versions(
             all_versions.pop(prov_version.item_id, None)
 
         # return the aggregated result
-        return all_versions.values()
+        return list(all_versions.values())
```
khers, line 59: I am sure there is a reason that dict.values() doesn't return a list and this had me digging around some but from what I see, this is the preferred solution. Not a request for a change, just an observation.

music_assistant/controllers/media/radio.py
```diff
@@ -52,10 +56,11 @@ async def versions(
...
 
-    async def _add_library_item(self, item: Radio) -> int:
+    async def _add_library_item(self, item: Radio, overwrite_existing: bool = False) -> int:
```
khers, line 61: Where is this new parameter used?

music_assistant/controllers/media/radio.py
```diff
@@ -52,10 +56,11 @@ async def versions(
...
         """Add a new item record to the database."""
+        assert self.mass.music.database is not None
```
khers: From what I can tell, asserts are disable-able at run time, is this intended for test/dev builds only?

music_assistant/controllers/media/radio.py
OzGav: It's not used in this method - it's just required to match the parent class signature. We could add something here to make that clear?

music_assistant/controllers/media/radio.py
OzGav: Its just for type checking. I dont think its possible for the database to not exist so thought an assert would be appropriate. I have added a comment to clarify? Thoughts?

music_assistant/controllers/media/radio.py
khers: Nope, I should have checked the base class for the signature. I think this is fine.

music_assistant/controllers/media/radio.py
khers: That sounds good.

khers (approved):
LGTM

PR #2580, opened 2025-11-01
Subsonic: Use loudness tags when present

khers:
The Open Subsonic API supports tagging items with loudness values and will report them, when present. If they are given to us, pass them onto MA.


music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -174,6 +174,17 @@ async def _run_async(
     ) -> RetType:
         return await asyncio.to_thread(call, *args, **kwargs)
 
+    def _set_loudness(self, item: SonicItem) -> None:
+        if item.replay_gain and item.replay_gain.track_gain and item.replay_gain.album_gain:
```
marcelveldt: album gain is an optional tag and will most likely not be set for a lot of tracks. Maybe make it a bit more loose and pass also if only the track _gain is present.

marcelveldt (approved):
LGTM
Thanks, @khers !

PR #2582, opened 2025-11-01
Refactor Smart fades

MarvinSchenkel:
This PR refactors Smart Fades into separate classes, such as Smart Fades and Filter classes.

What was done:
- Introduced base classes for Smart Fades and Filters, so that we can easily re-use filters and add new Smart fade modes
- Added the following filters:
    - TimeStretch
    - Trim
    - FrequencySweep
    - Crossfade
- Renamed the mode 'smart fades' to 'smart crossfade' to better represent what is is doing
- Refactored standard crossfade and smart crossfade to use the new base class


music_assistant/constants.py
```diff
@@ -310,12 +310,12 @@
     label="Enable Smart Fades",
     options=[
         ConfigValueOption("Disabled", "disabled"),
-        ConfigValueOption("Smart Fades", "smart_fades"),
+        ConfigValueOption("Smart Crossfade", "smart_crossfade"),
```
marcelveldt, line 313: this is now a breaking change

MarvinSchenkel: Within this beta round it is indeed a breaking change. Smart fades has not been released yet to stable. Do we want to auto migrate these kind of changes within the same beta round? Otherwise I'll add a migration script for this.

marcelveldt: I think so yes because the value is truly invalid and may result in errors

MarvinSchenkel: perfect, added the migration script.

marcelveldt:
some conflicts to rfesolve and then this is ready to merge

marcelveldt (approved):
Nice work, @MarvinSchenkel !

PR #2584, opened 2025-11-02
fix airplay not changing to idle

### LINKED ISSUE #4127: AirPlay player state does not change to idle after queue finished

2.5.5

Home Assistant OS Addon

- [x] Yes

### The problem

I was trying out Music Assistant Jukebox and discovered that it always started out with the same song, which was not queued by Jukebox. Turns out that when the last song is done playin, the `_handle_end_of_queue()` is executed, but the if statement that is supposed to call `_clear_queue_delayed()` evaluates to `false` due to `seconds_played` is 0.  

### How to reproduce

Queue up some songs and wait until all songs played through. The last song is still present in the queue.

### Music Providers

This is not related to the providers.

### Player Providers

Local, Airplay. 

### Full log output

The log does not show anything related to this. It is supposed to print `End of queue reached, clearing items`but does not.

### Additional information

None.

2025.7

Home Assistant OS

- [x] Yes

OzGav:
Fixes https://github.com/music-assistant/support/issues/4127

and maybe https://github.com/music-assistant/support/issues/4273 as my testing of that one showed an empty queue error which I think was triggering due to the starting state being PLAYING even though the queue is empty.

### DIFF
--- a/music_assistant/providers/airplay/stream_session.py
+++ b/music_assistant/providers/airplay/stream_session.py
@@ -345,6 +345,12 @@ class AirPlayStreamSession:
                     ],
                     return_exceptions=True,
                 )
+                # Set all players to IDLE state after queue finishes
+                # This ensures the state is updated even if the CLI binary doesn't
+                # reliably report "end of stream reached" via stderr
+                for sync_client in self.sync_clients:
+                    if sync_client.stream:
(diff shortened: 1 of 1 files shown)


marcelveldt:
this is already handled in a different way

PR #2585, opened 2025-11-02
 Fix for some HLS radio streams playback

### LINKED ISSUE #4322: Radio Italia doesn't work
### Discussed in https://github.com/orgs/music-assistant/discussions/4231

<div type='discussions-op-text'>

<sup>Originally posted by **fammastrodaiuto** September 11, 2025</sup>
Good evening,
I'm trying to play this radio station from Music Assistant, but it doesn't work.
If I go to TuneIN, it works fine; if I go to Music Assistant, click "Radio," and select "Radio Italia," I can't hear anything.
Link: https://tunein.com/radio/Radio-Italia-s295386/
Other radio stations, some I can hear, some I can't.
How can I fix this?
Thanks to anyone who will help me.</div>

OzGav:
Fixes https://github.com/music-assistant/support/issues/4322

  ### Summary

  This PR fixes a bug preventing certain HLS radio streams from playing, specifically those using unencrypted segments
  (METHOD=NONE). The issue caused an immediate stream abort with an IndexError during playlist parsing.

  ### Changes

###   1. Fixed HLS playlist parser to handle METHOD=NONE encryption (playlists.py)

  **Problem:**
  The M3U playlist parser assumed all #EXT-X-KEY: tags would contain a ,URI= parameter. However, per the
  https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.4, when METHOD=NONE is specified, no URI is present as the
  segments are unencrypted.

  Example of failing playlist:
  #EXT-X-KEY:METHOD=NONE
  #EXTINF:6,
  segment.ts

  **Fix:**
  - Check if METHOD=NONE is present and explicitly set key = None (per HLS spec, this clears any previous encryption)
  - Only attempt to extract URI when ,URI= is present in the line
  - Prevents IndexError: list index out of range crash

  Affected stream: https://radioitaliasmi.akamaized.net/hls/live/2093120/RISMI/stream01/streamPlaylist.m3u8

 **References**

  - https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.4 - Defines encryption key behavior including METHOD=NONE

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -672,6 +672,7 @@ async def get_media_stream(
         if isinstance(err, asyncio.CancelledError | GeneratorExit):
             # we were cancelled, just raise
             cancelled = True
+            raise
         logger.error("Error while streaming %s: %s", streamdetails.uri, err)
         # dump the last 10 lines of the log in case of an unclean exit
         logger.warning("\n".join(list(ffmpeg_proc.log_history)[-10:]))
--- a/music_assistant/helpers/playlists.py
+++ b/music_assistant/helpers/playlists.py
@@ -84,7 +84,12 @@ def parse_m3u(m3u_data: str) -> list[PlaylistItem]:
                 kev_value_parts = part.strip().split("=")
                 stream_info[kev_value_parts[0]] = kev_value_parts[1]
         elif line.startswith("#EXT-X-KEY:"):
-            key = line.split(",URI=")[1].strip('"')
+            # Extract encryption key URI if present
+            # METHOD=NONE means no encryption, so explicitly clear the key
+            if "METHOD=NONE" in line:
+                key = None
+            elif ",URI=" in line:
+                key = line.split(",URI=")[1].strip('"')
         elif line.startswith("#"):
             # Ignore other extensions
             continue


music_assistant/helpers/audio.py
marcelveldt, line 675: this is already fixed in dev, please rebase

OzGav: Fixed

marcelveldt (approved):
Looks good, thanks @OzGav !

fammastrodaiuto:
Good evening,
I apologize for not checking earlier, but unfortunately the problem is still present.
Radio Italia cannot be listened to via TuneIt-Radio.
I look forward to hearing from you.
I remain available for further clarification.
Thank you and have a good day,
Michael

fammastrodaiuto:
@marcelveldt  marcelveldt

OzGav:
Please dont ping people as we are busy enough without getting multiple message for the same thing. The URL you provided and that I showed in the PR comment works fine here on the lastest nightly. Make sure you are running the latest version of the MA server. I do see a problem with listening from TuneIn but that seems to be a problem with TuneIn not MA but if I get a chance I will look at it a bit more. For now, add the URL manually.

PR #2586, opened 2025-11-03
Typing fixes for the Webserver controller

OzGav:
Tested working locally. The problem that caused the webserver to crash after the last PR has been resolved with an update to the models.


music_assistant/controllers/webserver.py
```diff
@@ -130,7 +128,7 @@ async def get_config_entries(
     async def setup(self, config: CoreConfig) -> None:
         """Async initialize of module."""
         # work out all routes
-        routes: list[tuple[str, str, Awaitable]] = []
+        routes: list[tuple[str, str, Any]] = []
```
marcelveldt: this should probably be types as a callable returning a coroutine

marcelveldt (approved):
LGTM but needs good testing as last time this broke the entire webserver. Lets release in nightly so we can do a bit testing before it goes to beta

music_assistant/controllers/webserver.py
OzGav: There is getting beyond my level of knowledge but Claude says!
"I used Awaitable[web.StreamResponse] to match aiohttp's Handler type alias. Did you want the more specific Coroutine[Any, Any, web.StreamResponse] instead? Awaitable is the broader type that includes coroutines and other awaitables."

Thoughts?

marcelveldt: I think Awaitable[web.StreamResponse] is fine!

marcelveldt (approved):

OzGav:
Since the approval changes made to lines 281-283 and line 334 due to merge conflicts.

OzGav:
And yes the type fixes did break something! THIS DEVICE wasn't working due to a init signature mismatch with the core controller. This is now fixed. Changed lines 64 and 66

MarvinSchenkel:
Just did some quick testing and the webserver is looking fine at first sight. Let's do some more testing in the nightly

PR #2587, opened 2025-11-03
Add missing enforced flow mode config entry for Resonate

maximmaxim345:
With dc097bda407c Resonate no longer uses workarounds to directly access the queue flow stream.
This, however, requires a correct configuration to use the queue flow stream, which is fixed here.


marcelveldt (approved):
Thanks.
In the end I think we should handle this differently where we just let the queue controller decide to do flow mode if the player does not support enqueuing and we can have the config option to enforce flow mode. But that is for later.

PR #2589, opened 2025-11-03
Typing fixes for the squeezelite provider

OzGav:
Typing fixes and also changed runtimeerrors to the appropriate MA errors


music_assistant/providers/squeezelite/player.py
```diff
@@ -276,35 +286,42 @@ async def play_media(self, media: PlayerMedia) -> None:
...
         else:
             # assume url or some other direct path
             # NOTE: this will fail if its an uri not playable by ffmpeg
             audio_source = get_ffmpeg_stream(
                 audio_input=media.uri,
-                input_format=AudioFormat(ContentType.try_parse(media.uri)),
+                input_format=AudioFormat(
+                    content_type=ContentType.try_parse(media.uri) or ContentType.UNKNOWN
+                ),
                 output_format=master_audio_format,
```
marcelveldt: this whole block is outdated. rebase the branch to find out why

OzGav: OK I rebased but nothing showed up as changed?

OzGav: OK hang on I see Maxim did something...

marcelveldt: can you rebase again ?

OzGav: OK I think this is good now...

music_assistant/providers/squeezelite/player.py
```diff
@@ -251,16 +254,18 @@ async def play_media(self, media: PlayerMedia) -> None:
         # For queue streams, further cap to content sample rate
         if media.source_id and media.queue_item_id:
             queue_item = self.mass.player_queues.get_item(media.source_id, media.queue_item_id)
-            min_sample_rate = min(
-                min_sample_rate, queue_item.streamdetails.audio_format.sample_rate
-            )
+            if queue_item and queue_item.streamdetails and queue_item.streamdetails.audio_format:
+                min_sample_rate = min(
+                    min_sample_rate, queue_item.streamdetails.audio_format.sample_rate
+                )
```
marcelveldt: I really don't see why this is needed, I don't like the fact that a player provider is interacting directly with other core elements. Also it wont work because streamdetails may be known AFTER the play media command was issued. 

Why not simply pick one single flow format for the multi client stream ?
48/24 is a fine one.

And yes, that means that you cant play 192khz/24 on a group but that is nonsense anyways

OzGav: We certainly can do that. I was trying to have the same functionality as single players. The downside to the fixed format is a user will see one format for a single player but when grouped they might see it change.

marcelveldt: You can't have the same functionality as single players because grouped squeeze players use  flow mode by default so one fixed sample rate. Then we just pick a high but sane default sample rate by default. We can also set it to the maximum supported rate by all players. 

Note that it's technically possible to send individual streams even when grouped but that is an awful amount of work with less to gain, except maybe for the audio purists but they probably don't use grouped playback sessions for critical listening anyways

OzGav: Ok setting it to the maximum of all players is an option I considered at one point but was considering the CPU cost on smaller servers of up scaling unnecessarily? So maybe the 48/24 is the practicable solution….

OzGav: Then again setting the max supported by all would probably keep the most people happy… I think…

marcelveldt: Can we settle for a sane default then ? 96/24 ?

OzGav: OK I think this is settled!

marcelveldt (approved):
Thanks, @OzGav !

PR #2598, opened 2025-11-04
fix: abs - discarded progress

fmunkes:
Fixes a discarded progress in abs not being synced to MA, see music-assistant/support#4379. If a user discards an abs progress, the api does not return anything, and no socket event is emitted. So we need to compare items in MA's playlog with items still present in abs.

This PR adds a core function to obtain in-progress media item ids and types from the database. The abs provider uses this function to cleanup the playlog on regular provider syncs if necessary.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -517,6 +517,41 @@ class MusicController(CoreController):
             )
         return result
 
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
LGTM
A bit unfortunate that we have to deal with it this complex, but it is what it is. At least it's now user friendly!

PR #2602, opened 2025-11-06
Fix deadlock when trying to join a paused Airplay player to a group

MarvinSchenkel:


music_assistant/providers/airplay/player.py
```diff
@@ -515,6 +515,12 @@ async def set_members(
             if child_player_to_add := cast(
                 "AirPlayPlayer | None", self.mass.players.get(player_id)
             ):
+                if (
+                    child_player_to_add.playback_state == PlaybackState.PAUSED
+                    and child_player_to_add.stream
+                ):
+                    # Stop the paused stream to avoid a deadlock situation
+                    await child_player_to_add.stream.stop()
```
marcelveldt, line 523: why not unpause it ?

MarvinSchenkel: When I play Song X on Player A and then pause the player, player A becomes in a paused state.

Then, when I play Song Y on Player B and want to group Player A, this results in a deadlock. 

Un pausing is not an option here as it would continue to player Song X which is not what we want.

marcelveldt (approved):

PR #2605, opened 2025-11-07
fix: abs: another in progress fix

fmunkes:
Another abs progress fix - turns out, there is also the possibility of a progress being returned by the API, but the time set to 0. Fixed in this PR. Also discovered here: music-assistant/support#4379


marcelveldt (approved):
LGTM
Thanks, @fmunkes !

PR #2608, opened 2025-11-07
Plex: added remote control feature

anatosun:
# Plex Connect: new plugin for Plex discovery and control
This PR introduces Plex Connect, a new standalone plugin that enables Music Assistant players to be discovered and controlled by Plex clients like Plexamp when using Plex as a music source.

## Functionalities
Plex Connect acts as a bridge between your Plex server and Music Assistant players, allowing Plex clients like Plexamp to discover and control your MA players. This plugin requires the Plex music provider to be configured as it works specifically with Plex music sources.

Music Assistant players visible in Plexamp

### Setup:

- Enable Plex Connect plugin in Music Assistant
- Configure the plugin for each player you want to control via Plex clients

### Discovery & Connection:

- Plugin creates a dedicated remote control server for each enabled player
- Players are automatically advertised on the local network via GDM (Good Day Mate) broadcasts
- Plex clients (like Plexamp) discover these players as native Plex players - no manual configuration needed

### Control & Features:

- Control your Music Assistant players directly from any Plex client
- Standard playback controls: play, pause, skip, seek
- Queue management: add tracks, reorder, clear queue
- Access Plex advanced features: Sonic Adventures, Radio, DJs, Mixes

### Synchronization:

- Real-time state synchronization using Plex's timeline and navigation protocols
- Playback state, current track, progress, and queue stay synchronized
- All playback activity reported to your Plex server
- Listening history and scrobbles visible in Plex dashboard
- Enables "Continue Listening" and other Plex features

# Comments
-  The timeline report to the Plex server is now done on a per-player basis, the logic has been moved to `player_remote.py`.
- This feature, although tested extensively before the PR, needs testing by other fellow users!


music_assistant/providers/plex/__init__.py
```diff
@@ -295,6 +298,31 @@ async def get_config_entries(  # noqa: PLR0915
...
         entries.append(conf_libraries)
 
+    # remote control configuration entries
+    entries.extend(
+        [
+            ConfigEntry(
+                key=CONF_REMOTE_CONTROL_ENABLED,
+                type=ConfigEntryType.BOOLEAN,
+                label="Enable remote control",
+                description="Enable Plex remote control for Music Assistant players",
```
marcelveldt: Should this maybe be renamed to "Enable PlexAmp support" to better reflect what it does ?

music_assistant/providers/plex/__init__.py
```diff
@@ -295,6 +298,31 @@ async def get_config_entries(  # noqa: PLR0915
...
+                required=False,
+                default_value=False,
+                category="advanced",
+            ),
+            ConfigEntry(
```
marcelveldt: why would this be needed to be configured ?

music_assistant/providers/plex/gdm.py
```diff
@@ -0,0 +1,217 @@
...
+        return "\r\n".join(message_lines).encode("utf-8")
+
+    def start(self) -> None:
+        """Start GDM advertising and listening."""
+        if self._running:
+            return
+        self._running = True
+
+        # Cache local IP
+        self._local_ip = get_local_ip()
```
marcelveldt: I would instead use the mass.streams.publish_ip property because that will hold the value the user selected as the IP on the internal network, in case of multi homed network setups

music_assistant/providers/plex/helpers.py
```diff
@@ -79,3 +80,26 @@ def _discover_local_servers() -> tuple[str, int] | tuple[None, None]:
...
+        s.close()
+        return local_ip
+    except Exception:
+        # Fallback: try to get IP from hostname
+        try:
+            hostname = socket.gethostname()
+            return socket.gethostbyname(hostname)
+        except Exception:
+            # If all else fails, return localhost
+            return "127.0.0.1"
```
marcelveldt: please drop this helper. First of all because we already have this helper present and secondly because it would be better to use the mass.streams.publish_ip

music_assistant/providers/plex/player_remote.py
```diff
@@ -0,0 +1,1696 @@
...
+if TYPE_CHECKING:
+    from music_assistant_models.event import MassEvent
+
+    from music_assistant.providers.plex import PlexProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PlexPlayerManager:
```
marcelveldt: Reading all this code I'm almost starting to think this belongs in its own Plugin provider as this has strictly nothing to do with the Plex music provider but instead its creating a fake/virtual plex server on the network to trick plex clients.

music_assistant/providers/plex/player_remote.py
```diff
@@ -0,0 +1,1696 @@
...
+
+        # Track MA queue state to detect when we need to sync to Plex
+        self._last_synced_ma_queue_length: int = 0
+        self._last_synced_ma_queue_keys: list[str] = []
+
+        # Specific MA player this server controls (set by PlayerRemoteInstance)
+        self._ma_player_id = ma_player_id
+
+        # Shared HTTP client for timeline broadcasts
+        self._http_client: aiohttp.ClientSession | None = None
```
marcelveldt: let's try to always use the shared hrtp client that lives in mass.http_session and mass.http_session_no_ssl

music_assistant/providers/plex_connect/player_remote.py
```diff
@@ -0,0 +1,1696 @@
...
+            "X-Plex-Platform": "Music Assistant",
+            "X-Plex-Platform-Version": platform.release(),
+        }
+
+        self._setup_routes()
+
+    def _setup_routes(self) -> None:
+        """Set up all required endpoints."""
+        # Root endpoint
+        self.app.router.add_get("/", self.handle_root)
```
marcelveldt, line 187: why not host these routes on the streamserver, which allows you to register dynamic routes instead of hosting yet another webserver ?

mass.streams.register_dynamic_route (returns callback to unregister)

marcelveldt: Or are these endpoints set in stone ? They cant be prefixed or something ?

music_assistant/providers/plex_connect/player_remote.py
```diff
@@ -0,0 +1,1696 @@
...
+        if self._ma_player_id:
+            self._unsub_callbacks.append(
+                self.provider.mass.subscribe(self._handle_player_event, EventType.PLAYER_UPDATED)
+            )
+            self._unsub_callbacks.append(
+                self.provider.mass.subscribe(self._handle_queue_event, EventType.QUEUE_UPDATED)
+            )
+            self._unsub_callbacks.append(
+                self.provider.mass.subscribe(self._handle_queue_event, EventType.QUEUE_TIME_UPDATED)
+            )
```
marcelveldt, line 253: you can just subscribe to player_updated nowadays. we have updated the player model and player.current_media will now always contain the translated metadata from any source

marcelveldt:
As stated in the comments, my first feeling is that this is a separate, dedicated plugin provider and has little to do with the Plex music provider. It would be good to clarify exactly what this does and who can use it. Like for example, will this work regardless if you are an actual plex user ?

music_assistant/providers/plex_connect/player_remote.py
anatosun: Hosting those routes make it possible to receive commands from any Plex player. You can't host those routes on the stream servers since it's not "aware" of mass.

anatosun:
First let met thank you for reviewing the PR.

I understand your concern, my main motivation for making it part of the plex provider was to avoid duplicating Plex credentials. 

Looking at the amount of code this PR adds, I can agree that this should become its own separate plugin.

My question now is, could the `Plex Connect Plugin` use the Plex provider directly?

anatosun:
@marcelveldt 
I have refactored this functionality into its own separate plugin. Should I proceed by modifying the existing PR or doing a new one? I plan to move the playback report to the plugin, as it allows Plex to be aware of Music assistant players...

marcelveldt:
Does the plugin work without a Plex server? Does it just mimic a Plex server so you can allow any MA player with any MA music source?
Then it's simply a standalone plugin.

If it still relies on a Plex server, then you can can choose either between including it in the Plex music provider (while keeping the code maintainable) or isolate it into its own plugin. You can even add a "depends_on" in the manifest of the plugin to make it depend on Plex music provider.

marcelveldt:
> @marcelveldt 
> 
> I have refactored this functionality into its own separate plugin. Should I proceed by modifying the existing PR or doing a new one? I plan to move the playback report to the plugin, as it allows Plex to be aware of Music assistant players...

I have no preference. That is completely up to you

anatosun:
... and the [documentation](https://github.com/music-assistant/beta.music-assistant.io/pull/51)

music_assistant/providers/plex_connect/player_remote.py
anatosun: Is it still possible to subscribe to separate events?

marcelveldt: yes but that shouldnt be needed anymore as PLAYER_UPDATED is a catch-all. it will also update the player's current_media property if the queue track changed. But that is up to you. This is also fine

marcelveldt (approved):
Looks good to me.
Indeed good that you made this a dedicated provider, much better now!

music_assistant/providers/plex_connect/player_remote.py
```diff
@@ -0,0 +1,1725 @@
...
+            for mapping in item.media_item.provider_mappings:
+                if mapping.provider_instance == self.provider.instance_id:
+                    plex_key = mapping.item_id
+                    break
+
+            if plex_key:
+                try:
+                    # Fetch Plex item
+                    def fetch_item(key: object = plex_key) -> object:
+                        return self.plex_server.fetchItem(key)
```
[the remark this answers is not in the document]
marcelveldt: yeah, copilot is right here

marcelveldt:
Just have a look at the last copilot suggestions, some of them actually do make sense

anatosun:
Thank you, I will have a closer look at Copilot's suggestions later this week

OzGav:
Does this satisfy this feature request? https://github.com/orgs/music-assistant/discussions/3992

anatosun:
Yes, this PR addresses the feature request! It doesn't implement Plexamp Headless directly (which would be resource-heavy and require running the proprietary Plexamp software), but achieves the same end result.

anatosun:
I've implemented the necessary changes suggested by copilot. This PR is good to go!

OzGav:
Nice. If you can just resolve all the conversations with something short if its implemented or a short description as to why you chose not to.

music_assistant/providers/plex_connect/player_remote.py
```diff
@@ -0,0 +1,1725 @@
...
+        stale_clients = [
+            client_id
+            for client_id, sub in self.subscriptions.items()
+            if current_time - float(sub["last_update"]) > 90  # type: ignore[arg-type]
+        ]
+        for client_id in stale_clients:
+            del self.subscriptions[client_id]
+
+        for client_id in list(self.subscriptions.keys()):
+            await self._send_timeline(client_id)
```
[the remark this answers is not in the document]
anatosun: done

music_assistant/providers/plex_connect/player_remote.py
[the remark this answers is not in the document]
anatosun: implemented

anatosun:
Let me write this up here, this is simpler for me.

After reviewing all Copilot AI suggestions, 6 out of 17 recommendations should be implemented: two critical high-priority fixes (the lambda closure bug causing incorrect variable capture and improper queue ID mismatch handling that could corrupt state), and four medium-priority improvements (concurrent timeline broadcasts for better multi-client performance, robust error handling for stale client cleanup, more descriptive error messages, and parallel Plex item fetching for large queues). The remaining 11 suggestions were rejected because they either add unnecessary complexity without proven benefit (race condition locks, threading events), break intentional design decisions (port allocation starting at 32500 for Plex compatibility, shuffle disabling to prevent loops), address false-positive linter warnings, represent premature optimizations, or are cleanup items better handled separately (commented debug code, documentation links). This results in meaningful improvements to robustness and performance without bloating the codebase with speculative fixes.

music_assistant/providers/plex_connect/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "plugin",
+  "stage": "alpha",
+  "domain": "plex_connect",
+  "name": "Plex Connect Plugin",
+  "description": "Makes a Music Assistant player appear as a device in the official Plex apps.",
+  "codeowners": ["@anatosun"],
+  "requirements": ["plexapi==4.17.1"],
+  "icon": "mdi:plex",
+  "documentation": "https://music-assistant.io/plugin-support/plex_connect/",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/plugins/plex-connect/",
```

MarvinSchenkel:
Awesome job @anatosun <3

remon1496:
Legend! @anatosun

moonbit5:
Absolutely brilliant, thanks so much @anatosun

jmurph44:
Thank you for taking the time to make this @anatosun

PR #2611, opened 2025-11-07
Typing fixes for the player queues controller

OzGav:


music_assistant/controllers/player_queues.py
```diff
@@ -387,14 +393,14 @@ async def play_media(
             raise PlayerUnavailableError(f"Queue {queue_id} is not available")
         # always fetch the underlying player so we can raise early if its not available
         queue_player = self.mass.players.get(queue_id, True)
-        assert queue_player is not None  # for type checking
+        if queue_player is None:
+            raise PlayerUnavailableError(f"Player {queue_id} is not available")
```
marcelveldt: this is a rather useless change because self.mass.players.get will already raise if the player is not available

music_assistant/controllers/player_queues.py
```diff
@@ -406,15 +412,25 @@ async def play_media(
...
-                    media_item = media_from_dict(item)
                 else:
+                    # item is MediaItemType | ItemMapping at this point
                     media_item = item
+
+                # Resolve ItemMapping early
+                if isinstance(media_item, ItemMapping):
+                    # uri is always set in __post_init__, type hint should be str
+                    assert media_item.uri is not None
+                    media_item = await self.mass.music.get_item_by_uri(media_item.uri)
```
marcelveldt: let's not do this - this is a potentially heavy call.
Please keep these PR's to typing changes only

marcelveldt (changes requested):
I see changes that are not strictly typing. This has caused more often to unexpected surprises. One step at a time please

music_assistant/controllers/player_queues.py
OzGav: Noted. Reverted.

music_assistant/controllers/player_queues.py
OzGav: This was actually moved from resolve_media_items but I can make the type fixes without this logic change so I’ll do that shortly and revert this section

music_assistant/controllers/player_queues.py
```diff
@@ -406,18 +411,22 @@ async def play_media(
...
+                    # item is MediaItemType | ItemMapping at this point
                     media_item = item
+
                 # Save requested media item to play on the queue so we can use it as a source
                 # for Don't stop the music. Use FIFO list to keep track of the last 10 played items
-                if media_item.media_type in (
+                # Skip ItemMapping and BrowseFolder - they don't have media_type yet
+                if not isinstance(
+                    media_item, (ItemMapping, BrowseFolder)
+                ) and media_item.media_type in (
```
marcelveldt: this is still changing logic

music_assistant/controllers/player_queues.py
```diff
@@ -426,21 +435,32 @@ async def play_media(
                     queue.enqueued_media_items.append(media_item)
                     if len(queue.enqueued_media_items) > 10:
                         queue.enqueued_media_items.pop(0)
+
                 # handle default enqueue option if needed
                 if option is None:
-                    option = QueueOption(
-                        await self.mass.config.get_core_config_value(
+                    # Only MediaItemType has media_type attribute
```
marcelveldt: that is not true

marcelveldt (dismissed):
sorry, there are still logic changes here and not pure typing fixes.

music_assistant/controllers/player_queues.py
OzGav: Fixed

music_assistant/controllers/player_queues.py
OzGav: OK I think I have my head around ItemMapping and BrowseFolder now...at least enough to fix my mistakes here...I hope...

music_assistant/controllers/player_queues.py
```diff
@@ -406,18 +411,22 @@ async def play_media(
...
         radio_source: list[MediaItemType] = []
         # resolve all media items
-        for item in media:
+        for item in media_list:
             try:
                 # parse provided uri into a MA MediaItem or Basic QueueItem from URL
+                media_item: MediaItemType | ItemMapping | BrowseFolder
                 if isinstance(item, str):
                     media_item = await self.mass.music.get_item_by_uri(item)
-                elif isinstance(item, dict):
```
MarvinSchenkel, line 414: Are we sure this can be safely removed?

music_assistant/controllers/player_queues.py
```diff
@@ -453,14 +475,21 @@ async def play_media(
...
-        queue_items = [
-            QueueItem.from_media_item(queue_id, x) for x in media_items if x and x.available
-        ]
+        # _resolve_media_items should only return playable types, but we filter to be safe
+        queue_items: list[QueueItem] = []
+        for x in media_items:
+            if not x or not x.available:
+                continue
+            # Only playable media types can be added to queue
+            if not isinstance(x, (Track, Radio, Audiobook, PodcastEpisode)):
```
MarvinSchenkel: I think this is an unwanted side effect and we should rely on _resolve_media_items to only provide us with playable items. Now we basically put this responsibility in two places.

music_assistant/controllers/player_queues.py
```diff
@@ -1644,29 +1784,43 @@ async def _resolve_media_items(
         """Resolve/unwrap media items to enqueue."""
         # resolve Itemmapping to full media item
         if isinstance(media_item, ItemMapping):
+            if media_item.uri is None:
+                raise InvalidDataError("ItemMapping has no URI")
             media_item = await self.mass.music.get_item_by_uri(media_item.uri)
+
+        # Skip BrowseFolder items
+        if isinstance(media_item, BrowseFolder):
```
MarvinSchenkel: This needs to be removed. This will introduce a bug where playing a folder will result in 'no playable items found'

music_assistant/controllers/player_queues.py
```diff
@@ -1129,22 +1199,35 @@ async def _load_item(
...
+                and queue_item.media_item is not None
+                and hasattr(queue_item.media_item, "album")
+                and queue_item.media_item.album is not None
                 and queue_item.media_item.album.item_id == previous_item.media_item.album.item_id
             )
-        )
         playing_album_tracks = next_track_from_same_album or previous_track_from_same_album
         if queue_item.media_item and queue_item.media_item.media_type == MediaType.TRACK:
+            # Type narrowing with proper error handling
+            if not isinstance(queue_item.media_item, Track):
```
MarvinSchenkel: This type narrowing does not make sense to me. If mypy complains, we could simply replace the if:
`if queue_item.media_item and queue_item.media_item.media_type == MediaType.TRACK:`
to
`if queue_item.media_item and isinstance(queue_item.media_item, Track):`

music_assistant/controllers/player_queues.py
```diff
@@ -1153,13 +1236,31 @@ async def _load_item(
                 queue_item.media_item.item_id,
                 queue_item.media_item.provider,
             ):
-                queue_item.media_item = library_item
+                if not isinstance(library_item, Track):
```
MarvinSchenkel: Do we really need to guard this? Have there been occurrences where the library returns unexpected media types?

music_assistant/controllers/player_queues.py
```diff
@@ -1153,13 +1236,31 @@ async def _load_item(
...
+                else:
+                    queue_item.media_item = library_item
             elif not queue_item.media_item.image or queue_item.media_item.provider.startswith(
                 "ytmusic"
             ):
                 # Youtube Music has poor thumbs by default, so we always fetch the full item
                 # this also catches the case where they have an unavailable item in a listing
-                queue_item.media_item = await self.mass.music.get_item_by_uri(queue_item.uri)
+                fetched_item = await self.mass.music.get_item_by_uri(queue_item.uri)
+                if not isinstance(fetched_item, Track):
```
MarvinSchenkel: Do we really need to guard this? Have there been occurrences where the library returns unexpected media types?

music_assistant/controllers/player_queues.py
```diff
@@ -1168,25 +1269,34 @@ async def _load_item(
                     album.provider,
                 )
             ):
-                queue_item.media_item.album = library_album
+                if not isinstance(library_album, Album):
```
MarvinSchenkel: Do we really need to guard this? Have there been occurrences where the library returns unexpected media types?

OzGav: Not that I'm aware. Switched to an assert for mypy

music_assistant/controllers/player_queues.py
OzGav: Not that I'm aware. Switched to an assert for mypy

music_assistant/controllers/player_queues.py
OzGav: Not that I'm aware. Switched to an assert for mypy

music_assistant/controllers/player_queues.py
OzGav: Yep simple is better

music_assistant/controllers/player_queues.py
OzGav: Yep quite right. Removed.

music_assistant/controllers/player_queues.py
OzGav: OK. Using cast instead

music_assistant/controllers/player_queues.py
OzGav: After reviewing the API code, I discovered that a dict can indeed be passed to this function. When the API receives a request from a client, media items are sent as plain dictionaries (JSON data) that need to be converted into proper MediaItem objects using media_from_dict.
The mypy issue was that dict[str, Any] was missing from the function signature.

music_assistant/controllers/player_queues.py
```diff
@@ -365,7 +370,11 @@ def set_repeat(self, queue_id: str, repeat_mode: RepeatMode) -> None:
     async def play_media(
```
MarvinSchenkel, line 370: There's still something going wrong when trying to play music (an error is thrown). Looks like we need to change the ordering of the signature to:
```
async def play_media(
        self,
        queue_id: str,
        media: list[MediaItemType | ItemMapping | str | dict[str, Any]]
        | MediaItemType
        | ItemMapping
        | str
        | dict[str, Any],
        option: QueueOption | None = None,
        radio_mode: bool = False,
        start_item: PlayableMediaItemType | str | None = None,
    ) -> None:
```

Let's do a bit more of testing on this branch before merging.

OzGav: You are right. Sorry I was too rushed yesterday. After a bunch of testing across all media types I found one other issue which has been fixed. From my testing all is well but you will see some ignores in there now as I believe the actual mypy fix attempt was causing problems because the real issue is in api.py and I have added a TODO to look into that in the future (I have a couple of those types of TODOs added across these mypy PRs). 

In this case parse_value returns ANY which can let dicts slip through and then further down there is a block with the comment `# try if the value is a dict and we can convert it` but this will just return the dict if it has problems! Thats what I think anyway but I could be wrong!

music_assistant/controllers/player_queues.py
```diff
@@ -1153,13 +1225,17 @@ async def _load_item(
                 queue_item.media_item.item_id,
                 queue_item.media_item.provider,
             ):
+                assert isinstance(library_item, Track)
                 queue_item.media_item = library_item
```
MarvinSchenkel:
```suggestion
                queue_item.media_item = cast("Track", library_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1168,25 +1244,28 @@ async def _load_item(
                     album.provider,
                 )
             ):
+                assert isinstance(library_album, Album)
                 queue_item.media_item.album = library_album
```
MarvinSchenkel:
```suggestion
                queue_item.media_item.album = cast("Album", library_album)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1153,13 +1225,17 @@ async def _load_item(
...
                 queue_item.media_item = library_item
             elif not queue_item.media_item.image or queue_item.media_item.provider.startswith(
                 "ytmusic"
             ):
                 # Youtube Music has poor thumbs by default, so we always fetch the full item
                 # this also catches the case where they have an unavailable item in a listing
-                queue_item.media_item = await self.mass.music.get_item_by_uri(queue_item.uri)
+                fetched_item = await self.mass.music.get_item_by_uri(queue_item.uri)
+                assert isinstance(fetched_item, Track)
+                queue_item.media_item = fetched_item
```
MarvinSchenkel:
```suggestion
                queue_item.media_item = cast("Track", fetched_item)
```

MarvinSchenkel: \+ get rid of the asserts. I think after that we are good to merge :)

music_assistant/controllers/player_queues.py
OzGav: Fixed

music_assistant/controllers/player_queues.py
OzGav: Fixed

music_assistant/controllers/player_queues.py
OzGav: Thanks, casts added and asserts removed.

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
         """Resolve/unwrap media items to enqueue."""
         # resolve Itemmapping to full media item
         if isinstance(media_item, ItemMapping):
+            if media_item.uri is None:
+                raise InvalidDataError("ItemMapping has no URI")
             media_item = await self.mass.music.get_item_by_uri(media_item.uri)
         if media_item.media_type == MediaType.PLAYLIST:
+            assert isinstance(media_item, Playlist)
```
MarvinSchenkel:
```suggestion
            media_item = cast("Playlist", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
             self.mass.create_task(self.mass.music.mark_item_played(media_item))
-            return await self.get_playlist_tracks(media_item, start_item)
+            return list(await self.get_playlist_tracks(media_item, start_item))
         if media_item.media_type == MediaType.ARTIST:
+            assert isinstance(media_item, Artist)
```
MarvinSchenkel:
```suggestion
            media_item = cast("Artist", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
             self.mass.create_task(self.mass.music.mark_item_played(media_item))
-            return await self.get_artist_tracks(media_item)
+            return list(await self.get_artist_tracks(media_item))
         if media_item.media_type == MediaType.ALBUM:
+            assert isinstance(media_item, Album)
```
MarvinSchenkel:
```suggestion
            media_item = cast("Album", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
             self.mass.create_task(self.mass.music.mark_item_played(media_item))
-            return await self.get_album_tracks(media_item, start_item)
+            return list(await self.get_album_tracks(media_item, start_item))
         if media_item.media_type == MediaType.AUDIOBOOK:
+            assert isinstance(media_item, Audiobook)
```
MarvinSchenkel:
```suggestion
            media_item = cast("Audiobook", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
             # ensure we grab the correct/latest resume point info
             media_item.resume_position_ms = await self.get_audiobook_resume_point(
                 media_item, start_item
             )
             return [media_item]
         if media_item.media_type == MediaType.PODCAST:
+            assert isinstance(media_item, Podcast)
```
MarvinSchenkel:
```suggestion
            media_item = cast("Podcast", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
             self.mass.create_task(self.mass.music.mark_item_played(media_item))
-            return await self.get_next_podcast_episodes(media_item, start_item)
+            return list(await self.get_next_podcast_episodes(media_item, start_item))
         if media_item.media_type == MediaType.PODCAST_EPISODE:
-            return await self.get_next_podcast_episodes(None, media_item)
+            assert isinstance(media_item, PodcastEpisode)
```
MarvinSchenkel:
```suggestion
            media_item = cast("PodcastEpisode", media_item)
```

music_assistant/controllers/player_queues.py
```diff
@@ -1644,41 +1750,54 @@ async def _resolve_media_items(
...
+            return list(await self.get_next_podcast_episodes(None, media_item))
         if media_item.media_type == MediaType.FOLDER:
-            return await self._get_folder_tracks(media_item)
+            assert isinstance(media_item, BrowseFolder)
```
MarvinSchenkel:
```suggestion
            media_item = cast("BrowseFolder", media_item)
```

MarvinSchenkel (approved):
Thanks @OzGav

PR #2614, opened 2025-11-09
Spotify streaming improvement

OzGav:
## Reliability Improvement  
- Add retry logic to `stream_spotify_uri()` (up to 2 attempts)
- Add timeout on first chunk to detect librespot startup failures early
- Conditional stderr capture (only on verbose mode or final attempt)

### DIFF
--- a/music_assistant/providers/spotify/streaming.py
+++ b/music_assistant/providers/spotify/streaming.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-from collections import deque
 from collections.abc import AsyncGenerator
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


marcelveldt:
Sorry, as said before, I don't want this logic to be added. Why does this need to be added, please explain

marcelveldt:
Note that the logging has already been added and librespot itself will raise early on failure

OzGav:
Ok sorry I thought you meant move it to its own PR. Spotify used to have retry logic so I guess I was moving back to that and the error logging I didn’t know about. So I will close this.

marcelveldt:
Sorry about the confusion then!

PR #2615, opened 2025-11-09
Use secure connection to MusicBrainz API

mtdcr:
Currently, plain HTTP is used to connect to MusicBrainz, which isn't state of the art and imposes both privacy and security risks.

To quote https://musicbrainz.org/doc/MusicBrainz_API:
> The API root URL is https://musicbrainz.org/ws/2/.

So this PR switches to the documented and secure HTTPS URL.

I didn't observe any errors after the patch and verified that Music Assistant was indeed connecting to musicbrainz.org:443.


marcelveldt:
Not so sure about the privacy and security risks here but it never hurts. I wasn't even aware we were still using the http endpoint

marcelveldt (approved):

PR #2625, opened 2025-11-11
Typing fixes for artists controller

OzGav:


music_assistant/controllers/media/artists.py
marcelveldt: We should maybe consider changing the return type to just list (to be consistent with other methods) and if we have places where we want to deduplicate, use a temporary UniqueList in between

music_assistant/controllers/media/artists.py
```diff
@@ -280,6 +288,7 @@ async def _add_library_item(self, item: Artist | ItemMapping) -> int:
         if item.mbid == VARIOUS_ARTISTS_MBID:
             item.name = VARIOUS_ARTISTS_NAME
         # no existing item matched: insert item
+        assert self.mass.music.database is not None  # for type checking
```
marcelveldt: instead of doing 99999 asserts for a value that will always be there, we should consider making mass.music.database a property with the type check in there and the actual db conn as private variable (which we return in the property)

OzGav: OK I think (hope!) I have done what you are thinking here https://github.com/music-assistant/server/pull/2629/files

music_assistant/controllers/media/artists.py
OzGav: I can do. From a quick test it looks like it will require a one line change in the builtin provider as well.

OzGav: I have made that change. Should I do the same for the albums function as well?

music_assistant/controllers/media/artists.py
MarvinSchenkel: Yes that's exactly right! Now we can get rid of these asserts on mass.music.database in all the other PRs as well.

music_assistant/controllers/media/artists.py
OzGav: all done

music_assistant/controllers/media/artists.py
OzGav: all done

MarvinSchenkel (approved):
Nice job Gav!

PR #2627, opened 2025-11-11
Typing fixes for the podcasts controller

OzGav:


music_assistant/controllers/media/podcasts.py
```diff
@@ -32,12 +33,13 @@ class PodcastsController(MediaControllerBase[Podcast]):
...
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, mass: MusicAssistant) -> None:
         """Initialize class."""
-        super().__init__(*args, **kwargs)
+        super().__init__(mass)
         # register (extra) api handlers
         api_base = self.api_base
-        self.mass.register_api_command(f"music/{api_base}/podcast_episodes", self.episodes)
+        # TODO: This is a type mismatch that requires a logic change to fix properly
+        self.mass.register_api_command(f"music/{api_base}/podcast_episodes", self.episodes)  # type: ignore[arg-type]
```
MarvinSchenkel: @marcelveldt 
Can we just simply update the `handler` parameter of `mass.register_api_command` to `handler: Callable[..., Coroutine[Any, Any, Any] | AsyncGenerator[Any, Any, Any]]` ? This will also solve it for other controllers with AsyncGenerators such as the PlaylistController.

marcelveldt: Almost... it will still be a callable but it returns a generator

MarvinSchenkel: Then my type is correct I think? It's a `Callable` that either returns a `Coroutine` or an `AsyncGenerator` (note the second `]]`). The question here is if that makes sense to you to type `handler` this way. If so, we can implement it as such and get rid of these mypy errors in a more central place.

MarvinSchenkel: @OzGav pushed a fix for this to dev. Please sync with dev and remove the `type: ignore`, should work now.

OzGav: Fixed

MarvinSchenkel (approved):

PR #2632, opened 2025-11-12
Types fixes for the Albums Controller

OzGav:


MarvinSchenkel:
@marcelveldt

Just confirming based on your comments on the other PRs:
- Media Controller functions should just return `list[xxx]` instead of `UniqueList[xx]` and use a `UniqueList` in the function body to deduplicate whenever required.

I quite like this since this makes typing easy and trivial everywhere while still maintaining the functionality.

MarvinSchenkel (approved):
Thanks Gav!

PR #2633, opened 2025-11-12
Typing fixes for the base media controller

OzGav:


music_assistant/controllers/media/base.py
```diff
@@ -312,19 +317,19 @@ async def search(
         )
         match self.media_type:
             case MediaType.ARTIST:
-                return searchresult.artists
+                return [cast("ItemCls", x) for x in searchresult.artists]
```
marcelveldt: I'm not particularly liking the overhead of this. why isnt this just casted to a list of ItemCls ?

music_assistant/controllers/media/base.py
```diff
@@ -338,19 +343,22 @@ async def get_provider_mapping(self, item: ItemCls) -> tuple[str, str]:
...
+                provider = cast("MusicProvider", provider)
+                if prefer_unique and provider.is_streaming_provider:
+                    continue
+                return (prov_mapping.provider_instance, prov_mapping.item_id)
         # last resort: return just the first entry
         for prov_mapping in item.provider_mappings:
             return (prov_mapping.provider_domain, prov_mapping.item_id)
 
-        return (None, None)
+        return ("", "")
```
marcelveldt: shouldnt we just raise something here instead of returning magic values?

music_assistant/controllers/media/base.py
OzGav: Fixed.

music_assistant/controllers/media/base.py
OzGav: I have pushed the change but also I searched the codebase and get_provider_mapping doesn't appear to be used anywhere. Should I just remove it entirely?

MarvinSchenkel: Indeed does not seem to be used anywhere. Let's clean it up.

OzGav: Removed

MarvinSchenkel (approved):

PR #2634, opened 2025-11-12
Install JavaScript runtime

### LINKED ISSUE #4348: Youtube Music status: 🔴
Youtube Music is an experimental Music provider. Google does not provide an official API integration for us to integrate with, so this provider is built on a best-effort basis. We had to jump through A LOT of hoops in order to this to work.

Google is constantly changing things on their end to purposely make it harder for people to pirate videos / music. While we are against piracy, we have to use similar techniques in order to make streaming music from YT Music possible. This means that Music Assistant is also directly affected by Google's changes, now and then breaking the YT Music provider.

Music Assistant has implemented a method to automatically recover from these errors if our upstream friends have implemented a fix. A Music Assistant release is not required. Whenever a fix is available, simply reloading the Youtube Music provider inside Music Assistant is sufficient.

This issue will be updated to reflect the current status of the YT Music provider to keep you updated, so there is no need to post separate issues for this.

If you see errors like below in the logs, it is likely that Google has changed something and we will need to wait for a fix:
```
ERROR: [youtube] xxxxxxxxx : Requested format is not available. Use --list-formats for a list of available formats
```

RomuloVitoi:
A JavaScript runtime is now required for the Youtube Music provider to work, deno is the recommended and the default setting on yt-dlp: https://github.com/yt-dlp/yt-dlp/wiki/EJS

This fixes https://github.com/music-assistant/support/issues/4348

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -171,8 +171,13 @@ RUN set -x \
         libpopt0 \
         # pkg-config needed for PyAV (for aioresonate) to find system FFmpeg
         pkg-config \
+        # unzip needed for deno
+        unzip \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
+
+# Install deno, required for Youtube Music provider
+RUN curl -fsSL https://deno.land/install.sh | sh -s -- -y
 
 # Install Snapcast 0.34.0 from GitHub releases (requires at least 0.27)
 ARG SNAPCAST_VERSION=0.34.0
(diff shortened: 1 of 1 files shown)


Dnny44:
I may be completely wrong, but I think installing the dependency through the dockerfile is not the preferred method of adding the dependency. I have another pull request that I think is possible better but If I'm wrong I apologize!

[https://github.com/music-assistant/server/pull/2636](url)

PR #2635, opened 2025-11-12
Typing fixes for the tracks controller

OzGav:


MarvinSchenkel:
Now that `base.py` is in `dev` (and thus not excluded anymore), I discovered that your base.py change needs to be applied to two more function inside the base class:
- _update_library_item
- _add_library_item

I think this is good to merge after those changes.

MarvinSchenkel (approved):
Nice work Gav!

PR #2639, opened 2025-11-14
Add type hints to config controller functions

marcelveldt:
Add type hints to the config controller functions to get config entry values.
This prevents having to do a lot of type casts everywhere when retrieving config values.


marcelveldt:
@OzGav FYI - this helps your other PR(s)

OzGav:
I have made the required simplifying changes to the streams controller mypy PR

PR #2642, opened 2025-11-15
Fix squeezelite sync group source selection

OzGav:
When a squeezelite sync group was playing from a Music Assistant queue, the player would report its active source as its own player_id (e.g., "study player") instead of the queue ID.

Fixes https://discord.com/channels/753947050995089438/1439266061500289064


marcelveldt (approved):
LGTM, Thanks @OzGav !

PR #2650, opened 2025-11-17
Disable Zone Handling if needed for MusicCast players

fmunkes:
There are MusicCast players, where the aiomusiccast library used claims multiple zones despite them being disabled in the receiver. As I don't have access to such a player, this PR introduces a player option allowing to disable the zone handling altogether for a player with multiple zones.
(Hopefully) fixes the issues mentioned in music-assistant/support#4378.

@OzGav Can we update the docs please:
Settings, please add: - Disable zone handling completely: This disable the automatic source change if you play to another source. It is recommended to first try using the provider with this toggle disabled. However, should you encounter issues during playback, this toggle might fix your issues. This setting is only available on multi-zone players.

Known Issues / Notes, please add: - should your have trouble during playback on an AV receiver, i.e. the receiver switches the input source while skipping, then try to disable the zone handling in the player settings.

Thanks!


marcelveldt (approved):
Not entirely sure what this will do/fix but ok

fmunkes:
Thanks, there are players, where aiomusiccast returns multiple zones, but one of them is disabled/ does not exist. This leads to the zone handling being routed to the main zone on the receiver, such that it changes the current input source to non-MA on every subsequent play_media call. For these players users can disable the zone handling for the phantom zone, solving the issue. A user already confirmed, that this fixes the issue.

PR #2657, opened 2025-11-19
Fix TypeError when caching browse results with Sequence return type

OzGav:
The parse_value helper in api.py was not handling Sequence and Iterable types, only tuple and list. When the browse method returns Sequence[MediaItemType | ItemMapping | BrowseFolder], the cached data parsing would fail with:
  TypeError: isinstance() argument 2 cannot be a parameterized generic

This is affecting the SiriusXM and Radio Browser providers

Originally reported here https://discord.com/channels/753947050995089438/1437238396433993850

I tested this against radio browser and it worked but I can't test SiriusXM


marcelveldt:
Can't we just fix this at the roots and change the typing to be always list ?
Why are we using Sequence in the first place ?

OzGav:
This is all legacy code so I can't answer your second question. I had Claude search the codebase and I believe these are all the uses of Sequence. Did you want these changes made and revert api.py?

### Summary of all Sequence usages with @use_cache:
**Files that NEED changing (have both @use_cache and Sequence):**
music_assistant/providers/siriusxm/init.py (line ~250)
browse() → Sequence[MediaItemType | ItemMapping | BrowseFolder]

music_assistant/providers/radiobrowser/init.py (lines ~289, ~303)
get_by_popularity() → Sequence[Radio]
get_by_votes() → Sequence[Radio]

**Files that SHOULD change for consistency (base classes/interfaces):**
music_assistant/models/music_provider.py (line ~491)
Base browse() method signature

music_assistant/controllers/music.py (line ~415)
Music controller browse() method

**Other files using Sequence but NO caching (optional to change):**
All other browse() implementations in:

_demo_music_provider/__init__.py
ard_audiothek/__init__.py
audiobookshelf/__init__.py (and 11 internal _browse_* methods)
filesystem_local/__init__.py
phishin/provider.py
podcast_index/provider.py
radioparadise/provider.py

marcelveldt:
Yeah, I have always wondered why these are typed as sequence and not as list.
Maybe adjust it and see what happens ?

OzGav:
Changing from Sequence to list caused many mypy errors. Claude summarised the issue:

This is why Sequence was used in the first place. The mypy error is telling us that:

list is invariant - so list[Radio] is NOT compatible with list[MediaItemType | BrowseFolder]
Sequence is covariant - so Sequence[Radio] IS compatible with Sequence[MediaItemType | BrowseFolder]

The issue is that browse() returns Sequence[MediaItemType | BrowseFolder], and:

Sequence[Radio] → Sequence[MediaItemType | BrowseFolder] ✅ (covariant, works)
list[Radio] → Sequence[MediaItemType | BrowseFolder] ❌ (invariant, fails)

marcelveldt (approved):

PR #2660, opened 2025-11-20
POC: Add server endpoint for preloading next stream (backend)

elialbert:
[frontend PR](https://github.com/music-assistant/frontend/pull/1211)

this is a POC. But I think it may be fairly useful? I'm totally new to this project and open to feedback. It does cut off 1-2 seconds when clicking "next" in the browser!

### DIFF
--- a/music_assistant/controllers/streams/streams_controller.py
+++ b/music_assistant/controllers/streams/streams_controller.py
@@ -62,9 +62,7 @@ from music_assistant.constants import (
     VERBOSE_LOG_LEVEL,
 )
 from music_assistant.controllers.players.player_controller import AnnounceData
-from music_assistant.controllers.streams.smart_fades import (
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Hey there! Let me try to understand what you are trying to do here. I think you are trying to preload the next queue item in the front end, so that the built in player already has the stream available in order for the 'next' command to be a bit faster?

elialbert:
exactly!

elialbert:
link to related discussions: https://discord.com/channels/753947050995089438/1441399288352739399

https://discord.com/channels/753947050995089438/1098879369792983071/1441398451140165672

marcelveldt:
As discussed on discord, this is not the way this should work. The next track is already preloaded by other parts of the backend

PR #2661, opened 2025-11-21
Disallow malicious playlist updates

MarvinSchenkel:
This PR disallows malicious payloads when creating or updating playlists. We now implement the following security checks:
- Updating a provider_mappings' item_id is not allowed, ever
- Playlist validation is implemented when creating or updating playlists for non-streaming providers. Only those could interact with the local filesystem.


marcelveldt (approved):

music_assistant/controllers/media/base.py
```diff
@@ -154,10 +158,39 @@ async def _get_library_item_by_match(self, item: Track | ItemMapping) -> int | N
...
+        Verify that the update is allowed from a security perspective.
+
+        Validates that provider_mappings item_id values are immutable.
+        This prevents database poisoning attacks via path traversal.
+        """
+        # Build lookup dict: (provider_instance, provider_domain) -> item_id
+        current_mappings = {
+            (mapping.provider_instance, mapping.provider_domain): mapping.item_id
+            for mapping in current_item.provider_mappings
+        }
```
marcelveldt: I'm not sure this is the right approach because some of the backend does actually change provider mappings, or better said: adds additional provider mappings to an item (the matching logic for example).

This check here will already fail if one provider has multiple mappings for one item which is very common for streaming providers. 

I think this is going to cause side effects we don't want.

Also, what about adding a new item ? It will introduce the same issue.
So instead of preventing an item_id update, I would simply add sanity checks on the item_id.

MarvinSchenkel: This check will only raise when an existing provider_mapping's item_id is updated. Adding additional provider_mappings do not trigger the exception. If you still think this could have unwanted side effects, I could just remove this check from the base controller and add sanity checks on `item_id` to the `_update_library_item` function of the playlist controller only.

marcelveldt: problem is that you create a mapping per provider instance but a provider instance can have multiple mappings for the same item - for example a single track in spotify (or any other streaming provider) can have multiple mappings.

So I would suggest that we simply do not allow updating provider mappings at all in the update_item functions and instead we have a method to update item mappings not exposed on the api.
But at the same time we would want the provider mappings to be manageable, for example to allow a user to remove a mapping or simply link items together.

That brings us back to the real issue; we don't want to accept malicious paths to be injected through an api call. I'll have to chew on this for a little while to overthink our options.

In any case, the change you propose here isn't going to work

music_assistant/controllers/media/base.py
```diff
@@ -154,10 +158,39 @@ async def _get_library_item_by_match(self, item: Track | ItemMapping) -> int | N
...
+            if mapping_key in current_mappings:
+                # This provider mapping already exists, verify item_id hasn't changed
+                if current_mappings[mapping_key] != update_mapping.item_id:
+                    msg = (
+                        f"Cannot modify provider_mappings item_id: "
+                        f"attempted to change '{current_mappings[mapping_key]}' to "
+                        f"'{update_mapping.item_id}' for provider "
+                        f"{update_mapping.provider_instance}"
+                    )
+                    raise InvalidDataError(msg)
```
marcelveldt: This will raise as soon as a provider instance has multiple mappings

music_assistant/helpers/playlists.py
```diff
@@ -47,6 +48,24 @@ def is_url(self) -> bool:
         return all([result.scheme, result.netloc])
 
 
+def validate_playlist_filename(filename: str) -> None:
```
marcelveldt: This will actually fail for valid playlists that are nested within a filesystem provider's base folder.
Really the only way to protect all this is to simply disallow this on the API.

music_assistant/controllers/media/playlists.py
```diff
@@ -85,6 +93,8 @@ async def create_playlist(
         else:
             provider = self.mass.get_provider("builtin")
 
+        if provider and not provider.is_streaming_provider:
+            validate_playlist_filename(name)
```
marcelveldt:
```suggestion
        if "/" in name or "\" in name or ".." in name:
            msg = f".name} is not a valid Playlist name"
            raise InvalidDataError(msg)
```

just always do this check - the name may only be a name and nothing else

music_assistant/helpers/playlists.py
```diff
@@ -47,6 +48,26 @@ def is_url(self) -> bool:
...
+        msg = f"Invalid playlist filename: absolute paths not allowed: {filename}"
+        raise InvalidDataError(msg)
+
+    # Reject path traversal attempts (..)
+    # Check as path components to catch '../', or isolated '..'
+    if ".." in Path(filename).parts:
+        msg = f"Invalid playlist filename: path traversal not allowed: {filename}"
+        raise InvalidDataError(msg)
+
+
```
marcelveldt:
```suggestion
```

marcelveldt: this check can be simplified by not allowing any item_id changes for a non streaming provider at all.
so we can simply do that in the handler itself.

I wouldn't add this check to the playlists helper as this is not specific to playlists parsing and you may have situations where path traversal is actually required, then this helper set the wrong expectations

music_assistant/helpers/playlists.py
marcelveldt, line 1: so leave out all changes here in the playlists.py helper

music_assistant/controllers/media/playlists.py
```diff
@@ -44,6 +45,13 @@ def __init__(self, *args, **kwargs) -> None:
             "music/playlists/remove_playlist_tracks", self.remove_playlist_tracks
         )
 
+    def _verify_update_allowed(self, update: Playlist) -> None:
+        # Additional validation for non-streaming providers to prevent path traversal
+        for update_mapping in update.provider_mappings:
+            provider = self.mass.get_provider(update_mapping.provider_instance)
+            if provider and not provider.is_streaming_provider:
+                validate_playlist_filename(update_mapping.item_id)
+
```
marcelveldt: adjust this function to receive the old provider mappings and the new provider mappings.
then work out which provider mappings belong to a filesystem provider and simply deny changing the item _id for the same instance_id - that works because in case of non-streaming providers the playlist item will be unique

music_assistant/controllers/media/playlists.py
```diff
@@ -44,6 +45,35 @@ def __init__(self, *args, **kwargs) -> None:
...
+        for update_mapping in update.provider_mappings:
+            mapping_key = (update_mapping.provider_instance, update_mapping.provider_domain)
+
+            # Only check if this is an existing mapping being modified
+            if mapping_key in current_mappings:
+                current_item_id = current_mappings[mapping_key]
+
+                # Disallow item_id changes for non-streaming providers
+                if current_item_id != update_mapping.item_id:
+                    provider = self.mass.get_provider(update_mapping.provider_instance)
```
marcelveldt: if you temporary disable a provider, then do the update, this will still pass.
we are focusing on providers that store files locally so I would just do a manual check here:

if provider_instance.startswith(("filesystem", "builtin"))

also for the mapping table above you can simply only use the provider_instance

music_assistant/controllers/media/playlists.py
```diff
@@ -44,6 +45,33 @@ def __init__(self, *args, **kwargs) -> None:
             "music/playlists/remove_playlist_tracks", self.remove_playlist_tracks
         )
 
+    def _verify_update_allowed(self, current_item: Playlist, update: Playlist) -> None:
+        """Verify that the update is allowed from a security perspective.
+
+        Prevents updating item_id for non-streaming providers to prevent path traversal attacks.
+        """
+        # Build lookup dict of current mappings: (provider_instance, provider_domain) -> item_id
```
marcelveldt:
```suggestion
        # Build lookup dict of current mappings: provider_instance -> item_id
```

marcelveldt (approved):
Thanks, @MarvinSchenkel !

PR #2662, opened 2025-11-22
Fix for non-numeric artist ID in album controller

OzGav:
I'm not sure why the albums controller passed the tests but the tracks controller is failing a Jellyfin test when the artist ID is a non-numeric hash (like a0c459294295710546c81c20a8d9abfc or unknown). I had fixed it in the tracks controller but when albums got merged the same test is failing again (in tracks) but pointing to the albums controller.


music_assistant/controllers/media/albums.py
```diff
@@ -461,10 +461,12 @@ async def _set_album_artist(
...
-            # Type narrowing: if artist is an ItemMapping, convert it or handle it
             if isinstance(artist, ItemMapping):
-                # ItemMapping can't be added directly, use the existing or skip
-                db_artist = artist
+                # Can't add ItemMapping to library, must use existing
+                if not db_artist:
+                    msg = f"ItemMapping {artist.item_id} has no library entry"
+                    raise InvalidDataError(msg)
+                # else: use the existing db_artist
             else:
```
marcelveldt: artist is actually the only one for which we can actually allow an itemmapping, so you could just skip the entire block and only execute what is now in the else clause

OzGav: Following your comment that "artist is actually the only one for which we can actually allow an itemmapping", I removed the ItemMapping check and simplified this section to:
```
if not db_artist or overwrite:
    db_artist = await self.mass.music.artists.add_item_to_library(
        artist, overwrite_existing=overwrite
    )
```
However, I now have a mypy error on this line:
```
db_artist = await self.mass.music.artists.add_item_to_library(
    artist, overwrite_existing=overwrite
)
```
Error: Argument 1 to `add_item_to_library` of "MediaControllerBase" has incompatible type `Artist | ItemMapping`; expected `Artist`
Looking at the base controller, `add_item_to_library` has signature `item: ItemCls` but internally it calls `_get_library_item_by_match(item)` which accepts `ItemCls | ItemMapping`. So the type signature is narrower than what the implementation actually supports.
Should I just widen the signature in the base controller to `item: ItemCls | ItemMapping`? That would fix the mypy error and reflect what the code actually does.
Note: This same change would need to be applied to `_set_track_artist` in the tracks controller once we settle on the approach.

marcelveldt: yes you can widen it if you want but that will lead to another error as the type differs from the base class.
either ignore that or use that helper function that converts an itemmapping into an artist (I think its in the artists controller but not sure)

MarvinSchenkel:
This can be merged after [this](https://github.com/music-assistant/server/pull/2635#issuecomment-3575208940) is addressed and merged.

music_assistant/controllers/media/albums.py
OzGav: OK using the helper function!

OzGav:
Closing as added to here https://github.com/music-assistant/server/pull/2635

PR #2665, opened 2025-11-23
Subsonic: Give user control over pagination size

khers:
The selected value of 500 entries per invocation may not be ideal for all users. Given that there isn't a single value that is ideal for all users, let each configure their own. The default is shrinking to 200 as a starting point.


music_assistant/providers/opensubsonic/__init__.py
```diff
@@ -154,4 +155,14 @@ async def get_config_entries(
...
         ),
+        ConfigEntry(
+            key=CONF_PAGE_SIZE,
+            type=ConfigEntryType.INTEGER,
+            label="Number of items included per server request.",
+            required=True,
+            description="When enumerating items from the server, how many should be in each "
+            "request. Smaller will require more requests but is better for low bandwidth "
+            "connections. The Open Subsonic spec says the max value for this is 500 items.",
+            default_value=200,
```
marcelveldt, line 169:
```suggestion
            default_value=200,
            category="advanced"
```

MarvinSchenkel (approved):

PR #2669, opened 2025-11-24
Add extra info to scrobble data

OzGav:
Partially addresses https://github.com/music-assistant/support/issues/4101

This will change the scrobbled track name to the form `name (version)` when there is version info in the database

After discussing with Marcel we will not be trying to add artist names back into the title. (As was requested in the issue report)

@wjzijderveld @khers @ijc Just want to check in with you guys that this won't cause any issues with your providers?

Also based on here https://discord.com/channels/753947050995089438/1387755256665870346
I have added album_artist to the event for you to send with the scrobble (last.fm supports it I am not sure of the others...?)


music_assistant/controllers/player_queues.py
```diff
@@ -2173,7 +2177,7 @@ def _handle_playback_progress_report(
             data=MediaItemPlaybackProgressReport(
                 uri=item_to_report.media_item.uri,
                 media_type=item_to_report.media_item.media_type,
-                name=item_to_report.media_item.name,
+                name=track_name,
```
wjzijderveld: Maybe we should pass the version separate and make the scrobbler (or the helper) decide to append it to the scrobbled name?
I think that would make more sense, and also keeps it open for extension more. For example to make this a configuration option, instead of forcing it.

khers:
I don't maintain the subsonic_scrobble plugin (@Clusters does, IIRC) but I don't think this will impact the subsonic_scrobble plugin at all. The subsonic scrobble endpoint takes an item ID and doesn't match based on title.

ijc:
I don't foresee a problem for listenbrainz scrobbler, it passes on various mbids which are what matters not the name.

It could potentially make use of the new `album_artist*` fields in the future.

music_assistant/controllers/player_queues.py
OzGav: Sounds reasonable. Lets see what Marvin or Marcel have to say.

khers: I like this suggestion, let each scrobbling plugin decide what to do with the extra information.

OzGav: Will just have to make a change to the models which is easy enough.

marcelveldt: yes, I would go with that indeed! let the scrobbler decide how it can handle it.
maybe also have a dedicated artists (list with artist names) and a artist_str which contains the concatenated string of all artist names

OzGav: Models PR https://github.com/music-assistant/models/pull/136

Clusters:
@khers you are absolutely correct. This PR should not affect the subsonic_scrobble plugin, as it does not use the name property. 👍

music_assistant/controllers/player_queues.py
marcelveldt: the models PR is merged and bumped in the server (so you can rebase this branch/PR)

OzGav: New fields added

marcelveldt (approved):

OzGav:
@wjzijderveld @khers @ijc @Clusters Thanks for your feedback! This has been merged, Willem-Jan are you able to add the configurable option to suffix the version to the track name? We could then close that issue. edit: I see we already have! :-) Still let me know if you can do that change

wjzijderveld:
Yeah, I'll take a look! Thanks for adding the extra info!

PR #2678, opened 2025-11-25
Fix squeezelite UnboundLocalError exception when playing a plugin source.

sprocket-9:
```
 File "/opt/git/ma/server/squeezelite-queue-unbound/music_assistant/providers/squeezelite/player.py", line 429, in _handle_play_url_for_slimplayer
    if queue and queue.repeat_mode == RepeatMode.ONE:
       ^^^^^
UnboundLocalError: cannot access local variable 'queue' where it is not associated with a value
```

### DIFF
--- a/music_assistant/providers/squeezelite/player.py
+++ b/music_assistant/providers/squeezelite/player.py
@@ -407,6 +407,7 @@ class SqueezelitePlayer(Player):
             "source_id": media.source_id,
             "queue_item_id": media.queue_item_id,
         }
+        queue = None
         if media.source_id and (queue := self.mass.player_queues.get(media.source_id)):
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Since `queue` is used in two places, we could just define it outside the if without the walrus operator. So 

Instead of:
```
queue = None
if media.source_id and (queue := self.mass.player_queues.get(media.source_id)):
...
```

Let's just use:
```
queue = self.mass.player_queues.get(media.source_id)
if media.source_id and queue:
...
```

marcelveldt:
> Since `queue` is used in two places, we could just define it outside the if without the walrus operator. So

That won't work because media.source id is optional.
Let's leave it like this to fix the bug and optimize later

marcelveldt (approved):

PR #2682, opened 2025-11-26
Pre-build PyAV wheel in base image to speed up release builds

maximmaxim345:
- Add pyav-builder stage in Dockerfile.base that compiles PyAV against our custom FFmpeg build
- Auto-detect PyAV version from aioresonate's dependencies at build time
- Store pre-built wheel at /usr/local/share/pyav-wheels/ for use by release builds

This reduces release build times from ~20 minutes back to ~1 minute by avoiding PyAV compilation on every release, while still keeping the `--no-binary` option when installing `av` for reduced image sized (added in #2577).
The base image build only needs to run when the PyAV version changes, failing fast in all release builds to avoid a version mismatch.


marcelveldt (approved):
nice!

PR #2686, opened 2025-11-26
fix(bbc_sounds): Handle library exception more gracefully

kieranhogg:
An unexpected json response from the heartbeat API endpoint wasn't checked for `TypeError`s, handle this in both the library and the MA provider

### DIFF
--- a/music_assistant/providers/bbc_sounds/__init__.py
+++ b/music_assistant/providers/bbc_sounds/__init__.py
@@ -797,7 +797,7 @@ class BBCSoundsProvider(MusicProvider):
                     success = await self.client.streaming.update_play_status(
                         pid=media_item.item_id, elapsed_time=position, action=action
                     )
-                    self.logger.info(f"Updated play status: {success}")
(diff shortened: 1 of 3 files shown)


music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -793,10 +793,13 @@ async def on_played(
...
             if action:
-                success = await self.client.streaming.update_play_status(
-                    pid=media_item.item_id, elapsed_time=position, action=action
-                )
-                self.logger.info(f"Updated play status: {success}")
+                try:
+                    success = await self.client.streaming.update_play_status(
+                        pid=media_item.item_id, elapsed_time=position, action=action
+                    )
+                    self.logger.info(f"Updated play status: {success}")
```
MarvinSchenkel: Does this need to be info? Sounds like a debug message to me.

kieranhogg: Good spot, thanks!

MarvinSchenkel (approved):

PR #2688, opened 2025-11-26
BBC Sounds: Improve title metadata

kieranhogg:
With a variety of parent podcasts, radio series, stations, programs and networks, improve some of the edge cases for metadata titles.

### DIFF
--- a/music_assistant/providers/bbc_sounds/adaptor.py
+++ b/music_assistant/providers/bbc_sounds/adaptor.py
@@ -428,9 +428,21 @@ class PodcastConverter(BaseConverter):
                 seconds_streamed=(int(episode.position) if episode.position else 0),
             )
         elif episode and isinstance(episode, Track) and source_obj.stream:
-            metadata = StreamMetadata(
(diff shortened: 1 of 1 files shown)


music_assistant/providers/bbc_sounds/adaptor.py
```diff
@@ -428,9 +428,21 @@ async def get_stream_details(self, source_obj: ConvertableTypes) -> StreamDetail
...
+            elif source_obj.network:
+                title = f"BBC {source_obj.network.short_title}"
+            elif source_obj.container:
+                title = source_obj.container.title
+            elif episode.metadata and episode.metadata.description:
+                title = episode.metadata.description
+            elif source_obj.titles:
+                title = source_obj.titles["primary"]
+            else:
+                title = ""
```
MarvinSchenkel, line 443: Does it make sense to have a better fallback instead of `""`? Maybe "BBC" or something along those lines?

marcelveldt: It wil automatically fallback to the media item name if the title is empty

marcelveldt (approved):

PR #2689, opened 2025-11-26
BBC Sounds: Unauthenticated might see blank station list

kieranhogg:
Related to the menu API ordering, don't assume we've always set self.menu now it has moved places


MarvinSchenkel (approved):

MarvinSchenkel (dismissed):
Seems mypy is failing in the test. If you fix that,  I think this is fine to merge 👍

marcelveldt (approved):

PR #2702, opened 2025-11-27
Airplay2 improvements

bradkeifer:
- Sync's with RAOP playback
- Improved ntpstart handling


music_assistant/providers/airplay/helpers.py
```diff
@@ -291,6 +288,7 @@ def add_seconds_to_ntp(ntp_timestamp: int, seconds: float) -> int:
     fraction = seconds - whole_seconds
 
     # Convert to NTP format (upper 32 bits = seconds, lower 32 bits = fraction)
+    # What happens if the sum of existing fraction and new fraction exceeds 1 second?
```
marcelveldt: I don't know, do you ?

music_assistant/providers/airplay/constants.py
```diff
@@ -35,7 +35,13 @@ class StreamingProtocol(Enum):
 RAOP_DISCOVERY_TYPE: Final[str] = "_raop._tcp.local."
 DACP_DISCOVERY_TYPE: Final[str] = "_dacp._tcp.local."
 
+AIRPLAY_PRELOAD_CHUNKS: Final[int] = 5  # Number of PCM chunks to preload before throttling back
```
marcelveldt:
```suggestion
AIRPLAY_PRELOAD_SECONDS: Final[int] = 5  # Number of seconds (in PCM) to preload before throttling back
```

music_assistant/providers/airplay/stream_session.py
```diff
@@ -414,7 +433,13 @@ async def _start_client_ffmpeg(self, airplay_player: AirPlayPlayer) -> None:
             output_format=airplay_player.stream.pcm_format,
             filter_params=filter_params,
             audio_output=airplay_player.stream.audio_pipe.path,
-            extra_input_args=["-y", "-re"],
+            extra_input_args=[
+                "-y",
+                "-readrate",
+                "1",
+                "-readrate_initial_burst",
+                f"{AIRPLAY_PRELOAD_CHUNKS}",
```
marcelveldt:
```suggestion
                f"{AIRPLAY_PRELOAD_SECONDS}",
```

music_assistant/providers/airplay/stream_session.py
```diff
@@ -88,7 +96,18 @@ async def start(self, audio_source: AsyncGenerator[bytes, None]) -> None:
...
+        # The also have a fixed 2000ms output buffer. We will not be able to respect the
+        # ntpstart time unless we cater for all these time delays.
+        # RAOP clients need less due to less RTSP exchanges and different packet buffer
+        # handling
+        # Plus we need to cater for process spawn and initialisation time
+        wait_start = (
+            AIRPLAY2_CONNECT_TIME_MS
+            + AIRPLAY2_OUTPUT_BUFFER_DURATION_MS
+            + AIRPLAY_PROCESS_SPAWN_TIME_MS
+            + (250 * len(self.sync_clients))
```
marcelveldt, line 109: do we still need to add the extra 250ms per client ? I think that is fine to drop since we spawn all clients at the same time ? There will maybe be some overhead involved with that but not 250ms

music_assistant/providers/airplay/protocols/airplay2.py
```diff
@@ -59,9 +59,6 @@ async def start(self, start_ntp: int, skip: int = 0) -> None:
         player_id = self.player.player_id
         sync_adjust = self.mass.config.get_raw_player_config_value(player_id, CONF_SYNC_ADJUST, 0)
         assert isinstance(sync_adjust, int)
-        read_ahead = await self.mass.config.get_player_config_value(
-            player_id, CONF_READ_AHEAD_BUFFER, return_type=int
-        )
```
marcelveldt, line 64: shall we drop this entire setting ? So also drop it for RAOP I mean ?

music_assistant/providers/airplay/helpers.py
bradkeifer: Not for sure. I wanted to test, but forgot and ran out of time. I was thinking that it probably works fine. I think it would overflow the 32-bits for the fraction and add a 1 into the lowest bit for the seconds and hence all be good.

music_assistant/providers/airplay/protocols/airplay2.py
bradkeifer: I reckon that is probably worthwhile and just have it fixed at 2000ms. It serves a purpose by ensuring some buffering within cliraop, but it doesn't have anything to do with player latencies as such - especially now that we are using named pipes and can pump audio into them ahead of time to suit our needs.

music_assistant/providers/airplay/stream_session.py
bradkeifer: It's worth a try. I'm sure that you're like me and don't like the magnitude of the delay we have. I was hoping we could keep it to just a bit longer than the time taken to establish the AirPlay2 connection, but then I ran into playback problems with cliap2. 
It's a delicate balancing act and I think needs decent testing each time we adjust it. 
I'm happy to explore ways to reduce this time in the future, but I won't have the bandwidth to do so in the very short term sorry.

marcelveldt: yeah let'd just drop it, especially since you now tuned the optimal start value

music_assistant/providers/airplay/protocols/airplay2.py
marcelveldt: let's drop it then!

music_assistant/providers/airplay/stream_session.py
bradkeifer: I tried removing the additional 250ms/client and the result was not good. cliap2 ended up ~two seconds later than cliraop. I suspect cliraop could not handle the requested ntpstart and reverted to now. So, I think it is best to leave this extra 250ms/client until we can fully understand the dynamics between and within the two binaries and MA. There must be a way to reduce this delay time, but I think it requires some detailed debugging and possibly some coding changes within both cliraop and cliap2.

music_assistant/providers/airplay/helpers.py
bradkeifer: Tested and works perfectly. The 32-bit overflow does add to the seconds value.

music_assistant/providers/airplay/protocols/airplay2.py
bradkeifer: Done

bradkeifer:
Review comments addressed. Unfortunately, we need to retain the 250ms/client delay for the moment. Ready for your review again please @marcelveldt

marcelveldt (approved):
Looks good, thanks @bradkeifer !

PR #2704, opened 2025-11-28
Spotify connect: Fix sudden volume change when session starts

MarvinSchenkel:
This PR:
- Enables volume control from the Spotify app (e.g. your phone or the Spotify desktop app)
- Makes sure that the volume does not jump to 100% when a session starts
- Refactor stderr handling in a separate function to prevent `PLR0915`

### DIFF
--- a/music_assistant/providers/spotify/provider.py
+++ b/music_assistant/providers/spotify/provider.py
@@ -1074,8 +1074,9 @@ class SpotifyProvider(MusicProvider):
                 raise ResourceTemporarilyUnavailable(backoff_time=30)
             response.raise_for_status()
 
-    async def _put_data_unthrottled(self, endpoint: str, data: Any = None, **kwargs: Any) -> None:
-        """Put data on api without throttling."""
+    @throttle_with_retries
+    async def _put_data(self, endpoint: str, data: Any = None, **kwargs: Any) -> None:
+        """Put data on api."""
         url = f"https://api.spotify.com/v1/{endpoint}"
         auth_info = kwargs.pop("auth_info", await self.login())
         headers = {"Authorization": f"Bearer {auth_info['access_token']}"}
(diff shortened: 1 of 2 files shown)


music_assistant/providers/spotify/provider.py
```diff
@@ -1100,6 +1099,11 @@ async def _put_data(self, endpoint: str, data: Any = None, **kwargs: Any) -> Non
                 raise ResourceTemporarilyUnavailable(backoff_time=30)
             response.raise_for_status()
 
+    @throttle_with_retries
+    async def _put_data(self, endpoint: str, data: Any = None, **kwargs: Any) -> None:
+        """Put data on api with throttling."""
+        await self._put_data_unthrottled(endpoint, data, **kwargs)
+
```
marcelveldt: we have a ContextVar for this

marcelveldt (changes requested):
Please use the existing BYPASS_THROTTLER contexvar for this

music_assistant/providers/spotify/provider.py
MarvinSchenkel: Fixed

music_assistant/providers/spotify_connect/__init__.py
```diff
@@ -439,13 +511,13 @@ async def _librespot_runner(self) -> None:
                 self.named_pipe,
                 "--dither",
                 "none",
-                # disable volume control
+                # Enable volume control with linear scale for intuitive volume feel
                 "--mixer",
                 "softvol",
                 "--volume-ctrl",
-                "fixed",
+                "linear",
```
marcelveldt: This is wrong because this will enable software volume control on the output itself. We only want hardware volume control, meaning that the volume control is directed to the attached player

marcelveldt (changes requested):

music_assistant/providers/spotify_connect/__init__.py
MarvinSchenkel: Makes sense. I will revert this part and just fix the heart attack bug then. Will dive into librespot to see if I can get it to just send the volume_changed events without changing the stream's volume. Will do this in a different PR then.

marcelveldt (approved):

PR #2714, opened 2025-11-29
Fix for GET /all RTSP response handler in AirPlay

bradkeifer:
This is a fix for a problem created in the response handler for the RTSP GET /all request. 
A segment of OwnTone code was accidentally commented out, which will result in pairing issues for devices requiring PIN-based pairing or HomeKit pairing.


marcelveldt (approved):
Thanks, @bradkeifer !

PR #2732, opened 2025-12-02
Update Implemented Sendspin Version with included Volume Support

maximmaxim345:
This PR updates the implemented Sendspin specification (formerly known under the working title “Resonate”) to the latest version of the spec.

This includes a bunch of under the hood improvements to the protocol, but also finally adds volume support.

Almost everything is implemented according to the spec, but there are still a couple of exceptions:
- Player Format Changes through `stream/request-format` are not implemented
- the `paused` `playback_state` is never used. Only `playing` and `stopped` are ever sent.
- Streams are ended immediately when playback stops, skipping to the next track, or when seeking. Artwork is also cleared because of this. (related to the missing `paused` state)
- [Multi Server Support](https://github.com/Sendspin-Protocol/spec/pull/43) messages are implemented, but aren't fully utilized yet


music_assistant/providers/sendspin/provider.py
```diff
@@ -1,57 +1,57 @@
...
-        self.server_api = ResonateServer(
+        self.server_api = SendspinServer(
             self.mass.loop, mass.server_id, "Music Assistant", self.mass.http_session
         )
         self.unregister_cbs = [
             self.server_api.add_event_listener(self.event_cb),
             # For the web player
             self.mass.webserver.register_dynamic_route(
-                "/resonate", self.server_api.on_client_connect
+                "/sendspin", self.server_api.on_client_connect
```
marcelveldt, line 38: One downside of this is that the resonate handler is now fully exposed on the webserver, so we should maybe consider to require an auth command here later or find some other way to secure the connection. Something for later but keep in mind

marcelveldt (approved):
Nice work @maximmaxim345 !

PR #2737, opened 2025-12-02
Add Experimental Sendspin over Cast Support

maximmaxim345:
## Summary

Adds experimental support for using the Sendspin protocol with Chromecast devices via the [sendspin cast](https://sendspin.github.io/cast/sender.html) receiver app. This enables:

- **Cross-platform sync groups**: Group Chromecast devices with other Sendspin-compatible players (like the Home Assistant Voice Preview Edition, the future web player, or any other Sendspin compatible speakers) for synchronized multi-room audio
- **Flexible device combinations**: Group Chromecast Audio, Google Home, and Cast-enabled TVs with non-Google directly in Music Assistant, without going through the Google Home App

## How to use

⚠️ Sendspin over Cast is still highly experimental so expect bugs.

When "Sendspin mode" is enabled for a Chromecast player:
1. Power on a Chromecast device
2. This will create a new Sendspin player called `Speaker Name (Sendspin)`
3. You can group any Sendspin Speaker when grouping them through the `Speaker Name (Sendspin)` speaker

## Configuration

- **Enable Sendspin mode**: Per-player toggle (requires Sendspin provider enabled)
- **Sync delay (ms)**: Adjust timing offset (positive = delay, negative = advance)


marcelveldt:
> Gets very slowly out of sync over time. Stopping and starting playback seams to fix this.
This most probably means that sync is not working at all ? So the low level access to the browser's realtime clock I mean.
Did you test a long session with the web client to rule out an implementation error in the javascript client
?

maximmaxim345:
> This most probably means that sync is not working at all ? So the low level access to the browser's realtime clock I mean.
> Did you test a long session with the web client to rule out an implementation error in the javascript client?

I think its very likely that there is a bug in the js client somewhere.
I noticed a faint echo after playing a couple of hours to 5 js clients but they werent all synced from the start so I may am just imagining it 🤷

Edit: Time syncing was broken in `resonate-js`.
Fixed that so it now keeps playing in sync on Chromecast, Chromecast Audio, iOS Safari, and Android Firefox after playing for 2 hours!

marcelveldt (approved):
Let's give this to the beta testers for a good test drive!

PR #2741, opened 2025-12-03
Feat/genre search support

jozefKruszynski:
Initial draft PR in order to get some feedback on progress thus far.

### DIFF
--- a/music_assistant/controllers/media/albums.py
+++ b/music_assistant/controllers/media/albums.py
@@ -111,6 +111,7 @@ class AlbumsController(MediaControllerBase[Album]):
         offset: int = 0,
         order_by: str = "sort_name",
         provider: str | list[str] | None = None,
+        genre_filter: list[int] | None = None,
         extra_query: str | None = None,
         extra_query_params: dict[str, Any] | None = None,
         album_types: list[AlbumType] | None = None,
@@ -123,6 +124,7 @@ class AlbumsController(MediaControllerBase[Album]):
         :param offset: Number of items to skip.
         :param order_by: Order by field (e.g. 'sort_name', 'timestamp_added').
         :param provider: Filter by provider instance ID (single string or list).
+        :param genre_filter: Filter by genre library item IDs (list of integers).
         :param extra_query: Additional SQL query string.
         :param extra_query_params: Additional query parameters.
         :param album_types: Filter by album types.
@@ -167,6 +169,7 @@ class AlbumsController(MediaControllerBase[Album]):
             offset=offset,
             order_by=order_by,
             provider_filter=self._ensure_provider_filter(provider),
+            genre_filter=genre_filter,
             extra_query_parts=extra_query_parts,
             extra_query_params=extra_query_params,
             extra_join_parts=extra_join_parts,
@@ -194,6 +197,7 @@ class AlbumsController(MediaControllerBase[Album]):
                 limit=remaining_limit,
                 order_by=order_by,
                 provider_filter=self._ensure_provider_filter(provider),
+                genre_filter=genre_filter,
                 extra_query_parts=extra_query_parts,
                 extra_query_params=extra_query_params,
                 extra_join_parts=extra_join_parts,
--- a/music_assistant/controllers/media/albums.py
+++ b/music_assistant/controllers/media/albums.py
@@ -113,7 +113,7 @@ class AlbumsController(MediaControllerBase[Album]):
         provider: str | list[str] | None = None,
         extra_query: str | None = None,
         extra_query_params: dict[str, Any] | None = None,
-        genre_filter: list[int] | None = None,
+        genres: list[int] | None = None,
         album_types: list[AlbumType] | None = None,
     ) -> list[Album]:
         """Get in-database albums.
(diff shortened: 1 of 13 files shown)


music_assistant/controllers/media/albums.py
```diff
@@ -111,6 +111,7 @@ async def library_items(
         provider: str | list[str] | None = None,
         extra_query: str | None = None,
         extra_query_params: dict[str, Any] | None = None,
+        genre_items_ids: list[int] | None = None,
```
marcelveldt: I would just name this parameter genres and then explain in the docstring that you want library genre id's here

music_assistant/controllers/media/base.py
```diff
@@ -799,8 +828,17 @@ def _apply_filters(
         allowed providers.
         """
         # handle search
-        if search:
+        if search and genre_items_ids:
+            ids_str = ",".join(str(x) for x in genre_items_ids)
+            query_parts.append(
+                f"({self.db_table}.search_name LIKE :search OR "
+                f"{self.db_table}.item_id IN ({ids_str}))"
+            )
```
marcelveldt: This seems incorrect to me

music_assistant/controllers/media/base.py
```diff
@@ -799,8 +828,17 @@ def _apply_filters(
...
+        elif search:
             query_parts.append(f"{self.db_table}.search_name LIKE :search")
+        elif genre_items_ids:
+            ids_str = ",".join(str(x) for x in genre_items_ids)
+            query_parts.append(f"{self.db_table}.item_id IN ({ids_str})")
```
marcelveldt: this is also incorrect - you are modifying the base class here

music_assistant/controllers/media/genres.py
```diff
@@ -1,60 +1,691 @@
...
+    async def _on_item_added(self, event: MassEvent) -> None:
+        """Handle event when an item is added to the library."""
+        item = event.data
+        if not hasattr(item, "metadata") or not item.metadata.genres:
+            return
+        # trigger a rebuild of the genre index
+        # we debounce this to prevent too many rebuilds
+        self.mass.call_later(10, self.rebuild_index, task_id="rebuild_genre_index")
+
+    async def rebuild_index(self) -> None:
```
marcelveldt: I have no idea why this is needed. Cant we do this with pure db queries ?
You probably just need a genre_mappings tables

music_assistant/controllers/music.py
```diff
@@ -2117,6 +2195,38 @@ async def __create_database_tables(self) -> None:
...
+            [timestamp_added] INTEGER DEFAULT (cast(strftime('%s','now') as int)),
+            [timestamp_modified] INTEGER NOT NULL DEFAULT 0,
+            [search_name] TEXT NOT NULL,
+            [search_sort_name] TEXT NOT NULL,
+            [track_ids] json DEFAULT '[]',
+            [album_ids] json DEFAULT '[]',
+            [artist_ids] json DEFAULT '[]',
+            [playlist_ids] json DEFAULT '[]',
+            [podcast_ids] json DEFAULT '[]',
+            [audiobook_ids] json DEFAULT '[]'
```
marcelveldt: So this is the part I have doubts about - I think we can do better than this by having a genre_mappings table. So basically the same as how provider_mappings work currently

music_assistant/controllers/music.py
```diff
@@ -2117,6 +2195,38 @@ async def __create_database_tables(self) -> None:
...
+            );"""
+        )
+        await self.database.execute(
+            f"""
+            CREATE TABLE IF NOT EXISTS {DB_TABLE_GENRE_ALIASES}(
+            [genre_id] INTEGER NOT NULL,
+            [alias] TEXT NOT NULL,
```
marcelveldt: I would investigate storing aliases as json column on the genres table. 
sqlite has a native json handler nowadays which can also perform fast queries on the json and if I'm not mistaken can do indexes as well.

jozefKruszynski: Gotcha, sounds far more sensible. I will probably be able to do a bit on Friday morning, but after that I'm away in the UK until Late Monday night.

music_assistant/controllers/media/genres.py
```diff
@@ -1,60 +1,747 @@
...
+                {"item_id": genre_id},
+                {"genre_mappings": json_dumps(genre_mappings)},
+            )
+
+            count += 1
+            await asyncio.sleep(0)
+
+        self.logger.info("Finished background genre index scan. Updated %s genres.", count)
+
+    async def _scan_media_type(
```
MarvinSchenkel: I think this logic (along with the rebuild_index function) should not be in the controller. I think it makes more sense to create a proper database design for mappings between genres and media_items, rather than trying to constantly sync them off the metadata.genre attributes. The logic in these functions should in my opinion be a migration script that is run only once in order to migrate the existing genre metadata into the new database structure. Then we just need to update our other controllers to insert directly into the genre_mappings rather than writing to metadata.

So the database could look some thing like this
`[track] / [album] etc..` 1 -- * `[media_item_genre_mappings]` * -- 1 `[genre]`

Thoughts?

jozefKruszynski: Sounds like a reasonable suggestion, it makes genre a first class citizen rather than an after thought and I think that thinking about genre in such a way, is actually essential to making something meaningful.

PR #2752, opened 2025-12-05
refactor: remove builtin player (in favor of sendspin)

maximmaxim345:
The builtin player is no longer needed now that sendspin is the primary web player implementation.

Sendspin provides synchronized multi-room audio playback which the builtin player could not achieve.


marcelveldt (approved):

OzGav:
What documentation changes are needed as a result?

PR #2758, opened 2025-12-06
Fix race condition when transitioning from plugin source

sprocket-9:
When transitioning from a plugin source to a different source, player state update was happening before the cleanup in the StreamsController plugin handler finished, resulting in the plugin source still showing as the player's active_source in the frontend.

### DIFF
--- a/music_assistant/controllers/players/player_controller.py
+++ b/music_assistant/controllers/players/player_controller.py
@@ -889,7 +889,7 @@ class PlayerController(CoreController):
             with suppress(PlayerCommandFailed, RuntimeError):
                 # just try to stop (regardless of state)
                 await self.cmd_stop(player_id)
-                await asyncio.sleep(0.5)  # small delay to allow stop to process
+                await asyncio.sleep(2)  # small delay to allow stop to process
         # check if source is a pluginsource
         # in that case the source id is the instance_id of the plugin provider
         if plugin_prov := self.mass.get_provider(source):


music_assistant/controllers/players/player_controller.py
```diff
@@ -889,7 +889,7 @@ async def select_source(self, player_id: str, source: str | None) -> None:
             with suppress(PlayerCommandFailed, RuntimeError):
                 # just try to stop (regardless of state)
                 await self.cmd_stop(player_id)
-                await asyncio.sleep(0.5)  # small delay to allow stop to process
+                await asyncio.sleep(2)  # small delay to allow stop to process
```
marcelveldt, line 892: Better use the wait_for_state helper within the player controller to actually wait for the player to arrive at the idle state instead of guessing the amount of seconds we need to sleep...

await self.wait_for_state(player, wanted_state=PlaybackState.IDLE, timeout=10, minimal_time=1)

sprocket-9: I've tried the helper out but it doesn't seem to solve the race condition.   The streams controller plugin handler is a series of nested async generators so it's not easy to follow.  Is my understanding of what's going on correct:

The player's state changes to idle after a client.stop() and the http connection to the streams controller is closed. 

The cleanup (setting plugin_source.in_use_by = None) in the streams controller plugin source handler happens after the client disconnects and aiohttp cancels the handler task.

The cleanup in the generator occurs in the try.. finally clause, which gets called because it's about to be garbage collected due to the task cancellation?

> Yield expressions are allowed anywhere in a [try](https://docs.python.org/3.14/reference/compound_stmts.html#try) construct. If the generator is not resumed before it is finalized (by reaching a zero reference count or by being garbage collected), the generator-iterator’s [close()](https://docs.python.org/3.14/reference/expressions.html#generator.close) method will be called, allowing any pending [finally](https://docs.python.org/3.14/reference/compound_stmts.html#finally) clauses to execute.

marcelveldt: yeah that is correct.
Maybe we should decouple the "in_use_by" on the plugin source from the selected source on a player to prevent these kind of race conditions ?

marcelveldt: I'll merge this one now so we have a fix for the immediate issue and then next week I'll think about this a bit more

marcelveldt (approved):
Thanks, @sprocket-9 !

PR #2796, opened 2025-12-12
Add DTLS pinning

arturpragacz:
This pull request introduces a major refactor of the WebRTC remote access system, moving from randomly generated remote IDs to deterministic, certificate-based remote IDs. It adds persistent DTLS certificate management for WebRTC, enabling client-side certificate pinning, which significantly improves security.

The most important changes are:

**WebRTC Certificate Management and Remote ID Generation**
- Added a new module, `webrtc_certificate.py`, that handles persistent DTLS certificate creation, storage, loading, and renewal for WebRTC connections. This module also provides deterministic remote ID generation based on the certificate's fingerprint, replacing the previous random ID approach.
- Updated the remote access controller (`remote_access/__init__.py`) to use the persistent certificate and derive the remote ID from it, removing all logic related to random remote ID generation and storage. The controller now exposes the certificate and remote ID as properties.
- Updated the WebRTC gateway (`gateway.py`) to require the remote ID and certificate on initialisation, removing the ability to generate a random remote ID. All peer connections now use the persistent certificate for DTLS.

**API and Provider Integration**
- Updated the Sendspin provider to use the persistent certificate from the remote access controller when creating peer connections, ensuring consistent authentication and pinning.

Frontend PR: https://github.com/music-assistant/frontend/pull/1271

### DIFF
--- a/music_assistant/controllers/webserver/remote_access/__init__.py
+++ b/music_assistant/controllers/webserver/remote_access/__init__.py
@@ -16,12 +16,14 @@ from mashumaro import DataClassDictMixin
 from music_assistant_models.enums import EventType
 
 from music_assistant.constants import CONF_CORE
-from music_assistant.controllers.webserver.remote_access.gateway import (
-    WebRTCGateway,
-    generate_remote_id,
+from music_assistant.controllers.webserver.remote_access.gateway import WebRTCGateway
+from music_assistant.helpers.webrtc_certificate import (
+    get_or_create_webrtc_certificate,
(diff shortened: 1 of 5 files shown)


MarvinSchenkel:
Amazing work @arturpragacz . One question from my side, where is the actual certificate validation happening? Is this automatically done in the signalling server because we now set the certificate in the  `RTCPeerConnection`?

arturpragacz:
It happens on the client.

marcelveldt (approved):

PR #2803, opened 2025-12-13
abs/ gpodder: add configurable user for playlog sync

fmunkes:
adds a config option to the audiobookshelf and gpodder provider, such that the playlog sync is possible in the background with the new user system

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -1166,23 +1166,26 @@ class MusicController(CoreController):
     async def mark_item_unplayed(
         self,
         media_item: MediaItemType,
+        userid: str | None = None,
         all_users: bool = False,
     ) -> None:
         """
         Mark item as unplayed in playlog.
 
         :param media_item: The media item to mark as unplayed.
+        :param userid: The user ID to mark the item as unplayed for (instead of the current user).
         :param all_users: If True, mark the item as unplayed for all users.
         """
         current_user = get_current_user()
-        user_id = current_user.user_id if current_user else "system"
+        if userid is None:
+            userid = current_user.user_id if current_user else "system"
         params = {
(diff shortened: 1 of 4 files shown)


music_assistant/controllers/music.py
```diff
@@ -1166,23 +1166,26 @@ async def mark_item_played(
...
         Mark item as unplayed in playlog.
 
         :param media_item: The media item to mark as unplayed.
+        :param userid: The user ID to mark the item as unplayed for (instead of the current user).
         :param all_users: If True, mark the item as unplayed for all users.
         """
         current_user = get_current_user()
-        user_id = current_user.user_id if current_user else "system"
+        if userid is None:
+            userid = current_user.user_id if current_user else "system"
```
marcelveldt: we need to solve this in a better way.

option 1: ignore the request if we have no user
option 2: just pick the first user

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -167,6 +187,15 @@ async def get_config_entries(
             "you may provide an API key (ABS version >= 2.26). "
             "Please consult the docs.",
         ),
+        ConfigEntry(
+            key=CONF_PLAYLOG_USER_ID,
+            label="Music Assistant user to sync playlog with",
```
marcelveldt, line 192: Can you make this a bit more user friendly title ?

music_assistant/controllers/music.py
fmunkes: I made it identical to mark_item_played - wouldn't I render the frontend ability of marking it unplayed useless if the current user is not the default?

fmunkes:
A better solution came up while discussing in discord, closing this.

PR #2804, opened 2025-12-13
adapt get_playlog_provider_item_ids method of music controller

fmunkes:
adapts the method for multi user setups.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -650,14 +650,28 @@ class MusicController(CoreController):
         return result
 
     async def get_playlog_provider_item_ids(
-        self, provider_instance_id: str, limit: int = 0
+        self, provider_instance_id: str, limit: int = 0, userid: str | None = None
     ) -> list[tuple[MediaType, str]]:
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/music.py
```diff
@@ -650,14 +650,28 @@ async def in_progress_items(
         return result
 
     async def get_playlog_provider_item_ids(
-        self, provider_instance_id: str, limit: int = 0
+        self, provider_instance_id: str, limit: int = 0, userid: str | None = None
     ) -> list[tuple[MediaType, str]]:
         """Return a list of MediaType and provider_item_id of items in playlog of provider."""
+        # check if there is a provider user
+        # this method is not available in the frontend, so no need to check for session users.
```
marcelveldt, line 657: is this ever being called from the context of a play media session ?
Because then the user will be held for the playback context.

or is this backend-only for the provider sync ?

fmunkes: Exclusively during provider sync (full sync, so also not during any socket events)

marcelveldt (approved):
Thanks, @fmunkes !

PR #2805, opened 2025-12-14
feat(subsonic): allow self-signed certificates

Tooa:
Hi :wave:,

## Summary

this pull request (PR) adds support for self-signed certificates for the Subsonic provider. The `insecure` parameter is documented [here](https://github.com/khers/py-opensonic/blob/master/src/libopensonic/connection.py#L96).

Let me know if you want anything changed or have questions about the PR.

## Motivation

I have a self-signed certificate in my homelab for `*.home.internal` and host my services behind a reverse proxy. The change allows me to use: `https://navidrome.home.internal` as URL for the Subsonic provider.

## Type of Change

- [ ]  Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/music_assistant/providers/opensubsonic/__init__.py
+++ b/music_assistant/providers/opensubsonic/__init__.py
@@ -12,6 +12,7 @@ from music_assistant.constants import CONF_PASSWORD, CONF_PATH, CONF_PORT, CONF_
 from .sonic_provider import (
     CONF_BASE_URL,
     CONF_ENABLE_LEGACY_AUTH,
+    CONF_ALLOW_INSECURE_CONN,
     CONF_ENABLE_PODCASTS,
     CONF_NEW_ALBUMS,
     CONF_OVERRIDE_OFFSET,
@@ -115,6 +116,14 @@ async def get_config_entries(
             label="Enable Legacy Auth",
             required=True,
             description='Enable OpenSubsonic "legacy" auth support',
+            default_value=False,
+        ),
+        ConfigEntry(
+            key=CONF_ALLOW_INSECURE_CONN,
+            type=ConfigEntryType.BOOLEAN,
+            label="Allow Insecure Connection",
+            required=True,
+            description='This will allow you to use self signed certificates',
             default_value=False,
         ),
         ConfigEntry(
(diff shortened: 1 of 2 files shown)


khers:
Thanks for the PR. I believe this option was left in during a large refactor to use the Requests library and now it doesn't actually do anything. So making this work will required changes to py-opensonic as well.

I am not inclined to expose this because there are two alternatives for enabling it and neither is great (TLDR this isn't actually needed, if you have a home network, just use HTTP, otherwise just use Let's Encrypt).

1. We make this option disable certificate verification completely. If this is the route, why use HTTPS at all? It's doesn't buy you any security to allow for encrypted data if you will accept any cert from a server, including ones that are not for the domain hosting your Open Subsonic server.
2. We add the ability to include the root of trust chain to each request. This is the way for large corporations who host their own certificate authority internally but don't deploy it well enough to include that in the OS root of trust store. If we enable this, then we need to design a way to plumb that through MA (another provider option isn't quite it because you need to have the PEM file in the MA container, and if your going that route and building your own container, just insert your PEM into the system root of trust and be done). All this extra complexity, for what?

I think it's a great idea to try running your own internal CA, I tried it just to see what needed to change in my workflows to support it and how the whole system works. However, supporting this use case adds a lot of complexity to MA for no real benefit. If your Open Subsonic is accessible via your LAN or a VPN, just use HTTP. If you go over the internet, just use Let's Encrypt for your certificates as they are in almost all major distro roots of trust.

Tooa:
Thanks for your reply!

> I believe this option was left in during a large refactor to use the Requests library and now it doesn't actually do anything. So making this work will required changes to py-opensonic as well.

I was not aware that the `insecure` parameter is not functional and was overlooked in the refactoring. I get your point.

> We add the ability to include the root of trust chain to each request. This is the way for large corporations who host their own certificate authority internally but don't deploy it well enough to include that in the OS root of trust store. [...]

I tried mounting the OS root of trust containing the self-signed root CA in the container before creating this PR. It did not work.

```yaml
volumes:
  - /etc/ssl/certs:/etc/ssl/certs:ro
```

I assume this does not work, because MA does not use the container (OS) certificates. Haven't verified this though. By looking at the dependencies, I assume it uses [python-certifi](https://github.com/certifi/python-certifi).

I did a little bit of digging into the Requests library and found the `REQUESTS_CA_BUNDLE` env variable ([documentation](https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification)). I may use the OS CA bundle containing my self-signed root CA `/etc/ssl/certs/ca-certificates.crt` (created via `update-ca-certificates`) or go with  HTTP as you suggested.

Will close this MR then as this is more like support now.  Thank you for your time looking into this :heart:

PR #2810, opened 2025-12-15
Sendspin fixes

maximmaxim345:
Includes a bunch of other fixes from [aiosendspin](https://github.com/Sendspin/aiosendspin/releases/tag/1.1.1).

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -539,8 +539,13 @@ class SendspinPlayer(Player):
 
     async def _on_queue_update(self, event: MassEvent) -> None:
         """Extract and send current media metadata to sendspin players on queue updates."""
+        if self.synced_to is not None:
+            # Only leader sends metadata
+            return
         queue = self.mass.player_queues.get_active_queue(self.player_id)
         if not queue or not queue.current_item:
+            # Clear metadata when queue has no current item
+            self.api.group.set_metadata(Metadata())
             return
 
         current_item = queue.current_item
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -49,6 +49,7 @@ from music_assistant.constants import (
     CONF_ENTRY_HTTP_PROFILE_HIDDEN,
     CONF_ENTRY_OUTPUT_CODEC_HIDDEN,
     CONF_ENTRY_SAMPLE_RATES,
+    CONF_OUTPUT_CHANNELS,
     CONF_OUTPUT_CODEC,
     INTERNAL_PCM_FORMAT,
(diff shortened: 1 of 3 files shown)


music_assistant/providers/sendspin/player.py
```diff
@@ -533,10 +538,11 @@ async def _send_artist_artwork(self, current_item: QueueItem) -> None:
                 # Clear artist artwork if none available
                 await self.api.group.set_media_art(None, source=ArtworkSource.ARTIST)
 
-    async def _on_queue_update(self, event: MassEvent) -> None:
-        """Extract and send current media metadata to sendspin players on queue updates."""
+    async def _send_metadata(self) -> None:
+        """Extract and send current media metadata to sendspin players."""
         queue = self.mass.player_queues.get_active_queue(self.player_id)
         if not queue or not queue.current_item:
+            self.api.group.set_metadata(Metadata())
```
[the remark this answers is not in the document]
maximmaxim345: Its not async.

music_assistant/providers/sendspin/player.py
```diff
@@ -224,12 +226,8 @@ def __init__(self, provider: SendspinProvider, player_id: str) -> None:
             self._attr_volume_level = player_client.volume
             self._attr_volume_muted = player_client.muted
         self._attr_available = True
-        self._on_unload_callbacks.append(
-            self.mass.subscribe(
-                self._on_queue_update,
-                (EventType.QUEUE_UPDATED),
-            )
-        )
+        self._attr_needs_poll = True
```
marcelveldt: meh why ???

music_assistant/providers/sendspin/player.py
```diff
@@ -351,6 +349,13 @@ async def group_event_cb(self, group: SendspinGroup, event: GroupEvent) -> None:
             case GroupDeletedEvent():
                 pass
 
+    async def poll(self) -> None:
+        """Poll player to send metadata updates to Sendspin clients."""
+        if self.synced_to is not None:
+            # Only leader sends metadata
+            return
+        await self._send_metadata()
+
```
marcelveldt: I'm not liking the fact that we now are going to poll when that is really unneeded. everything is event driven. So if we are missing an event somewhere, somehow, we should fix that instead.

maximmaxim345: Agree, removed that. Lets see if it works now (maybe the metadata issues were just in aiosendspin, which I fixed already?).

marcelveldt: Maybe, but we could also be missing one callback in the chain when (only) the player.current_media changes. I will look into that

marcelveldt (approved):

PR #2822, opened 2025-12-15
Add user filter to scrobble providers

marcelveldt:
Allows scrobble providers (lastfm and listenbrainz) to only respond to media played by certain users


marcelveldt:
@OzGav FYI

PR #2825, opened 2025-12-16
Add album artist, year and track number to Sendspin metadata

maximmaxim345:
Extract additional metadata fields from queue items for Sendspin metadata clients:
- Track number from media item
- Album artist and year by fetching full album (since queue items only have simplified `ItemMapping`)

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -570,14 +570,29 @@ class SendspinPlayer(Player):
 
         shuffle = queue.shuffle_enabled if queue else False
 
+        # Extract extra metadata from queue item if available
+        album_artist: str | None = None
+        year: int | None = None
+        track_number: int | None = None
+        if queue_item and (media_item := queue_item.media_item):
+            track_number = getattr(media_item, "track_number", None)
+            # Album on queue item is simplified to ItemMapping, fetch full album for extra metadata
+            if album_mapping := getattr(media_item, "album", None):
+                try:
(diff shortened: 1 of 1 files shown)


music_assistant/providers/sendspin/player.py
```diff
@@ -570,14 +570,29 @@ async def send_current_media_metadata(self) -> None:
...
+        # Extract extra metadata from queue item if available
+        album_artist: str | None = None
+        year: int | None = None
+        track_number: int | None = None
+        if queue_item and (media_item := queue_item.media_item):
+            track_number = getattr(media_item, "track_number", None)
+            # Album on queue item is simplified to ItemMapping, fetch full album for extra metadata
+            if album_mapping := getattr(media_item, "album", None):
+                try:
+                    full_album = await self.mass.music.get_item_by_uri(album_mapping.uri)
```
marcelveldt, line 582: How important is it to have this info ? Because this is potentially a heavy api call being executed. 
Until we have the metadata proxy, we want less api calls, not more.

marcelveldt: suggestion: we should also have a method to get a library item by provider and item id, which is a cheap internal call. In that case you will only get this extra metadata if its an item existing in the lib but that is fine I think

maximmaxim345: Not important, I'll mark this PR as draft for now then.

OzGav:
Hi @maximmaxim345 just checking in about this PR?

PR #2828, opened 2025-12-17
(Roku) Media Assistant provider: bug fixes / cleanup

MedievalApple:
Fixes a few issues with state syncing with the Roku.

Cleans up some try catches to have better logging.

### DIFF
--- a/music_assistant/providers/roku_media_assistant/player.py
+++ b/music_assistant/providers/roku_media_assistant/player.py
@@ -164,7 +164,6 @@ class MediaAssistantPlayer(Player):
 
         logger = self.provider.logger.getChild(self.player_id)
         logger.info("Received PAUSE command on player %s", self.display_name)
-        self._attr_playback_state = PlaybackState.PAUSED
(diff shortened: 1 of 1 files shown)


music_assistant/providers/roku_media_assistant/player.py
```diff
@@ -215,6 +219,7 @@ async def play_media(self, media: PlayerMedia) -> None:
             )
             self._attr_powered = True
             self._attr_current_media = media
+            self._attr_playback_state = PlaybackState.PAUSED
```
marcelveldt: why is this ? That looks weird. Issue a play media command and the state gets set to paused

MedievalApple: My apologies that was left over from when I was debugging some of the bugs with syncing playback states with the Roku. I had thought I pushed the commit that removed those lines before opening the pull request. It is pushed now.

marcelveldt (approved):

PR #2851, opened 2025-12-19
Qobuz enable selection of stream quality level 

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/4538


marcelveldt:
This doesn't satisfy that requirement completely - first question should be what in the chain should be impacted.
If it's about the inbound stream from Qobuz into MA, then yes this will help.
But if its about streaming in the network or remote, then you would need to change MA's codec to players

OzGav:
The OP said it was for slow internet connections so I believe this is the solution?

Edit: They are clear where they say

> streaming to Music Assistant from Qobuz and other HQ providers

marcelveldt (approved):

PR #2852, opened 2025-12-19
Typing fixes for the chromecast provider

OzGav:


music_assistant/providers/chromecast/__init__.py
```diff
@@ -23,20 +23,24 @@
 )  # we don't have any special supported features (yet)
 
 # Monkey patch the Media controller here to store the queue items
+# TODO: Refactor this monkey patch into a proper MediaController subclass or wrapper
+# to avoid runtime attribute injection and `# type: ignore` lines.
+# The current approach dynamically adds `current_item_id` and `items` to MediaStatus,
+# which is not represented in pychromecast's type hints.
```
marcelveldt: remove the ai bs

music_assistant/providers/chromecast/helpers.py
```diff
@@ -171,14 +181,19 @@ def new_connection_status(self, status: ConnectionStatus) -> None:
...
         self.castplayer.on_new_connection_status(status)
 
-    def added_to_multizone(self, group_uuid) -> None:
+    def added_to_multizone(self, group_uuid: str) -> None:
         """Handle the cast added to a group."""
         self.castplayer.logger.debug(
             "%s is added to multizone: %s", self.castplayer.display_name, group_uuid
         )
-        self.new_cast_status(self.castplayer.cc.status)
+        # Check if the status attribute is available before passing it.
```
marcelveldt: remove the ai slob please

OzGav: Removed

music_assistant/providers/chromecast/__init__.py
OzGav: Removed

marcelveldt:
Let's hold off with merging until we dont want to backport any fixes to stable anymore.

music_assistant/providers/chromecast/provider.py
```diff
@@ -103,10 +107,10 @@ def _on_chromecast_discovered(self, uuid: str, _: object) -> None:
 
         assert self.browser is not None  # for type checking
         with self._discover_lock:
-            disc_info: CastInfo = self.browser.devices[uuid]
+            disc_info: CastInfo = self.browser.devices[UUID(uuid)]
 
             if disc_info.uuid is None:
-                self.logger.error("Discovered chromecast without uuid %s", disc_info)
+                self.logger.error("Discovered chromecast without uuid %s", disc_info)  # type: ignore[unreachable]
                 return
```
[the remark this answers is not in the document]
OzGav: In practice, CastInfo.uuid should not be None according to the type hints, but we’ve seen defensive checks like this throughout the Chromecast codebase because discovery data can be incomplete or malformed at runtime (especially during transient zeroconf states). Removing the guard would reduce runtime robustness, while changing upstream CastInfo typing is outside the scope of this PR.

The ignore does not hide a logic bug here:

If uuid is truly non-optional, the branch is effectively dead but harmless.

If uuid can be None at runtime, the guard is correct and the ignore merely compensates for overly strict upstream typing.

Given the constraints of this PR (mypy fixes only, no behavioral changes), suppressing unreachable locally with a comment is the least risky option. A proper fix would require upstream type adjustments in pychromecast, which can be revisited separately.

music_assistant/providers/chromecast/provider.py
```diff
@@ -155,7 +159,12 @@ async def _create_and_register_player(
...
-    def _on_chromecast_removed(self, uuid: str, service: object, cast_info: object) -> None:
+    def _on_chromecast_removed(
+        self,
+        uuid: str,
+        service: tuple[Any, ...],
+        cast_info: object,
+    ) -> None:
         """Handle zeroconf discovery of a removed Chromecast."""
         player_id = str(service[1])
         friendly_name = service[3]
```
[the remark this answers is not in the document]
OzGav: This is a conscious trade-off rather than an oversight.

This handler mirrors how pychromecast has historically invoked remove_callback, where service is a zeroconf service tuple and the positional indices are stable in practice. The earlier cast() is there purely to satisfy typing; it does not change the runtime contract we’re relying on here.

Switching to cast_info for player_id / friendly_name would indeed be cleaner, but that would be a behavioral change, not just a typing fix. This PR is intentionally scoped to mypy cleanup only, and altering which source of truth we use for removal events could have subtle side effects (e.g. differences between service-level removal and cast-info state).

music_assistant/providers/chromecast/player.py
```diff
@@ -242,7 +242,7 @@ async def _on_sendspin_player_event(self, event: MassEvent) -> None:
                     or sendspin_player.synced_to
                 )
             else:
-                self._attr_synced_to = None
+                self._attr_synced_to = ""
```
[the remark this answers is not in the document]
OzGav: This is fair. I have changed this to add a defintiion in __init__

music_assistant/providers/chromecast/provider.py
```diff
@@ -103,10 +107,10 @@ def _on_chromecast_discovered(self, uuid: str, _: object) -> None:
 
         assert self.browser is not None  # for type checking
         with self._discover_lock:
-            disc_info: CastInfo = self.browser.devices[uuid]
+            disc_info: CastInfo = self.browser.devices[UUID(uuid)]
```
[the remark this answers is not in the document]
OzGav: I don't exactly like Copilot suggestion but there is a possiblle behaviour change here. So line is reverted and typing is fixed in the function definition to satisfy mypy

music_assistant/providers/chromecast/provider.py
```diff
@@ -6,7 +6,9 @@
 import contextlib
 import logging
 import threading
-from typing import TYPE_CHECKING, cast
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Any, cast
+from uuid import UUID
```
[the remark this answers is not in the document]
OzGav: I am not seeing ruff complaining here

music_assistant/providers/chromecast/provider.py
```diff
@@ -49,9 +51,11 @@ def __init__(
...
         self.browser = CastBrowser(
             SimpleCastListener(
-                add_callback=self._on_chromecast_discovered,
-                remove_callback=self._on_chromecast_removed,
-                update_callback=self._on_chromecast_discovered,
+                add_callback=cast("Callable[[UUID, str], None]", self._on_chromecast_discovered),
+                remove_callback=cast(
+                    "Callable[[UUID, str, CastInfo], None]", self._on_chromecast_removed
+                ),
+                update_callback=cast("Callable[[UUID, str], None]", self._on_chromecast_discovered),
```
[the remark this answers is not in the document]
OzGav: I’ve updated the type hints and added some cast calls to the Chromecast provider to get mypy passing. I made sure to only touch the signatures and type annotations, so the actual logic and behavior haven't changed at all. I also aligned the callback arguments with what pychromecast sends to keep the static analysis happy.

music_assistant/providers/chromecast/player.py
```diff
@@ -283,7 +283,7 @@ async def on_unload(self) -> None:
             # Non-blocking disconnect: close socket, don't wait for thread.
             # Socket threads are daemon threads and die on process exit.
             # Blocking disconnect can stall shutdown if threads are slow to exit.
-            self.cc.disconnect(blocking=False)
+            self.cc.disconnect(0.001)
```
MarvinSchenkel: Why not `0`?

music_assistant/providers/chromecast/provider.py
```diff
@@ -55,9 +56,9 @@ def __init__(
         manual_ip_config = cast("list[str]", config.get_value(CONF_ENTRY_MANUAL_DISCOVERY_IPS.key))
         self.browser = CastBrowser(
             SimpleCastListener(
-                add_callback=self._on_chromecast_discovered,
-                remove_callback=self._on_chromecast_removed,
-                update_callback=self._on_chromecast_discovered,
+                add_callback=cast("Any", self._on_chromecast_discovered),
```
MarvinSchenkel: Changing the type of the `_` parameter of `_on_chromecast_discovered` to `str` should allow us to drop these casts I think?

music_assistant/providers/chromecast/provider.py
```diff
@@ -192,9 +193,13 @@ async def _create_and_register_player(
         finally:
             self._pending_discoveries.discard(player_id)
 
-    def _on_chromecast_removed(self, uuid: str, service: object, cast_info: object) -> None:
+    def _on_chromecast_removed(
+        self,
+        uuid: UUID,
+        service: str,
+        cast_info: Any,
```
MarvinSchenkel: `CastInfo` ?

music_assistant/providers/chromecast/provider.py
```diff
@@ -192,9 +193,13 @@ async def _create_and_register_player(
...
+    ) -> None:
         """Handle zeroconf discovery of a removed Chromecast."""
-        player_id = str(service[1])
-        friendly_name = service[3]
-        self.logger.debug("Chromecast removed: %s - %s", friendly_name, player_id)
+        player_id = str(uuid)
+        self.logger.debug("Chromecast removed: %s", player_id)
```
MarvinSchenkel: Why remove the player name here?

music_assistant/providers/chromecast/helpers.py
```diff
@@ -94,16 +95,21 @@ def fill_out_missing_chromecast_info(self, zconf: Zeroconf) -> None:
 
         # Get MAC address for device matching (not available for groups)
         if self.mac_address is None and self.cast_type != "group":
-            self.mac_address = get_mac_address(self.services, zconf)
+            self.mac_address = get_mac_address(list(self.services), zconf)
 
 
-def get_multizone_info(services: list[ServiceInfo], zconf: Zeroconf, timeout=30):
+def get_multizone_info(
+    services: list[ServiceInfo],
```
MarvinSchenkel: If we type this as a `set` we can drop all the conversions between list > set. It's already typed as `set` on line 36.

music_assistant/providers/chromecast/helpers.py
```diff
@@ -137,8 +143,9 @@ def get_mac_address(services: list[ServiceInfo], zconf: Zeroconf, timeout: int =
     :return: MAC address string or None if not available.
     """
     try:
+        services_set = cast("set[HostServiceInfo | MDNSServiceInfo]", set(services))
```
MarvinSchenkel: Also for these ones, if this is the true type of the set of service, let's type is like this everywhere. It looks like it is by lookg at the chromecast dataclas of `CastInfo`, it is typed as `set[HostServiceInfo | MDNSServiceInfo]`

MarvinSchenkel (changes requested):
Please have a look at my comments

music_assistant/providers/chromecast/player.py
OzGav: It was a defensive measure against truthiness bugs but I just looked at the pychromecast library and it says specifically to set 0 to not block so I have changed this.

music_assistant/providers/chromecast/provider.py
OzGav: Yes you are right. This has been fixed.

music_assistant/providers/chromecast/provider.py
OzGav: Yes just saw it when doing the above. Fixed

music_assistant/providers/chromecast/provider.py
OzGav: The original code was treating service as a tuple and indexing it with service[1] and service[3], but looking at the pychromecast callback signature it's actually a str. I think this would have crashed at runtime if it ever got called? Anyway, I initially just removed the friendly_name logging to fix the type error, but now that we're using CastInfo properly we can get it from cast_info.friendly_name instead. Have restored the logging now.

music_assistant/providers/chromecast/helpers.py
OzGav: Correct again. I have no reason for this other than didnt notice....

music_assistant/providers/chromecast/helpers.py
OzGav: I agree, aligning with the CastInfo types makes much more sense. I’ve updated the signatures in helpers.py to use set[HostServiceInfo | MDNSServiceInfo] directly. This allowed me to remove the manual casts and list-to-set conversions, keeping the typing consistent with pychromecast

MarvinSchenkel (approved):

PR #2855, opened 2025-12-19
Fix WiiM devices not starting as part of a group

MarvinSchenkel:
WiiM devices do not respond well to `pause_for` commands, resulting in them pausing forever. This PR changes the `pause_for` to a 'pause` + `unpause_at` which fixes WiiM devices not starting as part of a group.

Another issue with WiiM devices seems to be that they ignore Autostart = 0, meaning they will just start playback as soon as they signal a `STMl (buffer ready)`. This currently results in the WiiM having to pause for +- 5s when group playback starts, which is not great, but at least we have group playback working now.

Workaround for https://github.com/music-assistant/support/issues/4506


marcelveldt (approved):

darrenfreeman:
I have raised a support ticket with WiiM, regarding these implementation issues.

PR #2861, opened 2025-12-19
fix: podcast parser helpers not handling exception

fmunkes:
Reported on discord. No idea why the function did not take care of the exception.

### DIFF
--- a/music_assistant/helpers/podcast_parsers.py
+++ b/music_assistant/helpers/podcast_parsers.py
@@ -6,6 +6,7 @@ from typing import Any
 
 import aiohttp
 import podcastparser
+from aiohttp.client import ClientError
 from music_assistant_models.enums import ContentType, ImageType, MediaType
 from music_assistant_models.media_items import (
     AudioFormat,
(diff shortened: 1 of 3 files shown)


music_assistant/helpers/podcast_parsers.py
```diff
@@ -32,7 +33,11 @@ async def get_podcastparser_dict(
     # but, reports on discord show, that also the opposite may be true
     for headers in [{"User-Agent": "Mozilla/5.0"}, {}]:
         # raises ClientError on status failure
-        response = await session.get(feed_url, headers=headers, raise_for_status=True)
+        try:
+            response = await session.get(feed_url, headers=headers, raise_for_status=True)
+        except ClientError:
+            continue
+        break
     assert response is not None  # for type checking
```
marcelveldt: this is going to fail if you didnt got a response at all

fmunkes: you're right. I'm raising a MediaNotFound now. For itunes and the rss feed we can just raise this to the frontend, as the method is only called for single feeds. For gpodder during library sync we continue logging inaccessible podcasts.

marcelveldt (approved):

PR #2867, opened 2025-12-22
Add metadata support to HLS streams

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/4450

maybe this one? 
https://github.com/orgs/music-assistant/discussions/455

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -3,13 +3,13 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import logging
 import os
 import re
 import struct
 import time
 from collections.abc import AsyncGenerator
+from functools import partial
 from io import BytesIO
 from typing import TYPE_CHECKING, Final, cast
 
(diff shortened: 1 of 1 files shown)


music_assistant/helpers/audio.py
```diff
@@ -536,6 +537,12 @@ async def get_media_stream(
         substream = await get_hls_substream(mass, streamdetails.path)
         audio_source = substream.path
         if streamdetails.media_type == MediaType.RADIO:
+            # Start background task to monitor HLS playlist for metadata updates
+            if "hls_monitor_task" not in streamdetails.data:
```
marcelveldt: why not use the new callback we created for this purpose ?

OzGav: For no good reason so changed to use callback

music_assistant/helpers/audio.py
```diff
@@ -322,6 +323,12 @@ async def get_stream_details(
             resolved_url, stream_type = await resolve_radio_stream(mass, streamdetails.path)
             streamdetails.path = resolved_url
             streamdetails.stream_type = stream_type
+            # Set up metadata monitoring callback for HLS radio streams
+            if stream_type == StreamType.HLS:
+                streamdetails.stream_metadata_update_callback = partial(
+                    _update_hls_radio_metadata, mass, resolved_url
```
MarvinSchenkel: Just checking. Could we get rid of the `resolved_url` parameter by simply using `streamdetails.path` inside the `_update_hls_radio_metadata` function? Other than that, looks good to me.

OzGav: I think you are right. I have removed that.

MarvinSchenkel (approved):

PR #2868, opened 2025-12-22
Add support for legacy Shoutcast servers using ICY protocol

OzGav:
Adds metadata extraction support for legacy Shoutcast v1 servers that respond with `ICY 200 OK` instead of the standard `HTTP/1.1 200 OK`

Satisfies https://github.com/orgs/music-assistant/discussions/3333

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "ifaddr==0.2.0",
   "getmac==0.9.5",
   "mashumaro==3.18",
-  "music-assistant-frontend==2.17.117",
+  "music-assistant-frontend==2.17.118",
   "music-assistant-models==1.1.107",
   "mutagen==1.47.0",
   "orjson==3.11.5",
--- a/requirements_all.txt
+++ b/requirements_all.txt
@@ -41,7 +41,7 @@ liblistenbrainz==0.7.0
 librosa==0.11.0
 lyricsgenius==3.7.5
 mashumaro==3.18
-music-assistant-frontend==2.17.117
+music-assistant-frontend==2.17.118
 music-assistant-models==1.1.107
 mutagen==1.47.0
 niconico.py-ma==2.1.0.post1


marcelveldt:
This seems an overly complex solution for a simple problem and also I see a lot of code duplication.
Cant we just assume ICY support from the http headers alone ?
Or otherwise override aiohttp to support the ICY response ?

If unavoidable to do the manual parsing, lets try to prevent the duplicate code as much as we can

OzGav:
I simplified the ICY metadata handling by removing the redundant stream detection code and consolidating the duplicate metadata parsing logic into a single helper function. Instead of pre-checking if a stream is ICY-compatible with a separate socket connection, we now just try the normal HTTP request first and only fall back to raw socket handling if aiohttp fails to parse the response (which happens with old Shoutcast servers that return "ICY 200 OK" instead of "HTTP/1.1 200 OK"). The metadata parsing code that was duplicated in both the normal and fallback paths is now in one reusable function.

music_assistant/helpers/audio.py
```diff
@@ -768,9 +797,15 @@ async def resolve_radio_stream(mass: MusicAssistant, url: str) -> tuple[str, Str
             except IsHLSPlaylist:
                 stream_type = StreamType.HLS
 
-    except Exception as err:
-        LOGGER.warning("Error while parsing radio URL %s: %s", url, str(err))
-        return (url, stream_type)
+    except aiohttp.ClientError as err:
+        # aiohttp can't parse legacy "ICY 200 OK" responses from old Shoutcast servers
```
marcelveldt: Not sure if this is safe to assume ?
What about a 404 or 403 error etc. Can we try to handle this a tiny bit more specific OR do a double check if this is an icy stream

music_assistant/helpers/audio.py
```diff
@@ -768,9 +797,15 @@ async def resolve_radio_stream(mass: MusicAssistant, url: str) -> tuple[str, Str
...
+        # This is expected behavior, not an error - assume ICY and let streaming function handle it
+        LOGGER.debug(
+            "Could not resolve radio URL with aiohttp %s: %s - assuming ICY stream", url, err
+        )
+        return (url, StreamType.ICY)
```
marcelveldt: So, thinking about this a little bit more I actually think we should add an additional StreamType to the enum.... Have one StreamType.HTTP_ICY and StreamType.ICY (for these legacy ones)

or add a StreamType.SHOUTCAST ? then we can steer into the right handler

OzGav: Let me know what you think is best.

marcelveldt: What do you think ? The above message was more thinking out loud, hoping to spark a bit of discussion ;-)

OzGav: Yeah I agree. It will be better to identify the stream early in resolve_radio_stream and then branch appropriately in get_media_stream. I have opened a PR on the models repo to add a new SHOUTCAST stream type and once that is merged I will rework this

music_assistant/helpers/audio.py
OzGav: Fixed with new stream type

music_assistant/helpers/audio.py
OzGav: Have done the refactor but can’t fix the lint problems until I get home so marked as draft.

MarvinSchenkel:
We are going to investigate if we can implement a PR to the parser of `aiohttp` that also

1. Add ICY 200 to llhttp
2. Bump llhttp in aiohttp
3. Adjust the python parser in aiohttp to also accept ICY 200

I will try to do this and see if this could work. This would basically make this entire PR obsolete and fix the issue at the root :)

PR #2871, opened 2025-12-22
Add Bandcamp Music Provider

ALERTua:
# Bandcamp Music Provider

This PR adds support for Bandcamp Music Provider based on `bandcamp_async_api` Python library.

## API

- This Provider is based on [bandcamp_async_api](https://github.com/ALERTua/bandcamp_async_api) that I developed specifically for this.
- The API library has extensive tests both using mocks and the real data and covers everything the API touches.
- I encountered zero 5** or 429 errors during my testing, so I did not need to implement rate limiting. But I can if you need me to.

## Provider

- `ProviderFeature.BROWSE` would cause an immense load on the API.
- Alas, there are no Bandcamp API endpoints for `ProviderFeature.RECOMMENDATIONS` and `ProviderFeature.SIMILAR_TRACKS`.
- I can expand the code comments to further eliminate the Bus Factor, as I am from Kyiv, Ukraine.
- Please forgive me if you find my code imperfect, as I am not a real programmer.
- Discussion thread: https://github.com/orgs/music-assistant/discussions/4611
- You can test it using `d5369777_music_assistant_dev` addon with this configuration:
```yaml
server_repo: ALERTua/fork_music_assistant_server@bc
```
- You need the Bandcamp identity cookie only to browse your library. All other operations (search) do not need the identity.

## Documentation

- Documentation [PR](https://github.com/music-assistant/beta.music-assistant.io/pull/64)

## Maintaining the Provider Code in the Future

- I will happily maintain this provider and the code related to it in the future
- I will happily monitor and review other pull requests in case they are related to the code of this provider
- I would appreciate creating a label for this provider that I can subscribe to, which allows me to filter related issues

## Digressions

edit: PR for most of these items: https://github.com/music-assistant/server/pull/2871

- Music Assistant Server can be developed on Windows, or, should I say, on any OS. But there are caveats that can be easily fixed:
  - The pre-commit config uses `scripts/run-in-env.sh`. This is unneeded due to `uv`. `uv run pre-commit run` executes the same calls but on any OS.
  - The pre-commit config uses `python3 -m ...` which should also be substituted by `uv run -m ...`
  - Many pre-commit checks can be replaced with the "official" ones from [uv](https://github.com/astral-sh/uv-pre-commit), [ruff](https://github.com/astral-sh/ruff-pre-commit), [pre-commit](https://github.com/pre-commit/pre-commit-hooks), [mypy](https://github.com/pre-commit/mirrors-mypy).

<details>
  <summary>(Spoiler) My resulting working pre-commit config that does everything that the original one does, but on any OS</summary>
  
```yaml
repos:
  - repo: local
    hooks:
      # gen_requirements_all modifies files so it has to be first
      - id: gen_requirements_all
        name: gen_requirements_all
        entry: uv run -m scripts.gen_requirements_all
        pass_filenames: false
        language: system
        types: [text]
        files: ^(music_assistant/.+/manifest\.json|pyproject\.toml|\.pre-commit-config\.yaml|scripts/gen_requirements_all\.py)$

      - id: ruff-check
        name: 🐶 Ruff Linter
        language: system
        types: [python]
        entry: uv run ruff check --fix
        require_serial: true
        stages: [pre-commit, pre-push, manual]

      - id: ruff-format
        name: 🐶 Ruff Formatter
        language: system
        types: [python]
        entry: uv run ruff format
        require_serial: true
        stages: [pre-commit, pre-push, manual]

      - id: check-ast
        name: 🐍 Check Python AST
        language: system
        types: [python]
        entry: uv run check-ast

      - id: check-case-conflict
        name: 🔠 Check for case conflicts
        language: system
        entry: uv run check-case-conflict

      - id: check-docstring-first
        name: ℹ️  Check docstring is first
        language: system
        types: [python]
        entry: uv run check-docstring-first

      - id: check-executables-have-shebangs
        name: 🧐 Check that executables have shebangs
        language: system
        types: [text, executable]
        entry: uv run check-executables-have-shebangs
        stages: [pre-commit, pre-push, manual]

      - id: check-json
        name: ｛ Check JSON files
        language: system
        types: [json]
        entry: uv run check-json
        files: ^(music_assistant/.+/manifest\.json)|(tests/providers/.+/fixtures/.+\.json)$

      - id: check-merge-conflict
        name: 💥 Check for merge conflicts
        language: system
        types: [text]
        entry: uv run check-merge-conflict

      - id: check-symlinks
        name: 🔗 Check for broken symlinks
        language: system
        types: [symlink]
        entry: uv run check-symlinks

      - id: check-toml
        name: ✅ Check TOML files
        language: system
        types: [toml]
        entry: uv run check-toml

      - id: codespell
        name: ✅ Check code for common misspellings
        language: system
        types: [text]
        entry: uv run codespell

      - id: detect-private-key
        name: 🕵️ Detect Private Keys
        language: system
        types: [text]
        entry: uv run detect-private-key

      - id: end-of-file-fixer
        name: ⮐ Fix End of Files
        language: system
        types: [text]
        entry: uv run end-of-file-fixer
        stages: [pre-commit, pre-push, manual]

      - id: no-commit-to-branch
        name: 🛑 Don't commit to stable branch
        language: system
        entry: uv run no-commit-to-branch
        pass_filenames: false
        always_run: true
        args:
          - --branch=stable

      - id: trailing-whitespace
        name: ✄ Trim Trailing Whitespace
        language: system
        types: [text]
        entry: uv run trailing-whitespace-fixer
        stages: [pre-commit, pre-push, manual]

      - id: mypy
        name: mypy
        entry: uv run mypy
        language: system
        types: [python]
        require_serial: true
        pass_filenames: false
```
</details>

  - In pre-commit fixes, the end of lines is taken from the system. It should be fixed as `LF` in `pyproject.toml` and `.gitattributes` to allow for development on any OS.
  - WSL2 is not needed to either develop or test the project. Everything I tested works natively on Windows.
  - I could try implementing all this if you want in a separate PR.
- The documentation does not have templates for pages. If I want to add a provider or player, I have to copy the existing page and edit it. 
- I am happy with Music Assistant, and I am happy to be useful in expanding this great project.
- I am thrilled with SendSpin protocol, and I already use it everywhere I can on my Chromecast devices. Thank you for this <3


ALERTua:
OK, I forgot `mypy`. Fixing all issues. Sorry.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,404 @@
...
+SUPPORTED_FEATURES = {
+    ProviderFeature.LIBRARY_ARTISTS,
+    ProviderFeature.LIBRARY_ALBUMS,
+    ProviderFeature.LIBRARY_TRACKS,
+    # ProviderFeature.BROWSE,  # TODO: Consider
+    ProviderFeature.SEARCH,
+    ProviderFeature.ARTIST_ALBUMS,
+    ProviderFeature.ARTIST_TOPTRACKS,
+    # ProviderFeature.RECOMMENDATIONS,  # TODO: Consider
+    # ProviderFeature.SIMILAR_TRACKS,  # TODO: Consider
```
pljones: Could you use this for a "Don't stop the music" feature?  Just keep taking the first "similar track" that hasn't been played, storing the remainder (up to some limit)?

Recommendations - I guess you _could_ treat these as another "Don't stop the music" source and randomly pick one, if they're available, instead of a "similar track".

ALERTua: Alas, I have no idea where to get recommendations or similar tracks. More on the API: https://github.com/michaelherger/Bandcamp-API

pljones: Ah, right -- I thought you had a source for these features but no implementation.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,411 @@
...
+
+        # noinspection PyBroadException
+        try:
+            api_collection = await self._client.get_collection_items(CollectionType.COLLECTION)
+            for item in api_collection.items:
+                if item.item_type == "album":
+                    # noinspection PyArgumentList
+                    yield await self.get_album(f"{item.band_id}-{item.item_id}")
+                    await asyncio.sleep(0)  # Yield control to avoid blocking
+        except BandcampMustBeLoggedInError:
```
OzGav: Why not use the MA “LoginFailed” or “Invalidtoken” error here?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,411 @@
...
+            self.logger.error("Wrong Bandcamp identity token.")
+            return
+        except Exception:
```
OzGav: Generally it is desirable to work out what errors you are trying to catch instead of this broad catch-all

music_assistant/providers/bandcamp/__init__.py
ALERTua: Will do.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,411 @@
...
+        # Initialize the new async API client
+        self._client = BandcampAPIClient(session=self.mass.http_session, identity_token=identity)
+
+        self.throttler = ThrottlerManager(rate_limit=1, period=2)
+        self._converters = BandcampConverters(self.domain, self.instance_id)
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        return False
```
OzGav: How is this not a streaming provider?

music_assistant/providers/bandcamp/__init__.py
ALERTua: OK. On my way, changing all broad exceptions.

music_assistant/providers/bandcamp/__init__.py
ALERTua: Changing to True.

music_assistant/providers/bandcamp/__init__.py
ALERTua: Done. Revamped all exception catching and raising. I hope you are pleased with this implementation. If you aren't — please let me know and I'll improve it <3

music_assistant/providers/bandcamp/__init__.py
ALERTua: No more broad exceptions. I am sorry for that.

OzGav:
@ALERTua there is a test failure unfortunately

ALERTua:
I am sorry, I don't know how this happened. On my way to fix the tests v_v

ALERTua:
@OzGav the tests are fixed. Sorry I forgot to run tests after a refactoring v_v.

OzGav:
Looks pretty good. The only thing I would say is quite a few ignores being used which we don’t see in the other providers so it would be better to properly type the code (other than the arguments in get_config_entries which are pretty much always unused and ignored)

ALERTua:
Done. Thank you for your feedback.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    return (
+        ConfigEntry(
+            key=CONF_IDENTITY,
+            type=ConfigEntryType.SECURE_STRING,
+            label="Identity token",
+            required=False,
+            description="Identity token from Bandcamp cookies for account collection access."
+            " Log in https://bandcamp.com and extract identity from browser cookies.",
```
pljones: Could this either state whether _all cookies_ or a _specific named_ cookie is used, or else link to a page where that information is provided, please.

ALERTua:
```
Identity token from Bandcamp cookies for account collection access.
Log in https://bandcamp.com and extract browser cookie named 'identity'.
```
Done. Better now? Or should I rephrase it somehow?

ALERTua: Note that there will be a documentation after its [PR](https://github.com/music-assistant/beta.music-assistant.io/pull/64/changes) is merged.

OzGav: Yeah you dont need to get too into the detail here as the docs will cover these fine grain questions

pljones: Thanks!

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+# noinspection PyTypeHints,PyUnusedLocal
+async def get_config_entries(
+    mass: MusicAssistant,  # noqa: ARG001
+    instance_id: str | None = None,  # noqa: ARG001
+    action: str | None = None,  # noqa: ARG001
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    return (
+        ConfigEntry(
```
MarvinSchenkel, line 79: Would it be possible to implement OAuth? Would be a huge benefit to the user. For example, the Spotify provider uses this so you can have a look how this would work.

At first glance, it does look like the Bandcamp API [support OAuth](https://bandcamp.com/developer)?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+        return True
+
+    @use_cache(CACHE)
+    @throttle_with_retries
+    async def search(
+        self, search_query: str, media_types: list[MediaType], limit: int | None = None
+    ) -> SearchResults:
+        """Perform search on music provider."""
+        if limit is None:
+            limit = self.search_limit
```
MarvinSchenkel: No need for configuring / setting this limit. MA. This is actually typed as `limit: int = 5` in the base class. Also, our MusicController calls this with a default of 25, so no need to limit this yourself.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+            return results
+
+        try:
+            search_results = await self._client.search(search_query)
+        except BandcampNotFoundError as error:
+            raise MediaNotFoundError("No results for Bandcamp search") from error
+        except BandcampAPIError as error:
+            raise InvalidDataError("Unexpected error during Bandcamp search") from error
+
+        for item in search_results[:limit]:
```
MarvinSchenkel, line 156: Search results will be automatically limited for you already

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+
+        results = SearchResults()
+        if not self._client.identity:
+            return results
+
+        if not media_types:
+            return results
+
+        try:
+            search_results = await self._client.search(search_query)
```
MarvinSchenkel, line 150: Maybe introduce a `limit` parameter in the client?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+    action: str | None = None,  # noqa: ARG001
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    return (
+        ConfigEntry(
+            key=CONF_IDENTITY,
+            type=ConfigEntryType.SECURE_STRING,
+            label="Identity token",
+            required=False,
```
MarvinSchenkel, line 83: I think this is a required config item? Or are there parts of Bandcamp that do not require auth?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+            description="Identity token from Bandcamp cookies for account collection access."
+            " Log in https://bandcamp.com and extract browser cookie named 'identity'.",
+            value=values.get(CONF_IDENTITY) if values else None,
+        ),
+        ConfigEntry(
```
MarvinSchenkel: I think we can remove this (see my other comment in the search)

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -0,0 +1,431 @@
...
+            key=CONF_SEARCH_LIMIT,
+            type=ConfigEntryType.INTEGER,
+            label="Search items limit",
+            required=False,
+            description="Search items limit for one search.",
+            value=values.get(CONF_SEARCH_LIMIT) if values else DEFAULT_SEARCH_LIMIT,
+            default_value=DEFAULT_SEARCH_LIMIT,
+        ),
+        ConfigEntry(
```
MarvinSchenkel, line 88: As I understand it, Bandcamp does not have an endpoint for this and you are using the 'recent tracks' for this? Then I would put this under the 'advanced' category, since you already picked a sane default for this.

```
category="advanced",
```

MarvinSchenkel:
Few comments from my side, but all in all, awesome job on this new music provider <3. 

As for your pre-commit fixes. Feel free to open a separate PR for this :).

music_assistant/providers/bandcamp/__init__.py
ALERTua: @MarvinSchenkel sorry, I am struggling to understand your request.
- As far as I understand, the goal of the OAuth procedure is to get a token to use it to authenticate on the API requests. Bandcamp developer API covers music labels' and music performers' marketing methods, but not the consumer methods https://bandcamp.com/developer/account
- The consumer API **account library** methods authentication is performed using an identity token that one can get only using the `identity` cookie.
- The consumer API **search** methods do not require authentication.

What is the goal and the benefit for implementing the OAuth then?

music_assistant/providers/bandcamp/__init__.py
ALERTua: It is not.
- The consumer API **account library** methods authentication is performed using an identity token that one can get only using the identity cookie. User can have an empty library or choose to use only search or choose not to sync their library, in these cases they do not need an identity token to use the Bandcamp provider.
- The consumer API **search methods** do not require authentication.

music_assistant/providers/bandcamp/__init__.py
ALERTua: Alas, I am not sure the reverse engineered method that I took from https://github.com/impliedchaos/mopidy-bandcamp/blob/master/mopidy_bandcamp/bandcamp.py#L182 has the limit argument. And it would not be wise, as the search query returns all types of objects in a list of results. If you search for "stoner", it will return both artist "Stoner Rebel" and an album "Stoner Guard" and a song "Stoner Machine". If you want just artists with name that contains the query, and you limit the result size, you can get only albums but not an artist that exists :)
The Bandcamp search seems to be limited with 50 items by default.

music_assistant/providers/bandcamp/__init__.py
ALERTua: Alas, I have not found any API endpoints that get related or similar or top items for an entry, so top tracks is basically just first N tracks album by album.
Moving to "advanced" category.

music_assistant/providers/bandcamp/__init__.py
ALERTua: Not in this case, not by Music Assistant. But the Bandcamp search API endpoint seems to limit the reply with 50 items, so I am removing `[:limit]` here.

music_assistant/providers/bandcamp/__init__.py
ALERTua: done.

music_assistant/providers/bandcamp/__init__.py
MarvinSchenkel: I don't think there's a benefit then. Just wanted to make sure if there was a way to get this provider to work _without_ users needing to rip their own cookie (which will also expire). OAuth is simply a much better user experience than obtaining cookies every so often.

If you so there's no way to do this for the account library methods, then that's unfortunate, but something the user will need to do to get this provider to work. In that case, we will need to properly document this for the end user.

music_assistant/providers/bandcamp/__init__.py
MarvinSchenkel: In that case, I think we still need to limit the `search_results` themselves. (so reintroduce `for item in search_results[:limit]:`)

music_assistant/providers/bandcamp/__init__.py
MarvinSchenkel: Just had a quick look at the account link you sent. But they do seem to accept `Authentication: {Bearer token}` headers? In that case, following the instructions [here](https://bandcamp.com/developer) should just give you an access token (that can be used as bearer token) to authenticate requests to the account api (e.g. swap `Cookie` for the `Authentication` header [here](https://github.com/impliedchaos/mopidy-bandcamp/blob/c99d2997d15a2a7aaba636602d9941cfb8979457/mopidy_bandcamp/bandcamp.py#L128))? Or am I missing something?

ALERTua: The identity cookie does not expire. Bandcamp is very old school :)

> Just had a quick look at the account link you sent. But they do seem to accept `Authentication: {Bearer token}` headers? In that case, following the instructions [here](https://bandcamp.com/developer) should just give you an access token (that can be used as bearer token) to authenticate requests to the account api (e.g. swap `Cookie` for the `Authentication` header [here](https://github.com/impliedchaos/mopidy-bandcamp/blob/c99d2997d15a2a7aaba636602d9941cfb8979457/mopidy_bandcamp/bandcamp.py#L128))? Or am I missing something?

Looking at the developer documentation, "To register for API use, [contact us](https://bandcamp.com/contact?subj=API%20Access), and include in your email a brief description of how you intend to use API access.".
I am intending to scrap your search contents and mp3 data for an unknown number of users all around the world :D I don't see this possible :D If they were to freely give the tokens, then maybe, but this would mean that they at least control the quantity of requests from each API user. But they don't :D For now, they just shovel the money (traffic) without even 429 errors.
May I propose we return to this when and if there are any problems with users' authentication? :)

music_assistant/providers/bandcamp/__init__.py
ALERTua: This would lower the search results limit to from 50 to 5. May I at least introduce a hardcoded limit=50 then for this method? or return the config value with 50 as a default? :)

pljones: Looking at the requests, there `/api/bcsearch_public_api/1/autocomplete_elastic` parameter of `search_filter`.

| Tab | `search_filter` |
|--|--|
| All | empty |
| Artists & labels | `b` |
| Albums | `a` |
| Tracks | `t` |
| Fans | `f` |

so you should be able to provide those if the API you use is similar.

MarvinSchenkel: > This would lower the search results limit to from 50 to 5. May I at least introduce a hardcoded limit=50 then for this method? or return the config value with 50 as a default? :)

We actually use a default of 25 in the [music controller](https://github.com/music-assistant/server/blob/fca8a423c433aa89c2399ff1e541de1ea09a2d5b/music_assistant/controllers/music.py#L261). The 5 is just a default value when nothing gets passed. By actually using this parameter you will allow users to narrow down the search results.

ALERTua: > so you should be able to provide those if the API you use is similar.

The [library](https://github.com/ALERTua/bandcamp_async_api) does not use the `autocomplete_elastic`, but [/fuzzysearch/1/app_autocomplet](https://github.com/ALERTua/bandcamp_async_api/blob/main/src/bandcamp_async_api/client.py#L124)


> We actually use a default of 25 in the [music controller](https://github.com/music-assistant/server/blob/fca8a423c433aa89c2399ff1e541de1ea09a2d5b/music_assistant/controllers/music.py#L261). The 5 is just a default value when nothing gets passed. By actually using this parameter you will allow users to narrow down the search results.

OK, reintroducing `for item in search_results[:limit]:` then.

MarvinSchenkel (approved):
Awesome addition to the music providers, thanks a lot @ALERTua  🎉

music_assistant/providers/bandcamp/__init__.py
pljones: > > so you should be able to provide those if the API you use is similar.
> 
> The [library](https://github.com/ALERTua/bandcamp_async_api) does not use the `autocomplete_elastic`, but [/fuzzysearch/1/app_autocomplet](https://github.com/ALERTua/bandcamp_async_api/blob/main/src/bandcamp_async_api/client.py#L124)

That's right.  Did you check whether it takes a similar parameter?  You could try it with `curl` or in the browser.

ALERTua: > That's right. Did you check whether it takes a similar parameter? You could try it with `curl` or in the browser.

I did not. I'm not a reverse engineer. I would rather use a solution that is proven to work rather than rely on one I found myself.

PR #2873, opened 2025-12-23
Fix items not showing up in the library (filtering still not right)

arturpragacz:
Fix items not showing up in the library.

This PR reverts (partially) 1b60a3c (filtering is still not right).


MarvinSchenkel (approved):

MarvinSchenkel:
Thanks @arturpragacz , will discuss this with Marcel when he gets back from holiday. For now, I think this is fine to merge as it unblocks some use cases.

PR #2875, opened 2025-12-23
Audible: Fix authentication for new API token format

ztripez:
- Add compatible token refresh handling for new actor_access_token format
- Validate signing auth availability (preferred, stable auth method)
- Accept multiple authorization code parameter names in callback URL
- Improve error handling and diagnostic messages

Note: The python audible library hasn't been updated since 2024.
See https://github.com/mkb79/Audible/issues/832 for upstream issue.

### DIFF
--- a/music_assistant/providers/audible/audible_helper.py
+++ b/music_assistant/providers/audible/audible_helper.py
@@ -13,13 +13,15 @@ from collections.abc import AsyncGenerator
 from contextlib import suppress
 from datetime import UTC, datetime, timedelta
 from os import PathLike
-from typing import Any
+from typing import TYPE_CHECKING, Any
 from urllib.parse import parse_qs, urlparse
 
 import audible
 import audible.register
-import httpx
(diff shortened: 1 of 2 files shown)


music_assistant/providers/audible/audible_helper.py
```diff
@@ -41,14 +42,77 @@
...
+        "app_version": "3.56.2",
+        "source_token": refresh_token,
+        "requested_token_type": "access_token",
+        "source_token_type": "refresh_token",
+    }
+
+    target_domain = "audible" if with_username else "amazon"
+    url = f"https://api.{target_domain}.{domain}/auth/token"
+
+    async with httpx.AsyncClient() as client:
```
MarvinSchenkel: There should already be an http session available to you via `self.mass.http_session`. No need to create a new one.

MarvinSchenkel (approved):
Nice! 🚀

PR #2876, opened 2025-12-23
Plex Connect: Fix Plex Connect timeline reporting

anatosun:
Use queue.corrected_elapsed_time instead of player.corrected_elapsed_time to accurately track current track position. This fixes issues where:
- Timeline reported incorrect elapsed time when playing multiple tracks
- Huge time values were sent after track transitions
- Step forward/back commands used wrong elapsed time reference

The queue tracks per-track elapsed time while player tracks total session time.


MarvinSchenkel (approved):
Thanks @anatosun 🙏

PR #2877, opened 2025-12-23
Plex Connect: Ungroup player before starting playback

anatosun:
When a user selects a specific player for playback in Plex, automatically remove it from any sync groups or permanent groups. This should improve UX by ensuring playback only happens on the selected player, not all grouped players.

- Added _ungroup_player_if_needed() helper method
- Calls set_members directly to bypass static member restrictions
- Applied to handle_play_media (new playback) and handle_play (resume)
- Works with temporary syncs, permanent groups, and dynamic groups

### DIFF
--- a/music_assistant/providers/plex_connect/player_remote.py
+++ b/music_assistant/providers/plex_connect/player_remote.py
@@ -13,7 +13,12 @@ from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
 
 from aiohttp import ClientTimeout, web
-from music_assistant_models.enums import EventType, PlayerType, QueueOption, RepeatMode
+from music_assistant_models.enums import (
+    EventType,
+    PlayerType,
+    QueueOption,
+    RepeatMode,
+)
 from plexapi.playqueue import PlayQueue
 
 from .gdm import PlexGDMAdvertiser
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Nice work! Have a look at the Copilot comments, they make sense to implement to me. Then this will be good to merge 🚀

music_assistant/providers/plex_connect/player_remote.py
```diff
@@ -343,6 +351,35 @@ async def handle_poll(self, request: web.Request) -> web.Response:
...
+            return
+
+        LOGGER.debug("Ungrouping player %s before starting playback from Plex", player.display_name)
+        # Use set_members directly on the group to bypass static member check
+        if (
+            player.active_group
+            and (group := self.provider.mass.players.get(player.active_group))
+            and group.supports_feature(PlayerFeature.SET_MEMBERS)
+        ):
+            with suppress(UnsupportedFeaturedException, PlayerCommandFailed):
```
MarvinSchenkel: You can just leave the suppressions here I think, since we already check for the supported feature in the `if` above. Same comment for the `elifs` below.

MarvinSchenkel (approved):
Thanks @anatosun !

PR #2879, opened 2025-12-23
Enable multi-instance support for Audible provider

ztripez:
Allow users to add multiple Audible provider instances with separate logins. The provider implementation already supports this - just needed to add the multi_instance flag to the manifest. 

Fixes: music-assistant/discussions#862


OzGav (approved):

markjsmith:
Hi,
I cannot see the option to add a second instance of audible. Do I need to remove my existing one to add a second?

PR #2880, opened 2025-12-23
Add configuration for Zeroconf discovery interfaces

vossim:
- Add `zeroconf_interfaces` setting to allow choosing between 'default' and 'all' network interfaces for mDNS discovery.
- Move ConfigController setup before AsyncZeroconf initialization in `MusicAssistant` to ensure config is available.
- Initialize `AsyncZeroconf` based on the configured interface choice (`InterfaceChoice.All` vs `InterfaceChoice.Default`).

Not being able to bind the zeroconf interface to all network interfaces breaks the setup of some people. This leaves the default setting to the default network interface, but allows a user to configure zeroconf to bind to all network interfaces.

As mentioned in https://github.com/music-assistant/server/pull/2100#issuecomment-2811884657

### DIFF
--- a/music_assistant/controllers/streams/streams_controller.py
+++ b/music_assistant/controllers/streams/streams_controller.py
@@ -45,6 +45,7 @@ from music_assistant.constants import (
     CONF_ENTRY_ENABLE_ICY_METADATA,
     CONF_ENTRY_LOG_LEVEL,
     CONF_ENTRY_SUPPORT_GAPLESS_DIFFERENT_SAMPLE_RATES,
+    CONF_ENTRY_ZEROCONF_INTERFACES,
     CONF_HTTP_PROFILE,
     CONF_OUTPUT_CHANNELS,
     CONF_OUTPUT_CODEC,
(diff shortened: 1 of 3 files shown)


vossim:
Would be great to get this reviewed, helps at least myself and @jjhelin run music-assistant in our setups.

music_assistant/controllers/streams/streams_controller.py
```diff
@@ -294,6 +295,7 @@ async def get_config_entries(
                 default_value="GLOBAL",
                 category="advanced",
             ),
+            CONF_ENTRY_ZEROCONF_INTERFACES,
```
marcelveldt, line 298: This feels weird, hiding this setting into the streams controller.
But I get why it was done, because we dont have another place to put these global config options.

In the future we need to move all the mdns logic into a dedicated core controller (or builtin provider).

MarvinSchenkel: Noticed the same thing, but resolving this for now until we have a core controller for this.

MarvinSchenkel (approved):

PR #2885, opened 2025-12-24
Fix spotify podcast thumb image quality

OzGav:
Looks like Spotify changed the order images are delivered. Made the retrieval logic more robust to avoid this problem in the future.

### DIFF
--- a/music_assistant/providers/spotify/parsers.py
+++ b/music_assistant/providers/spotify/parsers.py
@@ -266,8 +266,6 @@ def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Pod
     if podcast_obj.get("description"):
         podcast.metadata.description = podcast_obj["description"]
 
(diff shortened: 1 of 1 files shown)


music_assistant/providers/spotify/parsers.py
```diff
@@ -45,9 +45,21 @@ def parse_images(
...
+
+    # First, try to find 640x640 image
+    for img in filtered_images:
+        if img.get("height") == 640:
+            best_image = img
+            break
+
+    # If no 640x640 found, select the largest image by height
+    if best_image is None:
+        best_image = max(
```
MarvinSchenkel: Can't all this logic be summarized by this one line if Spotify delivers 640x640 as the highest quality? In my mind it should just be: "sorting by size + grabbing the first one"

OzGav: Yes of course!

MarvinSchenkel (approved):
🎉

PR #2889, opened 2025-12-25
Add MCP Server plugin for AI assistant integration

ztripez:
## Summary

MCP (Model Context Protocol) server plugin that exposes Music Assistant to LLMs. Runs on port 8096 (configurable), uses MA tokens for auth.

**New dependency:** `mcp[cli]` (FastMCP SDK)

## What it does

- Playback: play, pause, stop, seek, skip, next/previous
- Volume: set, up/down, mute, group volume
- Queue: add, remove, move, shuffle, repeat, transfer between players
- Library: search, browse, get artists/albums/tracks, recommendations, similar tracks
- Playlists: create, delete, add/remove tracks, clear
- Players: power, group/ungroup, announcements
- Media: podcasts, audiobooks, radio stations
- Metadata: lyrics, artwork URLs

## Config UI

Settings grouped by category:
- **Server**: port, require auth
- **Query Permissions**: library, player, queue, playlist queries
- **Control Permissions**: playback, volume, player, queue control
- **Edit Permissions**: library, playlist, queue edits
- **Delete Permissions**: library, playlist, queue deletes (delete off by default)
- **MCP Resources**: player resources, library resources, prompts

## Test plan

- [x] 89 unit tests (tools, resources, prompts)
- [x] pre-commit + mypy pass

### Live validation

Tested with Claude Code, Gemini, and Codex against real MA instance:

| Category | Tests | Pass |
|----------|-------|------|
| Resources & Discovery | 5 | 5 |
| Search & Browse | 8 | 8 |
| Playback Control | 9 | 9 |
| Volume Control | 5 | 5 |
| Queue Management | 12 | 12 |
| Library Management | 11 | 11 |
| Playlist Management | 5 | 5 |
| Player Management | 9 | 9 |
| Player Capabilities | 4 | 4 |
| Feature Toggles | 11 | 10 |
| Advanced Filters | 8 | 8 |
| Discovery Features | 3 | 2 |
| New Resources | 4 | 4 |
| **Total** | **102** | **98** |

2 tests N/A due to client limitations (not server issues).

## Files

```
music_assistant/providers/mcp_server/
├── __init__.py      # Provider config, setup
├── auth.py          # Token verification
├── manifest.json    # Provider manifest
├── prompts.py       # MCP prompts
├── server.py        # FastMCP/uvicorn server
├── resources/       # MCP resources (players, library)
└── tools/           # MCP tools by category
    ├── library.py
    ├── media.py
    ├── metadata.py
    ├── playback.py
    ├── players.py
    ├── playlists.py
    ├── queue.py
    └── volume.py

tests/providers/mcp_server/
├── conftest.py      # Fixtures
├── test_prompts.py
├── test_resources.py
└── test_tools.py    # 89 tests
```

### DIFF
--- a/music_assistant/providers/mcp_server/__init__.py
+++ b/music_assistant/providers/mcp_server/__init__.py
@@ -1,0 +1,347 @@
+"""
+MCP Server Plugin for Music Assistant.
+
+Exposes Music Assistant functionality via the Model Context Protocol (MCP),
+enabling LLMs and AI assistants to control playback, query music library,
+and interact with speakers.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from music_assistant_models.config_entries import ConfigEntry, ConfigValueType
+from music_assistant_models.enums import ConfigEntryType
+
+from music_assistant.models.plugin import PluginProvider
+
+if TYPE_CHECKING:
+    from music_assistant_models.config_entries import ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+
+# Configuration keys
+CONF_PORT = "port"
+CONF_REQUIRE_AUTH = "require_auth"
+
+# Query permissions (read-only)
+CONF_LIBRARY_QUERY = "library_query"
+CONF_PLAYER_QUERY = "player_query"
+CONF_QUEUE_QUERY = "queue_query"
+CONF_PLAYLIST_QUERY = "playlist_query"
+
+# Control/Act permissions
+CONF_PLAYBACK_CONTROL = "playback_control"
+CONF_VOLUME_CONTROL = "volume_control"
+CONF_PLAYER_CONTROL = "player_control"
+CONF_QUEUE_CONTROL = "queue_control"
+
+# Edit permissions
+CONF_LIBRARY_EDIT = "library_edit"
+CONF_PLAYLIST_EDIT = "playlist_edit"
+CONF_QUEUE_EDIT = "queue_edit"
+
+# Delete permissions
+CONF_LIBRARY_DELETE = "library_delete"
+CONF_PLAYLIST_DELETE = "playlist_delete"
+CONF_QUEUE_DELETE = "queue_delete"
+
+# Resources and prompts
+CONF_PLAYER_RESOURCES = "player_resources"
+CONF_LIBRARY_RESOURCES = "library_resources"
+CONF_PROMPTS = "prompts"
+
+# Default port for MCP server
+DEFAULT_PORT = 8096
+
+SUPPORTED_FEATURES: set[object] = set()
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    return MCPServerProvider(mass, manifest, config)
+
+
+async def get_config_entries(
+    mass: MusicAssistant,
+    instance_id: str | None = None,  # noqa: ARG001
+    action: str | None = None,  # noqa: ARG001
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    # Get the configured port (use default if not set yet)
+    port = values.get(CONF_PORT, DEFAULT_PORT) if values else DEFAULT_PORT
+    if not isinstance(port, int):
+        port = DEFAULT_PORT
+
+    # Build the MCP server URL using the webserver's publish IP
+    try:
+        host = mass.streams.publish_ip
+    except Exception:
+        host = "localhost"
+    mcp_url = f"http://{host}:{port}/"
+
+    # Build connection info text (plain text, no markdown)
+    connection_info = (
+        f"MCP Server Endpoint: {mcp_url} - "
+        "When authentication is enabled, include a bearer token in your requests. "
+        f"Create tokens in Settings > Security at {mass.webserver.base_url}"
+    )
+
+    return (
+        # Connection info
+        ConfigEntry(
+            key="connection_info",
+            type=ConfigEntryType.LABEL,
+            label=connection_info,
+            required=False,
+        ),
+        # Server settings
+        ConfigEntry(
+            key=CONF_PORT,
+            type=ConfigEntryType.INTEGER,
+            label="Port",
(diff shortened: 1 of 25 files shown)


music_assistant/providers/sendspin/manifest.json
ALERTua, line 1: I guess this change was made by a mistake. This block does not seem to be related to this PR.

ztripez: Woops, my mistake, I'll make fix asap

ALERTua: It seems to me that `requirements_all.txt` and `pyproject.toml` also contain non-related changes.
`requirements_all.txt` should only contain two changes, described in the `mcp_server` `manifest.json`: `mcp==1.25.0` and `uvicorn==0.40.0`

ztripez: Pushed a fix, was some old labs i forgot to clean up.

requirements_all.txt
```diff
@@ -68,6 +69,7 @@ soundcloudpy==0.1.4
 sxm==0.2.8
 unidecode==1.4.0
 uv>=0.8.0
+uvicorn==0.40.0
```
marcelveldt, line 72: why the uvicorn dependency ? We cant do this with aiohttp ?

music_assistant/providers/mcp_server/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+CONF_PLAYLIST_DELETE = "playlist_delete"
+CONF_QUEUE_DELETE = "queue_delete"
+
+# Resources and prompts
+CONF_PLAYER_RESOURCES = "player_resources"
+CONF_LIBRARY_RESOURCES = "library_resources"
+CONF_PROMPTS = "prompts"
+
+# Default port for MCP server
+DEFAULT_PORT = 8096
```
marcelveldt: change this to something else because 8096 is I believe the default port for emby/jellyfin ?

music_assistant/providers/mcp_server/tools/media.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        try:
+            from music_assistant_models.enums import QueueOption  # noqa: PLC0415
+
+            # Convert chapter to string for play_media (it will be parsed internally)
+            start_item = str(chapter) if chapter is not None else None
+            await mass.player_queues.play_media(
+                queue_id=player_id,
+                media=audiobook_uri,
+                option=QueueOption.PLAY,
```
marcelveldt: If you leave option=None, it will pick the default configured option

requirements_all.txt
ztripez: Uvicorn is introduced here because the MCP server needs to run an ASGI application separately from the main MA aiohttp server. MCP itself (FastMCP SDK) is built around ASGI, and running it under an ASGI server like Uvicorn gives us a clean separation without blocking the core server.
If there’s a preferred pattern to host ASGI inside the existing aiohttp infrastructure  I’m open to exploring that, but as it stands this is the simplest and most reliable way to run MCP.

music_assistant/providers/mcp_server/__init__.py
ztripez: Yeah you are right, I'll change it.

ztripez:
Small note: most of the changes in this PR are DTOs, function-level wrapping and a homebrewed permission system. If MCP were ever moved into core, the majority of this logic could live as decorators and shared DTOs rather than provider-specific code.

music_assistant/providers/mcp_server/server.py
```diff
@@ -0,0 +1,328 @@
...
+        "transport_security": TransportSecuritySettings(
+            enable_dns_rebinding_protection=False,
+        ),
+    }
+
+    if require_auth:
+        from mcp.server.auth.settings import AuthSettings  # noqa: PLC0415
+        from pydantic import AnyHttpUrl  # noqa: PLC0415
+
+        from .auth import MusicAssistantTokenVerifier  # noqa: PLC0415
```
marcelveldt, line 179: what's up with all these in-line imports ? why are the imports not at module level ?

requirements_all.txt
marcelveldt: Yeah, unfortunately there is no (maintained) ASGI support for aiohttp so it is what it is then. 
At least the additional server is only loaded when the MCP server plugin is actually enabled.

marcelveldt:
> Small note: most of the changes in this PR are DTOs, function-level wrapping and a homebrewed permission system. If MCP were ever moved into core, the majority of this logic could live as decorators and shared DTOs rather than provider-specific code.

I like the fact that it is a provider which can be added/removed when folks see fit. Not everybody would be interested in a full fledged MCP server. However try to not duplicate auth/scope efforts so if we can make core more flexible to support more scenarios we should explore that (while keeping long term maintanability always in mind)

MarvinSchenkel:

music_assistant/providers/mcp_server/server.py
ztripez: Laziness and relying to much in the ruff pass I guess. I'll restructure.

ztripez:
> > Small note: most of the changes in this PR are DTOs, function-level wrapping and a homebrewed permission system. If MCP were ever moved into core, the majority of this logic could live as decorators and shared DTOs rather than provider-specific code.
> 
> I like the fact that it is a provider which can be added/removed when folks see fit. Not everybody would be interested in a full fledged MCP server. However try to not duplicate auth/scope efforts so if we can make core more flexible to support more scenarios we should explore that (while keeping long term maintanability always in mind)

Agreed and I wrote this before I did the jwt/claims PRs. in fact this is the reason I did them. I'll wait until they are merged and update this to match.

MarvinSchenkel:
> > > Small note: most of the changes in this PR are DTOs, function-level wrapping and a homebrewed permission system. If MCP were ever moved into core, the majority of this logic could live as decorators and shared DTOs rather than provider-specific code.
> > 
> > 
> > I like the fact that it is a provider which can be added/removed when folks see fit. Not everybody would be interested in a full fledged MCP server. However try to not duplicate auth/scope efforts so if we can make core more flexible to support more scenarios we should explore that (while keeping long term maintanability always in mind)
> 
> Agreed and I wrote this before I did the jwt/claims PRs. in fact this is the reason I did them. I'll wait until they are merged and update this to match.

FYI, The JWT PR has been merged. Still a few open items for you to look at for the claims PR, which should be good to go after that

MarvinSchenkel:
@ztripez any progress with this one?

MarvinSchenkel:
Politely closing this on in favour of https://github.com/music-assistant/server/pull/3858

PR #2891, opened 2025-12-26
Add JWT-based authentication with backward compatibility

ztripez:
## Summary

This PR introduces JWT-based authentication tokens for Music Assistant while maintaining full backward compatibility with existing hash-based tokens. The implementation provides a foundation for future OIDC integration and claims-based permissions.

### New Files
- `music_assistant/helpers/jwt_auth.py`: JWT encoding/decoding helper class
  - HS256 algorithm for token signing
  - Configurable secret key (auto-generated on first run)
  - Token encoding with standard JWT claims (sub, jti, iat, exp)
  - User metadata in claims (username, role, filters)
  - Token ID extraction without full validation

### Modified Files
- `music_assistant/controllers/webserver/auth.py`:
  - Updated `create_auth_token()` to generate JWT tokens
  - Updated `authenticate_with_token()` to accept both JWT and legacy tokens
  - Database expiration remains source of truth for both token types
  - Sliding window refresh for short-lived tokens (<30 days)

- `music_assistant/controllers/config.py`:
  - Added JWT secret key configuration entry
  - Auto-generates secure secret key on startup if not configured

## Token Format

JWT tokens include the following claims:
- `sub`: User ID
- `jti`: Token ID (for revocation via database)
- `iat`: Issued at timestamp
- `exp`: Expiration timestamp
- `username`: Username
- `role`: User role (admin/user)
- `player_filter`: Player filter configuration
- `provider_filter`: Provider filter configuration  
- `token_name`: Human-readable token name
- `is_long_lived`: Whether token is long-lived (>30 days)

## Backward Compatibility

- All existing hash-based tokens continue to work
- Token type auto-detected by format (JWT has 3 parts separated by dots)
- Database structure unchanged
- API endpoints unchanged
- Frontend requires no changes

## Security

- Tokens signed with HS256 algorithm using 256-bit secret key
- Secret key auto-generated using `secrets.token_urlsafe(32)`
- Database expiration checked as source of truth (prevents token reuse after revocation)
- Token ID (jti) allows individual token revocation
- No breaking changes to authentication flow

## Testing

All existing authentication tests pass (36/36):
- Hash token authentication
- JWT token authentication  
- Token creation and validation
- Token expiration handling
- Token revocation
- Sliding window refresh for short-lived tokens

## Future Work

This PR provides the foundation for:
- Claims-based permissions (follow-up PR)
- External OIDC provider support
- Provider-contributed custom claims
- OAuth2 refresh token flow

### DIFF
--- a/music_assistant/helpers/jwt_auth.py
+++ b/music_assistant/helpers/jwt_auth.py
@@ -14,13 +14,15 @@ Future OIDC Support:
 from __future__ import annotations
 
 import secrets
-from datetime import datetime, timedelta
-from typing import Any
+from datetime import datetime
+from typing import TYPE_CHECKING, Any
 
 import jwt
-from music_assistant_models.auth import User, UserRole
 
 from music_assistant.helpers.datetime import utc
+
+if TYPE_CHECKING:
+    from music_assistant_models.auth import User
 
 
 class JWTHelper:
@@ -53,12 +55,10 @@ class JWTHelper:
         """
         now = utc()
         payload = {
-            # Standard JWT claims
-            "sub": user.user_id,  # Subject (user ID)
-            "jti": token_id,  # JWT ID (token ID)
-            "iat": int(now.timestamp()),  # Issued at
-            "exp": int(expires_at.timestamp()),  # Expiration
-            # Custom claims
+            "sub": user.user_id,
+            "jti": token_id,
+            "iat": int(now.timestamp()),
+            "exp": int(expires_at.timestamp()),
             "username": user.username,
             "role": user.role.value,
             "player_filter": user.player_filter,
@@ -86,33 +86,6 @@ class JWTHelper:
         )
         return payload
 
-    def refresh_short_lived_token(
-        self,
-        token: str,
-        token_name: str,
-        expiration_days: int = 30,
-    ) -> str:
-        """Refresh a short-lived token with new expiration.
-
-        :param token: Original JWT token.
-        :param token_name: Token name.
-        :param expiration_days: Days until new expiration.
-        :return: New JWT token with updated expiration.
-        :raises jwt.InvalidTokenError: If token is invalid.
-        """
-        # Decode without verifying expiration (allow expired tokens to be refreshed)
-        payload = self.decode_token(token, verify_exp=False)
-
-        # Update timestamps
-        now = utc()
-        new_expires_at = now + timedelta(days=expiration_days)
-
-        payload["iat"] = int(now.timestamp())
-        payload["exp"] = int(new_expires_at.timestamp())
-        payload["token_name"] = token_name
-
-        return jwt.encode(payload, self.secret_key, algorithm=self.algorithm)
-
     @staticmethod
     def generate_secret_key() -> str:
         """Generate a secure random secret key for JWT signing.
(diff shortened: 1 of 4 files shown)


marcelveldt:
Please let me know what the fundamental issue is that this PR/feature tries to solve ?
I'd like to get a sales pitch why we really need to add this extra layer of complexity that needs to be maintained.

Don't get me wrong, I'm delighted that people are willing to invest time by doing PR's to extend functionality to core functionality but I kept the auth system simple on purpose to keep the scope limited. With adding more functionality, we increase maintenance burden and possible security issues.

ztripez:
> Please let me know what the fundamental issue is that this PR/feature tries to solve ? I'd like to get a sales pitch why we really need to add this extra layer of complexity that needs to be maintained.
> 
> Don't get me wrong, I'm delighted that people are willing to invest time by doing PR's to extend functionality to core functionality but I kept the auth system simple on purpose to keep the scope limited. With adding more functionality, we increase maintenance burden and possible security issues.

The main problem this addresses is that the current auth model is very coarse. As soon as a provider needs anything more granular than admin vs user, it has to invent its own permission checks. That is what happened with my MCP provider.

This PR does not change the end-user model and does not introduce per-user permissions or RBAC. Admin and user still exist as roles. What changes is that roles are expressed as predefined claim sets, and functionality can declare what it requires.

I want to be clear that I am not taking this the wrong way. I agree with you that the permission system needs to be kept on a very tight leash, because it can snowball quickly and become hard to maintain. That is also why I am explicitly not proposing user-level permission management inside MA.

The claims approach keeps MA simple while giving providers a consistent way to do fine-grained authorization, instead of each provider solving this differently. If someone wants real per-user permission granularity, my view is that this should be handled by an external identity provider via OIDC rather than modeled inside MA itself.

From an end-user perspective, nothing changes. From a provider and platform perspective, this avoids ad-hoc permission systems and makes authorization more explicit and predictable.

music_assistant/helpers/jwt_auth.py
```diff
@@ -0,0 +1,111 @@
...
+        now = utc()
+        payload = {
+            "sub": user.user_id,
+            "jti": token_id,
+            "iat": int(now.timestamp()),
+            "exp": int(expires_at.timestamp()),
+            "username": user.username,
+            "role": user.role.value,
+            "player_filter": user.player_filter,
+            "provider_filter": user.provider_filter,
```
marcelveldt: What is the advantage of having this stored in the token ?
I mean, it can dynamically be updated and then the token still has the old info.

marcelveldt:
We should consider switching to JWT by default so we can phase out the legacy tokens at some point.
So, in a follow-up PR, adjust the code to hand out JWT tokens by default and update the python client lib and frontend code.
Then we only accept legacy tokens (for a while, like 1 year or so) for token-auth only but login will always use JWT.
Does that make sense ?

music_assistant/helpers/jwt_auth.py
ztripez: Hm.. yes that is correct, and any future more fine grained access can be done with claims. I'll remove it.

MarvinSchenkel: @ztripez Could you have a look at this so we can get this merged please?

MarvinSchenkel (approved):
Thanks @ztripez 🎉

PR #2894, opened 2025-12-27
Subsonic Fixes

antoinepirlot:
What's changed (for Subsonic Provider):

* Refactor name of _get_artist_toptracks_ to _get_artist_top_tracks_ as now it respects the snake case
* When getting artist using _artistInfo2_, server returned an empty object or 404 error, so it is now ignored as when the server doesn't use a service like last.fm it returns empty object or a 404 error.
*  While getting _albumInfo2_ for an album view, it could returns error 404 if no last.fm like services is used on server. So it is ignored and data can be shown on screen. This error makes music assistant unable to play music from impacted albums even if music's data was returned by the server before that query.

(I hope I was clear on what I did).

Thanks for this project, I use it to stream on my IOT objects at home :D

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -15,6 +15,7 @@ from libopensonic.errors import (
     ParameterError,
     SonicError,
 )
+from mashumaro.exceptions import InvalidFieldValue
 from music_assistant_models.enums import ContentType, MediaType, StreamType
 from music_assistant_models.errors import (
     ActionUnavailable,
@@ -425,13 +426,20 @@ class OpenSonicProvider(MusicProvider):
                     )
                 },
             )
-
+        sonic_info = None
         try:
-            sonic_artist: SonicArtist = await self._run_async(
-                self.conn.get_artist, artist_id=prov_artist_id
-            )
-            sonic_info = await self._run_async(self.conn.get_artist_info2, aid=prov_artist_id)
-        except (ParameterError, DataNotFoundError) as e:
+            sonic_artist: SonicArtist = await self._run_async(self.conn.get_artist, artist_id=prov_artist_id)
+            try:
+                sonic_info = await self._run_async(self.conn.get_artist_info2, aid=prov_artist_id)
+            except HTTPError as e:
+                if e.response.status_code != 404:
+                    # If the user didn't activate last.fm connection it has 404 response code and it must be ignored
+                    raise e
+            except KeyError as e:
+                if e.args[0] != "":
+                    # If the user didn't activate last.fm connection it has an empty tuple and it must be ignored
+                    raise e
+        except (ParameterError, DataNotFoundError, Exception) as e:
             msg = f"Artist {prov_artist_id} not found"
             raise MediaNotFoundError(msg) from e
         return parse_artist(self.instance_id, sonic_artist, sonic_info)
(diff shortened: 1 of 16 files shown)


OzGav:
Sorry this has many problems. Did you run pre-commit before opening this? Also I don’t know if this will be accepted as it is touching core code when you have titled it subsonic fixes.

antoinepirlot:
Hello, no I did not.

I just took the dev branch and checking why it doesn't play some songs from my library as I didn't face any problem in an app I developed.

So I found the problem and just handle this specific error to fix the issue. I tested and the song was playing correctly (even other songs)

khers (changes requested):
This should be its own separate PR

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -365,9 +366,16 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
...
         """Return the requested Album."""
+        sonic_album: SonicAlbum
+        sonic_info = None
         try:
-            sonic_album: SonicAlbum = await self._run_async(self.conn.get_album, prov_album_id)
-            sonic_info = await self._run_async(self.conn.get_album_info2, aid=prov_album_id)
+            sonic_album = await self._run_async(self.conn.get_album, prov_album_id)
+            try:
+                sonic_info = await self._run_async(self.conn.get_album_info2, aid=prov_album_id)
+            except HTTPError as e:
```
khers, line 376: This is the wrong place to handle a 404. If the LastFM API key is not configured then your Open Subsonic server should be returning a response with the code for DataNotFoundError. The provider code would then need to place the call `to get_album_info2` in its own try block so that we can proceed with the basic info if that is all the server has.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -457,11 +465,14 @@ async def get_artist_albums(self, prov_artist_id: str) -> list[Album]:
         if prov_artist_id == UNKNOWN_ARTIST_ID or prov_artist_id.startswith(NAVI_VARIOUS_PREFIX):
             return []
 
-        try:
-            sonic_artist: SonicArtist = await self._run_async(self.conn.get_artist, prov_artist_id)
-        except (ParameterError, DataNotFoundError) as e:
-            msg = f"Album {prov_artist_id} not found"
-            raise MediaNotFoundError(msg) from e
+        sonic_artist: Artist = await self.get_artist(prov_artist_id=prov_artist_id)
+        # try:
```
khers: Why is this code commented out? If it is incorrect, remove it. But your commit message does not explain why this is removed.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -425,13 +426,20 @@ async def get_artist(self, prov_artist_id: str) -> Artist:
...
+                sonic_info = await self._run_async(self.conn.get_artist_info2, aid=prov_artist_id)
+            except HTTPError as e:
+                if e.response.status_code != 404:
+                    # If the user didn't activate last.fm connection it has 404 response code and it must be ignored
+                    raise e
+            except KeyError as e:
+                if e.args[0] != "":
+                    # If the user didn't activate last.fm connection it has an empty tuple and it must be ignored
+                    raise e
+        except (ParameterError, DataNotFoundError, Exception) as e:
```
khers, line 442: Catching the generic Exception is a a really bad idea IMO. As with the other commit, if your subsonic server is leaking the 404 back to the client, they are not following the API spec.

khers (changes requested):
This is not the way to handle the scenario described. The server should not be returning an HTTP 404 for a known endpoint and should instead return a Subsonic response with the error value of DataNotFound.

antoinepirlot:
Hello, I tested with another subsonic server and no error happened.

I apologize for this misunderstanding.

music_assistant/providers/opensubsonic/sonic_provider.py
antoinepirlot: Sorry, I forgot to remove the comment.

I did it as the get_artist(...) function has the same goal

OzGav:
@antoinepirlot So if no errors are being thrown with the other server does this mean that this PR is actually not required and can be closed?

antoinepirlot:
> @antoinepirlot So if no errors are being thrown with the other server does this mean that this PR is actually not required and can be closed?

Oh yes it can :)

PR #2901, opened 2025-12-30
Fix track name stripping too agressive

OzGav:
A user reported these songs were getting their title suffix stripped incorrectly:

CCF (I'm Gonna Stay with You)
The Catastrophe (Good Luck with That Man)

This PR should cover 99% of cases now (I think!)

### DIFF
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -99,7 +99,7 @@ IGNORE_TITLE_PARTS = (
 )
 WITH_TITLE_WORDS = (
     # words that, when following "with", indicate this is part of the song title
-    # not a featuring credit. Example: "(I'm Gonna Stay With You)" vs "(with John Doe)"
+    # not a featuring credit.
     "intro",
     "outro",
     "spoken",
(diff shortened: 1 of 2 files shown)


OzGav:
Did a longer search for titles with `(With` in them and there are very few so shortened the list of WITH_TITLE_WORDS to just catch the known tracks and not guess at others.

MarvinSchenkel:
The more I look at this PR, the more I think we should consider a different approach to solving this issue. Especially 'the' could still indicate a featuring artist right? (`Some title (with The Who / The Script`). 

Also, limiting the list now would not catch the edge case you mentioned in the PR I think? (The Catastrophe (Good Luck with That Man))

OzGav:
Let me look again but in my testing the listed tracks did now work correctly as with now has to be at the start of the phrase in brackets. Also I did extensive searching to find examples where there was a “with” at the start of the brackets but could not find many. I initially thought to provide a pre emptive list of words that MAY follow the with some time in the future but decided for simplicity not to?

marcelveldt:
Please extend the tests for this added functionality.

OzGav:
Tests added

MarvinSchenkel (approved):
Thanks @OzGav  🎉

PR #2905, opened 2026-01-01
Fix multiple spotify connect instances reporting to the latest registered webservice callback

kneirinck:
os.environ was used to communicate the callback url to the librespot binary.
When multiple instances are started simultaneously the env variable gets overwritten and only the last value was used.
Now every librespot instance gets its own env variables instead of using the shared os.environ.


MarvinSchenkel (approved):
Awesome, thanks @kneirinck 🎉

derekngibbs:
Same issue with trying to add Spotify and authenticate. I click the link "click here if the popup did not open". It opens authenticate.Spotify.com and I scroll down and select "agree". Then the browser pulls up a random page and Spotify is not setup. This is being done from IPhone. Running 2.8.0b4

PR #2907, opened 2026-01-02
fix: MusicCast Pause

fmunkes:
Could have sworn this worked once, but we use stop now when controlled by MA. Brought to attention by music-assistant/support#4564

If that PR is fine, maybe we could backport it to stable too?


MarvinSchenkel (approved):
Sweet, right on time for the next patch release :-)

PR #2908, opened 2026-01-02
Fix invalid timeout value for TLS connections

mtdcr:
[`aiohttp/connector.py`](https://github.com/aio-libs/aiohttp/blob/8b919d30a2339d3b9dbc53e8958f745b0f9703ea/aiohttp/connector.py#L1320-L1337) uses the total timeout value as handshake timeout for TLS connections:

```python
tls_transport = await self._loop.start_tls(
    [...]
    ssl_handshake_timeout=timeout.total,
```

This causes an exception with Deezer on slow network connections:
> ssl_handshake_timeout should be a positive number, got 0

According to https://github.com/aio-libs/aiohttp/issues/11859#issuecomment-3677980105, a zero value for total is not supposed to be used, but None means no timeout.

Fix it in all occurences throughout the repository.

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -736,7 +736,7 @@ async def resolve_radio_stream(mass: MusicAssistant, url: str) -> tuple[str, Str
         return (cache[0], StreamType(cache[1]))
     stream_type = StreamType.HTTP
     resolved_url = url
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
Thanks, @mtdcr

PR #2910, opened 2026-01-03
Subsonic: Update Subsonic library and use new AsyncConnection

khers:
py-opensonic 8.0.1 adds asyncio support with an AsyncConnection object that uses aiohttp. This simplifies the provider a great deal by allowing us to directly interact with the API instead of having to wrap each call. This commit should fix the stream and thread pool exhaustion errors we have been seeing since the 2.7 release.

This is going to need some extensive testing. We replaced all the synchronous IO and the wrappers to may it with an async compatible Subsonic library.

### DIFF
--- a/music_assistant/providers/opensubsonic/manifest.json
+++ b/music_assistant/providers/opensubsonic/manifest.json
@@ -6,7 +6,7 @@
   "description": "Stream music from your OpenSubsonic compatible server — your own cloud jukebox.",
   "codeowners": ["@khers"],
   "credits": ["[py-opensonic](https://github.com/khers/py-opensonic)"],
-  "requirements": ["py-opensonic==7.0.4"],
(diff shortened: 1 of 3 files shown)


marcelveldt (approved):
Looks good, @khers !
Great work getting this out of the way!

marcelveldt:
Please backport to stable branch once you have confirmation this fixes the issues that had been reported.

Clusters:
Just saw this now. @khers 
I think this also affects the scrobbler plugin.

Especially this lines of code:
```
# the opensubsonic library is not async friendly,
# so we need to run it in a executor thread
await asyncio.to_thread(handler, prov, item_id, report.uri)
```

Should just be a one-liner. Have to setup the dev again though. Will try to squeeze this in today.

khers:
I already have a pr up, sorry I forgot to tag you on it.

PR #2911, opened 2026-01-03
Party mode 

apophisnow:
This adds a party mode which does a few things.
* New Party Mode dashboard view shows played, currently playing, and upcoming tracks in a Jukebox style carousel.
* QR code on the Party Mode view allows guest access without having to log in. Guests are restricted to a page which only allows adding songs to the queue on the assigned party mode player.
* Configurable throttling for "Play Next" and "Add to queue" for guests.

There is also a models and frontend PR for this feature.


music_assistant/controllers/webserver/controller.py
```diff
@@ -379,6 +487,56 @@ async def close(self) -> None:
...
+    async def reload(self, config: CoreConfig | None = None) -> None:
+        """Reload the webserver controller.
+
+        :param config: Optional CoreConfig to use instead of fetching from config controller.
+        """
+        if config is None:
+            config = await self.mass.config.get_core_config(self.domain)
+
+        # Determine if we need a full restart
+        requires_restart = False
```
marcelveldt: I dont like this  as its unmaintainable long term.
I'd rather see a "require_reload" attribute on the configentry model itself and let this logic be handled by the core config controller.

music_assistant/controllers/webserver/controller.py
```diff
@@ -95,6 +103,10 @@ def __init__(self, mass: MusicAssistant) -> None:
         self.auth = AuthenticationManager(self)
         self.remote_access = RemoteAccessManager(self)
         self._sendspin_proxy = SendspinProxyHandler(self)
+        # Register party mode API commands
+        self.mass.register_api_command("webserver/party_mode_url", self.get_party_mode_url)
+        self.mass.register_api_command("webserver/party_mode_player", self.get_party_mode_player)
+        self.mass.register_api_command("webserver/party_mode_config", self.get_party_mode_config)
```
marcelveldt: This deserves its own (sub) controller or plugin provider instead of hacking in the webserver.

music_assistant/controllers/webserver/controller.py
```diff
@@ -1007,6 +1165,79 @@ async def _handle_setup(self, request: web.Request) -> web.Response:
                 {"success": False, "error": f"Setup failed: {e!s}"}, status=500
             )
 
+    async def _get_guest_token(self) -> str:
+        """
+        Get or create a guest access token for party mode.
+
+        :returns: The guest access token.
+        """
```
marcelveldt: I wonder if we can leverage the work of @ztripez here that is working on claims based auth - this sounds like a usecase for that approach ?

music_assistant/controllers/webserver/websocket_client.py
```diff
@@ -38,6 +38,51 @@
 MAX_PENDING_MSG = 512
 CANCELLATION_ERRORS: Final = (asyncio.CancelledError, futures.CancelledError)
 
+# Commands that guest users CANNOT execute (party mode blocklist)
+# All other commands are allowed for guests to enable full read-only access
+GUEST_BLOCKED_COMMANDS: Final = frozenset(
```
marcelveldt: isnt this unmaintainable long term ?
Cant we just iontroduce a new role "guest" here

apophisnow: Agreed, and I don't think it is needed at all since I locked the guest user to a special guest view.

music_assistant/controllers/webserver/controller.py
apophisnow: Agreed, refactored to us require_reload on config entry instead.

music_assistant/controllers/webserver/controller.py
apophisnow: Agreed, refactored to its own "Party Mode" provider. Should be a lot cleaner.

music_assistant/controllers/webserver/controller.py
apophisnow: I'm certainly interested. I would love to discuss with you both about the overall architectural model we want to follow for roles and permissions. Perhaps, better to discuss in depth on discord though?

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+                display_name="Party Guest",
+            )
+
+        # Generate a new long-lived token for guest access
+        # Note: Tokens are hashed and cannot be retrieved, so we generate a fresh one
+        # Old tokens remain valid until they expire or are manually revoked
+        return await self.mass.webserver.auth.create_token(
+            user=guest_user,
+            name="Party Mode Guest Token",
+            is_long_lived=True,
```
MarvinSchenkel: We might want to think about limiting the validity of guest tokens to something like 6-12 hours? Especially with remote access this would otherwise open up the scenario where guests can keep adding items to the queue after the party is over.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,271 @@
...
+        """Get the guest access URL for party mode.
+
+        :returns: The guest access URL with embedded token, or empty string if disabled.
+        """
+        if not self.config.get_value(CONF_ENABLE_GUEST_ACCESS):
+            return ""
+
+        token = await self._get_guest_token()
+        base_url_value = self.mass.webserver.config.get_value("base_url")
+        base_url = str(base_url_value) if base_url_value else f"http://localhost:{DEFAULT_PORT}"
```
MarvinSchenkel, line 498: I think we should use app.music-assistant.io for this combined with a short lived token that will immediately take the guest to the right page.

Using local urls would require guests to be on the wifi network.

music_assistant/controllers/webserver/controller.py
MarvinSchenkel: > I wonder if we can leverage the work of @ztripez here that is working on claims based auth - this sounds like a usecase for that approach ?

I think a simple guest role won't be sufficient if we want to make the 'guest speaker' configurable dynamically?

apophisnow: By guest speaker you mean the player that the guests are adding songs to right? In my current implementation, that is configured manually in the provider settings. I haven't thought about making it dynamic yet, maybe we could though. Not sure what the behavior for that should be though.

music_assistant/providers/party_mode/__init__.py
apophisnow: I changed some of this yesterday to make it so if you uncheck the QR code access or disable the provider the tokens are revoked. Shortening the life of the tokens, or making that configurable might be a good idea though.

music_assistant/providers/party_mode/__init__.py
apophisnow: I agree, or at least have it as a configurable option. I'll look into this.

apophisnow: I think if remote access is enabled, then it should automatically use that, otherwise fallback to a local URL.

MarvinSchenkel: Yep, makes sense to me.

music_assistant/controllers/webserver/controller.py
MarvinSchenkel: Well I meant 'configureable'. The fact that we can make it configureable means we somehow have to 'dynamically' assign access to that speaker. Meaning that the claims based auth could be a nice solution for this.

For context, we were asking about a use-case in the claims based auth PR but looks like we've got one now ;-)

apophisnow: Ok, so I have cherry-picked some of the JWT stuff from @ztripez. I have implemented short codes (6 char long) which get removed from the url and exchanged for JWT tokens. I did this to reduce the size of the data for the QR code for better scanning, and also to simplify the url in case users want to share it without the QR code.

music_assistant/providers/party_mode/__init__.py
apophisnow: I've reduced the token validity to 8 hours.

music_assistant/providers/party_mode/__init__.py
apophisnow: Remote access compatability has been implemented.

music_assistant/providers/party_mode/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "plugin",
+  "stage": "alpha",
+  "domain": "party_mode",
+  "name": "Party Mode",
+  "description": "Enable guest access with a shareable URL for party mode, allowing guests to add songs to the queue with rate limiting.",
+  "codeowners": ["@apophisnow"],
+  "requirements": [],
+  "documentation": "https://music-assistant.io"
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/plugins/party-mode"
```

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -192,6 +194,28 @@ def set_current_token(token: str | None) -> None:
     current_token.set(token)
 
 
+@contextmanager
```
marcelveldt: this belongs strictly taken in a dedicated PR

music_assistant/controllers/webserver/remote_access/__init__.py
```diff
@@ -85,6 +85,7 @@ async def close(self) -> None:
         await self.stop()
         for unload_cb in self._on_unload_callbacks:
             unload_cb()
+        self._on_unload_callbacks.clear()
```
marcelveldt: this belongs strictly taken in a dedicated PR

music_assistant/controllers/webserver/auth.py
marcelveldt, line 1: all of this needs discussion and belongs in a dedicated PR

music_assistant/controllers/webserver/auth.py
```diff
@@ -53,11 +53,16 @@
...
 
 # Database schema version
-DB_SCHEMA_VERSION = 4
+DB_SCHEMA_VERSION = 5
 
-# Token expiration constants (in days)
-TOKEN_SHORT_LIVED_EXPIRATION = 30  # Short-lived tokens (auto-renewing on use)
-TOKEN_LONG_LIVED_EXPIRATION = 3650  # Long-lived tokens (10 years, no auto-renewal)
+# Token expiration constants
+TOKEN_SHORT_LIVED_EXPIRATION_HOURS = 8  # Short-lived tokens (auto-renewing on use)
```
marcelveldt: This is a bad change slipping in a completely unrelated PR

music_assistant/controllers/webserver/auth.py
```diff
@@ -221,6 +243,12 @@ async def _create_database_indexes(self) -> None:
         await self.database.execute(
             "CREATE INDEX IF NOT EXISTS idx_tokens_hash ON auth_tokens(token_hash)"
         )
+        await self.database.execute(
+            "CREATE INDEX IF NOT EXISTS idx_join_codes_hash ON join_codes(code_hash)"
+        )
+        await self.database.execute(
+            "CREATE INDEX IF NOT EXISTS idx_join_codes_user ON join_codes(user_id)"
+        )
```
marcelveldt: I don't want any of this in the core code - the guest mode provider should be an isolated component.
We should only make changes to the core code if they make sense for the complete project to the benefit of all providers

music_assistant/controllers/player_queues.py
```diff
@@ -319,15 +320,36 @@ async def set_shuffle(self, queue_id: str, shuffle_enabled: bool) -> None:
...
+            x
+            for x in next_items
+            if x.extra_attributes.get("added_by_user_role") == UserRole.GUEST.value
+        ]
+        regular_items = [
+            x
+            for x in next_items
+            if x.extra_attributes.get("added_by_user_role") != UserRole.GUEST.value
+        ]
+
```
marcelveldt: all of this needs to be dropped. the guest mode should be an isolated provider that manipulates the queue on behalf of a user

marcelveldt (dismissed):
Please drop all unrelated code changes to core controlelrs.
A provider needs to be an ISOLATED component which interacts with the existing core controllers.

If a core controller needs an adjustment to extend its featureset towards providers, that needs dedicated, well explained PR's including a discussion why this (architectural) change is needed.

music_assistant/controllers/webserver/auth.py
apophisnow: Yep. reverted.

music_assistant/controllers/webserver/auth.py
apophisnow: I have I have isolated the short code functionality into a generic feature any provider can use. https://github.com/music-assistant/server/pull/3078

It is also showing in this PR since it isn't merged yet, but it is needed for the Party Mode feature in its current form.

music_assistant/controllers/webserver/auth.py
apophisnow: Created https://github.com/music-assistant/server/pull/3078 to start addressing this.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+            depends_on=CONF_ENABLE_BOOST,
+            range=(5, 120),
+            advanced=True,
+            category="Guest Features",
+        ),
+        # Skip Song feature
+        ConfigEntry(
+            key=CONF_ENABLE_SKIP_SONG,
+            type=ConfigEntryType.BOOLEAN,
+            default_value=True,
```
MarvinSchenkel: I would say this should be `False` by default?

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+    """
+    return (
+        ConfigEntry(
+            key=CONF_PARTY_MODE_PLAYER,
+            type=ConfigEntryType.STRING,
+            default_value="",
+            required=True,
+            label="Party Mode Player",
+            description="Select which player/queue guests will add songs to.",
+            depends_on=CONF_ENABLE_GUEST_ACCESS,
```
MarvinSchenkel, line 158: Since this depends on guest access, I would move the config for guest access up.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        """Revoke all guest access tokens and codes for party mode.
+
+        This is called when guest access is disabled or the plugin is removed.
+        We disconnect WebSocket connections to force the frontend to redirect to login,
+        revoke tokens so they can't reconnect, and invalidate pending join codes.
+        """
+        auth = self.mass.webserver.auth
+
+        # Revoke pending join codes for party mode only
+        codes_revoked = await auth.revoke_join_codes(provider_name="party_mode")
```
MarvinSchenkel:
```suggestion
        codes_revoked = await auth.revoke_join_codes(provider_name=self.provider.domain)
```

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        # Ensure we have a guest user for party mode
+        guest_user_id = await self._get_or_create_party_guest_user()
+
+        # Generate a new short code (valid for 24 hours, unlimited uses)
+        code, expires_at = await self.mass.webserver.auth.generate_join_code(
+            user_id=guest_user_id,
+            expires_in_hours=8,
+            max_uses=0,  # Unlimited uses
+            device_name="Party Mode Guest",
+            provider_name="party_mode",
```
MarvinSchenkel:
```suggestion
            provider_name=self.provider.domain,
```

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        if boost and not self.config.get_value(CONF_ENABLE_BOOST):
+            raise InvalidDataError("Boost feature is disabled")
+        if not self.config.get_value(CONF_ENABLE_ADD_QUEUE):
+            raise InvalidDataError("Add to queue feature is disabled")
+
+        # Get the party mode queue
+        queue_id = await self.get_party_mode_player()
+        if not queue_id:
+            # Fall back to first available player
+            players = list(self.mass.players.all(False, False))
```
MarvinSchenkel:
```suggestion
            players = list(self.mass.players.all_players(False, False))
```

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        :returns: Dictionary with url, code, and expires_at (or empty values if disabled).
+        """
+        if not self.config.get_value(CONF_ENABLE_GUEST_ACCESS):
+            return {"url": "", "code": "", "expires_at": None}
+
+        # Ensure we have a guest user for party mode
+        guest_user_id = await self._get_or_create_party_guest_user()
+
+        # Generate a new short code (valid for 24 hours, unlimited uses)
+        code, expires_at = await self.mass.webserver.auth.generate_join_code(
```
MarvinSchenkel: This will be called with every page refresh I think? Resulting in a new join code with every refresh, without deleting the other one. I think we should add some 'get_or_create' logic here, especially if we plan to add non-expiring join codes.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        queue logic including priority positioning for guest items.
+
+        :param uri: The URI of the media item to add (e.g., "spotify://track/xxx").
+        :param boost: If True, insert at the front of the guest section (play next).
+        :returns: Result dict with success status and queue position info.
+        """
+        user = get_current_user()
+
+        # Verify user is a guest
+        if not user or user.role != UserRole.GUEST:
```
MarvinSchenkel: What about ADMINS? They should still be able to access the UI I'd say?

music_assistant/providers/party_mode/__init__.py
MarvinSchenkel: Actually, I think we should just raise an error. It could otherwise lead to music being played to a totally unexpected speaker.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+
+        :returns: The queue ID for party mode, or None to use active player.
+        """
+        if not self.config.get_value(CONF_ENABLE_GUEST_ACCESS):
+            return None
+
+        player_id = self.config.get_value(CONF_PARTY_MODE_PLAYER)
+        return str(player_id) if player_id else None
+
+    async def get_party_mode_config(self) -> dict[str, int | bool | str]:
```
MarvinSchenkel: I would just add a class for typed support. Should be automatically serialized when the dataclass has to_dict:
```
@dataclass
class PartyModeConfig(DataClassDictMixin):
.....
```

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+                queue_id=queue_id,
+                media=uri,
+                option=QueueOption.PLAY,
+            )
+            started_playback = True
+            # Mark the newly added item
+            queue_items = self.mass.player_queues.items(queue_id)
+            if queue_items:
+                # The item we just added should be at current_index
+                current_idx = queue.current_index or 0
```
MarvinSchenkel: I think this icw the play_media above could lead to a race condition. Just use the same logic as in the _add_to_boost_section using get_item_by_uri and self.mass.player_queues.load. Then call play_index(queue_id, 0) to start playback.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        # Check if guest access and skip are enabled
+        if not self.config.get_value(CONF_ENABLE_GUEST_ACCESS):
+            raise InvalidDataError("Party mode guest access is disabled")
+        if not self.config.get_value(CONF_ENABLE_SKIP_SONG):
+            raise InvalidDataError("Skip song feature is disabled")
+
+        # Get the party mode queue
+        queue_id = await self.get_party_mode_player()
+        if not queue_id:
+            players = list(self.mass.players.all(False, False))
```
MarvinSchenkel:
```suggestion
            players = list(self.mass.players.all_players(False, False))
```

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+        # This forces the frontend to redirect to login
+        self.logger.debug(
+            "Disconnecting WebSocket connections for user '%s' (user_id=%s)",
+            guest_user.username,
+            guest_user.user_id,
+        )
+        self.mass.webserver.disconnect_websockets_for_user(guest_user.user_id)
+
+        # Revoke all tokens for the guest user so they can't reconnect
+        token_rows = await auth.database.get_rows(
```
MarvinSchenkel: We cannot interact with the db directly from a plugin. Let's use the revoke functionality in the web auth after the join code PR is merged

music_assistant/providers/party_mode/__init__.py
apophisnow: I suppose we could add that, but this is specifically for the guest UI page that guests get when the scan the QR code. They are locked to that page and are only allowed to add songs in this restricted way. I don't see how you would ever access this function as an admin?

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,846 @@
...
+
+        # If queue is empty or we're at the end, append at the end
+        if not queue_items or search_start >= len(queue_items):
+            return len(queue_items)
+
+        # Find the last consecutive guest item after current_index
+        guest_section_end = search_start
+        for i in range(search_start, len(queue_items)):
+            if queue_items[i].extra_attributes.get(ATTR_PARTY_MODE_GUEST):
+                guest_section_end = i + 1  # Insert after this guest item
```
MarvinSchenkel: Just checking. Could an admin theoretically insert in the queue in between guest items:
guest
guest
admin
guest
guest

If that's the case, then we would detect position 2 as guest end section instead of 5?

music_assistant/providers/party_mode/__init__.py
apophisnow: For context this is what I did instead of building out an entire permission system. If we had a way to very granularly control what behavior was accessable by guests we might not need a function like this at all, but it would take a lot more work to design that robust permission system. Maybe something we should work towards in the future?

music_assistant/providers/party_mode/__init__.py
apophisnow: Agreed

music_assistant/providers/party_mode/__init__.py
MarvinSchenkel: Ah I thought this was also used in the party mode page itself that is accessible from the menu within MA? If that's not the case, you can discard this comment.

music_assistant/controllers/webserver/auth.py
```diff
@@ -949,6 +950,24 @@ async def revoke_token(self, token_id: str) -> None:
         # Disconnect any WebSocket connections using this token
         self.webserver.disconnect_websockets_for_token(token_id)
 
+    async def revoke_tokens_for_user(self, user_id: str) -> int:
```
MarvinSchenkel: This can now also be `User`

music_assistant/controllers/webserver/auth.py
```diff
@@ -1541,32 +1560,29 @@ async def unlink_provider(self, user_id: str, provider_type: str) -> bool:
 
     async def generate_join_code(
         self,
-        user_id: str,
+        user: User,
+        provider_name: str,
```
MarvinSchenkel: I think we can drop this now, no?

music_assistant/controllers/webserver/auth.py
```diff
@@ -949,6 +950,24 @@ async def revoke_token(self, token_id: str) -> None:
...
+        """Revoke all auth tokens for a user.
+
+        This is an internal method for programmatic use (e.g., when disabling guest access).
+        Unlike revoke_token(), this does not require an authenticated user context.
+
+        :param user_id: The user ID whose tokens should be revoked.
+        :return: Number of tokens revoked.
+        """
+        token_rows = await self.database.get_rows("auth_tokens", {"user_id": user_id}, limit=1000)
+        revoked_count = 0
```
MarvinSchenkel: This seems unnecessarily expensive. We can just delete in 1 go: `await self.database.execute("DELETE FROM auth_tokens WHERE user_id = :user_id", {"user_id": user_id})`. You could leave the loop for the websocket disconnects though.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,901 @@
...
+
+    async def skip_current(self) -> dict[str, Any]:
+        """Skip the currently playing track.
+
+        :returns: Result dict with success status.
+        """
+        user = get_current_user()
+
+        # Verify user is a guest
+        if not user or user.role != UserRole.GUEST:
```
MarvinSchenkel: Should we also validate that the user is actually the party mode user, since we might use the guest role for different plugins in the future?

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,901 @@
...
+        await self.mass.player_queues.load(
+            queue_id=queue_id,
+            queue_items=[queue_item],
+            insert_at_index=insert_index,
+            keep_remaining=True,
+            keep_played=True,
+            shuffle=False,
+        )
+
+    def _find_boost_section_end(self, queue_items: list[QueueItem], current_index: int) -> int:
```
MarvinSchenkel: I think the logic for find_xxxx_section_end and add_to_xxxx_section contain a bit of code duplication. I think we can centralize most of the logic in a helper, call it from the boost/guest functions and just pass the extra_attributes that are required for the logic?

music_assistant/controllers/webserver/auth.py
apophisnow: yeah good point, that is dumb lol

music_assistant/controllers/webserver/auth.py
apophisnow: Yep. Done.

music_assistant/providers/party_mode/__init__.py
apophisnow: Fixed.

music_assistant/providers/party_mode/__init__.py
apophisnow: Maybe possible? I'm not sure how you would insert tracks in a specific position in the MA UI for a queue other than beginning, end, or next. You might know better than me. Either way, a fairly rare edge case that would be caused by the user doing something weird in the first place...

music_assistant/controllers/webserver/auth.py
```diff
@@ -949,6 +943,40 @@ async def revoke_token(self, token_id: str) -> None:
...
+        """Revoke all auth tokens for a user.
+
+        This is an internal method for programmatic use (e.g., when disabling guest access).
+        Unlike revoke_token(), this does not require an authenticated user context.
+
+        :param user: The user whose tokens should be revoked.
+        :return: Number of tokens revoked.
+        """
+        token_rows = await self.database.get_rows(
+            "auth_tokens", {"user_id": user.user_id}, limit=1000
```
MarvinSchenkel: Why is this limited to 1000?

music_assistant/controllers/webserver/auth.py
```diff
@@ -1645,52 +1706,51 @@ async def _exchange_join_code(self, code: str) -> str | None:
...
+
+        :param user: The user to look up codes for.
+        :return: The join code string if found, None otherwise.
+        """
+        now = utc()
+        cursor = await self.database.execute(
+            """
+            SELECT code FROM join_codes
+            WHERE user_id = :user_id
+            AND expires_at > :now
```
MarvinSchenkel, line 1743: We should also combine this check with `max_uses`. So OR it has expired OR max uses have been exhausted.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,828 @@
...
+        queue logic including priority positioning for guest items.
+
+        :param uri: The URI of the media item to add (e.g., "spotify://track/xxx").
+        :param boost: If True, insert at the front of the guest section (play next).
+        :returns: Result dict with success status and queue position info.
+        """
+        user = get_current_user()
+
+        # Verify user is a guest
+        if not user or user.username != PARTY_GUEST_USERNAME:
```
MarvinSchenkel: Let's check for BOTH the guest role and the username here.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,828 @@
...
+
+    async def skip_current(self) -> dict[str, Any]:
+        """Skip the currently playing track.
+
+        :returns: Result dict with success status.
+        """
+        user = get_current_user()
+
+        # Verify user is a guest
+        if not user or user.username != PARTY_GUEST_USERNAME:
```
MarvinSchenkel: Let's check for BOTH the guest role and the username here.

music_assistant/providers/party_mode/__init__.py
```diff
@@ -0,0 +1,828 @@
...
+        # Handle different scenarios based on queue state and boost mode
+        started_playback = False
+
+        if queue.state == PlaybackState.PLAYING:
+            pass  # Handled below via boost/guest section insertion
+        elif queue.current_index is not None:
+            # Not playing but queue has a position (paused/idle with loaded queue)
+            # Insert as next song after current position, then skip to it
+            insert_index = queue.current_index + 1
+            media_item = await self.mass.music.get_item_by_uri(uri)
```
MarvinSchenkel: Is it me or can we simply cut this function in half by moving some code out of the elif branch? Seems there's a fair bit of duplication

music_assistant/controllers/webserver/auth.py
apophisnow: I didn't intend to specifically limit this, will remove it.

music_assistant/providers/party_mode/__init__.py
apophisnow: The only user I would want hitting this is specifically the user that party mode created. I'm happy to add the check, but I don't understand the reasoning.

music_assistant/providers/party_mode/__init__.py
apophisnow: Only the specific party mode user should hit this. What reason should we check the role for?

music_assistant/controllers/webserver/auth.py
```diff
@@ -949,6 +948,40 @@ async def revoke_token(self, token_id: str) -> None:
...
+        # Disconnect any WebSocket connections using these tokens
+        for token_row in token_rows:
+            self.webserver.disconnect_websockets_for_token(token_row["token_id"])
+
+        # Delete all tokens in one go
+        await self.database.execute(
+            "DELETE FROM auth_tokens WHERE user_id = :user_id",
+            {"user_id": user.user_id},
+        )
+
```
MarvinSchenkel, line 974: Think we miss a `commit()` here?

MarvinSchenkel (approved):
It's PARTY TIME!!!! 🎉

PR #2913, opened 2026-01-03
ABS/ iTunes podcasts: allow translation of folders

fmunkes:
Adapts the abs and iTunes podcasts provider to make the recommendation and browse folders translatable.

Depends on music-assistant/frontend#1345


marcelveldt (approved):
Thanks, @fmunkes !

PR #2915, opened 2026-01-03
Enhanced Skip previous behavior

Bonusbartus:
change skip_previous behavior, with the new behavior, giving the previous track command, restarts the current song, when giving the command (again) within 4 seconds of starting it will skip back to the previous track


marcelveldt (approved):
Thanks, @Bonusbartus !

PR #2916, opened 2026-01-03
Fix sql injection vulnerability

marcelveldt:
Executing (custom) sql queries on the library should not be exposed on the api to prevent sql injection issues.


mweinelt:
Are there plans to issue a security advisory for this issue?

PR #2920, opened 2026-01-04
Allow music providers to provide the "date_added" field to library items

marcelveldt:


music_assistant/controllers/media/base.py
```diff
@@ -942,6 +943,7 @@ def _parse_db_row(db_row: Mapping[str, Any]) -> dict[str, Any]:
         db_row_dict["provider"] = "library"
         db_row_dict["favorite"] = bool(db_row_dict["favorite"])
         db_row_dict["item_id"] = str(db_row_dict["item_id"])
+        db_row_dict["date_added"] = datetime.fromtimestamp(db_row_dict["timestamp_added"])
```
fmunkes: If we merge https://github.com/music-assistant/models/pull/150 , then we need:
```python
db_row_dict["date_added"] = int(db_row_dict["timestamp_added"])
```

fmunkes:
```python
db_row_dict["date_added"] = datetime.fromtimestamp(
    db_row_dict["timestamp_added"]
).isoformat()
```
And then mashumaro creates a datetime from it.

fmunkes (approved):

PR #2928, opened 2026-01-05
Update old template references

dmcc:
This also includes some reformatting and a note about how to run a separate development server.

Let me know if you'd like me to remove any parts of this.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,18 +23,6 @@ With this repository cloned locally, execute the following commands in a termina
 * Make sure, that the python interpreter in VS Code is set to the newly generated venv.
 * Debug: Hit (Fn +) F5 to start Music Assistant locally (VS Code), or run `python -m music_assistant --log-level debug` from the command line
 * The pre-compiled UI of Music Assistant will be available at `localhost:8095` 🎉
-* To change the web server's port (e.g., if you're running a production server
-  on the same machine), set `bind_port` in your `settings.json`
-  (`~/.musicassistant/settings.json` by default). Under `core -> webserver`,
-  you'll want something like this:
-
-      "webserver": {
-        "values": {
-          "bind_port": <WEB_SERVER_PORT>
-        },
-        "domain": "webserver",
-        "last_error": null
-      }
 
 NOTE: Always re-run the setup script after you fetch the latest code because requirements could have changed.
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,8 +23,18 @@ With this repository cloned locally, execute the following commands in a termina
 * Make sure, that the python interpreter in VS Code is set to the newly generated venv.
 * Debug: Hit (Fn +) F5 to start Music Assistant locally (VS Code), or run `python -m music_assistant --log-level debug` from the command line
 * The pre-compiled UI of Music Assistant will be available at `localhost:8095` 🎉
-* To change the port (e.g., if you're running a production server on the same machine), edit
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -14,11 +14,17 @@ It is recommended to use Visual Studio Code as your IDE, since launch files to s
...
+      scripts/setup.sh
+
+    (creates a new separate virtual environment to nicely separate the project dependencies)
+
 * The setup script will create a separate virtual environment (if needed), install all the project/test dependencies and configure pre-commit for linting and testing.
 * Make sure, that the python interpreter in VS Code is set to the newly generated venv.
 * Debug: Hit (Fn +) F5 to start Music Assistant locally (VS Code), or run `python -m music_assistant --log-level debug` from the command line
 * The pre-compiled UI of Music Assistant will be available at `localhost:8095` 🎉
+* To change the port (e.g., if you're running a production server on the same machine), edit
+  `DEFAULT_SERVER_PORT` in `music_assistant/controllers/webserver/controller.py`
```
marcelveldt: This is just plain wrong - you need to change the port in the config

marcelveldt: If it helps, we can maybe adjust the port through a commandline argument

dmcc: Thanks for taking a look! I've updated the instructions for how to change it via `settings.json` -- let me know if this isn't what you had in mind.

I'd be in favor of adjusting the port through a CLI argument. That seems pretty standard for most tools and would let us simplify these docs (presumably we'd want to add the flags in the task in `.vscode/launch.json`?).

DEVELOPMENT.md
```diff
@@ -14,11 +14,27 @@ It is recommended to use Visual Studio Code as your IDE, since launch files to s
...
+  (`~/.musicassistant/settings.json` by default). Under `core -> webserver`,
+  you'll want something like this:
+
+      "webserver": {
+        "values": {
+          "bind_port": <WEB_SERVER_PORT>
+        },
+        "domain": "webserver",
+        "last_error": null
+      }
```
marcelveldt: why educating people on manually hacking in config files if there is just a UI setting for this ?
the settings.json should not be manually touched by either users or developers, ever.

DEVELOPMENT.md
marcelveldt: If the port is indeed an issue, I'm in favor of simply adding a cli argument.
I will add that later today or tomorrow. That has my preference over manually hacking in config files

DEVELOPMENT.md
dmcc: Ah, sorry, I should have explained my setup better -- maybe you can suggest the best way to do this: I have one machine where I run a production version in Docker but also want to do some local development. When I startup the development server via a venv in `vscode`, it crashes because the port is already in use, so I never get to the UI setting to change the port. I couldn't find docs on how to do this (but obviously don't know the Music Assistant ecosystem well :smiley:) hence this PR.

I suppose I could have Docker remap the port, but that's a bit cumbersome as lots other things depend on the production port number.

MarvinSchenkel: I agree this should be left out. Another way to workaround this is to (temporarily) change your port on your production instance. I think it's quite an edge case to be running production _and_ developing MA on the same host ;-)

MarvinSchenkel: All your other changes are valid, so if you take this out we can merge this PR 🚀

dmcc: Thanks @MarvinSchenkel! I've removed the custom port details.

MarvinSchenkel (approved):
Thanks! @dmcc

PR #2929, opened 2026-01-05
Add method to move item to end of queue

Bonusbartus:
Add method and API command to player_queues to move an item in queue to the end of the queue

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -629,8 +629,24 @@ class PlayerQueuesController(CoreController):
         - queue_id: id of the queue to process this request.
         - queue_item_id: the item_id of the queueitem that needs to be moved.
         """
-        pos_shift = len(self._queue_items[queue_id])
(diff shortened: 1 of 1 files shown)


OzGav:
Might be handy if you could add move to top as well to fully satisfy https://github.com/orgs/music-assistant/discussions/4329

Bonusbartus:
I planned to, but only then I noticed that would be doing the same as the “Play Next” option.
Or am I wrong? Would it make sense moving the song up till just after play next?

OzGav:
No you are right, I had forgotten about that!

Bonusbartus:
@OzGav, might be a good idea if you read the https://github.com/orgs/music-assistant/discussions/4329 discussion again.
please let me know how to proceed, I don't mind implementing the move to top functionality as described there, but not sure if you would want to have it in, in this pull request?

OzGav:
No I think the current “play next” is moving to the top so I don’t believe anything else is required.

marcelveldt (approved):

PR #2934, opened 2026-01-05
Add N818 mypy rule

OzGav:


MarvinSchenkel (approved):
Let's wait with merging this one till the beta period for 2.8 starts to avoid merge conflicts.

marcelveldt:
Not a fan of this change. Let's leave N818 out as it doesn't add any value except of annoying developers

OzGav:
Ok. I am just going through the existing list that says to enable this one. There are already many that have specific comments to not enable. Is there a way that list can be reviewed to identify any others that should remain disabled to avoid any wasted effort?

PR #2943, opened 2026-01-06
use instance_id instead of domain on provider level logging

fmunkes:
I have a (for me very hard to debug) issue report here music-assistant/support#4698, where the user has the progresses mixed up in a multi-user abs setup. If there is a bug, I personally do think it is likely to be outside of the abs provider, but would like to verify.

Is it acceptable to use the instance_id for logging on provider level, instead of the domain? This way, single instance providers still have the short domain in the log, but multi-instance providers are distinguishable in their log messages.


marcelveldt (approved):
Makes sense to me!

marcelveldt:
We could even consider using the provider.name as that will show the human friendly provider instance name ?

MarvinSchenkel:
Yep let's change this to `instance_name` and then we can merge this.

fmunkes:
if I use a pure
```
mass_logger.getChild(self.name)
```
and adapt the instance's name, the log looks a little weird I think:
```
2026-01-14 19:30:57.584 DEBUG (MainThread) [music_assistant.abs test instance] Updated 0 from full playlog.
```
Also, you can get rid of the provider type completely. I adapted it like so:
```
2026-01-14 19:37:48.086 DEBUG (MainThread) [music_assistant.audiobookshelf (abs test instance)] Updated 0 from full playlog.
```
Let me know what you think

halkeye:
fyi this casues issues for creating local providers, probably others - https://github.com/music-assistant/support/issues/4801#issuecomment-3762896297

MarvinSchenkel:
@fmunkes . Reverting this PR in `dev` as well. We have a few other ideas on solving this issue of identifying instances in log messages.

PR #2945, opened 2026-01-06
Fix players/get_by_name not always returning a result

OzGav:
Fixes https://github.com/orgs/music-assistant/discussions/4723

### DIFF
--- a/music_assistant/controllers/players/player_controller.py
+++ b/music_assistant/controllers/players/player_controller.py
@@ -385,10 +385,11 @@ class PlayerController(CoreController):
         if len(matches) > 1:
             player_ids = [p.player_id for p in matches]
             self.logger.warning(
-                "Multiple players found with name '%s': %s - returning first match. "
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
This will make the API endpoint ambiguous as it now searches for both the name and the display name. We could extend the endpoint to simply accept a new parameter (display_name) so people can search on display name as well.

Can you clarify the issue of the user? Was he searching by display_name?

OzGav:
Yes because that is pretty much all you know. The display name is obviously prominent in the UI but also if you call players/all then there are keys for name and display_name but they both show the display_name? So it is hard to get the value for name?

marcelveldt:
This entire logic can be simplified by just checking player.state.name as that will use the FINAL player name as it will also be visible in clients and api clients

MarvinSchenkel:
Looks good, let's merge this after today's release to be sure :)

MarvinSchenkel (approved):

PR #2953, opened 2026-01-08
Subsonic Scrobble: Fix Subsonic interactions after the async change

khers:
The Subsonic provider is now using aiohttp making it unnecessary for us to wrap calls to is in a new thread. Fix the scrobble plugin to use the async library as well.

### DIFF
--- a/music_assistant/providers/subsonic_scrobble/__init__.py
+++ b/music_assistant/providers/subsonic_scrobble/__init__.py
@@ -1,6 +1,5 @@
 """Allows scrobbling of tracks back to the Subsonic media server."""
 
-import asyncio
 import logging
 import time
 from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


Clusters:
Just checked out your branch @khers 
Works flawlessly with a fresh setup 👍

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -164,7 +152,12 @@ def handler(prov: OpenSonicProvider, item_id: str, uri: str) -> None:
 
         # the opensubsonic library is not async friendly,
```
Clusters: this comment can be removed I think

MarvinSchenkel (approved):
Looks good to me @khers . Just checking before I merge, is this safe to include in the next patch release (2.7.4)?

Clusters:
@MarvinSchenkel if #2910 is going to be included in the next patch release, this should be too, as the Scrobbler plugin won't work without. At least in my testing.

MarvinSchenkel:
Good catch, I already thought it was weird we did not have a dependency update :). If this needs backporting, we need to manually merge that PR too!

khers:
Thanks @Clusters that is what I was going to say. We need to back port both. I had wanted to leave them in beta for a cycle or two to shake out any bugs, but I have two reports that it solves the playback issues we are seeing in stable. So we should probably back port it now and I will just watch closely for bugs.

khers:
It the back port as simple as cherry-picking these to commits to the backport branch?

MarvinSchenkel:
It's as simple as adding the label to the PR (it's already there). GH actions will do the rest. I will make sure to also cherry pick #2910

PR #2954, opened 2026-01-08
Add track radio support to Qobuz

callym:
This is currently functional, **but** requires using the ID and secret from the official web interface, if not you get an unauthorized error.

I'm not sure if there's a way of getting Qobuz to give the official music-assistant app access to the radio API endpoints?

A lot of the code changes are to deal with having to use both app ids.

Another big chunk is because of different formats between what the other endpoints return for tracks vs what this endpoint returns. (also this endpoint seems to ignore the limit parameter when sent the same way as the other endpoints)

I've deliberately not included the webapp id needed for this PR to work as-is, so for testing you'll need to source it yourself.

### DIFF
--- a/music_assistant/providers/qobuz/__init__.py
+++ b/music_assistant/providers/qobuz/__init__.py
@@ -82,11 +82,15 @@ SUPPORTED_FEATURES = {
     ProviderFeature.SEARCH,
     ProviderFeature.ARTIST_ALBUMS,
     ProviderFeature.ARTIST_TOPTRACKS,
+    ProviderFeature.SIMILAR_TRACKS,
 }
 
 VARIOUS_ARTISTS_ID = "145383"
 
 CONF_QUALITY = "quality"
+
+WEBAPP_ID = "..."
+WEBAPP_SIG = "..."
 
 
 async def setup(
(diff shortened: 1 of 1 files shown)


music_assistant/providers/qobuz/__init__.py
```diff
@@ -655,6 +674,60 @@ async def _parse_album(
             album.metadata.explicit = True
         return album
 
+    async def _parse_track_radio(self, track_obj: dict[str, Any]) -> Track:
```
MarvinSchenkel, line 677: Why not re-use the existing `self._parse_track` here? We could enhance that cater for the differences between the endpoints.

MarvinSchenkel:
I am not a huge fan of all the changes that now require all the `_parse_xx` functions to be `async`. Can we avoid this by handling the token info `_auth_token` function (ie simply calling it once)?

MarvinSchenkel:
@marcelveldt Is there a way for us to get access to the radio endpoint via our own Qobuz application? I would prefer to use our own token instead of a (probably not allowed to use) token from Qobuz themselves.

marcelveldt:
> @marcelveldt Is there a way for us to get access to the radio endpoint via our own Qobuz application? I would prefer to use our own token instead of a (probably not allowed to use) token from Qobuz themselves.

No, Qobuz doesn't care about applications like us and they never respond to any of my enquiries.

marcelveldt:
> I am not a huge fan of all the changes that now require all the `_parse_xx` functions to be `async`. Can we avoid this by handling the token info `_auth_token` function (ie simply calling it once)?

Indeed, parse functions being async is a NO GO as it will bring a shitload of overhead. Parsers should be simple IO-free parsers of (raw) provider specific responses into MA models

MarvinSchenkel:

OzGav:
@callym any progress on this?

PR #2955, opened 2026-01-08
Prevent duplicate airplay player creation leaving player in invalid state

kdkavanagh:
Related to https://github.com/music-assistant/support/issues/4655

### DIFF
--- a/music_assistant/providers/airplay/provider.py
+++ b/music_assistant/providers/airplay/provider.py
@@ -45,8 +45,12 @@ class AirPlayProvider(PlayerProvider):
     _dacp_server: asyncio.Server
     _dacp_info: AsyncServiceInfo
 
+    _mdns_lock: asyncio.Lock
+
     async def handle_async_init(self) -> None:
         """Handle async initialization of the provider."""
+        self._mdns_lock = asyncio.Lock()
+
         # register DACP zeroconf service
         dacp_port = await select_free_port(39831, 49831)
         self.dacp_id = dacp_id = f"{randrange(2**64):X}"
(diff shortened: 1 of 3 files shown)


music_assistant/providers/airplay/provider.py
```diff
@@ -88,22 +92,27 @@ async def on_mdns_service_state_change(
...
-        if player := cast("AirPlayPlayer | None", self.mass.players.get(player_id)):
-            # update the latest discovery info for existing player
-            player.set_discovery_info(info, display_name)
-            return
-        await self._setup_player(player_id, display_name, info)
+
+        # We are going to process this event, ensure that they're processed in-sequence
+        # so as to avoid creating duplicate players and instead ensuring that one is
+        # created and then updated by subsequent state changes
+        async with self._mdns_lock:
```
marcelveldt: Instead of implementing it here, I think we should implement it at core level, where we keep a lock per provider(domain) as this can theoretically also happen for other/future providers that listen to multiple mdns services.

MarvinSchenkel: Just to help @kdkavanagh , I think you mean to implement it [here](https://github.com/music-assistant/server/blob/499012764e382047018214154029d10d858b03f1/music_assistant/mass.py#L915), right?

marcelveldt:
I adjusted your Pr so the lock is now applied at global level so it will work for any provider with mdns discovery.

music_assistant/providers/airplay/provider.py
marcelveldt: Considering this is a rather important bugfix, I moved ahead and adjusted the PR to do this fix at global level.

marcelveldt (approved):
Thanks @kdkavanagh !

PR #2957, opened 2026-01-09
fix: Ensure playlist ID is a string in Soundcloud provider

robsonke:
Convert playlist ID to string to prevent an error on startsWith. Most of the playlist id's are just names but sometimes it's an actual id. This should fix that.

See discussion here: https://github.com/orgs/music-assistant/discussions/1160#discussioncomment-15449676


MarvinSchenkel (approved):
Thanks @robsonke

PR #2958, opened 2026-01-10
Add Tag Player plugin for linking identifiers to media

ztripez:
## Summary

Adds a Tag Player **plugin** that links arbitrary identifiers (NFC tags, QR codes, or any string) to media items for quick playback.

**Complete rewrite** based on reviewer feedback from @marcelveldt and @MarvinSchenkel:
- Changed from `MusicProvider` → `PluginProvider` (manifest type: `plugin`)
- Replaced custom SQLite table with **provider mappings** (`available=False`)
- URI resolution uses existing `get_library_item_by_prov_id` SQL lookups — no custom DB code
- Cleanup is automatic: when the provider is removed, MA removes all its mappings

## Personal note

I built a tag player for my kids using NFC tags. Initially I made a Home Assistant automation that acted like an abomination of a tag database, but Music Assistant is already the source of truth for my media library — the tag mappings belong here.

## How it works

Tags are stored as **provider mappings** on existing library items with `available=False` (so MA never tries to stream through tagplayer). The tag ID becomes the `item_id` in the mapping, and resolution uses the standard `provider_mappings` table lookup.

### API Commands

| Command | Description |
|---------|-------------|
| `tagplayer/link` | Link a tag ID to a media item (e.g., `target="playlist/42"`) |
| `tagplayer/unlink` | Remove a tag mapping |
| `tagplayer/get` | Get a single tag's mapping info |
| `tagplayer/list` | List all tag mappings across all media types |
| `tagplayer/play` | Play a tag directly on a player (primary interface for NFC/QR → HA → playback) |

### URI format

`tagplayer://<media_type>/<tag_id>` — media type must match the linked item type (e.g., `tagplayer://playlist/party-mix` for a playlist tag).

### Usage example

```python
# Link any string to a playlist
await mass.send_command("tagplayer/link", tag_id="party-mix", target="playlist/42")

# Play directly via API (primary use case for HA automations)
await mass.send_command("tagplayer/play", tag_id="party-mix", player_id="living_room")

# Also works with NFC tag serials
await mass.send_command("tagplayer/link", tag_id="04:A3:22:1A:5E:3B:80", target="audiobook/5")
```

## Supported media types

Tracks, albums, playlists, artists, radio, audiobooks, podcasts

## Tests

24 unit tests covering lifecycle, link/unlink/get/list, playback, error handling, and target parsing.


music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+
+    async def _create_db_table(self) -> None:
+        """Create the tagplayer mappings table if it doesn't exist."""
+        await self.mass.music.database.execute(
+            f"""CREATE TABLE IF NOT EXISTS {DB_TABLE_TAGPLAYER}(
+                [tag_id] TEXT PRIMARY KEY,
+                [media_type] TEXT NOT NULL,
+                [item_id] INTEGER NOT NULL
+            );"""
+        )
```
marcelveldt: Why would you need a separate db ? You can simply use provider mappings to existing items.

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+            item_id = int(item_id_str)
+        except (ValueError, KeyError) as err:
+            msg = f"Invalid target format: {target}. Expected format: type/id (e.g., track/42)"
+            raise ValueError(msg) from err
+
+        await self._get_library_item(media_type, item_id)
+        await self.mass.music.database.insert_or_replace(
+            DB_TABLE_TAGPLAYER,
+            {"tag_id": tag_id, "media_type": media_type.value, "item_id": item_id},
+        )
```
marcelveldt: We cant accept music providers doing direct db modifications

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,235 @@
...
+
+    async def unlink_tag(self, tag_id: str) -> dict[str, str]:
+        """
+        Remove a tag mapping.
+
+        :param tag_id: The tag identifier to unlink.
+        """
+        await self.mass.music.database.delete(DB_TABLE_TAGPLAYER, {"tag_id": tag_id})
+        self.logger.debug("Unlinked tag '%s'", tag_id)
+        return {"status": "ok", "tag_id": tag_id}
```
marcelveldt: See my previous comment - this is unacceptable. My suggestion would be that you add a provider mapping to an existing item with the available flag set to False (to prevent playback being attempted to the provider).

music_assistant/providers/tagplayer/__init__.py
ztripez: I don't,
I think i was so deep into my own thing it never occurred to me to use provider mappings.

music_assistant/providers/tagplayer/__init__.py
ztripez: Fair enough.

music_assistant/providers/tagplayer/__init__.py
ztripez: Agreed

OzGav:
As you get closer to merging I will just need some text for the docs

MarvinSchenkel:
Just thinking out loud here.. Does it make sense that this functionality is implemented as a `MusicProvider`? I know technically it is now the only way to make it work, but to me this feels more like a plugin.

Thoughts?

marcelveldt:
> Just thinking out loud here.. Does it make sense that this functionality is implemented as a `MusicProvider`? I know technically it is now the only way to make it work, but to me this feels more like a plugin.
> 
> Thoughts?

I never tested this but theoretically I think you can just add any provider mapping to items. Might be worth a test.
So you have a plugin that can register additional provider mappings to existing music library items.

MarvinSchenkel:

ztripez:
> > Just thinking out loud here.. Does it make sense that this functionality is implemented as a `MusicProvider`? I know technically it is now the only way to make it work, but to me this feels more like a plugin.
> > Thoughts?
> 
> I never tested this but theoretically I think you can just add any provider mapping to items. Might be worth a test. So you have a plugin that can register additional provider mappings to existing music library items.

Do you want me to explore this instead?

ztripez:
Rewrite complete!

Based on the feedback from @marcelveldt and @MarvinSchenkel, this is now a **PluginProvider** that uses **provider mappings** instead of a custom SQLite table:

- `PluginProvider` subclass (manifest type: `plugin`)
- Tags stored as `ProviderMapping(available=False)` on existing library items
- `available=False` means MA never tries to stream through tagplayer, but URI resolution still works via the standard `get_library_item_by_prov_id` SQL lookup
- No direct DB access — everything goes through the existing controller API
- Automatic cleanup when the provider is removed

The 5 API commands (`link`, `unlink`, `get`, `list`, `play`) remain the same. The `play` command is the primary interface for NFC/QR → Home Assistant automation → playback.

24 unit tests included. Ready for review!

marcelveldt:
So my idea actually worked?

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,230 @@
...
+
+# Media types that support tag mappings
+TAGGABLE_MEDIA_TYPES = (
+    MediaType.TRACK,
+    MediaType.ALBUM,
+    MediaType.PLAYLIST,
+    MediaType.ARTIST,
+    MediaType.RADIO,
+    MediaType.AUDIOBOOK,
+    MediaType.PODCAST,
```
MarvinSchenkel, line 40: Let's add the new `MediaType.GENRE` here as well?

music_assistant/providers/tagplayer/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "plugin",
+  "domain": "tagplayer",
+  "stage": "alpha",
+  "name": "Tag Player",
+  "description": "Link NFC tags, QR codes, or any identifier to media items for quick playback.",
+  "codeowners": ["@ztripez"],
+  "requirements": [],
+  "documentation": "https://music-assistant.io/music-providers/tagplayer/",
```
MarvinSchenkel, line 9: This should point to `/plugins/` now

MarvinSchenkel:
Few small things and then this should be good to go. Would be awesome if you can add some documentation, especially about how to set it up from a HA perspective as well. FYI, we moved the docs [here](https://github.com/music-assistant/music-assistant.io/tree/beta) since we launched a new website. Marking this as draft again, please set it to 'ready for review' when you want us to have another look.

MarvinSchenkel:
@ztripez any progress on this one?

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    async def _find_tagged_item(self, tag_id: str) -> tuple[MediaType, MediaItemType] | None:
+        """Search all media types for a library item with the given tag mapping.
+
+        :param tag_id: The tag identifier to search for.
+        """
+        for media_type in TAGGABLE_MEDIA_TYPES:
+            controller = self.mass.music.get_controller(media_type)
+            if item := await controller.get_library_item_by_prov_id(tag_id, self.instance_id):
+                return (media_type, item)
+        return None
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in f76aa9bdc — `tag_id` is now validated in `_find_tagged_item`, so all tag commands (`unlink`/`get`/`play`) reject empty or whitespace-only input with a clean `ValueError` instead of an `AssertionError`. Tests added.

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+
+    async def loaded_in_mass(self) -> None:
+        """Register API commands after the provider is loaded."""
+        self._unregister_commands = [
+            self.mass.register_api_command("tagplayer/link", self.link_tag),
+            self.mass.register_api_command("tagplayer/unlink", self.unlink_tag),
+            self.mass.register_api_command("tagplayer/get", self.get_tag),
+            self.mass.register_api_command("tagplayer/list", self.list_tags),
+            self.mass.register_api_command("tagplayer/play", self.play_tag),
+        ]
```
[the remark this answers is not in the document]
OzGav: Added

music_assistant/providers/tagplayer/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+    async def list_tags(self) -> list[dict[str, Any]]:
+        """List all tag mappings."""
+        tags: list[dict[str, Any]] = []
+        for media_type in TAGGABLE_MEDIA_TYPES:
+            controller = self.mass.music.get_controller(media_type)
+            items = await controller.get_library_items_by_prov_id(
+                provider_instance=self.instance_id
+            )
```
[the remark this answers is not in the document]
OzGav: Fixed

OzGav (dismissed):
I think this is good to go

OzGav:
We are not seeing demand for this so maybe enthusiasm has waned?

OzGav:
Closing for now.

PR #2959, opened 2026-01-12
Fix issues with progress bar jumps and time overflow 

MarvinSchenkel:
This PR fixes a couple of issues that caused progress bar jumps and overflows:
- Clearing the playlog on a stream restart avoids duplicates in the playlog. This caused the overflow to happen in flow mode
- Send the server side 'elapsed_time_updated_at' as part of the QUEUE_TIME_UPDATED event ensures the frontend can accurately extrapolate elapsed_time. This avoids different elapsed_time with multiple browser open. [See Discord](https://discord.com/channels/753947050995089438/1458038874100994129)
- Pausing in Airplay caused the elapsed time to continue anyway. We now count the pause time to tackle this
- Pausing in Sqeezelite causes weird jumps because of inaccurate elapsed_time reported by some players. We now guard for this. 

Depends on PRs, will bump the frontend after its been merged:
- Frontend: https://github.com/music-assistant/frontend/pull/1364

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -59,7 +59,7 @@ from music_assistant_models.media_items import (
     media_from_dict,
 )
 from music_assistant_models.playback_progress_report import MediaItemPlaybackProgressReport
-from music_assistant_models.player_queue import PlayerQueue, QueueTimeUpdate
+from music_assistant_models.player_queue import PlayerQueue
 from music_assistant_models.queue_item import QueueItem
 
 from music_assistant.constants import (
@@ -2217,10 +2217,10 @@ class PlayerQueuesController(CoreController):
                 self.mass.signal_event(
                     EventType.QUEUE_TIME_UPDATED,
                     object_id=queue_id,
-                    data=QueueTimeUpdate(
-                        elapsed_time=queue.elapsed_time,
-                        elapsed_time_last_updated=queue.elapsed_time_last_updated,
-                    ),
+                    data={
+                        "elapsed_time": queue.elapsed_time,
+                        "elapsed_time_last_updated": queue.elapsed_time_last_updated,
+                    },
                 )
                 # also signal update to the player itself so it can update its current_media
                 self.mass.players.trigger_player_update(queue_id)
@@ -2302,6 +2302,7 @@ class PlayerQueuesController(CoreController):
     ) -> tuple[int | None, int]:
         """Calculate current queue index and current track elapsed time when flow mode is active."""
         elapsed_time_queue_total = player.corrected_elapsed_time or 0
+
         if queue.current_index is None and not queue.flow_mode_stream_log:
             return queue.current_index, int(queue.elapsed_time)
 
@@ -2340,6 +2341,7 @@ class PlayerQueuesController(CoreController):
             # if the player is not playing, we can't be sure that the elapsed time is correct
             # so we just return the queue index and the elapsed time
             return queue.current_index, int(queue.elapsed_time)
+
         return queue_index, int(track_time)
 
     def _parse_player_current_item_id(self, queue_id: str, player: Player) -> str | None:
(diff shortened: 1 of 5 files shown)


marcelveldt:
imo there is zero need for elapsed_time_updated_at being a dedicated schema as that will break backwards compatibility with clients (such as HA) while there is technical need to do so. You can simply use the current (corrected) elapsed time in the queue time update message and assume the "updated at" is NOW - it is offset by the latency of the connection which is just a few ms

music_assistant/controllers/streams/streams_controller.py
```diff
@@ -590,6 +590,9 @@ async def serve_queue_flow_stream(self, request: web.Request) -> web.StreamRespo
         if not start_queue_item:
             raise web.HTTPNotFound(reason=f"Unknown Queue item: {start_queue_item_id}")
 
+        # make sure to start the stream with a fresh log
```
marcelveldt:
```suggestion
```

music_assistant/controllers/player_queues.py
```diff
@@ -2214,7 +2216,10 @@ def _update_queue_from_player(
                 self.mass.signal_event(
                     EventType.QUEUE_TIME_UPDATED,
                     object_id=queue_id,
-                    data=queue.elapsed_time,
+                    data=QueueTimeUpdate(
+                        elapsed_time=queue.elapsed_time,
+                        elapsed_time_last_updated=queue.elapsed_time_last_updated,
+                    ),
```
marcelveldt: this is a breaking change I want to avoid

MarvinSchenkel: I think you are talking about app.music-assistant.io here? I added a check for that in the front-end to make it non-breaking.

MarvinSchenkel:
> imo there is zero need for elapsed_time_updated_at being a dedicated schema as that will break backwards compatibility with clients (such as HA) while there is technical need to do so. You can simply use the current (corrected) elapsed time in the queue time update message and assume the "updated at" is NOW - it is offset by the latency of the connection which is just a few ms

Well that was my first approach, but it does seem to not fully fix the issue. I will give it another go just to be sure though. I would also prefer if we can avoid the extra schema

MarvinSchenkel:
Indeed seems to work fine using just `Date.now() / 1000`. Removed the models change.

music_assistant/controllers/player_queues.py
marcelveldt: No, all clients will use this, including HA.
So this small change will break HA compatibility

music_assistant/providers/squeezelite/player.py
```diff
@@ -458,12 +464,14 @@ async def _handle_play_url_for_slimplayer(
...
-            # ignore server heartbeats when stopped
+        if self.playback_state != PlaybackState.PLAYING:
+            # ignore server heartbeats when not playing
+            # Some players keep sending heartbeat with increasing elapsed time
+            # even when paused (e.g. WiiM)
             return
         # elapsed time change on the player will be auto picked up
         # by the player manager.
-        self._attr_elapsed_time = self.client.elapsed_seconds
+        self._attr_elapsed_time = self.client._elapsed_milliseconds / 1000
```
marcelveldt: why is this ?
elapsed_seconds is already calculated here from milliseconds in the property

MarvinSchenkel: Because `elapsed_seconds` is also extrapolated using `time.time()`. This means we are doing double extrapolation (also in the Player base class). Thinking about this, another option would be to override that property and just return `self.client.elapsed_seconds` without extrapolation.

MarvinSchenkel: Just did some testing and it works fine with just `elapsed_seconds` now that heartbeat is fixed.

marcelveldt: yeah, that is what I thought

marcelveldt (approved):

PR #2963, opened 2026-01-13
Add ASYNC109 mypy rule

OzGav:


music_assistant/helpers/auth.py
```diff
@@ -58,27 +58,27 @@ async def __aexit__(
...
 
-    async def authenticate(self, auth_url: str, timeout: int = 60) -> dict[str, str]:
+    async def authenticate(self, auth_url: str, timeout_seconds: int = 60) -> dict[str, str]:
         """
         Start the auth process and return any query params if received on the callback.
 
         Params:
         - url: The URL the user needs to open for authentication.
-        - timeout: duration in seconds helpers waits for callback (default: 60).
+        - timeout_: duration in seconds helpers waits for callback (default: 60).
```
uSpike:
```suggestion
        - timeout_seconds: duration in seconds helpers waits for callback (default: 60).
```

MarvinSchenkel:
I think it's a bit weird that `ASYNC109` is choking on the parameter _names_. There's no mention about this [here](https://docs.astral.sh/ruff/rules/async-function-with-timeout/). @marcelveldt thoughts?

EDIT: Looking at this a bit more, this rule only _trigger_ based on parameters that are called `timeout`, so changing the parameter names to `timeout_seconds` is just a (dirty) workaround that (on the surface) satisfies the rule, but does not fix the underlying problem.

To be honest, I don't really see this as a problem, as the rule only triggers on helpers that intentionally have the timeout configurable, as they can be reused by multiple parts of the code (that might require different timeout values).

OzGav:
Is it that this is working around the rule or just making a change so this rule isn’t triggered when it doesn’t apply here? If the rule is one that is useful to have in place for when it is relevant then the change here would seem to be appropriate as it will allow the rule to be enabled?

Or are you saying that the rule has identified a problem here and I should properly address what is being flagged?

MarvinSchenkel:
> Is it that this is working around the rule or just making a change so this rule isn’t triggered when it doesn’t apply here? If the rule is one that is useful to have in place for when it is relevant then the change here would seem to be appropriate as it will allow the rule to be enabled?
> 
> Or are you saying that the rule has identified a problem here and I should properly address what is being flagged?

I'm saying that the rule doesn't trigger anymore because of renaming the parameter name, but that's actually not what this rule is trying to prevent.

However, in our case, I think we have a valid reason for passing timeouts as a parameter so I think we should keep ignoring this rule.

OzGav:
Still if this rule is a reasonable one to have (I.e. it will catch legitimate issues) then why not rename these parameters to fix the false positive and then enable the rule? Or have I misunderstood?

MarvinSchenkel:
Closing this one as it is not relevant for us.

PR #2965, opened 2026-01-13
Pin `av` to 16.0.1 for sendspin provider

maximmaxim345:
Pins PyAV to 16.0.1 to match the pre-built wheel in the base image. Without this pin, aiosendspin's `av>=16.0.0` constraint could resolve to a different version, causing Docker build failures.


maximmaxim345:
We only have a single base image for beta and stable builds.

PR #2968, opened 2026-01-13
orf radiothek provider

DButter:
This is a somewhat WIP PR at integrating the austrian public broadcast stations into ma 

https://sound.orf.at

It adds the live stations as stations, the older programs and actual podcasts as podcasts. I'm more of an infrastructure guy so tbh. I'm not too sure about the quality of my code.


DButter:
:D i maybe should mention, while it is a WIP, the base functionality is working, everything gets the correct pictures, etc

what I'm not too sure about is, the orf only publishes stuff for about one month, afterwards it gets depublished, as far as i understand it the caching settings should throw stuff out thats too old

music_assistant/providers/orf_radiothek/helpers.py
```diff
@@ -0,0 +1,380 @@
+"""Typed dataclasses + parsers for ORF Radiothek / ORF Sound provider."""
```
MarvinSchenkel, line 1: I would rename this file to something like `helpers.py`. Or split it into `models.py` and `parsers.py` to be a bit more descriptive. Up to you though!

music_assistant/providers/orf_radiothek/__init__.py
```diff
@@ -0,0 +1,1150 @@
...
+    instance_id: str | None = None,
+    action: str | None = None,
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return provider configuration entries."""
+    # ruff: noqa: ARG001
+    values = values or {}
+
+    return (
+        ConfigEntry(
```
MarvinSchenkel, line 124: Is there any reason why a user might want to configure this? If not, just pick a sane default and get rid of the configuration option. If there is a good reason, it might be good to hide this under the `category='advanced'`

music_assistant/providers/orf_radiothek/__init__.py
```diff
@@ -0,0 +1,1150 @@
...
+            label="Preferred ORF protocol",
+            required=False,
+            default_value="hls",
+            description=(
+                "Used for ORF stations (template-based). "
+                "Privates use explicit URLs from bundle.json."
+            ),
+            value=values.get(CONF_STREAM_PROTO),
+        ),
+        ConfigEntry(
```
MarvinSchenkel, line 137: Also sounds like an 'advanced' option to me

music_assistant/providers/orf_radiothek/__init__.py
```diff
@@ -0,0 +1,1150 @@
...
+            label="Catch-up retention (days)",
+            required=False,
+            default_value=30,
+            description=(
+                "Expose broadcasts as podcast episodes for the last N days. "
+                "Older items vanish on next sync."
+            ),
+            value=values.get(CONF_CATCHUP_DAYS),
+        ),
+        ConfigEntry(
```
MarvinSchenkel, line 157: Please move this to the advanced category

MarvinSchenkel:
Looks like you are on the right track 🎉 .

Two questions from my side:
- What's this concept of 'catchup stations' ?
- Why are you using the intermediate dataclasses? For example, you are first parsing a response to an `OrfPodcastEpisode`, and then parsing that to a `PodcastEpisode` (MA `MediaItem`). At first sight, this feels like a lot of unnecessary code / complexion. Why not parse straight from response -> `PodcastEpisode` ?

DButter:
> Looks like you are on the right track 🎉 .
> 
> Two questions from my side:
> 
>     * What's this concept of 'catchup stations' ?

On the official ORF site, there are three different kinds of audio categories:

1. the live broadcast, 
2. live broadcast material that you can replay for up to one month (to catch up), after which it is deleted, and 
3. actual podcasts, which are stored for a longer period of time.

 
>     * Why are you using the intermediate dataclasses? For example, you are first parsing a response to an `OrfPodcastEpisode`, and then parsing that to a `PodcastEpisode` (MA `MediaItem`). At first sight, this feels like a lot of unnecessary code / complexion. Why not parse straight from response -> `PodcastEpisode` ?

When I started, I didn’t like the idea of parsing the JSON directly, because developers can’t really know or rely on what an external API returns.

MarvinSchenkel:
> > Looks like you are on the right track 🎉 .
> > Two questions from my side:
> > ```
> > * What's this concept of 'catchup stations' ?
> > ```
> 
> On the official ORF site, there are three different kinds of audio categories:
> 
> 1. the live broadcast,
> 2. live broadcast material that you can replay for up to one month (to catch up), after which it is deleted, and
> 3. actual podcasts, which are stored for a longer period of time.
> 
> > ```
> > * Why are you using the intermediate dataclasses? For example, you are first parsing a response to an `OrfPodcastEpisode`, and then parsing that to a `PodcastEpisode` (MA `MediaItem`). At first sight, this feels like a lot of unnecessary code / complexion. Why not parse straight from response -> `PodcastEpisode` ?
> > ```
> 
> When I started, I didn’t like the idea of parsing the JSON directly, because developers can’t really know or rely on what an external API returns.

In that case, I would not make 'catch up days' configureable and just rely on the MA sync to remove those items when they become unavailable.

I would still suggest to parse into MA MediaItems (such as Podcast) directly. This is their exact purpose: parsing unstructured JSON responses into a typed data object. There's no need for the extra intermediate layer.

DButter:
fair, ill rework it

DButter:
> In that case, I would not make 'catch up days' configureable and just rely on the MA sync to remove those items when they become unavailable.

i do, but the api for the catchup is api/yyymmdd per day to be synced, the config value just tells music assistant how many days in the past it should query.

MarvinSchenkel:
> > In that case, I would not make 'catch up days' configureable and just rely on the MA sync to remove those items when they become unavailable.
> 
> i do, but the api for the catchup is api/yyymmdd per day to be synced, the config value just tells music assistant how many days in the past it should query.

Sure, but your code comment mentions `# default 30 (ORF keeps ~30 days)`. So my question is, why would the user need to change this value? Can we just hide it from the user and use a constant here?

DButter:
>  So my question is, why would the user need to change this value? Can we just hide it from the user and use a constant here?

This was only to reduce the amount of content in the “Podcast” stations, but you’re right—we could just pin it to 30 days.

> I would still suggest to parse into MA MediaItems (such as Podcast) directly. This is their exact purpose: parsing unstructured JSON responses into a typed data object. There's no need for the extra intermediate layer.

I started the refactor, and almost immediately began creating a JSON structure to ensure we get clear error messages if the ORF API responses inadvertently change and break the provider.

I guess for podcasts I can see why a dedicated data type isn’t really necessary. it’s basically just get podcasts, get podcast, get episodes, get episode. But for catch-up, I’m using data from the stations and the bundle, so having a data type here actually feels useful.

DButter:
@MarvinSchenkel I’m pretty happy with how it looks now. The data classes still feel useful to me, but if they really bother you, we can drop them and go simpler.

music_assistant/providers/orf_radiothek/__init__.py
```diff
@@ -0,0 +1,1055 @@
...
+
+
+async def get_config_entries(
+    mass: MusicAssistant,
+    instance_id: str | None = None,
+    action: str | None = None,
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return provider configuration entries."""
+    # ruff: noqa: ARG001
```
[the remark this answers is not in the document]
DButter: thats just plain wrong

`ARG001 Unused function argument: `mass`
   --> providers/orf_radiothek/__init__.py:115:5
    |
114 | async def get_config_entries(
115 |     mass: MusicAssistant,
    |     ^^^^
116 |     instance_id: str | None = None,
117 |     action: str | None = None,
    |

ARG001 Unused function argument: `instance_id`
   --> providers/orf_radiothek/__init__.py:116:5
    |
114 | async def get_config_entries(
115 |     mass: MusicAssistant,
116 |     instance_id: str | None = None,
    |     ^^^^^^^^^^^
117 |     action: str | None = None,
118 |     values: dict[str, ConfigValueType] | None = None,
    |

ARG001 Unused function argument: `action`
   --> providers/orf_radiothek/__init__.py:117:5
    |
115 |     mass: MusicAssistant,
116 |     instance_id: str | None = None,
117 |     action: str | None = None,
    |     ^^^^^^
118 |     values: dict[str, ConfigValueType] | None = None,
119 | ) -> tuple[ConfigEntry, ...]:
    |

Found 3 errors.`

music_assistant/providers/orf_radiothek/__init__.py
MarvinSchenkel: I think you forgot to move this to the advanced category?

MarvinSchenkel:
> @MarvinSchenkel I’m pretty happy with how it looks now. The data classes still feel useful to me, but if they really bother you, we can drop them and go simpler.

They don't bother me per se. The reason I mentioned this, is that it introduces another layer of complexity that might make it harder for other contributors to improve / contribute to this provider. Honestly, this mostly comes down to a matter of preference / style, so I am not gonna push this.

If you're okay with this, then we can merge this. I think you forgot to move 1 config entry to the advanced category, but other than that this should be good to go!

DButter:
> If you're okay with this, then we can merge this. I think you forgot to move 1 config entry to the advanced category, but other than that this should be good to go!

yup, I've fixed the last one, i'd be happy for it to be merged

MarvinSchenkel (approved):
Thanks @DButter for this awesome new provider :)

DButter:
@MarvinSchenkel this might be a stupid question, how is the documentation added on https://music-assistant.io/music-providers/radiothek?

MarvinSchenkel:
> @MarvinSchenkel this might be a stupid question, how is the documentation added on https://music-assistant.io/music-providers/radiothek?

Since we are running this off beta first, you can open a PR to [this repo](https://github.com/music-assistant/beta.music-assistant.io/pulls) in order to get the documentation merged. Have a look at the existing providers for a template, for example [Pandora](https://github.com/music-assistant/beta.music-assistant.io/pull/65/changes).

OzGav:
@DButter you can also just give me the info required and I will create the page.

DButter:
@OzGav that would be very nice of you, i've already prepared a fork https://github.com/DButter/beta.music-assistant.io the PR just showed me more commit's than i expected and i haven't had time to check what i did wrong

OzGav:
@DButter Probably nothing as that repo is a fork of the main site so it is a bunch of commits ahead. Let me make a start and I will have some questions for you soon.

OzGav:
@DButter questions:
Do you know what max stream quality is? The audio pipeline should show you.
Can you provide more detail about what each config option is for and how it should be used.
In your draft you had subscriptions free as no but I don’t see any config options to login so I assume this is actually yes?

OzGav:
Draft is here https://github.com/music-assistant/beta.music-assistant.io/pull/68

You will see the placeholders for the questions above

DButter:
> In your draft you had subscriptions free as no but I don’t see any config options to login so I assume this is actually yes?

| Subscription Required | No (some content may be region-locked) |

ive got subscription required and yours is subscription free, thats why the answer is the opposite but yes there is no subscription rquired

Prefered protocoll you can choose between # hls | shoutcast
quality you can choose between # hls: q1a/q2a/q3a/q4a/qxa ; shoutcast: q1a/q2a
there are some station that are hidden in the api, no idea why or if they are any good
catch-up stream you also got protocoll choices # progressive | hls
Catch-up stations, here you can choose which stations ma should query for the catchup stations, mostly good for reducing clutter all of them would be (bgl fm4 ktn noe oe1 oe3 ooe sbg stm tir vbg wie)

ngl. unless your connectivity is really shitty you never would need to touch any of these other than the last one

DButter:
@OzGav   44.1 kHz / 16bits

DButter:
container and codec aac

OzGav:
Thanks for that but for the users why would they change the protocol or quality? If there is no reason then perhaps we should just simplify this and always provide the highest quality? Most providers do that.
Edit: And what do all those q values equate to anyway?
Edit2: ok I see your comment about leaving these alone unless you have bandwidth problems. I can make mention of that but would still be good to explain what all the options are

DButter:
> And what do all those q values equate to anyway?

Q1A is the lowest, Q4A is the highest, and QXA is dynamic, it’s up to the radio stream provider, and I haven’t tested how they’ve implemented the different quality levels.

@OzGav

OzGav:
I think what would be good is to make all of the fixed value entries as a dropdown. For example,
```
            ConfigEntry(
                key=CONF_QUALITY,
                type=ConfigEntryType.STRING,
                label="Quality setting for Tidal:",
                description="High = 16bit 44.1kHz\n\nMax = Up to 24bit 192kHz",
                options=[
                    ConfigValueOption("High", "LOSSLESS"),
                    ConfigValueOption("Max", "HI_RES_LOSSLESS"),
                ],
                default_value="HI_RES_LOSSLESS",
            ),
        )
```

The description doesn’t need to list the value for each one but maybe “ Q1A = lowest, Q4A = highest, and QXA = dynamic”. I’ll add some info to the docs but if you have any more info about the hls / progressive option that would be good.

OzGav:
The docs are now live on beta.music-assistant.io but happy for any more improvements,

DButter:
@OzGav looks good, ill improve the inputs once ive collected some after improvements based on user feedback

OzGav:
Hi @DButter . We are looking to standardise the behaviour of all the radio providers. I have a question about ORF Radiothek. You have added `ProviderFeature.LIBRARY_RADIOS` but this is actually for when a provider has its own library and you want to sync that to MA. For example, TUNEIN you can "favourite" stations on their website and this then appears on a page called LIBRARY. So my question is, does ORF Radiothek have a library concept? If not, I will remove that provider feature and the `get_library_radios` method and implement `browse()` instead.

DButter:
@OzGav ah, i see, now i missunderstood what this method does then, it has a bunch of radios on the site, but since there is no account you cant really favorite any of them, it would only change if the stations themself changed

OzGav:
Thanks sounds good. I’ll add the change to my PR standardising this and I’ll let you know if I have any problems.

OzGav:
OK I have changed browse to this
image

That seems to be working ok but the only thing I am confused about is the config option for catch up stations. There seems to be a lot listed by default in the podcasts so why not just make them all available instead of having the config option?

This is browsing the podcasts
image

PR #2969, opened 2026-01-13
Adds date_added field to Tidal provider

FL550:
This adds the date_added field to artists, albums, tracks and playlists from the "created" tag in the Tidal api. This allows the user to sort based on when he added the entity to his tidal library.

### DIFF
--- a/music_assistant/providers/tidal/api_client.py
+++ b/music_assistant/providers/tidal/api_client.py
@@ -164,7 +164,6 @@ class TidalAPIClient:
         self,
         endpoint: str,
         item_key: str = "items",
-        nested_key: str | None = None,
         limit: int = 50,
         cursor_based: bool = False,
         **kwargs: Any,
@@ -192,10 +191,7 @@ class TidalAPIClient:
                 break
 
             for item in items:
-                if nested_key and nested_key in item and item[nested_key]:
-                    yield item[nested_key]
-                else:
-                    yield item
+                yield item
 
             if cursor_based:
                 cursor = response.get("cursor")
--- a/music_assistant/providers/tidal/library.py
+++ b/music_assistant/providers/tidal/library.py
@@ -31,22 +31,22 @@ class TidalLibraryManager:
     async def get_artists(self) -> AsyncGenerator[Artist, None]:
         """Retrieve library artists."""
         path = f"users/{self.auth.user_id}/favorites/artists"
-        async for item in self.api.paginate(path, nested_key="item"):
-            if item and item.get("id"):
+        async for item in self.api.paginate(path):
+            if item and item.get("item") and item["item"].get("id"):
                 yield parse_artist(self.provider, item)
 
     async def get_albums(self) -> AsyncGenerator[Album, None]:
         """Retrieve library albums."""
         path = f"users/{self.auth.user_id}/favorites/albums"
-        async for item in self.api.paginate(path, nested_key="item"):
-            if item and item.get("id"):
+        async for item in self.api.paginate(path):
+            if item and item.get("item") and item["item"].get("id"):
                 yield parse_album(self.provider, item)
 
     async def get_tracks(self) -> AsyncGenerator[Track, None]:
         """Retrieve library tracks."""
         path = f"users/{self.auth.user_id}/favorites/tracks"
-        async for item in self.api.paginate(path, nested_key="item"):
-            if item and item.get("id"):
+        async for item in self.api.paginate(path):
+            if item and item.get("item") and item["item"].get("id"):
                 yield parse_track(self.provider, item)
 
     async def get_playlists(self) -> AsyncGenerator[Playlist, None]:
@@ -60,8 +60,8 @@ class TidalLibraryManager:
 
         # 2. Get user playlists
         path = f"users/{self.auth.user_id}/playlistsAndFavoritePlaylists"
-        async for item in self.api.paginate(path, nested_key="playlist"):
-            if item and item.get("uuid"):
+        async for item in self.api.paginate(path):
+            if item and item.get("playlist") and item["playlist"].get("uuid"):
                 yield parse_playlist(self.provider, item)
 
     async def add_item(self, item: MediaItemType) -> bool:
--- a/music_assistant/providers/tidal/parsers.py
+++ b/music_assistant/providers/tidal/parsers.py
@@ -173,7 +173,7 @@ def parse_track(
     lyrics: dict[str, str] | None = None,
 ) -> Track:
     """Parse tidal track object to generic layout."""
-    track_obj_data = track_obj["item"]
+    track_obj_data = track_obj.get("item", track_obj)
     name, version = parse_title_and_version(
         track_obj_data.get("title", "Unknown"),
         track_obj_data.get("version") or None,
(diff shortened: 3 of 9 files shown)


MarvinSchenkel:
Please have a look at Copilot's suggestions. Mainly:
1. Wrap the parsing of timestamps  in a try/catch
2. Retain the old behaviour with pagination.
So instead of
```
async for item in self.api.paginate(path):
  if item:
```

Keep
```
async for item in self.api.paginate(path, nested_key="item"):
  if item and item.get("id"):
```
and move the parsing of the `date_added` field to the appropriate parsers in `parsers.py` (e.g. `parse_artists()`, `parse_albums()` etc).

That should fix the failing tests

MarvinSchenkel:
Looping in @jozefKruszynski for a second pair of eyes 👀

FL550:
@MarvinSchenkel Thanks for the review. 

Regarding the modified call to the pagination: the required key is one level above the nested_key "item", so I have to extract this first. And all ruff tests had no issues with this code change.

FL550:
Ah sorry, my mistake. I just noted the failing github actions. Nevertheless the issue remains the same and I have to fix the integration test therefore.

jozefKruszynski:
From what I can see when running some tests, this "created" field is none existent on any media item.

Albums and tracks have the field "streamStartDate", this is anyway passed through to the parser as there is no modification of the raw json data prior to hitting the parser.
Artists have no "created" or any other identifiable similar field,
Playlists have the "dateAdded" and "dateUpdated" field.
Mix type playlists also have no identifiable "created" or similar field

There is no need to add anything to the library.py module as this PR is related to parsing, and should therefore be handled in the parsers.py module.

Please show some examples of the data you are receiving where you see this "created" filed in the raw json, as I see none.

FL550:
@jozefKruszynski Thanks for tuning in.

Here is an extract of the data structure from the call to "https://api.tidal.com/v1/users/{user_id}/favorites/tracks": 
```json
{
  "limit": 10,
  "offset": 0,
  "totalNumberOfItems": 1403,
  "items": [
    {
      "created": "2023-05-11T05:56:46.760+0000",
      "item": {
        "id": 95476115,
        "title": "123",
```

jozefKruszynski:
> @jozefKruszynski Thanks for tuning in.
> 
> Here is an extract of the data structure:
> 
> ```json
> {
>   "limit": 10,
>   "offset": 0,
>   "totalNumberOfItems": 1403,
>   "items": [
>     {
>       "created": "2023-05-11T05:56:46.760+0000",
>       "item": {
>         "id": 95476115,
>         "title": "123",
> ```

Sorry, I just read your comment further up about the "created" field being one level up from the nested "item"
Apologies for that.

jozefKruszynski:
Can you take a look at the paginate function in the api_client module as we should be able to extract the necessary "created" key more cleanly there I suspect.

FL550:
Ok, I now have updated the code to do the field extraction in the api client and added the datetime parsing failure handling.

jozefKruszynski:
I gave it some more thought and realised that we can avoid further complication by simply not using the "nested_key" param.

This would then instruct the api_client to pass through the whole object, and then we parse out the "created" and "item" keys within the parsers themselves.

This feels like the most sensible solution, but I'm open to other opinions, sorry for messing you around @FL550

We paginate function already supports not passing the nested key param.

FL550:
That's also fine for me, however we then need to rename the existing *_obj variables in the functions and do something like that:
```python
artist_obj_data: dict[str, Any] = artist_obj["item"]
artist_id = str(artist_obj_data["id"])
```

And this has also be amended:
```python
async for item in self.api.paginate(path):
   if item and item.get("item") and item["item"].get("id"):
```

And the `nested_key` param and logic should then be removed?

jozefKruszynski:
> That's also fine for me, however we then need to rename the existing *_obj variables in the functions and do something like that:
> 
> ```python
> artist_obj_data: dict[str, Any] = artist_obj["item"]
> artist_id = str(artist_obj_data["id"])
> ```
> 
> And this has also be amended:
> 
> ```python
> async for item in self.api.paginate(path):
>    if item and item.get("item") and item["item"].get("id"):
> ```
> 
> And the `nested_key` param and logic should then be removed?

From a quick look we only use the paginate call in those library functions, and therefore removing the nested key logic makes sense as we no longer need it there.

Yes I think we should just pass the whole entry through to the parser, something like the following

```python
async def get_artists(self) -> AsyncGenerator[Artist, None]:
    """Retrieve library artists."""
    path = f"users/{self.auth.user_id}/favorites/artists"
    async for item in self.api.paginate(path): 
        yield parse_artist(self.provider, item)
```

Some care will be needed when handling the playlists and mixes because they are a little bit special, but I'm sure you already noticed this.

OzGav:
Also please also resolve all of the copilot suggestions by implementing or adding a comment as to why the suggestion is wrong.

FL550:
I've cleared the first commits with the disregarded solution and implemented your suggestions in a clean commit. 
The tests has also been updated, however 4 tests are still failing, due to a new test snapshot needed.

FL550:
@OzGav I'm not sure how to retrigger Copilot, because the old code references are now obsolete.

OzGav:
I think just comment and resolve the copilot suggestions. If it is no longer relevant just say that

jozefKruszynski:
> @OzGav I'm not sure how to retrigger Copilot, because the old code references are now obsolete.

In theory I just requested a second copilot review.
I'm away until Monday and will hopefully have a bit of time next week to work on the project in general.

FL550:
@jozefKruszynski Is there anything I can do for you to get this approved?

jozefKruszynski:
Hi sorry, I tested this out and everything seemed to work as expected.
From my perspective this is good to go.
I've re-requested a review from Marvin, and hopefully he will agree.

jozefKruszynski:
<strike>One question I have though, is this truly what we want here. I know that the date_added field was a relatively new addition onto the mediaItem objects, but I did not really see the discussion about it, and therefore I don't know what the expected outcome is here.

Sorry to throw a spanner in the works</strike>

@OzGav sent me some links. Thank you.

MarvinSchenkel:
Looks okay to me as well, but pytest thinks otherwise ;-). Could you have a look and address the test failures? After that, we can get this merged 👏

FL550:
The failing test are the test snapshot related ones. I haven't worked with this mechanism before, so I need help with this if I should fix this.

jozefKruszynski:
> The failing test are the test snapshot related ones. I haven't worked with this mechanism before, so I need help with this if I should fix this.

I'll pull your PR and have another look

jozefKruszynski:
Better to teach a man to fish and so on....

`python3 -m pytest tests/providers/tidal --snapshot-update`

The above command should update the snapshots accordingly @FL550

FL550:
> Better to teach a man to fish and so on....

Thanks for the hint. I updated the snapshot and have now something to read about snapshots :)

FL550:
@MarvinSchenkel The tests should now succeed.

MarvinSchenkel (approved):
Awesome work @FL550 , thanks 🎉

PR #2971, opened 2026-01-14
Make pre-commit configuration cross-platform compatible using uv

ALERTua:
This PR updates the pre-commit configuration with minimal changes to work consistently on both Linux and Windows, replacing shell script wrappers with direct `uv run` commands and ensuring proper line ending handling.
This PR is based on the digressions part of https://github.com/music-assistant/server/pull/2871

Music Assistant Server **can** be developed on Windows, or, should I say, on any OS. But there are caveats that can be easily fixed
  - The pre-commit config uses `scripts/run-in-env.sh`. This is unneeded due to `uv`. `uv run pre-commit run` executes the same calls but on any OS.
  - The pre-commit config uses `python3 -m ...` which should also be substituted by `uv run -m ...`
  - (Optional) Many pre-commit checks can be replaced with the "official" ones from [uv](https://github.com/astral-sh/uv-pre-commit), [ruff](https://github.com/astral-sh/ruff-pre-commit), [pre-commit](https://github.com/pre-commit/pre-commit-hooks), [mypy](https://github.com/pre-commit/mirrors-mypy).
  - In pre-commit fixes, the end of lines is taken from the system. It should be fixed as `LF` in `pyproject.toml` and `.gitattributes` to allow for development on any OS.
  - WSL2 is not needed to either develop or test the project. Everything I tested works natively on Windows.

__Changes:__

1. __`.gitattributes`__ (new file):

   - Added `* text=auto eol=lf` to enforce LF line endings for all text files across platforms to ensure LF endings on the git client's side
   - This file does not prevent `requirements_all.txt` from showing as modified on Windows due to CRLF conversion, as `gen_requirements_all` is responsible for the endings

2. __`.pre-commit-config.yaml`__:

   - Replaced `scripts/run-in-env.sh <command>` with `uv run <command>` for all hooks to get rid of Linux Shell usage and for `uv` usage consistency.
   - Changed hook execution from `language: script` to `language: system` for `mypy` and `gen_requirements_all`
   - Reordered `gen_requirements_all` hook to run first as it modifies files.

3. __`scripts/gen_requirements_all.py`__:

   - Added explicit `newline="\n"` parameter to `Path.write_text()` to ensure LF line endings regardless of OS

__Benefits:__

- Pre-commit hooks work on Windows
- Pre-commit hooks work identically on Linux and Windows
- No more spurious diffs from line ending differences on Windows
- Simpler configuration using `uv` directly without wrapper scripts

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Force LF line endings for all text files
+* text=auto eol=lf
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,102 +1,119 @@
 repos:
   - repo: local
     hooks:
+      # gen_requirements_all modifies files so it has to be first
+      - id: gen_requirements_all
+        name: gen_requirements_all
+        entry: uv run -m scripts.gen_requirements_all
+        pass_filenames: false
+        language: system
+        types: [text]
+        files: ^(music_assistant/.+/manifest\.json|pyproject\.toml|\.pre-commit-config\.yaml|scripts/gen_requirements_all\.py)$
+
       - id: ruff-check
         name: 🐶 Ruff Linter
         language: system
         types: [python]
-        entry: scripts/run-in-env.sh ruff check --fix
+        entry: uv run ruff check --fix
         require_serial: true
         stages: [pre-commit, pre-push, manual]
+
       - id: ruff-format
         name: 🐶 Ruff Formatter
         language: system
         types: [python]
-        entry: scripts/run-in-env.sh ruff format
+        entry: uv run ruff format
         require_serial: true
         stages: [pre-commit, pre-push, manual]
+
       - id: check-ast
         name: 🐍 Check Python AST
         language: system
         types: [python]
-        entry: scripts/run-in-env.sh check-ast
+        entry: uv run check-ast
+
       - id: check-case-conflict
         name: 🔠 Check for case conflicts
         language: system
-        entry: scripts/run-in-env.sh check-case-conflict
+        entry: uv run check-case-conflict
+
       - id: check-docstring-first
         name: ℹ️  Check docstring is first
         language: system
         types: [python]
-        entry: scripts/run-in-env.sh check-docstring-first
+        entry: uv run check-docstring-first
+
       - id: check-executables-have-shebangs
         name: 🧐 Check that executables have shebangs
         language: system
         types: [text, executable]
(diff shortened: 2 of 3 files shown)


marcelveldt:
Does this still run the pre-commit actions and scripts in a venv ?
Then it is fine.

marcelveldt:
BTW: We won't be able to support developing on Windows. Most likely large parts of the code will not be Windows friendly at all. For python development, better use macos or linux.

ALERTua:
> Does this still run the pre-commit actions and scripts in a venv ? Then it is fine.

`uv run` always runs in venv, and even sets up the venv if there is none.


> BTW: We won't be able to support developing on Windows. Most likely large parts of the code will not be Windows friendly at all. For python development, better use macos or linux.

I have been developing the [Bandcamp PR ](https://github.com/music-assistant/server/pull/2871), running tests, running the server natively, and debugging the server natively on Windows with zero problems. All I changed is the `pre-commit-config.yaml` from this PR. You don't need to support developing Music Assistant on Windows, as there is nothing to support after this PR. All you need is to get rid of the redundant bash scripts in favor of `uv run`, as this PR introduces :) At least go ahead and check out this PR locally and run `uv run pre-commit run -a` and you won't notice anything changed.

MarvinSchenkel:
There is one caveat with the new approach. The current `run-in-env.sh` script automatically detects the repo path (using `git rev-parse --show-toplevel`) and automatically activates the venv before running those commands. This ensures that for example tools such as Github desktop also run those pre-commit commands in the appropriate venv. 

Is there a way we can preserve this functionality with the `uv` approach?

ALERTua:
> There is one caveat with the new approach. The current `run-in-env.sh` script automatically detects the repo path (using `git rev-parse --show-toplevel`) and automatically activates the venv before running those commands. This ensures that for example tools such as Github desktop also run those pre-commit commands in the appropriate venv.

- I am using both [Fork](https://git-fork.com/) and [SourceGit](https://github.com/sourcegit-scm/sourcegit) to manage my Git repositories
- I always install my Git pre-commit hooks using `uv run pre-commit install`
- I have encountered zero issues with the `pre-commit-config.yaml` and the described tools


> Is there a way we can preserve this functionality with the `uv` approach?

`uv run`, launched at the repository root, always runs with an activated venv. you don't need these bash scripts to ensure this behavior.

Please try it yourself.

MarvinSchenkel (approved):
Just had a play and seems to work perfectly. Did not know that `uv run` automatically runs in a venv.. Nice!

PR #2973, opened 2026-01-14
Do not report progress on ffmpeg streaming error 

fmunkes:
Related to music-assistant/support#4754

This should prevent progress reporting if ffmpeg reports an error while streaming. What the actual problem in music-assistant/support#4754 is, is still not clear to me.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -2484,6 +2484,13 @@ class PlayerQueuesController(CoreController):
             return
         assert media_item.uri is not None  # uri is set in __post_init__
 
+        if item_to_report.streamdetails and item_to_report.streamdetails.stream_error:
+            self.logger.debug(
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/player_queues.py
```diff
@@ -2486,6 +2484,13 @@ def _handle_playback_progress_report(
             return
         assert media_item.uri is not None  # uri is set in __post_init__
 
+        if item_to_report.streamdetails and item_to_report.streamdetails.stream_error:
+            self.logger.debug(
+                "There was an error while streaming %s. Playback progress is not processed",
+                item_to_report.name,
+            )
```
```suggestion
            #  Ignore items that had a stream error
```

marcelveldt: No need to extra log this - it had already been logged in the streams controller.

marcelveldt (approved):

PR #2977, opened 2026-01-15
Add 24 and 32-bit audio support for Sendspin

maximmaxim345:
Enables true 24 and 32-bit audio support for Sendspin players (with both PCM and FLAC codecs).

The main stream now uses 32-bit PCM internally, with aiosendspin handling per-player conversion.

Tested with 15 Sendspin clients (with many different formats) on a Home Assistant Green with only 50% (of 400%) CPU usage during playback.


music_assistant/providers/sendspin/player.py
```diff
@@ -147,15 +147,25 @@ async def player_channel(
             # DSP is disabled and output is stereo, use main_stream
             return None
 
+        # Use mono for non-stereo output channels, it will upmix in aiosendspin in case
+        # only stereo is supported by the player
+        output_format = self.output_format
+        if output_format.channels != 1:
```
marcelveldt: shouldn't this be != 2 ?

MarvinSchenkel (approved):
🎉

PR #2981, opened 2026-01-16
Add SomaFM Radio provider

macegr:
image

### Implements SomaFM Radio provider
- Allows configuration of stream quality
- Fetches station data and thumbnails from SomaFM
- Parses playlist URL on stream start, selects random server with failover if one is unavailable

No known issues.

_Note that SomaFM is a free service **without** ads, unlike most of the paid/free streaming options currently available. For over 25 years, they have relied upon listener donations and store purchases to keep running. If you use SomaFM, please consider donating. I don't have any affiliation with SomaFM Radio other than a long-time listener and occasional donor._

- [x] TODO: Create provider documentation and file PR at https://github.com/music-assistant/beta.music-assistant.io


music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+            ],
+            default_value="high",
+        ),
+        ConfigEntry(
+            key=CONF_FORMAT,
+            type=ConfigEntryType.STRING,
+            label="Format",
+            description="High and Low quality streams are only available in AAC",
+            depends_on=CONF_QUALITY,
+            depends_on_value="highest",
```
MarvinSchenkel: Why does this depend on the other one?

music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    # ruff: noqa: ARG001
+    return (
+        ConfigEntry(
+            key=CONF_QUALITY,
+            type=ConfigEntryType.STRING,
+            label="Stream Quality",
+            description="Up to 320kbps in Highest",
+            options=[
```
MarvinSchenkel, line 69: I would pick 'highest' by default?

music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+
+async def get_config_entries(
+    mass: MusicAssistant,
+    instance_id: str | None = None,
+    action: str | None = None,
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    # ruff: noqa: ARG001
+    return (
```
MarvinSchenkel, line 63: I think both ConfigEntries can be moved to the 'advanced' category when we pick sane defaults. You can do this by adding `category="advanced"`

music_assistant/providers/somafm/__init__.py
macegr: I was trying to get the config options interdependent. For example, if you don’t select Highest you can’t select MP3 anymore. I didn’t find a good way to do that. Do you think we should just pick one since they all work and remove the format config?

music_assistant/providers/somafm/__init__.py
macegr: Some of them are 320kbps and SomaFM is donation supported, I didn’t feel right automatically using more of their bandwidth. The High profile still sounds pretty good?

music_assistant/providers/somafm/__init__.py
macegr: Yes with an agreed on default this is a good plan

macegr:
Thanks for the notes, will make some updates tomorrow

music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+            raise MediaNotFoundError("No valid SomaFM playlist available")
+
+        async def _get_stream_path(item_id: str) -> str:
+            """Pick correct playlist, fetch the playlist, and extract stream URL."""
+            stations = await self._get_stations()
+            station = stations.get(item_id)
+            if station:
+                playlist_url = _get_playlist_url(station)
+                playlist = await fetch_playlist(self.mass, playlist_url)
+                playlist_item: PlaylistItem = await _get_valid_playlist_item(playlist)
```
MarvinSchenkel, line 221: Why do we need to explicitly validate these playlist items? Is it common for Soma to return invalid urls for a certain station?

music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+            playlists: list[dict[str, str]] = station.get("playlists", [])
+            for playlist in playlists:
+                avail_quality = playlist.get("quality")
+                avail_format = playlist.get("format")
+                playlist_url = playlist.get("url")
+                if req_quality == avail_quality and req_format == avail_format and playlist_url:
+                    return playlist_url
+            # Nothing matched, just return the first (probably highest quality) item
+            self.logger.warning("Couldn't find SomaFM stream with requested quality and format")
+            playlist_url = playlists[0].get("url")
```
MarvinSchenkel, line 209: Would be nice if we sort this by quality first, so that `playlist[0]` returns the highest _available_ quality stream in case the requested quality is not found.

music_assistant/providers/somafm/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+            raise MediaNotFoundError("No valid SomaFM stream available")
+
+        def _get_playlist_url(station: dict[str, Any]) -> str:
+            """Pick playlist based on quality and format config values."""
+            req_quality = self.config.get_value(CONF_QUALITY)
+            req_format = self.config.get_value(CONF_FORMAT)
+            if req_quality in ["high", "low"]:
+                req_format = "aacp"
+                self.logger.info("SomaFM High and low quality streams are available only in AACP")
+            playlists: list[dict[str, str]] = station.get("playlists", [])
```
MarvinSchenkel, line 184: Let's add a guard after this:

```
if len(playlists) == 0:
  raise MediaNotFoundError("No valid SomaFM playlist available")
```

MarvinSchenkel:
Congrats on your first contribution , very well done 🎉 . Few small comments and then this is good to go!

music_assistant/providers/somafm/__init__.py
macegr: Done! Removed dependent config entry

music_assistant/providers/somafm/__init__.py
macegr: Done! Removed second config entry and moved quality config to Advanced section.

music_assistant/providers/somafm/__init__.py
macegr: Each SomaFM station offers PLS or M3U files, and the playlist contains several possible servers for the stream. Players will automatically fail over to the next good stream if one goes down or the stream server changes. Does this actually happen? I can only go from the message about this on SomaFM's website, which led me to implement this shuffle-and-check method to ensure we have a valid stream:

> Using the PLS or M3U links is the best, as these links are "permanent" and do not change. Direct server links can vary based on individual server availability and should only be used if PLS or M3U links won't work with your player. If a direct server link stops working, you'll need to get the updated links here.

macegr: It would be possible for me to NOT implement this method in my provider, if Radio paths could be playlists as well as direct streams, with the failover mechanism implemented deeper in the project. But perhaps SomaFM is the only one who does it this way.

music_assistant/providers/somafm/__init__.py
macegr: Done! Sorted the list with a key mapping and removed non AAC options.

music_assistant/providers/somafm/__init__.py
macegr: Done!

music_assistant/providers/somafm/__init__.py
macegr: Also, I have locked the format to AAC, but defaulted to Highest quality at the same time per your other comment. Since AAC is generally more efficient than high bitrate MP3, we are still saving bandwidth while offering good quality audio.

SomaFM is experimenting with HLS, so someday we may have another stream type to implement.

music_assistant/providers/somafm/__init__.py
macegr: @MarvinSchenkel Do you need more here?

MarvinSchenkel (approved):
Aaand another great addition to the music providers. Thanks for this contribution 👏

macegr:
> Aaand another great addition to the music providers. Thanks for this contribution 👏

Thanks Marvin, appreciate the detailed review and feedback!

PR #2982, opened 2026-01-16
fix(alexa): Fix issue with language on alexa skills for french and germany

vlacour97:
Location was stuck on en-US. However, for non-English users, the Alexa provider didn't work.
This PR updates the location settings using the configuration provided by the MA provider and implements the French Alexa command (my native language).

### DIFF
--- a/music_assistant/providers/alexa/__init__.py
+++ b/music_assistant/providers/alexa/__init__.py
@@ -51,6 +51,7 @@ CONF_API_URL = "api_url"
 CONF_ALEXA_LANGUAGE = "alexa_language"
 
 ALEXA_LANGUAGE_COMMANDS = {
+    "play_audio_fr-FR": "music assistant",
     "play_audio_de-DE": "sag music assistant spiele audio",
     "play_audio_en-US": "ask music assistant to play audio",
     "play_audio_default": "ask music assistant to play audio",
@@ -425,6 +426,8 @@ class AlexaProvider(PlayerProvider):
         if devices is None:
             return
 
+        alexa_locale = str(self.config.get_value(CONF_ALEXA_LANGUAGE, "en-US"))
+
         for device in devices:
             if device.get("capabilities") and "MUSIC_SKILL" in device.get("capabilities"):
                 dev_name = device["accountName"]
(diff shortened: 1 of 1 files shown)


music_assistant/providers/alexa/__init__.py
```diff
@@ -425,6 +426,8 @@ async def loaded_in_mass(self) -> None:
         if devices is None:
             return
 
+        alexa_locale = self.config.get_value(CONF_ALEXA_LANGUAGE) or 'en-US'
```
MarvinSchenkel:
```suggestion
        alexa_locale = self.config.get_value(CONF_ALEXA_LANGUAGE, "en-US")
```

music_assistant/providers/alexa/__init__.py
```diff
@@ -51,6 +51,7 @@
 CONF_ALEXA_LANGUAGE = "alexa_language"
 
 ALEXA_LANGUAGE_COMMANDS = {
+    "play_audio_fr-FR": "music assistant",
```
MarvinSchenkel, line 54: Shouldn't this be the French equivalent of 'ask music assistant to play audio', something like 'demander à l'assistant musical de lire l'audio'?

vlacour97: I tried using that type of phrase on my setup and it didn't work. Sending command with only "music assistant" sentence worked.
It might be interesting in the future to be able to customize this command in the provider's advanced options (in UI), but I don't yet have the necessary knowledge about the project for doing it

alams154: I've seen this behavior too a while back.  The US version of Alexa used to work without specifying the "play" intent but updates to Alexa in the US broke this.  May or may not be related to Alexa+.

alams154:
> Location was stuck on en-US. However, for non-English users, the Alexa provider didn't work. This PR updates the location settings using the configuration provided by the MA provider and implements the French Alexa command (my native language).

Thanks for finding the bug!

MarvinSchenkel:
> > Location was stuck on en-US. However, for non-English users, the Alexa provider didn't work. This PR updates the location settings using the configuration provided by the MA provider and implements the French Alexa command (my native language).
> 
> Thanks for finding the bug!

I am not an alexa user myself, are you able to verify this fix by any chance and confirm whether the change to the command is also required for the other languages?

alams154:
> elf, are you able to verify this fix by any chance and confirm whether the change to the command is also required for the other languages?

I can only confirm for en-US and the command needs to stay the same for that

OzGav:
@vlacour97 there is a mypy error to be fixed

vlacour97:
> @vlacour97 there is a mypy error to be fixed

@OzGav 
The mypy error is now fixed! :)

MarvinSchenkel (approved):
Thanks @vlacour97 !

PR #2986, opened 2026-01-16
Add HEOS Player provider

Tommatheussen:
Added initial native support for Denon HEOS players. There is still quite some work to be done, but I've had good progress so far.
Tested locally with 3 devices (2 Heos amps, and 1 AVR).

Features included to far:
- Basic controls (volume, next/previous, play/pause)
- Native grouping, handling both groups being made/changed by MA, as well as groups being made/changed by the native HEOS app
- Basic playback, there might be some more work to be done on this one
- Device change events from HEOS are handled to create/unregister devices in MA
- MA shows metadata when non-MA content is playing

Things to be done:
- mdns discovery, will do this in a future PR, more work involved than expected


OzGav:
A lot of people will be excited to see this as it highly requested.

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,192 @@
...
+        else:
+            self._attr_group_members.clear()
+
+        self.update_state()
+
+    async def _player_event_received(self, event: str) -> None:
+        """Handle player device events."""
+        await self.set_dynamic_attributes()
+
+    async def set_dynamic_attributes(self) -> None:
```
marcelveldt: this can be a regular function, no need for a coroutine here

music_assistant/providers/heos/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+) -> tuple[ConfigEntry, ...]:
+    """
+    Return Config entries to setup this provider.
+
+    instance_id: id of an existing provider instance (None if new instance setup).
+    action: [optional] action key called from config entries UI.
+    values: the (intermediate) raw values for config entries sent with the action.
+    """
+    # ruff: noqa: ARG001
+    # Config Entries are used to configure the Player Provider if needed.
```
OzGav: Can you start to tidy this up by removing all these placeholder comments and add your own as required.

music_assistant/providers/heos/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    # The ConfigEntry is a dataclass that represents a single configuration entry.
+    # The ConfigValueType is an Enum that represents the type of value that
+    # can be stored in a ConfigEntry.
+    # If your provider does not need any configuration, you can return an empty tuple.
+    return (
+        # example of a ConfigEntry for the number of players to create
+        ConfigEntry(
+            key=CONF_IP_ADDRESS,
+            type=ConfigEntryType.STRING,
+            label="Host name or IP address.",
```
OzGav: Looking at the description should this be more direct? “Main Controller host name or IP address.”?

music_assistant/providers/heos/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+  "stage": "alpha",
+  "name": "Denon HEOS",
+  "description": "HEOS Player provider for Denon/Marantz devices with HEOS Built-in",
+  "codeowners": [
+    "@Tommatheussen"
+  ],
+  "requirements": [
+    "pyheos==1.0.6"
+  ],
+  "documentation": "https://music-assistant.io/player-support/demo/",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/player-support/heos/",
```

music_assistant/providers/heos/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+  "type": "player",
+  "domain": "heos",
+  "stage": "alpha",
+  "name": "Denon HEOS",
+  "description": "HEOS Player provider for Denon/Marantz devices with HEOS Built-in",
```
OzGav:
```suggestion
  "description": "Play to Denon & Marantz devices via HEOS.",
```
This is used in the front end so the suggestion is action oriented and clear to the user what this provider is for

music_assistant/providers/heos/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+  "type": "player",
+  "domain": "heos",
+  "stage": "alpha",
```
OzGav:
```suggestion
  "stage": "beta",
```

I’m assuming it will be somewhat stable and feature complete?

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,270 @@
...
+        self.update_state()
+
+    def _update_player_volume(self) -> None:
+        """Update volume properties."""
+        self._attr_volume_level = self._device.volume
+        self._attr_volume_muted = self._device.is_muted
+
+    def _update_player_state(self) -> None:
+        """Update playback state."""
+        self._attr_playback_state = HEOS_PLAY_STATE_TO_PLAYBACK_STATE[self._device.state]
```
OzGav: I think that if pyheos returns an unexpected state/type not in the mappings, this will raise KeyError. Consider using .get() with a fallback.

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,270 @@
...
+        # We can only know we're playing from MA if we started this session.
+        if (now_playing.source_id != const.MUSIC_SOURCE_LOCAL_MUSIC) or (
+            self._attr_active_source != self.player_id
+        ):
+            self.logger.debug("[%s] Now playing change: %s", self._device.name, now_playing)
+
+            self._attr_active_source = str(now_playing.source_id)
+            self._attr_current_media = PlayerMedia(
+                uri=now_playing.media_id or media_uri_from_now_playing_media(now_playing),
+                media_type=HEOS_MEDIA_TYPE_TO_MEDIA_TYPE[now_playing.type],
```
OzGav: Same comment as above

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,270 @@
...
+    _device: pyheosPlayer
+
+    def __init__(self, provider: HeosPlayerProvider, client: pyheosPlayer) -> None:
+        """Initialize the Player."""
+        super().__init__(provider, str(client.player_id))
+
+        self._device: pyheosPlayer = client
+
+        # Keep internal reference so we don't need to check None on each call
+        assert self._device.heos
```
OzGav: Is this the best thing to do here vice a check and error message?

OzGav:
Generally looking pretty good to me. You could consider reviewing MusicCast's approach of marking players unavailable on connection errors.

Tommatheussen:
> Generally looking pretty good to me. You could consider reviewing MusicCast's approach of marking players unavailable on connection errors.

Thanks!
I'm going to attempt to finish it before the weekend, skipping mDNS for now and adding that into a new PR, that's a bit more work than I thought (also first time doing mDNS stuff)

music_assistant/providers/heos/player.py
Tommatheussen: Theoretically it shouldn't be possible for the `heos` property to be None (at least not in the code path I've found in the library), but adding an error message might be more future proof as a sanity check 👍

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,303 @@
...
+        self,
+        action: str | None = None,
+        values: dict[str, ConfigValueType] | None = None,
+    ) -> list[ConfigEntry]:
+        """Return all (provider/player specific) Config Entries for the player."""
+        return [
+            *await super().get_config_entries(action=action, values=values),
+            create_sample_rates_config_entry(
+                max_sample_rate=192000,
+                safe_max_sample_rate=192000,
```
Tommatheussen: I set the max rate and bit depth for FLAC and WAV here, mp3 and aac don't support that high, but I think MA outputs those at 48kHz/16 bit anyway? Is this fine or should these be reduced to what mp3 and aac support?

I've tested with selecting the max rate and depth on my devices and both mp3 and aac worked fine, MA did report lower rates being played.

music_assistant/providers/heos/provider.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        self._heos.add_on_user_credentials_invalid(self.invalid_credentials)
+
+        try:
+            await self._heos.connect()
+
+            self._heos.add_on_controller_event(self._handle_controller_event)
+        except HeosError as e:
+            self.logger.error(f"Failed to connect to HEOS controller: {e}")
+            raise MusicAssistantError("Failed to connect to HEOS controller") from e
```
Tommatheussen: Not sure if `MusicAssistantError` is fine here, or if I should raise `SetupFailedError`?

music_assistant/providers/heos/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+            description="Hostname or IP address of the HEOS device "
+            "to be used as the main controller. It is recommended to use a "
+            "wired device as the main controller.",
+        ),
+        ConfigEntry(
+            key=CONF_USERNAME,
+            type=ConfigEntryType.STRING,
+            label="Username",
+            required=False,
+            description="The username to authenticate to the HEOS system.",
```
MarvinSchenkel: Maybe add a `ConfigEntryType.LABEL` here explaining that the credentials are optional and the additional features they get when they do provider it.

music_assistant/providers/heos/player.py
```diff
@@ -0,0 +1,303 @@
...
+
+        self._device: PyHeosPlayer = device
+
+        if self._device.heos is None:
+            raise SetupFailedError("HEOS device has no controller assigned")
+
+        # Keep internal reference so we don't need to check None on each call
+        self._heos = self._device.heos
+
+        model_parts = device.model.split(maxsplit=1)
```
MarvinSchenkel: I would move the attribute assignments to the `setup()` function.

music_assistant/providers/heos/player.py
MarvinSchenkel: This is just what the player supports, you don't have to worry about codecs here. 

Do all HEOS devices support 192kHz/24? Otherwise you might want to lower the safe_xxx values here to maybe 48kHz/16 ?

music_assistant/providers/heos/provider.py
MarvinSchenkel: I think `SetupFailedError` is more appropriate here. `MusicAssistantError` is more of a generic error.

MarvinSchenkel:
Just a few minor things, but overall, this already looks really good. Nice job!

As for the mdns, what makes you think this is a lot of work? I see you have already identified the mdns key in the manifest, so it should just be implementing `on_mdns_service_state_change` in the `provider.py` and basing your setup off of those ip addresses. Have a look at the airplay and bluesound providers, they already have these implemented.

Tommatheussen:
> Just a few minor things, but overall, this already looks really good. Nice job!
> 
> As for the mdns, what makes you think this is a lot of work? I see you have already identified the mdns key in the manifest, so it should just be implementing `on_mdns_service_state_change` in the `provider.py` and basing your setup off of those ip addresses. Have a look at the airplay and bluesound providers, they already have these implemented.

Thanks!

So the main reason is that HEOS works with a single controller for all devices. mDNS finds all devices though, but we don't need all devices to be controllers. So instead of a simple mDNS device found -> MA player creation, we have to run through some kind of device host checking.
For all found devices we need to loop through this (inspired by Home Assistant code):
- Check if device is already in the system from the existing controller (skip if found)
- Connect to device, get system info (including players it found in the network)
- Disconnect from device
- System info reports version, connection method and players
- Based on this info (wired + latest version), the library builds a map of 'preferred hosts'.
- Pick 1 to connect to, store this in the integration information
- Provider connects to stored device, loads players it can control (should be all)

Don't get me wrong, it's not that hard to get done, I just need to put some time into this, but it might be good to get this basic in the beta already ot start getting testing ;) 
The HEOS system does report when devices change on the network, so those are handled after the initial setup was done

MarvinSchenkel:
> > Just a few minor things, but overall, this already looks really good. Nice job!
> > As for the mdns, what makes you think this is a lot of work? I see you have already identified the mdns key in the manifest, so it should just be implementing `on_mdns_service_state_change` in the `provider.py` and basing your setup off of those ip addresses. Have a look at the airplay and bluesound providers, they already have these implemented.
> 
> Thanks!
> 
> So the main reason is that HEOS works with a single controller for all devices. mDNS finds all devices though, but we don't need all devices to be controllers. So instead of a simple mDNS device found -> MA player creation, we have to run through some kind of device host checking. For all found devices we need to loop through this (inspired by Home Assistant code):
> 
> * Check if device is already in the system from the existing controller (skip if found)
> * Connect to device, get system info (including players it found in the network)
> * Disconnect from device
> * System info reports version, connection method and players
> * Based on this info (wired + latest version), the library builds a map of 'preferred hosts'.
> * Pick 1 to connect to, store this in the integration information
> * Provider connects to stored device, loads players it can control (should be all)
> 
> Don't get me wrong, it's not that hard to get done, I just need to put some time into this, but it might be good to get this basic in the beta already ot start getting testing ;) The HEOS system does report when devices change on the network, so those are handled after the initial setup was done

Ah gotcha, totally fine to do this in a follow up PR then. Will be nice to get people to start playing with it and get you some feedback :-)

music_assistant/providers/heos/player.py
Tommatheussen: I think they do, at least from the device specs I've come accross, but I will scour the internet to find specs on as many devices I can find, I know there are 1st gen devices that are quite old, they might not support it.

music_assistant/providers/heos/__init__.py
Tommatheussen: Will add the label, currently the credentials aren't used for anything yet (unless the library does something specific with it, but I don't think so). They are used to retrieve configured music services, playlists and favorites.
I have to check if it's possible to map these to MA somehow.
Should we keep the config options here actually while they are not used yet?

music_assistant/providers/heos/player.py
Tommatheussen: So the older Gen 1 devices don't support the higher sample rate/bit depth indeed, thanks for making me check!
I also looked at possibly making it dependent on the model, but the returned data does not make a distinction between generations so I just set it to a lower safer max for now.

music_assistant/providers/heos/__init__.py
MarvinSchenkel: Those would mostly be applicable to a `Music`Provider instead of a `Player`Provider I think. In that case I would just leave them out and add them once you've figured out a use for them..

I wouldn't worry about trying to import music services. If people want to add them, they should use the native music providers in MA.

Tommatheussen: Removed them for now, will think about it whether we need them later or not 👍

MarvinSchenkel:
I was thinking about this approach a bit more. Do we really need a controller IP for this provider? Or would we be fine by just detecting the individual players via mdns and connecting to them?

Or does the pyheos package only allow connecting to a controller vs individual players?

Tommatheussen:
> I was thinking about this approach a bit more. Do we really need a controller IP for this provider? Or would we be fine by just detecting the individual players via mdns and connecting to them?
> 
> Or does the pyheos package only allow connecting to a controller vs individual players?

Technically it's possible to connect to each player individually, but the pyheos library indeed only allows 1 connection to be set. I checked the HEOS connection specifications and they also state it's recommended to connect to a single speaker that will delegate the rest, to reduce network traffic.

MarvinSchenkel:
> > I was thinking about this approach a bit more. Do we really need a controller IP for this provider? Or would we be fine by just detecting the individual players via mdns and connecting to them?
> > Or does the pyheos package only allow connecting to a controller vs individual players?
> 
> Technically it's possible to connect to each player individually, but the pyheos library indeed only allows 1 connection to be set. I checked the HEOS connection specifications and they also state it's recommended to connect to a single speaker that will delegate the rest, to reduce network traffic.

Interesting approach. In that case, I think this is good to be merged. Just checking, anything you still want to add / change before we merge this?

Tommatheussen:
Nope, I think everything is in there that I initially planned 👍

MarvinSchenkel (approved):
Amazing stuff @Tommatheussen , thanks 🎉 .

sakowicz:
Nice! 
@MarvinSchenkel Can you do a nightly release with these changes so we can test?

OzGav:
That happens automatically

sakowicz:
Yeah - every 24h so we need to wait until tomorrow. 

It might be run manually too.

MarvinSchenkel:
Planning to do a new beta somewhere today, bit more patience ;-)

PR #2992, opened 2026-01-19
Use instance id in logger to distinguish log messages from multi instance providers

MarvinSchenkel:
A followup to [this](https://github.com/music-assistant/server/pull/2943) PR to distinguish multi instance provider. We cannot use `self.name` here, because this indirectly uses `instance_name_postfix` which in some cases, uses config values that are not available yet at that time.


MarvinSchenkel:
Another way to go about this is to just _always_ include the instance_id for logging. Just curious about your opinions here.

PR #2999, opened 2026-01-20
SyncGroupPlayer: Inherit MULTI_DEVICE_DSP feature from group members

gnumpi:
1. Add `PlayerFeature.MULTI_DEVICE_DSP` to `OPTIONAL_FEATURES` of `SyncGroupPlayer` so the feature is inherited from its group member players.
2. Allow reading supported features of a `SyncGroupPlayer` even when it is powered off (and therefore has no group leader yet) by inheriting supported features from the first player in its `group_members` list.

### DIFF
--- a/music_assistant/controllers/players/sync_groups.py
+++ b/music_assistant/controllers/players/sync_groups.py
@@ -113,14 +113,14 @@ class SyncGroupPlayer(GroupPlayer):
     def supported_features(self) -> set[PlayerFeature]:
         """Return the supported features of the player."""
         members = self.group_members
-        p: Player | None = self.sync_leader or (
+        reference_player: Player | None = self.sync_leader or (
             self.mass.players.get(members[0]) if members else None
         )
-        if p:
+        if reference_player:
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/players/sync_groups.py
```diff
@@ -111,11 +112,15 @@ async def on_config_updated(self) -> None:
     @property
     def supported_features(self) -> set[PlayerFeature]:
         """Return the supported features of the player."""
-        if self.sync_leader:
+        members = self.group_members
+        p: Player | None = self.sync_leader or (
```
marcelveldt: can we use a different variable name here ?
like "reference_player" or "base_player"

gnumpi: sure, will do...

gnumpi: done

MarvinSchenkel (approved):
Thanks @gnumpi , LGTM

PR #3004, opened 2026-01-21
Add AAC/M4A support to tag parsing

OzGav:
Problem reported here https://github.com/orgs/music-assistant/discussions/4823#discussioncomment-15552255 

The parse_tags_mutagen function previously only handled ID3 tags (MP3 files), ignoring MP4/M4A/AAC files which use a different tag format.

This commit adds comprehensive MP4 tag support including:

Basic tags:
- title (©nam), artist (©ART), albumartist (aART), album (©alb)
- genre (©gen), date (©day), lyrics (©lyr)
- track/disc numbers with totals (trkn, disk)
- compilation flag (cpil)

Sort tags:
- titlesort (sonm), albumsort (soal)
- artistsort (soar), albumartistsort (soaa)

MusicBrainz IDs (iTunes freeform tags):
- Album ID, Artist ID, Album Artist ID
- Release Group ID, Track ID, Recording ID
- Album Type

Additional metadata:
- ARTISTS (multi-value), BARCODE, ISRC
- ReplayGain track and album gain

The implementation:
- Detects MP4 files using isinstance(tags, MP4Tags)
- Handles MP4 freeform tags (----:com.apple.iTunes:*) with proper byte decoding for MusicBrainz and ReplayGain tags
- Maintains consistent value formats with ID3 handling (str vs list)
- Refactors ID3 parsing into separate _parse_id3_tags function

This also benefits ALAC and M4B files which use the same MP4 container.


marcelveldt:
This is an essential core functionality - please update the tests we have for parsing tags

OzGav:
Tests added.

I had to add some type ignores in the tests because AudioTags.tags is typed as dict[str, str] but some fields (like artistsort, albumartistsort) actually store lists to match mutagen’s ID3 behavior.
The ID3 test doesn’t have this issue because it doesn’t test sort tags - the MP3 fixture doesn’t have TSOP/TSO2 (artistsort/albumartistsort) tags, so the type conflict never surfaces there.
I believe in practice this should work fine because:
	∙	Properties that access these list fields use split_items() which already handles both strings and lists
	∙	Direct string operations (.split(), .lower()) are only used on fields that genuinely are strings (like date, track, compilation)
The type hint is inaccurate but the runtime behavior is correct. The ignores are only needed in tests where we explicitly assert the list values.

marcelveldt (approved):
Looks good, @OzGav !

PR #3006, opened 2026-01-21
fix(Tidal): Remove unnecessary deduping of recomendation rows

jozefKruszynski:
As we have proper multi user support along with multi account support, I would like to remove this deduping behaviour, as it is arguably not needed. It is also possibly the cause of some recent home screen issues for tidal users


MarvinSchenkel (approved):
Thanks @jozefKruszynski , fingers crossed it solves those few issues 🤞

PR #3008, opened 2026-01-22
Implement Apple Music playlist browsing with folders

MizterB:
Initial support for Apple Music playlist folders, accessible via the Browse view.

Partially addresses https://github.com/orgs/music-assistant/discussions/3039

### DIFF
--- a/music_assistant/providers/apple_music/__init__.py
+++ b/music_assistant/providers/apple_music/__init__.py
@@ -72,6 +72,7 @@ from music_assistant.helpers.playlists import fetch_playlist
 from music_assistant.helpers.throttle_retry import ThrottlerManager, throttle_with_retries
 from music_assistant.helpers.util import infer_album_type, parse_title_and_version
 from music_assistant.models.music_provider import MusicProvider
+from music_assistant.providers.apple_music.helpers import browse_playlists
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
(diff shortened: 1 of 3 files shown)


MarvinSchenkel:
Looks good to me! For structure and clarity, could you move all the browse related private functions into a file in a helpers folder? Something like `apple_music/helpers/browse.py`? The only function that should then we added to `__init__.py` is the `browse()` function itself.

MarvinSchenkel:

music_assistant/providers/apple_music/__init__.py
```diff
@@ -474,12 +486,19 @@ async def get_track(self, prov_track_id) -> Track:
     @use_cache()
     async def get_playlist(self, prov_playlist_id, is_favourite: bool = False) -> Playlist:
         """Get full playlist details by id."""
-        if not self.is_library_id(prov_playlist_id):
-            endpoint = f"catalog/{self._storefront}/playlists/{prov_playlist_id}"
-        else:
+        response: dict[str, Any] | None = None
```
MarvinSchenkel: What is it exactly that you are trying to do here? I see the typo with the extra `endpoint = f"catalog/{self._storefront}/playlists/{prov_playlist_id}"` (effectively always using the catalog playlist), but shouldn't simply removing that line be the fix here?

MizterB: I'm reverting that back (and removing the extra line).  I ran into some 404 errors when trying to pull up certain playlists, due to mismatched catalog vs. library IDs, and put this in as a fallback while troubleshooting.  It's no longer needed, and I've verified that the older code works fine (after removing the extra line).

MarvinSchenkel (approved):
Thanks @MizterB , looks good. Great addition :)

PR #3015, opened 2026-01-24
hass_players: Show now playing info from external playback

ulilicht:
This PR adds "Now Playing" information if playback is triggered from an external source in the hass_player provider. In addition, it allows prev/next handling for external sources. 

**Limitation:** 
For the functionality to work, Music Assistant needs to be able to detect if playback on the device is externally or internally. The best way I have found is to use `media_content_id` which is set by the HomeAssistant integration used for playback. If this is not set, the changes won't have an effect. 

Please let me know if there is a better way of detecting if playback is external. 

**AI Disclaimer:** 
The changes have been created assisted by AI. I have read the code and it looks fine to me. I have been testing the changes for a while on my devices. Manual testing has been done with Homeassistant 2026.1 and the Teufel Raumfeld Raumkernel integration as well the Chromecast integration. 

Solves https://github.com/music-assistant/support/issues/4763

Screenshot of MusicAssistant showing metadata of external media.


music_assistant/providers/hass_players/player.py
```diff
@@ -351,8 +387,35 @@ def update_from_compressed_state(self, state: CompressedState) -> None:
...
+        for source in self._attr_source_list:
+            if source.id == "External":
+                source.can_play_pause = MediaPlayerEntityFeature.PAUSE in hass_supported_features
+                source.can_next_previous = has_next_prev
+                source.can_seek = MediaPlayerEntityFeature.SEEK in hass_supported_features
+                break
+
     def _update_attributes(self, attributes: dict[str, Any]) -> None:
         """Update Player attributes from HA state attributes."""
+        self._hass_attributes.update(attributes)
```
MarvinSchenkel, line 408: I don't think we need to make this a private attribute of the Player. You seem to be only referring to `self._hass_attributes` inside this function, so you might as well use `attributes`

MarvinSchenkel:
Just a small comment, other than that, looks good 👏 . Marking this PR as draft so we can keep track of which PRs needs our attention. Please mark as 'Ready for review' when you want us to have another look 🙏 .

ulilicht:
thanks for the review, will need a few days until I can fix & test it and then mark it accordingly 🙌

music_assistant/providers/hass_players/player.py
ulilicht: I checked again, the issue here is that HA sends incremental attribute updates. So the content might be different each time we run through this function, and `media_content_id` might not be present in the current update, but we need it to know if we are currently having external playback. This is why this is stored on the player. 

Let me know if I oversaw something. 


Here an example, first update: 

Image

Second update: 

Image

MarvinSchenkel:
I think you forgot to address the feedback / push the commit? The only commits since my last review are the 'merge dev' commits?

ulilicht:
Hello @MarvinSchenkel , can you see this post above? 

I answered directly on your comment: https://github.com/music-assistant/server/pull/3015/changes/BASE..5c0a4b25f163af9570aa5c2e051391dc461e2a96#r2749975882

Best, 
Uli 

image

MarvinSchenkel:
> Hello @MarvinSchenkel , can you see this post above?
> 
> I answered directly on your comment: https://github.com/music-assistant/server/pull/3015/changes/BASE..5c0a4b25f163af9570aa5c2e051391dc461e2a96#r2749975882
> 
> Best, Uli
> 
> image

Ah that makes sense, in that case, this is ready to go :)

FYI, You 'started a review' rather than posted an individual comment, hence the 'Pending' message. So I can only see this when you press 'submit review' (or cancel the review and just send the individual)

MarvinSchenkel (approved):
Thanks for this nice improvement @ulilicht and congrats on your first contribution to MA!🎉

ulilicht:
🥳 thanks a lot for the support along the way & maintenance of this great project! Hope I manage to  contribute more from time to time in the future 🙌

ulilicht (commented):
for reference, comment which was not visible above

MarvinSchenkel:
> 🥳 thanks a lot for the support along the way & maintenance of this great project! Hope I manage to  contribute more from time to time in the future 🙌

And we'll be more than happy to support you. Looking forward to your first provider 😉

PR #3017, opened 2026-01-26
Allow nonexistent username for ingress auth

Weissnix4711:
https://github.com/music-assistant/support/issues/4851

### DIFF
--- a/music_assistant/controllers/webserver/helpers/auth_middleware.py
+++ b/music_assistant/controllers/webserver/helpers/auth_middleware.py
@@ -43,8 +43,10 @@ async def get_authenticated_user(request: web.Request) -> User | None:
         ingress_username = request.headers.get("X-Remote-User-Name")
         ingress_display_name = request.headers.get("X-Remote-User-Display-Name")
 
-        # Require all Ingress headers to be present for security
-        if not (ingress_user_id and ingress_username):
+        # Alternative auth providers may not provide username
+        ma_username = ingress_username or ingress_user_id
+
+        if not ingress_user_id:
             return None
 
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
I think you are trying to fix a non-existent problem here. Ingress auth (e.g. HA authenticating with MA) **must always** have a username.

MarvinSchenkel:
Just discussed this with the team. This is not something that should be fixed on the MA side, but rather on the HA side. @arturpragacz will have a look to see if he can figure out why the username is not being populated from the HA side. Little side note is that this a result from using a custom component, so we can't promise anything with regards to priority.

I will close this PR

Weissnix4711:
Hi,

Please feel free to correct me, but from what I can tell, home assistant users do not have a username, but it is part of the credential.

https://github.com/home-assistant/core/blob/ed70cacaa6e1bdfdd0c9594276b5d8502a891387/homeassistant/auth/models.py#L59

I also cannot find a reference to the 'username' field being actually required. From the documentation, it seems to suggest the credentials may include, but doesn't require a username.

> This will commonly be username and password [...]

https://developers.home-assistant.io/docs/auth_auth_provider/

As such from what I can tell, hass-oidc-auth is perfectly compliant here, and addons shouldn't be expecting `X-Remote-User-Name` to exist, but rather only depend on the user id.

Theweasels:
> Little side note is that this a result from using a custom component

This same issue occurs if using the built-in command line authentication as described here: https://www.home-assistant.io/docs/authentication/providers/#command-line

Users who log in this way do not have a username, only a Display Name and ID.
image

PR #3018, opened 2026-01-26
Sqeezelite: Only use low latency threshold values when a plugin is active

MarvinSchenkel:
Might fix this: https://github.com/music-assistant/support/issues/4756#issuecomment-3798247228


marcelveldt (approved):
This is fine for now - in the future we should consider adding a "is_love" or "realtime" flag to playermedia so that you dont have to do this plumbing within a player implementation

PR #3020, opened 2026-01-26
Add Last.fm Recommendations metadata provider

OzGav:
A new metadata provider that generates recommendation folders using the Last.fm API. Recommendations are resolved to playable items via streaming providers (Spotify, etc).

## Recommendation Rows

The provider generates up to 9 recommendation folders, arranged into four groups configurable via toggle:

### Personalized (based on user's MA library play counts)
- **Discover Similar Artists**: Top 5 most-played artists → 3 similar each → dedupe → first 10
- **Discover Similar Tracks**: Top 5 most-played tracks → 3 similar each → dedupe → first 10

### Global Charts (Last.fm worldwide data)
- **Global Top Artists**: Request 15 → resolve → dedupe → first 10
- **Global Top Tracks**: Request 15 → resolve → dedupe → first 10

### Genre-based (requires Last.fm username)
- **Discover {Genre} Artists**: User's top tag → 30 artists → filter library → top 3 + random 12 → resolve → first 10
- **Discover {Genre} Albums**: User's top tag → 30 albums → filter library → top 3 + random 12 → resolve → first 10
- **Discover {Genre} Tracks**: User's top tag → 30 tracks → filter library → top 3 + random 12 → resolve → first 10

### Geographic (configurable country)
- **Top Artists in {Country}**: Request 15 → resolve → dedupe → first 10
- **Top Tracks in {Country}**: Request 15 → resolve → dedupe → first 10

## Resolution Pipeline

1. Last.fm API returns artist/track/album names with optional MusicBrainz IDs (MBIDs)
2. For tracks: MBID → MusicBrainz API → ISRC lookup
3. Search streaming providers by ISRC (preferred) or name+artist fallback
4. Results cached at two levels: in-memory (session) and persistent (90 days)

## Key Features

- All recommendation groups are disabled by default (user enables what they want)
- Configurable refresh interval (default 6 hours)
- 20-second startup delay to allow streaming providers to load first
- Genre rows use "top 3 + random sample" strategy with daily seed for variety
- Library items filtered from genre recommendations to surface new music
- Refresh Now button in advanced settings for manual refresh


music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,389 @@
...
+            mass.create_task(provider._populate_recommendations())
+
+    return (
+        ConfigEntry(
+            key="api_key",
+            type=ConfigEntryType.SECURE_STRING,
+            label="Last.fm API Key",
+            required=True,
+            description="Get your API key from https://www.last.fm/api/account/create",
+            value=values.get("api_key") if values else None,
```
marcelveldt, line 113:
```suggestion
```
no need to set the value - that will be done by config controller

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,389 @@
...
+    ProviderFeature.RECOMMENDATIONS,
+}
+
+# Config action constants
+CONF_ACTION_CLEAR_CACHE = "clear_cache"
+
+# Curated list of popular countries for Last.fm geo charts
+# Last.fm API expects full country names (not ISO codes)
+# This list covers major music markets and can be expanded based on user requests
+GEO_COUNTRIES = [
```
marcelveldt, line 40: can these constants be moved to a dedicated constants.py in the provider folder?

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,389 @@
...
+        # Get the provider instance and clear its cache
+        provider = mass.get_provider(instance_id)
+        if isinstance(provider, LastFMRecommendationsProvider):
+            await provider.recommendations_manager.clear_cache()
+            # Trigger re-population after clearing
+            mass.create_task(provider._populate_recommendations())
+
+    return (
+        ConfigEntry(
+            key="api_key",
```
marcelveldt, line 108: is an api key required ?
we can also check if we can get an API key for our application, maybe through sponsorship etc.

music_assistant/providers/lastfm_recommendations/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "music",
+  "domain": "lastfm_recommendations",
```
marcelveldt, line 3: I'm not sure if I like the fact that we now have 2 lastfm providers.
We could potentially make this one the main provider and have the scrobbler depend on it or simply combine the 2 functions

OzGav: I looked at that but the recommendations feature is only available for music providers whereas the scrobbler is currently a plugin. Also you need your own API key and session key to scrobble but only an API key to pull info. So it gets a bit messy?

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Yes you do but they are easy to get individually

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Will do

music_assistant/providers/lastfm_recommendations/manifest.json
marcelveldt: ah ok that is clear. 
in that case I'm leaning towards making this a metadata provider and add support for recommendations from metadata modules to the core models because this is more a metadata provider than a music provider as it doesnt provide music on its own. Then suddenly also having similar tracks for filesystem items comes into reach meaning you can have radio mode for filesystem tracks

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Done

music_assistant/providers/lastfm_recommendations/manifest.json
OzGav: Agree that it is unusual to call this a music providers. Let me know what direction to head in

marcelveldt: give me a few days so I can overthink the best approach

OzGav: I asked Claude to come up with a proposal for a local radio mode and this is the result. Seems good to me? https://github.com/music-assistant/server/blob/claude/plan-lastfm-recommendations-lZsFl/RADIO_MODE_LOCAL_PLAN.md

marcelveldt: Yes but with one comment is that its all contained in the provider alone and we just add the get_similar_tracks etc. to the metadata model so we can utlize this for any metadata provider - for example one day I like to add an AI one

rendyhd: @marcelveldt is it confirmed to be as a metadata provider? I'm currently working on adding local file support to https://github.com/NeptuneHub/AudioMuse-AI with the plan to then use that as a radio feature for MA local files. I was looking at OzGrav's work for inspiration in how to implement it.

marcelveldt: Yes let's do this as metadata provider but we need a few small adjustments in core. I'll try to do so tomorrow or Tuesday but ping me if I forget

OzGav: @marcelveldt pinging as requested…

rendyhd: @marcelveldt any update on this?

marcelveldt:
I keep forgetting to come back on this one.
What we want is that metadata providers can also provide recommendations.

Long term goal is that we can have provider independent recommendations providers.
Based on services like listenbrainz, lastfm and ofcourse... AI

Now question is if this belongs to metadata providers or this should be plugins.
It definitely does not belong to the normal music providers as these providers do not provide music/audio on their own, they provide recommendations based on what you already have.

Also metadata is strictly about metadata so I'm inclined to say this should be plugin.
What do you think @OzGav ?

OzGav:
As discussed on Discord. Changed to a metadata provider.

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+            description="How often to refresh recommendations (0 to disable automatic refresh)",
+            category="recommendations",
+            range=(0, 168),  # 0 to 1 week
+        ),
+        ConfigEntry(
+            key="enable_similar_artists",
+            type=ConfigEntryType.BOOLEAN,
+            label="Enable Similar Artists (Personalized)",
+            default_value=False,
+            description="Show similar artists based on your listening history",
```
marcelveldt:
```suggestion
            description="Allow to use LastFM to provide similar artists based on your listening history",
```

Please change the wording of all these items because it provides something - it doesn't show anything.
Also wondering if we really need all this config toggles

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+        self.recommendations_manager = LastFMRecommendationManager(self)
+
+        # Initialize empty recommendation folders (will be populated in background)
+        # Always rebuild on provider load to ensure config changes take effect immediately
+        self._recommendation_folders: list[RecommendationFolder] = []
+        self._recommendations_populated = False
+
+        # Start background task to populate recommendations
+        # API key validation happens during first populate attempt
+        self.mass.create_task(self._populate_recommendations())
```
marcelveldt: can we do this with a scheduled task ?

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,336 @@
...
+
+        # Schedule periodic refresh using MA's scheduler
+        self._schedule_refresh()
+
+    async def _populate_recommendations(self) -> None:
+        """Populate recommendation folders in the background.
+
+        This runs immediately after initialization, building recommendation
+        folders progressively. Each folder appears as soon as it's populated,
+        allowing the frontend to display results incrementally.
```
marcelveldt: I don't see how this is related to the frontend.

OzGav: Complete tidy up of docstrings and comments has been done

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Descrptions reworded. 

As for the number of groups I initially did it this way to minimise the API calls. I didn't want the user to just hide the rows in the frontend and have a bunch of calls going on that were just wasted. However, I have looked at it again and due to the heavy caching being done after the first round of calls then extra calls is not that much. So I have collapsed the existing nine toggles into four.

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Yes. I switched to `mass.call_later` for the initial populate, same as the refresh path already uses. Dropped the              `asyncio.sleep(20)` inside `_populate_recommendations`; the 20-second delay is now expressed by scheduling the call 20 seconds out.  The populate function is now just "build the folders", and if the provider reloads during the wait the scheduled task gets cleaned up by its `task_id` instead of leaving a detached task running.

music_assistant/providers/lastfm_recommendations/manifest.json
OzGav: domain changed to metadata

OzGav: Changed to metadata provider

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: .

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+                    "Added %d geography-based recommendation folder(s)", len(geo_folders)
+                )
+
+            self._recommendations_populated = True
+            self.logger.info(
+                "Last.fm recommendations built (%d folders)",
+                len(self._recommendation_folders),
+            )
+            # TODO: signal the frontend to refresh the Discover view once a suitable
+            # event exists upstream (e.g. a RECOMMENDATIONS_UPDATED EventType).
```
marcelveldt: the frontend refreshes automatically if you open the discover page
as soon as the cache is expired and you (re)open the discover view, you will see fresh recommendations.
isn't that enough ?

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+        except MusicAssistantError as err:
+            self.logger.warning("Failed to populate recommendations: %s", err)
+
+    def _schedule_refresh(self) -> None:
```
marcelveldt: we have a task scheduler for this - please adjust the logic to use a scheduled task instead

music_assistant/providers/lastfm_recommendations/api_client.py
```diff
@@ -0,0 +1,330 @@
...
+                        )
+
+                    msg = f"Last.fm API error {error_code}: {error_msg}"
+                    raise InvalidDataError(msg)
+
+                return data
+
+    async def get_similar_artists(
+        self, artist_name: str, artist_mbid: str | None = None, limit: int = 10
+    ) -> list[dict[str, Any]]:
```
marcelveldt, line 82: I like to see this one refactored:

- Add a ProviderFeature.SIMILAR_ARTISTS
- Add this method (with a stub implementation that checks the feature) to the MetadataProvider model
- Add this method (with a stub implementation that checks the feature) to the MusicProvider model
- For the metadata provider, this takes an Artist object as input
- For the music provider, this takes an provider_artist_id as input

music_assistant/providers/lastfm_recommendations/api_client.py
```diff
@@ -0,0 +1,330 @@
...
+            # Last.fm returns a single dict when only one result is present.
+            if isinstance(similar_artists, dict):
+                return [similar_artists]
+
+            return similar_artists
+
+        except (TimeoutError, ClientError, InvalidDataError, KeyError):
+            return []
+
+    async def get_similar_tracks(
```
marcelveldt, line 120: I like to see this one refactored:

- Add the ProviderFeature.SIMILAR_TRACKS to your provider
- Add this method (with a stub implementation that checks the feature) to the MetadataProvider model
- The MusicProvider model already has this base method defined
- For the metadata provider, this takes an Track object as input
- For the music provider, this takes an provider_track_id as input

By doing this, we allow future expansion of using metadata providers to participate in "radio mode/dynamic playlists" - which also allows for people with only local tracks (or a streaming provider without similar tracks) to get this feature

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+        self._schedule_refresh()
+
+    async def _populate_recommendations(self) -> None:
+        """Populate recommendation folders in the background."""
+        try:
+            self.logger.info("Building Last.fm recommendations")
+
+            # Build folders incrementally so each category appears as soon as it's ready.
+            personalized_folders = (
+                await self.recommendations_manager._get_personalized_recommendations()
```
marcelveldt: why are you calling all these private methods on the recommendations_manager?

music_assistant/providers/lastfm_recommendations/api_client.py
marcelveldt: OK, I was looking at the api client apparently so ignore this.
We can do what I asked above in a follow-up PR because I think this can be more powerful than just providing recommendation rows.

music_assistant/providers/lastfm_recommendations/api_client.py
marcelveldt: OK, I was looking at the api client apparently so ignore this.
We can do what I asked above in a follow-up PR because I think this can be more powerful than just providing recommendation rows.

music_assistant/providers/lastfm_recommendations/mbid_resolver.py
```diff
@@ -0,0 +1,81 @@
...
+if TYPE_CHECKING:
+    from music_assistant.providers.lastfm_recommendations import LastFMRecommendationsProvider
+    from music_assistant.providers.musicbrainz import MusicbrainzProvider
+
+
+CACHE_CATEGORY_MBID_ISRC = 0  # Cache category for MBID->ISRC mappings
+
+
+class MBIDResolver:
+    """Resolves MusicBrainz recording IDs to ISRCs."""
```
marcelveldt: Can you explain why this is needed this way and we can't just add some (cached) helper methods to the MusicBrainz provider?

music_assistant/providers/lastfm_recommendations/api_client.py
OzGav: Yes absolutely that is what I was thinking as well.

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Ok. I was just looking at how the discover page was refreshed on MEDIA_ITEM_PLAYED and thought there was therefore a desire to have it refresh live while someone was looking at it. I’ll remove the TODO.

music_assistant/providers/lastfm_recommendations/mbid_resolver.py
OzGav: Yes I could. I did it this way because I wanted a 90-day extended cache on top of MusicBrainz’s 30-day cache since the ISRC hardly ever change, and I wanted the cache limited to this provider instance so clearing the Last.fm recommendations cache doesn’t affect MusicBrainz’s own cache. But if you’d prefer this as a helper method on the MusicBrainz provider, I can move it there - it’s essentially just calling get_recording_details() and extracting the ISRCs

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Yeah this isn’t right. I’ll add one public helper method instead

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Yes done. And refresh config entry removed as that is done via the task scheduler now

music_assistant/providers/lastfm_recommendations/mbid_resolver.py
OzGav: Actually looking at it again I am misunderstanding how the caching is working. The ISRC cache is scoped to the lastfm_recommendations provider instance, so clearing Last.fm’s cache wouldn’t affect MusicBrainz’s own cache in either approach. The only difference is whether ISRCs get cleared when the user clicks ‘Refresh Recommendations’ - and since ISRCs are stable data, I actually don’t want to clear them.
Moving it to a helper method on the MusicBrainz provider makes sense - it’s cleaner separation of concerns and means faster recommendation rebuilds since the ISRC mappings stay cached. I’ll move it there

music_assistant/providers/lastfm_recommendations/recommendations.py
```diff
@@ -0,0 +1,781 @@
...
+            similar = await self.api.get_similar_tracks(
+                artist_name=artist_name,
+                track_name=seed_track.name,
+                track_mbid=mbid,
+                limit=SIMILAR_ITEMS_PER_SEED,
+            )
+            all_similar.extend(similar)
+
+        # Deduplicate by MBID and by artist+name: Last.fm sometimes returns the same track
+        # twice, once with an MBID and once without.
```
OzGav:
```suggestion
        # twice, once with a MBID and once without.
```

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+            translation_key="background_task.refresh_lastfm_recommendations",
+        )
+
+        # Build on startup so the UI isn't empty until the next scheduled refresh.
+        # Delayed to let streaming providers finish loading before resolution runs.
+        self.mass.call_later(
+            20,
+            self._refresh_recommendations,
+            task_id=f"{REFRESH_TASK_ID}_initial_{self.instance_id}",
+        )
```
marcelveldt: this should be done with the background task - which should run immediately if the provider was just installed

marcelveldt: you can simply delete this block - the task defined above already takes care of this

OzGav: Block deleted

dmoo500:
Hi @OzGav — just wanted to flag that PR #3811 (currently in review) would allow Plugin Providers to implement music-related ProviderFeatures like `SIMILAR_TRACKS` and `RECOMMENDATIONS`.

Since this provider doesn't own any tracks itself but uses the Last.fm API to generate recommendations resolved via other providers, it might be worth integrating this into the existing lastfm_scrobble plugin once #3811 lands — rather than maintaining a separate Music Provider. That would also avoid users having to enter their Last.fm credentials twice (once for scrobbling, once for recommendations).

Might be worth keeping an eye on #3811 before investing more time here!

OzGav:
@dmoo500 i previously discussed this with Marcel. This is a metadata provider not a plugin. Also you need your own API key and session key to scrobble but only an API key to pull info. So it gets a bit messy. Will see if Marcel changes his mind.

music_assistant/providers/lastfm_recommendations/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "metadata",
+  "domain": "lastfm_recommendations",
+  "name": "Last.fm Recommendations",
+  "description": "Get music recommendations from Last.fm based on your listening history",
+  "codeowners": [
+    "ozGav"
```
MarvinSchenkel: Missing an `@` here?

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider."""
+    if action == CONF_ACTION_CLEAR_CACHE and instance_id:
+        provider = mass.get_provider(instance_id)
+        if isinstance(provider, LastFMRecommendationsProvider):
+            await provider.recommendations_manager.clear_cache()
+            mass.create_task(provider._refresh_recommendations())
+
+    return (
+        ConfigEntry(
```
MarvinSchenkel, line 76: We recently changed the lastfm provider to use the MA API Key. Maybe we can do that too here ?

music_assistant/providers/lastfm_recommendations/manifest.json
OzGav: Fixed

music_assistant/providers/lastfm_recommendations/api_client.py
```diff
@@ -0,0 +1,341 @@
...
+        if "error" in data:
+            error_code = int(data.get("error", 0))
+            error_msg = data.get("message", "Unknown error")
+            msg = f"Last.fm API error {error_code}: {error_msg} (method: {method})"
+            exc_cls = self._ERROR_MAP.get(error_code, InvalidDataError)
+            raise exc_cls(msg)
+
+        return data
+
+    async def get_similar_artists(
```
MarvinSchenkel, line 85: We could now expose the as a provider feature via init.py by calling this and defining `ProviderFeature.SIMILAR_ARTISTS`

music_assistant/providers/lastfm_recommendations/api_client.py
MarvinSchenkel: We could now expose the as a provider feature via init.py by calling this and defining `ProviderFeature.SIMILAR_TRACKS`

music_assistant/providers/lastfm_recommendations/recommendations.py
```diff
@@ -0,0 +1,704 @@
...
+            return items
+
+        top_items = items[:TOP_ITEMS_TO_TAKE]
+
+        remaining = items[TOP_ITEMS_TO_TAKE:]
+        random_count = target_count - TOP_ITEMS_TO_TAKE
+
+        # Daily seed keeps recommendations stable within the day and rotates them overnight.
+        seed = f"{datetime.datetime.now(tz=datetime.UTC).date().isoformat()}_{seed_suffix}"
+        random.seed(seed)
```
MarvinSchenkel: Please use a local random instance here to avoid the seed from interfering with the rest of MA

music_assistant/providers/lastfm_recommendations/parsers.py
```diff
@@ -0,0 +1,430 @@
...
+                for isrc in isrcs:
+                    external_ids.add((ExternalID.ISRC, isrc))
+            else:
+                LOGGER.debug("No ISRCs found for MBID %s", mbid)
+        else:
+            LOGGER.debug("MusicBrainz provider not available for ISRC lookup")
+    else:
+        LOGGER.debug("Track has no MBID, cannot resolve ISRCs")
+
+    image = None
```
MarvinSchenkel: I think these images are useless / dead code since you resolve this item mapping to an actual media item later right?

music_assistant/providers/lastfm_recommendations/parsers.py
```diff
@@ -0,0 +1,430 @@
...
+    if isinstance(artist_data, str):
+        artist_name = artist_data
+    else:
+        artist_name = artist_data.get("name", "Unknown Artist")
+
+    external_ids = set()
+    if mbid:
+        external_ids.add((ExternalID.MB_ALBUM, mbid))
+
+    image = None
```
MarvinSchenkel: I think these images are useless / dead code since you resolve this item mapping to an actual media item later right?

music_assistant/providers/lastfm_recommendations/parsers.py
```diff
@@ -0,0 +1,430 @@
...
+    """
+    name = lastfm_artist.get("name", "Unknown Artist")
+    mbid = lastfm_artist.get("mbid")
+
+    external_ids = set()
+    if mbid:
+        external_ids.add((ExternalID.MB_ARTIST, mbid))
+
+    image = None
+    if image_url := _extract_image_url(lastfm_artist.get("image", [])):
```
MarvinSchenkel: I think these images are useless / dead code since you resolve this item mapping to an actual media item later right?

MarvinSchenkel (changes requested):
Few minor things, already looks good 👍

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Done

music_assistant/providers/lastfm_recommendations/recommendations.py
OzGav: Fixed

music_assistant/providers/lastfm_recommendations/parsers.py
OzGav: Removed

music_assistant/providers/lastfm_recommendations/parsers.py
OzGav: Removed

music_assistant/providers/lastfm_recommendations/parsers.py
OzGav: Removed

music_assistant/providers/lastfm_recommendations/api_client.py
OzGav: OK added and ProviderFeature.SIMILAR_TRACKS as well

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -206,3 +216,34 @@ async def _refresh_recommendations(self) -> None:
...
+
+        :param artist: The reference artist.
+        :param limit: Maximum number of similar artists to return.
+        """
+        artist_mbid = artist.get_external_id(ExternalID.MB_ARTIST)
+        similar_raw = await self.api.get_similar_artists(artist.name, artist_mbid, limit)
+
+        results: list[Artist] = []
+        for raw in similar_raw:
+            if resolved := await self.recommendations_manager._get_or_resolve_artist(raw):
```
MarvinSchenkel: Let's use asyncio.gather here to prevent synchronous api calls (which slow down everything)

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -206,3 +216,34 @@ async def _refresh_recommendations(self) -> None:
...
+        :param track: The reference track.
+        :param limit: Maximum number of similar tracks to return.
+        """
+        artist_name = track.artists[0].name if track.artists else "Unknown Artist"
+        track_mbid = track.get_external_id(ExternalID.MB_RECORDING)
+        similar_raw = await self.api.get_similar_tracks(artist_name, track.name, track_mbid, limit)
+
+        results: list[Track] = []
+        for raw in similar_raw:
+            if resolved := await self.recommendations_manager._get_or_resolve_track(raw):
```
MarvinSchenkel: Let's use asyncio.gather here to prevent synchronous api calls (which slow down everything)

OzGav: Fixed

music_assistant/providers/lastfm_recommendations/__init__.py
OzGav: Fixed and I noticed I had a private method there after the last change so I fixed that as well.

MarvinSchenkel (approved):
Nice one @OzGav 🎉

PR #3025, opened 2026-01-26
Fix group volume mute in Sendspin provider

scyto:
## Summary
- Fix `volume_mute()` to propagate mute to all group members when called on the group leader
- Use `group.set_mute()` for leader/solo players (distributes to all members via aiosendspin)
- Use per-player `player.mute()`/`player.unmute()` for individual group members (`synced_to is not None`)

## Problem
When MA mutes a grouped player, `cmd_volume_mute()` calls `volume_mute()` on the leader only. The previous implementation called `player.mute()` which is per-player — only the leader was muted, other group members were unaffected.

## Fix
Check `self.synced_to` to distinguish the group leader from a member:
- **Leader or solo** (`synced_to is None`): call `group.set_mute(muted)` which iterates all players in the aiosendspin group
- **Group member** (`synced_to is not None`): call `player.mute()`/`player.unmute()` for individual control

## Test plan
- [x] Unit test with aiosendspin: 4 players grouped, `group.set_mute(True)` reaches all 4, `player.mute()` reaches only 1
- [x] End-to-end test: patched nightly Docker image with 4 Sendspin SDK clients — group mute/unmute works, individual member mute works

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -358,11 +358,16 @@ class SendspinPlayer(Player):
 
     async def volume_mute(self, muted: bool) -> None:
         """Handle VOLUME MUTE command on the player."""
-        if player_client := self.api.player:
-            if muted:
-                player_client.mute()
-            else:
-                player_client.unmute()
+        if self.synced_to is not None:
+            # This player is a group member — mute only this player
+            if player_client := self.api.player:
+                if muted:
+                    player_client.mute()
+                else:
+                    player_client.unmute()
+        else:
+            # This player is the leader (or solo) — mute the entire group
+            self.api.group.set_mute(muted)
 
     async def stop(self) -> None:
         """Stop command."""


scyto:
https://github.com/user-attachments/assets/17161409-8ed2-47db-bbf9-eafa80d5f040

marcelveldt (changes requested):
This is the wrong fix and causes unexpected behavior. 
User wants to mute one player (which happens to be the group leader) and then the whole group gets muted ? That is wrong. We have a dedicated api command to mute the entire group which is what needs to be used.

scyto:
> This is the wrong fix and causes unexpected behavior. User wants to mute one player (which happens to be the group leader) and then the whole group gets muted ? That is wrong. We have a dedicated api command to mute the entire group which is what needs to be used.

the default behaviour without this fix is that when muting the group if there are 4 players it will mute the leader as MA treats and the provider treat them as synomous (i.e it's not a bug i introduced) but nothing else in the group will be muted

i don't immediately see a way to make the leader and the group not be synomous (maybe its a bug in MA?)

~~i couldn't find that API dedicated to mute for group (if you mean the controller group message that only retuns as true once every player has been muted, but the provider only sends one message to one player to mute and so it never gets set as true in my testing), only  group_volume, group_volume_up and group_volume_down if you could point me in the right direction for the API you mention that would be good and i will go dig deeper with chris to revist if its a C# SDK issue or not~~

considering this more, sorry for the noise


+ @chrisuthe

marcelveldt:
It's actually a bug in the MA frontend...

scyto:
> It's actually a bug in the MA frontend...

i started playing with that and now realize it has many downstream implications on how the volume controls work and potentially other providers

i am an idiot-savant (and half decent PM and claude) i am bulding a running version of the front end that 'works how i want' as a starting point if this is something other would be interested in discussing (or is there already a plan for changing how this works in MA)?

marcelveldt:
I just fixed the bug in the frontend - available in next update

scyto:
> I just fixed the bug in the frontend - available in next update

thanks, i will evlauate after next nightly

PR #3027, opened 2026-01-26
Fix players unavailable of cast and bluesound provider

marcelveldt:


marcelveldt:
must be backported to stable if models bump is also going to be backported ( I dont it will though)

PR #3034, opened 2026-01-27
Add group volume mute support

scyto:
## Summary

- Add `cmd_group_volume_mute` API endpoint (`players/cmd/group_volume_mute`) that mutes/unmutes all powered group members concurrently, mirroring the existing `cmd_group_volume` pattern
- Add `ATTR_MUTE_LOCK` player attribute to track individually muted players, preventing group volume changes from auto-unmuting them

## Context

This is the server-side counterpart to the frontend PR music-assistant/frontend#TBD. The maintainer's frontend commit d1600778 includes a TODO: "revisit this when api/server supports group mute toggle" — this PR provides that server API.

Previously PR #3025 attempted to fix this in the Sendspin provider, but per maintainer feedback the fix belongs in the controller layer.

### What `cmd_group_volume_mute` does
- For group players or sync leaders: iterates `iter_group_members(only_powered=True)` and calls `cmd_volume_mute` on each concurrently via `asyncio.gather`
- For solo players: falls back to per-player `cmd_volume_mute`

### What `ATTR_MUTE_LOCK` does
- Set on a player when it's muted while part of a group
- Cleared when the player is explicitly unmuted
- Checked in `_handle_cmd_volume_set` — if set, the player stays muted even when group volume changes
- This keeps the public `cmd_volume_set` API unchanged while preserving mute state during group volume adjustments

## Test plan

- [ ] Verify `cmd_group_volume_mute` mutes/unmutes all group members when called on group leader
- [ ] Verify `cmd_group_volume_mute` falls back to per-player mute for solo players
- [ ] Verify group volume slider changes don't unmute individually muted players (mute lock behavior)
- [ ] Verify unmuting a player clears the mute lock
- [ ] Verify solo player behavior is unchanged

### DIFF
--- a/music_assistant/controllers/players/player_controller.py
+++ b/music_assistant/controllers/players/player_controller.py
@@ -60,6 +60,7 @@ from music_assistant.constants import (
     ATTR_FAKE_VOLUME,
     ATTR_GROUP_MEMBERS,
     ATTR_LAST_POLL,
+    ATTR_MUTE_LOCK,
     ATTR_PREVIOUS_VOLUME,
     CONF_AUTO_PLAY,
     CONF_ENTRY_ANNOUNCE_VOLUME,
@@ -680,16 +681,13 @@ class PlayerController(CoreController):
 
     @api_command("players/cmd/volume_set")
     @handle_player_command
-    async def cmd_volume_set(
-        self, player_id: str, volume_level: int, preserve_mute: bool = False
-    ) -> None:
+    async def cmd_volume_set(self, player_id: str, volume_level: int) -> None:
         """Send VOLUME_SET command to given player.
 
         :param player_id: player_id of the player to handle the command.
         :param volume_level: volume level (0..100) to set on the player.
-        :param preserve_mute: if True, do not auto-unmute the player.
         """
-        await self._handle_cmd_volume_set(player_id, volume_level, preserve_mute=preserve_mute)
+        await self._handle_cmd_volume_set(player_id, volume_level)
 
     @api_command("players/cmd/volume_up")
     @handle_player_command
@@ -813,9 +811,6 @@ class PlayerController(CoreController):
             ):
                 coros.append(self.cmd_volume_mute(child_player.player_id, muted))
             await asyncio.gather(*coros)
-            return
-        # treat as normal player mute
-        await self.cmd_volume_mute(player_id, muted)
 
     @api_command("players/cmd/volume_mute")
     @handle_player_command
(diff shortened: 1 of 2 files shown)


scyto:
demo docker image  `ghcr.io/scyto/ma-server-patched:group-mute`

music_assistant/controllers/players/player_controller.py
```diff
@@ -792,6 +795,32 @@ async def cmd_group_volume_down(self, player_id: str) -> None:
...
+            for child_player in self.iter_group_members(
+                player, only_powered=True, exclude_self=False
+            ):
+                coros.append(self.cmd_volume_mute(child_player.player_id, muted))
+            await asyncio.gather(*coros)
+            return
+        if player.synced_to and (sync_leader := self.get(player.synced_to)):
+            # redirect to sync leader
+            await self.cmd_group_volume_mute(sync_leader.player_id, muted)
+            return
```
marcelveldt: this is weird - this doesn't belong here

marcelveldt:
```suggestion
```

music_assistant/controllers/players/player_controller.py
```diff
@@ -2359,7 +2391,8 @@ async def _handle_cmd_volume_set(self, player_id: str, volume_level: int) -> Non
             )
 
         if (
-            player.mute_control not in (PLAYER_CONTROL_NONE, PLAYER_CONTROL_FAKE)
+            not preserve_mute
+            and player.mute_control not in (PLAYER_CONTROL_NONE, PLAYER_CONTROL_FAKE)
```
marcelveldt, line 2396:
```suggestion
            and player.mute_control != PLAYER_CONTROL_NONE
```

music_assistant/controllers/players/player_controller.py
scyto: fixed with your suggestion

music_assistant/controllers/players/player_controller.py
scyto: @marcelveldt as i am honest about my use of claude to help me, here is a summary of the analysis it helped me do on how i could accept your suggestion and still maintain the goal in the frontend.  I am open to any approach that achieves the front end objective as i don't have the expertise to know what the right approach is.
 
-----

On preserve_mute — the behavior I'm trying to enable is: in a group of 4 players, I can mute my bedroom speaker (so it doesn't wake someone up) while still adjusting the group volume for the other rooms. Right now, dragging the group volume slider unmutes the bedroom speaker because set_group_volume calls _handle_cmd_volume_set on each member, which auto-unmutes them. The frontend can't prevent this since it sends a single cmd_group_volume call and the server distributes it internally.

Would you be open to keeping preserve_mute as an internal-only parameter on _handle_cmd_volume_set (not exposed on the public API), used only by set_group_volume? That way the public API stays unchanged but individually-muted members stay muted during group volume adjustments.

marcelveldt: you can save it as an additional attribute in the extra_data (not exposed on api) or in extra_attributes (exposed on data). Make sure to define a constant for it.

scyto: i believe this resolves the feedback https://github.com/music-assistant/server/pull/3034/commits/9437f0e74ae15b74cdf6ef3352393c4e825209c0 also amended the front end PR to. work with this, built and tested docker container

marcelveldt (approved):
Looks good, thanks @scyto !

scyto:
@marcelveldt  thanks for your support and patience and direction, i appreciate it, i can imagine it might be frustrating to have all us 'vibe coders' come out of the woodwork :-)

now i look forward to your questions about the frontend PR :-)

PR #3035, opened 2026-01-27
Raise AudioException if ffmpeg encounters a demuxing error

fmunkes:
Related to https://github.com/music-assistant/support/issues/4754

Please triple check this code. I'd like to have an Audio Exception if the ffmpeg demuxer encounters an error to prevent progress reporting as finished.


marcelveldt (approved):
Thanks, @fmunkes !

PR #3038, opened 2026-01-28
Fix IndexError when Deezer track has no media available

sfortis:
## Summary
- Raise `MediaNotFoundError` instead of crashing with `IndexError` when the Deezer API returns an empty media list
- This allows the player queue to gracefully skip unavailable tracks instead of stopping playback

## Test plan
- Tested locally by patching Music Assistant server
- Playback now skips unavailable tracks with a warning: `Skipping unplayable item...`

Fixes music-assistant/support#4860


MarvinSchenkel (approved):
Great find, thanks @sfortis !

sfortis:
> Great find, thanks @sfortis !

It's my pleasure to contribute to this great project!

PR #3043, opened 2026-01-28
Add YouSee Musik provider

math625f:
This PR contains my initial implementation of the YouSee Musik provider for MA.
It solves the request: https://github.com/orgs/music-assistant/discussions/4558#discussion-9261232

I don't personally use the service (I've borrowed access to implement it), so I'd like to keep it tagged as experimental/alpha for now, until someone has used it over a period, to ensure everything works.

I have some questions:

1. Is there a more intended way for me to obtain custom variables from the streamdetails in the on_streamed method? The API requires me to send a "playbackContext", but I couldn't figure out a good way to store it in the stream, so I'm fetching the track again in the method for reporting playback.
2. I'm using the MA locale to obtain results from the api in that language, do you think this is the correct way, or should I have a separate toggle? I'm fairly certain only Danish and English is supported in the API.
3. I'm currently just adding "featured artists" as artists, is that right? I've been unable to find a track returned by the API that actually has featured artists populated, even if it says in the name of the track it "features" someone.
4. In YouSee Musik it's called "My music", the YouSee Musik calls it "favourites", should I be using the "Library" or "Favourites" functionality in Music Assistant? It's a streaming service similar to Spotify, you have a subscription, and then you can listen to their full catalog.
5. It looks like YouSee Musik has support for lyrics, some with timestamps and some without, should I implement it? And if yes, am I just supposed to populate the lyrics on the track in the get_track method, or is it fetched independently somehow?
6. My log sometimes shows the error
  ```
  Random aiohttp.client_exceptions.ClientPayloadError: 400, message:
    Can not decode content-encoding: br
  ```
  is this something my code is causing, or is it just something that happens once in a while? I didn't look that much into it, but I haven't been able to find the root cause.


OzGav:
I'll let the others answer 1 & 2
3. Where are these "featured artists" coming from? I think you will find the music controller will strip off the featured artists part from the artist name and add that as another artist. So A feat. B will show the track as being performed by A / B.
4. In MA favourites are just a subset of the library. If YouSee doesn't have two levels then just put their favourites into the MA library but don't mark it as a favourite (in MA). 
5. Have a look at Tidal as I think it has the same (or similar) functionality
6. I'll let the others look at that

init.py is getting pretty big. Have a look at Spotify to see how you can separate the code into managable files.

math625f:
>Where are these "featured artists" coming from? I think you will find the music controller will strip off the featured artists part from the artist name and add that as another artist. So A feat. B will show the track as being performed by A / B. 4.

It's the YouSee Musik API that doesn't return any featured artists, so I haven't been able to test it properly, I'm assuming MA would handle it appropriately if it had the chance to do so :)

Regarding point 4, I'll keep it as I've implemented now in that case, where "liked" tracks are just imported to the library.

MarvinSchenkel:
Please put your GraphQL queries in a separate file. It's really hard to properly review this. Suggestion:
`queries.py`
```
SEARCH_ARTISTS = """
....
"""
LIBRARY_GET_PLAYLISTS = """
...
"""
```

MarvinSchenkel:
First off, a huge thanks for taking the effort to contribute a new music provider 👏 .

To answer your questions:

1. I would not fetch the entire track, feels like a waste of resources. What is it exactly that needs to be provided there? Otherwise I would just use the track identifier from `streamdetails.item_id`
2. That is correct. Have a test with different locales to see if the API does not return errors. If it does, then you can use something like `local = 'danish if ma_local='danish' else 'english'`
3. Yes, they are usually the 'secondary artists' of a track or album, e.g. 'XXX ft. YYY`. The artist should then be XXX and Featured artists YYY, but within MA you can just provider 1 list of artists: `[artists + featured_artists]`
4. Just double checking, some music providers differentiate favourites from library. For example, in Apple music and YT Music I can add albums and track to my library (making them accessible quickly via library > albums), but I can also like/favourite/heart a track, which makes them show up within my likes. Spotify does not have this concept, there you can only 'add to liked songs'. If your provider supports both, then map them to library / favourites accordingly. If they do not, then you can map them to library. Have a look at the providers I mentioned to see the difference for yourself
5. Tidal is indeed a good example, so you obtain them through `get_track` and set them on `metadata.lyrics` in `parse_track`.
6. Does not look familiar to me. Would be good to trace where this error is coming from (api call? during streaming?) so we can see the origin.

I'll do a more thorough review once the initial feedback has been addressed 🙏 .

Marking this PR as draft again so we know which ones we still have to give feedback to ;-). Feel free to mark it as 'ready for review' again once you have made some progress 👍 .

helgedk:
Huge thanks to @math625f for picking this up so quickly! I'm the one who started the discussion, and I can clarify a few things from a user perspective:

Library/Favorites: YouSee Music is very similar to Spotify. We don't have two levels of 'saving'. We just use the 'Heart' icon to add tracks/albums to 'My Music'. So mapping them to the MA Library is the correct way to go.

Featured Artists: You are right, the API often leaves the featuredArtists field empty even when there's a feature. It seems they often just bake the 'feat.' part directly into the track title.

Lyrics: Yes, please implement them if possible! It's a popular feature, and as you noted, YouSee does support it with timestamps for many tracks.

I'm ready to test the Alpha/Alpha-version as soon as it's available in a build!

math625f:
I've now restructured the code largely based on the structure used by the Tidal provider, @MarvinSchenkel would you still like me to completely separate out the queries, or how do you feel about the current shape of it?

Regarding locales, it seems the API just ignores other languages, so it's probably fine as-is.

I didn't see the error again while restructuring, I'll keep an eye out for it.

I'll look into implementing Lyrics support soon.

As for question 1, this is what the official YouSee Musik website sends to its API when it reports playback:
```
mutation reportPlayback($report: ReportPlaybackInput!) {
  reportPlayback(report: $report) {
    ok
    __typename
  }
}
```
Variables: 
```
{
    "report": {
        "playbackUrl": "https://...",
        "playedSeconds": 42,
        "playbackContext": "dX....3M=",
        "playedAt": "2026-01-28T20:08:35.402Z"
    }
}
```
Everywhere else, when fetching tracks, albums, adding tracks to playlists etc, even when fetching the `stream_details` it uses proper `item_ids`, but for the `reportPlayback` functionality it uses this "playbackContext" string instead. I'm unsure what exactly it does, all I know is that `ReportPlaybackInput` seems to require it.

music_assistant/providers/yousee/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+) -> tuple[ConfigEntry, ...]:
+    """
+    Return Config entries to setup this provider.
+
+    instance_id: id of an existing provider instance (None if new instance setup).
+    action: [optional] action key called from config entries UI.
+    values: the (intermediate) raw values for config entries sent with the action.
+    """
+    # ruff: noqa: ARG001
+    return (
```
MarvinSchenkel, line 73: Just double checking. Does YouSee have an OAuth flow available voor authentication?

music_assistant/providers/yousee/api_client.py
```diff
@@ -0,0 +1,109 @@
...
+        self.auth = provider.auth
+        self.logger = provider.logger
+        self.mass = provider.mass
+
+    @throttle_with_retries  # type: ignore[type-var]
+    async def post_graphql(
+        self, query: str, variables: JsonLike, _headers: JsonLike | None = None
+    ) -> JsonLike:
+        """Post GraphQL query to YouSee endpoint with authorization."""
+        # TODO: Is this the right way to do determine locale?
```
MarvinSchenkel: Yes

music_assistant/providers/yousee/constants.py
```diff
@@ -0,0 +1,12 @@
...
+
+VARIOUS_ARTISTS_ID = "1776"
+
+PAGE_SIZE = 50
+# to avoid infinite loops, this effectively limits any album/playlist to
+# PAGE_SIZE * MAX_PAGES_PAGINATED items (1000 items with the current settings)
+MAX_PAGES_PAGINATED = 20
+GET_POPULAR_TRACKS_LIMIT = 25
+
+PLAYBACK_QUALITY = "KBPS_320"
```
MarvinSchenkel: Are there more options available than 320kbps?

music_assistant/providers/yousee/library.py
```diff
@@ -0,0 +1,253 @@
...
+                }
+            }
+        }
+        """
+        variables = {"imageSize": IMAGE_SIZE}
+
+        async for item in self.api.paginate_graphql(
+            query, variables, ["data", "me", "favorites", "artists"]
+        ):
+            self.logger.debug("Parsing artist item: %s", item)
```
MarvinSchenkel: We have a verbose level for these extensive log objects:
```
self.logger.log(VERBOSE_LOG_LEVEL, "....")
```

music_assistant/providers/yousee/library.py
```diff
@@ -0,0 +1,253 @@
...
+                }
+            }
+        }
+        """
+        variables = {"imageSize": IMAGE_SIZE}
+
+        async for item in self.api.paginate_graphql(
+            query, variables, ["data", "me", "favorites", "albums"]
+        ):
+            self.logger.debug("Parsing album item: %s", item)
```
MarvinSchenkel: Same comment as above, please use verbose :-)

music_assistant/providers/yousee/library.py
```diff
@@ -0,0 +1,253 @@
...
+                    }
+                }
+            }
+        """
+        variables = {"imageSize": IMAGE_SIZE}
+
+        async for item in self.api.paginate_graphql(
+            query, variables, ["data", "me", "favorites", "tracks"]
+        ):
+            self.logger.debug("Parsing track item: %s", item)
```
MarvinSchenkel: Same comment as above, please use verbose :-)

music_assistant/providers/yousee/library.py
```diff
@@ -0,0 +1,253 @@
...
+                        }
+                    }
+                }
+            }
+        """
+        variables = {"imageSize": IMAGE_SIZE}
+        async for item in self.api.paginate_graphql(
+            query, variables, ["data", "me", "playlists", "combinedPlaylists"]
+        ):
+            self.logger.debug("Parsing playlist item: %s", item)
```
MarvinSchenkel: Same comment as above, please use verbose :-)

helgedk:
Regarding the audio quality question by @MarvinSchenkel: I've checked the official app settings, and it confirms that YouSee Musik offers two tiers: Normal (192 kbit/s) and High (320 kbit/s).
There are no lossless or higher bitrate options available at this time. I've attached a screenshot from the app as documentation. So KBPS_320 is definitely the correct maximum to target.
Screenshot_20260129-113821

music_assistant/providers/yousee/parsers.py
```diff
@@ -0,0 +1,219 @@
...
+        duration=track_obj.get("duration", 0),
+        provider_mappings={
+            ProviderMapping(
+                item_id=str(track_obj["id"]),
+                provider_domain=provider.domain,
+                provider_instance=provider.instance_id,
+                available=track_obj.get("availableToStream", True),
+                audio_format=AudioFormat(
+                    content_type=ContentType.MP4,
+                    bit_rate=320000,
```
MarvinSchenkel: Bitrate is in kbps, so `320`. Maybe refer to your constant here?

music_assistant/providers/yousee/parsers.py
```diff
@@ -0,0 +1,219 @@
...
+        name=name,
+        version=version,
+        provider_mappings={
+            ProviderMapping(
+                item_id=str(album_obj["id"]),
+                provider_domain=provider.domain,
+                provider_instance=provider.instance_id,
+                audio_format=AudioFormat(
+                    content_type=ContentType.MP4,
+                    bit_rate=320000,
```
MarvinSchenkel: Bitrate is in kbps

music_assistant/providers/yousee/provider.py
```diff
@@ -0,0 +1,206 @@
...
+from __future__ import annotations
+
+from collections.abc import AsyncGenerator
+from typing import TYPE_CHECKING
+
+from music_assistant_models.errors import (
+    LoginFailed,
+)
+
+# It seems like this import is needed outside of TYPE_CHECKING, otherwise we run into runtime issues
```
MarvinSchenkel: Can you explain why?

MarvinSchenkel:
Few minor comments, but looks very good overall 👍 

As for your queries in a separate file. It already looks a lot better now everything else is split into multiple files. Only the big queries kill the flow of reading a bit. It's not an issue and comes down to personal preference, so I will it up to you to decide.

Question: What does the api expect you to pass as playbackContext? I would just try passing item_id and see what happens?

helgedk:
@MarvinSchenkel Regarding the playbackContext: Based on the JSON data I extracted earlier, the context string is a Base64 encoded string (like dXNlcjpwbGF5bGlzdDs0NDY2NDMyNQ==). When decoded, it often contains info about the source (e.g., 'user:playlist:44664325'). In my experience with this specific API, the backend is quite picky and expects this exact context to validate the stream request—it likely won't accept a simple item_id!?

MarvinSchenkel:
> @MarvinSchenkel Regarding the playbackContext: Based on the JSON data I extracted earlier, the context string is a Base64 encoded string (like dXNlcjpwbGF5bGlzdDs0NDY2NDMyNQ==). When decoded, it often contains info about the source (e.g., 'user:playlist:44664325'). In my experience with this specific API, the backend is quite picky and expects this exact context to validate the stream request—it likely won't accept a simple item_id!?

Is there any way to retrieve this playbackContext when requesting the stream url for example? I think your current approach also does not produce the expected base64 string? Any mention of the expected content in their api docs? Otherwise we have to play around with it. Decode a few different string and see if what they expect for different media types and build our own base64 string from the info we have.

On a different note, what does this actually do? Report play progress back to the api? From a MA perspective, this is not a mandatory part to implement. But of course, would be nice if we can get it to work.

math625f:
I too had another look at the playbackContext, and it seems I can omit it completely. I'm assuming YouSee Musik uses it internally to track how users "discover" songs. It's enough to just report the URL, which is what I believe is already contained within the streamdetails.path. 

I assumed it was mandatory, since it was always provided in requests sent from their official application.

As for your question regarding API docs, I haven't been able to find any official documentation, so I've manually tried to figure out what endpoints to use.

helgedk:
That's a great discovery @math625f! If the backend allows us to omit the playbackContext without any issues, it definitely makes the provider much cleaner and easier to maintain.
As long as we can fetch the stream URL and the metadata (like the 320kbps bitrate and lyrics we discussed), the reporting part is secondary. I agree with @MarvinSchenkel—let's keep it simple for the initial release so we can get it into the hands of the users for testing!

music_assistant/providers/yousee/constants.py
math625f: Yes - the service has support for streaming in `KBPS_320` and `KBPS_192`.
I found it accepts the following "qualities", mapped to the quality of the playback url it actually returns:

|Name|Returned|
-|-
|`KBPS_48`|`mp4-192kbps`|
|`KBPS_192`|`mp4-192kbps`|
|`KBPS_320`|`mp4-320kbps`|
|`FLAC_16`|`mp4-320kbps`|
|`FLAC_24`|`mp4-320kbps`|

Do you think we should make it possible for users to decide which quality they want to stream in?

helgedk: I agree that 320kbps should definitely be the default, as most MA users run this on a stable home connection and want the highest quality available.
However, having a setting to choose between the two real tiers (192 and 320) would be a great addition for those who might need to save bandwidth. Since your testing shows that requesting FLAC just returns the 320kbps stream anyway, we should probably just stick to offering the two actual qualities shown in the official app to keep it transparent for the users.

music_assistant/providers/yousee/provider.py
math625f: If I place the imports inside the `TYPE_CHECKING` block I get errors like:
```
2026-01-29 16:42:13.996 DEBUG (MainThread) [music_assistant.webserver] Handling command music/recommendations
2026-01-29 16:42:14.031 WARNING (MainThread) [music_assistant.music] Error while fetching recommendations from YouSee Musik: name 'RecommendationFolder' is not defined
  File "/path/to/server/music_assistant/controllers/music.py", line 1763, in _get_provider_recommendations
    return await provider.recommendations()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/controllers/cache.py", line 411, in wrapper
    type_hints = get_type_hints(func)
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 2499, in get_type_hints
    hints[name] = _eval_type(value, globalns, localns, type_params)
                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 474, in _eval_type
    return t._evaluate(globalns, localns, type_params, recursive_guard=recursive_guard)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 1081, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'RecommendationFolder' is not defined
```
Or
```
2026-01-29 16:42:30.615 WARNING (MainThread) [music_assistant] Exception in task Task-166 - target: <coroutine object MusicController._start_provider_sync.<locals>.run_sync at 0x7f3999dcf240>: name 'Album' is not defined
  File "/path/to/server/music_assistant/controllers/music.py", line 1793, in run_sync
    await provider.sync_library(media_type)
  File "/path/to/server/music_assistant/models/music_provider.py", line 689, in sync_library
    cur_db_ids = await self._sync_library_tracks()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/models/music_provider.py", line 983, in _sync_library_tracks
    async for prov_item in self.get_library_tracks():
    ...<37 lines>...
            )
  File "/path/to/server/music_assistant/providers/yousee/provider.py", line 92, in get_library_tracks
    async for track in self.library.get_tracks():
        yield track
  File "/path/to/server/music_assistant/providers/yousee/library.py", line 152, in get_tracks
    yield await parse_track(self.provider, item)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/providers/yousee/parsers.py", line 65, in parse_track
    album = await parse_album(provider, track_obj["album"])
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/providers/yousee/parsers.py", line 123, in parse_album
    return await provider.get_album(str(album_obj["id"]))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/controllers/cache.py", line 411, in wrapper
    type_hints = get_type_hints(func)
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 2499, in get_type_hints
    hints[name] = _eval_type(value, globalns, localns, type_params)
                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 474, in _eval_type
    return t._evaluate(globalns, localns, type_params, recursive_guard=recursive_guard)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 1081, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'Album' is not defined
2026-01-29 16:42:30.620 WARNING (MainThread) [music_assistant.music] Sync task for YouSee Musik/tracks completed with errors
  File "/path/to/server/music_assistant/controllers/music.py", line 1793, in run_sync
    await provider.sync_library(media_type)
  File "/path/to/server/music_assistant/models/music_provider.py", line 689, in sync_library
    cur_db_ids = await self._sync_library_tracks()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/models/music_provider.py", line 983, in _sync_library_tracks
    async for prov_item in self.get_library_tracks():
    ...<37 lines>...
            )
  File "/path/to/server/music_assistant/providers/yousee/provider.py", line 92, in get_library_tracks
    async for track in self.library.get_tracks():
        yield track
  File "/path/to/server/music_assistant/providers/yousee/library.py", line 152, in get_tracks
    yield await parse_track(self.provider, item)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/providers/yousee/parsers.py", line 65, in parse_track
    album = await parse_album(provider, track_obj["album"])
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/providers/yousee/parsers.py", line 123, in parse_album
    return await provider.get_album(str(album_obj["id"]))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/server/music_assistant/controllers/cache.py", line 411, in wrapper
    type_hints = get_type_hints(func)
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 2499, in get_type_hints
    hints[name] = _eval_type(value, globalns, localns, type_params)
                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 474, in _eval_type
    return t._evaluate(globalns, localns, type_params, recursive_guard=recursive_guard)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/typing.py", line 1081, in _evaluate
    eval(self.__forward_code__, globalns, localns),
    ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 1, in <module>
NameError: name 'Album' is not defined
```

I haven't had much experience with the `TYPE_CHECKING` feature in Python, so there could very well be something I'm missing.

math625f:
I noticed that the playbackContext returned from the API when just requesting a track directly by ID looks like `catalog:track;<id>`, so I changed the code to just use that always. I'm not sure we need to do it, but we might as well play nice with their API.

math625f:
I've implemented lyrics as well now, it seems to be working ( I can see the lyrics while the song is playing, I'm assuming that means it's working? )
Did I understand correctly that `lrc_lyrics` is for lyrics WITH timestamps, and `lyrics` is for lyrics WITHOUT timestamps?

music_assistant/providers/yousee/__init__.py
math625f: It would appear so, yes:
https://login.yousee.dk/.well-known/openid-configuration
I don't have much experience with OAuth, but you require a client_id - yes?

I suppose we could just grab the one the web application uses, but I don't know if that's standard?
We could also shoot them a message and ask if they'd want to provide an official client_id for us to use, although I'm not sure that's something they'd be interested in.

How do you usually handle OAuth?

MarvinSchenkel: Usually you'd request an application / client_id/secret and use that. However, since you say you don't have much experience and it looks like they don't support this 'officially' through their API, let's just leave it for now and implement it later as an enhancement (if possible)

music_assistant/providers/yousee/provider.py
```diff
@@ -0,0 +1,206 @@
...
+from music_assistant.providers.yousee.media import YouSeeMediaManager
+from music_assistant.providers.yousee.playlist import YouSeePlaylistManager
+from music_assistant.providers.yousee.recommendations import YouSeeRecommendationsManager
+from music_assistant.providers.yousee.streaming import YouSeeStreamingManager
+
+if TYPE_CHECKING:
+    from music_assistant_models.enums import (
+        MediaType,
+    )
+    from music_assistant_models.streamdetails import StreamDetails
```
MarvinSchenkel, line 48: This should work:
```suggestion
if TYPE_CHECKING:
    from music_assistant_models.enums import (
        MediaType,
    )
    from music_assistant_models.media_items import (
        Album,
        Artist,
        MediaItemType,
        Playlist,
        RecommendationFolder,
        SearchResults,
        Track,
    )
    from music_assistant_models.streamdetails import StreamDetails
```

MarvinSchenkel:
Looks great! Made a small suggestion for the `TYPE_CHECKING` stuff. There's one more thing, can you add the icons (icon.svg and icon_monochrome.png)? After that this should be good to merge :)

math625f:
Alright I've added the monochrome version, I had to change the svg a bit, but was able to just reuse the previous path as a mask instead to achieve a cutout effect.
I've also made the quality configurable in the provider settings.

MarvinSchenkel (approved):
Awesome job on this new provider 🎉 ! I'll get a new beta out today so people can play with it over the weekend 🙏

OzGav:
@math625f please create a PR against our beta docs repo. Look at other similar providers for the info which we typically include. Lyrics in this provider is unique so please cover a bit about that. Any questions let me know.

math625f:
> @math625f please create a PR against our beta docs repo. Look at other similar providers for the info which we typically include. Lyrics in this provider is unique so please cover a bit about that. Any questions let me know.

https://github.com/music-assistant/beta.music-assistant.io/pull/72

What would you like me to cover about the lyrics support? Do you mean how it's implemented, or just that it's there?

helgedk:
Huge thanks to @math625f for the YouSee integration and to the entire team for making this possible! I’ve just tested it on the new beta, and it works flawlessly with my HEOS speakers. Login was smooth, and even the synced lyrics are spot on. This is a huge Improvement for Danish users. Great job everyone!🤩

nicoleise:
Indeed thank you for this, @math625f 

I would like to ask, if it is possible to duplicate this provider for the "sibling-service" Telmore Musik? If so, it would be great for both users like me, and for Music Assistant to (in this case hopefully very easily) offer a large number of providers. 

I believe the services are essentially one and the same, differing only in branding. When you play a song in either, I'm convinced it's the same physical file being offered. I'm guessing there are different end points for at least authentication and possibly also different endpoints or aliases for the stream URLs themselves. But I think everything else is essentially identical. 

I did try to sign in to this provider using my Telmore Musik creditials, which fails authentication. Likely, each vendor (YouSee, Telmore,...) maintains their own user DB, to ensure that only active subscribers are allowed.

Is there anything I (as a user) can provide to you to help clarify if this is possible/easy? It sure seems like a low hanging fruit.


Thank you for your consideration, 
Nicolai

math625f:
> Indeed thank you for this, @math625f 
> 
> I would like to ask, if it is possible to duplicate this provider for the "sibling-service" Telmore Musik? If so, it would be great for both users like me, and for Music Assistant to (in this case hopefully very easily) offer a large number of providers. 
> 
> I believe the services are essentially one and the same, differing only in branding. When you play a song in either, I'm convinced it's the same physical file being offered. I'm guessing there are different end points for at least authentication and possibly also different endpoints or aliases for the stream URLs themselves. But I think everything else is essentially identical. 
> 
> I did try to sign in to this provider using my Telmore Musik creditials, which fails authentication. Likely, each vendor (YouSee, Telmore,...) maintains their own user DB, to ensure that only active subscribers are allowed.
> 
> Is there anything I (as a user) can provide to you to help clarify if this is possible/easy? It sure seems like a low hanging fruit.
> 
> 
> Thank you for your consideration, 
> Nicolai

Hi @nicoleise, I'm fairly certain you're correct, but I don't have access to credentials I can use to verify this fact. If they in fact do match exactly api-wise, it's a matter of changing the GraphQL endpoint.
The auth flow however is very flaky as I recall, and might not work for Telmore out-of-box.

nicoleise:
> Hi @nicoleise, I'm fairly certain you're correct, but I don't have access to credentials I can use to verify this fact. If they in fact do match exactly api-wise, it's a matter of changing the GraphQL endpoint. The auth flow however is very flaky as I recall, and might not work for Telmore out-of-box.

Is there anything I can do to assist in verifying? I obviously have credentials, but they are not really suited for sharing for two reasons: One, the username is the users phone number and two, the credentials also allow access to the control panel for the phone subscription itself including payment methods and such. 

I guess worst case I could "sponsor" a subscription for you to use, but for now, probably there's some simpler sanity check style thing, we could try? I'm not a developer but I can run a curl command or use developer mode in Chrome, for example, if any of those make sense?

(edit: @math625f in case GitHub doesn't ping you about the reply)

math625f:
> > Hi @nicoleise, I'm fairly certain you're correct, but I don't have access to credentials I can use to verify this fact. If they in fact do match exactly api-wise, it's a matter of changing the GraphQL endpoint. The auth flow however is very flaky as I recall, and might not work for Telmore out-of-box.
> 
> Is there anything I can do to assist in verifying? I obviously have credentials, but they are not really suited for sharing for two reasons: One, the username is the users phone number and two, the credentials also allow access to the control panel for the phone subscription itself including payment methods and such. 
> 
> I guess worst case I could "sponsor" a subscription for you to use, but for now, probably there's some simpler sanity check style thing, we could try? I'm not a developer but I can run a curl command or use developer mode in Chrome, for example, if any of those make sense?
> 
> (edit: @math625f in case GitHub doesn't ping you about the reply)

I'm sure we can figure something out, really I mostly need two bits of information: 

1. What is the GraphQL endpoint for Telmore,
2. and is the auth flow the same as for YouSee.

If you're at all comfortable reading, running, and altering a couple of lines of Python, you can verify these things yourself by pulling the codebase, otherwise I can try to look into it once I get back home to my pc in a couple of weeks.

nicoleise:
> 1. What is the GraphQL endpoint for Telmore,

https://graphql-1387.api.247e.com/graphql

Note that both YouSee Musik and Telmore Musik are essentially clients of a provider-aimed B2B company that offers the actual service. So the 1387 is Telmore Musiks "id" with that service (24/7 Entertainment) while YouSee Musik likely uses a different ID. 

> 2. and is the auth flow the same as for YouSee.

Auth flow **should** be identical, as both "services" are essentially side-ordered clients to the originating service. But as mentioned (since punching in my Telmore creditials into this Music Assistant provider didn't work), they validate users independently - or atleast against different endpoints. 

So I think most of the existing solution would work just fine. I don't have access to YouSee Musik and am not well enough versed in this specific discipline of coding to understand, if things are exactly the same. Instead, I focussed on providing insight into the auth flow in Telmore Musik. I imagine you'll immediately recognize it. 

I opened an incognito window and developer tools with preserve log enabled. I see this flow immediately after submitting on valid credentials: 

1) "internal-login" 

Endpoint: https://id.telmore.dk/internal-login
Method: POST
Payload ((censored by me)):
`{"session":"Tt6Ho+c8DCnf/s+UBvy+AZBUeJWqhdT/rWqASv/SSv0EMVA8fS9399erFxrnGAtQAJ4qHZ2ooYe4/GY0VuqVOZcgwL7TWv079yJmhP5j/cWqjPrn6Fqm4GvCc4xQYQ+mUjZD4bYageN((somecharacterschanged))Hq/dk2Fi49uv4/5elp1HPxluAA8pqBBJL7VcYwMRXtOrHDbHnMBW6bheHlpe7waWuYd+ugwb8DWOqL06Siu25dkvTlrqPkGKweqnP+1Sbb3TTA8iMYrECw8T1XHkjNU6JACwWC+e58ZmwBU987qW5QcZkZt/0mrIvzF4hVX+nhzw((somecharacterschanged))FBw6iU2Q7/Endrie4sxXYHWyzpDqWmD1DYJhfnYe0roKUbZbM2LyxJ2fyOS9o7UnK((somecharacterschanged))Qrd5l7zc+y4BvdWOsRr+aZDxtQRdRoM5ZJ82W1705HoU9F7vinOWuV12C1lBrujgXSsU/Y4BK5qZ2OCl4ZA8p1f5eFAkDTgKR5RaUIea5ovHFAJNHEI9MHxMd3aWUyhtZu8lT0VlosR0Y0D+iT3rPz4ytZCBdsnP1sb7JzMOIDwd01FtrEZccckMMtMlVAs10AW7H9Ngaxf1QpTYUmuPDhRo+xneCVvmpUJH3y8D/BRZg5JTeTsetKB00oyXoY=","username":"((my-username))","password":"((my-password))"}`

2) "Delegated login response" seems to then set a cookie based on server side authentication, and at the same time point to the correct 24/7 Entertainment environment:

Endpoint: https://musik.telmore.dk/delegatedloginresponse
Method: GET
Payload: 
`https://musik.telmore.dk/delegatedloginresponse?state=%7B%22installationId%22%3A%22bb3e2c56-5a66-4166-b92a-f1e4c2d7ad60%22%2C%22graphqlEnvironmentName%22%3A%22prod%22%2C%22apiPortalId%22%3A%221387%22%7D&code=rjfxFX((Anumberofcharacterschanged))bTgMrYsnAETIixmR7GGetYMvEOg4aYJyE%2FkrJZqcsoWH4d45LHXaD7%2BU((Anumberofcharacterschanged))W0WtpcQl7DZLird%2FniG7DWcEJounDim2%2FhvVTWNtfFBNF%2BuT9xAbKRVCNy2zDzc((Anumberofcharacterschanged))m98AJSZHDpVMOiKfbryf0zbLinKJ6UaZi8%2BwJPtgEIY0MfcTSA17hbcCmnRPz7FHVIjVEW42sdLBly5Sy6al4ff2CtXbROaGDx%2BEzgd6kNVONQJKcQxWfX7BeK851WuRBfm%2B1f9HAzKif%2FMl73qlzGOw3sku1niWqHI((Anumberofcharacterschanged))xCwtzCyPmbD%2B%2FWxpRrX27wmOoSwRWBXl%2FY%2FsU%2BBRwkq8hHDGxCKifKaKgo8YDTJR%2FbSow%3D%3D`

Note that the first bit translates to: 
`{"installationId":"bb3e2c56-5a66-4166-b92a-f1e4c2d7ad60","graphqlEnvironmentName":"prod","apiPortalId":"1387"}`

Here a cookie (TS82......) is set/modified as mentioned.


3) Graphql is then used directly to the backend service from this point onwards:

Endpoint: https://graphql-1387.api.247e.com/graphql
Method: POST
Payload: 
```
{operationName: "userId2", variables: {includeProfiles: false},…}
operationName
: 
"userId2"
query
: 
"query userId2($includeProfiles: Boolean!) {\n  me {\n    user {\n      id\n      username\n      useInviteFlow\n      trackingId\n      profiles @include(if: $includeProfiles) {\n        profiles {\n          items {\n            ...ProfileFragment\n            __typename\n          }\n          __typename\n        }\n        settings {\n          profilesNumberLimit\n          defaultColors\n          __typename\n        }\n        __typename\n      }\n      __typename\n    }\n    __typename\n  }\n}\n\nfragment ProfileFragment on Profile {\n  color\n  current\n  description\n  id\n  title\n  master\n  __typename\n}"
variables
: 
{includeProfiles: false}
includeProfiles
: 
false
```

Response to the above: 
```
{
    "data": {
        "me": {
            "user": {
                "id": "89((somedigitschanged))77",
                "username": null,
                "useInviteFlow": false,
                "trackingId": "7jCE7e5O((somecharacterschanged))52ZTc=",
                "__typename": "User"
            },
            "__typename": "Me"
        }
    }
}
```
The username/id in this above response is not anything I recognize. I'm censoring these things as they are publicly posted on the internet. 

From then on, it seems to all be "normal" graphql like listing my own playlists, that I may edit, fetching content, etc.

I hope this helps, and please don't be afraid to ask if I can do further, @math625f . :) Nicolai

tobsorbe-droid:
@math625f @nicoleise I have a live Telmore subscription and have been poking at 1387 for a few days, so here are some measurements that might save you a round of testing.

### Telmore is 1387 and it works
Confirmed against a real subscription: `https://graphql-1387.api.247e.com/graphql` answers normally. Top-level query fields are `catalog`, `playlists`, `search`, `promotions`, `charts`, `liveRadios`, `mixRadios`, `layout`, `authorize`, `me`, `playback`, `system` - and the bearer format is the same `Issuer=247e&ExpiresOn=…` string. So @math625f's guess that swapping the endpoint would mostly suffice looks right.

The tenant also shows up in every cover URL the API returns:
`https://aartwork.cdn.247e.com/1387/covers/<id>/320x320` - and those images are fetchable without auth once you have the URL. You still need an account to get a cover URL though.

**Unverified**: I've seen 1458 quoted as YouSee's tenant, and that endpoint does exist (403 rather than 404), but I have no YouSee account and haven't confirmed it.

### Heads-up on the quality setting: FLAC silently falls back
`StreamQuality` in the schema is:

`KBPS_48, KBPS_192, KBPS_320, FLAC_16, FLAC_24`

That reads like lossless is available, but on Telmore it isn't - the higher values are accepted and then quietly downgraded. Measured on one track across all five values, then re-checked `KBPS_320` and `FLAC_24` on two more tracks to make sure the FLAC fallback wasn't track-specific:

- **KBPS_48**: mp4-192kbps
- **KBPS_192**: mp4-192kbps
- **KBPS_320**: mp4-320kbps
- **FLAC_16**: mp4-320kbps
- **FLAC_24**: mp4-320kbps

No error, no warning - just a 320 kbps mp4 manifest. So 192 appears to be the floor and 320 the ceiling.

Worth knowing before exposing FLAC in the provider's quality options: on Telmore it would be a setting that appears to do something and doesn't. Whether YouSee's tier behaves the same, I can't say. Might be worth someone checking, since the enum is presumably shared across all 24-7 tenants and the difference would be entitlement, not API.

For reference the returned URL is a signed CloudFront HLS manifest - `ahls.cdn.247e.com/ahlsaes/assets/mp4-320kbps/…/index.m3u8`, with `Policy` / `Signature` / `Key-Pair-Id` query params. Fetching it returns a normal `#EXTM3U` playlist.

The playlist contains `#EXT-X-KEY`, so segments are AES-encrypted - consistent with the `ahlsaes` path segment.

### An auth route that needs no username or password
This might be interesting given the auth reliability wrinkles in the thread. If the user already has the service in Sonos, you can skip credentials entirely.

The Sonos music service is `auth_type = AppLink`, and **the authToken Sonos stores is base64 of the same signed key/value string the web player uses - and it authenticates against GraphQL**. Tested: decoded token as `Authorization: Bearer …` → HTTP 200 with real search results. The visible difference from a web token is `clientId=nuumusicapp-sonos-telmore` instead of `-web-telmore`.

```python
import base64
from soco.music_services import MusicService
from soco.music_services.token_store import JsonFileTokenStore

store = JsonFileTokenStore("/persistent/token_store.json")
ms = MusicService("TELMORE Musik", token_store=store)

url = ms.begin_authentication()
# real Telmore login form, in the user's own browser

ms.complete_authentication()
# after approval

sc = ms.soap_client
auth, _ = sc.token_store.load_token_pair(ms.service_id, sc._household_id)
bearer = base64.b64decode(auth + "=" * (-len(auth) % 4)).decode()
```

Renewal needs no user interaction at all:

```python
res = sc.call("refreshAuthToken", [])   # -> {"authToken": …, "privateKey": …}
sc.token_store.save_token_pair(
    ms.service_id, sc._household_id, (res["authToken"], res["privateKey"])
)
sc._cached_soap_header = None
```

How long the link survives I genuinely don't know yet - mine has only been running a couple of days. The reasoning is that it's the same credential Sonos itself maintains, but that's an expectation, not a measurement.

Two traps that cost me three re-logins, in case anyone tries it:

1. SoCo's default token store lives in the user config dir (`appdirs.user_config_dir`). In a container that's wiped on every image rebuild, so the link dies silently at the next update. Point it somewhere persistent.
2. `begin_authentication()` must run on a **MusicService** built with the same token store, or `complete_authentication()` writes the pair elsewhere and it looks like the login just didn't work.

`complete_authentication()` raises until the user has approved, so you can just call it on a loop and close the window on the first success. (In my case it sits behind an HTTP route that returns 409 until then, but that's my own wrapper, not theirs.)

### Two smaller notes on the web token path
Since @nicoleise documented the login flow - the piece after `delegatedloginresponse` is `POST /api/token` with `{"refresh_token": "<32 hex>"}`, returning `{status, tokenResult:{access_token, refresh_token, expires_in}}`. Tested: it works with `Content-Type` alone; `X-App-Version` and `X-Installation-Id` are sent by the web client but aren't currently required.

In my testing **the refresh token did not rotate** - same value returned every time, so it can be stored once. (One evening of testing, not weeks.) It lives in `localStorage` under a key starting with `refreshToken_`; note the suffix is not the `user_id` from the token, so match on the prefix.

The access token isn't opaque either - `Issuer=247e&ExpiresOn=<unix>&clientId=…&RSASHA256=…` - so you can refresh on a timer rather than waiting for a 401. `ExpiresOn` has been about 24 h out every time I've looked.

### One feature suggestion
`Track.similarTracks` maps straight onto `get_similar_tracks()` / `ProviderFeature.SIMILAR_TRACKS`, and `Artist.similarArtists` onto `get_similar_artists()`. Measured across six seeds (20 requested each): 12-20 unique artists per result. `Artist.tracks` looks like a fit for `get_artist_toptracks()`, and `Queries.charts` / `promotions` for `get_recommendations()`.

Also: covers come back at 128x128, but rewriting the trailing `/128x128` works - I verified `/320x320` and `/640x640` both return images at exactly those dimensions.

All of this is read out of the vendor's own web and Sonos clients, so it can change without notice.

math625f:
Hi @tobsorbe-droid, I'm glad someone was able to confirm that the Graphql API shares the same shape for both YouSee and Telmore. 

As for your point on quality, the enum represents the accepted values by the Graphql API, but the YouSee provider only accepts 320 kbps and 192 kbps in its configuration.

I'm really not sure I feel comfortable using the credentials stored in another providers Auth Store, I'd much rather we used a flow designed specifically for Telmore Musik.

I'm fairly certain that it _won't_ work out-of-the-box, I took a quick look, and I've tried to recreate the flow that Telmore uses. I am unable to see what happens next, since I don't have valid credentials. I've written a small script for someone with valid credentials to run, typing in their credentials, can either of you verify if it works?

If it _doesn't_ work, I'd really appreciate if I could take a look at an **obfuscated** print of what was returned after succesfully logging in. 

I'll just ping you as well @nicoleise, since either of you can run the script for me :)

```py
import getpass
import re
from urllib.parse import parse_qs, urlparse

import requests

DELEGATED_LOGIN = "https://musik.telmore.dk/api/delegatedlogin"
INTERNAL_LOGIN = "https://id.telmore.dk/internal-login"


def main() -> None:
    username = input("Telmore username: ").strip()
    password = getpass.getpass("Telmore password: ")

    session = requests.Session()

    resp = session.get(DELEGATED_LOGIN, allow_redirects=False)
    location = resp.headers.get("Location", "")
    login_session = parse_qs(urlparse(location).query).get("session", [None])[0]
    if not login_session:
        raise SystemExit(
            f"No session in redirect (status={resp.status_code}, location={location!r})"
        )
    print("session:", login_session)

    resp = session.post(
        INTERNAL_LOGIN,
        params={"session": login_session},
        json={"session": login_session, "username": username, "password": password},
    )
    if resp.status_code != 200:
        raise SystemExit(
            f"internal-login failed (status={resp.status_code}): {resp.text}"
        )
    login_result = resp.json()
    print("internal-login response:", login_result)
    login_url = login_result.get("url")
    if not login_url:
        raise SystemExit("No 'url' in internal-login response")

    page_response = session.get(login_url)
    page = page_response.text
    access = re.search(r'localStorage.setItem\("accesstoken", "([^"]+)"', page)
    refresh = re.search(r'localStorage.setItem\("refreshtoken", "([^"]+)"', page)
    if not access or not refresh:
        print("REQUESTED URL:", login_url)
        print("FINAL URL:", page_response.url)
        print("REDIRECT CHAIN:")
        for hop in [*page_response.history, page_response]:
            print(f"  {hop.status_code} {hop.url}")
        print(
            "PAGE CONTENT:\n-----------------------\n",
            page,
            "\n-----------------------",
        )
        print("Cookies:", page_response.cookies.get_dict())
        raise SystemExit("Tokens not found on the final page")

    print("\nSUCCESS")
    print("accesstoken:", access.group(1))
    print("refreshtoken:", refresh.group(1))


if __name__ == "__main__":
    main()

```

tobsorbe-droid:
@math625f Ran it - **your flow works**. Only the last step looks in the wrong place.

Both requests succeed:
```
1) GET  musik.telmore.dk/api/delegatedlogin   -> 302
   redirect: id.telmore.dk, session = 642 chars
2) POST id.telmore.dk/internal-login          -> 200
   response fields: ['url']          url = 880 chars
3) GET  <that url>                            -> 200   (913 bytes)
```
The final page never touches `localStorage`. It's a tiny relay that hands the tokens back to whoever opened it:
```js
let recipient = window.opener || window.parent;
if (recipient) {
    const origin = "*"; // todo: get correct origin
    recipient.postMessage({
        type: "tokens",
        accessToken: "<375 chars>",
        refreshToken: "<32 chars>",
    }, origin);
}
```
Two things there, and either alone would have broken the match:
1. It's **postMessage, not localStorage** - the web app receives this message and writes the values to localStorage itself afterwards. That's why I could find `refreshToken_<id>` in localStorage earlier, but it isn't on the login page.
2. The keys are **camelCase** - `accessToken` / `refreshToken`, not `accesstoken` / `refreshtoken`.
So step 3 just needs to parse the postMessage payload instead:
```python
acc = re.search(r'accessToken:\s*"([^"]+)"', page)
ref = re.search(r'refreshToken:\s*"([^"]+)"', page)
```
The lengths match what I measured independently: the 375-char access token is the signed `Issuer=247e&ExpiresOn=<unix>&clientId=...&RSASHA256=...` string, and the 32-char refresh token is the hex value that feeds `POST /api/token`.
### Renewal, once you have them
```http
POST https://musik.telmore.dk/api/token
Content-Type: application/json
{"refresh_token": "<32 hex chars>"}
-> {"status": 0, "tokenResult": {"access_token": "...", "refresh_token": "...", "expires_in": ...}}
```
Tested: it works with `Content-Type` alone - `X-App-Version` and `X-Installation-Id` are sent by the web client but aren't currently required. `status: 0` is success; I've only seen `4` when unauthenticated. In my testing the refresh token did not rotate - the same value came back on every refresh, so it can be stored once. (One evening of testing, not weeks.)

The access token carries its own expiry in `ExpiresOn`, roughly 24 h out, so you can refresh on a timer rather than waiting for a 401.
### On the auth store
My setup is a bridge that has to have Sonos anyway, so it cost nothing there; for a general provider a dedicated flow is clearly the right call.
### On quality
My point was only that the enum advertises `FLAC_16`/`FLAC_24` and Telmore silently downgrades them to 320 kbps rather than erroring, so widening the config would ship a setting that appears to do something and doesn't. If the provider already caps at 192/320, there's nothing to do.
### One aside worth flagging
They post to `origin: "*"` and left the comment `// todo: get correct origin` in shipped code. Any page that opens that window can read the tokens.

Happy to run anything else against a live Telmore account.

math625f:
Thanks a lot @tobsorbe-droid. Just to be completely sure, I've modified the dummy script with your recommended regex changes. Can you verify the full flow works now?

```py
import getpass
import re
from urllib.parse import parse_qs, urlparse

import requests

DELEGATED_LOGIN = "https://musik.telmore.dk/api/delegatedlogin"
INTERNAL_LOGIN = "https://id.telmore.dk/internal-login"


def main() -> None:
    username = input("Telmore username: ").strip()
    password = getpass.getpass("Telmore password: ")

    session = requests.Session()

    resp = session.get(DELEGATED_LOGIN, allow_redirects=False)
    location = resp.headers.get("Location", "")
    login_session = parse_qs(urlparse(location).query).get("session", [None])[0]
    if not login_session:
        raise SystemExit(
            f"No session in redirect (status={resp.status_code}, location={location!r})"
        )
    print("session:", login_session)

    resp = session.post(
        INTERNAL_LOGIN,
        params={"session": login_session},
        json={"session": login_session, "username": username, "password": password},
    )
    if resp.status_code != 200:
        raise SystemExit(
            f"internal-login failed (status={resp.status_code}): {resp.text}"
        )
    login_result = resp.json()
    print("internal-login response:", login_result)
    login_url = login_result.get("url")
    if not login_url:
        raise SystemExit("No 'url' in internal-login response")

    page_response = session.get(login_url)
    page = page_response.text
    access = re.search(r'accessToken:\s*"([^"]+)"', page)
    refresh = re.search(r'refreshToken:\s*"([^"]+)"', page)
    if not access or not refresh:
        print("REQUESTED URL:", login_url)
        print("FINAL URL:", page_response.url)
        print("REDIRECT CHAIN:")
        for hop in [*page_response.history, page_response]:
            print(f"  {hop.status_code} {hop.url}")
        print(
            "PAGE CONTENT:\n-----------------------\n",
            page,
            "\n-----------------------",
        )
        print("Cookies:", page_response.cookies.get_dict())
        raise SystemExit("Tokens not found on the final page")

    print("\nSUCCESS")
    print("accesstoken:", access.group(1))
    print("refreshtoken:", refresh.group(1))


if __name__ == "__main__":
    main()
```

tobsorbe-droid:
@math625f Ran your updated script - **it works**.
```
1) GET  /api/delegatedlogin        -> 302, session 626 chars
2) POST /internal-login            -> 200
   response fields: ['url'], url = 910 chars
3) GET  <that url>                 -> 200 (923 bytes)
SUCCESS - both regexes matched on a fresh login.
  accessToken:  385 chars  (ExpiresOn=True, RSASHA256=True)
  refreshToken: 32 chars   (hex-only=True)
```
Both matched on the first try, and the values are the right shape: the access token is the signed `Issuer=247e&ExpiresOn=<unix>&clientId=...&RSASHA256=...` string, and the refresh token is 32 hex chars, which is what `POST /api/token` takes.

One thing worth knowing: the lengths drift between logins. Session was 642 chars in my earlier run and 626 now, access token 375 then and 385 now.

Please also note: the script prints both tokens in full

math625f:
@tobsorbe-droid Alright, I've made a minimal-effort implementation on the basic assumption that everything works the same, aside from authentication. Are you able to clone my fork, checkout the `telmore-musik-provider` branch, and verify you can log in, and everything works?

By everything I mean

- Browsing music
- Searching albums, playlists, artists, tracks
- Viewing personal recommendations (recent tracks, top tracks, weekly discoveries, your mix, etc)
- Viewing tracks, albums, playlists, and artists added to "library"
  - Adding new items to library
  - Removing items from library
- Editing playlists
- Seeing similar tracks
- Browsing artist toptracks and albums
- Lyrics showing up for songs
- Streaming working, played tracks (from MA) showing up in recently played inside Telmore music (this is needed for proper recommendations)

As well as basic stuff such as titles and covers showing up correctly.

Edit: Forgot to link my fork :sweat_smile: 
https://github.com/math625f/music-assistant-server/tree/telmore-musik-provider

tobsorbe-droid:
@math625f Short version: everything on your list works except streaming, and that one is a config-key fix.

The setup flow works - form step with username/password, submit, `handle_async_init` runs the real login, provider comes up available. Browse and search through MA's music controller work too:

```
telmore://            6 folders
telmore://artists     16
telmore://albums      10
telmore://playlists   11
music/search          5 each of artists/albums/tracks/playlists, covers on all
```
---

## Bug 1 - the refresh flow posts form data, the endpoint wants JSON

`auth_manager.py`, the refresh branch:

```python
data={"refresh_token": self._refresh_token}     # -> HTTP 415
json={"refresh_token": self._refresh_token}     # -> HTTP 200
```

Same token, same endpoint, back to back. With `data=` the response is `application/problem+json` with status 415, so `refresh_result.get("status", 4) == 0` is False and it silently falls through to the full username/password login. It works, but the refresh path never actually fires - you re-login every time the token expires (~24 h).

---

## Bug 2 - the quality config key is the yousee one

`telmore/constants.py` registers `CONF_QUALITY = "telmore_quality"`, and the provider builds its `ConfigEntry` from that. But the inherited yousee code reads yousee's constant:

```python
# yousee/streaming.py:15
from music_assistant.providers.yousee.constants import CONF_QUALITY
# yousee/streaming.py:46
"quality": f"KBPS_{self.provider.config.get_value(CONF_QUALITY)}"
```

On a Telmore provider that lookup returns `None` (MA's `get_value` returns the `default` arg on `KeyError`, and there is no default passed), so the query sends `quality: "KBPS_None"` and the API answers 400.

```
config.get_value('yousee_quality') -> None
get_stream_details(...)            -> ClientResponseError 400
```

### It kills playback in a real MA

Hitting `/preview` on a configured Telmore provider, unpatched:

```
serve_preview_stream -> get_preview_stream -> provider.get_stream_details
  -> yousee/streaming.py:49 -> telmore/api_client.py:69
  -> ClientResponseError: 400, Bad Request
```

The browser just gets a closed connection. After correcting the key, same endpoint, same track:

```
ffprobe:        codec_name=aac  sample_rate=44100  channels=2
25 s decoded:   3528078 bytes of PCM (exactly 44100 * 2ch * 2B * 25s)
volumedetect:   mean_volume -14.7 dB, max_volume 0.0 dB
```

### Same class of bug in two more places

Here it silently produces empty metadata instead of an error:

```python
# yousee/parsers.py:48 and :140
bit_rate=try_parse_int(provider.config.get_value(CONF_QUALITY))
```

Since the subclasses inherit these methods wholesale, the cleanest fix is probably to stop reading a module-level constant and take the key from the provider (a class attribute on the manager, or `provider.CONF_QUALITY`), so the yousee subclass tree stays portable. Anything that reuses yousee's managers will hit this again otherwise.

---

## Bug 3 - removing an owned playlist silently does nothing

`library.py` `remove_item()` maps every media type onto the favorites mutation:

```graphql
favorites { removePlaylist(id: $id) { ok } }
```

For a playlist the user is merely following, that is right. For one they own - anything created through `playlists.create`, which is exactly what MA's own `create_playlist` makes - it is the wrong mutation. It returns `ok: true` and the playlist stays. I only caught it because my test playlist was still in browse afterwards.

Deleting an owned playlist needs the other mutation, which exists on 1387:

```graphql
playlists { delete(id: $id) { ok } }   # -> ok: true, playlist gone
```

Since this is in the shared yousee library manager, **yousee has it too**. Probably wants a branch on whether the playlist is owned (`playlists.delete`) or followed (`favorites.removePlaylist`).

---

## Your checklist

| Item | Result |
| --- | --- |
| Browsing music | same as yousee (neither declares `ProviderFeature.BROWSE`) |
| Search - artists | 5 |
| Search - albums | 10 |
| Search - tracks | 10 |
| Search - playlists | 10 |
| Personal recommendations | all 10 rows, 20-50 items each |
| Library - artists/albums/tracks/playlists | all four generators return |
| Adding to library | `library_add` -> True, verified present afterwards |
| Removing from library | `library_remove` -> True, verified gone (tracks) |
| Editing playlists | create -> add track -> remove track all OK; **deleting the playlist does not**, see bug 3 |
| Similar tracks | 10 |
| Artist toptracks | 25 |
| Artist albums | 38 |
| Lyrics | 1004 chars plain text (`lyrics_lrc` empty on that track) |
| Streaming | **broken**, see bug 2 - works once the key is fixed |
| reportPlayback | `ok: true` |
| Titles, covers, durations | correct throughout, covers on every result type |

Two things I got wrong at first and want to flag so you don't chase them: `get_recommendations()` returns rows with deliberately empty items (the mixin does `dataclasses.replace(folder, items=UniqueList())`), and the content comes from `get_recommendation_items(item_id)`. Fetched that way, all ten rows are populated. The row ids resolve fine on portal 1387 and the inline fragments match - `AlbumsRecommendation` and `TracksRecommendation` are the concrete typenames there too.

One real limit on 1387: `tracks(first: 50)` caps out at 30 items for the discover rows - fewer than asked for.

---

## On quality, since it came up earlier in the thread

Bug 2 makes the whole 192/320 discussion moot on Telmore: the setting the user picks never reaches the API, because the value is read from a key that does not exist. Fixing the key is what makes that config do anything at all.

I measured every value the 1387 schema advertises:

```
KBPS_48    ->  mp4-192kbps
KBPS_192   ->  mp4-192kbps
KBPS_320   ->  mp4-320kbps
FLAC_16    ->  mp4-320kbps
FLAC_24    ->  mp4-320kbps
```

So the silent downgrade I mentioned for lossless holds on Telmore's own portal too, not just YouSee, and `KBPS_48` quietly rounds up rather than erroring.

---

## Two small things

`reportPlayback` rejects `playedSeconds: 0` with BadRequest. In normal playback that never happens, but `report_playback` computes

```python
min(utc_timestamp() - streamdetails.data["start_ts"], streamdetails.seconds_streamed)
```

so a track skipped inside the first second reports 0 and the mutation fails. Worth a guard, since the only symptom is a logged warning and a gap in the listening history that recommendations feed on.

`auth_token()` returns `None` on failure, and `post_graphql` then interpolates it into the header as the literal string `"Bearer None"`. The 401 comes back and you get `LoginFailed`, which is the right outcome, but the traceback points at the GraphQL call rather than the login. Raising in `auth_token()` would make that easier to read.

math625f:
@tobsorbe-droid I've added your suggested changes for bugs 1 and 2, can you try again?

As for your other reports, I'd like to decide how much to preserve from YouSee before I start fixing things that affects both services.

tobsorbe-droid:
@math625f Retested - both are fixed.

## Bug 1 - refresh

Tested with the refresh token set and the credentials removed, so there is nothing to silently fall back to:

```
auth_token()  ->  OK in 0.41 s, 373-char token, expires in 24.0 h
```

Before the fix that same call returned `None`.

## Bug 2 - quality key

```
provider.playback_quality      '320'
get_stream_details(...)        format=mp4  bitrate=320  stream_type=hls
GET playback url               HTTP 200, application/vnd.apple.mpegurl, starts with #EXTM3U
```

And the two `parsers.py` sites now populate bitrate instead of leaving it empty:

```
Midt om natten                        bitrate=320
Midt om natten (Live - 2011 - Remas   bitrate=320
Midt Om Natten (Live at Danmarks T)   bitrate=320
```

Streaming works.

Happy to retest whenever you want, including the parts that touch both services.

PR #3045, opened 2026-01-28
Don't force reload on all config changes

apophisnow:
CoreController.update_config now stores the updated config and only reloads when at least one changed entry has requires_reload set (default is now False in the models package). Entries read dynamically should explicitly set requires_reload=False to skip the reload.

Provider.update_config is cleaned up to store config and handle log level changes separately, but still reloads on any other value change since provider reloads are lightweight.

Webserver base_url and allow_self_registration are marked requires_reload=False and refactored to be read dynamically: base_url is now read from self.config on each property access, and allow_self_registration is read by the HA OAuth provider from webserver config instead of being baked into LoginProviderConfig.

### DIFF
--- a/music_assistant/constants.py
+++ b/music_assistant/constants.py
@@ -159,6 +159,7 @@ CONF_ENTRY_LOG_LEVEL = ConfigEntry(
     ],
     default_value="GLOBAL",
     category="advanced",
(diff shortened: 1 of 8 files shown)


marcelveldt (approved):
Looks good, I have no comments

PR #3046, opened 2026-01-29
Radio Paradise. Switch to simple API

OzGav:
Steam metadata updates were failing as the endpoint used was returning an empty reponse. They have a simple now playing endpoint as well so this PR switches to that. As a result, all of the rich metadata code was redundant so has been removed.  I will monitor and if they fix the block_data endpoint then I will reimplement with a fallback option.

Also I realised I hadn't changed the provider to use the callback for metadata updates so that is fixed as well.


MarvinSchenkel (approved):
Works great @OzGav !

PR #3056, opened 2026-01-30
Add auto discovery to HEOS

Tommatheussen:
As mentioned in the initial HEOS PR, discovery was something still be done. With this PR, it's possible to do these things:

- Still manually configure the HEOS controller, useful to force a device to be used as controller. If manual IP is configured, mDNS does nothing.

- When mDNS players are found, take the first one that comes in and use that to connect to. If this device does have the right properties (supported firmware, wired connection), it will connect to 1 of the preferred hosts instead.

- When a player change event comes in from HEOS, run the player discovery again. Instead of unregistering a player when it is not found anymore, we mark it as unavailable.

### DIFF
--- a/music_assistant/providers/heos/__init__.py
+++ b/music_assistant/providers/heos/__init__.py
@@ -49,7 +49,7 @@ async def get_config_entries(
             key=CONF_IP_ADDRESS,
             type=ConfigEntryType.STRING,
             label="Main controller hostname or IP address.",
-            required=True,
+            required=False,
             description="Hostname or IP address of the HEOS device "
             "to be used as the main controller. It is recommended to use a "
             "wired device as the main controller.",
--- a/music_assistant/providers/heos/__init__.py
+++ b/music_assistant/providers/heos/__init__.py
@@ -53,5 +53,6 @@ async def get_config_entries(
             description="Hostname or IP address of the HEOS device "
             "to be used as the main controller. It is recommended to use a "
             "wired device as the main controller.",
+            category="advanced",
         ),
     )
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -75,7 +75,7 @@ class HeosPlayer(Player):
         await self.build_source_list()
 
     def set_device_info(self) -> None:
-        """Set all player static attributes."""
+        """Set all device info attributes."""
         # Extract manufacturer and model from device model string, if available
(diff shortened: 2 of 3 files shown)


OzGav:
Good news! Please PR the beta docs as well.

music_assistant/providers/heos/player.py
```diff
@@ -91,6 +90,8 @@ def set_static_attributes(self) -> None:
         self._attr_available = self._device.available
         self._attr_name = self._device.name
 
+        self.update_state()
+
```
marcelveldt:
```suggestion
```
I would not do this as it messes up with the init flow

music_assistant/providers/heos/provider.py
```diff
@@ -111,38 +125,72 @@ def input_source_list(self) -> list[MediaItem]:
...
                 player_id = str(device.player_id)
-                if player_id in already_registered:
-                    continue  # already registered
-                # ignore disabled players in discovery
+                if player := self.mass.players.get(player_id):
+                    self.logger.debug(
+                        "Updating existing HEOS player: %s (%s)", device.name, player_id
+                    )
+                    # Update properties such as name or availability
+                    cast("HeosPlayer", player).set_static_attributes()
```
marcelveldt: so if these values can actually change, they are not static are they ?

Tommatheussen: True, most values _should_ rarely change (like name and device_info), adjusted the updating call now

music_assistant/providers/heos/__init__.py
```diff
@@ -49,7 +49,7 @@ async def get_config_entries(
             key=CONF_IP_ADDRESS,
             type=ConfigEntryType.STRING,
             label="Main controller hostname or IP address.",
-            required=True,
+            required=False,
             description="Hostname or IP address of the HEOS device "
```
MarvinSchenkel, line 53: Just a thought, should we move this to the advanced category (`category='advanced'`)?

MarvinSchenkel: Besides that, all looks good to me 👍

Tommatheussen: Makes sense, this is only if you want to force a device to be used, not something that is generally needed, will adjust

MarvinSchenkel (approved):
Great work @Tommatheussen 👏

PR #3063, opened 2026-01-31
Add album year from Deezer release_date

sfortis:
## Summary
- Parse `release_date` from Deezer API and populate album `year` field
- Uses safe access with `getattr()` for backwards compatibility

image

## Test plan
- [x] Tested locally on dev server
- [x] Verified year is saved to database
- [x] Verified year displays in UI for provider albums


MarvinSchenkel (approved):
Thanks @sfortis , looks good to me

PR #3064, opened 2026-01-31
Add PlayerOptions to backend and MusicCast

fmunkes:
This PR adds additional player options for player providers to core.
For example, for MusicCast we can enable a different sound mode on an AV receiver (5.1, stereo etc.) or adjust bass and treble values. I only add the backend capability here, which I did test. Before I continue I'd like to get feedback if the approach taken is acceptable.
The next step for me would be to adjust the HA integration, such that these options are exposed to HA. I'm not sure if I have the ability though to adjust the MA frontend for it. Having it in HA though would give some value too I think.

Necessary models PR: https://github.com/music-assistant/models/pull/163
Feature Request: https://github.com/orgs/music-assistant/discussions/4856#discussioncomment-15648173


music_assistant/models/player.py
```diff
@@ -782,6 +818,24 @@ def source_list(self) -> UniqueList[PlayerSource]:
...
+    def sound_mode_list(self) -> UniqueList[PlayerSoundMode]:
+        """Return available PlayerSoundModes for Player."""
+        return UniqueList(self._attr_sound_mode_list)
+
+    @cached_property
+    @final
+    def player_option_list(self) -> UniqueList[PlayerOption]:
+        """Return all PlayerOptions for Player."""
+        return UniqueList(self._attr_player_option_list)
+
```
marcelveldt: these need to be in the top without the final
the pattern is that we allow any provider implementation to either use the _attr attributes or the public properties, following the Home Assistant entity model/pattern

the final properties and methods we have at the bottom are ONLY for things we override in core

marcelveldt (approved):
Nice work @fmunkes !

PR #3065, opened 2026-01-31
Fix track import with multiple artists and mixed separators

### LINKED ISSUE #4852: Artist names truncated or inconsistently indexed for multi-artist credits

2.7.5

Home Assistant OS Addon

### Carefully read the Troubleshooting FAQ and confirm that

- [x] I have fixed any errors in the logs that relate to tags
- [x] I have disabled tool such as AdGuard, Pi-hole or pfSense and retested

### Have you reviewed the [Open](https://github.com/music-assistant/support/issues) and [Closed](https://github.com/music-assistant/support/issues?q=is%3Aissue%20state%3Aclosed) Issues and [Discussions](https://github.com/orgs/music-assistant/discussions) to resolve this yourself?

- [x] Yes

## Issue summary

In Music Assistant, the artist **Nicholas Krgovich** is indexed and displayed in search results only as **“Krgovich”**, rather than their full credited name.

This appears to occur when the artist is associated with releases credited to multiple artists, such as *“Shabason & Krgovich”* or *“Shabason, Krgovich & Harris”*. In these cases, the artist library and search results collapse or truncate the name, even though:

- The `ARTISTS` tag is explicitly set to **“Nicholas Krgovich”**
- MusicBrainz Artist IDs are correct and consistent across all releases
- `Album Artist` reflects the as-credited collaborative name
- All albums were removed and fully re-synced after retagging

Attached log files and tag samples include both solo and collaborative releases.  Collaborative albums appear to influence how the artist is indexed or normalized globally.

A potentially related edge case may exist with artists such as **Crosby, Stills, Nash & Young**, where individual members also appear on releases credited under the combined group name. This suggests the issue may stem from how Music Assistant normalizes artist entities when multi-artist credits and individual artist identities overlap.

## Steps to reproduce

1. Add a music library containing both solo and collaborative releases by **Nicholas Krgovich**.
2. Ensure collaborative albums are credited with multi-artist `Album Artist` values (for example, *“Shabason & Krgovich”* or *“Shabason, Krgovich & Harris”*).
3. Tag tracks so that:
   - `ARTISTS` contains the full name **“Nicholas Krgovich”**
   - MusicBrainz Artist IDs are correct and consistent across all releases
4. Search for **Nicholas Krgovich** in Music Assistant or browse the artist list.

### Result
- The artist appears only as **“Krgovich”** in search and/or browse views.

### Expected result
- The artist should appear consistently as **“Nicholas Krgovich”**, matching the `ARTISTS` tag and MusicBrainz metadata.

### Music Providers

Filesystem

### Player Providers

Sonos

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/24847240/music-assistant.log)

### Additional information

[example_tags.txt](https://github.com/user-attachments/files/24847290/example_tags.txt)

Image

2025.12.5

Home Assistant OS

- [x] Yes

OzGav:
Fixes https://github.com/music-assistant/support/issues/4852

When importing tracks with album artists using mixed separators (e.g., "ArtistA, ArtistB & ArtistC"), the split_artists function produced malformed results by applying all matching splitters to the original string without breaking after finding one.

This has been fixed by adding a break after finding a splitter and recursive processing of split results, so mixed separators like "Shabson, Krgovich & Harris" correctly split into three separate artists.

tests have also been updated for this scenario

### DIFF
--- a/tests/core/test_tags.py
+++ b/tests/core/test_tags.py
@@ -106,20 +106,83 @@ async def test_parse_metadata_from_invalid_filename() -> None:
     assert _tags.musicbrainz_recordingid is None
 
 
-def test_split_artists_with_mixed_separators() -> None:
-    """Test splitting artists with mixed comma and ampersand separators."""
-    # Mixed comma and ampersand should split into 3 artists
-    result = split_artists("Shabson, Krgovich & Harris", allow_extra_splitters=True)
+def test_split_artists_with_expected_count() -> None:
+    """Test splitting artists guided by expected count (from MB IDs)."""
+    # With expected_count=3, should split on extra splitters to reach target
+    result = split_artists("Shabson, Krgovich & Harris", expected_count=3)
     assert result == ("Shabson", "Krgovich", "Harris")
 
-    # Only ampersands should also work
-    result = split_artists("Shabson & Krgovich & Harris", allow_extra_splitters=True)
+    # With expected_count=3, ampersands should split
+    result = split_artists("Shabson & Krgovich & Harris", expected_count=3)
     assert result == ("Shabson", "Krgovich", "Harris")
 
-    # Only commas should also work
-    result = split_artists("Shabson, Krgovich, Harris", allow_extra_splitters=True)
+    # With expected_count=3, commas should split
+    result = split_artists("Shabson, Krgovich, Harris", expected_count=3)
     assert result == ("Shabson", "Krgovich", "Harris")
 
-    # Without extra splitters, ampersand should not split
-    result = split_artists("Shabson & Krgovich", allow_extra_splitters=False)
+    # With expected_count=1, should NOT split at all
+    result = split_artists("Shabson & Krgovich", expected_count=1)
     assert result == ("Shabson & Krgovich",)
+
+    # With expected_count=None (no MB IDs), should NOT split on extra splitters
+    result = split_artists("Shabson & Krgovich", expected_count=None)
+    assert result == ("Shabson & Krgovich",)
+
+    # With expected_count=0 (no MB IDs), should NOT split on extra splitters
+    result = split_artists("Shabson & Krgovich", expected_count=0)
+    assert result == ("Shabson & Krgovich",)
+
+
+def test_split_artists_featuring() -> None:
+    """Test that featuring splitters always work regardless of expected_count."""
+    # "feat." should always split, even with no expected_count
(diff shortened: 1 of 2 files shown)


tests/core/test_tags.py
```diff
@@ -103,3 +104,22 @@ async def test_parse_metadata_from_invalid_filename() -> None:
     assert _tags.musicbrainz_artistids == ()
     assert _tags.musicbrainz_releasegroupid is None
     assert _tags.musicbrainz_recordingid is None
+
+
+def test_split_artists_with_mixed_separators() -> None:
+    """Test splitting artists with mixed comma and ampersand separators."""
+    # Mixed comma and ampersand should split into 3 artists
+    result = split_artists("Shabson, Krgovich & Harris", allow_extra_splitters=True)
```
marcelveldt: this is actually wrong because in some cases the performing artist is actually the combination.
Think of the Grease case where you have John Travolta & Olivia Newton John
In that case the performing artist is the combination of that two.

Also there are artists with an actual ampersand etc.

So the only time we allow this, is if there are also multiple artist musicbrainz id's
in fact we should maybe add the number of mb id's we discovered during this split action so we know how many artists to find in the string ?

OzGav: Yeah my Grease case was still OK as it had all the required tags but yes in cases where there was only a an ARTIST tag then ambiguity is there. So I have beefed up the logic to handle many edge cases.

The MusicBrainz Artist ID count is the source of truth for how many artists to extract from the ARTIST tag. Ideally an ARTISTS tag is used.
Logic:
  1. ARTISTS tag present → Split on ;, use all results (unchanged)
  2. ARTIST tag with MB Artist IDs:
    - 0 IDs: Only split on "featuring" splitters (feat., featuring, ft., duet with, vs.), never on &, ,, +, or with
    - 1 ID: Return as-is, no splitting at all
    - 2+ IDs: Split on featuring splitters first, then split on extra splitters (&, ,, +, with) only until we reach the MB
  ID count — then stop
  3. Warnings: Logged when we can't match the expected count or have to stop early or when number of artists found in the ARTISTS tag does not equal the number of MB ARTIST IDs

OzGav:
Also as a next PR I was thinking of adding support as requested here https://github.com/orgs/music-assistant/discussions/4595

marcelveldt (approved):
Looks good, thanks @OzGav !

PR #3068, opened 2026-01-31
Add a extensions.json file to .vscode

davidanthoff:


.vscode/extensions.json
```diff
@@ -0,0 +1,11 @@
+{
+    "recommendations": [
+        "ms-python.mypy-type-checker",
+        "ms-python.python",
+        "ms-python.debugpy",
+        "ms-python.vscode-python-envs",
```
MarvinSchenkel: I think python-envs does not add much value as we use `uv` and our own script (`scripts/setup.sh`) to do environment management. The others look fine to me 👍

MarvinSchenkel:

.vscode/extensions.json
davidanthoff: I removed it, but I think it is actually quite nice to have the UI to see the active env, and it seems that MS is starting to push it out by default with the Python extension as well. Oh, and It does use `uv` under the hood if it finds it.

MarvinSchenkel (approved):
Thanks @davidanthoff 🙏

PR #3078, opened 2026-02-02
Add generic short code authentication system

apophisnow:
Add a reusable short code authentication system that any provider can use for QR code login, device pairing, or similar flows.

Changes:
- Add join_codes database table
- Add generate_join_code(user_id, provider_name, ...) method
- Add exchange_join_code() to convert codes to JWT tokens
- Add auth/code public API endpoint
- Add revoke_join_codes() for cleanup
- Update login.html to handle ?join= parameter
- Add provider_name parameter to JWT token encoding

Providers can implement short code auth flows like:

    code, expires = await auth.generate_join_code(
        user_id=my_user.user_id,
        provider_name="my_provider",
        expires_in_hours=24,
    )

The provider_name is stored in the join code and passed to the JWT token, allowing providers to identify their authenticated sessions.


MarvinSchenkel:
Could you please update the auth tests to also include these new short code cases?

apophisnow:
> Could you please update the auth tests to also include these new short code cases?

Added tests.

marcelveldt:
Can you please walk me through the entire flow here you envision for the guest role + party mode ?

So, you create a user with GUEST role and then you can optionally request a short code token ?
How would a party mode guest user log in ? Remote access ?
Is there a single short code / join code for the whole "party" ? So all your guests share the same join code (and thus MA guest account) ?

At least, I think we should limit this feature to guest accounts only, for the sake of security.
Otherwise it looks fine, just some small feedback below:

**Security enhancements:**

- Default max_uses=1 instead of unlimited (safer default)
- Add automatic cleanup of expired codes (scheduled task)
- Log security events (code generation, exchange, rejection)

**API improvements:**

- Consider renaming to guest_access_code or quick_login_code to clarify purpose
- Add API endpoint to list active codes for a user (for management UI)
- Add ability to revoke a specific code (not just all codes for a user)

apophisnow:
> Can you please walk me through the entire flow here you envision for the guest role + party mode ?
> 
> So, you create a user with GUEST role and then you can optionally request a short code token ? How would a party mode guest user log in ? Remote access ? Is there a single short code / join code for the whole "party" ? So all your guests share the same join code (and thus MA guest account) ?
> 
> At least, I think we should limit this feature to guest accounts only, for the sake of security. Otherwise it looks fine, just some small feedback below:
> 
> **Security enhancements:**
> 
> * Default max_uses=1 instead of unlimited (safer default)
> * Add automatic cleanup of expired codes (scheduled task)
> * Log security events (code generation, exchange, rejection)
> 
> **API improvements:**
> 
> * Consider renaming to guest_access_code or quick_login_code to clarify purpose
> * Add API endpoint to list active codes for a user (for management UI)
> * Add ability to revoke a specific code (not just all codes for a user)

Sure thing. This specific PR is just to provide a mechanism for short join URLs which allow access without requiring manual login. So a URL like http://localhost:8095/?join=6Y6XQC can be provided to a user (through a text message, QR code, or any other means) which will allow them access without creating an account and logging in.

For party mode the idea is:
Enable party mode, it creates a party mode guest account with limited access, and a short join code which is used to generate the QR code. This all happens automatically when you enable party mode (not on an optional basis). It is a single account and join code (though each user will create a new session).

Since the join code is short it makes the QR code easier to scan, and if needed the URL could be easily shared with guests by writing it down, through text message, or even reading it aloud.

I think your suggestions make sense, i'll work to implement them.

marcelveldt:
Marked it as draft - please hit the "ready for review" button once you have implemented the suggestions. Thanks!

MarvinSchenkel (approved):
Thanks @apophisnow ! 👍

PR #3079, opened 2026-02-02
ABS: Use playback sessions, and optionally allow HLS stream

fmunkes:
The PR changes the provider such that it uses playback sessions for progress reporting. In that case, it is abs itself which calculates from the pure position and passed time if an item is finished or not etc. That also allowed me to remove some redundant code. So this fixes false marked_finished report by simply only reporting the current time, and not an is_finished at all.

ABS' web player either uses direct playback, which is the exact same as our MultiPartPath. This is still the default, and for me completely reliable. That is exactly what ABS does for its multi-files too.

In relation to https://github.com/music-assistant/support/issues/4754 and maybe https://github.com/music-assistant/support/issues/4586#issuecomment-3828430733 I added a by default disabled config option to use an hls stream instead.   The user may provide the file types / the extensions, where an hls stream shall be used. I made some good experience with aac (m4b), but not with mp3. I consider this an advanced config scenario.

The lib has been adjusted to account for some additional streams and session schema/ endpoints.


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -175,6 +189,26 @@ async def get_config_entries(
             required=False,
             hidden=True,
         ),
+        ConfigEntry(
```
marcelveldt, line 192: Why does this need to be configurable ? What is the benefit ?

fmunkes: The abs web player uses direct playback, aka our MultiFilePath, by default - this works for me and apparently most other users in MA without issues. The HLS stuff is only a fallback within abs if the player does not support the codec. I added it here in response to users having issues, but for me I don't see a benefit, and was indeed struggeling with e.g. mp3 files, but aac was fine. I don't want this playback type to be the default, to be honest, for me all formats work perfectly fine with MultiFilePath.

MarvinSchenkel: How about we merge it and validate whether this actually fixes those edge cases for those few users? If it does not, we can remove it?

fmunkes: Yeah, sounds good to me. I would then remove the HLS related code if needed, and would only keep the (non configurable) sessions, which do resolve the progress report issues.

MarvinSchenkel: Sounds like a plan 🙏

MarvinSchenkel (approved):

PR #3084, opened 2026-02-03
Fix squeezelite WAV playback

### LINKED ISSUE #4163: White noise instead of music, then output codec is WAV, 24 bit format and Squeezelite player is receiver

2.6.0b14

Home Assistant OS Addon

- [x] Yes

### The problem

Hello.
When I have the following streaming conditions from MA:
* as output stream codec is selected WAV
* stream format is 24-bit (with any sample rate 44.1, 48, 96, 192)
* receiver is Squeezelite players (in my case is Wiim Pro Plus or piCorePlayer)

...then both this players produces white noise, with music barely audible in the background.

If I limit the stream settings to 16-bits (in the MA player settings interface choose 16-bit formats only), when stream works correctly at any supported by the player sample rate.

Both this players connected to my DAC which can indicate the current format of the incoming signal. 
And in both cases, even when the MA streams in 24-bit format, the DAC indicator shows 16 bits.

I tried to run Moode player (which can also act as a squeezelite player) on my raspberry pi, instead of piCorePlayer and with this player, streaming in wav 24-bit format works correctly. I use the same players, but with DLNA protocol, and then the stream from MA in format WAV 24-bits is played correctly. I also installed LMS on my computer running Windows, and when streaming files in wav 24-bit format, all players work correctly without any noise.  

Thus, it turns out that when streaming in wav 24-bit format, some Squeezelite players, for some reason, process it as 16-bit. Moreover, the problem is typical for both the player installed on the raspberry pi and for the player running on a hardware device with the LinkPlay module. 

### How to reproduce

Enable player provider - squeezelite
Use as player piCorePlayer at raspberry pi, or Wiim Pro Plus (or maybe any device with LinkPlay module) with Squeezelite player enabled
In MA interface, on player settings page:
* Select output codec - WAV
* Select any supported format with 24 bit
* Start stream any track

### Music Providers

* Local files
* Tidal
* Qobuz

### Player Providers

Squeezelite (slimproto players)

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/21683435/music-assistant.log)

### Additional information

I tried to run Moode player (which can also act as a squeezelite player) on my raspberry pi, instead of piCorePlayer and with this player, streaming in wav 24-bit format works correctly. I use the same players, but with DLNA protocol, and then the stream from MA in format WAV 24-bits is played correctly. I also installed LMS on my computer running Windows, and when streaming files in wav 24-bit format, all players work correctly without any noise.  

2025.7.4

Home Assistant OS

Windows

- [x] Yes

OzGav:
REQUIRES: https://github.com/music-assistant/aioslimproto/pull/512

Fixes: https://github.com/music-assistant/support/issues/4163

When streaming WAV audio to slimproto players (squeezelite), include sample rate, bit depth, and channels in the Content-Type header (e.g., audio/wav;rate=44100;bitrate=24;channels=2). Without these parameters, slimproto defaults to 16-bit regardless of the actual stream bit depth, causing audio distortion when playing higher bit depth content.

### DIFF
--- a/music_assistant/controllers/streams/streams_controller.py
+++ b/music_assistant/controllers/streams/streams_controller.py
@@ -1844,11 +1844,23 @@ class StreamsController(CoreController):
             output_sample_rate = min(48000, output_sample_rate)
         if output_format_str == "pcm":
             content_type = ContentType.from_bit_depth(output_bit_depth)
+        output_channels = 1 if output_channels_str != "stereo" else 2
+        # For WAV format, explicitly set output_format_str with PCM parameters.
+        # This is needed for slimproto players which parse these parameters from the
+        # Content-Type header to configure the correct bit depth for decoding.
+        if content_type == ContentType.WAV:
+            wav_output_format_str = (
+                f"wav;rate={output_sample_rate};bitrate={output_bit_depth};"
+                f"channels={output_channels}"
+            )
+        else:
+            wav_output_format_str = ""
         return AudioFormat(
             content_type=content_type,
             sample_rate=output_sample_rate,
             bit_depth=output_bit_depth,
-            channels=1 if output_channels_str != "stereo" else 2,
+            channels=output_channels,
+            output_format_str=wav_output_format_str,
         )
 
     async def _select_flow_format(


MarvinSchenkel:
I think the better place for this would be in the`AudioFormat.__post_init__`. Add an `elif` :
```
elif not self.output_format_str and self.content_type == ContentType.WAV:
        self.output_format_str = (
            f"wav;rate={self.sample_rate};bitrate={self.bit_depth};channels={self.channels}"
        )
```

Looking at this a bit more. I think we also need to set the mime_types at `player.py` lines 438 and 464 to include that info, so `mime_type="audio/wav;rate=44100;bitrate=24;channels=2`.

Were you able to reproduce the white noise issue and check if this actually fixed it?

OzGav:
Yes I did a number of fix-revert-fix tests to confirm that it definately fixed it. I also checked all the other codecs with the fix in place to confirm there were no adverse impacts.

marcelveldt (approved):
Looks good, thanks @OzGav !

PR #3087, opened 2026-02-03
Fix cache check for expire

relic664:
`CacheController.get` checks to see if a query is expired or not when reading from cache. Crucially, it ignores the expiration time if no `checksum` is passed. This can cause unintended effects, like those raised in https://github.com/orgs/music-assistant/discussions/4865#discussioncomment-15683763. This simple fix has the `CacheController` always check `db_row["expires"]`.


OzGav:
Please run pre-commit as there is a lint problem

relic664:
Should be good to go now.

MarvinSchenkel (approved):
Great find @relic664! Thanks for this.

PR #3088, opened 2026-02-03
Fix(chromecast): Support HTTP Range for WAV streams

nicholas-gh:
## Summary

Some Chromecast “UA=Cast Lite” devices (e.g. Samsung soundbars) buffer a large portion of a track,
close the HTTP connection intentionally, then reconnect later with `Range: bytes=...-` to fetch more.
Music Assistant’s `/single` streams didn’t support Range, so these devices stop mid-track. Short tracks happen to play, but common length/longer stop mid-track.

This MR adds HTTP Range support for Chromecast `/single` streams when the output codec is WAV. Similar feature for the compressed streams is significantly harder so I don't have that done. I'm satisfied with using WAV.

The 'Flow mode' is a workaround via only feeding the device bytes slowly, I wasn't satisfied with that solution since it feels like it misses out on the benefit of the buffering (playback stability); I felt the next/previous controls were not quite as responsive; I think I still saw some disconnects/music stopping (just very rarely).

## AI

This code was written by Codex (GP-5.2 Extra High) with me guiding it to discover the "bug" (via having it write cast_stream_debug.py) and then the implementation. Before removing the 'draft' flag I intend to read through the code and ensure it doesn't do anything I'd consider unreasonable. I've raised the (draft) MR now in case other people want to see it early, it took me a while to discover what the problem with the Soundbar was, and I hope this might help others use their device.

(AI written from here)

## Why WAV (and not FLAC) for Range
- WAV is linear PCM, so byte offsets map directly to time offsets (bytes/sec).
- FLAC is lossless but compressed; for MA’s on-the-fly encoding pipeline there is no stable
  byte-offset → time mapping without pre-encoding/caching the full encoded stream. I've been working on a 'cache the full encoded stream' but don't have it working yet in a way that my Samsung soundbar accepts.

Default output remains FLAC for bandwidth reasons; WAV is an opt-in per player for devices that require Range.

## Changes
- Add `Range` parsing and `206 Partial Content` responses for Chromecast `/single` WAV streams:
  - `Accept-Ranges: bytes`, `Content-Range`, and correct `Content-Length`
  - Map requested byte offsets into WAV header vs PCM payload
  - Seek by whole seconds and skip remaining bytes within the first chunk for byte accuracy
  - Return `416 Requested Range Not Satisfiable` for invalid/unsatisfiable ranges
- Update Chromecast player config UI label to indicate WAV “supports range requests” while keeping default FLAC.
- Ensure Cast `contentType` is `audio/wav` when the stream URL ends in `.wav`.
- Add `scripts/cast_stream_debug.py` to reproduce the buffering/disconnect/Range-reconnect behavior in isolation. This script clearly shows the soundbar disconnecting and then later coming back and asking for a range. I understand this is acceptable/common Chromecast behaviour (according to ChatGPT 5.2)

## Testing / Repro
- Run the debug tool:
  - `./.venv/bin/python scripts/cast_stream_debug.py --cast-host <ip>`
  - Observe the Chromecast reconnecting with `Range: bytes=...-` requests.
- In Music Assistant, set the affected Chromecast player’s output codec to WAV and verify playback continues
  through reconnects (server should respond with `206` for Range requests).

## Limitations
- Range support is implemented for WAV `/single` streams with known duration.
- FLAC Range support is not included (would require full-file encode/caching to guarantee byte-stable ranges).

### DIFF
--- a/music_assistant/controllers/streams/streams_controller.py
+++ b/music_assistant/controllers/streams/streams_controller.py
@@ -19,7 +19,11 @@ from typing import TYPE_CHECKING, Final, cast
 
 from aiofiles.os import wrap
 from aiohttp import web
-from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption, ConfigValueType
+from music_assistant_models.config_entries import (
+    ConfigEntry,
+    ConfigValueOption,
+    ConfigValueType,
+)
 from music_assistant_models.enums import (
     ConfigEntryType,
     ContentType,
@@ -65,9 +69,12 @@ from music_assistant.constants import (
 from music_assistant.controllers.players.player_controller import AnnounceData
 from music_assistant.controllers.streams.smart_fades import SmartFadesMixer
 from music_assistant.controllers.streams.smart_fades.analyzer import SmartFadesAnalyzer
-from music_assistant.controllers.streams.smart_fades.fades import SMART_CROSSFADE_DURATION
+from music_assistant.controllers.streams.smart_fades.fades import (
+    SMART_CROSSFADE_DURATION,
+)
 from music_assistant.helpers.audio import LOGGER as AUDIO_LOGGER
 from music_assistant.helpers.audio import (
+    create_wave_header,
     get_buffered_media_stream,
     get_chunksize,
     get_media_stream,
@@ -117,7 +124,9 @@ CONF_ALLOW_BUFFER_DEFAULT = TOTAL_SYSTEM_MEMORY_GB >= 8.0
 def parse_pcm_info(content_type: str) -> tuple[int, int, int]:
     """Parse PCM info from a codec/content_type string."""
     params = (
-        dict(urllib.parse.parse_qsl(content_type.replace(";", "&"))) if ";" in content_type else {}
+        dict(urllib.parse.parse_qsl(content_type.replace(";", "&")))
+        if ";" in content_type
+        else {}
     )
     sample_rate = int(params.get("rate", 44100))
     sample_size = int(params.get("bitrate", 16))
@@ -131,7 +140,9 @@ class CrossfadeData:
 
     data: bytes
     fade_in_size: int
-    pcm_format: AudioFormat  # Format of the 'data' bytes (current/previous track's format)
+    pcm_format: (
+        AudioFormat  # Format of the 'data' bytes (current/previous track's format)
+    )
     fade_in_pcm_format: AudioFormat  # Format for 'fade_in_size' (next track's format)
     queue_item_id: str
 
@@ -402,7 +413,9 @@ class StreamsController(CoreController):
             fmt = ContentType.WAV.value
         else:
             fmt = plugin_source.audio_format.content_type.value
-        return f"{self._server.base_url}/pluginsource/{plugin_source.id}/{player_id}.{fmt}"
+        return (
+            f"{self._server.base_url}/pluginsource/{plugin_source.id}/{player_id}.{fmt}"
+        )
 
     async def serve_queue_item_stream(self, request: web.Request) -> web.StreamResponse:
         """Stream single queueitem audio to a player."""
@@ -429,7 +442,9 @@ class StreamsController(CoreController):
                     "Failed to get streamdetails for QueueItem %s: %s", queue_item_id, e
                 )
                 queue_item.available = False
-                raise web.HTTPNotFound(reason=f"No streamdetails for Queue item: {queue_item_id}")
+                raise web.HTTPNotFound(
+                    reason=f"No streamdetails for Queue item: {queue_item_id}"
+                )
 
         # pick output format based on the streamdetails and player capabilities
         if not queue_player:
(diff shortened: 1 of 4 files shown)


marcelveldt (changes requested):
I'm sorry we cant accept this PR in its current shape:

1) It modifies parts of the code that do not have to be modified (styling changes)
2) It duplicates a large part of the code
3) It is unacceptable to require WAV to stream to devices by default
4) Flow mode is a good working workaround for flaky devices
5) It is extremely inefficient as it consumes a full stream to throw away bytes.
6) It will break with smart fades completely

We might look into adding range request support one day (if even possible) but this is not the solution.

nicholas-gh:
> I'm sorry we cant accept this PR in its current shape:
> 
> 1. It modifies parts of the code that do not have to be modified (styling changes)
> 2. It duplicates a large part of the code
> 3. It is unacceptable to require WAV to stream to devices by default
> 4. Flow mode is a good working workaround for flaky devices
> 5. It is extremely inefficient as it consumes a full stream to throw away bytes.
> 6. It will break with smart fades completely
> 
> We might look into adding range request support one day (if even possible) but this is not the solution.

Many thanks for taking the initial look!

I'll dig in myself manually sometime over the next few weeks to see if I can do a better way (although not hopeful due to your 'if even possible').

I'll return to experimenting with Flow mode to see if I can get that robust with the Samsung Soundbar.

If not, I'll maintain my own WAV-range private fork for my house, a luxury of open source, thank you for making that possible!

nicholas-gh:
PS: I regret giving the early draft PR with generated code before I'd manually improved it; I won't do that again!

marcelveldt:
> I'll return to experimenting with Flow mode to see if I can get that robust with the Samsung Soundbar.

Yes please do so and inform us where you run into so we can fix any issues you see. We inform in our HTTP headers that we do not support range requests so the player should not attempt to do so. The idea behind flow mode is that it removes complexity from a player itself by just offering it one single "live" stream of audio of all queue tracks, including any crossfading etc. That should just work fine but we do have a readrate limit added to it because some players (eg. JBL etc.) stop playing without it. Question is finding the magic number that works with all players.

> If not, I'll maintain my own WAV-range private fork for my house, a luxury of open source, thank you for making that possible!

Sure but that can become a pain keeping it updated as our codebase moves very rapidly and you may miss out new cool features. Better fix the root issue(s) so everyone can benefit ;-)

nicholas-gh:
> We inform in our HTTP headers that we do not support range requests so the player should not attempt to do so.

Including this in case someone wants to refer to it in the future. Unfortunately the Samsung HW-QS750F Soundbar Chromecast implementation ignores the `Accept-Ranges: none` and asks for ranges anyway. This is demonstrated by a small debug utility, where we can see the initial GET, then it reads some bytes and drops the HTTP connection. After some time, it makes another GET this time with a range header:

```
HTTP server listening on 0.0.0.0:50890 (published as http://172.20.10.221:50890)
Range settings: advertise_range=False honor_range=True
Track URLs:
  - http://172.20.10.221:50890/track1.wav  [Track 1 (Content-Length, fast)]
  - http://172.20.10.221:50890/track2.wav  [Track 2 (Chunked, fast)]
  - http://172.20.10.221:50890/track3.wav  [Track 3 (Content-Length, throttled)]
  - http://172.20.10.221:50890/track4.wav  [Track 4 (Chunked, throttled)]
[21:03:41] CAST connection_status: CONNECTING
[21:03:41] CAST connection_status: CONNECTED
[21:03:41] CAST cast_status: app_id=None volume=0.07000000029802322
Connected to cast: Q-Series Soundbar
[21:03:42] CAST cast_status: app_id=CC1AD845 volume=0.07000000029802322
[21:03:42] CAST cast_status: app_id=CC1AD845 volume=0.07000000029802322
[21:03:42] CAST cast_status: app_id=CC1AD845 volume=0.07000000029802322
[21:03:42] CAST media_status: state=UNKNOWN t=0.0/None content_id=None
[21:03:42] CAST media_status: state=UNKNOWN t=0.0/None content_id=None
[21:03:42] CAST media_status: state=IDLE t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:42] HTTP GET /track1.wav from 172.20.10.228 range=None ua='Cast Lite'
[21:03:42] CAST cast_status: app_id=CC1AD845 volume=0.07000000029802322
Queue loaded (preloadTime=0).
Leave this running and watch for HTTP disconnects / cast socket drops.
Press Ctrl+C to stop.
[21:03:42] CAST media_status: state=PLAYING t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:42] CAST media_status: state=PLAYING t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:42] CAST media_status: state=PLAYING t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:42] CAST media_status: state=PLAYING t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:42] CAST media_status: state=PLAYING t=0/600 content_id=http://172.20.10.221:50890/track1.wav
[21:03:43] HTTP client disconnected while streaming /track1.wav (sent_header_bytes=44 sent_pcm_seconds≈81.0)
[21:04:29] HTTP GET /track1.wav from 172.20.10.228 range='bytes=10485760-' ua='Cast Lite'
[21:04:29] !!! CLIENT SENT Range header but server is NOT advertising range support (Accept-Ranges: none)
[21:04:30] HTTP client disconnected while streaming /track1.wav (sent_header_bytes=0 sent_pcm_seconds≈72.6)
[21:05:23] HTTP GET /track1.wav from 172.20.10.228 range='bytes=19939328-' ua='Cast Lite'
[21:05:23] !!! CLIENT SENT Range header but server is NOT advertising range support (Accept-Ranges: none)
[21:05:23] HTTP client disconnected while streaming /track1.wav (sent_header_bytes=0 sent_pcm_seconds≈73.0)
```

> Yes please do so and inform us where you run into so we can fix any issues you see.

Several times a day (with flow mode), the sound stops and I see:

```
2026-02-10 13:02:12.904 ERROR (Thread-10) [pychromecast.socket_client] [Q-Series Soundbar(172.20.10.228):8009] Error reading from socket: socket connection broken
2026-02-10 13:02:12.911 ERROR (Thread-10) [pychromecast.socket_client] [Q-Series Soundbar(172.20.10.228):8009] Failed to connect to service HostServiceInfo(host='172.20.10.228', port=8009), retrying in 5.0s
```

After that Music Assistant shows the device as 'off'.

I will see what debugging I can do.

(The device is 'disabled' in the HA cast integration, so I'm not expecting complications due to both MA and HA using the same device).

The audio quality of this device is excellent. I wish the firmware was better.

PR #3092, opened 2026-02-04
Snapcast: Stop the control scripts gracefully before shutting down the built-in snapcast server

gnumpi:
With the current Snapcast server version, control script subprocesses may not be shut down properly when the Snapcast server itself stops. This can leave behind running processes and open sockets.

This PR addresses the issue by explicitly stopping the control scripts before shutting down the built-in Snapcast server.

When the socket server shuts down, it sends a shutdown command to the control script, ensuring a clean and deterministic teardown.


MarvinSchenkel (approved):
Nice enhancement, thanks @gnumpi

PR #3094, opened 2026-02-04
Apple Music: Limit get_library_tracks to 150 songs per batch

MizterB:
Current batch size of 200 is triggering `504 Gateway Timeout` on every request, eventually causing synchronization of tracks to fail all retry attempts.  150 appears to be a safer value, only causing intermittent failures that are successful after retries.


MarvinSchenkel (approved):
Thanks @MizterB  🙏

PR #3096, opened 2026-02-05
Add Emby Music Provider

hatharry:


music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+    parse_artist,
+    parse_playlist,
+    parse_track,
+)
+
+if TYPE_CHECKING:
+    from music_assistant_models.provider import ProviderManifest
+
+CONF_URL = "url"
+CONF_USERNAME = "username"
```
OzGav: Have a look at the MA constants.py and import this and below

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+                f'Device="{self._device_name}", '
+                f'DeviceId="{device_id}", '
+                f'Version="{self.mass.version}"'
+            ),
+        }
+        try:
+            async with self._session.post(auth_url, json=payload, headers=headers) as resp:
+                resp.raise_for_status()
+                data = await resp.json()
+        except Exception as err:
```
OzGav: Broad exceptions are no good. Please identify the specific errors you want to trap and use the MA errors

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+        return False
+
+    async def _get(self, path: str, params: dict[str, Any] | None = None) -> dict[str, Any]:
+        url = urljoin(self._base_url, path.lstrip("/"))
+        try:
+            async with self._session.get(url, headers=self._headers, params=params) as resp:
+                resp.raise_for_status()
+                return await resp.json()  # type: ignore[no-any-return]
+        except ClientResponseError as err:
+            if err.status == 404:
```
OzGav, line 201: What about other possible HTTP error codes?

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+        track = await self.get_track(item_id)
+        # build universal audio URL (include token as query param for convenience)
+        container = ",".join(SUPPORTED_CONTAINER_FORMATS)
+        url = urljoin(self._base_url, f"Audio/{track.item_id}/universal")
+        params = {"static": "true", "Container": container, "api_key": self._token}
+        query = "&".join([f"{k}={v}" for k, v in params.items()])
+        return StreamDetails(
+            item_id=track.item_id,
+            provider=self.instance_id,
+            audio_format=AudioFormat(),
```
OzGav: Why is this empty?

music_assistant/providers/emby/parsers.py
```diff
@@ -0,0 +1,246 @@
...
+    Track,
+)
+from music_assistant_models.unique_list import UniqueList
+
+if TYPE_CHECKING:
+    from music_assistant.providers.emby import EmbyProvider
+
+
+def parse_track(
+    logger: logging.Logger,
```
OzGav: Why is this getting passed and then not used?

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+
+    async def get_library_playlists(self) -> AsyncGenerator[Playlist, None]:
+        """Yield all playlists from the music library."""
+        libs = await self._get_music_libraries()
+        for lib in libs:
+            params = {
+                "ParentId": lib[ITEM_KEY_ID],
+                "IncludeItemTypes": "Playlist",
+                "EnableUserData": "true",
+                "recursive": "true",
```
OzGav: There are three lower case “recursive” and many title case “Recursive”. I expect this should be consistent one way or the other?

music_assistant/providers/emby/parsers.py
```diff
@@ -0,0 +1,246 @@
...
+        provider_mappings={
+            ProviderMapping(
+                item_id=album_id,
+                provider_domain=provider.domain,
+                provider_instance=instance_id,
+            )
+        },
+    )
+
+    duration = int(item.get("RunTimeTicks", 0) / 10000000)  # Convert ticks to seconds
```
OzGav: Shouldn’t you be using your constant here? And also 10000000 would also be good as a constant

music_assistant/providers/emby/const.py
```diff
@@ -0,0 +1,43 @@
...
+    "Overview",
+    "ImageTags",
+    "DateCreated",
+]
+
+# Supported audio containers for streaming
+SUPPORTED_CONTAINER_FORMATS = ["mp3", "flac", "aac", "opus", "wav", "m4a"]
+
+# App/user agent name
+USER_APP_NAME = "MusicAssistant"
```
OzGav:
```suggestion
USER_APP_NAME = "Music Assistant"
```

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+            params = {
+                "ParentId": lib[ITEM_KEY_ID],
+                "IncludeItemTypes": "MusicArtist",
+                "EnableUserData": "true",
+                "Fields": ",".join(ARTIST_FIELDS),
+                "Recursive": "true",
+            }
+            page = 0
+            while True:
+                params["StartIndex"] = str(page * 100)
```
OzGav: Would be good to use a constant here instead of 100 as it is used in a similar fashion a number of times below

music_assistant/providers/emby/__init__.py
```diff
@@ -0,0 +1,506 @@
...
+            description="The username to authenticate to the remote server.",
+        ),
+        ConfigEntry(
+            key=CONF_PASSWORD,
+            type=ConfigEntryType.SECURE_STRING,
+            label="Password",
+            required=False,
+            description="The password to authenticate to the remote server.",
+        ),
+        ConfigEntry(
```
MarvinSchenkel: Is there a specific reason we want this configurable?

MarvinSchenkel:

music_assistant/providers/emby/__init__.py
```diff
@@ -473,10 +474,11 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
         url = urljoin(self._base_url, f"Audio/{track.item_id}/universal")
         params = {"static": "true", "Container": container, "api_key": self._token}
         query = "&".join([f"{k}={v}" for k, v in params.items()])
+        audio_format = next(iter(track.provider_mappings)).audio_format
```
MarvinSchenkel: This is incorrect. The idea is that you provide an audio format here based on whatever the server gives you back, e.g.
```
audio_format=AudioFormat(
                        content_type=ContentType.M4A,
                    )
```

However, if this is not known or could vary, you could just provider `AudioFormat()` here which will default to 'unknown' and MA will try to figure it out.

Other than this minor comment, this looks good to merge 👍

MarvinSchenkel:
@hatharry , maybe I wasn't clear in my previous comment, but `next(iter(track.provider_mappings)).audio_format` is not correct, as Emby is one of the providers that will provide this Track. So either:
- Try to obtain it from the remote (Emby) server
- If it's not available, just use `AudioFormat()`

MarvinSchenkel (approved):
Great work @hatharry and awesome addition to the music providers 🎉 !

PR #3106, opened 2026-02-06
Feature/queue modifier hooks

apophisnow:
Extracts the queue modification logic from Party Mode into a generic pattern which can be used by any provider.

This is useful if a provider wants custom control over how tracks are added to a queue. Party Mode uses this to prioritize guest requested songs over normal playlist items.

This PR is chained off of feature/generic-short-code-auth and will look much cleaner once that PR has been merged.


marcelveldt:
I'm not sure why this is even needed. It seems an overly complex solution for a simply task.
My proposal would be that the party mode provider adds custom api commands for enqueuing items an dthat the guest mode frontend talks to those custom commands.

In the provider code of the guest mode itself, it can simply call the existing public methods of the queue controller.
So, the guest mode provider is in control over the whole process of the guest mode party mode and accepts commands from the special/limited guest mode frontend and then handles all logic to finally call the queue controller for the actual action.

This also allows for the system admin to use normal queue operations to overrule the party mode.
Like I said before, keep the logic for party mode isolated in the party mode provider which I think is perfectly doable with the right architecture. We have added the "extra_attributes" attribute to both the PlayerQueue and PlayerQueueItem models so you can store custom data in these objects to help the provider.

PR #3108, opened 2026-02-07
Snapcast: Introduce fallback Snapcast setup for dev environments

gnumpi:
In some development environments, the default Snapcast configuration and plugin paths are not feasible.
This PR includes a default `snapserver.conf`, which is used if the system configuration file `/etc/snapserver.conf` does not exist.
Additionally, it falls back to a writable Snapcast plugin directory (subdirectory of the Music Assistant data directory) if the default `/usr/share/snapserver/plug-ins` is not accessible.

For convenience and readability, `_setup_controlscript()` is now called **before** starting the Snapcast server.
Hence, the SnapCastPlayer checks the newly introduced provider property `queue_control_available` to ensure that the control script is set up and that the Snapcast server has been started.


music_assistant/providers/snapcast/provider.py
```diff
@@ -138,36 +162,37 @@ async def _stop_builtin_server(self) -> None:
...
-            logger.warning(
-                "Could not copy controlscript (metadata/control disabled): %s",
-                err,
-            )
-            return False
+        if not CONTROL_SCRIPT.exists():
+            logger.warning("Control script does not exist: %s", CONTROL_SCRIPT)
+            return None
+
+        # fallback directory for dev environments
```
marcelveldt: This is not true - the control script always needs to be copied, otherwise snapcast doesn't allow it.
So in our docker container we install snapcast from the debian packages but we need to copy the controlscript.

So your change for the dev environment will now break the production environment

gnumpi: But it still does get copied, doesn't it? It first tires to copy the script to DEFAULT_SNAPSERVER_PLUGIN_DIR, which is the one as before, and only if it fails it tries to copy to the fallback_dir. At least that was what I intended to do, let me double check.

marcelveldt: Ah maybe just the inline comment is wrong then

gnumpi: I have double checked and created a test docker container. The control script is copied and used correctly.
I also moved the comment and fallback directory definition directly into the candidate list.

MarvinSchenkel (approved):
Nice work @gnumpi . Great to see some improvements to Snapcast!

PR #3110, opened 2026-02-07
Add artist artwork display for radio streams

OzGav:
With all the UI improvements I thought it would be good to enhance radio stations to rather than just show the station logo but instead show the current single cover/album/artist. This PR allows for the display of thumbnail images for radio streams when track metadata (Artist - Title) is available.

How it works:

- Parses "Artist - Title" from ICY/HLS stream metadata
- Checks user's library first for matching tracks/artists (fast, no API calls, respects user-curated images)
- Uses MusicBrainz recording search to verify artist + track for external lookups
- Tries single artwork first (exact track art), then falls back to album artwork via TheAudioDB/Fanart.tv
- Falls back to artist artwork lookup via TheAudioDB/Fanart.tv
- If no artwork found, uses the station logo as fallback
- Results cached (90 days hits, 7 days misses) to minimize API calls

Release group matching: Aggregates release groups from all matching MusicBrainz recordings and sorts by release date. Singles are tried first (for exact track artwork), then albums. Compilations and live albums are filtered out.

Title cleaning: Strips version suffixes (remastered, live, etc.) and featuring credits from track names to improve MusicBrainz matching.

Edge cases handled:

- Name format variations: Squier, Billy → Billy Squier, Bon_Jovi → Bon Jovi
- Swapped metadata: Detects stations sending "Track - Artist" instead of "Artist - Track"
- "The" prefix: Tries searches with and without leading "The " for artist names
- Diacritics: Blue Öyster Cult matches Blue Oyster Cult
- Collaborations: Artist1 feat. Artist2 extracts primary artist only
- Advertisements: Detects ad markers and shows station logo instead

### DIFF
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -1,4 +1,9 @@
-"""All logic for metadata retrieval."""
+"""All logic for metadata retrieval.
+
+TODO: This controller is getting large. Refactor into a dedicated subfolder
+with split files (controller.py, helpers.py, etc.) following the pattern
+of other controllers.
+"""
 
 from __future__ import annotations
 
@@ -77,7 +82,7 @@ from music_assistant.helpers.images import (
 from music_assistant.helpers.security import is_safe_path
 from music_assistant.helpers.tags import split_artists
 from music_assistant.helpers.throttle_retry import Throttler
-from music_assistant.helpers.util import clean_title_for_search, try_parse_int
+from music_assistant.helpers.util import parse_title_and_version, try_parse_int
 from music_assistant.models.core_controller import CoreController
 from music_assistant.models.music_provider import MusicProvider
 
@@ -730,7 +735,7 @@ class MetaDataController(CoreController):
         :returns: Tuple of (metadata, source_description, corrected_artist, corrected_track).
         """
         # Clean track name by stripping version suffixes and featuring credits
-        clean_track_name = clean_title_for_search(track_name)
+        clean_track_name, _ = parse_title_and_version(track_name, strip_for_search=True)
 
         # Check library track first - fast, no API calls, respects user-curated images
         if metadata := await self._get_library_track_metadata(artist_name, clean_track_name):
@@ -869,7 +874,7 @@ class MetaDataController(CoreController):
                     continue
                 if not compare_strings(track_name, track.name, strict=False):
                     continue
-                if image_url := await self._get_library_item_image(track):
+                if image_url := await self._get_library_item_thumb(track):
                     return MediaItemMetadata(
                         images=UniqueList(
                             [
@@ -932,7 +937,7 @@ class MetaDataController(CoreController):
                     return True
         return False
 
-    async def _get_library_item_image(self, track: Track) -> str | None:
+    async def _get_library_item_thumb(self, track: Track) -> str | None:
         """Get image URL for library track with fallback: track -> album -> artist.
 
         :param track: Track to get image for.
@@ -1048,17 +1053,22 @@ class MetaDataController(CoreController):
         # Standard flip (e.g., "Squier, Billy" -> "Billy Squier")
         return f"{after_comma} {before_comma}"
 
-    async def get_radio_stream_artwork(
+    async def get_image_url_by_name(
         self,
         artist_name: str,
         track_name: str,
         fallback_image_url: str | None = None,
     ) -> tuple[str | None, str | None, str | None]:
-        """Fetch artwork for radio stream based on current track metadata.
+        """
+        Look up artwork by artist and track name.
 
-        :param artist_name: Artist name (already normalized).
-        :param track_name: Track title.
-        :param fallback_image_url: Fallback image URL (e.g., station logo).
+        Searches library and external providers for matching artwork.
+        Also returns corrected artist/track names if the search detects
+        swapped metadata (e.g., "Track - Artist" instead of "Artist - Track").
+
+        :param artist_name: Artist name to search for.
+        :param track_name: Track title to search for.
+        :param fallback_image_url: Fallback image URL if no artwork found.
         :returns: Tuple of (image_url, corrected_artist, corrected_track).
         """
         if " / " in artist_name:
@@ -1152,7 +1162,7 @@ class MetaDataController(CoreController):
             fallback_url = streamdetails.stream_metadata.image_url
             original_artist = streamdetails.stream_metadata.artist
             original_title = streamdetails.stream_metadata.title
-            image_url, corrected_artist, corrected_track = await self.get_radio_stream_artwork(
+            image_url, corrected_artist, corrected_track = await self.get_image_url_by_name(
                 artist_name=original_artist,
                 track_name=original_title,
                 fallback_image_url=fallback_url,
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -142,12 +142,6 @@ LOCALES = {
 }
 
 DEFAULT_LANGUAGE = "en_US"
-REFRESH_INTERVAL_ARTISTS = 60 * 60 * 24 * 90  # 90 days
-REFRESH_INTERVAL_ALBUMS = 60 * 60 * 24 * 90  # 90 days
-REFRESH_INTERVAL_TRACKS = 60 * 60 * 24 * 90  # 90 days
-REFRESH_INTERVAL_AUDIOBOOKS = 60 * 60 * 24 * 90  # 90 days
-REFRESH_INTERVAL_PODCASTS = 60 * 60 * 24 * 90  # 90 days
-REFRESH_INTERVAL_PLAYLISTS = 60 * 60 * 24 * 14  # 14 days
 
 # Radio stream artwork cache settings
 CACHE_CATEGORY_RADIO_ARTWORK = 101
@@ -155,7 +149,6 @@ CACHE_EXPIRATION_RADIO_ARTWORK = 86400 * 90  # 90 days
 CACHE_EXPIRATION_RADIO_ARTWORK_MISS = 86400 * 7  # 7 days
 AD_DETECTION_PHRASES = ("asset link", "asset stop", "asset spot", "advert", "promo")
 
-PERIODIC_SCAN_INTERVAL = 60 * 60 * 6  # 6 hours
 REFRESH_INTERVAL = 60 * 60 * 24 * 90  # 90 days
 CONF_ENABLE_ONLINE_METADATA = "enable_online_metadata"
 MISSING_ARTIST_METADATA_SCAN_TASK_ID = "metadata_missing_artist_metadata_scan"
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -1,4 +1,5 @@
-"""All logic for metadata retrieval.
+"""
+All logic for metadata retrieval.
 
 TODO: This controller is getting large. Refactor into a dedicated subfolder
 with split files (controller.py, helpers.py, etc.) following the pattern
@@ -1059,8 +1060,7 @@ class MetaDataController(CoreController):
         track_name: str,
         fallback_image_url: str | None = None,
     ) -> tuple[str | None, str | None, str | None]:
-        """
-        Look up artwork by artist and track name.
+        """Look up artwork by artist and track name.
 
         Searches library and external providers for matching artwork.
         Also returns corrected artist/track names if the search detects
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -1055,7 +1055,7 @@ class MetaDataController(CoreController):
         :param fallback_image_url: Fallback image URL (e.g., station logo).
         """
         if " / " in artist_name:
-            artist_name = artist_name.split(" / ")[0].strip()
+            artist_name = artist_name.split(" / ", 1)[0].strip()
         else:
             artists_tuple = split_artists(artist_name)
             artist_name = artists_tuple[0] if artists_tuple else artist_name
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -11,7 +11,7 @@ import urllib.parse
 from base64 import b64encode
 from contextlib import suppress
 from time import time
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, cast
 from uuid import NAMESPACE_URL, uuid4, uuid5
 
 import aiofiles
@@ -21,12 +21,19 @@ from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption
 from music_assistant_models.enums import (
     AlbumType,
     ConfigEntryType,
+    ExternalID,
     ImageType,
     MediaType,
     ProviderFeature,
     ProviderType,
 )
-from music_assistant_models.errors import MediaNotFoundError, ProviderUnavailableError
+from music_assistant_models.errors import (
+    InvalidDataError,
+    MediaNotFoundError,
+    MusicAssistantError,
+    ProviderUnavailableError,
+    ResourceTemporarilyUnavailable,
+)
 from music_assistant_models.helpers import get_global_cache_value
 from music_assistant_models.media_items import (
     Album,
@@ -35,11 +42,13 @@ from music_assistant_models.media_items import (
     BrowseFolder,
     ItemMapping,
     MediaItemImage,
+    MediaItemMetadata,
     MediaItemType,
     Playlist,
     Podcast,
     Track,
 )
+from music_assistant_models.streamdetails import StreamMetadata
 from music_assistant_models.unique_list import UniqueList
 
 from music_assistant.constants import (
@@ -66,17 +75,19 @@ from music_assistant.helpers.images import (
     get_image_thumb,
 )
 from music_assistant.helpers.security import is_safe_path
+from music_assistant.helpers.tags import split_artists
 from music_assistant.helpers.throttle_retry import Throttler
-from music_assistant.helpers.util import try_parse_int
+from music_assistant.helpers.util import clean_title_for_search, try_parse_int
 from music_assistant.models.core_controller import CoreController
 from music_assistant.models.music_provider import MusicProvider
 
 if TYPE_CHECKING:
     from music_assistant_models.config_entries import CoreConfig
+    from music_assistant_models.streamdetails import StreamDetails
 
     from music_assistant import MusicAssistant
     from music_assistant.models.metadata_provider import MetadataProvider
-    from music_assistant.providers.musicbrainz import MusicbrainzProvider
+    from music_assistant.providers.musicbrainz import MusicbrainzProvider, MusicBrainzReleaseGroup
 
 
 def _detect_image_format(path: str) -> str:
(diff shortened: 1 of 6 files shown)


music_assistant/controllers/metadata.py
```diff
@@ -557,6 +563,88 @@ async def get_track_lyrics(
                 return metadata.lyrics, metadata.lrc_lyrics
         return None, None
 
+    async def get_artist_metadata_by_name(self, artist_name: str) -> MediaItemMetadata | None:
```
marcelveldt: I don't like this as it is error prone.
Better do a get_track_metadata_by_name where you need to send the artist string, track title and optional album name. Then the dynamic artwork can be a track image (if found), fallback to album image (if single match on album), last resort is artist image.

This is safer and also a bit nicer.

music_assistant/controllers/metadata.py
```diff
@@ -557,6 +563,88 @@ async def get_track_lyrics(
...
+                    copyright=library_artist.metadata.copyright,
+                    lyrics=library_artist.metadata.lyrics,
+                    review=library_artist.metadata.review,
+                )
+        except InvalidDataError:
+            pass
+
+        # Query metadata providers (only those supporting name-based search will return results)
+        # NOTE: No MusicBrainz ID available, so MBID-dependent providers like FanartTV won't match.
+        # TODO: Consider adding MusicBrainz lookup if a rate-limiting proxy is implemented.
```
marcelveldt: Now that we have our own musicbrainz mirror later this week, i;d even argue to ALWAYS use musicbrainz as it can do the matching and also has coverart image archive.

music_assistant/helpers/audio.py
```diff
@@ -791,6 +798,91 @@ async def resolve_radio_stream(mass: MusicAssistant, url: str) -> tuple[str, Str
...
+    """
+    normalized = artist_name.replace("_", " ")
+    if "," in normalized and " and " not in normalized.lower() and " & " not in normalized:
+        parts = normalized.split(",", 1)
+        if len(parts) == 2:
+            normalized = f"{parts[1].strip()} {parts[0].strip()}"
+    return normalized
+
+
+async def _fetch_artist_artwork_for_radio_stream(
```
marcelveldt: All this does not belong in a audio helper module at all

music_assistant/helpers/audio.py
```diff
@@ -838,6 +930,25 @@ async def get_icy_radio_stream(
                 )
                 streamdetails.stream_title = cleaned_stream_title
 
+                if " - " in cleaned_stream_title:
+                    parts = cleaned_stream_title.split(" - ", 1)
+                    artist_name_raw = parts[0].strip()
+                    track_name = parts[1].strip()
```
marcelveldt: yikes, all this duplicated logic again.

why not do streamdetails.stream_metadata.artist ?

music_assistant/providers/theaudiodb/__init__.py
```diff
@@ -147,12 +147,34 @@ async def get_artist_metadata(self, artist: Artist) -> MediaItemMetadata | None:
...
+                    if compare_strings(artist.name, db_name[4:], strict=False):
+                        return self.__parse_artist(item)
+
+        # Try search with "The" prefix added
+        if not artist.name.lower().startswith("the "):
+            result = await self._get_data("search.php", s=f"The {artist.name}")
+            if result and result.get("artists"):
+                for item in result["artists"]:
+                    if compare_strings(f"The {artist.name}", item["strArtist"], strict=False):
+                        return self.__parse_artist(item)
```
marcelveldt: Let's not do a artist search on name alone without providing any additional search context such as an album and/or track title. Artist names are simply not unique.

music_assistant/controllers/metadata.py
OzGav: Will now always use the MB lookup

music_assistant/controllers/metadata.py
OzGav: Yep removed and now will always lookup by track. Now uses `get_track_metadata_by_name` with MusicBrainz `search_recording` for artist+track matching.

music_assistant/providers/theaudiodb/__init__.py
OzGav: Per the above will now lookup by track

music_assistant/helpers/audio.py
OzGav: Moved `get_radio_stream_artwork`, `update_radio_stream_artwork`, `normalize_radio_artist_name`, and `get_radio_stream_station_image` to metadata controller.

music_assistant/helpers/audio.py
OzGav: Done - `update_radio_stream_artwork` now uses `streamdetails.stream_metadata.artist` and `.title` instead of re-parsing.

music_assistant/controllers/metadata.py
OzGav: Also went conservative with the album matching as it is not possible to know with certainty what album the track is from when there are more that more albums matched to the recording. For that reason I expect more often that not the artist image will be shown. If you would prefer to loosen that let me know.

OzGav:
Since we are trying to lookup on musicbrainz it is important clean artist and tracks names are sent. New functions added to support this.

OzGav:
Also I know some comments need to be tidied up or removed just wanted to keep my thought patterns there for now.

music_assistant/controllers/metadata.py
marcelveldt: If the album is the same as the track title, it is a single and then safe to grab the image from.
Tracks having an image is rare but a single will have it.

OzGav: I have impoved the album matching a lot and it is working pretty well now. There may still be some edge cases that I continue to monitor for

music_assistant/helpers/audio.py
```diff
@@ -995,6 +1026,28 @@ async def _update_hls_radio_metadata(
...
+                            artist=artist_normalized,
+                            image_url=station_image_url,
+                        )
+                        streamdetails.stream_metadata_last_updated = time.time()
+                        if streamdetails.queue_id:
+                            mass.player_queues.signal_update(streamdetails.queue_id)
+
+                        # Fetch artist artwork if not provided in stream metadata
+                        if artist and title and not image_url:
+                            asyncio.create_task(
```
marcelveldt-traveling: don't use asyncio.create_task but mass.create_task
also, when running into the situation where the metadata changes in quick succession you can have tasks overwriting eachother.

I would propose to use mass.call_later here with like a 200ms delay and add a task_id so the previous task is properly cancelled

music_assistant/helpers/util.py
```diff
@@ -172,6 +177,89 @@ def get_total_system_memory() -> float:
...
+_DISPLAY_SUFFIXES = (
+    "[Lyric Video]",
+    "[Official Video]",
+    "[Music Video]",
+    "[Official Audio]",
+    "(Official Video)",
+    "(Lyric Video)",
+    "(Music Video)",
+    "(Official Audio)",
+)
```
marcelveldt-traveling: why only use this for the search title ?
I would recommend to use this to use in our function that parses name + version.
Dont introduce new logic but fix existing logic.

music_assistant/helpers/util.py
```diff
@@ -172,6 +177,89 @@ def get_total_system_memory() -> float:
...
+
+
+def clean_title_for_search(title: str) -> str:
```
marcelveldt-traveling: see my previous comment - please drop this
we should already have our name+version parses that handles this

music_assistant/providers/musicbrainz/__init__.py
```diff
@@ -426,6 +426,113 @@ async def get_artist_details_by_resource_url(
                 return MusicBrainzArtist.from_raw(artist)
         return None
 
+    async def search_recording(
+        self, artist_name: str, track_name: str
+    ) -> tuple[MusicBrainzArtist, MusicBrainzReleaseGroup | None] | None:
```
marcelveldt-traveling: this is weird. search recording is the method name and you are returning a MusicBrainzReleaseGroup
This method should be renamed to get_release_group_by_track_name or something along those lines

marcelveldt-traveling:
I think we should always prefer the track metadata over anything else. 
We should try to fetch artwork/metadata in this order:

1) Try to get metadata of the actual recording/track (some popular tracks actually have track-level metadata!)
2) Fallback to single match (so album name is exactly the same as clean track name)
3) Fallback to first album that is NOT a compilation
4) Fallback to artist artwork

Also, I don't particular like the fact that this PR is duplicating logic, reinventing the wheel. Better improve existing logic instead of creating sidepaths that do more or less the same.

Then finally, we should aim at improving the lookup and metadata retrieval methods in the musicbrainz controller in a dedicated PR, not mangled into a PR that wants to fetch metadata for a radio stream.

Note that we can actually fetch images from MusicBrainz too using the coverart archive.

music_assistant/helpers/audio.py
OzGav: Changed in two places as I found another one the same in update_hls_radio_metadata

music_assistant/helpers/util.py
OzGav: Good point about not introducing new logic. I've made a few changes based on your feedback:

Replaced the hardcoded _DISPLAY_SUFFIXES list with a regex pattern that handles any casing/bracket variation of video/audio suffixes.

Eliminated _SEARCH_KEYWORDS entirely — the search patterns are now built directly from VERSION_PARTS (+ feat/ft), so there's a single source of truth for version keywords.

Added the missing keywords (edition, anniversary, stereo, album, bonus) to VERSION_PARTS so both `parse_title_and_version` and search cleaning benefit.

I kept `clean_title_for_display` as a separate function because it serves a different purpose than `parse_title_and_version` — for a title like "Elton John (feat. Kiki Dee) (Official Video)", the display function strips only the video suffix and preserves (feat. Kiki Dee), whereas `parse_title_and_version` would strip both since feat. is in IGNORE_TITLE_PARTS. For display purposes I think it would be better to keep the featuring credit visible.

music_assistant/helpers/util.py
OzGav: `clean_title_for_search` is modelled on the existing `genius_lyrics `helper. For search APIs (MusicBrainz, library search) we need to aggressively strip back to just the core artist/title to get reliable matches. `parse_title_and_version` doesn't go far enough because it can't strip bare featuring credits like "feat. Kiki Dee" that aren't in brackets — and adding that to `parse_title_and_version` would affect all 30+ providers at import time where we want to keep that data.

I can't see how I can use `parse_title_and_version` and maybe `split_artists` to make this work?

music_assistant/providers/musicbrainz/__init__.py
OzGav: yep. fixed

OzGav:
> Try to get metadata of the actual recording/track

This is not possible with MB or the Cover Art Archive as the endpoints are at the release level not recording. https://community.metabrainz.org/t/how-to-get-search-cover-image-of-track/337917/3

There is equally no way to query a track at TheAudioDB or Fanart.tv. MA, of course, does store images at track level and that is already the first thing I look for.

So your priortiy order is already how it works except for the single image as the second option so I have added that.

I have also created a PR just with the musicbrainz changes and marked this as draft until that is merged.

kieranhogg:
I appreciate it's an outlier, but just checking if there is a way to 'opt-out' of this feature when this functionality is already present in a provider as it is in the BBC Sounds one?

OzGav:
Since we made that change with the callback it should be fine I think?

kieranhogg:
Ah, I didn't even make the connection somehow! Strangely, one Sounds endpoint is missing track images, so I might well look at using these helpers as well, once it's landed.

music_assistant/controllers/streams/audio.py
```diff
@@ -607,6 +607,40 @@ async def get_icy_radio_stream(
                     )
                     streamdetails.stream_title = cleaned_stream_title
 
+                    if " - " in cleaned_stream_title:
```
marcelveldt, line 664: this could use a helper method

music_assistant/controllers/streams/audio.py
```diff
@@ -2050,6 +2087,34 @@ async def _update_hls_radio_metadata(
...
+                            streamdetails.stream_metadata = StreamMetadata(
+                                title=title or cleaned_title,
+                                artist=artist_normalized,
+                                image_url=station_image_url,
+                            )
+                            streamdetails.stream_metadata_last_updated = time.time()
+                            if streamdetails.queue_id:
+                                mass.player_queues.signal_update(streamdetails.queue_id)
+
+                            # Fetch artist artwork if not provided in stream metadata
```
marcelveldt: we fetch track/album/artist arywork, the artist is the last resort so this comment is wrong

music_assistant/controllers/metadata.py
```diff
@@ -131,6 +142,20 @@ def _detect_image_format(path: str) -> str:
 }
 
 DEFAULT_LANGUAGE = "en_US"
+REFRESH_INTERVAL_ARTISTS = 60 * 60 * 24 * 90  # 90 days
+REFRESH_INTERVAL_ALBUMS = 60 * 60 * 24 * 90  # 90 days
+REFRESH_INTERVAL_TRACKS = 60 * 60 * 24 * 90  # 90 days
+REFRESH_INTERVAL_AUDIOBOOKS = 60 * 60 * 24 * 90  # 90 days
+REFRESH_INTERVAL_PODCASTS = 60 * 60 * 24 * 90  # 90 days
+REFRESH_INTERVAL_PLAYLISTS = 60 * 60 * 24 * 14  # 14 days
```
marcelveldt: please rebase your branch as all this was removed

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +635,554 @@ async def get_track_lyrics(
...
+                return True
+            # Handle "The" prefix variations
+            if compare_strings(f"The {search_name}", artist.name, strict=False):
+                return True
+            if artist.name.lower().startswith("the "):
+                if compare_strings(search_name, artist.name[4:], strict=False):
+                    return True
+        return False
+
+    async def _get_library_item_image(self, track: Track) -> str | None:
```
marcelveldt: either change this to _get_library_item_thumb or add an optional ImageType param

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +635,554 @@ async def get_track_lyrics(
...
+        if streamdetails.queue_id and (
+            queue := self.mass.player_queues.get(streamdetails.queue_id)
+        ):
+            if queue.current_item and queue.current_item.media_item:
+                if station_image := queue.current_item.media_item.image:
+                    return station_image.path
+        return None
+
+    @staticmethod
+    def normalize_radio_artist_name(artist_name: str) -> str:
```
marcelveldt, line 1010: the metadata controller is getting big - lets remember to follow the example of the other controllers and move it to its own dedicated subfolder with split out files like controller.py, helpers.py etc

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +635,554 @@ async def get_track_lyrics(
...
+            return f"{after_comma} {before_comma}"
+
+        # Don't flip if 2+ words after comma (e.g., "Portugal, The Man")
+        if len(after_comma.split()) >= 2:
+            return normalized
+
+        # Standard flip (e.g., "Squier, Billy" -> "Billy Squier")
+        return f"{after_comma} {before_comma}"
+
+    async def get_radio_stream_artwork(
```
marcelveldt: why would this be limited to a radio stream ?
why not rename this to "get_image_by_name"

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +635,554 @@ async def get_track_lyrics(
...
+        if not streamdetails.stream_metadata:
+            return
+        if not streamdetails.stream_metadata.artist or not streamdetails.stream_metadata.title:
+            return
+
+        try:
+            fallback_url = streamdetails.stream_metadata.image_url
+            original_artist = streamdetails.stream_metadata.artist
+            original_title = streamdetails.stream_metadata.title
+            image_url, corrected_artist, corrected_track = await self.get_radio_stream_artwork(
```
marcelveldt: this is superweird to read
"get_radio_stream_artwork" --> it returns an image url and some names

music_assistant/helpers/util.py
marcelveldt: shouldnt we fix the parse_title_and_version helper then ?

OzGav: Ok I am thinking of doing this 

```
def parse_title_and_version(
    title: str, track_version: str | None = None, strip_featuring: bool = False
) -> tuple[str, str]:

```

and adding extra stripping logic to get to the single artist we need for search. How does that sound?

music_assistant/controllers/metadata.py
OzGav: Ok i will do that in a follow up PR

music_assistant/controllers/metadata.py
OzGav: fixed

music_assistant/helpers/util.py
OzGav: Now consolidated into options within parse_title_and_version

music_assistant/controllers/streams/audio.py
OzGav: Done

music_assistant/controllers/streams/audio.py
OzGav: I was being too general. Fixed

music_assistant/controllers/metadata.py
OzGav: renamed

music_assistant/controllers/metadata.py
OzGav: Renamed to get_image_url_by_name and updated the docstring to better reflect its general-purpose nature.

music_assistant/controllers/metadata.py
OzGav: The corrected artist/track names are a byproduct of the MusicBrainz search - when we detect swapped metadata (some stations send "Track - Artist"  instead of "Artist - Track"), we return the corrected names so the frontend display can be fixed. Renamed the method to get_image_url_by_name and clarified in the docstring that the corrected names are secondary outputs when swapped metadata is detected.

music_assistant/controllers/metadata.py
OzGav: TODO added at the top

music_assistant/controllers/metadata.py
```diff
@@ -1,4 +1,10 @@
-"""All logic for metadata retrieval."""
+"""
+All logic for metadata retrieval.
+
+TODO: This controller is getting large. Refactor into a dedicated subfolder
+with split files (controller.py, helpers.py, etc.) following the pattern
+of other controllers.
```
marcelveldt: this has nothing to do with this PR - strip it please

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +634,570 @@ async def get_track_lyrics(
...
+            "Looking up artwork for release group '%s' (mbid: %s)",
+            mb_release_group.title,
+            mb_release_group.id,
+        )
+        temp_album = Album(
+            item_id="temp",
+            provider="temp",
+            name=mb_release_group.title,
+            provider_mappings=set(),
+        )
```
marcelveldt, line 654: This somehow feels very weird to me. I suggest you add a comment here why this is needed this way.

music_assistant/controllers/metadata.py
marcelveldt: The conclusion is correct thoughl the metadata provider should be refactored into smaller chunks and logical subclasses and helpers, just like we did with streams, players etc.

Its just that it needs to be done in a separate PR

OzGav: It is just that above you mentioned that the refactoring should be done so my resolution was to add this TODO to remind to do it in a follow up? https://github.com/music-assistant/server/pull/3110#discussion_r3030353134 

I will take it out and just remember.

music_assistant/controllers/metadata.py
OzGav: This is because the metadata providers need an Album object passed to get_album_metadata, but they really only care about, ExternalID.MB_RELEASEGROUP and album.name.

The item_id="temp" and provider="temp" are just there to satisfy the Album constructor’s required fields — no provider actually looks at them. It’s essentially a carrier for the MusicBrainz ID.

I’ll add an associated note

music_assistant/controllers/metadata.py
```diff
@@ -610,6 +634,570 @@ async def get_track_lyrics(
...
+
+        :param mb_release_group: MusicBrainz release group to look up.
+        :returns: Tuple of (metadata, provider_name) or None if not found.
+        """
+        self.logger.debug(
+            "Looking up artwork for release group '%s' (mbid: %s)",
+            mb_release_group.title,
+            mb_release_group.id,
+        )
+        temp_album = Album(
```
OzGav, line 649:
```suggestion
        # Create a minimal Album object to pass the MusicBrainz release group ID
        # to metadata providers for artwork lookup.
        temp_album = Album(
```

marcelveldt (approved):
Nice work @OzGav !

marcelveldt:
For follow-up PR(s):
- Refactor metadata controller into smaller chunks
- Add configentry to enable metadata lookup for radio content (default enabled) so people that dont want it, can opt-out

kieranhogg:
> For follow-up PR(s):
> 
> * Refactor metadata controller into smaller chunks
> * **Add configentry to enable metadata lookup for radio content (default enabled) so people that dont want it, can opt-out**

As I've been following this, if radio providers implement custom lookup, is it expected to hook into this? I'll keep an eye out if so.

OzGav:
> As I've been following this, if radio providers implement custom lookup, is it expected to hook into this? I'll keep an eye out if so.

I’ll go out on a limb and say I wouldnt think so if it is integral to the provider.

PR #3111, opened 2026-02-07
Fix IPv6 address handling in bind config and ifaddr parsing

fmurodov:
- Fix `get_ip_addresses` to extract IPv6 address string from ifaddr tuples instead of stringifying the whole `(address, flowinfo, scope_id)` tuple
  - Add `::` (IPv6 all-interfaces) to bind IP options for both webserver and stream server
  - Treat `::` the same as `0.0.0.0` in webserver socket binding (map to `host=None`)

Follow-up to #3084


marcelveldt (approved):
Thanks, @fmurodov !

PR #3112, opened 2026-02-08
Sanitize queue_item.name in icy headers

fmunkes:
Fixes an aiohttp exception "Potential header injection attack." if there is a "\n" or similar in the queue_item.name in our headers.

See https://github.com/music-assistant/support/issues/4913


marcelveldt (approved):
Thanks, @fmunkes !

PR #3113, opened 2026-02-08
Print provider name in logging output

fmunkes:
Another attempt at better logging output especially for multi-instance providers.

Triggered by https://github.com/music-assistant/support/issues/4698

### DIFF
--- a/music_assistant/models/provider.py
+++ b/music_assistant/models/provider.py
@@ -74,7 +74,7 @@ class Provider:
         self.config = config
 
         # update log level if changed
-        if f"values/{CONF_LOG_LEVEL}" in changed_keys:
+        if f"values/{CONF_LOG_LEVEL}" in changed_keys or "name" in changed_keys:
             self._set_log_level_from_config(config)
 
         # reload if any non-log-level value keys changed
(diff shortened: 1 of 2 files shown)


music_assistant/models/provider.py
```diff
@@ -189,7 +189,13 @@ def _update_config_value(self, key: str, value: Any, encrypted: bool = False) ->
     def _set_log_level_from_config(self, config: ProviderConfig) -> None:
         """Set log level from config."""
         mass_logger = logging.getLogger(MASS_LOGGER_NAME)
-        self.logger = mass_logger.getChild(self.domain)
+        # self.name is only available after async_init. Otherwise we run into a race condition.
+        # see https://github.com/music-assistant/support/issues/4801
+        logging_name = self.domain
+        if getattr(self, "available", False):
+            # async_init completed
+            logging_name = f"{self.domain} ({self.name})"
```
marcelveldt: doesn't self.name already contain the provider name by default ?
Like, the default here is provider name + postfix.

fmunkes: That is true, but you could potentially remove it by editing the name - what do you prefer, we can go for only the name too.

marcelveldt: I would go with the name - the user will know what provider it is. Otherwise you end up with a duplicate in 99% of the cases

MarvinSchenkel (approved):

PR #3104, opened 2026-02-06
Snapcast: Fix player availability issues and align state with server truth

gnumpi:
### What this fixes

This PR resolves two issues in the Snapcast integration:

- **Players becoming unreachable / hidden when the group leader goes unavailable**  
  Players now correctly replicate the state of Snapcast server groups while taking their
  own connection status into account.

- **State mismatches between Snapcast controller apps and Music Assistant**  
  Player, group, and stream state is now replicated directly from the Snapcast server,
  ensuring consistent behavior across Music Assistant and external Snapcast controllers.
  
These fixes also enable **dynamic leader switching** for `SyncGroupPlayers`, which was
previously not possible due to fixed stream-to-player bindings.

---

#### Introduce `SnapcastMAStream`

A new abstraction representing Music Assistant–managed Snapcast streams.

`SnapcastMAStream`:
- Lets the **Snapcast provider keep track of open streams**
- Makes streams **independent of players**
  - A stream can continue even if the original player stops or leaves the group
- Manages its **own control script lifecycle**, including socket servers
- Encapsulates stream start/stop/restart logic and cleanup

---

#### Players replicate server state

- All relevant state is derived from Snapcast server updates:
  - player state
  - group membership
  - leader / follower roles
  - active stream
  - active media
  - availability

- This keeps Music Assistant fully aligned with Snapcast and other controller apps

---

#### Dynamic sync group leadership

- A `SyncGroupPlayer` can now switch leaders dynamically without interrupting the audio stream

---

### Developer-facing improvements

- Added **typing protocols** for the Snapcast Python library  
  (which currently does not provide any type hints)

- ~~**Snapcast server misconfiguration when running outside the official Docker container**  
  When Music Assistant is run outside the predefined Docker environment, the Snapcast
  server may silently ignore all command-line arguments if no valid configuration file
  is provided.  
  This is now handled explicitly by ensuring that a valid Snapcast configuration file
  is always supplied. Additionally, if `/usr/share/snapserver/plug-ins` is not writable, a writable plugin
  directory is automatically created under the Music Assistant data directory and used
  instead.~~ 
  **Split out into PR: #3108**

PR #3119, opened 2026-02-09
Snapcast: Fix elapsed time not updated issue

gnumpi:
Since PR #3104, the elapsed time is not reported by the SnapCastPlayer.

This PR fixes this by setting `elapsed_time_last_updated` to the (buffer time compensated) start time of the stream,
while `elapsed_time` returns 0 if the stream exists, else None.

### DIFF
--- a/music_assistant/providers/snapcast/ma_stream.py
+++ b/music_assistant/providers/snapcast/ma_stream.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import asyncio
 import random
+import time
 import urllib.parse
 from contextlib import suppress
 from typing import TYPE_CHECKING, cast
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Just double checking, is there _any_ way to obtain play time from the snapcast players themselves?

gnumpi:
> Just double checking, is there _any_ way to obtain play time from the snapcast players themselves?

It would be possible to let MA register itself as a client and read the timestamps, while keeping track of the amount bytes read from the MA pcm generator. This would give you an accurate play time (Neglecting server side re-syncs which we even could track from the logs). But I think this is overkill. 

Another approach would be to add an option to the Snapcast server which specifies the exact time when it should start streaming. Then we can derive the time directly and it would give us the possibility to even sync the streams with other providers like airplay or sendspin. If I find some time I will try to implement this.

MarvinSchenkel:
Agreed. As for the syncing with airplay and sendspin etc. We are working on a concept called 'sendspin bridges' that will eventually allow different protocols to sync with each other (via sendspin). We'll give you a heads up when we have an example for this :)

MarvinSchenkel (approved):

PR #3128, opened 2026-02-10
Fix builtin playlist genre display on refresh

### LINKED ISSUE #3899: "Random Album (from Library)" playlist genre does not update on refresh

2.5.1

- [x] Yes

### The problem

The "Random Album (from Library)" playlist includes a list of genres under the title (like all playlists).

However when you click the recycle button to pick a new random album it doesn't change, it sticks on whatever is already there, so e.g. you get things like:
Image

The album is tagged "GENRE=Progressive Rock", before I refreshed it was a Converge album which is tagged "GENRE=Hardcore".

Clicking the "Refresh metadata" menu item updates the genre to be correct -- although it can take a minute or more to actually update (I guess it refreshes other things too)

I expect this to happen for all the "Random * (from Library)" playlists, I can confirm it does happen for the artist one too.

### How to reproduce

Go to the random album playlist, observe that the genre list does not change when asking for a new choice by clicking the Image button.

### Music Providers

I only have a local storage provider and this is where I see the issue.

### Player Providers

Not specific to any player.

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/19820330/music-assistant.log)

### Additional information

I'm running in a docker container with an nginx reverse proxy in front. The issue is the same whether I visit MASS via an iframe dashboard in my HA instance or directly.

All my tracks are tagged using Picard. Nothing else in the troubleshooting seems relevant.

2025.4.2

Home Assistant Container

- [x] Yes

OzGav:
Fixes https://github.com/music-assistant/support/issues/3899 Builtin playlists (Random Album, etc.) not updating genres when refreshed.

Added genre counting during the track iteration when force_refresh=True. As we're already looping through every track to yield them, we grab the genre from each track (or its album if the track doesn't have one) and tally them up. At the end, we hand those counts off to metadata.save_playlist_genres().

I tried doing this the "proper" way through `_update_playlist_metadata` but that path has two problems:

1. It's behind a throttler (one call per 30 seconds) so the UI hangs waiting for it
2. It checks REFRESH_INTERVAL (90 days) and skips playlists that were recently refreshed

So I feel the genre collection needs to happen inline during the same iteration that's already happening. No extra iteration, no waiting on the throttler.

`metadata.py` - `save_playlist_genres() `

This takes the genre counts collected above and does the actual saving. It needs to do two things because I understand the UI reads genres from two places:

1. Write to genre_media_item_mapping table - this is what get_genres_for_media_item reads, which is what the UI actually calls to show genre tags
2. Update metadata.genres on the playlist object in the library DB - this fires MEDIA_ITEM_UPDATED which tells the UI to refresh

`metadata.py` - `_update_playlist_metadata()` genre filter improvement

The old filter (count > 5) was too aggressive for small playlists - a Random Album with 10 tracks might only have 1-2 genres with a count of 10. Now it keeps all genres for small playlists (<=20 total occurrences) and uses a proportional threshold for bigger ones. Also sorts by frequency and takes the top 8 rather than just grabbing an arbitrary set.

### DIFF
--- a/music_assistant/controllers/media/playlists.py
+++ b/music_assistant/controllers/media/playlists.py
@@ -154,6 +154,13 @@ class PlaylistController(MediaControllerBase[Playlist]):
                 yield track
             page += 1
 
+        # reset last_refresh on force_refresh so the metadata task
+        # will pick up this playlist and refresh genres + collage images
+        if force_refresh and library_item is not None:
+            library_item.metadata.last_refresh = None
+            await self.update_item_in_library(library_item.item_id, library_item)
+            self.mass.metadata.schedule_update_metadata(library_item)
+
     async def create_playlist(
         self,
         name: str,


music_assistant/controllers/media/playlists.py
```diff
@@ -89,10 +89,14 @@ async def tracks(
         item_id: str,
         provider_instance_id_or_domain: str,
         force_refresh: bool = False,
+        _from_metadata_update: bool = False,
```
marcelveldt: I dont like this pattern

music_assistant/controllers/media/playlists.py
```diff
@@ -110,6 +114,17 @@ async def tracks(
                 yield track
             page += 1
 
+        # trigger metadata update for builtin playlists when tracks are refreshed
```
marcelveldt: So if somebody smashes the update button it will cause a shitton of load ?
Let's not do this this way as the refresh metadata of a playlist can be heavy.
Instead we should mark the metadata as not fresh anymore and let the background task pick it up.

music_assistant/controllers/metadata.py
```diff
@@ -134,6 +134,8 @@ def __init__(self, mass: MusicAssistant) -> None:
         self._lookup_jobs: MetadataLookupQueue = MetadataLookupQueue(100)
         self._lookup_task: asyncio.Task[None] | None = None
         self._throttler = Throttler(1, 30)
+        # track pending metadata update requests per URI to cancel stale ones
+        self._pending_updates: dict[str, int] = {}
```
marcelveldt: this is duplicating _lookup_jobs

marcelveldt (changes requested):
Let's not do this and certainly not backport to stable. Think it through carefully first

music_assistant/controllers/media/playlists.py
OzGav: Changed approach

music_assistant/controllers/media/playlists.py
OzGav: Changed approach

music_assistant/controllers/metadata.py
OzGav: Changed approach

OzGav:
I have a looked at this a bunch of different ways. The solution I think will work is collecting genres inline during the tracks() iteration that already runs on force_refresh=True. This avoids a second iteration, scheduling, external calls, and the double UI refresh that the previous approach caused.
To avoid duplicating the genre logic between the playlist controller and _update_playlist_metadata, the extraction and filtering can be pulled out of both call sites into shared static methods on PlaylistController that both paths reuse. The full metadata path (collage images, timestamps) is otherwise unchanged.

Does that sound reasonable? I haven’t committed to here yet just wanted to sound you out to see if this might be the right way to go?

OzGav:
OK I have pushed through a solution based on the above. Tested working on the dev server

music_assistant/controllers/media/playlists.py
```diff
@@ -110,24 +114,38 @@ async def tracks(
...
+                    for genre in self.get_track_genres(track):
+                        genre_counts[genre] = genre_counts.get(genre, 0) + 1
                 yield track
             page += 1
+        # Save genres if collection was requested (genre_counts is not None).
+        # This includes empty dict {} to clear old genres when no genres found.
+        if genre_counts is not None:
+            db_item = await self.get_library_item_by_prov_id(prov_item_id, prov_instance)
+            if db_item:
+                await self._save_playlist_genres(db_item, genre_counts)
```
marcelveldt: this does not belong here

music_assistant/controllers/media/playlists.py
```diff
@@ -110,24 +114,38 @@ async def tracks(
...
+                prov_instance,
                 page=page,
                 force_refresh=force_refresh,
             )
             if not tracks:
                 break
             for track in tracks:
+                if genre_counts is not None:
+                    for genre in self.get_track_genres(track):
+                        genre_counts[genre] = genre_counts.get(genre, 0) + 1
```
marcelveldt: this does not belong here

music_assistant/controllers/media/playlists.py
```diff
@@ -110,24 +114,38 @@ async def tracks(
...
         if provider_instance_id_or_domain == "library":
             library_item = await self.get_library_item(item_id)
-            provider_instance_id_or_domain, item_id = self._select_provider_id(library_item)
+            prov_instance, prov_item_id = self._select_provider_id(library_item)
         # playlist tracks are not stored in the db,
-        # we always fetched them (cached) from the provider
+        # we always fetch them (cached) from the provider.
+        # when force_refresh is set, collect genres in the same pass
+        # to avoid a second iteration of the track list.
+        genre_counts: dict[str, int] | None = {} if force_refresh else None
```
marcelveldt: this does not belong here

there is also no more need to do it this way, since the fetching of the playlist tracks is now optimized

music_assistant/controllers/media/playlists.py
OzGav: Moved to metadata controller

music_assistant/controllers/media/playlists.py
OzGav: Moved to metadata controller

music_assistant/controllers/media/playlists.py
OzGav: I moved the genre persistence into `save_playlist_genres()` on the metadata controller — the playlists controller only collects the raw counts now and delegates the rest.

I did keep the counting inline in `tracks()` though. I tried routing it through `_update_playlist_metadata` but hit two issues: `schedule_update_metadata` skips anything refreshed within `REFRESH_INTERVAL` (90 days), and `update_metadata` is behind the throttler (30s) which blocks the UI. Happy to move it if there's another path I'm missing — could you point me at what you mean by the track fetching being optimised? I might be looking in the wrong place.

marcelveldt:
So I was thinking about this some more and actually... what are we trying to fix here ?
A whole lot of complex logic that causes overhead. Maybe we should simply not have any genres on these dynamic playlists because of the fact they are dynamic

OzGav:
So the issue is that genres are shown on the 3 randomly generated playlists (500, artist, album). Currently, those genres are not getting updated when the playlist is refreshed (and thus updated). 

Yes you could remove the genres. The only thing I would note is you would then have no genres against the album or artist on those playlists even though if you went to that album or artist you would see them. Not a major issue but worth noting.

music_assistant/controllers/media/playlists.py
marcelveldt: you need to reset the last_refresh in metadata so the metadata controller knows the metadata is stale and it will refresh it. This way you can keep the metadata collection (genres, images etc) isolated in the metadata controller and it doenst have to leak into the playlists controller

marcelveldt (changes requested):
the genres collection should remain in the metadata controller
only thing that you can do (if force refresh is true) is resetting the last_refresh on the playlist object

OzGav:
Will this change result in the frontend updating in a timely manner? What happens if a user presses the refresh button a couple of times?

PR #3130, opened 2026-02-10
Radio Paradise revert to using enriched API endpoint

OzGav:
I found the new endpoint for Radio Paradise's extended now playing info. Now we will use that primarily with a fallback to the simple endpoint. Upcoming track into is alternated with the artist name every 10secs to nice effect (I think!)


MarvinSchenkel (approved):
LGTM!

PR #3139, opened 2026-02-11
Fix library items not deleted after all providers have removed them 

MarvinSchenkel:
This PR 

- Fixes hanging library items when all providers have removed that item by checking if there are any prov_mappings left with in_library=1 after a sync
- This makes sure that MediaItems that were added to the MA library via search are still kept, as they are not included in the sync process
- The migration logic cleans up any library items that should have been deleted, with 1 exclusion: providers without 2 ways sync could have a valid db item entry with a provider_mapping.in_library = 0, since they might have been added via the search.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -46,6 +46,7 @@ from music_assistant_models.provider import SyncTask
 from music_assistant_models.unique_list import UniqueList
 
 from music_assistant.constants import (
+    CONF_PROVIDERS,
     DB_TABLE_ALBUM_ARTISTS,
     DB_TABLE_ALBUM_TRACKS,
     DB_TABLE_ALBUMS,
@@ -74,7 +75,7 @@ from music_assistant.helpers.tags import split_artists
 from music_assistant.helpers.uri import parse_uri
 from music_assistant.helpers.util import TaskManager, parse_title_and_version
 from music_assistant.models.core_controller import CoreController
-from music_assistant.models.music_provider import MusicProvider
+from music_assistant.models.music_provider import CACHE_CATEGORY_PREV_LIBRARY_IDS, MusicProvider
 from music_assistant.models.smart_fades import SmartFadesAnalysis, SmartFadesAnalysisFragment
 
 from .media.albums import AlbumsController
@@ -98,7 +99,7 @@ CONF_RESET_DB = "reset_db"
 DEFAULT_SYNC_INTERVAL = 12 * 60  # default sync interval in minutes
 CONF_SYNC_INTERVAL = "sync_interval"
 CONF_DELETED_PROVIDERS = "deleted_providers"
-DB_SCHEMA_VERSION: Final[int] = 26
+DB_SCHEMA_VERSION: Final[int] = 27
 
 CACHE_CATEGORY_LAST_SYNC: Final[int] = 9
 CACHE_CATEGORY_SEARCH_RESULTS: Final[int] = 10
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/music.py
```diff
@@ -1803,6 +1804,49 @@ async def _get_provider_recommendations(
...
+        :param provider_instance_id: The provider instance that just synced.
+        :param media_type: The media type that was synced.
+        :param prev_library_items: DB IDs from previous sync cycle.
+        """
+        cur_db_ids_list: list[int] | None = await self.mass.cache.get(
+            key=media_type.value,
+            provider=provider_instance_id,
+            category=CACHE_CATEGORY_PREV_LIBRARY_IDS,
+        )
+        cur_db_ids = set(cur_db_ids_list) if cur_db_ids_list else set()
```
marcelveldt, line 1827: so if the cache is empty we are assuming there are no current items ?
This is weird, we should either abort here or retrieve the item id's from db at this point

music_assistant/controllers/music.py
```diff
@@ -1823,7 +1867,16 @@ async def run_sync() -> None:
             # Wrap the provider sync into a lock to prevent
             # race conditions when multiple providers are syncing at the same time.
             async with self._sync_lock:
+                prev_library_items: list[int] | None = await self.mass.cache.get(
```
marcelveldt, line 1870: we need to replace this with a helper that checks cache first, db as fallback. purely relying on cache is not safe

music_assistant/controllers/music.py
```diff
@@ -2222,6 +2275,89 @@ async def __migrate_database(self, prev_version: int) -> None:  # noqa: PLR0915
...
 
+        if prev_version <= 27:
+            # Hard-delete items that were previously soft-deleted (all provider_mappings
+            # have in_library=0) and are exclusively from sync-back-enabled providers.
+            # Items from sync-back-disabled providers are kept (could be manually added).
+            sync_back_disabled: set[str] = set()
+            raw_providers = self.mass.config.get(CONF_PROVIDERS, {})
+            for instance_id in raw_providers:
+                if not self.mass.config.get_raw_provider_config_value(
+                    instance_id, "library_sync_back", True
```
marcelveldt, line 2286: I'm not fully understanding why this check is needed because when you manually add an item to the library, we will set the flag in_library to True anyways

MarvinSchenkel: This is for @arturpragacz 's use case where he disables two-way sync and just uses his music provider for music playback. He uses search to find content and then adds it to the MA library. This wil create a provider mapping with `in_library = False`, which is correct as it is not in the music provider's library: 
image

Basically, the whole challenge comes down to distinguishing:
"Media Items that have no provider mappings with in_library = True after a sync"
from
"Media Items that have no provider mappings with in_library = True due to MA only library management"

Purely from a database point of view, we cannot distinguish the two. Hence, this migration script uses the `library_sync_back` as an indicator that Media Items should be kept if they have a provider mapping with in_library=False for a provider that has 'sync back' disabled.

In the rest of the code changes, we rely on the sync context to detect this, but you were right about not purely relying on cache for this. Will fix that.

music_assistant/controllers/music.py
```diff
@@ -1803,6 +1804,49 @@ async def _get_provider_recommendations(
...
+                library_item = await controller.get_library_item(db_id)
+            except MediaNotFoundError:
+                continue
+
+            # check if any other provider mappings have this item in library
+            remaining_providers_in_library = {
+                x.provider_instance
+                for x in library_item.provider_mappings
+                if x.provider_instance != provider_instance_id and x.in_library
+            }
```
arturpragacz, line 1842: Why does it not check for `library_sync_back` here?

PR #3149, opened 2026-02-12
Add save_as_playlist command to player queue controller

chrisuthe:
Add a new player_queues/save_as_playlist API command that creates a new playlist from the current queue items. The method extracts URIs from playlist-compatible queue items and delegates to the existing PlaylistController.create_playlist() and add_playlist_tracks() methods.

Example Usage:
    command: "player_queues/save_as_playlist"
    args: {queue_id: "ma_jgesivgp4s", name: "My Queue Playlist"}
    
    Where ma_jgesivgp4s is the Queue ID we want to create from.

Tested with current dev frontend on HAOS + Music Assistant DEV package.


marcelveldt (approved):
Looks good, thanks @chrisuthe !

PR #3151, opened 2026-02-13
Fix case-sensitive comparison in compare_strings fuzzy matching

teancom:
## Summary

In non-strict mode, compare_strings lowercases str1 but passes the original-case str2 to SequenceMatcher on line 564. This causes fuzzy matching to penalize case differences, making it fail to match strings like "Track Feat. John" vs "TRACK FT. JOHN" that should be considered equivalent.

Also fix the elif branch (line 559) to replace on str2_lower instead of str2, so the result is consistently lowered.

  ## Labels
  `bugfix`


MarvinSchenkel (approved):
Good find! Thanks @teancom

PR #3152, opened 2026-02-13
Fix M3U parser truncating EXTINF duration to single character

teancom:
The EXTINF duration parser used `info[0].strip()[0]` which takes only the first character of the duration string. For multi-digit durations like "120", this truncates to "1". Also, the `-1` check on line 79 compared against "-1" but could only ever see "-" (a single char), so negative/unknown durations were never properly detected.

Remove the erroneous `[0]` index so the full duration string is used and the `-1` sentinel is correctly recognized.


MarvinSchenkel (approved):
Nice PR, to the point, nice tests, nothing to add. Thanks @teancom 🎉

PR #3153, opened 2026-02-13
Fix operator precedence and inverted dedup logic in audiobooks playlog

teancom:
Two bugs in the audiobook resume/playlog tracking:

1. `int(media_item.resume_position_ms or 0 / 1000)` never divides by 1000 because `/` binds tighter than `or`. The expression evaluates as `int(resume_position_ms or (0 / 1000))`, so non-zero values pass through as raw milliseconds instead of seconds.

2. The dedup check `abs(difference) > 2` skips the DB write when data HAS changed (difference > 2 seconds) and writes when it has NOT changed (difference <= 2 seconds). The comparison is inverted.

### DIFF
--- a/music_assistant/controllers/media/audiobooks.py
+++ b/music_assistant/controllers/media/audiobooks.py
@@ -305,12 +305,12 @@ class AudiobooksController(MediaControllerBase[Audiobook]):
                 "provider": "library",
             },
         )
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
Thanks @teancom

PR #3154, opened 2026-02-13
Fix library items not being deletes after providers have removed them

MarvinSchenkel:
This PR:
- Re adds the library changes from this commit: https://github.com/music-assistant/server/commit/1b60a3ca39faaed045520f43ae6d2d314f46f927
- Adds a migration script that sets all current provider mappings of streaming providers to `in_library = True` to avoid losing items in the overview pages
- Adds a new advanced property to music providers 'sync back deletions' (default enabled). When disabled, library deletions are not processed from the music provider. This supports a use case where a user mainly uses MA for library management. 
- Adds tests for library syncing


music_assistant/controllers/config.py
```diff
@@ -450,6 +451,8 @@ async def get_provider_config_entries(  # noqa: PLR0915
                 }
             ):
                 extra_entries.append(CONF_ENTRY_LIBRARY_SYNC_BACK)
+            if provider and isinstance(provider, MusicProvider) and provider.is_streaming_provider:
+                extra_entries.append(CONF_ENTRY_LIBRARY_SYNC_DELETIONS)
```
marcelveldt, line 455:
```suggestion
              if provider and isinstance(provider, MusicProvider) and provider.is_streaming_provider:
                  extra_entries.append(CONF_ENTRY_LIBRARY_SYNC_DELETIONS)
```
I think you also need that edit check from above ?

music_assistant/controllers/music.py
```diff
@@ -898,6 +898,8 @@ async def add_item_to_library(
             )
         # add to provider(s) library first
         for prov_mapping in full_item.provider_mappings:
+            # we optimistically set in library to True to prevent items from disappearing
```
```suggestion
            # we optimistically set in library to True to prevent items 
            # from disappearing when the provider doesn't support library edit
            # or 2-way sync is disabled.
```

music_assistant/controllers/music.py
```diff
@@ -938,6 +939,12 @@ async def refresh_item(  # noqa: PLR0915
         ctrl = self.get_controller(media_type)
         library_id = media_item.item_id if media_item.provider == "library" else None
 
+        # cache in_library state before the provider fetch overwrites media_item
```
marcelveldt, line 944: can you walk me through why this is needed ?

music_assistant/controllers/music.py
```diff
@@ -2222,6 +2234,15 @@ async def __migrate_database(self, prev_version: int) -> None:  # noqa: PLR0915
                 "WHERE media_type = 'playlist' AND in_library = 0;"
             )
 
+        if prev_version <= 27:
+            # set streaming provider mappings to in_library=True
+            await self._database.execute(
+                f"UPDATE {DB_TABLE_PROVIDER_MAPPINGS} SET in_library = 1 "
```
marcelveldt, line 2244: I think we can do better and only set the flag for items that do not have any mapping in library now.
Otherwise we cause troubles for people that actually have a multi provider instance setup tide to user accounts

music_assistant/controllers/music.py
```diff
@@ -2222,6 +2234,15 @@ async def __migrate_database(self, prev_version: int) -> None:  # noqa: PLR0915
...
+                "WHERE provider_domain NOT IN "
+                "('filesystem_local', 'builtin', 'test', 'jellyfin', 'emby', "
+                "'plex', 'opensubsonic', 'audiobookshelf', 'gpodder', 'podcastfeed')"
```
marcelveldt: I'm pretty sure we have some cached variable somewhere that returns you all non streaming providers

music_assistant/models/music_provider.py
```diff
@@ -700,36 +701,39 @@ async def sync_library(self, media_type: MediaType) -> None:
...
+                        }
+                        if not remaining_providers_in_library and library_item.favorite:
+                            # unmark as favorite since no providers have it in library anymore
+                            await controller.set_favorite(db_id, False)
+                        # unmark this provider mapping as in_library = False
+                        # we keep it in the library database so we can keep the metadata
+                        for prov_map in library_item.provider_mappings:
+                            if prov_map.provider_instance == self.instance_id:
+                                prov_map.in_library = False
+                        await controller.set_provider_mappings(
```
marcelveldt, line 733: Now, within set_provider_mappings we can do an additional check (which we may already do, I'm not sure): If the item has no mappings anymore AND it is not referenced in any of the link tables (e.g. track artists etc) we can safely clean it up from db

music_assistant/controllers/music.py
MarvinSchenkel: Without it, when you refresh an item, provider mapping's in library status is not maintained. It re-adds them with `in_library = None` causing them to disappear from the overview pages. This logic basically restores the original statuses after a refresh.

music_assistant/models/music_provider.py
MarvinSchenkel: Do we already delete those records from the link tables? Because otherwise this check will never be `True` if we keep for example `track_artist` records for tracks without any `in_library` provider mappings?

music_assistant/controllers/music.py
MarvinSchenkel: Double check: Only set all provider_mappings of an Item to inlibrary=True for Media Items where ALL provider_mappings.in_library are currently set to False?

music_assistant/controllers/config.py
MarvinSchenkel: Why? The former is to sync back additions in MA > Streaming provider. The new toggle prevents deletions in the streaming provider from flowing back into MA. 

I think this behaviour does not depend on a music provider supporting writing back library changes from MA > the Streaming provider?

So:
- Music provider does not support `LIBRARY_EDIT_XX`
- User disables sync back deletions
- User adds Album X in MA
- User adds Album X in music provider
- User removes Album X in music provider
- User syncs the music provider -> Should not delete Album X in MA

music_assistant/controllers/music.py
MarvinSchenkel: There is, but I think this is only populated _after_ we run the database migrations?

music_assistant/controllers/music.py
marcelveldt: yeah or just one - as long there is at least one, its already fine

music_assistant/controllers/music.py
marcelveldt: hmm yeah, true

music_assistant/models/music_provider.py
marcelveldt: there is already some logic to handle that yeah

music_assistant/controllers/media/albums.py
```diff
@@ -113,7 +113,9 @@ async def library_items(
         offset: int = 0,
         order_by: str = "sort_name",
         provider: str | list[str] | None = None,
+        library_items_only: bool = True,
```
marcelveldt: so I was thinking about this a bit and I don't think we should expose this on the public method as the whole purpose of this method is to get the library items.
We should just implicitly call the internal method with the library_items_only filter set to true

music_assistant/controllers/config.py
marcelveldt: ah yeah, I was reading it the other way around. deletions in MA to provider but that is already where the 2 way sync is for. So: nvm

marcelveldt (approved):

PR #3156, opened 2026-02-13
Workaround for "Youtube Music playlist stalls on uploaded music" music-assistant/support#4469

whitty:
This is a workaround for the "stalling" aspect of music-assistant/support#4469

MA tries to resolve albums providing the `prov_album_id` of the form `FEmusic_library_privately_owned_release_detailb_po_<probably private_letters_and_numbers_id>` which triggers an exception in `ytmusic.get_album()` as it checks that the ID `.startswith('MPRE')`.

The handling of the exception seems to lead to playback stalling in this case.  Catching and returning `None` (presumably like "not found") instead leads to a warning as follows, but playback doesn't stall.

```
2026-02-13 23:05:25.159 WARNING (MainThread) [music_assistant.music.track] Unable to fetch album details for ytmusic--xxxxxxx://track/JB2LoK_zfD4 - album://FEmusic_library_privately_owned_release_detailb_po_xxxxxxxxxx_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx not found on provider ytmusic--xxxxxxx
```

This may not resolve the base issues:

1. Why do we have a providerID for the album that cannot be looked up - has the wrong field been stored somewhere?
2. Why does reloading MA often allow that album to be played successfully (though I don't see calls to `get_album()` in that case)

----

Approach:

I preferred to catch the exception rather than check the prefix, in case `ytmusicapi` resolves the issue and allows non MPRE prefixes in the future.

Since the Exception is a generic `UserError` type I specifically check the message and propagate any other exceptions..

----

Additionally I removed a duplicate lookup rather than expand the `try` block further.

### DIFF
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -41,12 +41,21 @@ async def get_album(headers: dict[str, str], prov_album_id: str, language: str =
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album():
-        ytm = ytmusicapi.YTMusic(language=language)
-        album = ytm.get_album(browseId=prov_album_id)
+        if prov_album_id.startswith("FEmusic_library_privately_owned_release"):
+            ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
+            album = ytm.get_library_upload_album(browseId=prov_album_id)
+        else:
+            ytm = ytmusicapi.YTMusic(language=language)
+            album = ytm.get_album(browseId=prov_album_id)
+
         if "audioPlaylistId" in album:
             # Track id's from album tracks do not match with actual album tracks. E.g. a track
             # points to the videoId of the original version, while we want the album version
-            album_playlist = ytm.get_playlist(playlistId=album["audioPlaylistId"], limit=None)
+            try:
+                album_playlist = ytm.get_playlist(playlistId=album["audioPlaylistId"], limit=None)
+            except ytmusicapi.YTMusicError:
+                return album
+
             # Do some basic checks
             if len(album_playlist.get("tracks", [])) != len(album.get("tracks", [])):
                 return album
@@ -57,8 +66,7 @@ async def get_album(headers: dict[str, str], prov_album_id: str, language: str =
                     album_track["videoId"] = playlist_track["videoId"]
                     album_track["isAvailable"] = playlist_track.get("isAvailable", True)
                     album_track["likeStatus"] = playlist_track.get("likeStatus", "INDIFFERENT")
-            return album
-        return ytm.get_album(browseId=prov_album_id)
+        return album
 
     return await asyncio.to_thread(_get_album)
 
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -37,7 +37,9 @@ async def get_artist(
     return await asyncio.to_thread(_get_artist)
 
 
-async def get_album(headers: dict[str, str], prov_album_id: str, language: str = "en", user: str | None = None) -> dict[str, str]:
+async def get_album(
+    headers: dict[str, str], prov_album_id: str, language: str = "en", user: str | None = None
+) -> dict[str, str]:
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album():
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -37,7 +37,7 @@ async def get_artist(
     return await asyncio.to_thread(_get_artist)
 
 
-async def get_album(prov_album_id: str, language: str = "en") -> dict[str, str]:
+async def get_album(headers: dict[str, str], prov_album_id: str, language: str = "en", user: str | None = None) -> dict[str, str]:
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album():
@@ -238,7 +238,7 @@ async def library_add_remove_album(
     headers: dict[str, str], prov_item_id: str, add: bool = True, user: str | None = None
 ) -> bool:
     """Add or remove an album or playlist to the user's library."""
-    album = await get_album(prov_album_id=prov_item_id)
+    album = await get_album(headers=headers, prov_album_id=prov_item_id, user=user)
 
     def _library_add_remove_album():
         ytm = ytmusicapi.YTMusic(auth=headers, user=user)
--- a/music_assistant/providers/ytmusic/__init__.py
+++ b/music_assistant/providers/ytmusic/__init__.py
@@ -337,7 +337,7 @@ class YoutubeMusicProvider(MusicProvider):
     @use_cache(3600 * 24 * 30)  # Cache for 30 days
     async def get_album(self, prov_album_id) -> Album:
         """Get full album details by id."""
-        if album_obj := await get_album(prov_album_id=prov_album_id, language=self.language):
+        if album_obj := await get_album(headers=self._headers, prov_album_id=prov_album_id, language=self.language, user=self._yt_user):
             return self._parse_album(album_obj=album_obj, album_id=prov_album_id)
         msg = f"Item {prov_album_id} not found"
         raise MediaNotFoundError(msg)
@@ -345,7 +345,7 @@ class YoutubeMusicProvider(MusicProvider):
     @use_cache(3600 * 24 * 30)  # Cache for 30 days
     async def get_album_tracks(self, prov_album_id: str) -> list[Track]:
         """Get album tracks for given album id."""
-        album_obj = await get_album(prov_album_id=prov_album_id, language=self.language)
+        album_obj = await get_album(headers=self._headers, prov_album_id=prov_album_id, language=self.language, user=self._yt_user)
         if not album_obj.get("tracks"):
             return []
         tracks = []
--- a/music_assistant/providers/ytmusic/__init__.py
+++ b/music_assistant/providers/ytmusic/__init__.py
@@ -337,7 +337,12 @@ class YoutubeMusicProvider(MusicProvider):
     @use_cache(3600 * 24 * 30)  # Cache for 30 days
     async def get_album(self, prov_album_id) -> Album:
         """Get full album details by id."""
-        if album_obj := await get_album(headers=self._headers, prov_album_id=prov_album_id, language=self.language, user=self._yt_user):
+        if album_obj := await get_album(
+            headers=self._headers,
+            prov_album_id=prov_album_id,
+            language=self.language,
+            user=self._yt_user,
+        ):
             return self._parse_album(album_obj=album_obj, album_id=prov_album_id)
         msg = f"Item {prov_album_id} not found"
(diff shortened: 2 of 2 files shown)


MarvinSchenkel:
I would like to solve the root of the issue rather than implementing an 'except workaround'. Can you let me know what the album id is of your custom uploaded track that is triggering the issue? That might shed some light on this case.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -42,7 +43,14 @@ async def get_album(prov_album_id: str, language: str = "en") -> dict[str, str]:
 
     def _get_album():
         ytm = ytmusicapi.YTMusic(language=language)
-        album = ytm.get_album(browseId=prov_album_id)
+        try:
+            album = ytm.get_album(browseId=prov_album_id)
+        except ytmusicapi.exceptions.YTMusicUserError as e:
+            if prov_album_id is not None and "must start with MPRE" in str(e):
+                logging.getLogger("music_assistant").getChild("YouTube Music").debug("ytmusicapi refused to handle prov_album_id: '%s'", prov_album_id)
+                return None
```
whitty: Note `None` is not a normal return path for `ytm.get_album()` there is no obvious "not-found" result

whitty:
> I would like to solve the root of the issue rather than implementing an 'except workaround'. 

Understood.  As I've pointed out the `None` return isn't the same as a "not found"

> Can you let me know what the album id is of your custom uploaded track that is triggering the issue?

Can you explain exactly what information you are after?

As described the `prov_album_id` is a string of the form `'FEmusic_library_privately_owned_release_detailb_po_CP39qciXhNa_LxIUaG90dGVzdCAxMDAgdm9sdW1lIDgaCHRyaXBsZSBqIgNncG0'` - I do not know if that is the album ID you are after - I did mention it in the PR description.

I do not know where this value came from - presumably from a query of the youtube api elsewhere and/or stored by MA itself from some source.

MarvinSchenkel:
That's exactly the info I was after. It looks like YTM internally stores this value as the album id for a track that was manually uploaded. Now of course we cannot link this to any album out there, which ultimately causes this error.

My suggestion is to **prevent this value from ever creeping into MA**. So instead of failing on a invalid album id in the `helpers.py`, we could just simply extend [this](https://github.com/music-assistant/server/blob/fdf1ad5e6ad69dfbb8361cc753cc1e21bd3a05d0/music_assistant/providers/ytmusic/__init__.py#L928) check by checking if the album id is a valid id (e.g. `and track_obj["album"]["id"].startswith("MPRE")`). This will only add the Album to the track if the id is valid, preventing your album id from ever being synced to MA in the first place.

MarvinSchenkel (changes requested):
Please have a look at my suggestion and mark this PR 'Ready for review' again once you have made the changes.

OzGav:
@whitty any progress on this? Would be good to get it into 2.8

whitty:
> That's exactly the info I was after. It looks like YTM internally stores this value as the album id for a track that was manually uploaded. Now of course we cannot link this to any album out there, which ultimately causes this error.
> 
> My suggestion is to **prevent this value from ever creeping into MA**. So instead of failing on a invalid album id in the `helpers.py`, we could just simply extend [this](https://github.com/music-assistant/server/blob/fdf1ad5e6ad69dfbb8361cc753cc1e21bd3a05d0/music_assistant/providers/ytmusic/__init__.py#L928) check by checking if the album id is a valid id (e.g. `and track_obj["album"]["id"].startswith("MPRE")`). This will only add the Album to the track if the id is valid, preventing your album id from ever being synced to MA in the first place.

Sorry - I've been busy and not gotten back to this.  I haven't had a chance to play around here.

I'm trying to parse your suggestion, and also clarify the behaviour I see (before attempting the change).

I know these are true:
1. There's nothing inherently "wrong" with those IDs "in general" because from MA's perspective the tracks are still organised correctly into albums - ie they still have identity value, even if some metadata isn't available.  MA correctly shows the album name (possibly the album Art, but that may be embedded in the track).
2. Playback can continue without the album request that failed in the [original stack-trace](https://github.com/user-attachments/files/23843454/music-assistant.log) (player-queues -> music -> media/tracks -> media/albums)  
[as described in this update](https://github.com/music-assistant/support/issues/4469#issuecomment-3896960813) - once the fail is "cached" (after a restart) it doesn't trigger the fail and playback seems normal.

> This will only add the Album to the track if the id is valid,

My concern with this, is that it might "throw out the baby with the bathwater".   Would that change leave the track "orphaned" in MA - just a track, but not on an album?  

Is `track.album = self._get_item_mapping(MediaType.ALBUM, album["id"], album["name"])` presumably is why we have at least the album name attached?

What I can't tell from the stack-traces is what information is being fetched, but I'd prefer to keep the album for its "Identity value", just return any metadata queries as if there was no data attached to that album.  Not knowing the structure of how MA manages things I might be saying the wrong thing.  I know my change doesn't really do _that_, but it kind of does by effect.  I suspect there's a point in the middle where this could be managed more cleanly.

Trying to look a little deeper:

https://github.com/music-assistant/server/blob/fdf1ad5e6ad69dfbb8361cc753cc1e21bd3a05d0/music_assistant/controllers/media/tracks.py#L133C1-L136C18

```
            elif isinstance(track.album, ItemMapping) or (track.album and not track.album.image):
                track.album = await self.mass.music.albums.get(
                    track.album.item_id, track.album.provider, recursive=False
                )
```

This is replacing `track.album` - either if its an `ItemMapping`, or if it doesn't have an `.image`.  Is `ItemMapping` just a stub for later resolving?

Did we have enough information in `track_obj["album"]` in ? do we need to create a mapping (assuming that's a deferred lookup)?  Is there another object that can be inserted other than a mapping - this code from above `(track.album and not track.album.image)` suggests there is an alternative to an `ItemMapping` there (sorry for the silly questions I C++ by day, so I find it hard when I don't know _what sort of object_ I'm pointing at).

I will instrument the code you have mentioned, then try preventing the mapping from being created, though I'm worried about the side-effects and whether `@cache` will make it hard to be sure what the actual result is.

whitty:
OK - I've looked deeper and I'm happier with this, though I still class it as "workaround" at this point - it prevents the playback errors.

Digging through ytmusicapi `ytm.get_library_upload_album()` is the correct API for [fetching uploaded albums with the given prefix](https://github.com/sigma67/ytmusicapi/blob/a8a120f37c5363ffb4d36d226c1afb2051bd4d79/ytmusicapi/mixins/uploads.py#L197).

This seems to resolve an OK `Album` object.  However there are some rough edges.  The track-list doesn't seem to resolve properly though its in the returned data.

Is there a place to put test-data - eg unittests for `_parse_album()` - where I could store returned data for reference?  Its 90% the same in shape to a regular album return, with a few possible rough edges that might be causing the failure to cross-link with tracks.

MarvinSchenkel:
Yep there is a `tests` folder with tests. A few providers have already set up tests for this, you can just follow their pattern to setup some tests for album parsing in YTM :)

MarvinSchenkel:

MarvinSchenkel:
@whitty any progress on this one?

whitty:
> @whitty any progress on this one?

Yes and no

The current status of this change is that it resolves the "MPRE" stall and allows playback.  There seems to be a minor issue where the private/uploaded album referenced in the track doesn't have a backward link back to the track in the MA database - I think it does on YTM though.

That doesn't affect anything other than the navigation through the UI, you can click through a track -> album, but there's no link back to the tracks in the album (album -> track)

I started looking at trying to fill out the missing MA database entries related to the albums so they get bi-directional linkage, but haven't had the time to work through it.

> Yep there is a tests folder with tests. A few providers have already set up tests for this, you can just follow their pattern to setup some tests for album parsing in YTM :)

I added unit-tests with data as suggested, but realised it was completely trivial in what it actually tested.  I might make a separate PR in case _any_ of the test code is useful, but it didn't help me validate much of the code anyway, as most of the code in the album metadata parsing function is optional and/or not filled in even with the non upload album.

whitty:
> @whitty any progress on this one?

TL;DR

I think this change "as-is" resolves the issue, without having to fix every underlying problem in YTMusic provider.  If you're willing, we should look at landing it.

MarvinSchenkel:
> That's exactly the info I was after. It looks like YTM internally stores this value as the album id for a track that was manually uploaded. Now of course we cannot link this to any album out there, which ultimately causes this error.
> 
> My suggestion is to **prevent this value from ever creeping into MA**. So instead of failing on a invalid album id in the `helpers.py`, we could just simply extend [this](https://github.com/music-assistant/server/blob/fdf1ad5e6ad69dfbb8361cc753cc1e21bd3a05d0/music_assistant/providers/ytmusic/__init__.py#L928) check by checking if the album id is a valid id (e.g. `and track_obj["album"]["id"].startswith("MPRE")`). This will only add the Album to the track if the id is valid, preventing your album id from ever being synced to MA in the first place.

Before we merge this as-is, can you explain why my suggestion here does not work? Does your current fix actually resolve an Album object?

whitty:
> Before we merge this as-is, can you explain why my suggestion here does not work? Does your current fix actually resolve an Album object?

That may depend upon what you mean by an Album object.  The existing YTM code that runs in response to the call that was failing calls `_parse_album()` with the new upload result and creates an `Album` object and returns it in answering the call `provider.get_album()`.

The result from `_parse_album()` is not noticeably different from the type of result you get from parsing a "regular" album as captured when I instrumented the code.  You can see the test I added which calls `_parse_album()` against [both regular and upload responses from `ytmusicapi` here](https://github.com/whitty/homeassistant-server/commit/34480c31c7a101515006c326d3392d7e49631159)

That said, I think the Album isn't "entirely" in the database because it would not be returned by the provider's enumeration function `get_library_albums()`.  I've been playing with a candidate change to [enumerate the upload albums here](https://github.com/whitty/homeassistant-server/commit/f7a0ce90b90d070f18e9133a276ac7afdeac8583).  However I've _never_ seen `get_library_albums()` get called (with or without my patch), so I have no way to verify if that change is good or bad.

Again the only observed issue with it being missing is that when you look at an upload album with _this_ patch it has name, details etc,... just no track-list.  That said, `parse_album()` doesn't seem to enumerate the "regular" album tracks anyway.

Sorry if I've being obtuse as I don't really understand the structure of MA and how a provider is _supposed_ to work.  I'm only looking at what _does_ work.

---

If I understood the shape of things better (and had infinite time) I would:
1. create a unittest that instantiates the provider itself and runs all the database filling code.  I see that's done with some providers who's APIs provide test hooks or fake-apis for testing.
2. try to force my MA to rebuild its database to see if enumerating the albums works as expected.

music_assistant/providers/ytmusic/helpers.py
whitty: Note this code is no longer present as I now just call the correct API

whitty:
> My suggestion is to **prevent this value from ever creeping into MA**

Please make sure you've looked at the change as it was since March - I do not think this idea makes sense in the light of the changes I force pushed on March 20.

OzGav:
Please check on the lint error while you wait

MarvinSchenkel (approved):
Thanks @whitty , fixed the lint and tested for regressions. Seems to work fine on my end 🙏

PR #3159, opened 2026-02-13
Fix RuntimeError from dict/set mutation during iteration

teancom:
Several collections are iterated while concurrent callbacks can modify them, causing "Set/dictionary changed size during iteration" RuntimeErrors:

- mass.py: _subscribers set iterated in signal_event() while subscribe/unsubscribe callbacks modify it
- mass.py: _providers dict iterated in get_provider() and _on_mdns_service_state_change() while providers load/unload
- player_controller.py: _players dict iterated in get_players(), get_player_by_name(), and update_player_control() while players register/unregister
- music.py: in_progress_syncs list iterated while done-callbacks call .remove() on it

Snapshot via list() before iterating in all affected locations.

### DIFF
--- a/music_assistant/mass.py
+++ b/music_assistant/mass.py
@@ -366,7 +366,7 @@ class MusicAssistant:
                 return None
             provider_instance_or_domain = prov.domain
         # fallback to match on domain
-        for prov in self._providers.values():
+        for prov in list(self._providers.values()):
             if prov.domain != provider_instance_or_domain:
                 continue
             if return_unavailable or prov.available:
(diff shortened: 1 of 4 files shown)


teancom:
Rebased off dev to sort out the merge conflict.

music_assistant/mass.py
MarvinSchenkel: I think we also need it here then?

MarvinSchenkel: And here?

teancom:
Good eyes! Yup, totally missed those two spots. I remonstrated with Claude and it actually found a third place in mass.py (~~and a few other places but I'm going to open a new PR once I've verified them~~ and one other, as mentioned below. The rest were false positives).

teancom:
Added one more that's of the same type/class. That *should* be it for this PR.

MarvinSchenkel (approved):
LGTM, thanks @teancom 🙏

PR #3160, opened 2026-02-13
Fix index_in_buffer or current_index treating index 0 as unset

teancom:
index_in_buffer and current_index are int | None where 0 is a valid value (first track in queue). Two locations use chained `or` to select between them:

- set_shuffle: `queue.index_in_buffer or queue.current_index`
- play_media_from_queue: `queue.index_in_buffer or queue.current_index or 0`

When index_in_buffer is 0 (first track buffered), Python's `or` treats it as falsy and falls through to current_index. This causes shuffle reshuffling and queue insertion to use the wrong position.

Replace with explicit `is not None` checks in both locations.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -313,7 +313,9 @@ class PlayerQueuesController(CoreController):
             return  # no change
         queue.shuffle_enabled = shuffle_enabled
         queue_items = self._queue_items[queue_id]
-        cur_index = queue.index_in_buffer or queue.current_index
+        cur_index = (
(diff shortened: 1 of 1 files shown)


marcelveldt:
In some situations this is handled as fallback value when the queue in newly populated so careful testing is adviced

teancom:
That *should* still be fine. The behavior with the new logic is just "Use queue.index_in_buffer if it exists even if the index of the track in the buffer is the first track, zero". If it doesn't, then it continues along and uses queue.current_index if that exists (even if it's zero) and finally, if neither exists, use zero. So there shouldn't be any new fall-through / undefined behavior.

MarvinSchenkel (approved):
Thanks @teancom , nice find!

PR #3161, opened 2026-02-13
Fix custom base URL with subpath causing 404 errors

chrisuthe:
When base_url contains a subpath (e.g., https://link.com/music), extract the path component and prefix all route registrations so the server works correctly behind a reverse proxy with path-based routing. Also redirect bare root to the subpath and mark base_url config as requires_reload since routes are registered at startup.

Fixes music-assistant/support#4940

Tested on HAOS With DEV Plugin
image

**This is a small amount of code with GIANT testing implications**


chrisuthe:
Updated the PR to handle both ingress and reverse proxy. I have tested with both but that does not cover all possible reverse proxy configurations.

music_assistant/helpers/webserver.py
```diff
@@ -182,6 +212,42 @@ async def serve_static(self, file_path: str, request: web.Request) -> web.FileRe
         headers = {"Cache-Control": "no-cache"}
         return web.FileResponse(file_path, headers=headers)
 
+    async def _handle_root_redirect(self, request: web.Request) -> web.Response:
+        """Redirect root path to the base path when a subpath is configured."""
+        raise web.HTTPFound(self._base_path + "/")
+
+    async def _handle_prefixed_catch_all(
```
MarvinSchenkel, line 219: This function introduces quite a bit of duplicate code. Could you refactor this into a function that can be used by both `_handle_prefixed_catch_all` and `_handle_catch_all`?

MarvinSchenkel (changes requested):
Since, as you mentioned, this as quite some test implications, I would love to see some unit tests for this that handle:
- Current use cases (correct, should stay correct)
- Current failing use cases (fail without your fix, pass with your fix applied) 

Marking the PR as draft so we can keep track of things that need our attention. Please mark it as ready for review again when you want us to have another look 🙏

music_assistant/helpers/webserver.py
chrisuthe: created _resolve_dynamic_route taking request and lookup path then made _handle_prefixed and _handle_catch_all both thin wrappers.

chrisuthe:
> Since, as you mentioned, this as quite some test implications, I would love to see some unit tests for this that handle:
> 
> * Current use cases (correct, should stay correct)
> * Current failing use cases (fail without your fix, pass with your fix applied)
> 
> Marking the PR as draft so we can keep track of things that need our attention. Please mark it as ready for review again when you want us to have another look 🙏

I added a suite of tests but _be warned_ they are generated by AI. They all make sense to me who didn't know exactly where to start on unit tests for these changes. LMK If they missed the mark.

music_assistant/helpers/webserver.py
```diff
@@ -81,16 +89,33 @@ async def setup(
...
                 self._webapp[key] = value
         self._apprunner = web.AppRunner(self._webapp, access_log=None, shutdown_timeout=10)
-        # add static routes
+        # add static routes at root level (works for ingress and direct access)
+        # when base_path is set, also register at the prefixed path (for reverse proxy)
         if self._static_routes:
             for method, path, handler in self._static_routes:
                 self._webapp.router.add_route(method, path, handler)
+                if self._base_path:
+                    self._webapp.router.add_route(method, self._base_path + path, handler)
```
marcelveldt, line 98: This is a bit weird tbh - the MA webserver should not have to deal with the fact that it's hosted through a reverse proxy. The base path for MA perspective stays the same. The reverse proxy will rewrite the URL's.

So I think this change is incorrect. Maybe start by describing the exact problem.

chrisuthe: Entirely fair point, I tunneled a little on the experience described, and fixed something that wasn't broken. 

If I look at the webserver configuration here:

image


And I change that to have /Music on the end of it and hit save, what am I telling MA? In what situations would I do such a thing?

I think likely this change should be completely discarded, as we probably wouldn't intend someone to look at that and think "Oh, if I put /Music here, I can then go that URL and access it, assuming the base url resolves to my MA server, and I expect the ingress at HAOS to work properly still as well"

marcelveldt: that base URL is there only if you have proxied the MA webinterface behind a reverse proxy so we can use the correct links and discovery info to the FINAL url. That does however change nothing on MA's webserver itself which is by default simply hosted at the root of the server.

chrisuthe: Yea, this use of it was not intended use case, this "fix" is not needed, it would make future work on pathing more difficult and isn't needed functionality. Closing!

PR #3163, opened 2026-02-14
abs: report correct time_listened in sessions

fmunkes:
The sessions in the abs provider report, apart from the listening position, a time_listened within that session to abs, which goes into its statistics. Since the provider is not aware of the user pausing the stream, we need to reset the time_listened to not report some potentially very long listening time. The same is true if we already have an open session.

Quite a bit in the abs provider is dependent on the 30s reporting interval of the player_queues controller. While I don't expect that value to change, I made it a constant I can use in the abs provider as well.

Related: https://github.com/music-assistant/support/issues/4754

### DIFF
--- a/music_assistant/constants.py
+++ b/music_assistant/constants.py
@@ -934,6 +934,9 @@ SOUNDTRACK_INDICATORS = [
     r"\boriginal.*cast.*recording\b",
 ]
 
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
Thanks @fmunkes 🎉

PR #3164, opened 2026-02-14
Feat/genres-v2-implementation

jozefKruszynski:
# Genre V2: Complete Genre System with Aliases

## Overview

This PR implements a complete overhaul of the genre system in Music Assistant, replacing the legacy `metadata.genres` approach with a flexible, user-managed genre system featuring genre aliases, full CRUD operations, and powerful filtering capabilities.

**Branch:** `feat/genres-v2-implementation`
**Target:** `dev`
**Type:** Feature

---

### 1. Core Genre System

**Genre Management**
- Full CRUD operations for genres (Create, Read, Update, Delete)
- User-created custom genres with metadata support
- 100+ pre-configured default genres with restore functionality
- Genre favorites and search capabilities
- Translation keys for internationalization support

**Alias System**
- Genre aliases for flexible music tagging (e.g., "indie rock", "classic rock" -> "Rock")
- Many-to-many relationships: genres <-> aliases <-> media items
- Automatic self-alias creation (genre automatically gets its own name as an alias)
- Full CRUD operations for aliases
- Reverse lookup: aliases can show which genres they belong to

**Media Integration**
- Genre filtering on tracks, albums, and artists via `genre_ids` parameter
- Multi-genre filtering with OR logic (`genre_ids=[1,2,3]`)
- Transitive relationship: Genre -> Aliases -> Media
- Provider sync integration for automatic genre tagging

### 2. New API Endpoints

**Genre Endpoints** (`music/genres/`)

| Method | Endpoint | Description |
|--------|----------|-------------|
| `GET` | `/count` | Get total genre count |
| `GET` | `/library_items` | Get all genres with filtering/search |
| `GET` | `/get` | Get single genre by ID |
| `GET` | `/overview` | Get genre overview (recommendation rows) |
| `GET` | `/radio_mode_base_tracks` | Get tracks for radio mode |
| `POST` | `/add` | Create new genre (admin) |
| `PUT` | `/update` | Update existing genre (admin) |
| `DELETE` | `/remove` | Delete genre (admin) |
| `POST` | `/add_alias_mapping` | Map alias to genre (admin) |
| `DELETE` | `/remove_alias_mapping` | Unmap alias from genre (admin) |
| `POST` | `/restore_defaults` | Restore default genres (partial or full) (admin) |

**Alias Endpoints** (`music/aliases/`)

| Method | Endpoint | Description |
|--------|----------|-------------|
| `GET` | `/library_items` | Get all aliases with filtering/search |
| `GET` | `/get` | Get single alias by ID |
| `POST` | `/add` | Create new alias (admin) |
| `PUT` | `/update` | Update existing alias (admin) |
| `DELETE` | `/remove` | Delete alias (admin) |
| `POST` | `/add_media_mapping` | Tag media with alias (admin) |
| `DELETE` | `/remove_media_mapping` | Remove alias from media (admin) |
| `POST` | `/promote_to_genre` | Promote alias to parent genre (admin) |

**Media Filtering** (Updated existing endpoints)

All media endpoints now support:
```
genre_ids: int | list[int]  - Filter by genre(s)

Examples:
  music/tracks/library_items?genre_ids=123
  music/albums/library_items?genre_ids=[123,456]
  music/artists/library_items?genre_ids=123
```

### 3. Alias Promotion

**Purpose:** Allows promoting an alias to become a standalone parent genre when it deserves its own top-level category.

**What it does:**
- Creates a new Genre with the alias's name, sort_name, and metadata
- Remaps the alias to belong to the new genre (as its primary alias)
- Preserves all media->alias mappings (no media items are affected)
- Automatically creates a self-alias for the new genre

**Example:**
```
Before:
  Genre: "Rock"
    -> Alias: "indie rock" -> [1000 tracks]

After promotion:
  Genre: "Indie Rock" (newly created)
    -> Alias: "indie rock" -> [1000 tracks] (same media preserved)
```

**Protection:**
- Self-aliases cannot be promoted (would create duplicates)
- All media associations are preserved
- Requires admin role

### 4. Self-Alias Protection

**What are self-aliases?**
Self-aliases are automatically created when a genre is added. They match the genre's name exactly (e.g., "Rock" genre has a "Rock" alias). These are fundamental to the genre system.

**Protection Mechanisms:**

1. **Cannot delete self-aliases** (`music/aliases/remove`)
   - Raises `ValueError` with clear guidance
   - Directs users to delete the parent genre instead
   - Non-self-aliases can be deleted normally

2. **Cannot unlink genre from its own self-alias** (`music/genres/remove_alias_mapping`)
   - Raises `ValueError` explaining self-aliases cannot be unlinked
   - Prevents breaking the genre<->self-alias relationship
   - Unlinking self-aliases from *other* genres is allowed (edge case)

3. **Cannot promote self-aliases** (`music/aliases/promote_to_genre`)
   - Raises `ValueError` to prevent creating duplicate genres
   - Non-self-aliases can be promoted normally

**Implementation:**
```python
# Check if alias name matches parent genre name (case-insensitive)
if genre.name.lower() == alias.name.lower():
    raise ValueError("Cannot [delete/unlink/promote] self-alias...")
```

### 5. Genre Overview Endpoint

**Purpose:** Optimized single-call solution for building genre detail pages.

**What it returns:**
- Pre-organized recommendation rows (similar to Netflix/Spotify UI)
- Up to 7 media type rows: Artists, Albums, Tracks, Playlists, Radio, Podcasts, Audiobooks
- Empty rows automatically filtered out
- Browse paths included for "View All" navigation
- Consistent limits across all media types

**Benefits:**
- 7x fewer API calls — single request instead of 7 separate queries
- Pre-filtered — empty rows automatically removed
- Pre-organized — RecommendationFolder structure ready for UI
- Browse paths — "View All" links included
- Faster page load — single round-trip to backend

**Example Usage:**
```typescript
const overview = await api.call("music/genres/overview", {
  item_id: genreId,
  provider_instance_id_or_domain: "library"
});

// Returns array of RecommendationFolder objects ready to render
overview.forEach(row => {
  renderMediaRow(row.label, row.items, row.path);
});
```

### 6. Reverse Lookup for Aliases

**Feature:** GenreAlias objects now include a `genres` field showing parent genres.

**Use Cases:**
- Admin interfaces showing alias -> genre relationships
- Genre page search showing "indie rock -> Rock"
- Debugging genre mapping issues
- Understanding genre structure when managing library

**Implementation:**
```python
async def _attach_genres(self, aliases: list[GenreAlias]) -> None:
    """Attach parent genres to alias objects by querying mapping tables."""
    # Queries genre_alias_mapping + genres tables
    # Populates alias.genres field
```

**Benefits:**
- Always populated on GenreAlias objects
- Minimal performance impact (one extra query per alias fetch)
- Helps users understand and organize their genre structure

### 7. Restore Default Genres (Dual Mode)

**Feature:** The `music/genres/restore_defaults` endpoint supports two restore modes for different use cases.

**Partial Restore (Default - Non-destructive):**
- Adds any missing default genres from `genre_mapping.json`
- Ensures self-aliases exist for all existing genres
- Ensures configured aliases exist for all existing genres
- Does NOT delete or modify existing genres
- Safe to run multiple times (idempotent)

```python
await api.call("music/genres/restore_defaults", {
    full_restore: False  # or omit (defaults to False)
})
```

**Full Restore (Destructive):**
- Deletes ALL existing genres and aliases
- Deletes ALL genre→media associations
- Recreates everything from default mapping
- Essentially a "reset to factory defaults"
- Requires explicit user confirmation in UI

```python
await api.call("music/genres/restore_defaults", {
    full_restore: True
})
```

**Use Cases:**
- **Partial:** First-run setup, recovering deleted defaults, adding new defaults after updates
- **Full:** Corrupted genre library, user wants fresh start, testing/development

**UI Integration:**
- Partial restore: Single confirmation, shown in welcome flow and settings
- Full restore: Double confirmation with typed text, hidden behind "Advanced" toggle
- See [Use Case #8 in Frontend Guide](genre-v2-frontend-guide.md#8-system-settings-genre-management)


music_assistant/controllers/media/albums.py
```diff
@@ -113,6 +113,7 @@ async def library_items(
         offset: int = 0,
         order_by: str = "sort_name",
         provider: str | list[str] | None = None,
+        genre_ids: int | list[int] | None = None,
```
marcelveldt: if we want to be consistent with the provider filter above, we may want to name this just genre ?

jozefKruszynski: Makes sense, I'll also update that this evening

music_assistant/controllers/media/base.py
```diff
@@ -968,6 +997,17 @@ def _apply_filters(
         if favorite is not None:
             query_parts.append(f"{self.db_table}.favorite = :favorite")
             query_params["favorite"] = favorite
+        # handle genre filter
+        if genre_ids:
+            query_params["genre_ids"] = genre_ids
+            query_params["genre_media_type"] = self.media_type.value
+            query_parts.append(
```
MarvinSchenkel, line 1021: It's more performant to rewrite this as a `WHERE EXISTS` + `JOIN` to the other tables. This also makes use of the indexes you defined on the table. So:
```
WHERE EXISTS(
  SELECT 1
  FROM alias_media_item_mapping ami
  INNER JOIN genre_alias_mapping gam
    ON gam.alias_id = ami.alias_id
  WHERE ami.media_id = {self.db_table}.item_id
    AND ami.media_type = :genre_media_type
    AND gam.genre_id IN :genre_ids
)
```

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+                DB_TABLE_ALIAS_MEDIA_ITEM_MAPPING,
+                {
+                    "media_type": media_type.value,
+                    "media_id": media_id_int,
+                    "alias_id": alias_id,
+                },
+                allow_replace=True,
+            )
+
+    async def _attach_aliases(self, genres: list[Genre]) -> None:
```
MarvinSchenkel: Just wondering if we really need this function? We should just extend the base query and populate `aliases` as a json field using `JSON_GROUP_ARRAY` and deserialize the entire Genre object in one go. We do the same for artists in the Album query for example

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+                f"{'DESC' if order_by.endswith('desc') else 'ASC'}"
+            )
+        rows = await self.mass.music.database.get_rows_from_query(
+            sql_query, query_params, limit=limit, offset=offset
+        )
+        items = [Genre.from_dict(self._parse_genre_row(row)) for row in rows]
+        await self._attach_aliases(items)
+        return items
+
+    async def get_library_item(self, item_id: int | str) -> Genre:
```
MarvinSchenkel: I think we can remove this one and use the base version when we have implemented the aliases in the base query

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+            {"genre_id": genre_id, "alias_id": alias_id},
+            allow_replace=True,
+        )
+
+    async def _get_description(self, item_id: int) -> str | None:
+        if db_row := await self.mass.music.database.get_row(DB_TABLE_GENRES, {"item_id": item_id}):
+            return dict(db_row).get("description")
+        return None
+
+    def _parse_genre_row(self, db_row: Any) -> dict[str, Any]:
```
MarvinSchenkel: I think once we move aliases as inline to the base query, we can remove both parse_genre_row and parse_alias_row and rely on base.parse_db_row?

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+        self,
+        item: Genre,
+        limit: int = 0,
+        offset: int = 0,
+    ) -> tuple[list[Track], list[Album], list[Artist]]:
+        """Return tracks, albums, and artists mapped to aliases for a genre."""
+        db_id = int(item.item_id)
+        subquery = f"SELECT alias_id FROM {DB_TABLE_GENRE_ALIAS_MAPPING} WHERE genre_id = :genre_id"
+
+        track_query = (
```
MarvinSchenkel, line 324: These can also be rewritten with EXISTS for better performance / index usage

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+            (MediaType.TRACK, "Tracks"),
+            (MediaType.PLAYLIST, "Playlists"),
+            (MediaType.RADIO, "Radio"),
+            (MediaType.PODCAST, "Podcasts"),
+            (MediaType.AUDIOBOOK, "Audiobooks"),
+        ]
+        rows: list[RecommendationFolder] = []
+        for media_type, title in media_rows:
+            ctrl = self.mass.music.get_controller(media_type)
+            query = (
```
MarvinSchenkel, line 424: This can also be rewritten with EXISTS for better performance / index usage

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+        return [await self.get_library_item(item_id) for item_id in created_ids]
+
+    async def remove_item_from_library(self, item_id: str | int, recursive: bool = True) -> None:
+        """Delete genre record from the database."""
+        db_id = int(item_id)
+        item = await self.get_library_item(db_id)
+        await self.mass.music.database.delete(DB_TABLE_GENRE_ALIAS_MAPPING, {"genre_id": db_id})
+        await self.mass.music.database.delete(DB_TABLE_GENRES, {"item_id": db_id})
+        self.mass.signal_event(EventType.MEDIA_ITEM_DELETED, item.uri, item)
+        self.logger.debug("deleted genre with id %s from database", db_id)
```
MarvinSchenkel: Think we also need to call `await super().remove_item_from_library()` here?

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+                ),
+                "search_name": create_safe_string(name, True, True),
+                "search_sort_name": create_safe_string(sort_name or "", True, True),
+                "timestamp_added": UNSET,
+            },
+        )
+        await self._ensure_self_alias(db_id, name)
+        self.logger.debug("updated %s in database: (id %s)", update.name, db_id)
 
     async def search(
```
MarvinSchenkel: I think we can also remove this and rely on the base implementation after fixing the base query

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1345 @@
...
+        :param offset: Number of items to skip.
+        :param order_by: Sort order (name, name_desc, sort_name, sort_name_desc).
+        :param provider: Filter by provider (not used for genres).
+        :param genre_ids: NOT SUPPORTED - Filtering genres by genres doesn't make sense.
+        """
+        if genre_ids is not None:
+            msg = "genre_ids parameter is not supported for Genre.library_items()"
+            raise ValueError(msg)
+
+        query_parts: list[str] = []
```
MarvinSchenkel: I think that once we have the aliases in the base query, we can also rewrite this as: 
```
return await super().library_items(
        favorite=favorite,
        search=search,
        limit=limit,
        offset=offset,
        order_by=order_by,
        provider=provider,
    )
```

music_assistant/controllers/player_queues.py
```diff
@@ -1597,6 +1607,34 @@ async def get_album_tracks(self, album: Album, start_item: str | None) -> list[T
             result.append(album_track)
         return result
 
+    async def get_genre_tracks(self, genre: Genre, start_item: str | None) -> list[Track]:
```
MarvinSchenkel, line 1622: I think this can potentially blow up for broad genres. Should we implement some limiting + randomization in here? (just thinking out loud)

jozefKruszynski: Sounds like a sensible plan

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1198 @@
...
+        for media_type, title in media_rows:
+            ctrl = self.mass.music.get_controller(media_type)
+            query = (
+                f"EXISTS(SELECT 1 FROM {ami} ami "
+                f"INNER JOIN {gam} gam ON gam.alias_id = ami.alias_id "
+                f"WHERE ami.media_id = {ctrl.db_table}.item_id "
+                "AND ami.media_type = :media_type "
+                "AND gam.genre_id = :genre_id)"
+            )
+            items = await ctrl.get_library_items_by_query(
```
MarvinSchenkel, line 430: We should rewrite this using `asyncio.gather` to prevent sequential db queries

music_assistant/controllers/music.py
```diff
@@ -2222,6 +2227,214 @@ async def __migrate_database(self, prev_version: int) -> None:  # noqa: PLR0915
...
+            # 2a: Extract all unique raw genre names from metadata via json_each
+            union_parts = [
+                f"SELECT DISTINCT TRIM(g.value) AS raw_name "
+                f"FROM {table}, json_each(json_extract({table}.metadata, '$.genres')) AS g "
+                f"WHERE json_extract({table}.metadata, '$.genres') IS NOT NULL "
+                f"AND json_extract({table}.metadata, '$.genres') != '[]'"
+                for table, _ in media_tables
+            ]
+            unique_names_sql = " UNION ".join(union_parts)
+            print(f"Genre migration - unique names query:\n{unique_names_sql}\n")  # noqa: T201
```
MarvinSchenkel: Can we remove all these prints and replace them with logger.INFO? Might be nice to give some feedback during the migration as this might be one of the bigger migrations

jozefKruszynski: I'm an idiot. I meant to change them immediately after we spoke about it yesterday. 🤦‍♂️

music_assistant/controllers/media/genres.py
jozefKruszynski: Will do so

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1198 @@
...
+        sort_name = name
+        search_name = create_safe_string(name, True, True)
+        if not search_name:
+            return None
+        search_sort_name = create_safe_string(sort_name or "", True, True)
+        return name, sort_name, search_name, search_sort_name
+
+    def _on_sync_tasks_updated(self, event: MassEvent) -> None:
+        """Trigger genre mapping scan when all sync tasks complete."""
+        if not event.data:  # empty list = all syncs done
```
MarvinSchenkel: I think it's safer to use `self.mass.music.in_progress_syncs` because theoretically another sync could have started after event dispatch? Also set self._scanner_running here to be safe. So something like:
```
if self.mass.music.in_progress_syncs or self._scanner_running:
        return
    self._scanner_running = True
```

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1198 @@
...
+            self.mass.create_task(self._scan_genre_mappings())
+
+    async def _scan_genre_mappings(self) -> None:
+        """Scan media items with metadata.genres and map them to genre aliases.
+
+        Triggered after library sync completes or via manual API call.
+        Processes all unmapped items across tracks, albums, and artists.
+        """
+        if self._scanner_running:
```
MarvinSchenkel: Then we can replace this with another check to be sure we are firing after the last sync:

```
if self.mass.music.in_progress_syncs:
        self.logger.debug("Syncs still in progress, deferring genre scan")
        self._scanner_running = False
        return
```

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1198 @@
...
+                {
+                    "name": name_value,
+                    "sort_name": sort_name,
+                    "favorite": 0,
+                    "metadata": serialize_to_json({}),
+                    "external_ids": serialize_to_json(set()),
+                    "play_count": 0,
+                    "last_played": 0,
+                    "search_name": search_name,
+                    "search_sort_name": search_sort_name,
```
MarvinSchenkel, line 1066: Think we are missing `"timestamp_added": UNSET,` here

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1198 @@
...
+            "AND ami.media_type = 'album' AND gam.genre_id = :genre_id)"
+        )
+        artist_query = (
+            f"EXISTS(SELECT 1 FROM {ami} ami "
+            f"INNER JOIN {gam} gam ON gam.alias_id = ami.alias_id "
+            "WHERE ami.media_id = artists.item_id "
+            "AND ami.media_type = 'artist' AND gam.genre_id = :genre_id)"
+        )
+
+        tracks = await self.mass.music.tracks.get_library_items_by_query(
```
MarvinSchenkel: We can also paralellize these three queries with `asyncio.gather`

music_assistant/controllers/player_queues.py
```diff
@@ -1597,6 +1607,39 @@ async def get_album_tracks(self, album: Album, start_item: str | None) -> list[T
...
+        """Return tracks for given genre, based on alias mappings.
+
+        Limits results to avoid loading thousands of tracks for broad genres.
+        Directly mapped tracks are fetched with random ordering, then supplemented
+        with tracks from a limited set of mapped albums and artists.
+        """
+        result: list[Track] = []
+        start_item_found = False
+        self.logger.info(
```
MarvinSchenkel, line 1631: debug?

jozefKruszynski: The other functions also log as info for this, we could either change all of them or none.

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1080 @@
...
+                    "translation_key": entry.get("translation_key"),
+                    "description": None,
+                    "favorite": 0,
+                    "metadata": serialize_to_json({}),
+                    "external_ids": serialize_to_json(set()),
+                    "genre_aliases": serialize_to_json(all_aliases),
+                    "play_count": 0,
+                    "last_played": 0,
+                    "search_name": search_name,
+                    "search_sort_name": search_sort_name,
```
MarvinSchenkel, line 506:
```suggestion
                    "search_sort_name": search_sort_name,
                    "timestamp_added": UNSET,
```

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1080 @@
...
+        existing = set(genre.genre_aliases) if genre.genre_aliases else set()
+        updated = existing | set(aliases)
+        if updated != existing:
+            await self.mass.music.database.update(
+                self.db_table,
+                {"item_id": genre_id},
+                {"genre_aliases": serialize_to_json(list(updated))},
+            )
+
+    async def _ensure_genre_for_alias(self, name: str) -> int | None:
```
MarvinSchenkel: Shouldn't this return a list[int]? Since we flatten all the hiarchies, a certain alias can actually be mapped to two genres. E.g. a user wants to extract a 2nd level genre into its own genre, then all 3rd levels below that will be linked to both the top level and the 2nd level. Example: Pop > Pop 2 > Pop 3, If the user wants Pop 2 as it's own genre, then we will have flattened genres of Pop > Pop 2 > Pop 3 and Pop 2 > Pop 3.

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1080 @@
...
+        self._scanner_running = True
+        self.mass.create_task(self._scan_genre_mappings())
+
+    async def _scan_genre_mappings(self) -> None:
+        """Scan media items with metadata.genres and map them to genres.
+
+        Triggered after library sync completes or via manual API call.
+        Callers must set _scanner_running = True before calling this method.
+        """
+        self._last_scan_time = time.time()
```
MarvinSchenkel, line 1112:
```suggestion
        if self.mass.music.in_progress_syncs:
            self.logger.debug("Syncs still in progress, deferring genre scan")
            self._scanner_running = False
            return
        self._last_scan_time = time.time()
```

music_assistant/controllers/music.py
```diff
@@ -2581,6 +2823,15 @@ async def __create_database_indexes(self) -> None:
             f"CREATE INDEX IF NOT EXISTS {DB_TABLE_SMART_FADES_ANALYSIS}_idx "
             f"on {DB_TABLE_SMART_FADES_ANALYSIS}(item_id,provider,fragment);"
         )
+        # indexes on genre_media_item_mapping table
+        await self.database.execute(
+            f"CREATE INDEX IF NOT EXISTS {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}_media_idx "
+            f"on {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}(media_type,media_id);"
```
MarvinSchenkel: Better query coverage:
```suggestion
            f"on {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}(media_id,media_type);"
```

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1159 @@
...
+    async def remove_alias(self, genre_id: str | int, alias: str) -> Genre:
+        """Remove an alias string from a genre.
+
+        :param genre_id: Database ID of the genre.
+        :param alias: Alias string to remove.
+        :raises ValueError: If trying to remove the genre's own name.
+        """
+        db_id = int(genre_id)
+        genre = await self.get_library_item(db_id)
+        if alias.lower() == genre.name.lower():
```
MarvinSchenkel: Shouldn't we use `create_safe_string` here as well when comparing?

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1159 @@
...
+        original genre.
+
+        :param genre_id: Database ID of the source genre.
+        :param alias: The alias string to promote.
+        :return: The newly created Genre.
+        """
+        db_genre_id = int(genre_id)
+        source_genre = await self.get_library_item(db_genre_id)
+
+        if alias.lower() == source_genre.name.lower():
```
MarvinSchenkel: Shouldn't we use create_safe_string here as well when comparing?

tests/core/test_genres.py
```diff
@@ -0,0 +1,866 @@
...
+    DB_TABLE_GENRES,
+    DEFAULT_GENRE_MAPPING,
+)
+from music_assistant.controllers.media.genres import GenreController
+from music_assistant.mass import MusicAssistant
+
+# ---------------------------------------------------------------------------
+# Fixtures & helpers
+# ---------------------------------------------------------------------------
+
```
MarvinSchenkel, line 35: I think we are missing a test here that tests 1 alias mapping to 2 genres?

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1159 @@
...
+            )
+            for row in rows:
+                gid = int(row["item_id"])
+                if gid not in found_ids:
+                    found_ids.append(gid)
+
+            # Fallback: normalize each alias with create_safe_string and compare.
+            # Handles cases like "Synthpop" matching "synth-pop" where LOWER
+            # alone can't bridge the difference (hyphens, spaces, unicode).
+            if not found_ids:
```
MarvinSchenkel: I think this short circuits some valid use cases, e.g. genre A > synthpop, genre B > synth-pop. With the current logic, it would only return genre A

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1159 @@
...
+                    if genre_id not in alias_to_genre[norm]:
+                        alias_to_genre[norm].append(genre_id)
+
+        # Extract unique raw genre names only from unmapped items
+        union_parts = [
+            f"SELECT DISTINCT TRIM(g.value) AS raw_name "
+            f"FROM {table}, json_each(json_extract({table}.metadata, '$.genres')) AS g "
+            f"WHERE json_extract({table}.metadata, '$.genres') IS NOT NULL "
+            f"AND json_extract({table}.metadata, '$.genres') != '[]' "
+            f"AND {table}.item_id NOT IN "
```
MarvinSchenkel: I think this wont work for incremental changes to a media item, because the NOT IN will fail if there is already a mapping for a media item. I think we need to change this to a NOT EXISTS based on both item_id, media_type  and alias value.

music_assistant/controllers/media/genres.py
```diff
@@ -1,64 +1,1159 @@
...
+        for table, media_type in media_tables:
+            full_query = (
+                f"{cte} INSERT OR REPLACE INTO {gm}"
+                f"(genre_id, media_id, media_type, alias) "
+                f"SELECT gl.genre_id, {table}.item_id, "
+                f"'{media_type.value}', TRIM(g.value) "
+                f"FROM {table}, "
+                f"json_each(json_extract({table}.metadata, '$.genres')) AS g "
+                f"JOIN genre_lookup gl ON gl.raw_name = LOWER(TRIM(g.value)) "
+                f"WHERE json_extract({table}.metadata, '$.genres') IS NOT NULL "
```
MarvinSchenkel, line 751: I think here as well? But based on (genre_id, media_id, media_type)

MarvinSchenkel (approved):
Awesome job @jozefKruszynski 🎉 👏

PR #3165, opened 2026-02-14
Expand PIN based auth in airplay 2

hmonteiro:
This MR adds PIN based authentication to devices other than AppleTV MacOS.
I spent some time testing owntone and pyatv with my device (an LG TV) and realised that we need very few changes to broaden support.

I had a look at pyatv and adapted their pairing requirement test to MA.

This is my first contribution to the repo, so I am sorry in advance for any issues 🙂 

Also, I only own one Airplay compatible device, so this probably needs to be tested by a few people with different devices before it can be merged.


music_assistant/providers/airplay/player.py
```diff
@@ -245,17 +247,23 @@ async def get_config_entries(
 
         return base_entries
 
+    def _get_flags(self) -> int:
+        # Flags are either present via "sf" or "flags. Taken from pyatv.protocols.airplay.utils"
+        if self.airplay_discovery_info is None:
+            return 0
+        flags = (
+            self.airplay_discovery_info.properties.get(b"sf")
```
marcelveldt: what about raop only devices ? they will only have raop discovery info

hmonteiro: good point, fix incoming

hmonteiro: I have addressed your comment. As I mentioned in the description, I don't have an actual device that exercises that branch, so I would rely on other people to confirm it still works. From what I understand this logic is used for both airplay v1 and v2 in pyatv, so I am hopefull it will be ok.

MarvinSchenkel:
@hmonteiro I feel this is ready for review, but it's still marked as draft?

MarvinSchenkel (approved):
Thanks @hmonteiro , great addition 👏

PR #3166, opened 2026-02-14
Improve track retrieval for Soundcloud artists with fallback

robsonke:
Solves [#4936](https://github.com/music-assistant/support/issues/4936)


music_assistant/providers/soundcloud/__init__.py
```diff
@@ -338,11 +338,38 @@ async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> lis
...
+        if not collection:
+            try:
+                popular_tracks_obj = await self._soundcloud.get_popular_tracks_user(
+                    prov_artist_id, 100
+                )
+                if popular_tracks_obj:
+                    collection = popular_tracks_obj.get("collection") or popular_tracks_obj.get(
+                        "items"
+                    )
+            except Exception as error:
```
MarvinSchenkel: Can we narrow down this Exception to the one you are trying to catch here?

MarvinSchenkel:
Just had a quick look and it seems that `get_similar_tracks` can benefit from this fix too, as it is just blindly iterating over `tracks_obj["collection"]`. Could you apply it there as well ?

music_assistant/providers/soundcloud/__init__.py
robsonke: Done, thanks!

robsonke:
> Just had a quick look and it seems that `get_similar_tracks` can benefit from this fix too, as it is just blindly iterating over `tracks_obj["collection"]`. Could you apply it there as well ?

I'm not completely sure but it is a safe fallback, I've added it here as well.

robsonke:
Argh, forgot I already had an open pr, now pushed another fix in the same. Trying to revert.

MarvinSchenkel (approved):
Looks good, thanks @robsonke 🙏

PR #3167, opened 2026-02-15
Update Alexa player provider

alams154:
This PR should bring some improvements to the Alexa Player Provider and help bring it more in line with the current state of the Alexa skill.

- Change default port to 5000 to match the default Alexa skill port
- Add api_request function for generic endpoints
- Pull metadata that allows for display of metadata on Show devices
- Migrate to custom commands in AlexaPy calls since the API calls do not work for APL devices


music_assistant/providers/alexa/__init__.py
```diff
@@ -244,6 +245,97 @@ async def delete_cookie(cookiefile: str) -> None:
...
+
+    auth = None
+    if username is not None and password is not None:
+        auth = BasicAuth(str(username), str(password))
+
+    api_url = provider.config.get_value(CONF_API_URL) or ""
+    url = f"{api_url.rstrip('/')}/{endpoint.lstrip('/')}"
+
+    mass_session = None
+    try:
```
MarvinSchenkel: I think you can just rely on mass.http_session to be there. No need for all this fallback logic.

music_assistant/providers/alexa/__init__.py
```diff
@@ -314,47 +433,32 @@ async def volume_set(self, volume_level: int) -> None:
...
         """Handle PLAY MEDIA on the player."""
-        username = self.provider.config.get_value(CONF_API_BASIC_AUTH_USERNAME)
-        password = self.provider.config.get_value(CONF_API_BASIC_AUTH_PASSWORD)
-
-        auth = None
-        if username is not None and password is not None:
-            auth = BasicAuth(str(username), str(password))
-
         stream_url = await self.provider.mass.streams.resolve_stream_url(self.player_id, media)
+        # Prefer the player's current_media (may contain enriched metadata),
```
MarvinSchenkel: Why prefer the current metadata? If MA is playing track A and you request play_media with Track B, the metadata of track B will be in `media`. I think you should just use `media.xxx or ""` here ?

music_assistant/providers/alexa/__init__.py
```diff
@@ -244,6 +245,97 @@ async def delete_cookie(cookiefile: str) -> None:
...
+        resp_text = await resp.text()
+        if resp.status < 200 or resp.status >= 300:
+            msg = f"Failed API request to {url}: Status code: {resp.status}, Response: {resp_text}"
+            _LOGGER.error(msg)
+            raise ActionUnavailable(msg)
+        return resp_text
+
+
+async def api_request(
+    provider: Any,
```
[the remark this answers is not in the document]
MarvinSchenkel: Please use appropriate type here

music_assistant/providers/alexa/__init__.py
```diff
@@ -432,3 +538,50 @@ async def loaded_in_mass(self) -> None:
...
                 player = AlexaPlayer(self, player_id, device_object)
                 await self.mass.players.register_or_update(player)
+
+        # Prefetch intents so players can use cached utterances
+        try:
+            await self._load_intents()
+        except Exception:  # don't fail provider load on intent fetch issues
+            _LOGGER.debug("Could not prefetch Alexa intents, will lazy-load on use.")
+
+    async def _load_intents(self) -> None:
```
MarvinSchenkel, line 545: Can we add an `@lock` decorator here to prevent concurrent calls?

music_assistant/providers/alexa/__init__.py
```diff
@@ -181,7 +182,7 @@ async def proxy_handler(request: web.Request) -> Any:
             key=CONF_API_URL,
             type=ConfigEntryType.STRING,
             label="API Url",
-            default_value="http://localhost:3000",
+            default_value="http://localhost:5000",
```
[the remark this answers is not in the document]
MarvinSchenkel: Can you explain why we need to change the port number?

MarvinSchenkel (changes requested):
Could you please update the PR description with the goal? I miss the context to properly review the functional changes of the player commands. What is this PR improving?

Also gave some inline feedback. Please have a look, address the comments and let us know when you want us to have another look by selecting 'Ready for review' again

music_assistant/providers/alexa/__init__.py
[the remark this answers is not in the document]
alams154: This is to bring the provider more in line with the skill since the default port on the skill has been changed to 5000.

music_assistant/providers/alexa/__init__.py
alams154: I thought that would be the case but when I was testing, media didn't contain relevant metadata.  Album and artist were null and title was set to "Music Assistant" and image_url was the Music Assistant logo.

If there is some other field that contains the relevant metadata, I can update the assignments, or if there is a bug in the provider somewhere I could address that.

music_assistant/providers/alexa/__init__.py
alams154: Should be added now

music_assistant/providers/alexa/__init__.py
alams154: Should be removed now

music_assistant/providers/alexa/__init__.py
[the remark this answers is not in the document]
alams154: Should be the appropriate type now

music_assistant/providers/alexa/__init__.py
MarvinSchenkel: This is due to flow mode playing as 1 big continuous stream, so `play_media` is only called once in those cases. If you want rich metadata in flow mode, you need an endpoint on the player side that allows you to _just_ update the metadata. You can implement `_on_player_media_updated` for this. Have a look at the Airplay or Chromecast provider as an example.

music_assistant/providers/alexa/__init__.py
```diff
@@ -244,6 +246,89 @@ async def delete_cookie(cookiefile: str) -> None:
...
+    mass_session = provider.mass.http_session
+
+    if mass_session:
+        try:
+            return await _request_with_session(mass_session, method, url, json_data, timeout, auth)
+        except Exception as exc:  # pylint: disable=broad-except
+            _LOGGER.error("Failed API request to %s: %s", url, exc)
+            raise ActionUnavailable("Failed to connect to the configured API endpoint.")
+
+    async with aiohttp.ClientSession() as session:
```
MarvinSchenkel: I think this is practically dead code. I cannot think of a reason that `mass.http_session` is None? Have you encountered this in your development.

music_assistant/providers/alexa/__init__.py
```diff
@@ -244,6 +246,89 @@ async def delete_cookie(cookiefile: str) -> None:
...
+
+    api_url = str(provider.config.get_value(CONF_API_URL) or "")
+    url = f"{api_url.rstrip('/')}/{endpoint.lstrip('/')}"
+
+    mass_session = provider.mass.http_session
+
+    if mass_session:
+        try:
+            return await _request_with_session(mass_session, method, url, json_data, timeout, auth)
+        except Exception as exc:  # pylint: disable=broad-except
```
MarvinSchenkel: Would be nice to narrow this down to specific Exceptions you are trying to catch here.

MarvinSchenkel: I see more of those "catch all" patterns. I think that, in combination with my comments above is all that's left. Then we should be good to merge this :)

alams154: I think I've removed the extra catch all patterns now.  The original exception from api_request should propagate up now.

music_assistant/providers/alexa/__init__.py
alams154: No, I haven't.  Should be removed now.

music_assistant/providers/alexa/__init__.py
alams154: This is still not fully implemented on the device/skill side but some placeholder pushes to the skill side API are in place now.

The initial loading on the device will not get the metadata but sending commands in the same stream will cause the metadata to load in.

alams154: The latest version of the skill now supports this and updates the metadata appropriately.

MarvinSchenkel (approved):
Nice improvements, thanks @alams154 👏

PR #3169, opened 2026-02-15
Adapt playlog for all users if no user present

fmunkes:
I get a `sqlite3.IntegrityError: NOT NULL constraint failed: playlog.userid` reproducibly when I start MA (as a single user instance) in the background, and then get an externally triggered progress update.

The code already says that we are syncing for all users then, but I think the table expects a userid.


MarvinSchenkel (approved):
Thanks @fmunkes , looks indeed the code does not match the comment, good catch. 

Issue: https://github.com/music-assistant/support/issues/4629

PR #3077, opened 2026-02-02
feat(deezer): add Flow, Radios and Top Charts as recommendations

sfortis:
## Summary

   Restructures Deezer recommendations to include personalized content:

   - **Made for you** folder containing:
     - Flow (personalized endless radio)
     - Recommended tracks
     - Top Charts (global top 100)
     - Deezer's recommended playlists
   - **Deezer Radios** folder with 14 curated genre radios (Hits, Pop, Rock, Metal, Jazz, etc.)
   - **Recommended albums** folder
   - **Recommended artists** folder

image

image

image

image

Flow, Recommended tracks, Top Charts, and Radios are implemented as virtual playlists (`in_library=false`) since they represent dynamic content.

## Test Plan

- Tested on dev environment, working as expected

**Suggested label:** `new-feature`

PR #3171, opened 2026-02-15
Deezer: Add Genre and mood flows

jdaberkow:
## Summary

Extends the Deezer provider with mood and genre Flow playlists (Happy, Chill, Focus, Party, Rock, Metal, etc.) building on top of #3077.

These Flow variants are only available through Deezer's unofficial GW API (`radio.getUserRadio`), so the changes go into `gw_client.py` and `__init__.py`.

Available Flows are discovered dynamically via the `page.get` GW endpoint instead of being hardcoded, so user-specific variations are picked up automatically.

The recommendations view now includes two new folders:
- **Deezer Mood Flows**: Happy, Chill, Focus, Melancholy, Party, Love, Motivation
- **Deezer Genre Flows**: Rock, Metal, Electronic, Classical, etc. (varies by user)

image

Also wraps `get_user_recommended_albums` and `get_user_recommended_artists` in try/except since the Deezer API occasionally returns errors on those endpoints, which would otherwise prevent all recommendations from loading.

## Test Plan

- Tested locally, mood and genre Flows show up in recommendations and play back correctly
- Verified that recommendations still load when `get_user_recommended_artists` returns an error

Suggested label: `enhancement`

### DIFF
--- a/music_assistant/providers/deezer/__init__.py
+++ b/music_assistant/providers/deezer/__init__.py
@@ -102,28 +102,6 @@ TOP_CHARTS_PLAYLIST_ID = "top_charts"
 RADIO_PLAYLIST_PREFIX = "radio_"
 MOOD_FLOW_PREFIX = "mood_flow_"
 
-# Mood-based Flow variants (personalized, always available)
-MOOD_FLOWS = [
-    ("happy", "Flow: Happy"),
-    ("motivation", "Flow: Motivation"),
-    ("party", "Flow: Party"),
-    ("chill", "Flow: Chill"),
-    ("melancholy", "Flow: Melancholy"),
-    ("you_and_me", "Flow: Love"),
-    ("focus", "Flow: Focus"),
-]
-
-# Genre-based Flow variants (may vary by region/user)
-GENRE_FLOWS = [
-    ("genre-rock", "Flow: Rock"),
-    ("genre-metal", "Flow: Metal"),
-    ("genre-electronic", "Flow: Electronic"),
-    ("genre-classical", "Flow: Classical"),
-    ("genre-danceedm", "Flow: Dance & EDM"),
-    ("genre-house", "Flow: House"),
-    ("genre-lofi", "Flow: Lofi"),
-]
-
 # Curated Deezer radio station IDs
 CURATED_RADIO_IDS = [
     37151,  # Hits
@@ -268,6 +246,26 @@ class DeezerProvider(MusicProvider):
         """
         return await self.gw_client.get_user_radio(config_id)
 
+    @use_cache(3600 * 24)  # Cache for 24 hours
+    async def _get_available_flows(self) -> list[tuple[str, str, str | None]]:
+        """Discover available mood/genre Flow variants from the Deezer home page.
+
+        Genre flows have config_ids starting with 'genre-'.
+        Returns a list of (config_id, display_name, cover_url) tuples.
+        """
+        items = await self.gw_client.get_home_flows()
+        flows: list[tuple[str, str, str | None]] = []
+        for item in items:
+            config_id = item["data"]["id"]
+            if config_id == "default":
+                continue
+            title = f"Flow: {item['title']}"
+            cover_url = None
+            if pictures := item.get("pictures"):
+                cover_url = f"https://e-cdns-images.dzcdn.net/images/misc/{pictures[0]['md5']}/264x264-000000-80-0-0.jpg"
+            flows.append((config_id, title, cover_url))
+        return flows
+
     @use_cache(3600 * 24 * 7)  # Cache for 7 days
     async def search(
         self, search_query: str, media_types: list[MediaType], limit: int = 5
@@ -397,9 +395,10 @@ class DeezerProvider(MusicProvider):
                 raise MediaNotFoundError(f"Radio {prov_playlist_id} not found on Deezer") from err
         if prov_playlist_id.startswith(MOOD_FLOW_PREFIX):
             config_id = prov_playlist_id.removeprefix(MOOD_FLOW_PREFIX)
-            flow_names = dict(MOOD_FLOWS + GENRE_FLOWS)
-            display_name = flow_names.get(config_id, f"Flow: {config_id}")
-            return self._create_virtual_playlist(prov_playlist_id, display_name)
+            all_flows = await self._get_available_flows()
+            flow_info = {cid: (name, cover) for cid, name, cover in all_flows}
+            name, cover_url = flow_info.get(config_id, (f"Flow: {config_id}", None))
+            return self._create_virtual_playlist(prov_playlist_id, name, image_url=cover_url)
         try:
             return self.parse_playlist(
                 playlist=await self.client.get_playlist(playlist_id=int(prov_playlist_id)),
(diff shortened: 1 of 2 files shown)


music_assistant/providers/deezer/__init__.py
```diff
@@ -569,32 +609,64 @@ async def recommendations(self) -> list[RecommendationFolder]:
...
+                    RecommendationFolder(
+                        item_id="recommended_artists",
+                        provider=self.instance_id,
+                        name="Recommended artists",
+                        items=UniqueList(
+                            [self.parse_artist(artist=artist) for artist in recommended_artists]
+                        ),
+                    )
+                )
+        except Exception as err:
```
MarvinSchenkel: Could we narrow this down a bit to the exceptions we expect here?

music_assistant/providers/deezer/__init__.py
```diff
@@ -423,6 +458,11 @@ async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> lis
                 self.logger.debug("Failed to get radio tracks %s: %s", radio_id, err)
                 return []
 
+        if prov_playlist_id.startswith(MOOD_FLOW_PREFIX):
+            config_id = prov_playlist_id.removeprefix(MOOD_FLOW_PREFIX)
+            gw_tracks = await self._get_mood_flow_tracks(config_id)
+            return [await self.get_track(track["SNG_ID"]) for track in gw_tracks]
```
MarvinSchenkel: This calls the API for every single track in the list. Is there really no other way to do this?

music_assistant/providers/deezer/gw_client.py
```diff
@@ -127,6 +130,38 @@ async def _gw_api_call(
...
+    async def get_user_radio(self, config_id: str) -> list[dict[str, Any]]:
+        """Get personalized Flow tracks for a specific mood or genre.
+
+        :param config_id: The Flow config identifier (e.g. "happy", "chill", "genre-rock").
+        """
+        result = await self._gw_api_call(
+            "radio.getUserRadio",
+            args={"config_id": config_id, "user_id": self._user_id},
+        )
+        return cast("list[dict[str, Any]]", result["results"]["data"])
```
MarvinSchenkel, line 144: Consider adding an early return check here
```
if "data" not in result["results"]:
  return []
```

music_assistant/providers/deezer/gw_client.py
```diff
@@ -127,6 +130,38 @@ async def _gw_api_call(
...
+
+    async def get_home_flows(self) -> list[dict[str, Any]]:
+        """Discover available Flow variants from the Deezer home page.
+
+        :param: None
```
MarvinSchenkel: `:param None?`

music_assistant/providers/deezer/gw_client.py
```diff
@@ -127,6 +130,38 @@ async def _gw_api_call(
...
+                "PAGE": "home",
+                "VERSION": "2.5",
+                "SUPPORT": {"filterable-grid": ["flow"]},
+            }
+        )
+        result = await self._gw_api_call(
+            "page.get",
+            params={"gateway_input": gateway_input},
+        )
+        for section in result["results"]["sections"]:
```
MarvinSchenkel: Also check here if "sections" actually exists in `result["results"]` to prevent KeyErrors

MarvinSchenkel (changes requested):
Nice addition to the Deezer provider :-). Few minor comments, please have a look in line. Marking this PR as draft. Hit the 'Ready for review' button once you want us to have a look again. 🙏

music_assistant/providers/deezer/__init__.py
jdaberkow: There is. I just figured out that I could also simply do a full parse of the response from the gw api. :)

jdaberkow:
> Nice addition to the Deezer provider :-). Few minor comments, please have a look in line. Marking this PR as draft. Hit the 'Ready for review' button once you want us to have a look again. 🙏

@MarvinSchenkel:
Thank you for your quick review. I adressed all of your findings. :)

Besides I'm quite motivated to implement even more of Deezer's functionalities (shaker playlists, livestreams, audiobooks...) but this would be way easier and straightforward by using their graphql API.

I opened a discussion regarding that in the discord:
https://discord.com/channels/753947050995089438/1472946309353967767

music_assistant/providers/deezer/__init__.py
```diff
@@ -910,6 +982,68 @@ def _create_virtual_playlist(
             owner="Deezer",
         )
 
+    def _parse_gw_track(self, gw_track: dict[str, Any]) -> Track:
```
MarvinSchenkel: Hm, not a fan of duplicating the parsing logic. If we cannot parse a `Track` from a gw_track using the existing parse function, then I think we should go back to the previous solution where we called `get_track`. `get_similar_tracks` already does this and we cache the result, so in that case, that's preferable.

music_assistant/providers/deezer/gw_client.py
```diff
@@ -5,6 +5,7 @@
 """
 
 import datetime
+import json as json_module
```
MarvinSchenkel: Just use `import json` here as it is a common pattern

jdaberkow:
@MarvinSchenkel 

Alright. Switched back to the previous approach. Although the performance when opening a flow playlist becomes then noticeably worse.

While maintaining two different parsing functions are probably not a good idea, the main issue currently that lead to this situation is that we have two different APIs that we use for fetching tracks.
(Another possible argument for switching maybe everything to the graphql API in the long term?)

Re PR workflows: How do you handle the resolving of conversations? Shall the PR creator resolve them, when he worked on them? Is is the reviewer in charge of resolving them, after approving they have been solved?

MarvinSchenkel:
> @MarvinSchenkel
> 
> Alright. Switched back to the previous approach. Although the performance when opening a flow playlist becomes then noticeably worse.
> 
> While maintaining two different parsing functions are probably not a good idea, the main issue currently that lead to this situation is that we have two different APIs that we use for fetching tracks. (Another possible argument for switching maybe everything to the graphql API in the long term?)
> 
> Re PR workflows: How do you handle the resolving of conversations? Shall the PR creator resolve them, when he worked on them? Is is the reviewer in charge of resolving them, after approving they have been solved?

I don't really have a preference, either resolve or +1 the comments is fine. 

As for switching to the graph api, I think it is worth investigating as this would solve both the 'double parsing issue' and the 'performance issue'. Since this is a community maintained provider, I would be very happy if you could take a stab at it? For now, this PR looks good to go for me, thanks 👍 .

MarvinSchenkel (approved):

elementc:
Hey, I've been testing this feature in 2.8.0b20 and I have noticed that when the first batch of songs finish, even if Don't Stop The Music is enabled, playback ceases. 

I'll go file a proper support ticket this weekend, when I have the chance to repro at a time that I am able to grab logs, but wanted to flag this behavior here at the implementation just in case the author or reviewer note something obvious! 

Love this feature, thanks for getting it across the line!

jdaberkow:
Thanks for your feedback @elementc ! This is in fact something we should have a look into.

There is also a more general problem here. On fetch, we receive a fixed set of around 24 tracks. (One workaround there could be of course to do multiple fetches, to increase the initial playlist size, since each fetch gives us another set of tracks)

From my understanding there is currently no built-in functionality from music assistant to fetch new tracks on demand via a specified api call, besides the generic "Don't Stop The Music" function, right? @MarvinSchenkel 

The fact that "Don't Stop The Music" is currently not working with the flow is strange. I will have a look into that. But even if that is going to work, it wouldn't exactly be the desired functionality, since this will simply generate new tracks, based on the ones that were initially in the playlist, which would then loose the context of the flow. Still better than simply stopping the play for sure though.

elementc:
> The fact that "Don't Stop The Music" is currently not working with the flow is strange. I will have a look into that.

I am actually unable to reproduce the behavior I saw as of this morning: during a controlled experiment I observed one of the mood flows (Chill, naturally, for Saturday morning Coffee) start with a batch of 12 and then re-up with an additional 25 while the 12th track was playing. Not sure what went wrong here. I will poke around some more this morning, but consider the complaint rescinded. Sorry for the false alarm! From what I have observed this morning, Don't Stop The Music is a per-queue property, and not a system-global property like I previously thought. I am a new MA user and hadn't realized that, thought I now totally prefer that it be queue-specific. It almost certainly contributed to my confusion the other day, Don't Stop The Music wasn't actually on when I thought it was.

> But even if that is going to work, it wouldn't exactly be the desired functionality, since this will simply generate new tracks, based on the ones that were initially in the playlist, which would then loose the context of the flow. Still better than simply stopping the play for sure though.

I did observe this this morning, the second batch contained 2-3 songs I would definitely not classify as "Chill".

PR #3182, opened 2026-02-18
Fix multi disc audiobook

OzGav:
Fixes https://github.com/orgs/music-assistant/discussions/4956

Tested using a multi disc audio CD. Also verified the previous test audiobooks that I have for a previous PR were unaffected.

I didn't want to regress the changes from the last PR I had which fixed a number of audiobook issues. Claude's assessment is:

The fix from commit d6399d2 (supporting untagged multi-file audiobooks) is not regressed because:

The alphabetical fallback path (use_alphabetical=True) was already using position via the ternary condition
The embedded chapters path is completely separate code
Only the tagged files path changes, which is exactly where the multi-disc duplicate position bug lived

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -1882,7 +1882,7 @@ class LocalFileSystemProvider(MusicProvider):
                     continue
                 chapters.append(
                     MediaItemChapter(
-                        position=position if use_alphabetical else (chapter_tags.track or position),
+                        position=position,
                         name=chapter_tags.title,
                         start=total_duration,
                         end=total_duration + chapter_tags.duration,


MarvinSchenkel:
If AudioBooks can have multiple discs, why don't we add it to the `MediaItemChapter` to represent it  as such?

OzGav:
So you mean change the model? If so to what end? To show in the frontend similar to audio CDs?

MarvinSchenkel:
> So you mean change the model? If so to what end? To show in the frontend similar to audio CDs?

Exactly, we do it for multi disc albums, so if multi disc audio books exist, why not show it as such in the frontend? @marcelveldt thoughts?

MarvinSchenkel (approved):
Had a chat to Marcel. Apparently all the streaming services out there do not distinguish discs, so I don't think it is worth the hassle of implementing disc for audiobooks. 

Thanks @OzGav

PR #3183, opened 2026-02-18
Gracefully skip files/folders with emoji names on SMB mounts

OzGav:
Emoji and other 4-byte UTF-8 characters in folder/file names on SMB/CIFS mounts cause EINVAL errors due to a Linux kernel limitation in the CIFS client's NLS layer. Instead of crashing the library scan, these items are now skipped with a clear warning-level log message indicating the cause.

Also restores iocharset=utf8 mount option and corrects the misleading comment that previously suggested omitting it would help.

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -4,13 +4,12 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-import errno
 import logging
 import os
 import os.path
 import time
 import urllib.parse
-from collections.abc import AsyncGenerator, Iterator, Sequence
+from collections.abc import AsyncGenerator, Sequence
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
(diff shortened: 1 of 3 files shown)


music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -353,30 +354,53 @@ async def sync_library(self, media_type: MediaType) -> None:
...
-                if item.is_dir(follow_symlinks=False):
-                    yield from listdir(item.path)
-                elif item.is_file(follow_symlinks=False):
-                    # skip files without extension
-                    if "." not in item.name:
-                        continue
-                    ext = item.name.rsplit(".", 1)[1].lower()
-                    if ext not in SUPPORTED_EXTENSIONS:
-                        # skip unsupported file extension
+            try:
```
marcelveldt: this has grown to large now for an inline method - its hard to read as well.
look if we can make this into a small helper func

OzGav: Extracted it to recursive_iter() in helpers.py as a standalone function that takes base_path, supported_extensions, and the provider's logger as parameters.

marcelveldt (approved):
Looks good @OzGav, thanks!

PR #3184, opened 2026-02-18
Fix podcast episode retrieval in builtin playlist tracks

OzGav:
When loading playlist tracks, the builtin provider used media_controller.get_provider_item() which passes the controller's own media_type to the provider. For podcast episodes, get_controller() returns PodcastsController (media_type=PODCAST), causing get_podcast() to be called instead of get_podcast_episode() - resulting in a RuntimeError.

Fix by calling provider.get_item() directly with the correct media_type parsed from the URI.

### DIFF
--- a/music_assistant/providers/builtin/__init__.py
+++ b/music_assistant/providers/builtin/__init__.py
@@ -392,9 +392,12 @@ class BuiltinProvider(MusicProvider):
                     # controller's own media_type, which is wrong for podcast
                     # episodes (controller has PODCAST, not PODCAST_EPISODE).
                     item_prov = self.mass.get_provider(provider_instance_id_or_domain)
-                    if item_prov:
-                        item_prov = cast("MusicProvider", item_prov)
-                        media_item = await item_prov.get_item(media_type, item_id)
+                    if not item_prov:
+                        raise ProviderUnavailableError(
+                            f"{provider_instance_id_or_domain} is not available"
+                        )
+                    item_prov = cast("MusicProvider", item_prov)
+                    media_item = await item_prov.get_item(media_type, item_id)
                 if media_item is not None and media_item.media_type in PLAYLIST_MEDIA_TYPES:
                     playlist_item = cast("PlaylistPlayableItem", media_item)
                     playlist_item.position = index
--- a/music_assistant/providers/builtin/__init__.py
+++ b/music_assistant/providers/builtin/__init__.py
@@ -24,13 +24,11 @@ from music_assistant_models.errors import (
 )
 from music_assistant_models.media_items import (
     Artist,
-    Audiobook,
     AudioFormat,
     MediaItemImage,
     MediaItemMetadata,
     MediaItemType,
     Playlist,
-    PodcastEpisode,
     ProviderMapping,
     Radio,
     Track,
@@ -40,6 +38,7 @@ from music_assistant_models.streamdetails import StreamDetails
 
 from music_assistant.constants import (
     MASS_LOGO,
+    PLAYLIST_MEDIA_TYPES,
     VARIOUS_ARTISTS_FANART,
     PlaylistPlayableItem,
 )
@@ -396,9 +395,10 @@ class BuiltinProvider(MusicProvider):
                     if item_prov:
                         item_prov = cast("MusicProvider", item_prov)
                         media_item = await item_prov.get_item(media_type, item_id)
-                    if isinstance(media_item, (Track, Radio, PodcastEpisode, Audiobook)):
-                        media_item.position = index
-                        result.append(media_item)
+                if media_item is not None and media_item.media_type in PLAYLIST_MEDIA_TYPES:
+                    playlist_item = cast("PlaylistPlayableItem", media_item)
+                    playlist_item.position = index
+                    result.append(playlist_item)
                 else:
                     self.logger.warning(
                         "Unsupported media type in playlist %s: %s",
--- a/music_assistant/providers/builtin/__init__.py
+++ b/music_assistant/providers/builtin/__init__.py
@@ -380,22 +380,30 @@ class BuiltinProvider(MusicProvider):
                 media_type, provider_instance_id_or_domain, item_id = await parse_uri(uri)
                 media_controller = self.mass.music.get_controller(media_type)
                 # prefer item already in the db
-                track = await media_controller.get_library_item_by_prov_id(
+                media_item: (
+                    MediaItemType | None
+                ) = await media_controller.get_library_item_by_prov_id(
                     item_id, provider_instance_id_or_domain
                 )
-                if track is None:
-                    # get the provider item and not the full track from a regular 'get' call
-                    # as we only need basic track info here
-                    track = await media_controller.get_provider_item(
-                        item_id, provider_instance_id_or_domain
-                    )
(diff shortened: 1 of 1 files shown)


music_assistant/providers/builtin/__init__.py
```diff
@@ -380,22 +380,30 @@ async def get_playlist_tracks(  # type: ignore[override]
                 media_type, provider_instance_id_or_domain, item_id = await parse_uri(uri)
                 media_controller = self.mass.music.get_controller(media_type)
                 # prefer item already in the db
-                track = await media_controller.get_library_item_by_prov_id(
+                media_item: (
+                    MediaItemType | None
+                ) = await media_controller.get_library_item_by_prov_id(
                     item_id, provider_instance_id_or_domain
```
marcelveldt, line 387:
```suggestion
                media_item = await media_controller.get_library_item_by_prov_id(
                    item_id, provider_instance_id_or_domain
```

music_assistant/providers/builtin/__init__.py
```diff
@@ -380,22 +380,30 @@ async def get_playlist_tracks(  # type: ignore[override]
...
+                if media_item is None:
+                    # Call provider.get_item directly with the correct media_type.
+                    # Using media_controller.get_provider_item would pass the
+                    # controller's own media_type, which is wrong for podcast
+                    # episodes (controller has PODCAST, not PODCAST_EPISODE).
+                    item_prov = self.mass.get_provider(provider_instance_id_or_domain)
+                    if item_prov:
+                        media_item = await cast("MusicProvider", item_prov).get_item(
+                            media_type, item_id
+                        )
```
marcelveldt, line 398:
```suggestion
                    item_prov = self.mass.get_provider(provider_instance_id_or_domain)
                    if item_prov:
                        item_prov = cast("MusicProvider", item_prov)
                        media_item = await item_prov.get_item(media_type, item_id)
```

music_assistant/providers/builtin/__init__.py
```diff
@@ -380,22 +380,30 @@ async def get_playlist_tracks(  # type: ignore[override]
...
+                if isinstance(media_item, get_args(PlaylistPlayableItem)):
+                    playlist_item = cast("PlaylistPlayableItem", media_item)
```
marcelveldt: what is this ? why do we need the isinstance check (which is slow) and then also a cast.
this looks weird

music_assistant/providers/builtin/__init__.py
OzGav: The explicit MediaItemType | None annotation is required for mypy. Without it:

media_item = await media_controller.get_library_item_by_prov_id(...)

Infers type as: Artist | Album | Track | Radio | Playlist | Audiobook | Podcast | Genre | None (no PodcastEpisode)

Then when we assign from provider.get_item() which returns MediaItemType (includes PodcastEpisode), mypy errors:

error: Incompatible types in assignment (expression has type "... | PodcastEpisode | ...", variable has type "... | None")  [assignment]

The annotation widens the type to accept all media items including podcast episodes.

music_assistant/providers/builtin/__init__.py
OzGav: Agree. Fixed.

music_assistant/providers/builtin/__init__.py
```diff
@@ -380,22 +382,28 @@ async def get_playlist_tracks(  # type: ignore[override]
...
+                if media_item is None:
+                    # Call provider.get_item directly with the correct media_type.
+                    # Using media_controller.get_provider_item would pass the
+                    # controller's own media_type, which is wrong for podcast
+                    # episodes (controller has PODCAST, not PODCAST_EPISODE).
+                    item_prov = self.mass.get_provider(provider_instance_id_or_domain)
+                    if item_prov:
+                        item_prov = cast("MusicProvider", item_prov)
+                        media_item = await item_prov.get_item(media_type, item_id)
+                    if isinstance(media_item, (Track, Radio, PodcastEpisode, Audiobook)):
```
MarvinSchenkel: Why not `PlaylistPlayableItem`? I think you created that alias specifically for this?

OzGav: I can't use PlaylistItem directly in isinstance because it's a union type. I could create a tuple just for this but previously I had feedback that it wasnt worth creating a constant for use in one place

music_assistant/providers/builtin/__init__.py
MarvinSchenkel: You could also just use a `media_item = cast("MediaItemType", await item_prov.get_item(media_type, item_id))` a bit below instead of this, which should also satisfy mypy?

music_assistant/providers/builtin/__init__.py
MarvinSchenkel: Yes you can by using `get_args(PlaylistPlayableItem) `. However, you could also use `media_item.media_type in PLAYLIST_MEDIA_TYPES` here that completely bypasses `isinstance` checking (way cheaper). This requires a `playlist_item = cast("PlaylistPlayableItem", media_item)` though, which I think is preferable over `isinstance` in this case.

Another IMPORTANT note here is that the indentation is wrong. `if isinstance(media_item, (Track, Radio, PodcastEpisode, Audiobook)):` should be at the same level of indentation as `if media_item is None:`, not inside it.

music_assistant/providers/builtin/__init__.py
OzGav: I had a look at this with Claude's help and it didn't work. I think the issue is that mypy locks in the type of media_item from the first assignment (get_library_item_by_prov_id), and that return type doesn't include PodcastEpisode. Even if we cast the value on the second assignment, mypy still complains because the variable's type was already set too narrow. The cast changes the type of the value but not what the variable is allowed to hold.

So I think the annotation is the only clean way to make both assignments work. Happy to be proven wrong!

music_assistant/providers/builtin/__init__.py
OzGav: Yes quite right. Both of these are fixed.

music_assistant/providers/builtin/__init__.py
```diff
@@ -380,22 +381,29 @@ async def get_playlist_tracks(  # type: ignore[override]
...
-                    playlist_item = cast("PlaylistPlayableItem", track)
+                if media_item is None:
+                    # Call provider.get_item directly with the correct media_type.
+                    # Using media_controller.get_provider_item would pass the
+                    # controller's own media_type, which is wrong for podcast
+                    # episodes (controller has PODCAST, not PODCAST_EPISODE).
+                    item_prov = self.mass.get_provider(provider_instance_id_or_domain)
+                    if item_prov:
+                        item_prov = cast("MusicProvider", item_prov)
+                        media_item = await item_prov.get_item(media_type, item_id)
```
MarvinSchenkel:
```suggestion
                    item_prov = self.mass.get_provider(provider_instance_id_or_domain)
                    if not item_prov:
                        raise ProviderUnavailableError(
                            f"{provider_instance_id_or_domain} is not available"
                        )
                    item_prov = cast("MusicProvider", item_prov)
                    media_item = await item_prov.get_item(media_type, item_id)
```

MarvinSchenkel (approved):
Thanks @OzGav

PR #3186, opened 2026-02-18
Switch to MA MusicBrainz mirror

MarvinSchenkel:
Switch to the MA MusicBrainz mirror and relax the throttler.


marcelveldt (approved):

marcelveldt:
Make sure to also backport this to stable once its confirmed working on nightly+beta.
Follow-up should be adjusting the metadata flows to more proactively grab metadata (especially image) for artists without image.

PR #3189, opened 2026-02-19
4892 myplex auth

Stand1531:
For issue 4892

Allow non-account owners to auth using myplex. Previously if you were not the account owner it will give a 401. This will look up the server token based on local server and IP.

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -161,7 +161,6 @@ def resolve_plex_server(
     if auth_token and auth_token != AUTH_TOKEN_UNAUTH:
 
         # Ensure we have a MyPlexAccount
-        print(myplex_account)
         if not myplex_account:
             try:
                 myplex_account = MyPlexAccount(token=auth_token)
@@ -176,8 +175,13 @@ def resolve_plex_server(
                 session=session,
             )
             return plex_server
-        except Exception:
-            pass
+        except plexapi.exceptions.Unauthorized as err:
+            logging.getLogger("music_assistant.providers.plex").debug(
+                "Account-owner auth failed for %s:%s, continuing to check shared servers: %s",
+                local_server_ip,
+                local_server_port,
+                err,
+            )
 
         # If local-owner token fails, iterate over shared servers
         try:
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -446,7 +446,13 @@ class PlexProvider(MusicProvider):
         _, library_name = str(self.config.get_value(CONF_LIBRARY_ID)).split(" / ", 1)
 
         def connect() -> PlexServer:
+            """
+            - Plex.tv for discovery of accessible servers (including shared)
+            - Find the local server by IP or friendly name
+            - Use the server-specific token for a direct local connection
+            """
             try:
+                # Setup session for local connection
                 session = requests.Session()
                 session.verify = (
                     self.config.get_value(CONF_LOCAL_SERVER_VERIFY_CERT)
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
@anatosun Could you al have a look at this PR please? 🙏

music_assistant/providers/plex/helpers.py
```diff
@@ -45,17 +54,42 @@ def _get_libraries() -> list[str]:
...
-        if auth_token is None:
-            plex_server = PlexServer(
-                f"{local_server_protocol}://{local_server_ip}:{local_server_port}"
-            )
+        base_url = f"{local_server_protocol}://{local_server_ip}:{local_server_port}"
+
+        plex_server: PlexServer | None = None
+
+        # Plex.tv for resource discovery
+        if auth_token and auth_token != "local_auth":
```
MarvinSchenkel: We have a constant for this `AUTH_TOKEN_UNAUTH`

music_assistant/providers/plex/__init__.py
```diff
@@ -462,24 +468,39 @@ def connect() -> PlexServer:
...
+                                    ):
+                                        plex_server = PlexServer(
+                                            f"{conn.protocol}://{conn.address}:{conn.port}",
+                                            token=resource.accessToken,
+                                            session=session,
+                                        )
+                                        break
+                                if plex_server:
+                                    break
+                            except Exception:
```
MarvinSchenkel: Please narrow down the type of exception you are trying to catch here

music_assistant/providers/plex/__init__.py
```diff
@@ -462,24 +468,39 @@ def connect() -> PlexServer:
...
-                    plex_server = PlexServer(
-                        plex_url,
-                        token,
-                        session=session,
-                    )
-                # I don't think PlexAPI intends for this to be accessible, but we need it.
+                    # If we have a Plex.tv token, use it to discover resources
+                    plex_server = None
+                    account = self._myplex_account
+                    if account:
```
MarvinSchenkel: Looking at this logic, we seem to change the existing behaviour. Shouldn't this just be an *additional* way of authentication in case the previous logic in the `else` fails?

music_assistant/providers/plex/helpers.py
```diff
@@ -45,17 +54,42 @@ def _get_libraries() -> list[str]:
...
+            try:
```
MarvinSchenkel: @anatosun might be able to answer this one: Seems we are duplicating the connect logic from the init.py here. Any particular reason we cannot create a reusable function out of this so we keep the logic on one place?

MarvinSchenkel:

music_assistant/providers/plex/helpers.py
Stand1531: I made an attempt to consolidate connection logic but would appreciate feedback if this is the best way to handle it.

music_assistant/providers/plex/__init__.py
Stand1531: I used existing logic https://github.com/Stand1531/server/blob/1911ef1e774ba5a3905247c147d1aa3b36265e4c/music_assistant/providers/plex/__init__.py#L171 then added shared library lookup after

music_assistant/providers/plex/helpers.py
Stand1531: Moved to consolidated call and used constant https://github.com/Stand1531/server/commit/43c03910bc32793ba053293d7a710e3a6ecbd6f3

anatosun:
Hey @Stand1531 , 
First let me thank you for taking the time to open a PR. I have some small suggestions that might improve the user experience. The way I understand it is that currently the flow is: try owner token -> might fail -> iterate over resources -> retry and use appropriate token.

Would checking the resources first make more sense? 

Also the error message could be more verbose like
```python
raise LoginFailed(f"Server {ip}:{port} not accessible in your Plex account")
```

  A couple other small things to consider:
  - Could bump the fallback log from DEBUG to INFO so users know when resource tokens are being used
  - Line 378 comment says MyPlexAccount will be created/fetched but it's just set to None: might be leftover?

Stand1531:
> * 378

Thanks for the feedback! I believe addressed the suggestions in latest commit. 

Resource logger shows like
Not owned
2026-02-20 22:20:35.165 INFO (ThreadPoolExecutor-0_7) [music_assistant.providers.plex] Matched Plex resource 'My-Resource' (owned=False)

Owned
2026-02-20 22:27:25.451 INFO (ThreadPoolExecutor-0_0) [music_assistant.providers.plex] Matched Plex resource 'My-Resource' (owned=True)

Stand1531:
Hi @anatosun. Should I move this to ready for review?

Stand1531:
@MarvinSchenkel Can you help review?

music_assistant/providers/plex/__init__.py
```diff
@@ -112,6 +112,107 @@
...
+    :param local_server_ssl: Use HTTPS for local connection
+    :param local_server_verify_cert: Verify SSL certificate
+    :param auth_token: Plex auth token (or AUTH_TOKEN_UNAUTH for local)
+    :param myplex_account: Optional MyPlexAccount object
+    :return: PlexServer instance
+    :raises LoginFailed: if authentication fails or server not found
+    """
+
+    # silence loggers
+    logging.getLogger("plexapi").setLevel(logging.WARNING)
```
MarvinSchenkel, line 139: This should be part of the handle_async_init (where it was), not a helper function

music_assistant/providers/plex/__init__.py
```diff
@@ -112,6 +112,107 @@
...
+    plex_server: PlexServer | None = None
+
+    # MyPlex authentication path
+    if auth_token and auth_token != AUTH_TOKEN_UNAUTH:
+
+        # Ensure we have a MyPlexAccount
+        if not myplex_account:
+            try:
+                myplex_account = MyPlexAccount(token=auth_token)
+            except Exception as err:
```
MarvinSchenkel, line 167: Please catch specific exceptions here such as `plexapi.exceptions.Unauthorized` or `requests.exceptions.ConnectionError`

music_assistant/providers/plex/__init__.py
```diff
@@ -112,6 +112,107 @@
...
+                    continue
+
+                # Match server by address/port
+                for conn in resource.connections:
+                    if (
+                        conn.address == local_server_ip
+                        and int(conn.port) == int(local_server_port)
+                    ):
+
+                        logging.getLogger("music_assistant.providers.plex").info(
```
MarvinSchenkel, line 183: You can use `self.logger` if you make this function a private function of the provider.

music_assistant/providers/plex/__init__.py
```diff
@@ -437,69 +553,36 @@ class PlexProvider(MusicProvider):
...
+                myplex_account=self._myplex_account,
+                client_id=self.instance_id,
+                client_version=self.mass.version,
+            )
 
-        self._myplex_account = await self.get_myplex_account_and_refresh_token(
-            str(self.config.get_value(CONF_AUTH_TOKEN))
-        )
         try:
             self._plex_server = await self._run_async(connect)
```
MarvinSchenkel, line 583: You can directly pass arguments to this funcitn. No need to wrap `resolve_plex_server` in a `connect`

MarvinSchenkel: Also, this new wrapper does not seem to handle "Invalid Token" properly anymore?

MarvinSchenkel (changes requested):
I unfortunately think this PR needs some structural changes. I don't have a Plex setup here that I can use to test this, so it's hard for me to help you out. Hopefully @anatosun can jump in and help you restructure / test the changes in this PR?

Stand1531:
Thanks for checking. I am just going to close this. Like I previously stated, i am not a programmer and was just trying to help. I'll make the change in my local.

PR #3191, opened 2026-02-19
Fix DSP not applying for AirPlay and Sendspin players

maximmaxim345:
DSP config is stored on the parent player, but was being looked up by the protocol player ID. Resolving to the parent player fixes DSP for AirPlay and Sendspin.

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -1402,10 +1402,18 @@ def get_player_filter_params(
     """Get player specific filter parameters for ffmpeg (if any)."""
     filter_params = []
 
-    dsp = mass.config.get_player_dsp_config(player_id)
+    # In case this is a protocol player, their DSP config is stored
+    # under the parent (native or universal) player that wraps them.
+    dsp_player_id = player_id
+    if (player := mass.players.get_player(player_id)) and (
+        player.type == PlayerType.PROTOCOL and player.protocol_parent_id
+    ):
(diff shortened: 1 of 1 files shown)


music_assistant/helpers/audio.py
```diff
@@ -1402,10 +1402,18 @@ def get_player_filter_params(
...
-    dsp = mass.config.get_player_dsp_config(player_id)
+    # In case this is a protocol player, their DSP config is stored
+    # under the parent (native or universal) player that wraps them.
+    dsp_player_id = player_id
+    if (player := mass.players.get_player(player_id)) and (
+        player.type == PlayerType.PROTOCOL and player.protocol_parent_id
+    ):
+        dsp_player_id = player.protocol_parent_id
+
+    dsp = mass.config.get_player_dsp_config(dsp_player_id)
```
marcelveldt, line 1411:
```suggestion
    player = mass.players.get_player(player_id)
    # In case this is a protocol player, their DSP config is stored
    # under the parent (native or universal) player that wraps them.
    dsp_player_id = player_id
    if player and player.protocol_parent_id:
        dsp_player_id = player.protocol_parent_id
    dsp = mass.config.get_player_dsp_config(dsp_player_id)
```

marcelveldt: protocol_parent_id only has a value for protocol players

marcelveldt (approved):

PR #3192, opened 2026-02-19
Fix grouping for for players whos native protocol is a protocol of other players

MarvinSchenkel:
For example, a homepod which native protocol is Airplay can only group with `apxxxx` players. We need to translate those visible parents so that a home pod can also group with a Sonos that support Airplay.

Similarly, we also need to do this the other way around, where a Sonos with Airplay can also group with the homepod.

### DIFF
--- a/music_assistant/models/player.py
+++ b/music_assistant/models/player.py
@@ -1784,8 +1784,7 @@ class Player(ABC):
 
         Protocol players are hidden and users interact with visible players
         (native or universal). This method translates protocol player IDs
-        back to the visible (parent) players. Non-protocol players (e.g.,
-        native Apple AirPlay players) are already visible and included as-is.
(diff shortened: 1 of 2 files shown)


music_assistant/models/player.py
```diff
@@ -1777,15 +1786,19 @@ def _translate_protocol_ids_to_visible(self, player_ids: set[str]) -> set[Player
         (native or universal). This method translates protocol player IDs
         back to the visible (parent) players.
 
-        :param player_ids: Set of player IDs (protocol player IDs).
+        :param player_ids: Set of player IDs (may include both protocol and non-protocol IDs).
```
marcelveldt:
```suggestion
        :param player_ids: Set of player IDs.
```

marcelveldt (approved):

PR #3194, opened 2026-02-19
Validate queue item ID in Sonos pause path

rjbutler:
There is an apparent race condition where after calling media_player.pause, the player's current media URI can become just 'mass:<queue_id>' (without the item ID suffix).   When this happens, Music Assistant loses its place in the queue when paused and starts over with the first item in the queue on resume.

get_item as used in the two code paths below this change validates correctly and returns None in this case.  Changing the code to be consistent with that approach seems to solve the queue reset on pause issue.


marcelveldt (approved):
Nice find @rjbutler !

PR #3197, opened 2026-02-20
Fix RAOP password handling

bradkeifer:
This fixes a bug which was passing the MA encrypted format of the device password to cliraop. We now will pass the password in an unencrypted format as a command line argument.
Additional error logging incorporated if the device requires a password, but none has been defined in the Player Settings.


MarvinSchenkel (approved):
Thanks @bradkeifer !

PR #3198, opened 2026-02-20
Add API to handle playback speed

andykelk:
Add handling of the custom speed set by the UI (see https://github.com/music-assistant/frontend/pull/1476).

A new `player_queues/set_playback_speed` API command stores the speed in `queue.extra_attributes["playback_speed"]`. `get_player_filter_params()` then applies an ffmpeg `atempo` filter at stream time, which changes tempo without affecting pitch.

If the queue is actively playing a seekable item, the stream is restarted at the current position so the new speed takes effect immediately. `extra_attributes` was chosen over player config because: it serialises automatically in `QUEUE_UPDATED` events (so the frontend gets live updates for free), it's cached and survives server restarts, and it's semantically tied to the queue/content context rather than the player hardware.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -379,6 +379,31 @@ class PlayerQueuesController(CoreController):
             if next_item := self.get_next_item(queue_id, queue.index_in_buffer):
                 self._enqueue_next_item(queue_id, next_item)
 
+    @api_command("player_queues/set_playback_speed")
+    async def set_playback_speed(self, queue_id: str, speed: float) -> None:
+        """Set the playback speed for the given queue.
+
+        :param queue_id: queue_id of the queue to configure.
+        :param speed: playback speed multiplier (0.5 to 2.0). 1.0 = normal speed.
+        """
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/player_queues.py
```diff
@@ -379,6 +379,31 @@ def set_repeat(self, queue_id: str, repeat_mode: RepeatMode) -> None:
...
+        self.signal_update(queue_id)
+        # If currently playing a seekable item, restart the stream at the current position
+        # so the new speed takes effect immediately without waiting for the next track.
+        if (
+            queue.state == PlaybackState.PLAYING
+            and queue.current_item
+            and queue.current_item.duration
+            and queue.current_index is not None
+        ):
+            await self.seek(queue_id, int(queue.corrected_elapsed_time))
```
marcelveldt: I think you can also just call resume here

andykelk: Good call. Updated.

marcelveldt (approved):
Nice QoL improvement @andykelk !

PR #3200, opened 2026-02-20
OpenSubsonic: Use server provided version tag if present

khers:
The Open Subsonic spec has a field for album version and some implementations are populating this field. We should use the provided value if it is present.


marcelveldt (approved):
LGTM!

PR #3206, opened 2026-02-21
Fix HEOS source switching back to Local Music after starting stream

Tommatheussen:
Fix for a reported issue on Discord. 
When an external source is active, then start playing from MA, the UI briefly shows the correct metadata (from MA's queue).
Then the player switches it's source back to Local Music and just shows URL Stream.

The problem is that HEOS sends an event that there internal queue changed, AFTER it sends it's event that the playing media was changed. Because HEOS doesn't distinguish between different local sources, we shouldn't update the active source and media on those queue change events, as we can't properly check the source ID.

Additionally, we specifically log the errors on a playback error event now, helping the debug process if that event occurs.

### DIFF
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -63,7 +63,7 @@ class HeosPlayer(Player):
     async def setup(self) -> None:
         """Set up the player."""
         self.set_device_info()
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
Thanks, @Tommatheussen !

PR #3209, opened 2026-02-21
Snapcast: Fixes for hard switching of group leaders

gnumpi:
This PR fixes an snapcast issue where the previous leader player was not stopped when switching leaders in sync group players while using hard switching (which enforces a stop between re-creation).


marcelveldt (approved):
LGTM
Thanks, @gnumpi !

PR #3210, opened 2026-02-21
Airplay2-configurable-latency

bradkeifer:
Implements a configurable output buffer duration as an advanced setting for AirPlay 2 players. The default value is 250ms, which significantly reduces the latency experienced for commencement of playback. 
Changes also made to the derivation of ntpstart time to cater for the longer duration experienced in establishing AirPlay 2 connections. This should improve synchronisation issues being experienced when AirPlay 2 players are in a sync group.


music_assistant/providers/airplay/protocols/airplay2.py
```diff
@@ -98,12 +103,12 @@ async def start(self, start_ntp: int) -> None:
             str(self._cli_loglevel),
             "--dacp_id",
             prov.dacp_id,
-            "--active_remote",
-            self.active_remote_id,
```
marcelveldt, line 102: you do confirm that the active remote is still functioning with this change ?
so e.g. volume changes on the device itself get processed as DACP commands

music_assistant/providers/airplay/player.py
```diff
@@ -223,22 +237,21 @@ async def get_config_entries(
...
+                default_value=AIRPLAY_OUTPUT_BUFFER_MIN_DURATION_MS,
+                range=(AIRPLAY_OUTPUT_BUFFER_MIN_DURATION_MS, AIRPLAY_OUTPUT_BUFFER_DURATION_MS),
+                label="Milliseconds of data to buffer",
                 description=(
-                    "The AirPlay protocol allows devices to report their own volume "
-                    "level. \n"
-                    "For some devices this is not reliable and can cause unexpected "
-                    "volume changes. \n"
-                    "Enable this option to ignore these reports."
+                    "The number of milliseconds of data to buffer in the CLI binary\n"
```
marcelveldt:
```suggestion
                    "The number of milliseconds of data to buffer\n"
```

marcelveldt (approved):
Thanks, @bradkeifer !

music_assistant/providers/airplay/protocols/airplay2.py
bradkeifer: Yes. The active remote id being natively used by OwnTone is identical to the on we were passing as an argument.

PR #3211, opened 2026-02-22
Fix inverted track_number condition in Bandcamp converter

teancom:
The condition checked output.track_number instead of track.track_number, meaning track numbers from the API were only applied when the output already had a non-None default.


MarvinSchenkel (approved):
Good catch 👍

PR #3213, opened 2026-02-22
Set audio format on Bandcamp track ProviderMapping

teancom:
Set audio format and fetch fresh streaming URLs for Bandcamp

Extract _fetch_api_track from get_track so get_stream_details can bypass the 30-day cache and always retrieve a fresh streaming URL. Set AudioFormat (content_type, bit_rate) on both StreamDetails and ProviderMapping instead of stashing the URL in metadata.links.

Also: guard split_id against malformed IDs, handle BandcampMustBeLoggedInError explicitly, and fix album_from_api for albums with no release_date.

Tests cover _fetch_api_track paths, streaming URL priority/format selection, library async generators, split_id validation, and error mapping.

### DIFF
--- a/music_assistant/providers/bandcamp/__init__.py
+++ b/music_assistant/providers/bandcamp/__init__.py
@@ -15,9 +15,14 @@ from bandcamp_async_api import (
     SearchResultArtist,
     SearchResultTrack,
 )
-from bandcamp_async_api.models import CollectionType
+from bandcamp_async_api.models import BCAlbum, BCTrack, CollectionType
 from music_assistant_models.config_entries import ConfigEntry, ConfigValueType, ProviderConfig
-from music_assistant_models.enums import ConfigEntryType, MediaType, ProviderFeature, StreamType
+from music_assistant_models.enums import (
+    ConfigEntryType,
+    MediaType,
+    ProviderFeature,
+    StreamType,
+)
 from music_assistant_models.errors import (
     InvalidDataError,
     LoginFailed,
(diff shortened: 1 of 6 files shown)


music_assistant/providers/bandcamp/converters.py
```diff
@@ -165,6 +166,7 @@ def track_from_api(
     ) -> MATrack:
         """Convert a Track object from the API to MA Track format."""
         album_id = album_id or 0
+        streaming_url, bitrate = self.streaming_url_from_api(track.streaming_url)
```
MarvinSchenkel: I am not entirely sure how Bandcamp works under the hood, but I can imagine it uses urls that expire. Why not just call `self.streaming_url_from_api` from `get_stream_details` so you also have the bitrate available?

teancom: I'd actually considered that approach but for some reason decided this was less 'invasive' of a change. I went back and forth a bit this morning and decided to take your suggested approach and then also address the reasons why I'd hesitated before. Keeping this in Draft for you to take a look at, it's become a bit of a bigger change.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -283,43 +289,56 @@ async def get_album(self, prov_album_id: str) -> Album:
...
+        """Fetch a raw API track and its parent album by compound item ID.
+
+        Uses get_album when album_id is present (most tracks), falling back
+        to get_track for standalone tracks (album_id=0). Not cached or
+        throttled directly; callers apply those decorators as needed.
+
+        :param item_id: Compound track ID in the form artist_id-album_id-track_id.
+        """
+        artist_id, album_id, track_id = split_id(item_id)
+        if track_id is None:
```
MarvinSchenkel: This is never `True` since `split_id` does not return `None` (but 0). Don't think that's intended?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -283,43 +289,56 @@ async def get_album(self, prov_album_id: str) -> Album:
         except BandcampAPIError as error:
             raise MediaNotFoundError(f"Failed to get album {prov_album_id}") from error
 
-    @use_cache(CACHE)
-    @throttle_with_retries
-    async def get_track(self, prov_track_id: str) -> Track:
-        """Get full track details by id."""
-        artist_id, album_id, track_id = split_id(prov_track_id)
-        if track_id is None:  # artist_id-track_id
+    async def _fetch_api_track(self, item_id: str) -> tuple[BCTrack, BCAlbum | None]:
```
MarvinSchenkel, line 292: Since this is actually the function that requires throttling, I would add the decorator to this function and remove it from get_track and get_streamdetails. So:

get_track -> @use_cache
get_streamdetails -> no decorator
_fetch_api_track -> @throttle_with_retries

music_assistant/providers/bandcamp/converters.py
MarvinSchenkel: Few last small things. Please mark as ready for review when you want me to have another look :)

teancom:
Just a note that I'm testing this branch locally and will report back after I have a few hours of audio playback and manually poking it with a stick under my belt.

teancom:
@MarvinSchenkel I think it's ready for re-review. 🙇

MarvinSchenkel (approved):
Thanks @teancom !

PR #3215, opened 2026-02-22
Bandcamp: validate login on init when credentials are configured

teancom:
The provider works without login (search-only mode), but if the user has explicitly configured credentials, validate them at startup with a lightweight API call. Bad/expired logins fail hard with LoginFailed so the user can fix it immediately. Transient errors (rate limits, network issues) are logged as warnings and the provider continues since the login may still be valid.

### DIFF
--- a/music_assistant/providers/bandcamp/__init__.py
+++ b/music_assistant/providers/bandcamp/__init__.py
@@ -124,6 +124,19 @@ class BandcampProvider(MusicProvider):
         )
         self._converters = BandcampConverters(self.domain, self.instance_id)
 
+        # The provider can function without login (search-only),
+        # but if credentials were explicitly configured, validate them now.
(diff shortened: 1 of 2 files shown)


OzGav:
Tests are failing

teancom:
Well I feel very silly. Sorry about that. I'm so used to the email saying the 'label' test failed I ignored the legit one :-/ 

Now, I fixed the band camp tests but there are other failing that aren't related to this change (they were broken in c6347f27). I have a fix which I can add to this PR or I can open a new one. Any preference?

teancom:
That was a silly question. Of course I shouldn't add it to this PR. I made a new one, and over there, those tests are now passing. https://github.com/music-assistant/server/pull/3217

MarvinSchenkel (approved):
Thanks @teancom 👍

PR #3216, opened 2026-02-22
Adapt core to allow multiple media types in playlists

fmunkes:
Adapts core to allow playlists of different media types.

https://github.com/music-assistant/models/pull/169


tests/providers/tidal/test_provider.py
```diff
@@ -333,7 +333,7 @@ async def test_create_playlist_delegates_to_playlists(provider: TidalProvider) -
     with patch.object(provider.playlists, "create", new_callable=AsyncMock) as mock_create:
         mock_create.return_value = Mock(spec=Playlist)
 
-        await provider.create_playlist("New Playlist")
+        await provider.create_playlist("New Playlist", media_types={MediaType.TRACK})
```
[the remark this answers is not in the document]
fmunkes: This is calling the provider method, so this should be fine.

music_assistant/models/music_provider.py
```diff
@@ -1029,20 +1030,26 @@ async def _sync_playlist_tracks(self, provider_playlist: Playlist) -> None:
...
                 elif not self._check_provider_mappings(library_track, prov_track, True):
                     # existing library track but provider mapping doesn't match
-                    library_track = await self.mass.music.tracks.update_item_in_library(
-                        library_track.item_id, prov_track
+                    library_track = await controller.update_item_in_library(
+                        library_track.item_id,
+                        prov_track,  # type: ignore[arg-type]
                     )
                 fallback_genres = (
                     set(prov_track.metadata.genres)
```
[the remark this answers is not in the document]
fmunkes: I do not see an issue here, controller is just the respective one for the media item, i.e. the playlist, audiobook, podcast or tracks. I don't see a genre issue.

music_assistant/controllers/music.py
```diff
@@ -2481,6 +2499,32 @@ async def _get_or_create_genre(
             # to the crossfade mixer. Truncate the table so all analyses are re-computed.
             await self._database.execute(f"DELETE FROM {DB_TABLE_SMART_FADES_ANALYSIS}")
 
+        if prev_version <= 30:
+            # add supported_mediatypes column to playlist table, and populated it with
+            # {MediaType.TRACK}, i.e. ["track"], as this was the only media type supported.
+            await self._database.execute(
+                f"ALTER TABLE {DB_TABLE_PLAYLISTS} ADD COLUMN supported_mediatypes json DEFAULT '' "
```
marcelveldt: should the default be '[\"track\"]' ?

music_assistant/controllers/music.py
```diff
@@ -2481,6 +2499,32 @@ async def _get_or_create_genre(
...
+            )
+            await self._database.execute(
+                f"UPDATE {DB_TABLE_PLAYLISTS} SET supported_mediatypes = '[\"track\"]'"
+            )
+            # The builtin provider is special, as it supports all media types for
+            # user-created playlists already prior to this schema version.
+            query = (
+                "SELECT * FROM provider_mappings WHERE provider_domain = 'builtin' and "
+                "provider_item_id not in "
+                f"('{BUILTIN_ALL_FAVORITE_TRACKS}','{BUILTIN_RANDOM_ALBUM}','{BUILTIN_RANDOM_TRACKS}','{BUILTIN_RECENTLY_ADDED_TRACKS}','{BUILTIN_RECENTLY_PLAYED}','{BUILTIN_RANDOM_ARTIST}')"
```
marcelveldt: I wonder if this part is needed. We can also just adjust the update logic to detect changes in supported_mediatypes and then it will auto recover itself when the sync is run

music_assistant/controllers/music.py
fmunkes: Changed it at both places.

music_assistant/controllers/music.py
fmunkes: yeah, that is cleaner - I also forgot the update_libray_item method altogether, so this is removed now, and the latter added.

marcelveldt (approved):
Looks good, nice job @fmunkes !

PR #3217, opened 2026-02-22
Fix player controller tests broken by handle_player_command decorator

teancom:
Commit c6347f27 added @handle_player_command(lock=True) to cmd_set_members, which wraps all exceptions in PlayerCommandFailed. This broke two tests:

1. test_set_members_requires_feature: Now expects PlayerCommandFailed instead of UnsupportedFeaturedException since the decorator wraps it.

2. test_cannot_group_incompatible_players: This test was a false positive before c6347f27 — it never called update_state(), so SET_MEMBERS wasn't visible in state.supported_features, and the test was hitting the "doesn't support SET_MEMBERS" check (raising UnsupportedFeaturedException) instead of ever reaching the incompatibility check it claimed to test. The actual code warns and filters incompatible players rather than raising, so the fix asserts the player was not added to the group.

### DIFF
--- a/tests/core/test_player_controller.py
+++ b/tests/core/test_player_controller.py
@@ -15,7 +15,7 @@ from unittest.mock import MagicMock
 
 import pytest
 from music_assistant_models.enums import PlayerFeature
-from music_assistant_models.errors import UnsupportedFeaturedException
+from music_assistant_models.errors import PlayerCommandFailed
 
 from music_assistant.controllers.players import PlayerController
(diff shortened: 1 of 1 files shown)


marcelveldt:
tests are fine on my end. Did you rebase your branch ? You may have seen a snapshot in between some try outs of me.

teancom:
Weird. You can a failed run here: https://github.com/music-assistant/server/actions/runs/22280753244/job/64450901920?pr=3215 but yeah, it all looks good now? Sorry for the noise.

PR #3228, opened 2026-02-24
Add custom URL endpoint for listenbrainz scrobble

relic664:
The current ListenBrainz scrobbler was limited to scrobbling to the official API only, due to an upstream limitation in `liblistenbrainz`. I've contributed a fix there such that the `liblistenbrainz` client will now accept a custom URL, instead of a hardcoded value. I'm opening this PR to also add the capability in music assistant to specify a custom ListenBrainz endpoint. 

I added the URL specification as an optional configuration option. If no URL is specified, then the provider will fallback to the official, public ListenBrainz instance (the current, hard coded endpoint).


relic664:
@ijc would you mind reviewing?

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -155,6 +163,15 @@ async def get_config_entries(
...
             value=values.get(CONF_USER_TOKEN) if values else None,
         ),
+        ConfigEntry(
+            key=CONF_API_BASE_URL,
+            type=ConfigEntryType.STRING,
+            label="Base URL",
+            required=False,
+            value=values.get(CONF_API_BASE_URL) if values else None,
+            description="URL for listenbrainz endpoint. Leave blank to default"
+            "to the public listenbrainz API.",
```
MarvinSchenkel, line 169: Can we add this to the advanced category (`category="advanced"`)?

music_assistant/providers/listenbrainz_scrobble/__init__.py
```diff
@@ -37,12 +39,18 @@ async def setup(
 ) -> ProviderInstanceType:
     """Initialize provider(instance) with given configuration."""
     token = config.get_value(CONF_USER_TOKEN)
+    api_base_url = config.get_value(CONF_API_BASE_URL)
```
MarvinSchenkel: You should be able to use the default value in 1 go via `api_base_url = config.get_value(CONF_API_BASE_URL, LISTENBRAINZ_API_URL)`

MarvinSchenkel (changes requested):
Nice addition! Few small comments and then this should be good to merge 🙏 . Marking as draft so we can stay on top of PRs that need our attention. Mark as ready for review again once you want us to have another look.

music_assistant/providers/listenbrainz_scrobble/__init__.py
ijc: The end result is the same but I wonder if it would be better to default to None and let the underlying library pick the endpoint in that case?

ijc:
No objections from me, looks good, thanks!

music_assistant/providers/listenbrainz_scrobble/__init__.py
relic664: Sure, I'll make this change

music_assistant/providers/listenbrainz_scrobble/__init__.py
relic664: I cleaned up the `api_base_url` construction. 

The client construction looks like:

```
def __init__(self, api_base_url = 'https://api.listenbrainz.org'):
```

So passing `None` for `api_base_url` would raise an error (there's no parsing/checking of `api_base_url` upstream).  I think having an explicit fallback like this makes the code easier to follow without looking at the upstream lib to see what the default endpoint is.

MarvinSchenkel (approved):
Awesome, thanks @relic664 👏

vegardhw:
@MarvinSchenkel Which release will include this?

MarvinSchenkel:
> @MarvinSchenkel Which release will include this?

2.8

PR #3231, opened 2026-02-24
Fix language and genre mapping for Spotify podcasts and audiobooks

OzGav:


music_assistant/providers/spotify/parsers.py
```diff
@@ -261,9 +261,10 @@ def parse_podcast(podcast_obj: dict[str, Any], provider: SpotifyProvider) -> Pod
     if "explicit" in podcast_obj:
         podcast.metadata.explicit = podcast_obj["explicit"]
 
-    # Convert languages list to genres for categorization
-    if "languages" in podcast_obj:
-        podcast.metadata.genres = set(podcast_obj["languages"])
+    if podcast_obj.get("languages"):
+        podcast.metadata.languages = UniqueList(podcast_obj["languages"])
+
+    podcast.metadata.genres = {"Spoken Word"}
```
MarvinSchenkel, line 267: Just one double check, is there any `metadata.genres` returned from spotify?

OzGav: Not that I could see when I tried the web API

OzGav: For clarity. Looking at the sample api responses only artists and albums return a genre key

MarvinSchenkel (approved):
Thanks Gav!

MarvinSchenkel (approved):

PR #3235, opened 2026-02-24
Fix IPv6 support across core and providers

fmurodov:
### Changes:
- Fix zeroconf crash on IPv6-only hosts (graceful fallback when no IPv4 multicast).
- Bracket IPv6 in URLs across providers (RFC 2732).
- Fix sendspin proxy using 127.0.0.1 on IPv6-only hosts.
- Fix device discovery always preferring IPv4 even when server is on IPv6.
- Fix RAOP -if flag passing mismatched address family.
- Fix get_ip_from_host only resolving IPv4.
- Remove duplicate :: bind option (Default 0.0.0.0 does also listen on ::).

Affected: AirPlay, Chromecast, MusicCast, Snapcast, Sonos S1/S2, VBAN Receiver

Also fixed cliap2 (AirPlay 2) IPv6 in a separate PR: https://github.com/music-assistant/cliairplay/pull/86
That was the main blocker for me — AirPlay 2 playback to HomePod Mini on IPv6-only network.

Tested on IPv6-only and dual-stack Docker/Kubernetes deployments with HomePod Mini and Macbook as Airplay 2 players.

No breaking changes on dual-stack networks.

### Known upstream limitations on IPv6-only networks
- **cliraop (AirPlay 1/RAOP)**: IPv4-only, exits on IPv6 addresses. Use AirPlay 2 instead.
- **aiosendspin**: Hardcodes IPVersion.V4Only for zeroconf, mDNS advertisement fails (non-fatal, HTTP still works).
- **DLNA (async_upnp_client)**: UPnP/SSDP is IPv4 multicast only. Protocol limitation.
- **Internet services without AAAA records**: Expected failures on IPv6-only (all providers from not requiring auth I tested).


marcelveldt:
Nice to see some love for IPv6 !
Did you test these changes on MacOS as well ?
As that is the de-facto development environment nowadays and I know pyzeroconf has some issues with IPv6 on mac.

music_assistant/mass.py
```diff
@@ -168,10 +168,37 @@ async def start(self) -> None:
...
-            ip_version=IPVersion.All if use_all_interfaces else IPVersion.V4Only,
-            interfaces=InterfaceChoice.All if use_all_interfaces else InterfaceChoice.Default,
-        )
+        # Try to create zeroconf with the preferred IP version, falling back if the
+        # network doesn't support it (e.g. IPv6-only host has no IPv4 multicast).
+        # IPv6 multicast requires InterfaceChoice.All (Default only works for IPv4).
+        if use_all_interfaces:
+            zc_attempts = [
+                (IPVersion.All, InterfaceChoice.All),
+                (IPVersion.V6Only, InterfaceChoice.All),
```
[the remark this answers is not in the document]
fmurodov: will fix

music_assistant/providers/snapcast/control.py
```diff
@@ -20,6 +20,12 @@
 
 import shortuuid
 
+
+def _format_ip(ip: str) -> str:
+    """Wrap IPv6 addresses in brackets for use in URLs (RFC 2732)."""
+    return f"[{ip}]" if ":" in ip else ip
```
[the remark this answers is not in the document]
fmurodov: fixed. using a single helper now

music_assistant/helpers/util.py
```diff
@@ -632,22 +649,35 @@ async def remove_file(file_path: str) -> None:
...
-    for address in discovery_info.parsed_addresses(IPVersion.V6Only):
-        if address.startswith(("::1", "fe80")):
-            # filter out loopback and link-local addresses
-            continue
-        return address
+def get_primary_ip_address_from_zeroconf(
+    discovery_info: AsyncServiceInfo,
+    prefer_ipv6: bool = False,
+) -> str | None:
+    """Get primary IP address from zeroconf discovery info.
```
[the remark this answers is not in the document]
fmurodov: added tests

music_assistant/mass.py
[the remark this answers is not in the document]
fmurodov: done

fmurodov:
> Nice to see some love for IPv6 ! Did you test these changes on MacOS as well ? As that is the de-facto development environment nowadays and I know pyzeroconf has some issues with IPv6 on mac.

Tested and confirmed that all is good as much as it can be.
Tests are passing in macOS (I was using Docker before).

Tested running the app on macOS in an IPv6-only network. It requires setting Players/Zeroconf network interfaces to All interfaces to discover players. Otherwise, it tries default V4Only, which works because of loopback having IPv4. (still airplay won't work because of linux specific binaries)

I don’t think that it’s worth fixing this behavior, as just switching to all interfaces works and I'm afraid of breaking default behaviour.

fmurodov:
Can it be merged?

trisk:
Marcel informed me about this PR since I was investigating a zeroconf discovery failure with `InterfaceChoice.Default` on a dual-stack host running Music Assistant on Home Assistant OS with only one enabled interface, and one disabled wifi interface. I was comparing Music Assistant's zeroconf interface configuration strategy to to Home Assistant's, since HA's zeroconf doesn't encounter the same issues.

As you know, we pass `IPVersion.V4Only` and `InterfaceChoice.Default` when constructing `zeroconf.AsyncZeroconf()`, by default. If `use_all_interfaces` is enabled, we use  `IPVersion.All` and `InterfaceChoice.All` instead. Your changes attempt to fall back to other combinations of `IPVersion` and `InterfaceChoice` in case either of these cause `zeroconf.AsyncZeroconf()` to fail.

I'm not sure if it's possible for `IPVersion.All` and`InterfaceChoice.All` or `InterfaceChoice.Default` to fail on a dual-stack host but if it is we might end up with a suboptimal configuration where the host is dual-stack but we are using either `IPVersion.V4Only` or `IPVersion.V6Only`. The issue I was debugging also suggests that it's not safe to use `InterfaceChoice.Default` without additional validation. These are areas where Home Assistant's interface configuration strategy differs.

Home Assistant also uses the `zeroconf` library, but enumerates the available interfaces with the `ifaddr` library's `get_adapters()` first.
* It checks for IPv6 and IPv4 on all of the interfaces to determine which `IPVersion` flag to use.
* Then, if using `IPVersion.V4Only`, it does a check to see if the default interface is the _only_ enabled interface, in which case (and only in this case) it also uses `InterfaceChoice.Default`.
* Otherwise, it passes an explicit list of addresses derived from `ifaddr.get_adapters()` (after excluding loopback addresses, globally-routable IPv6 addresses, and addresses that should be impossible in v4 or v6-only mode). Home Assistant _never_ uses `InterfaceChoice.All`.
See: [homeassistant/components/zeroconf/__init__.py](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zeroconf/__init__.py)

I think we might get a more optimal interface configuration using this strategy and avoid `zeroconf.AsyncZeroconf()` failures on v4/v6-only hosts. This would probably eliminate the need for the `use_all_interfaces` setting as well, since it makes the default just as robust (and certainly HA has never needed that option in its implementation). @fmurodov what do you think?

music_assistant/helpers/util.py
```diff
@@ -632,22 +649,35 @@ async def remove_file(file_path: str) -> None:
...
+
+    :param discovery_info: The zeroconf service info to extract the address from.
+    :param prefer_ipv6: If True, prefer IPv6 addresses over IPv4.
+    """
+    if prefer_ipv6:
+        # Prefer IPv6, fall back to IPv4
+        for address in discovery_info.parsed_addresses(IPVersion.V6Only):
+            if address.startswith(("::1", "fe80")):
```
bdraco: Please use the built-in ipaddress helpers like `ip_addr.is_link_local` and not `ip_addr.is_unspecified` loopback, etc

bdraco: https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_loopback

bdraco: You can get ip address objects from https://github.com/python-zeroconf/python-zeroconf/blob/a7f13b182675d26ce938058a2015278bbcaac736/src/zeroconf/_services/info.py#L328

music_assistant/mass.py
```diff
@@ -158,10 +158,38 @@ async def start(self) -> None:
         )
         # IPv6 requires InterfaceChoice.All, so only enable when zeroconf_interfaces is "all"
         use_all_interfaces = zeroconf_interfaces == "all"
-        self.aiozc = AsyncZeroconf(
-            ip_version=IPVersion.All if use_all_interfaces else IPVersion.V4Only,
-            interfaces=InterfaceChoice.All if use_all_interfaces else InterfaceChoice.Default,
-        )
+        # Try to create zeroconf with the preferred IP version, falling back if the
```
bdraco: The brute force strategy of trying something until it works is not great here. It would be better to work out the correct values ahead of time rather than trying until it works

MarvinSchenkel:

OzGav:
@fmurodov would be great if we can get this into v2.8 which is due for release in a couple of weeks.

fmurodov:
> @fmurodov would be great if we can get this into v2.8 which is due for release in a couple of weeks.

Let me try to resolve current issues then

music_assistant/mass.py
fmurodov: this is fixed

music_assistant/helpers/util.py
fmurodov: @bdraco Thanks for help. I think its good now

fmurodov: Thanks for the detailed analysis @trisk! I've updated the PR to address the key points:
- get_zeroconf_args() now uses ifaddr.get_adapters() to determine the correct IPVersion and interface list upfront. no more brute-force fallback chain
- Switched to ip_addresses_by_version() with proper is_loopback/is_link_local/is_unspecified checks instead of string matching
- Explicit IP lists are passed to zeroconf in most cases, only using InterfaceChoice.Default for IPv4-only

Fully adopting HA's strategy (removing use_all_interfaces, never using InterfaceChoice.All, etc.) would be a good follow-up but I'd rather not expand the scope of this PR further. I'm not confident enough to make those architectural changes, so leaving that for a future effort.

music_assistant/helpers/util.py
```diff
@@ -715,6 +735,60 @@ def get_port_from_zeroconf(discovery_info: AsyncServiceInfo) -> int | None:
...
+        # User explicitly requested all interfaces — pass explicit IP list
+        # to avoid issues with InterfaceChoice.Default on multi-interface hosts.
+        if interface_ips:
+            return {"ip_version": ip_version, "interfaces": interface_ips}
+        return {"ip_version": ip_version, "interfaces": InterfaceChoice.All}
+
+    # Default mode: use InterfaceChoice.Default for IPv4-only single-interface,
+    # otherwise pass explicit interface list for reliability.
+    if ip_version == IPVersion.V4Only:
+        return {"ip_version": ip_version, "interfaces": InterfaceChoice.Default}
```
trisk, line 797: HA does not use `InterfaceChoice.Default` if any adapters exist which are "enabled" but not "default" (i.e., not single-interface). We may also want to have this condition here.
On the other hand, adding that condition would mean the default configuration will behave the same as `use_all_interfaces`  except in this very narrow single-interface case (where it will likely have the same result in practice).

trisk (changes requested):
Thank you for implementing these changes to the interface configuration (and adding all the unit tests as well!). I think it will make discovery a lot more robust. I am wondering whether we should make the `InterfaceChoice.Default` usage more consistent with HA (limit to configurations where there is only one interface enabled).

music_assistant/helpers/util.py
fmurodov: This is a good idea. Especially in my setup where I always have multiple interfaces, so I have to switch to all interfaces from the UI.
But I understand that my setup is more of an exception than most users have. And I'm not the one who can decide on this

music_assistant/controllers/webserver/controller.py
```diff
@@ -226,10 +226,10 @@ async def get_config_entries(
                 key=CONF_BIND_IP,
                 type=ConfigEntryType.STRING,
                 default_value="0.0.0.0",
-                options=[ConfigValueOption(x, x) for x in {"0.0.0.0", "::", *ip_addresses}],
+                options=[ConfigValueOption(x, x) for x in {"0.0.0.0", *ip_addresses}],
                 label="Bind to IP/interface",
                 description="Bind the (web)server to this specific interface. \n"
-                "Use 0.0.0.0 or :: to bind to all interfaces. \n"
+                "Use 0.0.0.0 to bind to all interfaces (both IPv4 and IPv6). \n"
```
trisk, line 234: As with `streams_controller.py`, I think asking the user to leave the field empty for the default would make more sense.

music_assistant/helpers/util.py
```diff
@@ -357,6 +360,17 @@ def call() -> tuple[str, ...]:
             primary_ip = ""
         finally:
             _sock.close()
+        # fall back to IPv6 if no IPv4 primary found (e.g. IPv6-only networks)
+        if not primary_ip:
+            _sock6 = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
+            _sock6.settimeout(0)
+            try:
+                _sock6.connect(("2001:db8::1", 1))
```
trisk, line 369: Maybe this is a common practice I'm not aware of, but it doesn't look like RFC 3849 intends the `2001:db8::/32` prefix to be used outside of actual documentation. Maybe another place to use `ifaddr`?

music_assistant/controllers/streams/streams_controller.py
```diff
@@ -281,10 +281,11 @@ async def get_config_entries(
                 key=CONF_BIND_IP,
                 type=ConfigEntryType.STRING,
                 default_value="0.0.0.0",
-                options=[ConfigValueOption(x, x) for x in {"0.0.0.0", "::", *ip_addresses}],
+                options=[ConfigValueOption(x, x) for x in {"0.0.0.0", *ip_addresses}],
                 label="Bind to IP/interface",
                 description="Start the stream server on this specific interface. \n"
-                "Use 0.0.0.0 or :: to bind to all interfaces, which is the default. \n"
+                "Use 0.0.0.0 to bind to all interfaces (both IPv4 and IPv6), "
+                "which is the default. \n"
```
trisk: From a UX perspective, leaving the field empty for the default behaviour would make more sense.

music_assistant/providers/airplay/provider.py
```diff
@@ -179,7 +180,7 @@ async def _setup_player(
         # shairport-sync instances running on other machines
         if model == "ShairportSync":
             # Check if this is a local address (127.x.x.x or matches our server's IP)
-            if address.startswith("127.") or address == self.mass.streams.publish_ip:
+            if address.startswith(("127.", "::1")) or address == self.mass.streams.publish_ip:
```
trisk: Can we use `addr.is_loopback or address.is_link_local` here?

music_assistant/providers/sonos/helpers.py
```diff
@@ -20,4 +20,10 @@ def get_primary_ip_address(discovery_info: AsyncServiceInfo) -> str | None:
             # filter out APIPA address
             continue
         return address
+    # fall back to IPv6 addresses if no usable IPv4 address found
+    for address in discovery_info.parsed_addresses(IPVersion.V6Only):
+        if address.startswith(("::1", "fe80")):
```
trisk: `address.is_loopback or address.is_link_local`?

trisk (changes requested):
I've looked through the full PR and have a few suggestions.

OzGav:
Some minor conflicts to resolve in addition to the review comments above. FYI 2.8 will be shipping on 25 March.

fmurodov:
> Some minor conflicts to resolve in addition to the review comments above. FYI 2.8 will be shipping on 25 March.

Hope we can take it to 2.8 but it's too close date and it takes more time for me. Resolved conflicts, working suggestions

music_assistant/controllers/webserver/controller.py
fmurodov: I don’t think that leaving an empty option is good here and dont know how it would look in UI. But it’s something outside of the scope of this PR.

music_assistant/controllers/streams/streams_controller.py
fmurodov: Same as for webserver, it's out of scope. Or can be decided between maintainers

music_assistant/helpers/util.py
fmurodov: The UDP socket trick with connect() never sends traffic, so it could be any address (same pattern is used for IPv4 with 10.254.254.254). It just asks kernel which source address it would route through to determine the default IP. 

ifaddr can enumerate all addresses but doesn't provide routing preference, so replacing this would need additional logic. Can be done but outside the scope of this PR.

music_assistant/providers/airplay/provider.py
fmurodov: yes, fixed

music_assistant/providers/sonos/helpers.py
fmurodov: fixed

music_assistant/controllers/webserver/controller.py
trisk: Understood, I can raise a follow-up issue for this.

music_assistant/helpers/util.py
trisk: I know we're just using this to query the routing table, but it would be nice to avoid using a reserved prefix regardless. Unfortunately IPv6 doesn't have a direct equivalent of an IPv4 private address, and using a link-local address might not have the result we want. It doesn't seem like there's any other IPv6 reserved prefix meant for this specific use, and this is harmless enough so I'm ok with leaving it for now.

trisk (approved):
Thanks for the update. I think this looks good to go, with the future work beyond the scope of this PR noted in the previous comments.

trisk:
@bdraco @marcelveldt is this ready to merge? Not sure what's causing the test failure but it's affecting all other PRs right now.

marcelveldt:
I'll have a look at the CI tests

music_assistant/helpers/util.py
```diff
@@ -715,6 +735,60 @@ def get_port_from_zeroconf(discovery_info: AsyncServiceInfo) -> int | None:
...
+    has_ipv6 = False
+    interface_ips: list[str] = []
+    for adapter in adapters:
+        for ip_config in adapter.ips:
+            if ip_config.is_IPv6:
+                ip_tuple = cast("tuple[str, int, int]", ip_config.ip)
+                addr = ip_address(ip_tuple[0])
+                if isinstance(addr, IPv6Address) and not addr.is_loopback:
+                    has_ipv6 = True
+                    if not addr.is_global:
```
MarvinSchenkel, line 764: I just did some local testing and noticed the following error:

```
2026-03-23 13:57:28.589 WARNING (MainThread) [zeroconf] Error with socket 26 (('fe80::ce81:b1c:bd2c:69e', 5353, 0, 25))): [Errno 40] Message too long
  File "/Users/marvin/.pyenv/versions/3.13.11/lib/python3.13/asyncio/selector_events.py", line 1276, in sendto
    self._sock.sendto(data, addr)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
OSError: [Errno 40] Message too long
```

According to Claude this is caused by link local ipv6 sockets having small packet size limits and we should exclude them here as well. Should we change this to:
```
if addr.is_link_local or addr.is_global:
  continue
interface_ips.append(f"{ip_tuple[0]}%{ip_tuple[2]}")
```

MarvinSchenkel: I think we can still merge this as part of 2.8 release after this.

fmurodov: Its issue of link local with small mtu, no need to exclude global. Easy fix

fmurodov: @MarvinSchenkel Done. Can you have another look?

MarvinSchenkel: Yep, this particular fix looks good now. However, I just discovered a blocking issue. On this branch, the Chromecast protocol for my Wiim device is not discovered. It is always grayed out. When I switch back to the `dev` branch and restart, it instantly shows up. Only my WiiM seems to be affected, other Chromecast protocls (Nest hub Nvidia shield) show up. 

It seems to be due to the `ip_version` being set to `IPVersion.All`. Whenever I force, `IPVersion.V4Only` in your branch, it instantly works. Talking about changing line 792 of util.py to `return {"ip_version": IPVersion.V4Only, "interfaces": interface_ips}`

Any idea? Honestly, merging this PR as part of 2.8 is starting to feel a bit risky. I would be more comfortable merging this _after_ the release after ironing out the above issue.

fmurodov: Agree that 2.8 is too close and too risky. But also tired of keeping this open when every new merge creates more conflicts. 

About chromecast issue. I will try to reproduce it (or guess with Claude what could cause the issue), but I don't have real chromecast devices.

MarvinSchenkel: Yeah totally feel you. More than happy to get this merged as soon as possible after the 2.8 release and do the chromecast fix as a follow up to avoid those merge conflicts.

trisk: I have a variety of Chromecast devices and can try to reproduce. I assume @MarvinSchenkel has a mostly dual-stack network? I wonder if the affected device is IPv4-only.

MarvinSchenkel: Yep, dual stack network. Just ran a few command line tools and it looks like the WiiM Amp is indeed ipv4 only:
* ARP (IPv4): Found at 192.168.1.243 with MAC xxxxxxxxx
* NDP (IPv6): No entry — the WiiM has no IPv6 address on your network
* ping6: Fails — no IPv6 connectivity
* mDNS IPv6 resolution: No response

MarvinSchenkel: This turned out to be a macos only issue, works fine in our Docker image :). See [here](https://github.com/music-assistant/server/pull/3235/changes/815e675ddebe278510d7f2528b2da03a533979de) for more info

marcelveldt (approved):
Nice work @fmurodov !

PR #3246, opened 2026-02-26
Fix Qobuz provider using HTTP instead of HTTPS for all API calls

teancom:
- Switch both _get_data and _post_data URLs from http:// to https://
- Remove ssl=False on POST requests that was disabling TLS verification


MarvinSchenkel (approved):
LGTM, thanks @teancom

PR #3247, opened 2026-02-26
Qobuz provider tidy up and add tests

teancom:
A few "tidy up" stuff (typo of Qobux, an extra forward slash in a URL, some defensive programming for list access) and adding some ground-truth unit tests. I'm going through the Qobuz provider and wanted to make sure I wasn't going to break anything (as much as I can prevent that), so wanted to get tests in place to help with that.

The AI-generated summary:

Harden Qobuz API response parsing against missing/null fields:
- Guard against missing 'artist' key in get_artist_albums results
- Handle missing 'upc' field on albums without raising KeyError
- Fix performer field access (dict, not string) when filtering Various Artists
- Handle missing 'performers' key in track fallback artist resolution
- Skip malformed performer strings that lack role delimiter
- Fix indentation bug that appended artist outside conditional block
- Remove errant leading slash from reportStreamingEnd endpoint
- Fix "Qobux" typo in class docstring

Add comprehensive unit tests covering all parse methods (_parse_artist,
_parse_album, _parse_track, _parse_playlist) with edge cases for missing
fields, Various Artists filtering, and playlist editability.

### DIFF
--- a/music_assistant/providers/qobuz/__init__.py
+++ b/music_assistant/providers/qobuz/__init__.py
@@ -623,7 +623,8 @@ class QobuzProvider(MusicProvider):
         ):
             album.album_type = AlbumType.SINGLE
         elif (
-            album_obj.get("product_type", "") == "compilation" or "Various" in album.artists[0].name
+            album_obj.get("product_type", "") == "compilation"
+            or album.artists[0].item_id == VARIOUS_ARTISTS_ID
         ):
             album.album_type = AlbumType.COMPILATION
         elif (
@@ -689,7 +690,10 @@ class QobuzProvider(MusicProvider):
         )
         if isrc := track_obj.get("isrc"):
             track.external_ids.add((ExternalID.ISRC, isrc))
-        if track_obj.get("performer") and "Various" not in track_obj["performer"].get("name", ""):
+        if (
+            track_obj.get("performer")
+            and str(track_obj["performer"].get("id", "")) != VARIOUS_ARTISTS_ID
+        ):
             artist = self._parse_artist(track_obj["performer"])
             if artist:
                 track.artists.append(artist)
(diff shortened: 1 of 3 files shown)


music_assistant/providers/qobuz/__init__.py
```diff
@@ -683,22 +689,24 @@ async def _parse_track(self, track_obj: dict[str, Any]) -> Track:
         )
         if isrc := track_obj.get("isrc"):
             track.external_ids.add((ExternalID.ISRC, isrc))
-        if track_obj.get("performer") and "Various " not in track_obj["performer"]:
+        if track_obj.get("performer") and "Various" not in track_obj["performer"].get("name", ""):
```
MarvinSchenkel: Not sure if removing the space is the right approach here? Then we would also match "Variouslywhatever" ?

tests/providers/qobuz/test_parsers.py
```diff
@@ -0,0 +1,516 @@
...
+        assert result.metadata.images is not None
+        assert len(result.metadata.images) == 1
+
+    async def test_parse_track_performer_various_artists(
+        self, mock_provider: QobuzProvider
+    ) -> None:
+        """Test parsing track with Various Artists performer is filtered."""
+        track_with_various = deepcopy(TRACK_OBJ)
+        track_with_various["performer"] = {
+            "id": 145383,
```
MarvinSchenkel: Please use the constant from the init for all these various artist IDs

music_assistant/providers/qobuz/__init__.py
teancom: Great catch, I had noted that as probably not being right and then got distracted and never came back. Updated this and one other place where we could be using the constant instead and stop parsing strings. 🙇

marcelveldt:
I'm assuming you tested the Qobuz provider with your own account after these changes ?

teancom:
Yep, for the last few days.

MarvinSchenkel (approved):
Thanks @teancom 👍

PR #3250, opened 2026-02-26
Fix sendspin metadata sending wrong progress when paused

balloob:
## Summary
- Set `playback_speed=0` when not playing (was hardcoded to `1000`), so the server and clients know progress shouldn't advance during pause
- Use `Player.corrected_elapsed_time` (which only interpolates during PLAYING) instead of `PlayerMedia.corrected_elapsed_time` (which always interpolates regardless of playback state, accumulating hours of wall-clock time while paused)

These two issues caused sendspin clients to show wildly incorrect progress (e.g. `795:54 / 04:23`) after reconnecting to a paused stream.

## Test plan
- [x] Play a track via sendspin, pause, wait, reconnect with a sendspin client — progress should show the correct pause position
- [x] Verify progress still advances correctly during normal playback


maximmaxim345 (approved):
Works now when paused and playing, tested with sendspin-js.

PR #3254, opened 2026-02-26
Remove HLS streams from the Audiobookshelf provider based on user feedback

fmunkes:
Given the feedback I got in https://github.com/music-assistant/support/issues/4586 and https://github.com/music-assistant/support/issues/4754, we can safely remove the HLS stream option introduced in https://github.com/music-assistant/server/pull/3079 from the Audiobookshelf provider again.

Progress syncing via sessions seems to have solved these issues.


MarvinSchenkel (approved):
Always good to clean up, thanks @fmunkes

PR #3259, opened 2026-02-27
Make Smart Fades respect the PCM format

MarvinSchenkel:


music_assistant/controllers/streams/smart_fades/analyzer.py
```diff
@@ -66,7 +67,7 @@ async def analyze(
                 len(audio_data),
             )
             # Convert PCM bytes to numpy array and then to mono for analysis
-            audio_array = np.frombuffer(audio_data, dtype=np.float32)
+            audio_array = self._pcm_bytes_to_float32(audio_data, pcm_format)
```
marcelveldt, line 70: why are we converting ? why are we not passing the audio format along so the data will be in that shape ?

MarvinSchenkel: Because librosa expects as float 32 numpy array for analysis. Before we just naively assumed we were receiving F32 pcm data for the conversion to a numpy array, but this change makes it aware of the source format and correctly applies the necessary calculations.

music_assistant/controllers/streams/smart_fades/analyzer.py
```diff
@@ -255,3 +256,18 @@ async def _analyze_track_beats(
...
+
+    def _pcm_bytes_to_float32(
+        self,
+        audio_data: bytes,
+        pcm_format: AudioFormat,
+    ) -> npt.NDArray[np.float32]:
+        """Convert raw PCM bytes to a normalised float32 numpy array."""
+        if pcm_format.content_type == ContentType.PCM_S16LE:
+            return (np.frombuffer(audio_data, dtype="<i2").astype(np.float32)) / np.float32(32768.0)
+        if pcm_format.content_type == ContentType.PCM_S32LE:
```
marcelveldt, line 276: if you want to be complete, you should add also add S24LE here...

marcelveldt (approved):

PR #3265, opened 2026-02-28
fix: delay stream/end until client buffers drain at end of album

teancom:
When the last track in a queue finishes, the server must wait for clients to finish playing their buffered audio before sending stream/end (which clears client buffers per the SendSpin spec).

The drain sleep runs inline in _run_playback's finally block so MA still considers the track "playing" and the UI stays in sync with actual audio output. On cancellation (e.g. play_media starts a new track), the sleep is interrupted via CancelledError.

After the drain completes, the queue is checked for newly enqueued tracks (e.g. user pressed "play next" during the last track). If found, play_index is triggered instead of stopping — playback continues after the current track finishes naturally.

Also:
- Active channel filtering for stream end delay computation
- Debug logging for exception fallback in delay computation

### DIFF
--- a/music_assistant/providers/sendspin/playback.py
+++ b/music_assistant/providers/sendspin/playback.py
@@ -735,6 +735,25 @@ class SendspinPlaybackSession:
                 commit_task.cancel()
                 with suppress(asyncio.CancelledError, Exception):
                     await commit_task
+            # On clean EOF, wait for clients to finish playing their
+            # buffered audio before sending stream/end (which clears
+            # client buffers per the SendSpin spec).  The sleep keeps
+            # _run_playback alive so MA still considers the track
+            # "playing" and the UI stays in sync with actual audio.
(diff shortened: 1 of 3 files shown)


marcelveldt:
Thanks for the PR !
I'm going to look if we can solve this without having to rely on querying the queue controller for 2 reasons:

1) playback may also be another source, like spotify connect
2) A player implementation should not hav eto interact with the queue controller directly

teancom:
Makes complete sense. I'll close this. Purely as a favor / optional, could you tag me here or on the discord when you have a solution? I'd love to read through it and learn. 🙇‍♂️

marcelveldt:
Fixed in these 2 commits:

https://github.com/music-assistant/server/commit/19524668a784fc25ccb2d0156bb1eb834c4bc966

https://github.com/music-assistant/server/commit/ebc6ea74db9a69b088ff8ba1f43540a356632a87

PR #3271, opened 2026-03-01
Fix helper for safe type evaluation

apophisnow:
Fixes unsafe use of eval.


music_assistant/helpers/api.py
```diff
@@ -403,8 +417,13 @@ def parse_value(  # noqa: PLR0911
         logging.getLogger(__name__).warning(err)
         return None
     if origin is type:
+        # Resolve string to a type using a safe allowlist.
+        # Never use eval() here — the value comes from API input.
         assert isinstance(value, str)  # for type checking
-        return eval(value)
+        if value not in _SAFE_TYPE_FROM_STRING:
```
marcelveldt: How would this work if the type string is a list of something, or a tuple of some other type etc. ?

apophisnow: You mean like list[str] or something like that?

apophisnow: Revised the approach. And added tests.

marcelveldt: will it still work with our own models ?
so if the type hint e.g. "list[Track'"

apophisnow: I think this needs more thought. I will make it a draft for now.

marcelveldt (approved):
Looks good, thanks @apophisnow !

PR #3276, opened 2026-03-02
Add NFS Filesystem Provider

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/1198


music_assistant/providers/filesystem_nfs/provider.py
```diff
@@ -0,0 +1,112 @@
...
+            msg = f"NFS mount failed: {err}"
+            raise ProviderUnavailableError(msg) from err
+        await self.check_write_access()
+
+    async def unload(self, is_removed: bool = False) -> None:
+        """Handle unload/close of the provider.
+
+        Called when provider is deregistered (e.g. MA exiting or config reloading).
+        """
+        await self.unmount(ignore_error=True)
```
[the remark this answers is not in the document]
OzGav: There are pros and cons here.

I understand NFS mounts can hang indefinitely if the server goes away, which would mean umount blocks and stops the MA shutdown. That’s the scenario ignore_error=True protects against.
I understand SMB is less prone to this because cifs mounts tend to time out rather than hang, but it’s arguably still a risk there too.
So I’d prefer to keep ignore_error=True here. A failed unmount during shutdown is harmless — the mount point gets cleaned up when the container restarts or the /tmp directory is cleared. A hung shutdown of MA is worse.​​​​​​​​​​​​​​​​

music_assistant/providers/filesystem_nfs/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    # check if valid dns name is given for the host
+    server = str(config.get_value(CONF_HOST))
+    if not await get_ip_from_host(server):
+        msg = f"Unable to resolve {server}, make sure the address is resolvable."
+        raise SetupFailedError(msg)
+    # check if export path is valid
+    export_path = str(config.get_value(CONF_EXPORT_PATH))
+    if not export_path or not export_path.startswith("/"):
```
MarvinSchenkel: We have a helper for path validation under helpers/security.py > is_safe_path

MarvinSchenkel: I would actually move this check to handle_async_init so it is always checked before the provider loads.

music_assistant/providers/filesystem_nfs/provider.py
```diff
@@ -0,0 +1,112 @@
...
+
+    async def handle_async_init(self) -> None:
+        """Handle async initialization of the provider."""
+        if not await exists(self.base_path):
+            await makedirs(self.base_path)
+        try:
+            # unmount first to cleanup any unexpected state
+            await self.unmount(ignore_error=True)
+            await self.mount()
+        except (LoginFailed, SetupFailedError, ProviderUnavailableError):
```
MarvinSchenkel: This can be removed

music_assistant/providers/filesystem_nfs/provider.py
```diff
@@ -0,0 +1,112 @@
...
+        self.logger.debug("Mounting %s:%s to %s", server, full_export, self.base_path)
+        self.logger.log(VERBOSE_LOG_LEVEL, "Using mount command: %s", " ".join(mount_cmd))
+        returncode: int
+        output: bytes
+        returncode, output = await check_output(*mount_cmd)
+        if returncode != 0:
+            msg = f"NFS mount failed with error: {output.decode()}"
+            raise ProviderUnavailableError(msg)
+
+    def _get_mount_options(self) -> list[str]:
```
MarvinSchenkel, line 95: We should consider adding `soft` mounting in the options here along with timeout and retries to prevent MA hanging when the NFS host is unreachable. Something like `"soft", "timeo=30", "retrans=5"`

MarvinSchenkel (changes requested):
Few minor comments, other than that looks nice @OzGav 👏

music_assistant/providers/filesystem_nfs/__init__.py
OzGav: That check is just to make sure it is a valid absolute path so did you mean supplement with is_safe_path or replace it?
in the mean time I have moved the check but I note that I based that on the SMB provider so would you like me to change that as well in a separate PR?

music_assistant/providers/filesystem_nfs/provider.py
OzGav: Ok I understand you are referring to the redundant re-raise. I have removed that, but again I note that the SMB does something similar so I can fix that separately as well if you want

music_assistant/providers/filesystem_nfs/provider.py
OzGav: Done

music_assistant/providers/filesystem_nfs/__init__.py
MarvinSchenkel: It's a security measure to disallow path traversals etc by injecting malicious configs. So technically the SMB provider should also have it.

MarvinSchenkel: Ok in hindsight, the validation should be moved back into `setup()` as that is the more logical place for it. This seems to be called as well during a reload. So basically what you add before but then add the `is_safe_path` to the export_Path as an additional check

music_assistant/providers/filesystem_nfs/provider.py
```diff
@@ -0,0 +1,128 @@
...
+        ]
+
+        self.logger.debug("Mounting %s:%s to %s", server, full_export, self.base_path)
+        self.logger.log(VERBOSE_LOG_LEVEL, "Using mount command: %s", " ".join(mount_cmd))
+        returncode: int
+        output: bytes
+        returncode, output = await check_output(*mount_cmd)
+        if returncode != 0:
+            msg = f"NFS mount failed with error: {output.decode()}"
+            raise ProviderUnavailableError(msg)
```
MarvinSchenkel: I would throw a `SetupFailedError` here as well

music_assistant/providers/filesystem_nfs/provider.py
```diff
@@ -0,0 +1,128 @@
...
+            msg = "Invalid export path: must be an absolute path starting with /"
+            raise SetupFailedError(msg)
+
+        if not await exists(self.base_path):
+            await makedirs(self.base_path)
+        try:
+            # unmount first to cleanup any unexpected state
+            await self.unmount(ignore_error=True)
+            await self.mount()
+        except OSError as err:
```
MarvinSchenkel, line 48: And also catch the `SetupFailedError` here because `mount` can raise it.

MarvinSchenkel:
Could you also add the `CONF_ENTRY_PROPAGATE_GENRES` which was recently added to SMB / File system for genre propagation on files?

music_assistant/providers/filesystem_nfs/__init__.py
OzGav: OK that is all done

music_assistant/providers/filesystem_nfs/provider.py
OzGav: done

OzGav:
CONF_ENTRY_PROPAGATE_GENRES added to the config entries

music_assistant/providers/filesystem_nfs/provider.py
OzGav: I think we might not need to explicitly catch SetupFailedError here — it should propagate up naturally since it's not an OSError. The OSError catch raises a SetupFailedError so that can be handled further up

MarvinSchenkel (approved):
Looks good now Gav, thanks!

PR #3283, opened 2026-03-02
Fix userid missing in player queue controller's resolve media item method

fmunkes:
Fixes the userid not being passed to the resolve_media_items method of the player_queues controller


marcelveldt (approved):
Thanks, @fmunkes !

PR #3295, opened 2026-03-03
Check if HEOS queue exists before clearing

Tommatheussen:
I recently added internal HEOS queue clearing when we start playing media from MA (otherwise the HEOS players start playing older content after a stream finishes). However, it seems HEOS can't handle the clear command when the queue is empty, throwing `Requested data not available (4)` errors.
This became more clear with the merge player providers, as playing from Airplay already clears the internal HEOS queue, thus resulting in not being able to start playing anything on the native provider anymore. 

With the PR we first check if there is anything in the queue, only then do we clear it.


marcelveldt (approved):
Thanks, @Tommatheussen !

PR #3296, opened 2026-03-03
Ensure soundcloud playlist genre is added as a list

jozefKruszynski:
Similar fix was added for tracks here:
https://github.com/music-assistant/support/issues/3107
This fix addresses playlist genre parsing


robsonke (approved):
Thanks!

PR #3301, opened 2026-03-04
Handle outdated TLS/SSL configuration in radio streams

OzGav:
Partially addresses https://github.com/orgs/music-assistant/discussions/5016

A user is trying to stream from an Icecast server. The stream server uses an outdated TLS/SSL configuration (it lacks forward secrecy ciphers). Python's default security policy rejects these older ciphers. This PR seeks to relax the requirements for radio streams specifically if a handshake failure occurs, since the audio data is public broadcast content this seems reasonable.

Whilst the stream will play without this fix, we won't receive any track metadata since we have to negotiate a TLS connection to get the ICY metadata.

### DIFF
--- a/music_assistant/helpers/audio.py
+++ b/music_assistant/helpers/audio.py
@@ -9,13 +9,14 @@ import re
 import struct
 import time
 from collections.abc import AsyncGenerator
+from contextlib import asynccontextmanager
 from functools import partial
 from io import BytesIO
-from typing import TYPE_CHECKING, Final, cast
+from typing import TYPE_CHECKING, Any, Final, cast
 
 import aiofiles
 import shortuuid
-from aiohttp import ClientTimeout
+from aiohttp import ClientConnectorSSLError, ClientTimeout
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Looks good to me. One additional question: Shouldn't we also add this as wrappers for `_update_hls_radio_metadata` and `get_hls_substream`?

OzGav:
We could but I don’t know if this would be encountered with any decent stream. I think it is just these 20 year old stations using shoutcast that will be an issue but we can if you think there is value.

MarvinSchenkel:
> We could but I don’t know if this would be encountered with any decent stream. I think it is just these 20 year old stations using shoutcast that will be an issue but we can if you think there is value.

If you think those do 20 year old stations are not routed through those functions, then I think we are fine as is

OzGav:
Yeah I don’t think so. They aren’t using HLS to my knowledge. Indeed, in the context of icecast it can’t natively stream HLS

MarvinSchenkel (approved):

PR #3302, opened 2026-03-04
Fix album image missing from queue item responses

### LINKED ISSUE #40: Add artist and album images to get_queue response
Recently get_queue was enhanced to include extra fields. https://github.com/music-assistant/backlog/issues/15#issue-3513453658 Two of those are artist and album images. However, the response is always showing NULL for these items. The following example is while playing an in-library album from the local file system provider wheer all imagery is available locally

```
media_player.unnamed_room:
  queue_id: RINCON_74CA607DC13E01400
  active: true
  name: Sonos Roaming
  items: 30
  shuffle_enabled: false
  repeat_mode: "off"
  current_index: 1
  elapsed_time: 14
  current_item:
    queue_item_id: 642d175da99c4fe5b10b54b193995c72
    name: AC/DC - Highway to Hell
    duration: 208
    media_item:
      media_type: track
      uri: library://track/72
      name: Highway to Hell
      version: ""
      image: >-
        http://192.168.1.166:8095/imageproxy?path=ACDC/Highway%2520to%2520Hell/folder.jpg&provider=filesystem_smb--oR6kM33Q&size=0
      favorite: false
      explicit: null
      artists:
        - media_type: artist
          uri: library://artist/3
          name: AC/DC
          version: ""
          image: null
      album:
        media_type: album
        uri: library://album/8
        name: Highway to Hell
        version: ""
        image: null
    stream_title: null
    stream_details:
      provider: filesystem_smb--oR6kM33Q
      item_id: ACDC/Highway to Hell/ACDC -- Highway To Hell.mp3
      content_type: mp3
      sample_rate: 44100
      bit_depth: 16
      bit_rate: 128
```

chrisuthe:
Include artist image metadata in the SQL queries for tracks and albums so that ItemMapping objects have their image field populated. Also set the image field on the album dict in _parse_db_row for ItemMapping compatibility.

Previously, the artist subqueries in tracks.py and albums.py did not extract metadata, so artist images were always null. The album image was extracted into an images array but never set as a singular image field on the album dict itself.
This is to fix https://github.com/music-assistant/backlog/issues/40

@OzGav  I think this accomplishes the ask in that frontend issue, if you want to take a look.

### DIFF
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -1109,8 +1109,10 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
             if (album_images := track_album.get("images")) and (
                 album_thumb := next((x for x in album_images if x["type"] == "thumb"), None)
             ):
-                # copy album image to itemmapping single image
+                # copy album image to itemmapping single image (on the track)
                 db_row_dict["image"] = album_thumb
+                # also set image on the album dict for ItemMapping compatibility
+                track_album["image"] = album_thumb
                 if db_row_dict["metadata"].get("images"):
                     # merge album image with existing images
                     db_row_dict["metadata"]["images"] = [
@@ -1119,6 +1121,7 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
                     ]
                 else:
                     db_row_dict["metadata"]["images"] = [album_thumb]
+
         return db_row_dict
 
     @final


OzGav (approved):
Yep tested working!

music_assistant/controllers/media/albums.py
```diff
@@ -67,7 +67,8 @@ def __init__(self, mass: MusicAssistant) -> None:
                 'provider', 'library',
                     'name', artists.name,
                     'sort_name', artists.sort_name,
-                    'media_type', 'artist'
+                    'media_type', 'artist',
+                    'images', json_extract(artists.metadata, '$.images')
```
marcelveldt: why do we want the artist image to be present on an album ?

music_assistant/controllers/media/base.py
```diff
@@ -1119,6 +1121,20 @@ def _parse_db_row(db_row: Mapping[str, Any]) -> dict[str, Any]:
...
+            if isinstance(artists_list, list):
+                for artist_dict in artists_list:
+                    if not isinstance(artist_dict, dict):
+                        continue
+                    if artist_images := artist_dict.get("images"):
+                        artist_thumb = next(
+                            (x for x in artist_images if x.get("type") == "thumb"), None
+                        )
+                        if artist_thumb:
+                            artist_dict["image"] = artist_thumb
```
marcelveldt: this adds a bit of overhead so I'm wondering how important it is that you have the artist image on a simple/linked artist within an album. You wont be looking at that anyways

music_assistant/controllers/media/albums.py
chrisuthe: @marcelveldt  Valid, in the case of Album we have album art, we wouldn't want/need to provide artist image for additional use/fallback unless explicitly built for. I'm curious if we want to somehow aknowledge that get_queue is expected to return null for artist image on album items, even if album art is null? I'll pull this out for album.

music_assistant/controllers/media/base.py
chrisuthe: The intent of your original post seems to really focus on album art, I agree (https://github.com/music-assistant/backlog/issues/15#issue-3513453658). 

I see a use for artist photos (Plexamp, for instance, uses them heavily as things like backgrounds of what we would consider a player card, especially when they have rich ones like concert photos etc) but I am not in that space, so we can pull artist complete if you want, I would ask if we want to pull it from get_queue completely, and if we do that, should supporting media be seperated completely and left to Queue to decide it needs to fetch (Album art, Artist Art, Lyrics come to mind).  That seems more expensive but only if you want to use it. 

Leaving album art (it's most used) and saying "you can get artist photos another way" seems like a solid compromise, but how do you want me to tackle it?

music_assistant/controllers/media/albums.py
marcelveldt: all listings with multiple items are supposed to be compact objects, because these listings can be large.
That is also why we have ItemMapping, so we dont have to drag in complete child objects.

But... for the media controllers we also have the "get item" methods that lets you fetch a single item with full details. 

If an API client wants to get all fancy details of a media item in the queue, they can use those calls.

chrisuthe: Makes sense, get_item can be iterated through if you want to get fancy when it's played. I think that covers both situations and I suggest we pull the artist image from get_queue completely.

chrisuthe: Cleaned up, we can decide seperately if removing the artist info from get_queue is rational.

music_assistant/controllers/media/base.py
chrisuthe: removed per conversation on other bit, this lighter version gives minimum needed to dispaly a queue without additional per-item fetches, other bits when a media is being played can be achieved using get_item as that item is the focus of now playing or some sort of action.

chrisuthe: @OzGav  This does impact the original ask, we now are providing album art but intentionally not providing artist image, we should consider removing it from the get_queue as it was an overstep of what that required.

OzGav:
Yes I would say we should remove the keys from the get_queue response if we aren’t going to supply the info. However, I think we should have kept it as it’s a valid use case for card builders to show both the artist fanart and album image

PR #3306, opened 2026-03-04
Cleanup genre, alias and media item mapping orphans during sync

jozefKruszynski:
The genre sync process should cleanly handle orphaned entries, across genres themselves, but also aliases and media mappings.

### DIFF
--- a/music_assistant/controllers/media/genres.py
+++ b/music_assistant/controllers/media/genres.py
@@ -46,6 +46,17 @@ if TYPE_CHECKING:
     from music_assistant import MusicAssistant
 
 
+MEDIA_TABLES: tuple[tuple[str, MediaType], ...] = (
+    (DB_TABLE_TRACKS, MediaType.TRACK),
+    (DB_TABLE_ALBUMS, MediaType.ALBUM),
+    (DB_TABLE_ARTISTS, MediaType.ARTIST),
+    (DB_TABLE_PLAYLISTS, MediaType.PLAYLIST),
+    (DB_TABLE_RADIOS, MediaType.RADIO),
+    (DB_TABLE_AUDIOBOOKS, MediaType.AUDIOBOOK),
+    (DB_TABLE_PODCASTS, MediaType.PODCAST),
+)
+
+
 class GenreController(MediaControllerBase[Genre]):
     """Controller for Genre entities."""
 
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/media/genres.py
```diff
@@ -682,6 +682,76 @@ async def _bulk_scan_media_genres(self) -> None:
...
+            DB_TABLE_GENRES,
+            f"translation_key IS NULL "
+            f"AND NOT EXISTS ("
+            f"  SELECT 1 FROM {gm} WHERE {gm}.genre_id = {DB_TABLE_GENRES}.item_id"
+            f")",
+        )
+        genres_deleted = genres_before - await db.get_count(DB_TABLE_GENRES)
+        if genres_deleted:
+            self.logger.info("Genre scan: deleted %d empty non-default genres", genres_deleted)
+
```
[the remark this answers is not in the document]
jozefKruszynski: Deleting the playlog entries ensures that the genre cannot be clicked in the playlog after it has been removed.

The use of translation_key is a simple way to find the none default genres, as there is no way to add a translation key to a genre that is not part of the default set.

The deletion of genres that are empty and none default, was discussed with @MarvinSchenkel in a private discord chat.

music_assistant/controllers/media/genres.py
```diff
@@ -682,6 +682,76 @@ async def _bulk_scan_media_genres(self) -> None:
...
+            self.logger.info("Genre scan: removed %d stale genre mappings", mappings_removed)
+
+        # Delete playlog entries for empty non-default genres before removing them, to avoid
+        # orphaned playlog rows pointing to genres that no longer exist.
+        await db.delete_where_query(
+            DB_TABLE_PLAYLOG,
+            f"media_type = '{MediaType.GENRE.value}' "
+            f"AND item_id IN ("
+            f"  SELECT item_id FROM {DB_TABLE_GENRES} "
+            f"  WHERE translation_key IS NULL "
```
MarvinSchenkel, line 730: Can you add a comment here about `translation_key IS NULL` being the indicator of non default genres?

music_assistant/controllers/media/genres.py
```diff
@@ -682,6 +682,76 @@ async def _bulk_scan_media_genres(self) -> None:
...
+
+        A mapping is considered stale when the alias stored in the mapping is no longer present
+        in the media item's current metadata.genres. This includes items where metadata.genres
+        is empty or null — all mappings for such items are removed. Empty non-default genres
+        (those without a translation_key) are also deleted.
+        """
+        db = self.mass.music.database
+        gm = DB_TABLE_GENRE_MEDIA_ITEM_MAPPING
+
+        media_tables = (
```
MarvinSchenkel: This is now duplicated across a couple of functions, we could turn this into a constant ?

MarvinSchenkel (approved):
Thanks @jozefKruszynski 👏

PR #3308, opened 2026-03-04
Add configurable in-library podcast feeds syncing time to the iTunes Podcast Provider

fmunkes:
This is per this issue / feature request https://github.com/music-assistant/support/issues/5029

I use the sync library method to allow feed syncing of in-library items. If the approach is fine, I'll add something to the docs too. I decreased the cache time of the search as well.

### DIFF
--- a/music_assistant/providers/itunes_podcasts/__init__.py
+++ b/music_assistant/providers/itunes_podcasts/__init__.py
@@ -385,15 +385,15 @@ class ITunesPodcastsProvider(MusicProvider):
         return parsed_podcast  # type: ignore[no-any-return]
 
     async def _cache_set_podcast(self, feed_url: str, parsed_podcast: dict[str, Any]) -> None:
-        # We cache for 12 hours. However, the user may overwrite the cache with a library sync. So
-        # 12 hours is only true for a non-library item, e.g. a search result not added to the
+        # We cache for 24 hours. However, the user may overwrite the cache with a library sync. So
+        # 24 hours is only true for a non-library item, e.g. a search result not added to the
         # library.
         await self.mass.cache.set(
             key=feed_url,
             provider=self.instance_id,
             category=CACHE_CATEGORY_PODCASTS,
             data=parsed_podcast,
-            expiration=60 * 60 * 12,  # 12h
+            expiration=60 * 60 * 24,  # 24h
         )
 
     async def _cache_set_top_podcasts(self, top_podcast_helper: TopPodcastsHelper) -> None:
--- a/music_assistant/providers/itunes_podcasts/__init__.py
+++ b/music_assistant/providers/itunes_podcasts/__init__.py
@@ -231,15 +231,17 @@ class ITunesPodcastsProvider(MusicProvider):
             provider_instance=self.instance_id
         )
         for podcast in podcasts:
-            feed_url: str | None = None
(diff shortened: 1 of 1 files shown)


music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -213,6 +219,54 @@ def _get_podcast_list(self, results: list[PodcastSearchResult]) -> list[Podcast]
...
+                await self._cache_set_podcast(feed_url=feed_url, parsed_podcast=parsed_podcast)
+                self.logger.debug("Synced podcast %s.", podcast.name)
+            except MediaNotFoundError:
+                # If we are not able to refresh the podcast, we must prevent the sync
+                # from deleting the podcast from the library - that is both a breaking change
+                # (pre March 2026) and certainly not desired just because of some downtime.
+                self.logger.warning("Was unable to sync podcast %s (%s).", podcast.name, feed_url)
+                podcast.item_id = feed_url
+                podcast.provider_mappings = {our_provider_mapping}
+                yield podcast
```
MarvinSchenkel, line 282:
```suggestion
                yield podcast
                continue
```

music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -213,6 +219,54 @@ def _get_podcast_list(self, results: list[PodcastSearchResult]) -> list[Podcast]
...
+
+            assert parsed_podcast is not None  # for type checking
+            yield parse_podcast(
```
MarvinSchenkel: Think this assert is not required anymore after `continue` above?

music_assistant/providers/itunes_podcasts/__init__.py
```diff
@@ -134,7 +140,7 @@ async def handle_async_init(self) -> None:
         # 20 requests per minute, be a bit below
         self.throttler = ThrottlerManager(rate_limit=18, period=60)
 
-    @use_cache(3600 * 24 * 7)  # Cache for 7 days
+    @use_cache(3600 * 12)  # Cache for 12 hours
```
MarvinSchenkel: Why do we also need to reduce this cache?

MarvinSchenkel (changes requested):
Few minor things, I think it's a clever trick to offer refreshes for a provider without a proper library.

music_assistant/providers/itunes_podcasts/__init__.py
fmunkes: We don't have too, but I'm pretty sure I didn't put that decorator there. This is the only place, where the itunes API is used, and I added a ThrottlerManager to obey their rules. It is an open API after all. I just grepped through other providers, some have a cache there, some don't. I personally think 7 days is quite extreme, as you are using this method to search for something in your interest.

I reverted it for now, but IMHO we could remove that decorator completely - see for example the podcastindex provider, where we don't have any throttling and caching.

Feel free to adjust this part as you see fit :)

fmunkes:
Added a small follow-up commit, which gives the sync option a proper description, inspired by the builtin-provider. If the user selects "0", no sync is enabled, so the other option is hidden, as is the same with builtin.
Makes it a bit clearer, I think.

MarvinSchenkel (approved):
Thanks @fmunkes 👏

PR #3311, opened 2026-03-05
Feature: Bandcamp Browse support

teancom:
Following on https://github.com/ALERTua/bandcamp_async_api/pull/15, this adds Browse support to the Bandcamp provider, as well as using the newly-available-from-the-library Fan/Follower data to expose Bandcamp's social graph features. So you can browse and stream your friend's music as well. I've tested it down to fairly ridiculous depths, far past what can even been seen in the breadcrumb trail in the Browse window, and it seems to be pretty rock solid. Because this is a decently big change, I also had the AI write the absolute heck outta tests, both ahead of time for TDD and after to catch bugs I found while testing and edge-case stuff. 

All of which is to say that conceptually this is one thing ("Enable Browse with all of the types of things Bandcamp lets you browse") but also it's a pretty big change. If you want me to carve this into two PRs - one to enable just Artists/Albums/Track browsing, and a follow-up to enable the fan/follower stuff, I can probably do that. Just let me know what you'd prefer.

Thanks again for all the work you've done on this. I was *really* missing having a Bandcamp provider and when I saw that you'd written this, I was over the moon.

Commit message follows:

Add Fans/Followers browse with social graph traversal, bump bandcamp-async-library to 0.0.9

Bump bandcamp-async-api to 0.0.9 which adds FanItem, FOLLOWING_FANS,
FOLLOWERS collection types, and fan_id parameter for cross-user queries.

Expose FOLLOWING_FANS and FOLLOWERS as top-level browse folders alongside
Wishlist and Following. Each person is a browsable folder revealing their
Collection, Wishlist, Following, Fans, and Followers — with Fans/Followers
being recursive for social graph exploration at arbitrary depth.

Browse paths use URL slugs (e.g. fans/teancom/followers/yerhot) instead
of numeric fan_ids for human-readable breadcrumbs. An in-memory
_slug_to_fan_id dict maps slugs back to fan_ids; eviction is intentionally
omitted since it would break back-navigation.

Consolidates browse helpers into 3 parameterized methods:
- _browse_person_content(person_id, collection_type) for Collection/Wishlist
- _browse_person_following(person_id) for followed artists
- _browse_person_people(collection_type, base_path, person_id) for Fans/Followers

Adds _PERSON_SUB_ROUTES dispatch table for routing person sub-paths, and
_map_api_errors context manager to centralize the 3-exception mapping
pattern (BandcampMustBeLoggedInError, BandcampRateLimitError,
BandcampAPIError) used across all browse methods.

Splits the old CACHE constant into three tiers:
- CACHE_METADATA (30 days) for artist/album/track data
- CACHE_USER_LISTS (4 hours) for wishlists/following/fans/followers
- CACHE_EMPTY_RESULTS (5 minutes) to avoid hammering API for empty lists

Path parsing hardened against malformed input: empty segments filtered,
person IDs validated as positive integers, unrecognized sub-categories
rejected with InvalidDataError.


music_assistant/providers/bandcamp/__init__.py
```diff
@@ -418,6 +433,290 @@ async def get_artist_toptracks(self, prov_artist_id: str) -> list[Track]:
 
         return tracks[: self.top_tracks_limit]
 
+    # Maps person sub-route names to (method_suffix, CollectionType).
+    # "content" → _browse_person_content, "following" → _browse_person_following,
+    # "people" → _browse_person_people.
+    _PERSON_SUB_ROUTES: dict[str, tuple[str, CollectionType]] = {
```
MarvinSchenkel: Why is this a private constant? Why not just a normal constant at the top of the file?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -418,6 +433,290 @@ async def get_artist_toptracks(self, prov_artist_id: str) -> list[Track]:
...
+            BrowseFolder(
+                item_id=f"person_{person_id}_{sub_id}",
+                provider=self.instance_id,
+                path=f"{base_path}/{sub_id}",
+                name=name,
+            )
+            for sub_id, name in sub_folders
+        ]
+
+    @contextmanager
```
MarvinSchenkel: Shouldn't this be `@asynccontextmanager` ?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -418,6 +433,290 @@ async def get_artist_toptracks(self, prov_artist_id: str) -> list[Track]:
...
+
+        :param path: The path to browse, (e.g. provider_id://artists).
+        """
+        subpath = path.split("://")[1] if "://" in path else ""
+        # Filter empty segments from double-slashes or trailing slashes
+        path_parts = [p for p in subpath.split("/") if p]
+        base = f"{self.instance_id}://"
+
+        # Route fan/follower paths (supports arbitrary nesting depth)
+        if path_parts and path_parts[0] in ("fans", "followers"):
```
MarvinSchenkel: I would also make those slugs constant, which you can reuse everywhere. "fans", "followers" etc. Might event be nice to introduce a `constants.py` in the provider folder if that makes this file too verbose?

MarvinSchenkel (changes requested):
Few small nits, other than that, looks good :-)

music_assistant/providers/bandcamp/__init__.py
teancom: Good question! Fixed (along with the other constants).

music_assistant/providers/bandcamp/__init__.py
teancom: I'd actually argued with the robot over that a couple of times. Non-async was technically correct (the best kind of correct) in that no async work was being done. But it was flagged by reviewer-bots a couple of times and now by you so I am convinced that using async is better if just for consistency's sake. Done.

music_assistant/providers/bandcamp/__init__.py
teancom: I liked the idea of a constants.py, done. (__init__.py is getting pretty big, a split-up refactoring is probably in the near-future, but didn't want to overburden this PR).

teancom:
I discovered during testing that we were doing too many API calls when browsing. I'm going to fix that as part of this, will mark as ready for review once I have a solution I'm happy with.

teancom:
Alright, I'm much happier with it now. Much, much fewer calls to the Bandcamp API when browsing, we now see real usernames in the breadcrumb instead of user IDs, and we're caching those username<->id mappings in a way that survives through a server reboot (😅).

MarvinSchenkel (approved):
Looks good, thanks @teancom !

ALERTua:
I have deployed the Music-Assistant Dev in my Home Assistant and pulled `dev` backend and `main` frontend. I can confirm that Browse for Bandcamp provider works correctly.

PR #3324, opened 2026-03-07
Fix in progress items recommendation for multi-user setups

fmunkes:
Small fix for the recommendation of in-progress item in multi-user setups. For in-library items (i.e. only audiobooks), we need to check the provider_mappings table.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -652,30 +652,47 @@ class MusicController(CoreController):
         # An audiobook can be part of the library, in contrast to podcast episodes.
         # We then need to check the provider mappings table.
         query = (
-            "SELECT p.item_id,p.media_type,p.name,p.image, "
-            "IF(p.provider = 'library',m.provider_instance,p.provider) as provider "
+            "SELECT p.item_id, p.media_type, p.name, p.image, p.provider "
             f"FROM {DB_TABLE_PLAYLOG} p "
-            f"LEFT JOIN {DB_TABLE_PROVIDER_MAPPINGS} m ON m.item_id = p.item_id "
-            "WHERE p.media_type in ('audiobook', 'podcast_episode') AND p.fully_played = 0 "
-            f"AND p.provider in {available_providers_str} "
+            "WHERE p.media_type IN ('audiobook', 'podcast_episode') "
+            "AND p.fully_played = 0 "
             "AND p.seconds_played > 0 "
         )
+        query += (
+            "AND ( "
+            "CASE WHEN p.provider = 'library' THEN "
+            f"EXISTS (SELECT 1 FROM {DB_TABLE_PROVIDER_MAPPINGS} m "
+            "WHERE m.item_id = p.item_id AND m.media_type = p.media_type "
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/music.py
```diff
@@ -667,9 +667,21 @@ async def in_progress_items(
...
         for db_row in db_rows:
             provider = db_row["provider"]
+
             # Apply user provider filter
-            if user_provider_filter and provider not in user_provider_filter:
-                continue
+            if user_provider_filter:
+                if provider == "library" and (
+                    # podcast episodes are never part of the library, this has to be an audiobook
+                    db_row_audiobook := await self.mass.music.database.get_row(
```
MarvinSchenkel: I think we can improve this by moving this check out of the loop and into the original query above? Would save a couple of db roundtrips.

fmunkes: Implemented a single DB query - but please triple check ;) sql queries are something I usually don't have to write.

Ok, this is not going to work, as item_id is not unique in the provider_mappings table, will have to filter by audiobook too.

MarvinSchenkel: Ok, let me walk through the problem we are actually fixing here so I completely understand what you are trying to do. As far as I understand it is this:

 `db_row["provider"]` current only returns "library" from the playlog, causing recommendations to be empty for users that have a user_provider_filter configured. Correct?

If that is the case, we should be able to simply move the filter inside the SQL query and then we can just blindly create an `ItemMapping` for every row we get back. Something like this:
```
SELECT p.item_id, p.media_type, p.name, p.image, p.provider
FROM playlog p
WHERE p.media_type IN ('audiobook', 'podcast_episode')
  AND p.fully_played = 0
  AND p.provider IN (<available_providers>)
  AND p.seconds_played > 0
  AND (
    CASE WHEN p.provider = 'library' THEN
      EXISTS (SELECT 1 FROM provider_mappings m
              WHERE m.item_id = p.item_id AND m.media_type = p.media_type
              AND m.provider_instance IN (<user_provider_filter>))
    ELSE p.provider IN (<user_provider_filter>)
    END
  )
ORDER BY timestamp DESC
```

If this is not the problem we are trying to solve, please explain it to me :)

fmunkes: Yes, you understood the problem correctly, and the SQL statement looks very much like what I need from just reading it. I'll give that a test tonight and will report back 🙂.

MarvinSchenkel: Cool, glad we're on the same page. It's always a good idea to let the database handle the filtering as much as we can to prevent unnecessarily dragging data into memory. :)

fmunkes: That sql query is perfect! It also made me spot/ fix another bug - in progress items listed, which were not part of MA anymore, because the provider was removed. That was reported every once in a while on discord, without me being able to figure out the issue - we now always check availability against provider mappings, if it is a library item as well.

MarvinSchenkel (approved):
Awesome, thanks @fmunkes  👏

PR #3325, opened 2026-03-07
Add password pairing support for Airplay 2 devices

hmonteiro:
This MR depends on https://github.com/music-assistant/cliairplay/pull/91

It adds support for password based authentication for airplay 2 devices.

Password based pairing in Airplay 2 turned out to be the same as the PIN based one, except that the pin is predefined. The changes required were related to bypassing the first stage of pin based pairing (getting the device to show the pin) and adding the ability to dynamically decide which pairing fields we require.


OzGav:
@bradkeifer

bradkeifer:
It looks good to me and is a welcomed improvement in capability. I would appreciate if @marcelveldt or @MarvinSchenkel could review the code to confirm compliance with standards/conventions please.

hmonteiro:
I have fixed a small issue with the reset sequence and added tests. It should be ready for review now, but I won't take the Draft tag until the MR on which it depends is merged and released to avoid any accidental merging.

MarvinSchenkel:
@bradkeifer Now that the cliairplay PR has been merged, I think we should also add the binaries to this PR right?

bradkeifer:
Yes, I agree. The appropriate version is available from the main branch of cliairplay repo. The binaries are built by github actions.

hmonteiro:
> Yes, I agree. The appropriate version is available from the main branch of cliairplay repo. The binaries are built by github actions.

I have updated the MR with the binaries from the latest main branch action: https://github.com/music-assistant/cliairplay/actions/runs/22935555709#artifacts

music_assistant/providers/airplay/player.py
```diff
@@ -419,6 +462,19 @@ def _get_pairing_config_entries(
                     category="protocol_generic",
                 )
             )
+        if protocol is StreamingProtocol.AIRPLAY2:
```
MarvinSchenkel: This is always `True` because of the loop above. Did you mean to indent this `if`?

music_assistant/providers/airplay/player.py
```diff
@@ -439,64 +496,101 @@ async def _handle_pairing_action(
...
+    ) -> None:
+        """Complete an in-progress pairing session.
 
+        ``values`` may contain a PIN or a password supplied by the user when required.
+        """
+        self.logger.debug(f"_finish_pairing for protocol: {protocol_name} with values: {values}")
+        if not values:
+            return
+
+        if self._requires_password_pairing():
```
MarvinSchenkel: For the config entries, you first check for pin pairing and secondly for password pairing. This could result in a race if the device requires both (not sure if this is a valid use case?). 

So let's say a device requires both, then the UI will only show a ConfigEntry for pin while this logic checks for a password.

We should be able to easily fix this by applying the same order as for the config entries. Something like 
```
if self._requires_pin_pairing():
    pin = values.get(CONF_PAIRING_PIN)
    ...
elif self._requires_password_pairing():
    pin = values.get(CONF_PAIRING_PASSWORD)
    ...
```

MarvinSchenkel (changes requested):
Few minor comments, but this is almost ready to merge 👍 . Please mark as 'Ready for review' again when you want us to have another look.

music_assistant/providers/airplay/player.py
hmonteiro: Good catch, this happened to work just because protocol would always be AIRPLAY2 due to the order of the loop.

music_assistant/providers/airplay/player.py
hmonteiro: I don't think it can require both, but it can't hurt to make the order consistent, so I will make that change to the order

music_assistant/providers/airplay/protocols/airplay2.py
```diff
@@ -55,6 +56,7 @@ def _cli_loglevel(self) -> int:
 
     async def start(self, start_ntp: int) -> None:
         """Start cliap2 process."""
+        self.logger.debug(f"Starting AirPlay2 stream with config: {self.player.config}")
```
MarvinSchenkel: Can we please remove this one?

MarvinSchenkel: Sorry I missed this before. It'll be good to merge after :).

hmonteiro: Done

MarvinSchenkel (approved):
Awesome, right on time for the new beta. Thanks @hmonteiro  🙏

PR #3326, opened 2026-03-07
Add support for OGG Vorbis radio streams

OzGav:
DEPENDS ON: https://github.com/music-assistant/models/pull/175

SATISFIES: https://github.com/orgs/music-assistant/discussions/5016

This PR adds support for OGG Vorbis and Opus radio streams with in-band metadata extraction. Internet radio stations using OGG containers send "chained" streams where each track is a separate logical bitstream with its own headers. FFmpeg's OGG demuxer cannot handle these chain transitions—it treats the end-of-stream marker followed by new headers as an error and exits.

The solution, inspired by VLC's OGG demuxer, is to parse the raw OGG pages and stitch multiple logical bitstreams into a single continuous stream. The first chain's headers (OpusHead/OpusTags or Vorbis identification/comment) are forwarded to FFmpeg, while subsequent chains have their headers stripped and EOS pages skipped entirely. Page sequence numbers and granule positions are rewritten to maintain continuity.

Metadata is extracted from OpusTags or Vorbis comment headers at chain boundaries and propagated via callback, enabling track title updates during playback without interrupting the audio stream.

tested with http://stream.tilos.hu:80/tilos.opus and http://srv.tsmod.net:8008/dnb


music_assistant/helpers/audio.py
```diff
@@ -899,6 +927,57 @@ async def get_icy_radio_stream(
...
+                    chunk_count,
+                    reconnect_count,
+                )
+                reconnect_count += 1
+                await asyncio.sleep(0.1)  # Brief delay before reconnect
+
+        except asyncio.CancelledError:
+            LOGGER.debug("Radio stream cancelled for %s", url)
+            raise
+        except ClientError as err:
```
marcelveldt-traveling: Do you want to be specific which errors we want to catch ?
Because a 404 should maybe just fail at once for example...

music_assistant/helpers/audio.py
```diff
@@ -1003,6 +1082,89 @@ async def _update_hls_radio_metadata(
         )
 
 
+async def _update_inband_radio_metadata(
```
marcelveldt-traveling: Am I reading this right that we are connecting twice to the same stream to get the metadata ?
That feels weird and unneeded. Why don't we read the data and extract audio+metadata in one go, such as the ICY stream ? Or would it not be possible ?

music_assistant/helpers/audio.py
OzGav: Yes. I have added more targetted error handling

music_assistant/helpers/audio.py
OzGav: You're right, the original implementation was wrong to open a second connection. The new code wires a metadata callback into get_chained_ogg_stream which already parses every Ogg page as it flows through to ffmpeg. When it encounters an OpusTags or Vorbis comment page, it extracts the metadata and updates streamdetails inline — same single-connection pattern as ICY. The polling callback and _update_inband_radio_metadata are removed.

marcelveldt-traveling (approved):
Looks good, well done!

MarvinSchenkel (approved):

PR #3327, opened 2026-03-07
Add genre exclusion feature to media items

jozefKruszynski:
This allows a user to exclude discovered genres from a mapped media item ensuring that it does not get remapped during background genre sync

### DIFF
--- a/music_assistant/controllers/media/genres.py
+++ b/music_assistant/controllers/media/genres.py
@@ -24,6 +24,7 @@ from music_assistant.constants import (
     DB_TABLE_ALBUMS,
     DB_TABLE_ARTISTS,
     DB_TABLE_AUDIOBOOKS,
+    DB_TABLE_GENRE_MEDIA_ITEM_EXCLUSION,
     DB_TABLE_GENRE_MEDIA_ITEM_MAPPING,
     DB_TABLE_GENRES,
     DB_TABLE_PLAYLISTS,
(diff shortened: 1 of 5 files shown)


music_assistant/controllers/music.py
```diff
@@ -2489,6 +2490,27 @@ async def _get_or_create_genre(
...
 
+        if prev_version <= 30:
+            # create the genre_media_item_exclusion table (new in schema 31)
+            await self._database.execute(
+                f"""
+                CREATE TABLE IF NOT EXISTS {DB_TABLE_GENRE_MEDIA_ITEM_EXCLUSION}(
+                [genre_id] INTEGER NOT NULL,
+                [media_id] INTEGER NOT NULL,
+                [media_type] TEXT NOT NULL,
+                FOREIGN KEY([genre_id]) REFERENCES [genres]([item_id]) ON DELETE CASCADE,
```
MarvinSchenkel: We don't have cascade delete enabled for the DB so you'd need to cleanup this table yourself.

music_assistant/controllers/media/genres.py
MarvinSchenkel, line 554: I think we need to delete the exclusions here as well?

MarvinSchenkel, line 929: We also need to delete from the genre_exclusion table here since cascade delete is not enabled on the db.

music_assistant/controllers/music.py
jozefKruszynski: Will do.

music_assistant/controllers/music.py
```diff
@@ -101,7 +102,7 @@
 DEFAULT_SYNC_INTERVAL = 12 * 60  # default sync interval in minutes
 CONF_SYNC_INTERVAL = "sync_interval"
 CONF_DELETED_PROVIDERS = "deleted_providers"
-DB_SCHEMA_VERSION: Final[int] = 30
+DB_SCHEMA_VERSION: Final[int] = 31
```
MarvinSchenkel: Let's bump this to 32 to correct the schema / migration logic offset. It now runs the migration logic more than once.

MarvinSchenkel (approved):
Looks good,  thanks @jozefKruszynski !

PR #3329, opened 2026-03-08
Downgrade DLNA SSDP discovery socket errors to a warning

balloob-travel:
## Summary
- catch `OSError` from DLNA SSDP discovery so the recurring background task logs a warning instead of dumping an unhandled task traceback
- add a regression test for the discovery socket failure path

## Notes
This error is probably rare and mostly experienced during development, for example when the computer loses Wi-Fi or goes to sleep while the server is running.

## Testing
- `uv run --extra test pytest tests/providers/dlna/test_provider.py`
- `uv run --extra test ruff check music_assistant/providers/dlna/provider.py tests/providers/dlna/__init__.py tests/providers/dlna/test_provider.py`


OzGav:
A couple of mypy errors on this one.

MarvinSchenkel (approved):
Thanks @balloob-travel 🙏

PR #3331, opened 2026-03-08
Make tracks sent to enqueue provider use correct position

apophisnow:
Bug fix for enqueue players. For example, when adding a song as next to Sonos it gets added an extra song back instead of being the very next song.


apophisnow:
Need to do a deeper dive and confirm the bug better before implementing this.

PR #3333, opened 2026-03-08
Add user awareness to podcast controller

fmunkes:
Adds user awareness to set_resume_position method in the podcast controller. This has an impact in multi-user setups which uses a single instance provider without own playlog tracking (e.g. itunes podcast). It doesn't hurt in all other cases, although provider mapping to a user is the preferred way then.

### DIFF
--- a/music_assistant/controllers/media/podcasts.py
+++ b/music_assistant/controllers/media/podcasts.py
@@ -223,20 +223,20 @@ class PodcastsController(MediaControllerBase[Podcast]):
 
         # Get user who initiated the query. Querying the userid as well is most useful
         # in a multi-user environment where a single instance provider is used.
-        podcast = await self.get(
-            item_id=item_id,
-            provider_instance_id_or_domain=provider_instance_id_or_domain,
-        )
-
         user: User | None = None
         if session_user := get_current_user():
             # this is the active session user that triggered the action
             user = session_user
-        elif provider_user := await self.mass.music._get_user_for_provider(
-            podcast.provider_mappings
-        ):
-            # based on configured provider filter we can try to find a user
-            user = provider_user
+        else:
+            podcast = await self.get(
+                item_id=item_id,
+                provider_instance_id_or_domain=provider_instance_id_or_domain,
+            )
+            if provider_user := await self.mass.music._get_user_for_provider(
+                podcast.provider_mappings
+            ):
+                # based on configured provider filter we can try to find a user
+                user = provider_user
 
         async def set_resume_position(episode: PodcastEpisode) -> None:
             if episode.fully_played is not None or episode.resume_position_ms:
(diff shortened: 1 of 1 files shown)


music_assistant/controllers/media/podcasts.py
```diff
@@ -219,19 +221,39 @@ async def _get_provider_podcast_episodes(
...
+        podcast = await self.get(
+            item_id=item_id,
+            provider_instance_id_or_domain=provider_instance_id_or_domain,
+        )
+
+        user: User | None = None
+        if session_user := get_current_user():
+            # this is the active session user that triggered the action
+            user = session_user
+        elif provider_user := await self.mass.music._get_user_for_provider(
```
MarvinSchenkel: We cannot access a private function outside of the class definition. We could consider making this a public function of the AuthController if it is needed outside of the music controller.

music_assistant/controllers/media/podcasts.py
```diff
@@ -219,19 +221,39 @@ async def _get_provider_podcast_episodes(
         if not isinstance(prov, MusicProvider):
             return
 
+        # Get user who initiated the query. Querying the userid as well is most useful
+        # in a multi-user environment where a single instance provider is used.
+        podcast = await self.get(
```
MarvinSchenkel: We could defer this call when we know that `get_current_user()` is `None`. We now always obtain the podcast object which in most cases is not necessary.

MarvinSchenkel (changes requested):
Few small improvements, but almost there :)

music_assistant/controllers/media/podcasts.py
fmunkes: See here https://github.com/music-assistant/server/pull/3253 for a similar PR targeting the audiobook controller, where Marcel agreed on the very last resolved copilot comment, that promoting it is not desired.

MarvinSchenkel: Okay, fair. Still think it's a bit strange as the expectation is that a private method is *never* called outside of a class definition. But okay, let's leave it in that case.

fmunkes: Yeah, I get your point - shall we add a "revisit" to the backlog? - maybe we can agree on some common naming scheme of core only, yet public methods or similar.

music_assistant/controllers/media/podcasts.py
fmunkes: done

music_assistant/controllers/media/podcasts.py
```diff
@@ -219,19 +221,39 @@ async def _get_provider_podcast_episodes(
...
             return
 
+        # Get user who initiated the query. Querying the userid as well is most useful
+        # in a multi-user environment where a single instance provider is used.
+        user: User | None = None
+        if session_user := get_current_user():
+            # this is the active session user that triggered the action
+            user = session_user
+        else:
+            podcast = await self.get(
```
MarvinSchenkel: Looking at this a bit more, can't we just pass `provider_instance_id_or_domain` to `_get_user_for_provider` and skip the entire `self.get`?

fmunkes: Uh, yeah, you are completely right - fixed.

MarvinSchenkel (approved):
Looks good to me , thanks:)

PR #3337, opened 2026-03-09
Add MPD Player Provider

OzGav:
Satisfies: https://github.com/orgs/music-assistant/discussions/1170

Each provider instance represents one MPD server, configured with host, port, and an optional password. MA delivers audio by having MPD fetch MA's HTTP stream URL; player state is received via MPD's idle mechanism and supplemented by periodic polling for elapsed time.
Flow mode is enforced and FLAC is excluded from codec options, as MPD cannot probe infinite HTTP streams. MP3 is the default output codec.
Tested against MPD running on Windows (via [Music Player Daemon for Windows](https://www.musicpd.org/download.html))

edit: I dont think the Trusted Sources failure in the security check should block this PR because as far as I can tell this is because the python-mpd2 project is just missing a line in their pyproject.toml. They have a bunch of entries which point to the repo except for "Source Code"


music_assistant/providers/mpd/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "player",
+  "domain": "mpd",
+  "name": "MPD Players",
+  "description": "Stream music from Music Assistant to any device running MPD (Music Player Daemon).",
+  "codeowners": ["@OzGav"],
+  "requirements": ["python-mpd2>=3.1.1"],
+  "documentation": "https://music-assistant.io/music-providers/mpd/",
+  "multi_instances": true
```
MarvinSchenkel: "multi_instance" *

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,311 @@
...
+        await self._disconnect()
+        await self._connect()
+
+    async def poll(self) -> None:
+        """Poll MPD for current state (elapsed time while playing)."""
+        await self._fetch_and_sync_state()
+
+    async def on_unload(self) -> None:
+        """Handle cleanup when the player is unloaded."""
+        await self._disconnect()
```
MarvinSchenkel: on_unload should not be overridden at player level. Instead, register an _on_unload_callbacks in the setup() to disconnect when the player gets unloaded

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,311 @@
...
+
+            status = await self._client.status()
+            self._attr_available = True
+            self._attr_device_info = DeviceInfo(
+                model=f"MPD {self._client.mpd_version}",
+                manufacturer="Music Player Daemon",
+            )
+            self._attr_device_info.add_identifier(IdentifierType.IP_ADDRESS, self.host)
+            self.logger.info("Connected to MPD at %s:%s", self.host, self.port)
+            self.mass.create_task(self._idle_loop())
```
MarvinSchenkel: You need to store this task on `self` so you can cancel it when `disconnect()` gets called on an unload to prevent this from hanging.

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,311 @@
...
+
+    async def _disconnect(self) -> None:
+        """Disconnect both MPD clients."""
+        for client in (self._client, self._idle_client):
+            if client:
+                client.disconnect()
+        self._client = None
+        self._idle_client = None
+
+    async def _reconnect(self) -> None:
```
MarvinSchenkel: Use mass.call_later icw the delay and a task id to automatically deduplicate reconnect tasks. See how Sonos does this as an example.

MarvinSchenkel (changes requested):
Few minor comments, almost there :)

music_assistant/providers/mpd/manifest.json
OzGav: fixed

music_assistant/providers/mpd/player.py
OzGav: done

music_assistant/providers/mpd/player.py
OzGav: Removed on_unload and registered _disconnect via _on_unload_callbacks in __init__ instead.

music_assistant/providers/mpd/player.py
OzGav: Done — _idle_task is now stored on self and cancelled in _disconnect().

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+
+# MPD receives a single continuous HTTP stream from MA, so flow mode must always be on.
+_CONF_ENTRY_FLOW_MODE_ENFORCED = ConfigEntry.from_dict(
+    {
+        **CONF_ENTRY_FLOW_MODE.to_dict(),
+        "default_value": True,
+        "value": True,
+        "hidden": True,
+    }
+)
```
marcelveldt: this is not the way to define this
you need to add a property to your player object "requires_flow_mode"

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+# Map MPD state strings to MA PlaybackState
+MPD_STATE_MAP: dict[str, PlaybackState] = {
+    "play": PlaybackState.PLAYING,
+    "pause": PlaybackState.PAUSED,
+    "stop": PlaybackState.IDLE,
+}
+
+
+class MPDPlayer(Player):
+    """Represents a single MPD server as a Music Assistant player.
```
marcelveldt:
```suggestion
    """
    Represents a single MPD server as a Music Assistant player.
```

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+        return ELAPSED_POLL_INTERVAL
+
+    async def get_config_entries(
+        self,
+        action: str | None = None,
+        values: dict[str, ConfigValueType] | None = None,
+    ) -> list[ConfigEntry]:
+        """Return player config entries."""
+        return [
+            _CONF_ENTRY_FLOW_MODE_ENFORCED,
```
marcelveldt:
```suggestion
```

music_assistant/providers/mpd/player.py
marcelveldt: flow mode is automaticaly enforced when enqueue is not defined in the supported features

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+        """Schedule a reconnect, deduplicating any pending reconnect tasks."""
+        task_id = f"mpd_reconnect_{self.player_id}"
+        self.mass.call_later(RECONNECT_DELAY, self._connect, task_id=task_id)
+
+    # ------------------------------------------------------------------
+    # Background idle loop
+    # ------------------------------------------------------------------
+
+    async def _idle_loop(self) -> None:
+        """Listen for MPD state changes and sync to MA.
```
marcelveldt:
```suggestion
        """
        Listen for MPD state changes and sync to MA.
```

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+
+        MPD's idle command blocks until a subsystem changes, then yields
+        the subsystem name. We act on player/mixer/playlist changes.
```
marcelveldt: implementation details do not belong in a docstring

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+            self._attr_available = False
+            self.update_state()
+            self.reconnect()
+
+    # ------------------------------------------------------------------
+    # State sync
+    # ------------------------------------------------------------------
+
+    async def _fetch_and_sync_state(self) -> None:
+        """Fetch current MPD status and update MA player state."""
```
marcelveldt, line 231:
```suggestion
        """Fetch current MPD status and update player state."""
```
MA is a bit redundant here as we are in the MA code

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+            return
+        try:
+            status = await self._client.status()
+            await self._sync_state(status)
+            self.update_state()
+        except MPDError as err:
+            self.logger.warning("Failed to fetch MPD status: %s", err)
+
+    async def _sync_state(self, status: dict[str, Any]) -> None:
+        """Map MPD status dict onto MA player attributes."""
```
marcelveldt:
```suggestion
        """Map MPD status dict onto player attributes."""
```

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,315 @@
...
+            self._attr_elapsed_time_last_updated = time.time()
+        else:
+            self._attr_elapsed_time = 0
+
+    # ------------------------------------------------------------------
+    # Player commands
+    # ------------------------------------------------------------------
+
+    async def play_media(self, media: PlayerMedia) -> None:
+        """Send play command with MA stream URL to MPD."""
```
marcelveldt:
```suggestion
        """Handle play media command on MPD Player."""
```

no implementation details in a docstring

music_assistant/providers/mpd/provider.py
```diff
@@ -0,0 +1,45 @@
...
+
+from music_assistant.constants import CONF_PASSWORD, CONF_PORT
+from music_assistant.models.player_provider import PlayerProvider
+
+from .constants import CONF_HOST
+from .player import MPDPlayer
+
+
+class MPDPlayerProvider(PlayerProvider):
+    """MPD Player provider.
```
marcelveldt:
```suggestion
    """
    MPD Player provider.
```

music_assistant/providers/mpd/provider.py
```diff
@@ -0,0 +1,45 @@
...
+
+    One provider instance represents one MPD server. The host, port and
```
marcelveldt: so it does not support discovery of players ?
what about a config entry to enter IP addresses and then have the optional username and password in the player config ? the same way as how airplay pairing works.

the one provider per player has always bothered me about the fully kiosk provider (and I really think we should change that) so if possible try to avoid that behavior.

music_assistant/providers/mpd/player.py
OzGav: All removed.

music_assistant/providers/mpd/player.py
OzGav: Removed

music_assistant/providers/mpd/player.py
OzGav: Fixed

music_assistant/providers/mpd/provider.py
OzGav: I have implemented this using a single comma-separated config entry, where each MPD server is entered in the format host, host:port, or host:port#password. Port defaults to 6600 if not specified.

Regarding the password — it authenticates the TCP connection to the MPD server, not the player. Every client connecting to that server uses the same password. Placing it in the player config would also create a chicken-and-egg problem: the player cannot be registered until the first connection succeeds, but the connection requires the password before the player config screen is even available. Keeping it inline with the host address avoids both issues.

OzGav: The other thing to note about the password is it is stored as plain text. Whilst not ideal my rationale for suggesting this is that the password is stored in MA’s config, which is a local file on the user’s own server. It’s not transmitted anywhere beyond the local TCP connection to MPD. The risk level is relatively low — anyone with access to MA’s config already has full access to MA itself.

marcelveldt: Can't you follow the same approach as AirPlay where the player is already added with requires setup?

So you add the host based on IP and port in provider setup, create a disconnected player

OzGav: Yeah ok I see it now. I wasn’t fully across how AirPlay was working. Let me rework this.

marcelveldt: Yeah, and for specifying the players, you may want to consider a multi value config entry (see the manual ips for sonos and cast for example)

by specifying the password in the player config, you can also store it as secure string (=encrypted)

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,316 @@
...
+        ],
+    }
+)
+
+# Map MPD state strings to MA PlaybackState
+MPD_STATE_MAP: dict[str, PlaybackState] = {
+    "play": PlaybackState.PLAYING,
+    "pause": PlaybackState.PAUSED,
+    "stop": PlaybackState.IDLE,
+}
```
marcelveldt: potential candidates for the constants.py

music_assistant/providers/mpd/player.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def __init__(
+        self,
+        provider: MPDPlayerProvider,
+        player_id: str,
+        host: str,
+        port: int = 6600,
+        password: str | None = None,
+    ) -> None:
+        """Initialize MPDPlayer.
```
marcelveldt:
```suggestion
        """
        Initialize MPDPlayer.
```

music_assistant/providers/mpd/player.py
OzGav: Moved

music_assistant/providers/mpd/provider.py
OzGav: OK this is now working as suggested. Player is registered. If password is required then it is marked as needing setup otherwise it is immediately available.

marcelveldt (commented):
Looks good, thanks @OzGav !

marcelveldt:
hmmm except for the dependency warning

OzGav:
I looked into it and added this to the PR description

> edit: I dont think the Trusted Sources failure in the security check should block this PR because as far as I can tell this is because the python-mpd2 project is just missing a line in their pyproject.toml. They have a bunch of entries which point to the repo except for "Source Code"

edit: I looked into it again with Claude and this is apparently the situation?

> This is very helpful. According to the PyPA well-known URLs spec, Repository is normalised to repository and rendered as "Source Code" by package indices like PyPI. Similarly, GitHub is rendered as "Source Code (GitHub)". 

> They already have a Repository line?
```
[project.urls]
Homepage = "https://github.com/Mic92/python-mpd2"
Repository = "https://github.com/Mic92/python-mpd2"
Issues = "https://github.com/Mic92/python-mpd2/issues"
```

> Well that's interesting — they already have Repository pointing to the correct URL, which should satisfy the check. So the GitHub security warning may simply be a false positive or a bug in the checker itself, since the package has all the right metadata.
In that case there's nothing to raise with the maintainer — the issue is on GitHub's side and not something we or they can fix. Worth noting that in the PR so Marcel understands it's not actionable.

OzGav:
OH hang on I see that new warning now... Let me look into that.

edit: False positive. Rebase fixed it.

MarvinSchenkel (approved):

PR #3340, opened 2026-03-09
Preserve Sendspin players queue across reconnects

maximmaxim345:
## Summary
Do not unregister offline Sendspin players to keep the queue in memory.

With this PR, Sendspin players now remain registered and are only marked as unavailable, preserving the queue.

## Description

The Sendspin provider was previously calling `unregister` whenever a player/client went offline.
This had one major drawback: unregistering a player also deleted that player's queue.
While merged players created a Universal player, which solved this issue for most clients (since they have the type `PROTOCOL`), Sendspin-only clients are of type `PLAYER`.
This means that the clients most susceptible to network interruptions are also the ones whose queue gets cleared most often: the Web Player and Mobile Apps.

## Alternative Solutions

Alternatively, we could keep Sendspin using `unregister` for offline players and instead preserve queues after calling `PlayerController.unregister(permanent=False)`.
This would also fix the issue for other providers that unregister players, but it might introduce unexpected issues with dangling queues.

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -1373,6 +1373,10 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
         if player.player_id in self._players:
             self._players[player.player_id] = player
             player.update_state()
+            # Re-evaluate protocol links on protocol player updates so parent wrappers
+            # can refresh metadata (e.g. reconnect with updated device info).
+            if player.state.type == PlayerType.PROTOCOL:
+                self._evaluate_protocol_links(player)
             # Also schedule update when replacing existing player
             self._schedule_update_all_players()
             return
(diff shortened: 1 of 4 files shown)


marcelveldt-traveling:
There seems to be a number of unrelated commits in the PR - was that intentional ?
If so, please update the PR description to match what was changed. 

I have these 2 comments:

1) in general, a Queue should always be restored when a player reconnects again, this is done in the on_player_register callback from Queue controller, where it restores the queue from cache. If that is not the case, then we have some other challenge to worry about or we accidentally permanently remove the player from the player controller ?

2) By only setting the available flag, it keeps a player forever loaded in the player controller, even if a player is indeed unavailable for a longer period of time. Would it make sense to schedule a unregister after e.g. 120 seconds ? So player goes offline --> mark unavailable --> it returns within 120 seconds --> mark available again. If it doesn't return within 120 seconds, unregister from player controller.

MarvinSchenkel:
Related issue: https://github.com/music-assistant/support/issues/5049

maximmaxim345:
Closing since this PR was superseded by #3352 and #3431.

PR #3347, opened 2026-03-09
Validate player control config references before using them

teancom:
Myself and at least one other person have seen that, post-universal player consolidation, our WiiM devices always show 'zero' as the volume, and any changes aren't reflected. I believe that's not anything WiiM-specific, but that we previously had DNLA/Chromecast/whatever entries that aren't used anymore, but *are* being referenced on start. This fixes that, at least for me locally (I've been testing it all day).

The volume_control, mute_control, and power_control properties trusted config values without verifying the referenced player or control still exists. After the universal player consolidation, stale config entries (e.g. old Chromecast player IDs) caused volume commands to silently fail.

Now validates that config IDs resolve to known constants or existing players/controls before using them, falling through to auto-detection otherwise.

### DIFF
--- a/music_assistant/models/player.py
+++ b/music_assistant/models/player.py
@@ -756,30 +756,37 @@ class Player(ABC):
         return cast("str", self._config.get_value(CONF_ENTRY_PLAYER_ICON.key))
 
     def _resolve_control_config(self, conf_key: str) -> str | None:
-        """Validate a control config value references an existing player or control.
+        """Return the explicitly configured control value, or None if not set.
 
-        Returns the config value if it's a known constant or resolves to an existing
-        player/control, or None if stale (so the caller can fall through to auto-detection).
+        If a value is explicitly configured (raw value is not None), it is always returned
+        — even if it doesn't currently resolve to a registered player or control. This
+        handles the case where protocol players register asynchronously and the configured
+        player will appear shortly after startup.
+
+        Returns None only when no value has been explicitly configured, signaling the
+        caller to fall through to auto-detection.
 
         :param conf_key: The config key to look up (e.g. CONF_VOLUME_CONTROL).
         """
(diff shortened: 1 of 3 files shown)


teancom:
One thing I didn't do in this PR is automatically clean up the old stale entries. I'm a little worried about anything along those lines possibly removing configuration for things that just haven't been initialized yet. If we want, I can look into that in a follow-up PR or add it to this one if we want to do it all at once. Or we can leave it as is and let people clean up their configs manually.

music_assistant/models/player.py
```diff
@@ -755,12 +755,38 @@ def icon(self) -> str:
...
+        if conf := self.mass.config.get_raw_player_config_value(self.player_id, conf_key):
+            conf_str = str(conf)
+            if conf_str in (PLAYER_CONTROL_NATIVE, PLAYER_CONTROL_FAKE, PLAYER_CONTROL_NONE):
+                return conf_str
+            if self.mass.players.get_player(conf_str) or self.mass.players.get_player_control(
+                conf_str
+            ):
+                return conf_str
+            # configured ID doesn't resolve to any known player or control
+            self.logger.warning(
```
marcelveldt-traveling: There is a high chance that this will now be logged every time at startup due to the fact that players and controls get added asynchronously - this might be confusing to the user. I would use debug log level here.

On the other hand, if the value is explicitly set by a user, it might be strange to just fallback to something else. We may want to consider only doing a fallback if the raw value is actually None, which means its not set by the user.

self.config.get_value --> contains calculated value 
self.mass.config.get_raw_player_config_value --> contains the actually stored value

If the raw value is None, it means it has not been explicitly set

teancom: That leads to the original issue (pointing to non-existent controls). But there may be a way to reconcile at the end. I'm gonna go have a hard think about it.

marcelveldt:
superseded by https://github.com/music-assistant/server/pull/3503

PR #3349, opened 2026-03-10
Add derived genres for local and smb providers

jozefKruszynski:
Adds the option to derive genre from tagged tracks for artists and albums for the local and smb file share provider.

Is dependant on PR before being merged.
https://github.com/music-assistant/server/pull/3327


music_assistant/controllers/music.py
MarvinSchenkel: We now need to bump this to 33 and create a separate <= 32 migration

music_assistant/controllers/media/genres.py
```diff
@@ -919,10 +923,92 @@ async def _bulk_scan_unmapped_genres(self) -> int:
...
+        enabled_instance_ids: list[str] = []
+        for p in self.mass.music.providers:
+            if p.domain in {"filesystem_local", "filesystem_smb"}:
+                enabled = await self.mass.config.get_provider_config_value(
+                    p.instance_id, "propagate_track_genres", default=False
+                )
+                if enabled:
+                    enabled_instance_ids.append(p.instance_id)
+        if not enabled_instance_ids:
+            return
```
MarvinSchenkel: Should we add some logic here that also deletes the existing derived genres? Now I can have this situation:
- Enable propagation.. genres are added
- Disable propagation -> propagated genres remain active

music_assistant/controllers/media/genres.py
```diff
@@ -919,10 +923,92 @@ async def _bulk_scan_unmapped_genres(self) -> int:
             )
             await db.execute(full_query)
         await db.commit()
+        await self._propagate_genre_mappings_to_parents()
         count_after = await db.get_count(gm)
 
         return count_after - count_before
 
+    async def _propagate_genre_mappings_to_parents(self) -> None:
```
MarvinSchenkel, line 932: Let's add some tests for this method

music_assistant/controllers/media/genres.py
```diff
@@ -1020,10 +1110,20 @@ async def remove_media_mapping(
...
         :param media_type: Type of media item (track, album, artist).
         :param media_id: Database ID of the media item.
         """
+        row = await self.mass.music.database.get_row(
+            DB_TABLE_GENRE_MEDIA_ITEM_MAPPING,
+            {"genre_id": int(genre_id), "media_id": int(media_id), "media_type": media_type.value},
+        )
+        if row and row["is_derived"]:
+            await self.exclude_genre_from_media_item(genre_id, media_type, media_id)
+            return
```
MarvinSchenkel: I think we do not need the return here? Otherwise we do not delete the derived mapping

MarvinSchenkel (approved):
👏 👏 👏 👏

PR #3352, opened 2026-03-10
Fix queue state loss on player re-register

maximmaxim345:
Fixes queues getting lost for players that get re-registered.
Particularly fixes the Sendspin Web Player loosing its queue.


maximmaxim345:
Partially superseded #3340

MarvinSchenkel (approved):
Nice, thanks @maximmaxim345 🙏

PR #3357, opened 2026-03-11
Fix db migration for second run of supported_mediatypes

fmunkes:
I copied the approach of other db alteration by catching any exception, and looking for a duplicate in the string.


MarvinSchenkel (approved):
Thanks @fmunkes 👍

PR #3358, opened 2026-03-11
Handle HEOS internal queue timeouts

Tommatheussen:
In previous PRs, we made sure the internal HEOS queue gets cleared when starting a new stream from MA.

It seems in some cases, the `get_queue` command takes longer to respond than the default 15 seconds timeout. 
Catching this exception, skipping the queue clearing and continuing to send a new stream to the device, makes sure the devices still are able to play.
I do not know yet how the devices get into this bad state where `get_queue` commands are delayed. But I had it happen on 1 of my devices, where I tested this fix with.

Additionally, I've added a timeout configuration to the provider. This could reduce these errors if they occur very often. When running a full HEOS network wirelessly, it might increase response times on all commands, allow to tweak this can help reduce these issues.
I also fixed the config entry for the manually configured IP, as it wasn't showing up anymore.

### DIFF
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -276,9 +276,24 @@ class HeosPlayer(Player):
 
     async def play_media(self, media: PlayerMedia) -> None:
         """Handle PLAY MEDIA command on given player."""
-        queue_items = await self._device.get_queue(range_start=0, range_end=1)
-        if queue_items:
-            await self._device.clear_queue()
+        self.logger.debug(
+            "[%s] Received PLAY_MEDIA command with media_type=%s uri=%s",
+            self._device.name,
+            media.media_type,
+            media.uri,
+        )
+
+        try:
+            queue_items = await self._device.get_queue(range_start=0, range_end=1)
+        except Exception as err:
+            self.logger.warning(
+                "[%s] Failed to fetch HEOS queue before play_media, continuing: %s",
+                self._device.name,
+                err,
+            )
+        else:
+            if queue_items:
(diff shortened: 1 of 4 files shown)


music_assistant/providers/heos/player.py
```diff
@@ -276,9 +276,24 @@ async def previous_track(self) -> None:
...
+        self.logger.debug(
+            "[%s] Received PLAY_MEDIA command with media_type=%s uri=%s",
+            self._device.name,
+            media.media_type,
+            media.uri,
+        )
+
+        try:
+            queue_items = await self._device.get_queue(range_start=0, range_end=1)
+        except Exception as err:
```
MarvinSchenkel: Should we narrow this to `HeosError` ?

music_assistant/providers/heos/player.py
```diff
@@ -276,9 +276,24 @@ async def previous_track(self) -> None:
...
+            self.logger.warning(
+                "[%s] Failed to fetch HEOS queue before play_media, continuing: %s",
+                self._device.name,
+                err,
+            )
+        else:
+            if queue_items:
+                await self._device.clear_queue()
```
MarvinSchenkel: Just checking , shouldn't this be part of the `try` too? or won't this one throw a timeout error?

Tommatheussen: Will add it to the try part, haven't seen it happen but it's safer if it does happen

MarvinSchenkel (approved):
Looks good, thanks @Tommatheussen

PR #3359, opened 2026-03-11
Fix possible race conditions during HEOS startup

Tommatheussen:
While working on https://github.com/music-assistant/server/pull/3358, I noticed a race condition happening from time to time:

HEOS goes through this setup:
- 1 mDNS entry start controller setup
- Controller setup will grab system information (including players)
- As HEOS prefers using wired hosts, it's possible a re-connection needs to happen to the preferred host
- In the meantime, additional mDNS reports might trigger HEOS player discovery.

This results in 3 scenarios:
- controller setup is fully done before player discovery, no problem here
- Player discovery happens before the re-connection, everything works fine, there's just a minor blip in connectivity, but not noticable as it's still being setup
- Player discovery happens when the re-connection checks happen. The player discovery command fails here, sometimes being able to recover because the controller setup specifically calls the player discovery method. But sometimes failing totally due to command conflicts and timeouts.

This fix makes sure we have finished the controller setup (also the re-connection) before trying to discover any players.
Manual IP setup force player discovery after setup, discovery setup already runs player discovery after mDNS entry is handled (by MA itself)

### DIFF
--- a/music_assistant/providers/heos/provider.py
+++ b/music_assistant/providers/heos/provider.py
@@ -43,6 +43,9 @@ class HeosPlayerProvider(PlayerProvider):
             ip_address = cast("str", ip_address)
             await self._setup_controller(ip_address)
 
+            # Explicitly discover players now
+            await self.discover_players()
+
     async def _setup_controller(self, controller_ip: str, connect_preferred: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
LGTM :)

PR #3360, opened 2026-03-12
Add min and max volume functionality per player

OzGav:
FRONTEND CHANGES: https://github.com/music-assistant/frontend/pull/1569
Satisifies: https://github.com/orgs/music-assistant/discussions/3288

Add configurable min/max volume limits per player

### Summary
- Adds per-player minimum and maximum volume configuration options (advanced settings under player controls)
- Volume commands (volume_set, volume_up, volume_down) are scaled from the UI’s 0-100 range to the configured min-max range before being sent to the device
- External volume changes (physical buttons, other apps) are reverse-scaled back to 0-100 for reporting, so the UI always displays a consistent 0-100 range
- Group volume adjustments work correctly since scaling is applied per child player
- Config validation rejects min > max to prevent invalid states
- Works with all API clients (frontend, Home Assistant, third-party) without any client-side changes

### Edge cases and functionality covered
**cmd_volume_set** — The 0-100 input is linearly mapped to `[min_volume, max_volume]` via `min_volume + (volume_level * (max_volume - min_volume)) // 100` before reaching the device. Applies to all programmatic and API invocations.

**cmd_volume_up / cmd_volume_down** — Step-based changes still operate in the 0-100 UI range (step sizes of 1-3 depending on position). Scaling happens downstream in _handle_cmd_volume_set, so no special casing is needed here.

**External volume changes** — When a device reports a volume change (e.g. user pressed a physical volume button, or used a manufacturer app), signal_player_state_update reverse-scales the reported value back to the 0-100 range before it reaches the UI. This keeps the slider position consistent with user expectations regardless of how the volume was changed.

**Group volume** — Each child player is scaled independently using its own min_volume/max_volume, so a group volume of 80 translates correctly for each member (e.g. a soundbar with max=50 receives 40, a regular speaker with max=100 receives 80).

**Config validation** — on_player_config_change validates that min_volume <= max_volume and raises InvalidDataError if not.

**Falsy value handling** — Uses explicit is not None checks rather than truthiness when reading config values, so min_volume=0 (a valid and common default) isn’t incorrectly treated as missing.

**Advanced settings** — Both config entries are marked as advanced so they don’t clutter the basic player settings for users who don’t need them.

**Default behavior unchanged** — With defaults of min=0 and max=100, the scaling formula is the identity function, so all existing behavior is preserved.


oleost:
Thanks for trying to fix https://github.com/orgs/music-assistant/discussions/3288

Was wondering, would it be possible to implement also a option to scale the volume ?

Something like this:
https://github.com/oleost/server/tree/dev-max-volum-scale


image

image

oleost:
Readjusted the mentioned pull request above to only scale volume.

marcelveldt:
1) Agreed on scaling the volume - that is imo the only way to implement this properly.
2) Do we really need min volume ? Isn't this always about max volume ?

oleost:
> 1. Agreed on scaling the volume - that is imo the only way to implement this properly.
> 
>     2. Do we really need min volume ? Isn't this always about max volume ?

For me personally I don't need min volume, however was someone that requested it, and cant see what it hurts implementing it at the same time. Mute still works.

music_assistant/controllers/players/controller.py
```diff
@@ -1577,6 +1582,97 @@ def delete_player_config(self, player_id: str) -> None:
...
+        min_volume, max_volume = self._get_volume_limits(player_id)
+        if min_volume == 0 and max_volume == 100:
+            return
+        device_volume = player.volume_level
+        clamped = max(min_volume, min(max_volume, device_volume))
+        if clamped != device_volume:
+            # Device volume is outside allowed range, correct it
+            self.mass.create_task(player.volume_set(clamped))
+
+    def _forward_state_updates_to_related_players(
```
marcelveldt: where is this coming from ? this is an existing method in the controller?

marcelveldt: also its outdated!

music_assistant/controllers/players/controller.py
```diff
@@ -3097,24 +3184,25 @@ async def _handle_cmd_volume_set(self, player_id: str, volume_level: int) -> Non
...
         # Only fire if this player is the direct owner of the plugin source,
         # not when it merely inherits active_source from a parent group —
         # group volume changes handle the callback once at the group level.
         if plugin_source := self._get_active_plugin_source(player):
             if plugin_source.on_volume and plugin_source.in_use_by == player.player_id:
                 await plugin_source.on_volume(volume_level)
+
         # Handle native volume control support
         if player.volume_control == PLAYER_CONTROL_NATIVE:
-            # player supports volume command natively: forward to player
```
marcelveldt, line 3202: why are existing comments removed ?

music_assistant/controllers/players/controller.py
```diff
@@ -3097,24 +3184,25 @@ async def _handle_cmd_volume_set(self, player_id: str, volume_level: int) -> Non
...
-            await player.volume_set(volume_level)
+            await player.volume_set(device_volume)
             return
         # Handle fake volume control support
         if player.volume_control == PLAYER_CONTROL_FAKE:
-            # user wants to use fake volume control - so we (optimistically) update the state
```
marcelveldt, line 3207: why are existing comments removed?

music_assistant/constants.py
```diff
@@ -271,6 +273,30 @@ def load_genre_mapping() -> list[dict[str, Any]]:
...
+CONF_ENTRY_MIN_VOLUME = ConfigEntry(
+    key=CONF_MIN_VOLUME,
+    type=ConfigEntryType.INTEGER,
+    range=(0, 100),
+    default_value=0,
+    label="Minimum volume",
+    description="Minimum device volume. "
+    "The volume slider (0-100) will be scaled to this as the lower bound.",
+    category="player_controls",
+    advanced=True,
```
marcelveldt, line 292:
```suggestion
    advanced=True,
    depends_on=CONF_VOLUME_CONTROL,
    depends_on_value_not=VOLUME_CONTROL_NONE
```

music_assistant/constants.py
```diff
@@ -271,6 +273,30 @@ def load_genre_mapping() -> list[dict[str, Any]]:
...
+CONF_ENTRY_MAX_VOLUME = ConfigEntry(
+    key=CONF_MAX_VOLUME,
+    type=ConfigEntryType.INTEGER,
+    range=(0, 100),
+    default_value=100,
+    label="Maximum volume",
+    description="Maximum device volume. "
+    "The volume slider (0-100) will be scaled to this as the upper bound.",
+    category="player_controls",
+    advanced=True,
```
marcelveldt, line 306:
```suggestion
    advanced=True,
    depends_on=CONF_VOLUME_CONTROL,
    depends_on_value_not=VOLUME_CONTROL_NONE
```

music_assistant/controllers/players/controller.py
OzGav: Got carried away with trying to simplify the comments and docstrings. Comments re-added.

music_assistant/controllers/players/controller.py
OzGav: As above

music_assistant/controllers/players/controller.py
OzGav: Yeah the updates outside this PR have made this redundant so it has been removed

marcelveldt (approved):
Nice work, thanks @OzGav !

MicheleLa:
Nice Work!
Would it be possible to add per player "Default Volume" ?
The idea behind this request is that so when you group multiple speakers while playing , you will have prediactable behaviour for devices with powerful output like AVR receiver . :-)

OzGav:
You will need to open a new feature request for that

oleost:
@OzGav Tried latest dev, and for me this volume limit does not work.
First time I change the max volume, it reflects in the UI. example if volume is 50 and I change max volume to 25, the UI now reports 240 in volume. But as soon as I change volume it completly ignored and still regular 0-100, where 100 is 100 and not actually clamped down to 25.

This is on a Q-series soundbar running Chromecast protocol.

Edit: This is what claude said: 

> **Note: the following analysis was done with AI (Claude)**
>
> The bug is in `_handle_cmd_volume_set`. The PR introduces `device_volume = self.scale_volume_to_device(...)` but only uses it in the `PLAYER_CONTROL_NATIVE` branch. The **protocol player redirect** — which is how Chromecast volume control works — was left unchanged and still passes the unscaled `volume_level`:
>
> ```python
> if protocol_player := self.get_player(player.state.volume_control):
>     await self._handle_cmd_volume_set(protocol_player.player_id, volume_level)  # should be device_volume
> ```
>
> The fix is to overwrite `volume_level` in-place at the top of the function (before any redirect logic), so all code paths — including the protocol player redirect — automatically use the scaled value. That way no redirect branch needs to be individually updated.

oleost:
@OzGav Did you ever look into my previous comment?

edit, seems like someone else did fix it in 4461

OzGav:
You are commenting on a closed PR which is why I am not seeing these. Please open a discussion Q&A

PR #3362, opened 2026-03-12
Guard against trailing spaces in auth URLs

OzGav:
When users configure their Home Assistant URL with trailing whitespace (e.g. "https://ha.example.com "), the OAuth authorization URL becomes invalid (e.g. "https://ha.example.com /auth/authorize?..."). Strip whitespace from HA URLs at all points where they feed into auth URLs.


MarvinSchenkel (approved):
Nice catch!

PR #3368, opened 2026-03-12
Fix _transitioning_players leak causing frozen queue state

chrisuthe:
Issue: skip to a track that can't be played (expired stream URL, unavailable provider, network hiccup during the fetch) — or anything else goes wrong during the transition between tracks. The UI could freeze on whatever state it was in. If it showed "Playing," it keeps showing "Playing" forever — even though nothing is coming out of the speakers. If you press pause, the server pauses the player internally, but the queue controller never finds out. Press play again, same thing — the actual player state changes but the queue display never updates. The seekbar stops advancing or shows a stale position.

play_index() adds the queue_id to _transitioning_players but only removes it on the success path. If any exception is raised (media not found, player unavailable, stream error), the queue_id remains in the set permanently. While stuck, on_player_update() silently ignores all state updates from that player, leaving the queue frozen with a stale state that never reflects the actual player status.

Additionally, next() and previous() add to the set before scheduling play_index via call_later, but have early return paths where play_index is never scheduled. Methods that cancel the debounced play_index timer (stop, pause, play_media) also leave the flag set since the scheduled cleanup in play_index never executes.

Changes:
- Wrap play_index body in try/finally to guarantee discard on any exit
- Add discard before early returns in next() and previous()
- Add discard after cancel_timer calls in stop(), pause(), play_media(), and on_player_remove() to clean up when the debounced play_index is cancelled

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -490,6 +490,7 @@ class PlayerQueuesController(CoreController):
         """
         # cancel any pending play_index calls for this queue to prevent conflicts
         self.mass.cancel_timer(f"queue_play_index_{queue_id}")
+        self._transitioning_players.discard(queue_id)
         # ruff: noqa: PLR0915
         # we use a contextvar to bypass the throttler for this asyncio task/context
         # this makes sure that playback has priority over other requests that may be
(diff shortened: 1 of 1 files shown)


marcelveldt-traveling (approved):
Nice catch+fix!

MarvinSchenkel (approved):
Thanks @chrisuthe 🙏

PR #3370, opened 2026-03-12
Fix: don't reassign the metadata callback for HLS streams, if already set

kieranhogg:


MarvinSchenkel (approved):
Thanks @kieranhogg !

kieranhogg:
@MarvinSchenkel is there any way this might have evaded 2.8.0? I'm seeing the original buggy behaviour in 2.8.0, but not dev.

OzGav:
Check the code base is it still there?

kieranhogg:
Hm, it looks like it, I'll keep digging.

MarvinSchenkel:
Yeah this should be in 2.8

PR #3180, opened 2026-02-17
Add Dashie Kiosk player provider

jwlerch78:
## Summary

Adds a new **player provider** for [Dashie Kiosk](https://dashieapp.com/dashie-kiosk) Android tablets. This provider communicates directly with the tablets via their REST API for audio playback control.

- **Direct audio control**: play, pause, stop, seek, and volume via HTTP REST API
- **HA entity discovery**: auto-discovers Dashie devices from the Home Assistant entity registry (filtered by the `dashie` integration platform), presented as a multi-select dropdown in the provider config
- **Offline device retry**: devices that are unreachable at startup are automatically retried every 60 seconds in the background
- **Device polling**: polls each tablet every 10 seconds for playback state and position updates
- **Flow mode**: uses flow mode for gapless audio streaming

### Requirements

- Home Assistant Plugin (must be configured first)
- [Dashie HA Integration](https://github.com/jwlerch78/dashie-ha-integration) installed in Home Assistant (provides the `dashie` platform entities and device `configuration_url` used for direct connection)

### Files

| File | Description |
|------|-------------|
| `__init__.py` | Provider setup and HA entity dropdown config |
| `provider.py` | Device discovery, lifecycle, and retry logic |
| `player.py` | Player implementation (playback, volume, polling) |
| `client.py` | Async HTTP client for the Dashie Kiosk REST API |
| `constants.py` | Shared constants (`CONF_PLAYERS`, `DASHIE_HA_DOMAIN`, `RETRY_INTERVAL`) |
| `manifest.json` | Provider metadata (`depends_on: hass`, `stage: beta`) |
| `icon.svg` | Dashie logo (color) |
| `icon_monochrome.svg` | Dashie logo (monochrome, vector-traced) |

### How it works

1. User selects Dashie media player entities from a dropdown (filtered to only show entities from the `dashie` HA integration)
2. Provider looks up each entity's device in the HA device registry and extracts the `configuration_url` (e.g. `http://192.168.86.30:2323`)
3. Creates a direct REST API client for each device
4. Registers each device as a Music Assistant player with flow mode enabled
5. Polls devices for state (volume, playback position) and sends commands directly

## Test plan

- [x] Tested with 5 physical Dashie Kiosk devices (Fire tablet, Samsung SM-X200, Echo Show 5 with Lineage OS, Mio 15" kiosk, ONN streaming stick)
- [x] Verified all 5 devices register and connect at startup
[...]

PR #3371, opened 2026-03-12
Fix Dashie Kiosk manual_players config validation error

jwlerch78:
## Summary

Fixes a config validation error when saving the Dashie Kiosk provider settings. The `manual_players` field used `multi_value=True` without predefined options, causing the frontend to send a plain string instead of a list. The backend then rejected the value with:

```
value for manual_players must be a list
```

**Changes:**
- `__init__.py`: Remove `multi_value=True` from the `manual_players` config entry. Users now enter comma-separated IP:port addresses in a plain string field.
- `provider.py`: Parse the comma-separated string into individual addresses.

The HA players multi-select field (which has predefined options) is unchanged and works correctly.

## Test plan

- [ ] Add Dashie Kiosk provider with only HA players selected — should save without error
- [ ] Add Dashie Kiosk provider with manual addresses (e.g. `192.168.1.100:2323`) — should save and connect
- [ ] Add multiple manual addresses comma-separated — should connect to all

Follows up on #3180.

### DIFF
--- a/music_assistant/providers/dashie_kiosk/__init__.py
+++ b/music_assistant/providers/dashie_kiosk/__init__.py
@@ -79,12 +79,11 @@ async def get_config_entries(
         ConfigEntry(
             key=CONF_MANUAL_PLAYERS,
             type=ConfigEntryType.STRING,
-            multi_value=True,
             label="Manual Dashie Kiosk addresses",
             required=False,
-            description="Manually add Dashie Kiosk tablets by IP address and port "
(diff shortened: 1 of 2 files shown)


OzGav:
@jwlerch78 while I have you can you confirm that Dashie has the same functionality as Fully Kiosk per the linked table?

MarvinSchenkel:
Closing this one as these changes are in your followup PR as well

PR #3372, opened 2026-03-12
Narrow down genre linking during scan

jozefKruszynski:
Media items link to fewer genres, due to the scan now returning early after finding match. Previously we still continued to look for the genre alias in all other genres even when finding an exact match.


MarvinSchenkel (approved):
Thanks @jozefKruszynski

PR #3374, opened 2026-03-12
Use internal time in socket based progress updates in Audiobookshelf.

fmunkes:
Fixes an edge case, where the abs server and ma server are not synced in time. The progress update via socket then occurs a bit unpredictably. Solved by using the internal time.

Also, abs updates every 10s now, so I adjusted the minimum time required to pass between updates.

### DIFF
--- a/music_assistant/providers/audiobookshelf/helpers.py
+++ b/music_assistant/providers/audiobookshelf/helpers.py
@@ -96,15 +96,7 @@ class ProgressGuard:
         mass external updates. Here, we compare this property against a potential
         stored one.
         """
-        item_id = abs_progress.library_item_id
-        episode_id = abs_progress.episode_id
-        stored_progress = self._get_progress(item_id=item_id, episode_id=episode_id)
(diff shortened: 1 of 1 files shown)


music_assistant/providers/audiobookshelf/helpers.py
MarvinSchenkel, line 92: Isn't this now functionally equal to `guard_ok_mass`?

MarvinSchenkel: You could just `return guard_ok_mass(item_id, episode_id)` I think

MarvinSchenkel (changes requested):
1 minor nit :)

music_assistant/providers/audiobookshelf/helpers.py
fmunkes: You're right. Did exactly that :)

MarvinSchenkel (approved):
Thanks @fmunkes 🙏

PR #3376, opened 2026-03-13
Fix DLNA volume feedback, poll state updates, and RC polling

sfortis:
Fixes several DLNA/UPnP player issues observed on devices like JBL Authentics:

**Volume feedback delay**: After volume_set() or volume_mute(), internal state waited up to 30s for the next poll cycle. Now explicitly polls RenderingControl and pushes state immediately.

**Poll not pushing state**: poll() called async_update() but never _update_player(), so updated state was not pushed to WebSocket clients. Added the missing call.

**Explicit RenderingControl polling**: Some DLNA devices (e.g. JBL Authentics) do not send reliable LastChange volume events even with active subscriptions. Added _poll_rendering_control() helper that directly queries GetVolume/GetMute actions.

### DIFF
--- a/music_assistant/providers/dlna/player.py
+++ b/music_assistant/providers/dlna/player.py
@@ -96,6 +96,7 @@ class DLNAPlayer(Player):
         self.bootid: int | None = None
         self.last_seen = time.time()
         self.last_command = time.time()
+        self._stale_resume_pending = False  # guard against repeated resume attempts
 
     def set_available(self, available: bool) -> None:
         """Set the availability of the player."""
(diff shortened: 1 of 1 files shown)


music_assistant/providers/dlna/player.py
```diff
@@ -382,6 +383,30 @@ def _get_playback_state(self) -> PlaybackState | None:
             TransportState.PLAYING,
             TransportState.TRANSITIONING,
         ):
+            # Detect stale PLAYING state after server restart:
```
MarvinSchenkel: I don't think detecting stale streams is something we should implement on provider level. I would leave this code out of this PR. If you want to implement this somewhere I think the player_controller would be the place to do this, but this should be a separate PR so we can move the discussion there.

sfortis: Done, removed the stale detection code from this PR. I'll submit a separate PR implementing it at the player_queues controller level where it can cover all providers.

MarvinSchenkel:
I am so sorry, I totally missed this PR and accidentally fixed this issue myself in [this one](https://github.com/music-assistant/server/pull/3474) when I bumped into it. Thanks!

PR #3377, opened 2026-03-13
Extract date_added from Deezer API for library items

### LINKED ISSUE #5094: [Feature Request] Sorting by 'Recently Added' broken for Deezer/Spotify (timestamp_added always 0)

2.7.10

Docker Container

- [ ] If using HA, I have confirmed the internal URL is set correctly (N/A, standalone Docker)
- [ ] If the problem relates to a device then I have checked the device settings (N/A, not device related)
- [ ] For voice problems, I have sought help elsewhere before returning here (N/A)
- [ ] For playback problems, I have recycled power to the physical device (N/A, not a playback issue)

### The problem

Sorting favorite tracks by "Recently Added" (or `timestamp_added`) does not work correctly for providers that do not supply a `date_added` field. All tracks end up with `timestamp_added = 0` in the database, so sorting by date added silently falls back to alphabetical order.

**Affected providers:** Deezer, Spotify, Jellyfin, and others that do not set `date_added` during parsing.

**Working providers:** Tidal (uses `created` field), filesystem_local (uses file modification time).

**API evidence (Deezer, 28 favorite tracks):**

```
GET music/tracks/library_items {favorite: true, limit: 30}
```

Response returns tracks in alphabetical order (Achilles, Appoloduction, A Song for You, Assez...) with no `date_added` or `timestamp_added` field present.

The sorting order is identical regardless of whether `order_by: "timestamp_added"` or `order_by: "timestamp_added_desc"` is specified, confirming all values are the same (0).

**Root cause (from source analysis):**

In `controllers/media/tracks.py` line 536:

```python
"timestamp_added": int(item.date_added.timestamp()) if item.date_added else UNSET,
```

When `date_added` is None (most providers), `UNSET` is used, which means the field is excluded from the database INSERT (line 215 of `helpers/database.py`):

```python
values = {k: v for k, v in values.items() if v is not UNSET}
```

The database column then gets its default value (0), making date-based sorting meaningless.

**Suggested fix:**

When the provider does not supply `date_added`, fall back to the current time (`utc_timestamp()`) instead of `UNSET`. This way, tracks are at least sorted by the time they were added to the MA library:

```python
"timestamp_added": int(item.date_added.timestamp()) if item.date_added else int(utc_timestamp()),
```

This would affect all media types (tracks, albums, artists, playlists, etc.) in their respective controllers. An alternative would be to set a database column default of `CURRENT_TIMESTAMP` instead of 0.

### How to reproduce

1. Use Deezer (or Spotify) as music provider
2. Add several tracks to favorites at different times
3. Go to Tracks > Favorites
4. Sort by "Recently Added"
5. Observe that the order is alphabetical instead of chronological

### Music Providers

**Problem:** Deezer (confirmed). Spotify, Jellyfin, and others that do not set `date_added` during parsing are also affected based on source analysis.

**Working:** Tidal and filesystem_local correctly populate `date_added`.

### Player Providers

Not player-related. Using DLNA (JBL Authentics 200) and Sonos S1 (Sonos Move), both unaffected by this issue.

### Full log output

N/A. This is a data model issue, not a runtime error. No relevant log entries are produced.

### Additional information

This also affects albums and artists from the same providers. The `timestamp_added` column in the database is 0 for all Deezer items.

Related: #1921 (reported inverted sort order with Spotify, likely the same root cause since all timestamps are 0)

N/A, standalone Docker

Home Assistant Container

Proxmox

- [x] Yes

sfortis:
Fixes music-assistant/support#5094

The Deezer API returns a `time_add` field (Unix timestamp) for library items indicating when the user added them to favorites. This field was available on the deezer-python Resource objects but was never extracted.

Without this, all library items get `timestamp_added` set to the sync time (via SQLite column default), making "Sort by Recently Added" show sync order instead of the actual order the user favorited items.

This extracts `time_add` and sets `date_added` for tracks, albums, artists, and playlists. Tracks added individually at different times now sort correctly. Items added in bulk still share the same narrow time window (expected behavior from Deezer's side).

Tested with a Deezer account: confirmed `time_add` is present in API responses and timestamps are correctly stored in the database after sync.

Note: Spotify has a similar issue (`added_at` field in wrapper objects is discarded), but that is not addressed here.

### DIFF
--- a/music_assistant/providers/deezer/__init__.py
+++ b/music_assistant/providers/deezer/__init__.py
@@ -5,6 +5,7 @@ import uuid
 from asyncio import TaskGroup
 from collections.abc import AsyncGenerator
 from dataclasses import dataclass
+from datetime import UTC, datetime
 from math import ceil
 from typing import Any, Literal, cast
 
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
Thanks @sfortis 🙏

PR #3382, opened 2026-03-13
Fix Sendspin playback stability issues

maximmaxim345:
Several race conditions could cause Sendspin players to get stuck, immediately pause, or get out of sync, especially under rapid group membership changes and when DSP was enabled.

## aiosendspin 4.3.4

Bumps aiosendspin to [4.3.4](https://github.com/Sendspin/aiosendspin/releases/tag/4.3.4) which include these fixes:
- Reconnection loop after connection takeover
- Stale chunk log spam in the debug console after web player disconnects
- Out of sync playback when DSP clients join or reconnect

## Sendspin Provider fixes

- Prevent duplicate join attempts when multiple group events fire at once
- Cancel stale joins that no longer match the current group
- Clean up resources when a join fails partway through
- Stop the correct stream instance instead of tearing down the whole group
- Flush audio pipelines on track end so DSP players don't get stuck in "playing" state

### DIFF
--- a/music_assistant/providers/sendspin/manifest.json
+++ b/music_assistant/providers/sendspin/manifest.json
@@ -7,7 +7,7 @@
   "documentation": "https://music-assistant.io/player-support/sendspin/",
   "codeowners": ["@music-assistant"],
   "credits": ["[Sendspin](https://sendspin-audio.com)"],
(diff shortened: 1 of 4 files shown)


MarvinSchenkel (approved):
LGTM, thanks @maximmaxim345

PR #3384, opened 2026-03-13
Add musicbrainz get_release_group functions

OzGav:
Add functions in support of https://github.com/music-assistant/server/pull/3110


MarvinSchenkel:
Let's wait for 2.9 with adding additional load to our musicbrainz mirror. It seems to run fine in beta, but with the 2.8 release coming we will already increase the load on the mirror by the increased number of users.

marcelveldt (approved):
Thanks, @OzGav !

PR #3385, opened 2026-03-13
Fix albums not being able to be added to playlists during playlist creation

fmunkes:
The frontend is not able to add albums to playlist anymore - I introduced that bug accidentally here https://github.com/music-assistant/server/pull/3216

If a playlist supports tracks, it supports albums too, as we acquire individual tracks for the playlist anyways.

Frontend PR: https://github.com/music-assistant/frontend/pull/1577

### DIFF
--- a/music_assistant/controllers/media/playlists.py
+++ b/music_assistant/controllers/media/playlists.py
@@ -130,6 +130,10 @@ class PlaylistController(MediaControllerBase[Playlist]):
 
         # Default is track for backwards compatibility.
         media_types_set = {MediaType.TRACK} if not media_types else set(media_types)
+        if MediaType.ALBUM in media_types_set:
+            # an album is unwrapped, so we remove that and use tracks instead
+            media_types_set.remove(MediaType.ALBUM)
+            media_types_set.add(MediaType.TRACK)
         if not provider_instance_or_domain and not media_types_set:
             # builtin can handle all media_types
             media_types_set.update(
@@ -146,8 +150,6 @@ class PlaylistController(MediaControllerBase[Playlist]):
         ):
             # PLAYLIST_CREATE is deprecated
             supported_types.add(MediaType.TRACK)
-            # an album is unwrapped to individual tracks in self.add_playlist_tracks
-            supported_types.add(MediaType.ALBUM)
         if ProviderFeature.PLAYLIST_CREATE_AUDIOBOOKS in provider.supported_features:
             supported_types.add(MediaType.AUDIOBOOK)
         if ProviderFeature.PLAYLIST_CREATE_PODCAST_EPISODES in provider.supported_features:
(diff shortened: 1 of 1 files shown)


fmunkes:
I thought about this a bit. The current approach uses the knowledge, that the backend unwraps albums, but somehow this feels wrong - the allowed media_types are not album, but tracks for playlist. So my question is, should I revert the change to the playlist_create method, and then adjust the playlist_add_tracks method to only the specific media types? It would then be the client's behavior, i.e. the frontend, to be able to add albums to playlist, by receiving the tracks first (or maybe call a new helper method in the backend), before adding the tracks. Let me know what you think.

chrisuthe:
My 2cents is that organizational containers  should all be considered at once here - albums, genres, seasons, years, artists, search results etc... 

I  suggest that containers that could contain playlist-compatible items should be able to pass the items to the playlist creation (aka tracks, or episodes or stations) so that as future organizational work happens (incoming genre's are a great example) they can provide their own way to give playlist handling the appropriate data...

MarvinSchenkel:
Kinda agreeing with @chrisuthe here. I think playlists should only be able to hold **directly playable items**, such as Tracks, Podcast episodes, Audiobooks and radio streams. 

I think we should postpone support for 'Containers' such as genres, albums etc. when we introduce the 'dynamic playlist' feature we have been talking about for a while now. The (still not spec'ed) idea is that we add some sort of rule-based mechanism for dynamic playlist, e.g. "30% genre Pop + 20% Album X + 50% radio mode over playlist Y".

I feel this is the more logical place for these kind of MediaTypes as opposed to static playlists. Thoughts?

fmunkes:
The problem is, before my changes adding an album to a playlist was possible, as the backend silently took the individual tracks of an album and added them. The remnants of this code are still in the add_playlist_tracks. The frontend shows an add to playlist as an option for albums as well, which fails now. So my merged changes to frontend/ backend agree with both of you, they are strict to the playable type.
At the same time, I discovered on the weekend that I accidentally broke a previously available functionality.

On your last comment - I personally like to be able to add a whole album to a static playlist. The cleanest could be a new backend method, which then unwraps the album to single tracks, and adds these?

MarvinSchenkel:
> The problem is, before my changes adding an album to a playlist was possible, as the backend silently took the individual tracks of an album and added them. The remnants of this code are still in the add_playlist_tracks. The frontend shows an add to playlist as an option for albums as well, which fails now. So my merged changes to frontend/ backend agree with both of you, they are strict to the playable type. At the same time, I discovered on the weekend that I accidentally broke a previously available functionality.
> 
> On your last comment - I personally like to be able to add a whole album to a static playlist. The cleanest could be a new backend method, which then unwraps the album to single tracks, and adds these?

If there is existing logic for this I would use that and postpone any improvements to when we introduce the dynamic playlist feature. So with the changes in this PR, does the server automatically add the Album tracks to the Playlist?

fmunkes:
Yes. This PR and the one linked for the frontend. Then the previous functionality is back while keeping the new one.

MarvinSchenkel (approved):
In that case, lgtm, thanks @fmunkes

PR #3386, opened 2026-03-13
Add support for legacy Shoutcast servers using ICY protocol

OzGav:
Other PR was corrupted during a merge https://github.com/music-assistant/server/pull/2868


music_assistant/controllers/streams/audio.py
```diff
@@ -549,9 +554,70 @@ async def resolve_radio_stream(self, url: str) -> tuple[str, StreamType]:
...
                     stream_type = StreamType.HLS
 
-        except Exception as err:
-            self.logger.warning("Error while parsing radio URL %s: %s", url, str(err))
-            return (url, stream_type)
+        except TimeoutError as err:
+            self.logger.warning("Timeout while parsing radio URL %s", url)
+            raise InvalidDataError(f"Timeout connecting to {url}") from err
+
+        except aiohttp.ClientError as err:
```
marcelveldt, line 611: this logic has become huge. can we split some of it off into an additional (private) helper method maybe ?

music_assistant/controllers/streams/audio.py
```diff
@@ -1952,8 +2076,191 @@ def clear_crossfade_data(self, queue_id: str) -> None:
...
+
+        if cleaned_stream_title != streamdetails.stream_title:
+            self.logger.log(VERBOSE_LOG_LEVEL, "ICY Radio streamtitle original: %s", stream_title)
+            self.logger.log(
+                VERBOSE_LOG_LEVEL, "ICY Radio streamtitle cleaned: %s", cleaned_stream_title
+            )
+            streamdetails.stream_title = cleaned_stream_title
+
+    async def _validate_shoutcast_stream(self, url: str) -> bool:
+        """Validate if a URL is a legacy Shoutcast stream that returns "ICY 200 OK".
```
```suggestion
        """
        Validate if a URL is a legacy Shoutcast stream that returns "ICY 200 OK".
```

music_assistant/controllers/streams/audio.py
```diff
@@ -1952,8 +2076,191 @@ def clear_crossfade_data(self, queue_id: str) -> None:
...
+                    break
+
+        finally:
+            writer.close()
+            await writer.wait_closed()
+
     # --- Private methods ---
 
+    def _parse_icy_metadata(self, meta_data: bytes, streamdetails: StreamDetails) -> None:
+        """Parse ICY metadata and update streamdetails with stream title.
```
```suggestion
        """
        Parse ICY metadata and update streamdetails with stream title.
```

music_assistant/controllers/streams/audio.py
OzGav: Extracted Shoutcast/ICY detection logic into `_handle_client_error_for_radio_stream` and added `_cache_and_return_radio_result` to reduce code duplication.

music_assistant/controllers/streams/audio.py
```diff
@@ -493,6 +498,43 @@ def _on_inband_metadata(metadata: dict[str, str]) -> None:
...
+            self.logger.debug("ICY response detected for %s, validating Shoutcast stream", url)
+            if await self._validate_shoutcast_stream(url):
+                return await self._cache_and_return_radio_result(url, StreamType.SHOUTCAST)
+            self.logger.warning("ICY response detected but Shoutcast validation failed for %s", url)
+            return await self._cache_and_return_radio_result(url, fallback_stream_type)
+
+        # Other aiohttp errors - might still be Shoutcast, check it
+        self.logger.debug("aiohttp error for %s, checking if legacy Shoutcast stream", url)
+        if await self._validate_shoutcast_stream(url):
+            return await self._cache_and_return_radio_result(url, StreamType.SHOUTCAST)
```
marcelveldt, line 532: isn't this a bit redundant ?

music_assistant/controllers/streams/audio.py
```diff
@@ -493,6 +498,43 @@ def _on_inband_metadata(metadata: dict[str, str]) -> None:
...
+
+        # Unknown error - still try to stream
+        self.logger.warning(
```
marcelveldt, line 535: what is the point of still trying to stream here ?
shouldn't we catch common errors like 404 etc

music_assistant/controllers/streams/audio.py
```diff
@@ -549,63 +590,115 @@ async def resolve_radio_stream(self, url: str) -> tuple[str, StreamType]:
...
-        ) as resp:
-            headers = resp.headers
-            meta_int = int(headers["icy-metaint"])
-            while True:
-                try:
-                    yield await resp.content.readexactly(meta_int)
-                    meta_byte = await resp.content.readexactly(1)
-                    if meta_byte == b"\x00":
+
+        # Try aiohttp first (works for most ICY streams)
```
marcelveldt: why do we try this ? we have the detection logic that should catch this upfront so we can do shoutcast streaming if needed and do regular http based icy streaming otherwise

music_assistant/controllers/streams/audio.py
OzGav: You’re right, this was redundant.

I’ve removed the fallback from get_icy_radio_stream() and now only detect once in resolve_radio_stream().

music_assistant/controllers/streams/audio.py
OzGav: I’ve updated it to properly handle HTTP status errors.
Only aiohttp parsing errors (like when it can’t parse “ICY 200 OK”) trigger the Shoutcast check now. Real connection failures raise exceptions instead of silently continuing.

edit: error 400 is passed through to check as that is being returned when commencing playback…

music_assistant/controllers/streams/audio.py
OzGav: Yeah, this was a mistake on my part. I didn’t trust the detection logic so I added a fallback.
I’ve fixed it by:
	1.	Removing the try/except fallback from get_icy_radio_stream()
	2.	Creating a separate get_shoutcast_stream() method with the raw socket code
	3.	The routing now uses the stream type from detection to call the right method
So resolve_radio_stream() detects the type once, and the streaming code just does what it’s told. Cleaner separation between detection and streaming.

music_assistant/controllers/streams/audio.py
```diff
@@ -2034,8 +2071,193 @@ def clear_crossfade_data(self, queue_id: str) -> None:
             self.logger.debug("Clearing crossfade data for queue %s", queue_id)
             del self._crossfade_data[queue_id]
 
+    async def get_shoutcast_stream(
+        self, url: str, streamdetails: StreamDetails
+    ) -> AsyncGenerator[bytes, None]:
+        """Get (radio) audio stream from legacy Shoutcast server using raw socket connection.
```
OzGav:
```suggestion
        """
        Get (radio) audio stream from legacy Shoutcast server using raw socket connection.
```

music_assistant/controllers/streams/audio.py
```diff
@@ -495,27 +500,65 @@ def _on_inband_metadata(metadata: dict[str, str]) -> None:
...
+        """Cache and return a radio stream resolution result."""
+        result = (url, stream_type)
+        await self.mass.cache.set(
+            url,
+            result,
+            expiration=3600 * 3,
+            provider=CACHE_PROVIDER,
+            category=CACHE_CATEGORY_RESOLVED_RADIO_URL,
+        )
+        return result
```
marcelveldt: why is this ?
a method with just one call

marcelveldt: OK, reading further down below I now know why this was added. Then I think the name is just a bit weird.
Why not just call it "_cache_radio_result" or something like that ?

OzGav: Yep simpler is better

music_assistant/controllers/streams/audio.py
```diff
@@ -584,45 +625,61 @@ async def resolve_radio_stream(self, url: str) -> tuple[str, StreamType]:
...
+                raise MediaNotFoundError(f"Radio stream not found: {url}") from err
+            if err.status == 403:
+                raise InvalidDataError(f"Access denied to radio stream: {url}") from err
+            if err.status >= 500:
+                raise InvalidDataError(
+                    f"Radio stream server error (HTTP {err.status}): {url}"
+                ) from err
+            if err.status == 400:
+                # 400 errors might be from legacy Shoutcast servers
+                return await self._handle_client_error_for_radio_stream(url, err, stream_type)
```
marcelveldt, line 750: Redirected legacy Shoutcast streams can still miss this path. When `aiohttp` follows a redirect and then raises on the final `ICY 200 OK`, `err.request_info.url` points at the final stream URL, but this code validates/caches the original `url`. Since `_validate_shoutcast_stream` uses a raw socket and does not follow redirects, common radio URLs that redirect to the actual ICY endpoint will fall back to HTTP and fail playback. Can we validate/cache `str(err.request_info.url)` when present?

music_assistant/controllers/streams/audio.py
```diff
@@ -2120,8 +2140,207 @@ def clear_crossfade_data(self, queue_id: str) -> None:
...
+            return
+
+        if cleaned_stream_title != streamdetails.stream_title:
+            self.logger.log(VERBOSE_LOG_LEVEL, "ICY Radio streamtitle original: %s", stream_title)
+            self.logger.log(
+                VERBOSE_LOG_LEVEL, "ICY Radio streamtitle cleaned: %s", cleaned_stream_title
+            )
+            streamdetails.stream_title = cleaned_stream_title
+
+        if " - " in cleaned_stream_title:
```
marcelveldt, line 2534: This now refreshes radio metadata/artwork even when the `StreamTitle` has not changed. Previously the artwork update only happened inside the title-changed branch, so repeated ICY metadata blocks for the same song did not keep signaling the queue and scheduling artwork refreshes. Can we keep the `_update_radio_stream_metadata` call gated on `cleaned_stream_title != streamdetails.stream_title`?

music_assistant/controllers/streams/audio.py
```diff
@@ -2120,8 +2140,207 @@ def clear_crossfade_data(self, queue_id: str) -> None:
...
+        try:
+            parsed = urlparse(url)
+            host = parsed.hostname
+            port = parsed.port or 80
+            path = parsed.path or "/"
+            if parsed.query:
+                path = f"{path}?{parsed.query}"
+
+            # Open raw socket connection with timeout
+            reader, writer = await asyncio.wait_for(asyncio.open_connection(host, port), timeout=10)
```
marcelveldt, line 2561: If the connection opens successfully but `writer.drain()` or the following `reader.readline()` times out/errors, the writer is never closed because cleanup only happens on the success path below. Can we wrap the opened writer in a `finally` so the validation socket is closed on all error paths?

marcelveldt (commented):
Found a few line-level issues to address.

music_assistant/controllers/streams/audio.py
OzGav: Done

music_assistant/controllers/streams/audio.py
OzGav: Yes, fixed.

music_assistant/controllers/streams/audio.py
OzGav: Fixed. The handler now pulls the post-redirect URL from err.request_info.url and uses it for both validation and the cached result, so redirected legacy Shoutcast streams get probed at their real endpoint and future plays skip the redirector.

marcelveldt (approved):
LGTM, Thanks @OzGav !

marcelveldt:
@OzGav small merge conflict

PR #3388, opened 2026-03-13
Adding provider support for YouTube

giantorth:
I wanted support for youtube direct playback, not just via YT Music.  Also have A documentation PR to submit to go with this provider plugin.


OzGav:
In its current form this would allow playback of YouTube content without a subscription and bypassing the ads. This is not allowed.

MarvinSchenkel:
We definitely cannot accept this PR as it violates the TOS of Youtube. You are already able to play most videos from Youtube via Youtube Music by searching the exact video title and/or adding the video to your likes. What are you currently missing from the YT Music provider?

OzGav:
@giantorth please close this if you wont be progressing any changes.

marcelveldt:
Would be an idea maybe to add YouTube as metadata provider so you can links to the YouTube video for songs?

MarvinSchenkel:
> Would be an idea maybe to add YouTube as metadata provider so you can links to the YouTube video for songs?

They're the same thing. Any song on YT Music will share the same id with a video on Youtube itself.

giantorth:
Closing, if anybody wants this plugin it's packaged here: https://github.com/giantorth/music-assistant-yt-dlp

PR #3390, opened 2026-03-14
Fix artist name splitting with semicolons and Vorbis multi-field handling

OzGav:
1. music_assistant/helpers/tags.py:
- Fixed artists and album_artists properties to not split on semicolons when values are already in a list (multiple ARTIST fields from Vorbis)
- Added MB ID count check for singular ARTIST/ALBUMARTIST tags - if count is 1, don't split even on semicolons
- Added warning in _parse_vorbis_artist_tags() for non-standard ARTISTS/ALBUMARTISTS tags in Vorbis files
- Added proper comments with spec references
2. tests/core/test_tags.py:
- Added test_vorbis_multiple_artist_fields_semicolon_in_name - mock test for multiple ARTIST fields with semicolon
- Added test_flac_multiple_artist_fields_semicolon_e2e - end-to-end test with real FLAC file
- Added test_id3_artist_tag_semicolon_single_mbid - test for single ARTIST with semicolon and 1 MB ID
- Added test_id3_artist_tag_semicolon_multiple_mbids - test for ARTIST with semicolon and multiple MB IDs
- Added test_id3_albumartist_tag_semicolon_single_mbid - same for album artist
3. tests/fixtures/ArtistWithSemicolon.flac:
- New test fixture with multiple ARTIST fields, one containing ave;new


kepstin:
As long as you are [recommending that people use MusicBrainz Picard](https://www.music-assistant.io/music-providers/filesystem/#:~:text=Recommended%20Tagger) for tagging, you should not print a warning when handling the `ARTISTS` tag which is [documented to be used by MA](https://www.music-assistant.io/music-providers/filesystem/#tags-used-by-ma) to disambiguate multiple artists and is [set by Picard](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#artists) in order to store [the separate artists as a multi-value tag](https://picard-docs.musicbrainz.org/en/variables/tags_basic.html#:~:text=artists,-A).

Note that Picard does not (by default) store an `ALBUMARTISTS` tag, although the information needed to do so is available in the internal [`_albumartists` tagger script variable](https://picard-docs.musicbrainz.org/en/variables/variables_basic.html#:~:text=%5Falbumartists).

music_assistant/helpers/tags.py
```diff
@@ -331,9 +344,20 @@ def writers(self) -> tuple[str, ...]:
     @property
     def album_artists(self) -> tuple[str, ...]:
         """Return (all) album artists (if any)."""
-        # prefer multi-artist tag (ALBUMARTISTS plural)
+        # Check for "albumartists" key which can come from two sources:
+        # 1. ID3: TXXX:ALBUMARTISTS - a MusicBrainz/Picard extension using semicolon-delimited
+        #    values (TPE2 multi-value support is inconsistent across players).
+        #    See: https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html
```
kepstin: Picard does not store an `ALBUMARTISTS` tag. The provided reference is invalid, it doesn't mention `ALBUMARTISTS` anywhere.

[`_albumartists`](https://picard-docs.musicbrainz.org/en/variables/variables_basic.html#:~:text=%5Falbumartists) (note the leading `_`) is an internal tagger script variable, not written to files by default.

music_assistant/helpers/tags.py
```diff
@@ -976,10 +1002,23 @@ def _parse_vorbis_artist_tags(tags: VCommentDict, result: dict[str, Any]) -> Non
...
 
-    # Explicit ARTISTS tag takes precedence if present
+    # ARTISTS (plural) is non-standard in Vorbis - it's a MusicBrainz/Picard ID3 convention.
+    # Vorbis spec recommends multiple ARTIST (singular) fields instead.
+    # Accept it but warn. See: https://xiph.org/vorbis/doc/v-comment.html
     if artists := _vorbis_get_multi(tags, "ARTISTS"):
+        LOGGER.warning(
+            "ARTISTS tag found in Vorbis file. Use multiple ARTIST fields instead: %s",
+            artists,
+        )
```
kepstin: The [Music Assistant documentation](https://www.music-assistant.io/music-providers/filesystem/#tags-used-by-ma) "strongly" recommends using the [Artists tag as defined by Picard](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#artists) to disambiguate multiple artist names. It's inappropriate to print a warning when people are using a recommended tag which is set by default by the [recommended tagger application](https://www.music-assistant.io/music-providers/filesystem/#:~:text=Recommended%20Tagger).

kepstin (changes requested):

music_assistant/helpers/tags.py
```diff
@@ -282,9 +282,20 @@ def album(self) -> str | None:
     @property
     def artists(self) -> tuple[str, ...]:
         """Return track artists."""
-        # prefer multi-artist tag (ARTISTS plural)
+        # Check for "artists" key which can come from two sources:
+        # 1. ID3: TXXX:ARTISTS - a MusicBrainz/Picard extension using semicolon-delimited values
+        #    in a single field (ID3 doesn't support multiple instances of the same tag).
+        #    See: https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html
+        # 2. Vorbis: Multiple ARTIST (singular) fields per the spec, stored as list by mutagen.
+        #    See: https://xiph.org/vorbis/doc/v-comment.html
```
kepstin: This comment is incorrect, the following code loads the `ARTISTS` (plural) vorbiscomment tag, not the `ARTIST` (singular) tag.

music_assistant/helpers/tags.py
```diff
@@ -976,10 +1002,23 @@ def _parse_vorbis_artist_tags(tags: VCommentDict, result: dict[str, Any]) -> Non
         else:
             result["albumartist"] = albumartist_values[0]
 
-    # Explicit ARTISTS tag takes precedence if present
+    # ARTISTS (plural) is non-standard in Vorbis - it's a MusicBrainz/Picard ID3 convention.
+    # Vorbis spec recommends multiple ARTIST (singular) fields instead.
+    # Accept it but warn. See: https://xiph.org/vorbis/doc/v-comment.html
     if artists := _vorbis_get_multi(tags, "ARTISTS"):
```
kepstin, line 1020: I'm a bit confused by this code. It seems redundant with the code on line 294 which should already be handling multivalue vorbis tags?

tests/core/test_tags.py
kepstin, line 1: Some tests should be added which cover the multi-value `ARTISTS` tag. I would recommend using a fixture file which is tagged by Picard to ensure you have representative tags.

music_assistant/helpers/tags.py
OzGav: I just changed the documentation to differentiate between id3 and vorbis tags.

music_assistant/helpers/tags.py
OzGav: Thanks. Cut and paste error. Docstring and comment improved

music_assistant/helpers/tags.py
OzGav: I think there may be some confusion here between the internal key name and the tag name. The artists key in self.tags is set by the parser when it encounters multiple ARTIST (singular) fields in Vorbis files. The comment was referring to those multiple ARTIST fields, not the non-standard ARTISTS (plural) tag.

The ARTISTS (plural) tag is non-standard in Vorbis - we accept it with a warning for compatibility with mis-tagged files, but that's an implementation detail rather than expected behavior.

That said, the comment was overly verbose. I've simplified the docstring to match the style used elsewhere.

music_assistant/helpers/tags.py
OzGav: See above - the ARTISTS (plural) tag handling is separate from the multiple ARTIST (singular) fields handling. They're not redundant; they handle two different tagging approaches. The warning is intentional for mis-tagged files.

tests/core/test_tags.py
OzGav: This is already covered by test_parse_metadata_from_id3tags

music_assistant/helpers/tags.py
```diff
@@ -281,10 +287,20 @@ def album(self) -> str | None:
...
+        """Return track artists.
+
+        Vorbis: parser returns artists (plural) for multiple ARTIST fields.
+        ID3: parser returns artists from TXXX:ARTISTS, or artist (singular) split on delimiter here.
+        MP4/APEv2: parser returns artist (singular), split on delimiter here.
+        """
         if tag := self.tags.get("artists"):
-            artists = split_items(tag)
+            # Runtime check: mutagen returns list[str] for Vorbis multi-field
+            if isinstance(tag, list) and len(tag) > 1:  # type: ignore[unreachable]
```
MarvinSchenkel, line 296: These unreachable mypy warnings actually indicate an incorrect type declaration for `self.tags`. This should be `dict[str, str | list[str]]` which will also remove the need for these 'type: ignores'

OzGav: The tricky thing is that only the Vorbis parser returns lists (for artists/albumartists) - all other parsers return strings. Changing the type to handle this results in 40+ changes elsewhere. We can't normalize to semicolon-joined strings because the whole point is that Vorbis multi-field tags are already properly separated without delimiter ambiguity - joining and re-splitting would break artist names containing semicolons. Happy to do the 40+ changes if you think it's worth it

OzGav:
Afer getting the ID3v2.4 PR merged I realised the same could be done here to add more support for semi-colon in name edge case. So this has been added along with tests.

music_assistant/helpers/tags.py
OzGav: Agreed to keep the ignore for now.

music_assistant/helpers/tags.py
```diff
@@ -995,10 +1014,23 @@ def _parse_vorbis_artist_tags(tags: VCommentDict, result: dict[str, Any]) -> Non
...
+            "ARTISTS tag found in Vorbis file. Use multiple ARTIST fields instead: %s",
+            artists,
+        )
         result["artists"] = artists
 
+    if albumartists := _vorbis_get_multi(tags, "ALBUMARTISTS"):
+        LOGGER.warning(
+            "ALBUMARTISTS tag found in Vorbis file. Use multiple ALBUMARTIST fields instead: %s",
+            albumartists,
+        )
```
marcelveldt: should we really warn here ? the fact that we can handle it is fine right ?
i would only warn if something is broken and can be fixed by the user or we couldnt parse it. In this case you made our tags reader more flexible so you dont need the warning

marcelveldt (approved):

music_assistant/helpers/tags.py
OzGav: Yeah. It was more to alert the user that they should do something about fixing their tags. Remove or drop to info?

marcelveldt: I would remove it. You fixed it so they don't have to change tags, which is always better

OzGav: ok, removed.

PR #3393, opened 2026-03-14
Add MusicMe music provider

JulienDeveaux:
## Summary

Add a new music provider for **MusicMe** ([musicme.com](https://www.musicme.com/)), a French legal music streaming service operated by ApachNetwork with a catalogue of 13M+ tracks from major and independent labels.

  ### Features

  - **Search**: artists, albums, tracks (with web search fallback for niche queries)
  - **Browse**: new releases (by genre), top artists, radios by theme, homepage highlights
  - **Recommendations**: powers the Discover page with 4 sections (featured, new releases, top artists, radios)
  - **Artist details**: albums, top tracks (derived from recent albums)
  - **Playlists**: user playlist listing
  - **Streaming**: AAC in MP4 container, 44.1kHz stereo, ~111kbps via ephemeral ticket-based URLs

  ## Supported features

  - `LIBRARY_PLAYLISTS`
  - `SEARCH`
  - `ARTIST_ALBUMS`, `ARTIST_TOPTRACKS`
  - `BROWSE`, `RECOMMENDATIONS`

  ## Notes

  - Requires a MusicMe account (paid subscription for full streaming)
  - MusicMe is primarily available in France
  - Streaming tickets are single-use and ephemeral (generated on-demand before each playback)
  - The dataservice API uses a lightweight XOR cipher for response obfuscation (not encryption)


music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+    ProviderFeature.BROWSE,
+    ProviderFeature.RECOMMENDATIONS,
+}
+
+DATASERVICE_BASE = "https://dataservice.musicme.com/dataservice/v3"
+STREAM_BASE = "https://stream.hosting-media.net/musicme"
+WEB_BASE = "https://www.musicme.com"
+PARTNER_ID = "22876"
+_CLIENT_JSON = json.dumps(
+    {"type": "desktop-web", "context": "www.musicme.com", "app": "mmplayer"}
```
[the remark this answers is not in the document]
JulienDeveaux: the MusicMe dataservice expects literal JSON in the query string. URL-encoding breaks the API.

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+            "Successfully logged in to MusicMe (catalog=%s)",
+            player_init.get("catalogSize", "?"),
+        )
+
+    def _build_base_params(self) -> str:
+        """Build the common query string for dataservice API calls."""
+        parts = ["format=json", f"partnerid={PARTNER_ID}", f"client={_CLIENT_JSON}"]
+        if self._user_id:
+            parts.append(f"userid={self._user_id}")
+        return "&".join(parts)
```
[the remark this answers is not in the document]
JulienDeveaux: the MusicMe dataservice expects literal JSON in the query string. URL-encoding breaks the API.

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+
+        async with self._session.get(url, allow_redirects=True) as resp:
+            if resp.status != 200:
+                return None
+            raw_bytes = await resp.read()
+            content = raw_bytes.decode("latin-1", errors="replace")
+
+        result = SearchResults()
+
+        match = re.search(r"window\.playerInit\s*=\s*(\{.*?\});", content, re.DOTALL)
```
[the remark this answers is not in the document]
JulienDeveaux: Already fixed

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        try:
+            async with self._session.get(f"{WEB_BASE}/?f=1") as resp:
+                resp.raise_for_status()
+                raw_bytes = await resp.read()
+                content = raw_bytes.decode("latin-1", errors="replace")
+        except aiohttp.ClientResponseError as err:
+            msg = f"MusicMe page load failed: HTTP {err.status}"
+            raise LoginFailed(msg) from err
+
+        match = re.search(r"window\.playerInit\s*=\s*(\{.*?\});", content, re.DOTALL)
```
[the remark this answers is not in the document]
JulienDeveaux: Already fixed

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        return [self._parse_album(a) for a in albums if a.get("barcode")]
+
+    @use_cache(3600 * 3)
+    async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> list[Track]:
+        """Get playlist tracks."""
+        data = await self._api_get(f"/playlist/{prov_playlist_id}?resources=tracks")
+        if not data:
+            return []
+        tracks = data.get("results", {}).get("tracks", [])
+        return [self._parse_track(t) for t in tracks if t.get("barcode")]
```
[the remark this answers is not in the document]
JulienDeveaux: The page parameter is part of the MusicProvider base class signature. MusicMe returns all tracks in one call, so pagination isn't needed, but the parameter must stay for interface compliance

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+
+    from music_assistant_models.config_entries import ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.LIBRARY_PLAYLISTS,
+    ProviderFeature.LIBRARY_RADIOS,
```
OzGav: This should only be declared if the user creates a curated set of radio stations that are pulled from the provider.

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+
+    _user_id: str | None = None
+    _http_session: aiohttp.ClientSession | None = None
+    throttler: ThrottlerManager
+
+    async def handle_async_init(self) -> None:
+        """Handle async initialization of the provider."""
+        if not self.config.get_value(CONF_USERNAME) or not self.config.get_value(CONF_PASSWORD):
+            msg = "Invalid login credentials"
+            raise LoginFailed(msg)
```
OzGav:
```suggestion
            raise SetupFailedError(msg)
```

MIssing credentials is not a login failure

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        """Handle unload/close of the provider."""
+        if self._http_session and not self._http_session.closed:
+            await self._http_session.close()
+            self._http_session = None
+
+    @property
+    def _session(self) -> aiohttp.ClientSession:
+        """Return the provider's dedicated HTTP session."""
+        if self._http_session is None or self._http_session.closed:
+            self._http_session = aiohttp.ClientSession()
```
OzGav: Maybe I'm wrong but this seems a bit of a hack to address an error state by restarting a closed session?

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        return self._http_session
+
+    # ---- search ----
+
+    @use_cache(3600 * 24)
```
OzGav: This would mean a user would see the same search results for 24 hours regardless of what they search for

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+                if isinstance(tracks, list):
+                    result.tracks = [
+                        self._parse_track(t)
+                        for t in tracks
+                        if t.get("barcode") and t.get("streamable")
+                    ]
+
+        if not result.artists and not result.albums and not result.tracks:
+            self.logger.debug("API search empty, trying web search fallback")
+            web_result = await self._web_search_fallback(search_query, media_types, limit)
```
OzGav: Why is this needed? What is wrong with the API?

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+            if web_result:
+                result = web_result
+
+        return result
+
+    # ---- browse ----
```
OzGav: Assuming the data returned by these methods is fairly static they should be cached

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+                if item.get("id"):
+                    folder.items.append(self._parse_radio(item))
+            if folder.items:
+                result.append(folder)
+
+        return result
+
+    # ---- library ----
+
+    async def get_library_radios(self) -> AsyncGenerator[Radio, None]:
```
OzGav: Related to LIBRARY_RADIOS. This method is only for when the provider allows the user to create a curated list of stations (for example TuneIn). If the user selects stations to be added to the library from a large list then that is what browse is for. If MusicMe allows user to save a list of favourite stations then all good.

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        for item in data.get("results", {}).get("theme-airplays", []):
+            if item.get("id"):
+                yield self._parse_radio(item)
+
+    async def get_radio(self, prov_radio_id: str) -> Radio:
+        """Get a single radio by id, loading full details from the airplay endpoint."""
+        data = await self._api_get(f"/airplay/{prov_radio_id}?resources=tracks")
+        if data and "item" in data:
+            return self._parse_radio(data["item"])
+        async for radio in self.get_library_radios():
```
OzGav: Why would this be required and why would you not raise MediaNotFoundError immediately?

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        msg = f"Radio {prov_radio_id} not found"
+        raise MediaNotFoundError(msg)
+
+    async def get_library_playlists(self) -> AsyncGenerator[Playlist, None]:
+        """Retrieve library playlists from MusicMe."""
+        data = await self._api_get('/playlists?resources=home{details:"list"}')
+        if not data:
+            return
+        items: Any = data.get("results", {})
+        if isinstance(items, dict) and "items" in items:
```
OzGav: What is going on here. Why dont you know what the API will return?

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+                disc_track_parts = disc_track.split("_")
+                track.disc_number = int(disc_track_parts[0])
+                track.track_number = int(disc_track_parts[1])
+            except (ValueError, IndexError):
+                pass
+        for art_obj in track_obj.get("artists", []):
+            if art_obj.get("id"):
+                track.artists.append(self._parse_artist(art_obj))
+        if track_obj.get("album"):
+            track.album = self._parse_album(
```
OzGav: I think an ItemMapping should be created here as there is no data to populate the album?

music_assistant/providers/musicme/__init__.py
```diff
@@ -0,0 +1,1008 @@
...
+        """Make a GET request to MusicMe dataservice and decrypt the response.
+
+        :param endpoint: API endpoint path (e.g. '/search?query=...')
+        """
+        url = f"{DATASERVICE_BASE}{endpoint}"
+        sep = "&" if "?" in url else "?"
+        url = f"{url}{sep}{self._build_base_params()}"
+
+        self.logger.debug("GET %s", endpoint.split("?")[0])
+        async with self._session.get(url) as response:
```
OzGav: There is no error handling here if there is some sort of connection problem?

OzGav:
Also this is very long for one file. We generally like to see the following structure
init.py
constants.py
helpers.py - Put your decrypt function in here
provider.py

and also install and run pre-commit before pushing any changes. Currently this provider has a number of mypy issues.

music_assistant/providers/musicme/__init__.py
JulienDeveaux: It's indeed a fixed catalog, i've removed the flag

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Indeed, changed

music_assistant/providers/musicme/__init__.py
JulienDeveaux: That's a cache for query, I didn't taught that in 24h it would change dramatically, I've reduced to 3 hours

music_assistant/providers/musicme/__init__.py
JulienDeveaux: MusicMe's dataservice /search is an autocomplete endpoint that only returns popular/exact matches. Many valid artists are not returned by it but are accessible via the PHP web search. This fallback catches those cases

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Indeed, added

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Removed along the LIBRARY_RADIOS first comment

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Removed with the LIBRARY_RADIOS stuff

music_assistant/providers/musicme/__init__.py
JulienDeveaux: MusicMe's playlist API wraps results in varying nesting levels depending on the response. MusicMe don't have any documentation of their api and it's just assumption from my part

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Indeed, fixed

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Fixed

music_assistant/providers/musicme/__init__.py
JulienDeveaux: Ideed, I've adressed that in the latest commit

OzGav:
There are still mypy errors. You need to run pre-commit before pushing. Also based on a quick glance you still have places where you are raising generic python errors instead of the MA equivalent.

JulienDeveaux:
I've added the french word connexion to the config of the linter as it's part of the login url and flagged
Is that ok ?

marcelveldt-traveling:
> Uses a dedicated HTTP session to isolate auth cookies from other providers

Why? There is no point in that. Please just use the global aiohttp client session

JulienDeveaux:
> > Uses a dedicated HTTP session to isolate auth cookies from other providers
> 
> Why? There is no point in that. Please just use the global aiohttp client session

At that time, I thought it made sense, changed it to the global client

OzGav:
Please resolve all of the copilot comments either by fixing the problem or explaining why copilot is wrong. In either case click the resolve button at the end

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,866 @@
...
+    _http_session: aiohttp.ClientSession | None = None
+    throttler: ThrottlerManager
+
+    async def handle_async_init(self) -> None:
+        """Handle async initialization of the provider."""
+        if not self.config.get_value(CONF_USERNAME) or not self.config.get_value(CONF_PASSWORD):
+            msg = "Missing MusicMe email or password"
+            raise SetupFailedError(msg)
+        self._http_session = aiohttp.ClientSession()
+        self.throttler = ThrottlerManager(rate_limit=1, period=1)
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/constants.py
```diff
@@ -0,0 +1,23 @@
...
+from music_assistant_models.enums import ProviderFeature
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.LIBRARY_PLAYLISTS,
+    ProviderFeature.SEARCH,
+    ProviderFeature.ARTIST_ALBUMS,
+    ProviderFeature.ARTIST_TOPTRACKS,
+    ProviderFeature.BROWSE,
+    ProviderFeature.RECOMMENDATIONS,
+}
```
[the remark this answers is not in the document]
JulienDeveaux: PR descc updated

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,866 @@
...
+
+    # ---- browse ----
+
+    async def browse(self, path: str) -> Sequence[MediaItemType | ItemMapping | BrowseFolder]:
+        """Browse MusicMe content.
+
+        :param path: The path to browse.
+        """
+        if not path or "://" not in path:
+            return [
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/constants.py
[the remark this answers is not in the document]
JulienDeveaux: already fixed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,866 @@
...
+
+    async def handle_async_init(self) -> None:
+        """Handle async initialization of the provider."""
+        if not self.config.get_value(CONF_USERNAME) or not self.config.get_value(CONF_PASSWORD):
+            msg = "Missing MusicMe email or password"
+            raise SetupFailedError(msg)
+        self._http_session = aiohttp.ClientSession()
+        self.throttler = ThrottlerManager(rate_limit=1, period=1)
+        await self._login()
+
```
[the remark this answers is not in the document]
JulienDeveaux: added

music_assistant/providers/musicme/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+
+def decrypt(encrypted: str) -> str:
+    """Decrypt a MusicMe API response or ticket string.
+
+    Reverse-engineered from audioFrame-bundle.min.js (apache_crypto module).
+    Algorithm: strip padding, find marker, rearrange halves, hex-decode XOR 0xAA.
+    """
+    if len(encrypted) < 20:
+        msg = f"Encrypted payload too short ({len(encrypted)} chars)"
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
JulienDeveaux: added

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,852 @@
...
+        elif content_type == "album" and MediaType.ALBUM in media_types:
+            album_data = await self._api_get(f"/album/{content_id}?resources=tracks")
+            if album_data and "item" in album_data:
+                result.albums = [self._parse_album(album_data["item"])]
+                if MediaType.TRACK in media_types:
+                    tracks = album_data.get("results", {}).get("tracks", [])
+                    result.tracks = [
+                        self._parse_track(t)
+                        for t in tracks
+                        if t.get("barcode") and t.get("streamable")
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,852 @@
...
+                albums = res.get("albums", [])
+                if isinstance(albums, list):
+                    result.albums = [self._parse_album(a) for a in albums if a.get("barcode")]
+            if MediaType.TRACK in media_types:
+                tracks = res.get("tracks", [])
+                if isinstance(tracks, list):
+                    result.tracks = [
+                        self._parse_track(t)
+                        for t in tracks
+                        if t.get("barcode") and t.get("streamable")
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,852 @@
...
+                if response.status in (502, 503):
+                    raise ResourceTemporarilyUnavailable(backoff_time=30)
+                if response.status == 404:
+                    return None
+                response.raise_for_status()
+                raw = await response.text()
+        except ClientConnectionError as err:
+            path = endpoint.split("?", maxsplit=1)[0]
+            self.logger.warning("Connection error for %s: %s", path, err)
+            return None
```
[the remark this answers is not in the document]
JulienDeveaux: done

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,852 @@
...
+                content_type=ContentType.AAC,
+                sample_rate=44100,
+                bit_depth=16,
+                channels=2,
+            ),
+            stream_type=StreamType.HTTP,
+            path=stream_url,
+            can_seek=True,
+            allow_seek=True,
+        )
```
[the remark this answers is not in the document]
JulienDeveaux: done

music_assistant/providers/musicme/constants.py
```diff
@@ -0,0 +1,23 @@
...
+import json
+import re
+
+from music_assistant_models.enums import ProviderFeature
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.LIBRARY_PLAYLISTS,
+    ProviderFeature.SEARCH,
+    ProviderFeature.ARTIST_ALBUMS,
+    ProviderFeature.ARTIST_TOPTRACKS,
```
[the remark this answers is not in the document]
JulienDeveaux: PR desc already updated

music_assistant/providers/musicme/__init__.py
OzGav: Sorry I wasn’t clear in what I wrote but even within a query we don’t normally cache search. You could make it short like 15 mins as you want any changes in the provider to show up quickly.

music_assistant/providers/musicme/__init__.py
OzGav: Ok then your reasoning needs to be added as a comment so this isn’t flagged again in the future.

music_assistant/providers/musicme/__init__.py
OzGav: I think you need to test it then and work it out otherwise this will silently not work which I would suggest is worse than just not advertising the feature.

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,853 @@
...
+        try:
+            async with self.mass.http_session.get(url) as response:
+                if response.status == 429:
+                    try:
+                        backoff = min(int(response.headers.get("Retry-After", 10)), 300)
+                    except (ValueError, TypeError):
+                        backoff = 10
+                    raise ResourceTemporarilyUnavailable("Rate limited", backoff_time=backoff)
+                if response.status in (502, 503):
+                    raise ResourceTemporarilyUnavailable(backoff_time=30)
```
OzGav: Keep consistency with the raise above and include a message here

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,853 @@
...
+                    path=cover,
+                    provider=self.instance_id,
+                    remotely_accessible=True,
+                )
+            )
+        return playlist
+
+    # ---- private helpers ----
+
+    async def _web_search_fallback(
```
OzGav, line 712: Why no exception handling in here?

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,853 @@
...
+                data={
+                    "email": email,
+                    "password": password,
+                    "act": "login",
+                    "login_cookie": "1",
+                },
+                allow_redirects=True,
+            ) as resp:
+                resp.raise_for_status()
+        except ClientResponseError as err:
```
OzGav: Probably should also catch ClientConnectioEerror

music_assistant/providers/musicme/provider.py
JulienDeveaux: Indeed, fixed

music_assistant/providers/musicme/provider.py
JulienDeveaux: Added a try catch for errors

music_assistant/providers/musicme/provider.py
JulienDeveaux: Added

music_assistant/providers/musicme/constants.py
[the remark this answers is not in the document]
JulienDeveaux: Already fixed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,859 @@
...
+        return [self._parse_album(a) for a in albums if a.get("barcode")]
+
+    @use_cache(3600 * 3)
+    async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> list[Track]:
+        """Get playlist tracks."""
+        data = await self._api_get(f"/playlist/{prov_playlist_id}?resources=tracks")
+        if not data:
+            return []
+        tracks = data.get("results", {}).get("tracks", [])
+        return [self._parse_track(t) for t in tracks if t.get("barcode")]
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/constants.py
```diff
@@ -0,0 +1,23 @@
...
+    ProviderFeature.BROWSE,
+    ProviderFeature.RECOMMENDATIONS,
+}
+
+DATASERVICE_BASE = "https://dataservice.musicme.com/dataservice/v3"
+STREAM_BASE = "https://stream.hosting-media.net/musicme"
+WEB_BASE = "https://www.musicme.com"
+LOGIN_URL = "https://www.musicme.com/mon-musicme/connexion/"
+PARTNER_ID = "22876"
+CLIENT_JSON = json.dumps({"type": "desktop-web", "context": "www.musicme.com", "app": "mmplayer"})
```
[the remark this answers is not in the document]
JulienDeveaux: fixed

music_assistant/providers/musicme/constants.py
```diff
@@ -0,0 +1,26 @@
+"""Constants for the MusicMe provider."""
+
+import json
+import re
+
+from music_assistant_models.enums import ProviderFeature
+
+SUPPORTED_FEATURES = {
```
MarvinSchenkel, line 8: Just wondering, doesn't MusicMe support adding artists and albums to your library? I only see `LIBRARY_PLAYLISTS` here, but would then also expect `LIBRARY_ARTISTS` and `LIBRARY_ALBUMS`

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,861 @@
...
+                    tracks = album_data.get("results", {}).get("tracks", [])
+                    result.tracks = [
+                        self._parse_track(t)
+                        for t in tracks
+                        if t.get("barcode") and t.get("streamable") == 2
+                    ]
+
+        return result if (result.artists or result.albums or result.tracks) else None
+
+    async def _login(self) -> None:
```
MarvinSchenkel, line 776: Just a check with you. Is this method solely used to verify that a user has a paid subscription, or do you implicitly rely on the cookies it creates? If the latter, then this will break multi-instance support as we have 1 shared mass.session per running MA

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,861 @@
...
+
+        for item in data.get("results", {}).get("theme-airplays", []):
+            if item.get("id"):
+                results.append(self._parse_radio(item))
+        return results
+
+    # ---- recommendations ----
+
+    @use_cache(3600 * 3)
+    async def recommendations(self) -> list[RecommendationFolder]:
```
MarvinSchenkel, line 287: The 4 api calls in this function can be parallelized using `asyncio.gather` to increase the speed of this function

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,861 @@
...
+                    return None
+                raw_bytes = await resp.read()
+                content = raw_bytes.decode("latin-1", errors="replace")
+        except (ClientConnectionError, ClientResponseError) as err:
+            self.logger.debug("Web search fallback failed: %s", err)
+            return None
+
+        result = SearchResults()
+
+        match = re.search(r"window\.playerInit\s*=\s*(\{.*?\});", content, re.DOTALL)
```
MarvinSchenkel, line 739: Shouldn't we use `match = re.search(r"window\.playerInit\s*=\s*(\{.*\});", content)` here to be a bit more defensive?

MarvinSchenkel (changes requested):
Please have a look at my comments. Marking this PR as draft so we can keep track of things that need our attention. When you want us to have another look, mark this PR as 'ready to review' again 🙏

music_assistant/providers/musicme/provider.py
JulienDeveaux: Indeed, it's much better

music_assistant/providers/musicme/constants.py
JulienDeveaux: I don't personally use the library/favorites features on MusicMe, so I'm not confident I could properly test them. The favorites API is only available through web AJAX endpoints (not the dataservice), which makes it more fragile

music_assistant/providers/musicme/provider.py
JulienDeveaux: .*? (non-greedy) is actually the safer choice here, as .* would greedily match past the closing }; if there's another one later in the page. .*? stops at the first match, which is what we want

music_assistant/providers/musicme/provider.py
JulienDeveaux: You're right, the login cookies on the shared session would get overwritten by the second instance's login, which would break the web search fallback for the first one. The dataservice API calls themselves are fine since they rely on the userId query param, not cookies. I haven't found a clean way around this without introducing a dedicated session (which was noticed previously as a bad practice to have multiple http sessions). Open to suggestions if you have ideas

pyproject.toml
```diff
@@ -72,7 +72,8 @@ mass = "music_assistant.__main__:main"
 [tool.codespell]
 # explicit is misspelled in the iTunes API
 # "additionals" is a fixed field name from the Nicovideo API fixtures
-ignore-words-list = "provid,hass,followings,childs,explict,additionals,commitish,nam,"
+# "connexion" is a French word in the MusicMe login URL path
+ignore-words-list = "provid,hass,followings,childs,explict,additionals,commitish,nam,connexion,"
```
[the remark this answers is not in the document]
JulienDeveaux: codespell don't have this kind of inline comment, I already tried before adding that

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,867 @@
...
+        except (ClientResponseError, ClientConnectionError) as err:
+            msg = f"MusicMe login request failed: {err}"
+            raise LoginFailed(msg) from err
+
+        try:
+            async with self.mass.http_session.get(f"{WEB_BASE}/?f=1") as resp:
+                resp.raise_for_status()
+                raw_bytes = await resp.read()
+                content = raw_bytes.decode("latin-1", errors="replace")
+        except (ClientResponseError, ClientConnectionError) as err:
```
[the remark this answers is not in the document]
JulienDeveaux: already responded

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,871 @@
...
+            async with self.mass.http_session.post(
+                LOGIN_URL,
+                data={
+                    "email": email,
+                    "password": password,
+                    "act": "login",
+                    "login_cookie": "1",
+                },
+                allow_redirects=True,
+            ) as resp:
```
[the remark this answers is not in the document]
JulienDeveaux: already answered to reviewer

music_assistant/providers/musicme/provider.py
MarvinSchenkel: I think the only way is to create separate client Sessions. I think the Pandora provider also does this. Could you have a check there and see if we can work with something similar?

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,870 @@
...
+
+        MusicMe has no dedicated top-tracks endpoint, so we collect the first
+        tracks from the artist's most recent albums (sorted by date desc).
+        """
+        data = await self._api_get(f"/artist/{prov_artist_id}?resources=albums{{maxResults:5}}")
+        if not data:
+            return []
+        albums = data.get("results", {}).get("albums", [])
+        top_tracks: list[Track] = []
+        for album_obj in albums:
```
MarvinSchenkel: Could we use asyncio.gather here? Fetching albums sequentially could cause some slowdoen.

MarvinSchenkel (changes requested):

music_assistant/providers/musicme/provider.py
JulienDeveaux: Done, album fetches are now parallelized

music_assistant/providers/musicme/provider.py
JulienDeveaux: Good idea, checked the Pandora provider as you suggested. Now using create_clientsession with a dedicated CookieJar per instance, same pattern

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+            self.logger.debug("Web search fallback failed: %s", err)
+            return None
+
+        result = SearchResults()
+
+        match = re.search(r"window\.playerInit\s*=\s*(\{.*?\});", content, re.DOTALL)
+        if not match:
+            return None
+
+        player_init = json_loads(match.group(1))
```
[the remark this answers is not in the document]
JulienDeveaux: In practice, MusicMe's playerInit is always a single-line minified JSON object without nested braces. We've validated this against real responses. A brace counter would be more defensive but adds complexity for a case we haven't encountered. Happy to revisit if it breaks.

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+            msg = f"MusicMe page load failed: {err}"
+            raise LoginFailed(msg) from err
+
+        match = re.search(r"window\.playerInit\s*=\s*(\{.*?\});", content, re.DOTALL)
+        if not match:
+            msg = "Login failed — could not find playerInit in MusicMe response"
+            raise LoginFailed(msg)
+
+        player_init = json_loads(match.group(1))
+        self._user_id = player_init.get("userId")
```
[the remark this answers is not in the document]
JulienDeveaux: same as above

music_assistant/providers/musicme/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+    first_length = len(test) // 2 - (len(test) // 2) % 2
+    sec_length = len(test) - first_length
+    chars: list[str] = [""] * len(test)
+    for j in range(first_length):
+        chars[j + sec_length] = test[j]
+    for j in range(sec_length):
+        chars[j] = test[first_length + j]
+
+    decrypted_chars: list[int] = []
+    for i in range(len(chars) // 2):
```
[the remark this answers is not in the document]
JulienDeveaux: applyed

tests/providers/musicme/test_streaming.py
```diff
@@ -0,0 +1,166 @@
...
+    manifest = Mock()
+    manifest.domain = "musicme"
+    config = Mock()
+    config.instance_id = "musicme--test123"
+    config.name = "MusicMe Test"
+    config.enabled = True
+    config.get_value.side_effect = lambda key, default=None: {
+        "log_level": "GLOBAL",
+    }.get(key, default)
+    return MusicMeProvider(mass, manifest, config, SUPPORTED_FEATURES)
```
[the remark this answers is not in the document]
JulienDeveaux: applyed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+        return playlist
+
+    # ---- private helpers ----
+
+    async def _web_search_fallback(
+        self,
+        search_query: str,
+        media_types: list[MediaType],
+        limit: int = 10,
+    ) -> SearchResults | None:
```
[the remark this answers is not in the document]
JulienDeveaux: applyed

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+        stream_url = await self._get_stream_url(track_barcode)
+        return StreamDetails(
+            item_id=item_id,
+            provider=self.instance_id,
+            audio_format=AudioFormat(
+                content_type=ContentType.AAC,
+                sample_rate=44100,
+                bit_depth=16,
+                channels=2,
+            ),
```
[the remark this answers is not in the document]
JulienDeveaux: done

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+                    item_id=barcode,
+                    provider_domain=self.domain,
+                    provider_instance=self.instance_id,
+                    available=album_obj.get("streamable", 0) == 2,
+                    audio_format=AudioFormat(
+                        content_type=ContentType.AAC,
+                        sample_rate=44100,
+                        bit_depth=16,
+                    ),
+                    url=f"{WEB_BASE}{album_obj['uri']}" if album_obj.get("uri") else "",
```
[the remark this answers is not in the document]
JulienDeveaux: done, same as above

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+                    item_id=barcode,
+                    provider_domain=self.domain,
+                    provider_instance=self.instance_id,
+                    available=track_obj.get("streamable", 0) == 2,
+                    audio_format=AudioFormat(
+                        content_type=ContentType.AAC,
+                        sample_rate=44100,
+                        bit_depth=16,
+                    ),
+                )
```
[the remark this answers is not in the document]
JulienDeveaux: done, same as above

music_assistant/providers/musicme/provider.py
```diff
@@ -0,0 +1,883 @@
...
+            album_data = await self._api_get(f"/album/{content_id}?resources=tracks")
+            if album_data and "item" in album_data:
+                result.albums = [self._parse_album(album_data["item"])]
+                if MediaType.TRACK in media_types:
+                    tracks = album_data.get("results", {}).get("tracks", [])
+                    result.tracks = [
+                        self._parse_track(t)
+                        for t in tracks
+                        if t.get("barcode") and t.get("streamable") == 2
+                    ]
```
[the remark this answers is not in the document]
JulienDeveaux: limit applied

MarvinSchenkel (approved):
Thanks @JulienDeveaux 🙏 . Another great music provider 🎉

OzGav:
@JulienDeveaux can you raise a PR against the beta branch of our docs following the style of the others. Thanks!

JulienDeveaux:
> @JulienDeveaux can you raise a PR against the beta branch of our docs following the style of the others. Thanks!

Pr opened here https://github.com/music-assistant/music-assistant.io/pull/604 🎉

PR #3394, opened 2026-03-14
Add PAUSE support to sendspin player provider

teancom:
Add PlayerFeature.PAUSE to the supported features and implement pause() which stops the audio stream and sets the group state to PAUSED, preserving current media for resume.

This isn't a functional change, in that using 'stop' as pause has worked just fine. But seeing the little log message* bugged me, and also I know some players *do* visually distinguish between 'paused' and 'stopped', and I figured if we could enable that with a little change, great.

*Player/protocol macbookair does not support pause, using STOP instead

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -165,6 +165,7 @@ class SendspinPlayer(Player):
         self._attr_name = sendspin_client.name
         self._attr_supported_features = {
             PlayerFeature.PLAY_MEDIA,
+            PlayerFeature.PAUSE,
             PlayerFeature.SET_MEMBERS,
             PlayerFeature.VOLUME_SET,
             PlayerFeature.VOLUME_MUTE,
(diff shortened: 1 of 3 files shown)


maximmaxim345:
@teancom Why should we implement `PAUSE` on the provider level for Sendspin?
I mean, Sendspin does not support pausing a stream, only stopping it.
From what I see, this only adds complexity for the provider.

MarvinSchenkel:
Agree here, `PAUSE` is not in the spec so the Sendspin player provider should not implement it. It is then up to MA core to handle this accordingly.

teancom:
My thought (and I'm very open to being wrong) is that the 'spec' in this case _is_ the MA provider interface, which supports PAUSE as a separate state. And while the sendspin audio  protocol doesn't differentiate between stopped/paused, by supporting it in the provider it allows MA to have them as separate states that could be reflected in the UI of a given MA player. And very specifically, one of my back-burner projects is to port a Winamp-skin compatible player I wrote to MA from Roon, and the Winamp UI differentiates between 'stopped' and 'paused'. But! This is *not* a must-have, even for that, so if the consensus is not to bother, I can close this, no harm, no foul. Just giving context as to why I was looking into it at all.

MarvinSchenkel:
No the spec is the 'Sendspin spec' which can be found [here](https://www.sendspin-audio.com/spec/). It's basically tells how clients and servers should interact. MA always needs to implement a player provider according to the 'player provider specs'. If MA wants to do something on top of that, MA needs to counter for it (e.g. by sending a stop instead of a pause).

I think this still allows all your use cases because MA can definitely pause sendspin players, the only thing it will do under the hood is it will stop the stream and then start it from t=30s for example.

teancom:
Got it, I just had it backwards in my head. Thanks for the explanations, everyone. 🙇

PR #3395, opened 2026-03-15
Support id3v2.4 multi value tags

OzGav:
ID3v2.4 supports null-separated lists in text frames like TPE1/TPE2. This adds support for parsing multi-value artist and album artist fields alongside the existing semicolon-delimited fallback.

https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html


music_assistant/helpers/tags.py
```diff
@@ -281,7 +281,13 @@ def album(self) -> str | None:
...
     @property
     def artists(self) -> tuple[str, ...]:
-        """Return track artists."""
+        """Return track artists.
+
+        Vorbis: parser returns artists (plural) for multiple ARTIST fields.
+        ID3: parser returns artists from TXXX:ARTISTS or multi-value TPE1,
+            otherwise artist (singular) split on delimiter here.
+        MP4/APEv2: parser returns artist (singular), split on delimiter here.
+        """
```
marcelveldt-traveling: Are these verbose (AI slop) docstrings really needed ?

marcelveldt-traveling (approved):

music_assistant/helpers/tags.py
OzGav: Its because I have a small brain and can't remember all the path but can remove if you wish. I am probably harking back to my C days when we had almost a comment per line :-)

marcelveldt-traveling: I think comments are fine to clarify but then I would put the comments next to the lines of implementation instead of the docstring. The doscstring should provide clarity to the caller of the method, the implementation details that need clarification is for inline comments

OzGav: That makes sense I will move them down.

MarvinSchenkel (approved):

PR #3397, opened 2026-03-15
Update beta release versioning to use leading zeros

Getslow6:
Hi,

Out of the box tools like Renovate don't identify the new Beta releases correctly, e.g.:

`2.8.0b18 → 2.8.0b9`

To fix this, I've proposed this pull request to make any numbers below 10 have a leading 0. 

I think I've added all the changes needed (and let Claude double check them), but I'm not sure how I could test this.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -2,7 +2,7 @@ name: Auto Release
 
 # Automatically creates releases with proper version increments
 # - Nightly: runs at 02:00 UTC daily if there are 2+ commits (format: 1.2.3.dev20251025HH)
-# - Beta: manual trigger (format: 1.2.0b1, 1.2.0b2, etc.)
+# - Beta: manual trigger (format: 1.2.0b01, 1.2.0b02, etc.)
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
We are using PEP440 for our versioning. Merging this PR would cause inconsistencies so I would prefer not to merge this one if it's not necessary. I think you can tell Renovate which versioning strategy to use when evaluating package versions. Can you see if [this](https://docs.renovatebot.com/modules/versioning/#examples-of-versioning-overrides) works for you?

Getslow6:
Hi Marvin,

I've now overruled the versioning in Renovate to use 'pep440' for Music Assistant indeed, and it works. 

I didn't know it could break other parts of the system if this merges - if that's the case feel free to close this PR. 

I also have to admit it is a minor inconvenience I thought I could quickly fix for everyone else. It is only an 'issue' if a user uses AND the docker container AND is in the beta AND there are more than 9 betas. 

I'll let you guys decide. Great work and really looking forward to the new release. 

Groeten!

MarvinSchenkel:
In that case I will close this PR, as it will also require changes in other areas. Thanks!

PR #3398, opened 2026-03-15
Add Dashie player provider

jwlerch78:
## Summary

Apologies for the incremental review request so quickly. I didn't realize there was a dev server and hadn't thoroughly tested the last release. This one is tested and working well.

This PR adds the **Dashie player provider** for Music Assistant, enabling Android tablets running the [Dashie](https://dashieapp.com) family dashboard app to appear as MA players with full playback control, volume, and metadata display. It also adds several streams server endpoints needed for the integration.

### Provider (`music_assistant/providers/dashie/`)
- Renamed from `dashie_kiosk` to `dashie`
- Poll-based player with 3s interval for state sync
- Pushes track metadata to device via REST API (`setMediaInfo`)
- Supports play/pause, next, previous, stop, volume_set
- Manual IP configuration for devices not auto-discovered via HA
- Response type validation and proper error handling in client
- Debug-level logging for routine operations

### Streams controller additions
- `/player_state/{queue_id}` and `/player_state` (active queue) endpoints
- `/players` endpoint listing available MA players with friendly names
- `/recently_played/{queue_id}` endpoint
- Extended `serve_command_request` with `volume_set`, `previous`, `play`, `pause`, `play_pause`, `stop` commands
- Helper: `_friendly_name()` derives readable names from entity IDs
- Helper: `_get_player_state()` extracted to avoid code duplication
- Elapsed time clamping uses `min()` instead of modulo
- Consistent error handling: all endpoints return `{"error": str(err)}` with status 500

## Test plan
- [x] Dashie tablet discovered and registered as MA player
- [x] Playback controls (play/pause/next/previous/stop) working
- [x] Volume control via `volume_set` command working
- [x] Track metadata displayed on device during playback
- [x] Speaker transfer between devices working
- [x] `/player_state` endpoint returns correct state
- [x] `/players` endpoint lists all available players with friendly names
- [x] Manual IP configuration tested for devices without HA entity


music_assistant/controllers/streams/streams_controller.py
MarvinSchenkel, line 1: Can you explain to me why all these changes are needed? The streamscontroller is meant to handle streaming related actions only. You should not have to worry about this when developing a player plugin and changes to this file should not be necessary.

Could you explain what functionality you require in your player provider so that I can steer you in the right direction? The endpoints you register in the streamscontroller don't even seem to be used in the player provider?

MarvinSchenkel (changes requested):
Please have a look at my comment about the streamscontroller and mark this PR as ready for review when you want us to have another look.

OzGav:
@jwlerch78 Can you confirm that Dashie has the same functionality as Fully Kiosk per the linked table? I need to add this as a new column

OzGav:
@jwlerch78  n addition to the above . I am updating the beta docs in preparation for the 2.8 release. As we now have the merged player protocol functionality I need to add a SETTINGS section to each player provider but I dont have any Dashie devices. Refer to the Cast provider <https://beta.music-assistant.io/player-support/google-cast/> Can you PR the beta branch or give me the text that I can paste in? Thanks!

MarvinSchenkel:
@jwlerch78 the 2.8 release is next Wednesday. It would be great to get this PR over the finish line before then, so people do not have to reconfigure their Dashie providers after the rename.

jwlerch78:
Sorry for the delay.  I was on vacation with the family last week

Thanks for the feedback @MarvinSchenkel — great question. Dashie is different than Fully Kiosk here in that it's both a **player** (receives audio from MA) and has a **display/controller** built-in natively. It'll show music playing information as a pop-up and alongside the screensaver as well as support transferring music between speakers.

Here's what the tablet UI looks like in practice:
dashie_screensaver_overlay

dashie_screensaver_2

dashie_speaker_transfer


These features are why the streams controller endpoints were added — the tablet needs to query MA for player state, player list, and recently played items. The provider itself only handles the player direction (MA → device), but the display/controller direction (device → MA) needs these endpoints.

I agree they don't belong in the streams controller. Should I move them into the provider using `mass.streams.register_dynamic_route()` (same pattern as Sonos, Squeezelite, DLNA, etc.), or would you prefer a different approach for this kind of display/controller use case?

jwlerch78:
@OzGav — two responses:

**Capabilities table:** Dashie matches Fully Kiosk exactly — same Android MediaPlayer under the hood. All ❌ across the board (no hi-res, no lossless, no sync, no crossfade, no stereo pair, no device buttons, no voice control). Max sample rate: 48kHz/16 bits.

**Settings section for beta docs:** The existing page needs to be renamed from "Dashie Kiosk" to "Dashie" (PR #3398 renames the provider). Here's the updated settings text:

---

In addition to the [Individual Player Settings](/settings/individual-player/) the Dashie provider has the following settings:

**Provider Settings:**
- **Dashie devices (via Home Assistant)** — Select Dashie tablets discovered through the Dashie HA integration. Requires the Home Assistant Plugin.
- **Manual Dashie addresses** *(advanced)* — Manually add Dashie tablets by IP address and port (e.g. 192.168.1.100:2323). Use this if you don't have the Dashie HA integration installed.

**Individual Player Settings:**
- **Output codec to use for streaming audio to the player.** The default is `MP3` but other options are `FLAC`, `AAC` or `WAV`. Trying to send lossless to these devices can be problematic.
- **Output channel mode.** The default is `Stereo` but other options are `Left channel only`, `Right channel only` or `Mono (both channels)`.
- **Sample rates supported by this player.** This setting is automatically set upon player discovery but the sample rates and bit depths supported by the player can be manually set. Content with unsupported sample rates will be resampled.
- **HTTP profile used for send audio.** This is considered to be a very advanced setting and should only be adjusted if needed.
- **Try to inject metadata into stream (ICY).** Enabling this option attempts to provide metadata to the player which can be used to show track info, even when flow mode is enabled.

---

Happy to submit a docs PR to the beta branch if you'd prefer that instead.

OzGav:
@jwlerch78 Thanks. I have added this to a broader update I am doing to the docs

MarvinSchenkel:
> Sorry for the delay. I was on vacation with the family last week
> 
> Thanks for the feedback @MarvinSchenkel — great question. Dashie is different than Fully Kiosk here in that it's both a **player** (receives audio from MA) and has a **display/controller** built-in natively. It'll show music playing information as a pop-up and alongside the screensaver as well as support transferring music between speakers.
> 
> Here's what the tablet UI looks like in practice: dashie_screensaver_overlay
> 
> dashie_screensaver_2 dashie_speaker_transfer
> These features are why the streams controller endpoints were added — the tablet needs to query MA for player state, player list, and recently played items. The provider itself only handles the player direction (MA → device), but the display/controller direction (device → MA) needs these endpoints.
> 
> I agree they don't belong in the streams controller. Should I move them into the provider using `mass.streams.register_dynamic_route()` (same pattern as Sonos, Squeezelite, DLNA, etc.), or would you prefer a different approach for this kind of display/controller use case?

Those are two completely different use cases. Let's scope this PR to what we intended to do with this: Add the Dashie player provider. 

Generally, when you want to obtain information from MA (eg player states), you would use the default MA API for this. Users can generate a long lived token and add that in Dashie, so that dashie can query the MA api for this information. We cannot register dynamic routes for this use case, since this is not the intended purpose.

jwlerch78:
Thanks @MarvinSchenkel — that makes sense. I've removed all the streams controller changes and scoped this PR to the player provider only. The display/controller features will use the MA API with a long-lived token separately.

PR is updated and ready for review.

jwlerch78:
Tested and confirmed working — player discovery, playback controls, volume, and metadata push all functioning correctly with the streams controller changes removed.

PR is ready for review. @MarvinSchenkel

music_assistant/providers/dashie/provider.py
```diff
@@ -155,9 +155,23 @@ async def _setup_player(
...
         # Create and register the player
-        player = DashieKioskPlayer(self, entity_id, client, f"{host}:{port}", dev_info)
+        player = DashiePlayer(self, entity_id, client, f"{host}:{port}", dev_info)
         player.set_attributes()
         await self.mass.players.register(player)
+        # Push player ID + server URL immediately so the device knows its identity
+        streams = self.mass.streams
+        ma_server_url = streams.base_url or f"http://{streams.publish_ip}:{streams.publish_port}"
+        try:
+            await client.set_player_id(entity_id, ma_server_url)
```
MarvinSchenkel, line 165: Why is this needed? Players should receive a full urls to any stream when `play_media` is invoked, which is basically this url + the path to the correct stream.

music_assistant/providers/dashie/player.py
```diff
@@ -118,24 +125,91 @@ async def play_media(self, media: PlayerMedia) -> None:
...
+            self.logger.debug("Failed to push media info: %s", err)
+
     async def poll(self) -> None:
         """Poll player for state updates."""
         try:
             async with asyncio.timeout(15):
                 await self.client.get_device_info()
                 self.set_attributes()
+                # Push this device's player ID on first successful connect
+                if not self._player_id_sent:
```
MarvinSchenkel, line 177: Same comment as in setup player, why is this needed? Also, why do we need to do this in every poll?

MarvinSchenkel (changes requested):
Please have a look at my questions

MarvinSchenkel:

MarvinSchenkel:
@jwlerch78 any progress on this one?

jwlerch78:
Closing — Dashie has pivoted its Music Assistant integration to a Sendspin client (the tablet registers itself directly with MA as a player), which supersedes this REST-API provider. The already-merged `dashie_kiosk` provider is being removed in #4192. Thanks for the reviews!

PR #3399, opened 2026-03-15
Fix group volume balance drift and plugin volume feedback loops

theturtle32:
## Summary

- Replace the additive-delta group volume algorithm with a ratio-based model that preserves relative child volume balance across the full 0–100 range
- Isolate plugin volume callbacks to prevent feedback loops with Spotify Connect and similar integrations
- Group volume automatically tracks max(child_volumes), so the group slider always reflects the loudest child and scales everything proportionally

### Balance drift from additive deltas

The existing `set_group_volume` computes a flat `±delta` and applies it to every child player, clamping each to 0–100. Clamping is lossy and irreversible: once a child's volume hits the ceiling or floor, the relative balance between children is permanently destroyed. There is no stored state to restore from.

**Example:** Children A=80, B=40. Raise group by +30 → A clamps to 100 (lost 10), B=70. Lower group by -30 → A=70, B=40. A was 80, now 70 — balance has drifted.

### Plugin volume feedback loop

When Spotify Connect is the active source on a group, every group volume change fires per-child `on_volume` callbacks to Spotify's API, each sending a different volume value. Spotify picks one, echoes it back as a `volume_changed` event, which triggers another group volume change — creating an oscillating feedback loop until Spotify rate-limits (HTTP 429).

### Ratio-based group volume with max-tracking

The new model stores two pieces of state per group:

- **Group volume level** (0–100): always equals the loudest child's volume
- **Child ratios** (0.0–1.0 per child): each child's volume proportion relative to the group level

Child hardware volume is always `group_vol * ratio`. The loudest child always has ratio 1.0. Because ratios are never modified by group volume changes, raising the group until a child clamps at 100 and then lowering it back restores the original balance exactly — the ratio remembers what clamping would otherwise destroy.

The group volume tracks the maximum child volume as an invariant:

- **Overflow**: When a child's volume is set above the current group level, the group volume rises to accommodate it, the child's ratio becomes 1.0, and sibling ratios are rescaled to preserve their hardware output.
- **Deflation**: When the loudest child's volume decreases, the group volume drops to the new max, and all sibling ratios are rescaled accordingly. Hardware volumes don't change for siblings — only the ratio representation and displayed group volume.
- **Normal**: When a non-max child's volume changes within the current group level, only that child's ratio updates.

Sibling ratio rescaling uses pure float math (`old_ratio * old_group_vol / new_max`) rather than re-deriving from integer hardware volumes, avoiding compounding rounding errors across multiple rescaling cycles.

This means the group slider always reflects the loudest child, and "scale everything proportionally" is the consistent semantic in both directions.

### Plugin callback isolation

`set_group_volume` fires the plugin `on_volume` callback once at the group level with the group volume value. Individual child volume changes update ratios silently without notifying plugins. Overflow and deflation (which change the group volume) do notify plugins with the new group level. Protocol players skip plugin and ratio logic entirely during volume command redirects.

Additionally, the Spotify Connect inbound `volume_changed` handler now suppresses processing when the incoming volume matches the last value we sent, preventing self-triggered echoes from re-entering the volume control path.

### Data ownership for sync groups

A new `_resolve_group_data_owner` helper ensures that dynamic sync groups (where a sync leader and its virtual `SyncGroupPlayer` coexist) always route group volume operations to the canonical `SyncGroupPlayer` entity, preventing state divergence.

### Persistence

Static groups (`PlayerType.GROUP`) persist group volume and ratios to player config, surviving restarts. Dynamic sync groups store state in runtime `extra_data` and re-derive ratios when the group re-forms.

## Changes

- **`music_assistant/constants.py`**: Add `ATTR_GROUP_VOLUME_LEVEL` and `ATTR_GROUP_CHILD_RATIOS` constants
- **`music_assistant/controllers/players/controller.py`**: Rewrite `set_group_volume` with ratio-based logic; add helpers for ratio initialization, membership change tracking, overflow/deflation rescale, persistence, and data ownership resolution; restructure `_handle_cmd_volume_set` with `from_group_volume` flag, protocol player guard, and selective plugin callback firing; add `[GroupVolume]` debug logging; extract `_handle_group_members_changed` from `signal_player_state_update`
- **`music_assistant/models/player.py`**: `group_volume` property reads stored level instead of deriving from child volume average; added fallback to owning `SyncGroupPlayer` for sync leaders
- **`music_assistant/providers/spotify_connect/__init__.py`**: Add inbound echo suppression and debug logging to volume handlers
- **`tests/core/test_group_volume.py`** (new): 35 tests covering ratio preservation, overflow rescale, deflation rescale, plugin callback isolation, membership changes, persistence, sync group data ownership, and edge cases

## Test plan

- [x] All pre-commit hooks pass (ruff, mypy, formatting)
- [x] All 35 new tests pass
- [x] All existing tests pass (1073 passed; 3 pre-existing failures in `test_tags.py` unrelated)
- [x] Manual: Spotify Connect → static group, child volume changes don't trigger Spotify feedback
- [x] Manual: Spotify Connect → static group, group volume changes notify Spotify once
- [x] Manual: overflow rescale raises group volume and notifies Spotify correctly
- [x] Manual: dynamic sync group volume routes through canonical SyncGroupPlayer
- [x] Manual: deflation lowers group volume when loudest child decreases
- [x] Manual: group slider scales all children proportionally in both directions
- [x] Manual: restart server, verify persisted ratios restored for static groups
- [x] Manual: add/remove child from active group, verify ratio derived correctly


marcelveldt-traveling (changes requested):
I think the PR is solving a real edge case, but it is doing it with a lot more machinery than the problem seems to require.

The current issue appears to be that group volume is derived from the live child volumes, and group volume changes are then applied as a delta to each child. That works until one or more children hit 0/100 and get clamped, at which point the original balance can drift. So the bug itself looks legitimate.

That said, the proposed fix feels too complex for the scope of the problem. It introduces persistent ratio state, canonical owner resolution, sync-group indirection, and plugin feedback-loop handling all in one change. Those are separate concerns, and combining them makes the volume model much harder to reason about. It also seems to move group volume away from being derived from the actual current child state and toward a stored internal model, which risks getting out of sync with provider-driven or hardware-originated child volume changes.

I think a simpler model would be better:

Pick one canonical place where group volume state lives:
dedicated group player if one exists, otherwise the sync leader for ad-hoc/manual sync.
When group volume changes, calculate the relative change for the child players from that one owner.
When an individual child volume changes, update that child’s relative trim/balance against the current group volume.
Keep plugin/source feedback-loop prevention separate from the group-volume algorithm itself.
In other words, I would avoid introducing persisted ratio state unless there is a very strong reason for it. The main problem here seems to be preserving relative child balance when changing group volume, and that should be solvable with a much smaller abstraction around a single group-volume owner plus per-child relative adjustments.

So overall: yes, I think your hunch is correct. The issue is probably real, but the solution in this PR looks over-engineered and duplicates logic that should be centralized around a single “group volume owner” concept.

theturtle32:
@marcelveldt-traveling Hey, thanks for taking a look at this! I appreciate you taking the time. I also apologize, as I meant to just open this draft PR to keep track of where I'm at with this and what's been done and begin to prepare for submission. However, I fully agree that this PR is not ready for prime time, which is why I've left it in draft status for now.

I'm currently running my own personal installation of Music Assistant off of this branch to see concretely how it operates and if it introduces any unexpected problems.

Also, of note, the plugin volume change feedback loops are also solved on this branch, and that is a FAR more pressing issue, as it essentially means that plugins like Spotify Connect are unusable with sync groups outside of this branch: when you change the volume of a sync group member without the fix on this branch, it triggers a recalculation of the sync group volume, which is sent to the Spotify Connect provider, which sends it to the Spotify API to update the remote state, which then triggers a volume change event, which then triggers recalculation of the group member volumes, which then triggers the overall sync group volume change event, which sends the update back to the spotify API, etc., etc., in an endless loop until the Spotify API starts rate limiting. Meanwhile the volumes of the individual group members jump around wildly and uncontrollably until the Spotify API rate limits break the cycle.

I believe it is unavoidably necessary to keep separate state of what the relative sync group member volumes should be relative to the sync group's overall volume, and in most circumstances, changing an individual sync group member volume should NOT reflect as a change to the sync group's overall volume level. I believe that the conceptualization of the primary sync group volume as being a derived, calculated value based on the individual member volumes is fundamentally flawed.

theturtle32:
### Update: max-tracking group volume and revised thinking on derived group volume

After further iteration and real-world testing, the branch now includes **max-tracking**: the group volume always equals `max(child_volumes)`. When the loudest child's volume decreases, the group volume deflates to the new max and all sibling ratios are rescaled (in float math to avoid compounding rounding errors from integer volumes). When a child's volume exceeds the group level, the group volume rises and sibling ratios are rescaled downward. The result is that the group slider always reflects the loudest child, and moving it in either direction scales all children proportionally — which feels like the most intuitive behavior for end users.

I want to walk back something I said in my earlier comment: that the group volume being derived from child volumes is a "fundamentally flawed concept." That statement came from a particular mental model — one rooted in analog signal flow, where the group volume acts like a potentiometer that attenuates the input signal before it reaches the individual player volume controls, which then further attenuate before the final amplification stage. In that model, the group volume is an upstream "source" control, and deriving it from downstream output levels doesn't make sense. That's an intuitive conceptualization if you have audio engineering experience.

But Music Assistant is targeting a general consumer audience, not audio engineers. For most users, the group volume slider is just "the overall volume for all my speakers." The intuitive expectation is: it shows something representative of how loud the group is, and moving it makes everything louder or quieter together, preserving the relative balance. Having the slider track the loudest child satisfies that intuition — you can see at a glance how loud the group is, and the slider behaves symmetrically in both directions.

So rather than the group volume being a fixed upstream control that only changes on explicit user action, it now floats with the max child volume while preserving ratios, which turns out to feel the most natural in practice. The key properties are maintained: ratios are stable (only change on individual child adjustments), balance survives clamping at 0/100, and plugin callbacks only fire when the group volume actually changes.

As far as group volume intuition goes from a user's perspective, the edge case this approach doesn't help with is if you have a few player volumes set at a low level, say 20%, but one player in the group is at 100%, you cannot use the group volume slider to raise the other player volumes. You have to raise them individually. Some people may find this surprising, unexpected, or frustrating. Personally, I find it much more frustrating to lose my carefully set relative volumes every time I manipulate the group volume.

The *advantage* of the max-tracking approach implemented here is that you can dial in the relative volumes once the way you want them, and then just use the group volume to adjust up and down within the ranges that you configured without worrying about losing that balance every time you move the slider.

Thoughts/ideas for improvement welcome.

theturtle32:
Also, I was thinking it might be useful to extract the fix for the group volume plugin feedback loop bug (which effectively effectively makes plugins like Spotify Connect entirely unusable with sync groups) into its own PR separate from the discussion of intuitive group volume change algorithms. What do you think?

theturtle32:
I decided to close this draft PR for now and move my work entirely to my own fork to avoid creating unnecessary noise and distraction for the maintainers until I have a clearer proposal.

marcelveldt:
I think I found a way to fix issue number 1 (the volume balance drift) in this PR: https://github.com/music-assistant/server/pull/3548

As for the feedback loops of volume I need to have a better look and reproduction scenario.

PR #3400, opened 2026-03-16
Emby music provider fix album track order

hatharry:


MarvinSchenkel (approved):
That was fast! Thanks @hatharry 👏

Hans-Haefner:
Thank you as well @hatharry 
As soon as the beta is available, I will test it and let you know.

PR #3401, opened 2026-03-16
Support for SoundCloud AAC HQ streams (Go+)

mgranell:
*Not Ready to Merge*

This is initial support for high quality streams (up to 256 kb/s), available to Go+ subscribers.

Issue:
Earlier builds of ffmpeg (e.g. ffmpeg 4) are extremely slow to open progressive m4a aac streams for soundcloud, which can be 10 seconds or more. Using a latest source code compile of ffmpeg has lowest latency with HLS (starting in a couple of seconds), and slightly longer for progressive.

I have added an initial configuration option to enable the AAC support, as well as a configuration option to prefer HLS to progressive for testing purposes in different environments.

Welcome any comments or discussion on the PR.

Thanks

### DIFF
--- a/music_assistant/helpers/ffmpeg.py
+++ b/music_assistant/helpers/ffmpeg.py
@@ -455,12 +455,13 @@ async def check_ffmpeg_version() -> None:
     # use globals as in-memory cache
     await set_global_cache_values({CACHE_ATTR_LIBSOXR_PRESENT: libsoxr_support})
 
-    major_version = int("".join(char for char in version.split(".")[0] if not char.isalpha()))
-    if major_version < MINIMAL_FFMPEG_VERSION:
-        raise AudioError(
-            f"FFmpeg version {version} is not supported. "
-            f"Minimal version required is {MINIMAL_FFMPEG_VERSION}."
-        )
+    if version != "N":  # N-versions are development versions, so we assume they are recent enough
+        major_version = int("".join(char for char in version.split(".")[0] if not char.isalpha()))
+        if major_version < MINIMAL_FFMPEG_VERSION:
+            raise AudioError(
+                f"FFmpeg version {version} is not supported. "
+                f"Minimal version required is {MINIMAL_FFMPEG_VERSION}."
+            )
 
     LOGGER.info(
         "Detected ffmpeg version %s %s",
--- a/music_assistant/providers/soundcloud/__init__.py
+++ b/music_assistant/providers/soundcloud/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import time
+from collections.abc import Generator
 from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import parse_qs, urlparse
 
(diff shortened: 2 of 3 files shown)


music_assistant/helpers/ffmpeg.py
```diff
@@ -455,12 +455,13 @@ async def check_ffmpeg_version() -> None:
     # use globals as in-memory cache
     await set_global_cache_values({CACHE_ATTR_LIBSOXR_PRESENT: libsoxr_support})
 
-    major_version = int("".join(char for char in version.split(".")[0] if not char.isalpha()))
-    if major_version < MINIMAL_FFMPEG_VERSION:
-        raise AudioError(
-            f"FFmpeg version {version} is not supported. "
-            f"Minimal version required is {MINIMAL_FFMPEG_VERSION}."
-        )
+    if version != "N":  # N-versions are development versions, so we assume they are recent enough
```
OzGav, line 458: What is this about? MA already builds with ffmpeg 7. The version check exists for a reason — bypassing it for 'N-versions' with no lower bound is unsafe. If you need a dev build, it should still be a recent ffmpeg 7+ dev build, and the version string parsing should handle that properly rather than skipping the check entirely

music_assistant/providers/soundcloud/__init__.py
```diff
@@ -97,6 +105,30 @@ async def get_config_entries(
             label="Authorization",
             required=True,
         ),
+        ConfigEntry(
```
OzGav, line 108: Regarding the two experimental config entries — your own descriptions label these as experimental, which suggests this may not be ready for users yet. Can you explain the reasoning behind exposing both options? Ideally we'd want to settle on the best-performing option as a default and not expose the protocol preference to users at all. Is there a reason that's not possible here?

music_assistant/providers/soundcloud/__init__.py
```diff
@@ -408,70 +440,153 @@ async def get_similar_tracks(self, prov_track_id: str, limit: int = 25) -> list[
...
+
+                    yield (
+                        transcoding,
+                        stream_type,
+                        AudioFormat(
+                            content_type=ContentType.M4A
+                            if content_type == ContentType.AAC
+                            else ContentType.MP3
+                            if content_type == ContentType.MP3
+                            else ContentType.OGG,
```
OzGav, line 524: Is OGG the only remaining format SoundCloud delivers, and if so can you add a comment explaining that? Because without that context it looks like a bug.

music_assistant/providers/soundcloud/__init__.py
```diff
@@ -408,70 +440,153 @@ async def get_similar_tracks(self, prov_track_id: str, limit: int = 25) -> list[
...
+                        else ContentType.AAC
+                        if mime_type.startswith("audio/mp4")
+                        else ContentType.UNKNOWN
+                    )
+                    stream_type = (
+                        StreamType.HTTP
+                        if protocol == "progressive"
+                        else StreamType.HLS
+                        if protocol == "hls"
+                        else StreamType.UNKNOWN
```
OzGav, line 513: Can protocol ever be anything other than progressive or hls? If not, the UNKNOWN branch is dead code and should be removed. If it can, what should MA actually do in that case?

OzGav:
Also there are conflicts to be resolved

MarvinSchenkel:
@mgranell any progress on this one?

OzGav:
@mgranell do you intend to progress this? It is OK if not but maybe close the PR if that is the case?

MarvinSchenkel:
Closing this PR for now. Feel free to re-open when and if you feel like bringing this to our attention again 🙏

PR #3402, opened 2026-03-16
Fix PyAV version detection when building the base image Dockerfile

maximmaxim345:
Dynamic resolution via pip dry-run was broken and completely failed once PyAV 17.0.0 was released.
We now read from the Sendspin provider manifest so building the base image will always install the correct pinned version.


MarvinSchenkel (approved):
Thanks @maximmaxim345 👏

PR #3416, opened 2026-03-18
Detect stale streams after server restart in player queue controller

sfortis:
After MA restarts, some players (e.g. DLNA devices) keep reporting PLAYING with a flow URL from the previous session. The stream is dead but the player has no way to know.

This adds session ID comparison in _parse_player_current_item_id(): if the URL's session ID doesn't match the current queue session, it triggers a delayed resume so playback continues seamlessly with a fresh stream. Uses create_task(task_id=...) for deduplication so repeated poll cycles don't spawn multiple resume attempts.

Works at the controller level so all player providers benefit, not just DLNA.

Related: #3376 (review comment by @MarvinSchenkel suggesting controller-level implementation)

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -2725,8 +2725,6 @@ class PlayerQueuesController(CoreController):
             return
         # wait briefly for the player to settle after detection
         await asyncio.sleep(3)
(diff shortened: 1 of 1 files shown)


sfortis:
This needs more investigation, I'm closing this.

PR #3418, opened 2026-03-18
Only show non empty genres for media type in library views

jozefKruszynski:
Backend filtering logic to allow library view genre filters to only display non empty genres per media type

### DIFF
--- a/music_assistant/controllers/media/genres.py
+++ b/music_assistant/controllers/media/genres.py
@@ -308,13 +308,12 @@ class GenreController(MediaControllerBase[Genre]):
         """Get genres in the library.
 
         :param genre: NOT SUPPORTED - Filtering genres by genres doesn't make sense.
-        :param hide_empty: Controls which genres are returned when media_type is not set.
-            True: only return genres that have media mappings.
+        :param hide_empty: Only applies when media_type is not set.
+            True: only return genres that have at least one media mapping.
             False: return all genres including unmapped ones.
             None (default): only return default genres (those with a translation_key).
-        :param media_type: When set, return ALL genres (including non-defaults) that have
-            at least one media item of this type mapped. Overrides hide_empty completely —
-            this is intended for library views that need type-scoped genre filters.
+        :param media_type: When set, return all genres (including non-defaults) that have
+            at least one mapping for this media type. Takes precedence over hide_empty.
         """
         if genre is not None:
             msg = "genre parameter is not supported for Genre.library_items()"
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/media/genres.py
```diff
@@ -645,11 +659,16 @@ async def _bulk_scan_media_genres(self) -> None:
...
+        # rows with malformed metadata produce no rows rather than raising an error.
+        # A WHERE-only json_valid guard is insufficient because SQLite evaluates the
+        # FROM clause before the WHERE filter.
         union_parts = [
             f"SELECT DISTINCT TRIM(g.value) AS raw_name "
-            f"FROM {table}, json_each(json_extract({table}.metadata, '$.genres')) AS g "
-            f"WHERE json_extract({table}.metadata, '$.genres') IS NOT NULL "
-            f"AND json_extract({table}.metadata, '$.genres') != '[]'"
+            f"FROM {table}, "
+            f"json_each(CASE WHEN json_valid({table}.metadata) "
```
MarvinSchenkel: Why would we have invalid json in the metdata field?

music_assistant/controllers/music.py
```diff
@@ -2981,6 +2981,10 @@ async def __create_database_indexes(self) -> None:
             f"CREATE INDEX IF NOT EXISTS {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}_genre_alias_idx "
             f"on {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}(genre_id,alias);"
         )
+        await self.database.execute(
+            f"CREATE INDEX IF NOT EXISTS {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}_genre_type_idx "
+            f"on {DB_TABLE_GENRE_MEDIA_ITEM_MAPPING}(genre_id,media_type);"
```
MarvinSchenkel: If this is meant to quickly get 'all genres for media_type xxx', the order should be swapped. Otherwise, I think we can drop this as it is already partly covered by the unqiue constraint

jozefKruszynski: I'll drop it, I was too eager to follow the suggestions of copilot

music_assistant/controllers/media/genres.py
jozefKruszynski: Very good question, as below, I was too eager to follow the suggestions of copilot, who was trying to write overly defensive code.

music_assistant/controllers/media/genres.py
```diff
@@ -644,12 +657,11 @@ async def _bulk_scan_media_genres(self) -> None:
...
         alias_to_genre, primary_name_to_genre = await self._build_genre_lookup()
 
-        # Extract all unique raw genre names from metadata across all media tables
         union_parts = [
             f"SELECT DISTINCT TRIM(g.value) AS raw_name "
-            f"FROM {table}, json_each(json_extract({table}.metadata, '$.genres')) AS g "
-            f"WHERE json_extract({table}.metadata, '$.genres') IS NOT NULL "
-            f"AND json_extract({table}.metadata, '$.genres') != '[]'"
+            f"FROM {table}, "
+            f"json_each(json_extract({table}.metadata, '$.genres')) AS g "
```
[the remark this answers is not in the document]
jozefKruszynski: Not needed as previously discussed.

tests/core/test_genres.py
```diff
@@ -288,6 +288,86 @@ async def test_library_items_rejects_genre_param(self, genre_ctrl: GenreControll
...
+        search_names = {g.name for g in search_results}
+        assert "MT_FilterSharedGenre" in search_names
+        assert "MT_FilterTrackOnlyGenre" not in search_names
+
+        # No mappings for the requested type returns an empty list
+        playlist_results = await genre_ctrl.library_items(media_type=MediaType.PLAYLIST)
+        playlist_names = {g.name for g in playlist_results}
+        assert "MT_FilterTrackOnlyGenre" not in playlist_names
+        assert "MT_FilterAlbumOnlyGenre" not in playlist_names
+        assert "MT_FilterSharedGenre" not in playlist_names
```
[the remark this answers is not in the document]
jozefKruszynski: Copilot has a point technically — the test asserts three specific genres are absent rather than assert playlist_results == []. However, the test uses a shared genre_ctrl fixture with DB state from earlier in the test class, so other genres legitimately exist in the DB.
The three assertions are the meaningful contract check — that none of the mapped genres leak through. Asserting == [] would be brittle and wrong given the shared state. Ignore.

MarvinSchenkel (approved):
THanks @jozefKruszynski

PR #3419, opened 2026-03-18
Support playback of radio station PLS playlist URLs with query parameters

OzGav:
Radio station URLs like `http://www.radiofeeds.net/playlists/bauerflash.pls?station=absoluteclassicrockhigh-aac` use query parameters to determine which station's stream URL is returned in the PLS file. These URLs were failing to be added because ffprobe cannot analyze PLS files directly, it sees them as text/subtitle content rather than audio.

This PR adds `_resolve_url()` to the builtin provider that detects playlist URLs and extracts the actual audio stream URL before passing to ffprobe. The original playlist URL is preserved as the `item_id`, ensuring fresh tokens are fetched at playback time via the existing `resolve_radio_stream()` flow.


music_assistant/providers/builtin/__init__.py
```diff
@@ -536,6 +538,31 @@ async def resolve_image(self, path: str) -> str | bytes:
...
+
+        ffprobe cannot analyze PLS/M3U files directly as it sees them as text.
+        This method extracts the actual audio stream URL from playlist files.
+
+        :param url: The URL to check and potentially resolve.
+        :returns: The resolved stream URL, or the original URL if not a playlist.
+        """
+        parsed = urlparse(url)
+        path_lower = parsed.path.lower()
+        is_playlist = path_lower.endswith(".pls")
```
marcelveldt, line 555: The docstring mentions m3u and pls, yet the check is only checking pls

music_assistant/providers/builtin/__init__.py
```diff
@@ -544,8 +571,9 @@ async def _get_media_info(self, url: str, force_refresh: bool = False) -> AudioT
         )
         if cached_info and not force_refresh:
             return AudioTags.parse(cached_info)
+        resolved_url = await self._resolve_playlist_url(url)
```
marcelveldt: this looks weird, the helper function should probably be renamed to "_resolve_url" ?

music_assistant/providers/builtin/__init__.py
OzGav: Yeah missed that. Marvin and I chatted and decided to switch to pls only for this PR.

music_assistant/providers/builtin/__init__.py
OzGav: Renamed

marcelveldt (approved):

PR #3421, opened 2026-03-18
Fix timeouts and slowdowns on play_media action for HEOS 

Tommatheussen:
In earlier HEOS queue-handling PRs (https://github.com/music-assistant/server/pull/3219, https://github.com/music-assistant/server/pull/3295, https://github.com/music-assistant/server/pull/3358), we fixed the issue where “ghost” items could keep playing after Music Assistant finished its queue. Those ghost items were leftover entries in the internal HEOS queue, so once the MA stream ended, HEOS could continue with the next stale queue item. We also added a guard to avoid clearing an already empty HEOS queue, because HEOS does not handle that case gracefully.

That previous fix solved the functional problem, but it exposed a second issue: queue inspection and cleanup can be very slow on HEOS. In particular, `player/get_queue` can enter a command under process state and only return its final response about 20 seconds later. Because pyheos serializes commands on a single connection, those slow queue operations can block other HEOS commands and make `play_media` appear stalled for a fixed ~20 second interval.

This PR moves the queue cleanup from `play_media` to `queue_changed` events coming from HEOS.
The cleanup now happens in two stages:
- A `queue_changed` event marks that the HEOS queue needs cleanup.
- The actual cleanup is only triggered once the player state has reached PLAYING.

The two-stage flow is needed because sometimes `queue_changed` arrives too early to treat the HEOS queue as stable. During a track transition, HEOS can briefly report queue changes while the player is still in STOP and before the new stream has become the active queue item. Cleaning the queue at that moment can remove the newly queued MA stream instead of the stale leftover item. By first marking cleanup as pending and then waiting until HEOS reports PLAYING, the provider avoids that race and only clears the queue once playback has settled far enough to identify the current queue item correctly.

I have also increased the default timeout to 25 seconds. It might be worth to just remove the timeout property again, I have never seen the commands take longer than 20 seconds (it's either basically instant or 20 second delay, doesn't matter if it's wired or wireless, wireless just has more cases where it's 20 seconds)

If the `get_queue` command is running, all other player events are queued until this command is done, that's a current limitation of `pyheos`, will check with the author if there's some way around that.


music_assistant/providers/heos/player.py
```diff
@@ -283,26 +303,83 @@ async def play_media(self, media: PlayerMedia) -> None:
...
+            1,
+            self._start_queue_cleanup_task,
+            task_id=f"heos_queue_cleanup_timer_{self.player_id}",
+        )
+
+    def _start_queue_cleanup_task(self) -> None:
+        """Start the queue cleanup task if not already running."""
+        if not self._ma_controls_playback or not self._queue_cleanup_pending:
+            return
+        self.mass.create_task(
```
MarvinSchenkel, line 339: Shouldn't we add `abort_existing=True` here to prevent creating multiple cleanup Tasks with the same id? Or we could check if a task is already running using something like `if self._queue_cleanup_lock.locked()`?

Tommatheussen: The tricky bit is we can't just abort the task, as the `get_queue` command is already send to the system, and we're waiting for the response. pyheos (or rather HEOS itself) doesn't report any request IDs, other than the command itself, so cancelling the task won't do anything other than sending another queue command that might take another 20 seconds.

Doesn't the `create_task` internally already track if a task is running and return the existing task, thus preventing a new one from starting?

MarvinSchenkel (approved):
Danku! @Tommatheussen 👍

PR #3424, opened 2026-03-19
Open Subsonic Lyric support

khers:
The Open Subsonic spec has lyrics support and we have users who would like to see it enabled in the provider. This _should_ work but I don't use lyrics and have not tested it. We need to ask someone who does to test it in beta!


Schnurzel700:
When will this be merged into the main branch? I want to use it in my Home Assistant Music Assistant instance :)

khers:
I suspect our reviewers are taking a moment after the big push for the 2.8 release. They'll get to the review when they can ;)

marcelveldt (approved):
Nice addition, @khers !

PR #3428, opened 2026-03-19
Fix sync group members not hiding in UI

maximmaxim345:
`__final_active_group` checked the raw `group_members` property of sync group players, which returns untranslated protocol-level IDs. Universal players were never matched, so `active_group` stayed `None` and the UI showed them as separate playing group leaders instead of hiding them behind the sync group.

Using `state.group_members` (the finalized, translated list) fixes the ID matching.


marcelveldt-traveling (approved):

MarvinSchenkel (approved):
🎉

PR #3429, opened 2026-03-19
Add YuTorah provider for streaming Torah shiurim

russlevy:
Adds a new music provider for [YuTorah Online](https://yutorah.org/), streaming over 400,000 Torah audio lectures via the official mobile app API.

**Features:**
* Browse by series, teacher, or topic category
* Search returns series (Podcasts), teachers (Artists), and individual shiurim (Tracks)
* Login optional — unauthenticated users get a preview; authenticated users get full paginated episode lists and search
* Breadcrumbs use human-readable name slugs (not raw numeric IDs)
* Backward-compatible with any previously saved browse paths

**Data model:**
* Series → Podcast / PodcastEpisode
* Teacher → Artist / Track (top tracks)
* Individual shiur → Track (search results)


russlevy:
Label should be `new-provider` (I cannot add it)

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1307 @@
...
+    # -----------------------------------------------------------------------
+    # Streaming
+    # -----------------------------------------------------------------------
+
+    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Return stream details for a shiur.
+
+        We first check if the URL was cached in the ProviderMapping.details
+        field (set when building the episode), and only fall back to an API
+        call if it's missing.
```
teancom: The docstring says "We first check if the URL was cached in the ProviderMapping.details field... and only fall back to an API call if it's missing." But the implementation (line 698) always calls _api_get("shiur/details", ...) unconditionally — it never reads from ProviderMapping.details.

teancom (changes requested):

music_assistant/providers/yutorah/__init__.py
russlevy: yah that's the problem with docstrings! I forgot to update it after testing and needing it there. the issue was that get_stream_details only gets the item_id and media_type, so we needed to call the API to get the rest of the info. f0afe67014b9392235bd800742f15d0bcce55951 fixes the docstring

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.BROWSE,
+    ProviderFeature.SEARCH,
+    ProviderFeature.LIBRARY_PODCASTS,
+    ProviderFeature.ARTIST_TOPTRACKS,
+}
+
+CONF_EMAIL = "email"
+CONF_PASSWORD = "password"
```
OzGav: Should be using the MA constants here

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+    # Field names differ between shiur/details (YuShiur) and search/get (YuSearchDoc)
+    shiur_id = str(shiur.get("shiurID") or shiur.get("shiurid") or "")
+    if not shiur_id:
+        return None
+
+    mp3_url = shiur.get("shiurFileURL") or shiur.get("shiurdownloadurl") or ""
+    media_type = shiur.get("shiurMediaType") or shiur.get("mediatypename") or ""
+
+    # Only handle audio (MP3); skip video/PDF/HTML
+    if media_type.upper() not in ("MP3", "AUDIO", "") and not mp3_url:
```
OzGav: Is this right? If the type is not known allow it?

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+        },
+        metadata=MediaItemMetadata(
+            description=description or None,
+            images=UniqueList(images) if images else None,
+            links=links or None,
+        ),
+    )
+
+
+def _shiur_to_track(
```
OzGav: You already have _shiur_to_episode which maps a shiur to a PodcastEpisode. So why does _shiur_to_track exist right after it mapping the same data to a Track? A shiur can't be both.

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+                resp.raise_for_status()
+                data = await resp.json(content_type=None)
+            if data and data.get("loginSuccess") and data.get("userToken"):
+                self._user_token = data["userToken"]
+                self.logger.info("YuTorah login successful — full episode access enabled.")
+            else:
+                self.logger.warning(
+                    "YuTorah login failed (bad credentials?). Running in anonymous mode."
+                )
+        except aiohttp.ClientError as exc:
```
OzGav: You should be raising LoginFailed or AuthenticationFailed as appropriate here.

Generally I think the error handling in this function should be reviewed.

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+            self.logger.warning("YuTorah login error: %s. Running in anonymous mode.", exc)
+
+    async def get_library_podcasts(self) -> AsyncGenerator[Podcast, None]:
```
OzGav: This should be deleted

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+                },
+            )
+
+        series_list = await self._fetch_series_list()
+        for series in series_list:
+            sid = str(series.get("ID") or series.get("seriesID") or "")
+            if sid == str(prov_podcast_id):
+                return _series_to_podcast(series, self.instance_id)
+
+        self.logger.warning("Series %s not found in browse list", prov_podcast_id)
```
OzGav: Why are you doing this instead of raising MediaNotfoundError

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+        episode = _shiur_to_episode(data, 0, self.instance_id)
+        if not episode:
+            raise MediaNotFoundError(f"YuTorah shiur {prov_episode_id} has no playable audio")
+        return episode
+
+    # -----------------------------------------------------------------------
+    # Artist — teachers
+    # -----------------------------------------------------------------------
+
+    async def get_artist(self, prov_artist_id: str) -> Artist:
```
OzGav: I don't think modelling a teacher as an Artist is going to work as the Artist/Track model carries a lot of baggage that doesn't fit — tracks have albums, artists have discographies, MA will try to match them against music metadata providers, normalise them, look up MusicBrainz IDs etc. None of that makes sense for a lecture. Browsing by teacher can be handled entirely through the browse tree and search returning PodcastEpisode results, which is what the content actually is.

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+        p2 = parts[2] if len(parts) > 2 else ""
+        p3 = parts[3] if len(parts) > 3 else ""
+
+        if section == "":
+            return [
+                BrowseFolder(
+                    item_id="series",
+                    provider=self.instance_id,
+                    path=f"{self.domain}://series",
+                    name="Browse by Series",
```
OzGav: Should use translation keys for these

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+                None,
+            )
+            return ([series_item] if series_item else []) + teacher_folders
+
+        if section == "teachers" and not p1:
+            return await self._browse_all_teachers()
+        if section == "teachers" and p1:
+            teacher_id = await self._resolve_teacher_id(p1)
+            return await self._browse_teacher_episodes(teacher_id)
+        # Legacy path: keep backward compat for any saved/bookmarked yutorah://teacher/{id} paths
```
OzGav: When you say 'legacy path' and 'backward compat' — where have the  users saved paths from, or did you change the path structure during development? If the latter, these can just be removed

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+            for raw in docs:
+                episode = _shiur_to_episode(raw, len(episodes), self.instance_id, parent_series_id)
+                if episode:
+                    episodes.append(episode)
+            if len(docs) < PAGE_SIZE:
+                break
+            page += 1
+        return episodes
+
+    async def _api_get(self, endpoint: str, **params: Any) -> Any:
```
OzGav: This is logging but not raising any errors?

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+    The wrapper structure is not fully documented; we try common shapes.
+    """
+    if not data:
+        return []
+    if isinstance(data, list):
+        return data
+
+    if isinstance(data, dict):
+        # Try common wrapper keys
+        for key in ("results", "docs", "shiurim", "items", "data"):
```
OzGav: You mention in the module docstring that the API was discovered via APK reverse engineering, so you must have seen real API responses. What does search/get actually return? This loop trying 5 different key names suggests the response structure isn't known, but it should be

music_assistant/providers/yutorah/manifest.json
```diff
@@ -0,0 +1,31 @@
+{
+    "type": "music",
+    "domain": "yutorah",
+    "name": "YuTorah",
+    "description": "Stream over 400,000 Torah shiurim (audio lectures) from YUTorah Online as podcasts. Browse by series (e.g. Daf Yomi, Ten Minute Halacha), search by topic or teacher, and subscribe to any series.",
+    "stage": "beta",
+    "icon": "podcast",
+    "codeowners": ["@russlevy"],
+    "requirements": [],
+    "documentation": "https://github.com/music-assistant/server/discussions/",
```
OzGav:
```suggestion
    "documentation": "https://music-assistant.io/music-providers/yutorah",
```

music_assistant/providers/yutorah/manifest.json
```diff
@@ -0,0 +1,31 @@
...
+    "multi_instance": false,
+    "config_entries": [
```
OzGav, line 10: Remove this hallucination

music_assistant/providers/yutorah/manifest.json
```diff
@@ -0,0 +1,31 @@
+{
+    "type": "music",
+    "domain": "yutorah",
+    "name": "YuTorah",
+    "description": "Stream over 400,000 Torah shiurim (audio lectures) from YUTorah Online as podcasts. Browse by series (e.g. Daf Yomi, Ten Minute Halacha), search by topic or teacher, and subscribe to any series.",
+    "stage": "beta",
+    "icon": "podcast",
```
OzGav: Delete this. You need to supply colour and monochrome SVG icons

music_assistant/providers/yutorah/__init__.py
```diff
@@ -0,0 +1,1305 @@
...
+    # -----------------------------------------------------------------------
+
+    async def _browse_all_series(self) -> list[BrowseFolder]:
+        """Return all series as browse folders (each expands to teacher sub-folders)."""
+        series_list = await self._fetch_series_list()
+        folders = [
+            BrowseFolder(
+                item_id=str(s.get("ID") or s.get("seriesID") or ""),
+                provider=self.instance_id,
+                path=f"{self.domain}://series/{_slugify(s.get('name') or 'unknown')}",
```
OzGav: Other MA providers put the numeric ID directly in the browse path, e.g. yutorah://series/12345, and put the display name only in BrowseFolder.name. You're putting the name slug in the path and then having to fetch the entire series list to reverse-lookup the ID every time MA navigates to it. Can you use the numeric ID in the path instead?

music_assistant/providers/yutorah/manifest.json
```diff
@@ -0,0 +1,31 @@
+{
+    "type": "music",
+    "domain": "yutorah",
+    "name": "YuTorah",
+    "description": "Stream over 400,000 Torah shiurim (audio lectures) from YUTorah Online as podcasts. Browse by series (e.g. Daf Yomi, Ten Minute Halacha), search by topic or teacher, and subscribe to any series.",
```
OzGav:
```suggestion
    "description": "Stream over 400,000 Torah shiurim (audio lectures) from YUTorah Online.",
```
This is shown in the front end so can’t be too long.

OzGav (changes requested):

OzGav:
Also having the whole provider in one file is too much. Please split this into init.py, constants.py, helpers.py and provider.py

OzGav:
There is a bit to consider here. Marking as draft. Set it back to review required when it’s ready.

music_assistant/providers/yutorah/__init__.py
russlevy: yah, really old stuff, before they had video/pdf/html didn't have a type and it wasn't updated.

music_assistant/providers/yutorah/__init__.py
russlevy: This was to enable the ability to search by artist vs browse -- the UI there was much better. Do you really feel that this can't work? I understand if you feel I should take it out again.

music_assistant/providers/yutorah/__init__.py
russlevy: yah that's a mistake, I was playing around with a few things and had both at some point.

music_assistant/providers/yutorah/manifest.json
russlevy: that's what i get for using too much claude :)

music_assistant/providers/yutorah/__init__.py
russlevy: I had it using the numeric ID earlier, but it made the browsing a bit unwieldy -- the slug at the top just looked wrong, you didn't know where you were. Is there a way around it?

teancom: We do some similar name / id mapping in the Bandcamp provider for the fans/followers stuff. It was actually pretty tricky to get the mapping+caching right, I kept messing it up while trying to handle cases of the client restarting, the server restarting, etc. But it's working now (and pretty well, in my experience).

music_assistant/providers/yutorah/__init__.py
russlevy: This is similar to the teacher -> artist mapping. we want to be able to subscribe to a class (shiur) via podcasts, but we also want them discoverable in search.

music_assistant/providers/yutorah/__init__.py
russlevy: OK, it seems to work! going to copy bbc. not sure what i did wrong last time, but if it works in other places, it's a me problem

music_assistant/providers/yutorah/__init__.py
OzGav: That makes sense, but can you add a comment explaining that the empty string case is for legacy content predating the media type field? As written it just looks like a bug

music_assistant/providers/yutorah/__init__.py
russlevy: 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: I updated to LoginFailed -- what do you mean by reviewing the error handling in the function? just _login()? I think it's a bit more robust now 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: I think I left this here from a previous iteration, when I was debugging. 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
russlevy: leftover from testing. fixed 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
OzGav: If searching by teacher name isn't returning results, that's what needs fixing — you want the search implementation to return matching items as PodcastEpisode and matching series as Podcast when a teacher name is searched. Adding Artist/Track is the wrong fix for a search gap. Can you test what search/get?searchTerm=<teacher name> actually returns from the API? If it returns podcasts by that teacher then the search implementation just needs to surface them correctly.

music_assistant/providers/yutorah/__init__.py
russlevy: An earlier iteration was using cookies + website APIs which had very diff forms, and was trying all at the same time with failovers... this should be better now that I got rid of it. 7f905ba00d8a370359b48082522fd2e366a67382

music_assistant/providers/yutorah/__init__.py
OzGav: Looking at the models, PodcastEpisode has no author/teacher field. Podcast has publisher which could hold a teacher name for single-teacher series. Is that sufficient, or is teacher discoverability important enough to raise a model enhancement request rather than misusing Artist/Track?

russlevy:
> There is a bit to consider here. Marking as draft. Set it back to review required when it’s ready.

Should I mark it as ready when I'm done getting through everything? Just moving functions around right now

music_assistant/providers/yutorah/__init__.py
russlevy: Going through this -- how should we approach this? the class / teacher (shiur / teacher) pair is a podcast I'd want to subscribe to, but the provider doesn't publish it in any way from what I can tell. When I removed this, I couldn't subscribe to a class anymore, as the provider no longer provides podcasts (as this function is needed to be implemented if the provider supports podcasts)

music_assistant/providers/yutorah/__init__.py
OzGav: If _login raises LoginFailed, handle_async_init will let it propagate (as there is no handling there currently) and setup will fail entirely. Is that intentional? If so, you should catch LoginFailed and re-raise as SetupFailedError which will surface the error to the user in the UI. Option B (falling back to anonymous mode silently) isn't great UX — the user entered credentials and got no feedback that they were rejected.

music_assistant/providers/yutorah/__init__.py
OzGav: get_library_podcasts is for providers that maintain their own subscription list server-side that MA syncs down — like Spotify subscribed podcasts. If YuTorah doesn't have that, then the base class no-op is correct and this override should still be deleted. Subscribing to a series happens through browse — the user adds it to the MA library directly, which then calls get_podcast and get_podcast_episodes

OzGav:
As your resolve each comment then mark it as such by clicking the resolve conversation button. Yes once you think it is ready for review towards merging then click the “Ready for Review” button. Just don’t rush it.

russlevy:
> As your resolve each comment then mark it as such by clicking the resolve conversation button. Yes once you think it is ready for review towards merging then click the “Ready for Review” button. Just don’t rush it.

OK great. My general practice has been to make the reviewer resolve when they are happy with the fix, but happy to do it too! Of course not rushing, going to spend a bit of time thinking through the whole podcast / teacher stuff. probably needs a bunch of testing too, when I'm happy with the result I'll mark it as ready for review.

music_assistant/providers/yutorah/__init__.py
russlevy: Actually what do you think about the apple music implementation with music_assistant/providers/apple_music/helpers/browse.py

basically it seems to put both the ID and the name in there separated by | and the UI just knows to show what's before the pipe?

russlevy: I committed this -- let me know what you think, if this fits standard practices. It means we don't have to do any lookups anymore. @teancom I missed your suggestion before I started to copy apple music...

music_assistant/providers/yutorah/__init__.py
russlevy: Yah, I removed the fallback. It just added a bunch of code too. I added the reraise here too, in handle_async_init(). It does look much better now :)

russlevy:
> As your resolve each comment then mark it as such by clicking the resolve conversation button. Yes once you think it is ready for review towards merging then click the “Ready for Review” button. Just don’t rush it.

This should be ready for re-review. I've tried to manually test every path as well too ensure it's working. As well, I felt that provider.py was going to be too big still, so I copied Apple Music and created a browse.py file. Finally, felt some of the functions were a bit unwieldy so broke them up into helper functions.

russlevy (commented):
This should be ready for re-review. I've tried to manually test every path as well too ensure it's working. As well, I felt that provider.py was going to be too big still, so I copied Apple Music and created a browse.py file. Finally, felt some of the functions were a bit unwieldy so broke them up into helper functions.

music_assistant/providers/yutorah/browse.py
```diff
@@ -0,0 +1,312 @@
...
+            return await self._browse_recent()
+        return []
+
+    def _browse_root(self) -> list[BrowseFolder]:
+        """Return the four top-level browse folders."""
+        sections = [
+            ("series", "Browse by Series", "podcasts", False),
+            ("teachers", "Browse by Teacher", "artists", False),
+            ("categories", "Browse by Topic", None, False),
+            ("recent", "Recent Shiurim", None, True),
```
OzGav: Why are you are using keys which are not the same as the text you have as the default?

Why only have translations (which are incorrect) for two of four?

music_assistant/providers/yutorah/helpers.py
```diff
@@ -0,0 +1,358 @@
...
+        },
+        metadata=MediaItemMetadata(
+            description=description or None,
+            images=UniqueList(images) if images else None,
+            links=links or None,
+        ),
+    )
+
+
+def _shiur_to_track(
```
OzGav, line 147: I don't think returning shiurim as Track objects in search results is right. A shiur has no artist and no album, so the track will appear incomplete in the search UI. More importantly, MA will treat it as a music track and attempt metadata enrichment — MusicBrainz lookups etc. — for content that is fundamentally a podcast episode. If SearchResults has no podcast_episodes field, should we raise a model enhancement request to add one rather than misusing Track?

@MarvinSchenkel thoughts on this?

music_assistant/providers/yutorah/provider.py
```diff
@@ -0,0 +1,408 @@
...
+        return results
+
+    # -----------------------------------------------------------------------
+    # Internal — API calls
+    # -----------------------------------------------------------------------
+
+    async def _fetch_episodes_paged(
+        self,
+        parent_series_id: str | None = None,
+        **filter_params: Any,
```
OzGav: Can this really be "Any"?

music_assistant/providers/yutorah/browse.py
```diff
@@ -0,0 +1,312 @@
...
+    # Interface required from the host class (YuTorahProvider)
+    # Declared under TYPE_CHECKING so they are visible to mypy but do not
+    # override the final attributes declared in the MusicProvider base class.
+    # ------------------------------------------------------------------
+
+    if TYPE_CHECKING:
+        domain: str
+        instance_id: str
+
+    async def _api_get(self, endpoint: str, **params: Any) -> Any:
```
OzGav, line 49: Your comment says these are declared under TYPE_CHECKING but they aren't — they are real method definitions with raise NotImplementedError bodies?

OzGav:
There are a bunch of tests failing (although one looks unrelated to your code so ignore that)

music_assistant/providers/yutorah/provider.py
```diff
@@ -195,6 +196,35 @@ async def get_track(self, prov_track_id: str) -> Track:
...
+            provider_mappings={
+                ProviderMapping(
+                    item_id=prov_artist_id,
+                    provider_domain="yutorah",
+                    provider_instance=self.instance_id,
+                )
+            },
+        )
+
+    async def get_artist_toptracks(self, prov_artist_id: str) -> list[Track]:
```
OzGav, line 221: I’m still not sure about teachers as artists but on this particular method, it won’t get called as you haven’t added the relevant provider feature

OzGav:
Please update the PR description with any changes made through this process. This will probably get looked at further next week.

music_assistant/providers/yutorah/helpers.py
MarvinSchenkel: Yeah agree with @OzGav here. Looking at this provider, it exclusively delivers podcasts / podcast episodes. I think we can remove any references to Artists and Tracks. Since there are not library_xxxx functions (because I don't think YuToarh has a library?), you can rely on browse endpoints so people can still browse the content by teacher for example.

music_assistant/providers/yutorah/helpers.py
```diff
@@ -0,0 +1,374 @@
...
+        t = teachers[0]
+        teacher_id = str(t.get("teacherID") or "")
+        teacher_name = t.get("teacherName") or ""
+        image_url = t.get("teacherPhotoURL") or t.get("teacherAlbumURL") or ""
+    else:
+        teacher_id = str(shiur.get("teacherid") or "")
+        teacher_name = shiur.get("teacherfullname") or ""
+        photo = shiur.get("PHOTO") or shiur.get("photo") or ""
+        if photo and not photo.startswith("http"):
+            photo = f"https://cdnyutorah.cachefly.net/_images/roshei_yeshiva/{photo}"
```
MarvinSchenkel, line 232: We can move this into constants.py too

music_assistant/providers/yutorah/helpers.py
russlevy: The issue is the discoverability. Almost all of the episodes are standalone, and you may want to search for a specific topic to find it across different teachers and series. This is how it works on yutorah.org as well -- though there are teacher, and teacher / series pairs (which we are modeling as podcasts), the primary search is by class / shiur / episode

OzGav: Then the fix is possibly to adjust the search to return results which match the teacher name as well as the podcast name. You can also implement different browse folders if there are topics that users might typically be interested in.

MarvinSchenkel:
Hey @russlevy, marking this PR as draft so we can keep on top of which PRs need our attention. Just flag it as 'ready for review' when you want us to have another look!

MarvinSchenkel:
@russlevy any progress on this one?

OzGav:
@russlevy did you intend to keep progressing this? It is OK if not but then probably close the PR.

russlevy:
I am going to close. The use case for search etc doesn't really fit as a music provider and the current interfaces don't lend well to it.

PR #3432, opened 2026-03-19
Add support for dynamic playlists to the Queue controller

dmoo500:
This PR extracts the provider-agnostic queue behavior improvements from the Apple Music work.

## Changes
- Treat RADIO items as continuous sources when no next index exists.
- Allow end-of-queue handling to continue RADIO playback instead of clearing immediately.
- Surface stream metadata title/artist for RADIO items in player media output.

## Why separate PR
These changes are controller-level and benefit any provider that uses track-based radio behavior, not only Apple Music.

## Related

The queue behavior here is a workaround for the lack of a first-class dynamic/endless track source type in MA. A proper solution is tracked in:
- **[music-assistant/models#192](https://github.com/music-assistant/models/pull/192)** — Draft PR adding `MediaType.DYNAMIC_PLAYLIST` and a `DynamicPlaylist` dataclass
- **[orgs/music-assistant/discussions/5129](https://github.com/orgs/music-assistant/discussions/5129)** — Design discussion

Once the models PR is released, this PR's workarounds may be superseded by a dedicated queue path for `DYNAMIC_PLAYLIST` items.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -952,7 +952,9 @@ class PlayerQueuesController(CoreController):
         # debounce rapid next button presses using call_later
         self.mass.call_later(
             1,
-            self.play_index(queue_id, next_index),
+            self.play_index,
+            queue_id,
+            next_index,
             task_id=f"queue_play_index_{queue_id}",
         )
 
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -554,6 +554,18 @@ class PlayerQueuesController(CoreController):
                     # item is MediaItemType | ItemMapping at this point
                     media_item = item
 
+                if (
+                    isinstance(media_item, ItemMapping)
+                    and media_item.media_type == MediaType.PLAYLIST
+                ):
+                    # Resolve mapping once so playlist-specific logic can use
+                    # a full Playlist object without duplicating branches.
+                    with suppress(MusicAssistantError):
+                        media_item = await self.mass.music.playlists.get(
+                            media_item.item_id,
+                            media_item.provider,
+                        )
+
                 # Save requested media item to play on the queue so we can use it as a source
                 # for Don't stop the music. Use FIFO list to keep track of the last 10 played items
                 # Skip ItemMapping and BrowseFolder - only queue full MediaItemType objects
@@ -572,21 +584,6 @@ class PlayerQueuesController(CoreController):
                         media_item, "is_dynamic", False
                     ):
                         radio_source.append(media_item)
-                elif (
-                    isinstance(media_item, ItemMapping)
-                    and media_item.media_type == MediaType.PLAYLIST
-                ):
-                    # Resolve mapping to full playlist so dynamic playlists can hook into
-                    # existing radio_source refill behavior.
-                    try:
-                        mapped_playlist = await self.mass.music.playlists.get(
-                            media_item.item_id,
-                            media_item.provider,
-                        )
-                    except MusicAssistantError:
-                        mapped_playlist = None
-                    if mapped_playlist and getattr(mapped_playlist, "is_dynamic", False):
-                        radio_source.append(mapped_playlist)
 
                 # handle default enqueue option if needed
                 if option is None:
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -2101,11 +2101,15 @@ class PlayerQueuesController(CoreController):
                     QueueItem.from_media_item(queue_id, x) for x in dynamic_tracks if x.available
                 ]
                 if queue_items:
+                    cur_index = self._queues[queue_id].current_index or 0
                     await self.load(
                         queue_id,
                         queue_items,
-                        insert_at_index=len(self._queue_items[queue_id]) + 1,
+                        insert_at_index=cur_index + 1,
+                        keep_remaining=False,
+                        keep_played=True,
                     )
+                    return
             except MusicAssistantError as err:
                 self.logger.warning(
                     "Failed to refill dynamic playlist %s for queue %s: %s",
@@ -2113,7 +2117,6 @@ class PlayerQueuesController(CoreController):
                     queue.display_name,
                     err,
                 )
-            return
         radio_tracks = await self._get_radio_tracks(queue_id=queue_id, is_initial_radio_mode=False)
         # fill queue - filter out unavailable items
         queue_items = [QueueItem.from_media_item(queue_id, x) for x in radio_tracks if x.available]
@@ -2883,10 +2886,11 @@ class PlayerQueuesController(CoreController):
                             if x.available
                         ]
                         if queue_items:
+                            cur_index = queue.current_index or 0
                             await self.load(
                                 queue.queue_id,
                                 queue_items,
-                                insert_at_index=len(self._queue_items[queue.queue_id]),
+                                insert_at_index=cur_index + 1,
                                 keep_remaining=False,
                                 keep_played=True,
                                 shuffle=False,
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -1913,9 +1913,12 @@ class PlayerQueuesController(CoreController):
             "Fetching tracks to play for playlist %s",
             playlist.name,
         )
+        force_refresh = bool(getattr(playlist, "is_dynamic", False))
         # TODO: Handle other sort options etc.
         async for playlist_track in self.mass.music.playlists.tracks(
-            playlist.item_id, playlist.provider
+            playlist.item_id,
+            playlist.provider,
+            force_refresh=force_refresh,
         ):
             if not playlist_track.available:
                 continue
@@ -2825,6 +2828,39 @@ class PlayerQueuesController(CoreController):
                         )
                         await self.play_index(queue.queue_id, next_index)
                     return
+            # If the queue was started from a dynamic playlist, fetch fresh tracks and continue.
+            dynamic_playlist = next(
+                (
+                    item
+                    for item in reversed(queue.enqueued_media_items)
+                    if isinstance(item, Playlist) and getattr(item, "is_dynamic", False)
+                ),
+                None,
+            )
+            if dynamic_playlist is not None:
+                dynamic_tracks = await self.get_playlist_tracks(dynamic_playlist, start_item=None)
+                if dynamic_tracks:
+                    queue_items = [
+                        QueueItem.from_media_item(queue.queue_id, x)
+                        for x in dynamic_tracks
+                        if x.available
+                    ]
+                    if queue_items:
+                        await self.load(
+                            queue.queue_id,
+                            queue_items,
+                            insert_at_index=len(self._queue_items[queue.queue_id]),
+                            keep_remaining=False,
+                            keep_played=True,
+                            shuffle=False,
+                        )
+                        if queue.current_index is not None and (
+                            next_item := self.get_next_item(queue.queue_id, queue.current_index)
+                        ):
+                            next_index = self.index_by_id(queue.queue_id, next_item.queue_item_id)
+                            if next_index is not None:
+                                await self.play_index(queue.queue_id, next_index)
+                                return
             # If the last item was a radio station, continue playing instead of clearing.
             # For track-based radio (Apple Music Artist Radio etc.) this fetches the next
             # track from the station; for live radio it would reconnect to the stream.
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -568,6 +568,25 @@ class PlayerQueuesController(CoreController):
                     queue.enqueued_media_items.append(media_item)
                     if len(queue.enqueued_media_items) > 10:
                         queue.enqueued_media_items.pop(0)
+                    if isinstance(media_item, Playlist) and getattr(
+                        media_item, "is_dynamic", False
+                    ):
+                        radio_source.append(media_item)
+                elif (
+                    isinstance(media_item, ItemMapping)
+                    and media_item.media_type == MediaType.PLAYLIST
+                ):
+                    # Resolve mapping to full playlist so dynamic playlists can hook into
+                    # existing radio_source refill behavior.
+                    try:
+                        mapped_playlist = await self.mass.music.playlists.get(
+                            media_item.item_id,
+                            media_item.provider,
+                        )
+                    except MusicAssistantError:
+                        mapped_playlist = None
+                    if mapped_playlist and getattr(mapped_playlist, "is_dynamic", False):
+                        radio_source.append(mapped_playlist)
 
                 # handle default enqueue option if needed
                 if option is None:
@@ -782,8 +801,6 @@ class PlayerQueuesController(CoreController):
         queue.radio_source = []
         if queue.state != PlaybackState.IDLE and not skip_stop:
             self.mass.create_task(self.stop(queue_id))
-        self.mass.cancel_timer(f"queue_play_index_{queue_id}")
-        self._transitioning_players.discard(queue_id)
         queue.current_index = None
         queue.current_item = None
         queue.elapsed_time = 0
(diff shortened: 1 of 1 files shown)


OzGav:
@dmoo500 there are conflicts that need to be resolved.

MarvinSchenkel:
Can you explain to me in your own words what problem these changes are solving?

dmoo500:
These changes fix a queue-handling gap for radio playback (track-based radio from Apple Music).
The queue logic was treating radio like a finite list of tracks, so when it reached the “end” (no next index), playback could stop, clear too early, or fail to continue smoothly. For track-based radio (like artist radio), that “end” is artificial, because the provider is expected to fetch the next track continuously.

This PR makes radio behave as a continuous source: when no next queue item exists, it refreshes stream details and re-requests the next radio track instead of ending playback. It also improves end-of-queue behaviour so radio can resume rather than being cleared immediately, and it surfaces live stream metadata (title/artist) so players show what is actually playing. In short, it turns radio playback from “finite queue semantics” into “continuous stream semantics”.

MarvinSchenkel:
I think we are mixing different concepts of 'Radio' here. In MA, a `MediaType.RADIO` is a continuous radio stream (e.g. FM radio). 

What you are referring to (Apple Music Radio), are dynamic playlists that keep changing after they 'run out of tracks'. I don't think we have something for this inside MA, but we do plan on adding dynamic playlists as some point.

dmoo500:
Fair point on the naming — I mapped it to `MediaType.RADIO` because Apple Music itself calls this feature "Radio" (e.g. "Artist Radio", "Station"). I can see how that clashes with MA's definition of a continuous live stream.

That said, the behavior this PR implements **does work** end-to-end on my setup: MA fetches the next track from Apple Music's `next-tracks` API each time the current one ends, giving seamless continuous playback. Note that this PR depends on #3433 — both together are required for Apple Music track-based radio to work.

Could you point me in the right direction on how to implement this correctly within MA's architecture? I'd like to make sure this fits the existing design rather than working around it.

marcelveldt:
Marked as draft because I think this PR needs more discussion. It's trying to fix a non-existent issue

dmoo500:
@marcelveldt, @MarvinSchenkel:

I've opened a discussion to address this topic and gather input from the maintainers: [https://github.com/orgs/music-assistant/discussions/5129](https://github.com/orgs/music-assistant/discussions/5129)

MarvinSchenkel:
> Marked as draft because I think this PR needs more discussion. It's trying to fix a non-existent issue

@marcelveldt 
I do think there is a bit of a gap here in MA. Some music providers have 'Radio stations' which are similar to our 'radio mode' (e.g. Apple Music's Discover Radio). Those stations deliver dynamic content and therefore need to be refreshed/refetched after let's say the initial 30 tracks have played. I do think we could combine this with a new MediaType we have been talking about for a while now: `MediaType.DYNAMIC_PLAYLIST` which we can use within MA for our fancy playlists, but can also be used to represent dynamic content in music providers.

dmoo500:
### What was added
- `get_playlist_tracks(...)` now uses `force_refresh` for dynamic playlists.
- End-of-queue handling now:
  - detects when playback originated from a dynamic playlist,
  - fetches fresh tracks,
  - appends them to the queue,
  - and resumes playback automatically.

### Scope note
This is queue/controller behavior and is intentionally separated from provider-specific Apple Music changes in **PR #3433**.

music_assistant/controllers/player_queues.py
```diff
@@ -2736,6 +2781,49 @@ async def _clear_or_resume_delayed() -> None:
...
+                            keep_played=True,
+                            shuffle=False,
+                        )
+                        if queue.current_index is not None and (
+                            next_item := self.get_next_item(queue.queue_id, queue.current_index)
+                        ):
+                            next_index = self.index_by_id(queue.queue_id, next_item.queue_item_id)
+                            if next_index is not None:
+                                await self.play_index(queue.queue_id, next_index)
+                                return
```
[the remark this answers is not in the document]
dmoo500: Thanks for catching that! You're absolutely right—the dynamic-playlist refresh can fail with provider/network errors, and we need to gracefully fall back to the radio/clear handling.

✅ **Fix applied:** Wrapped the dynamic playlist fetch/load in `try/except(MusicAssistantError)` with logging. On failure, the code now logs a warning and continues to the radio station or queue-clear fallback, so the background task won't abort and the queue stays responsive.

Commit: updating now...

marcelveldt:
Can we focus on one thing at a time in the PR please ?

- MediaType.RADIO needs to changes now - that is for ordinary radio streams and already works fine today. If not, open a focused PR on that alone.
- We want to add support for playlists with the "is_dynamic" set to True so lets handle that in this PR.
- Let's please try to use as much of the existing logic as possible where we already watch the end of the queue and refill dynamic tracks.
- When we add a dynamic playlist to the queue, let's automatically add that to the "radio items" of the queue, so teh user is informed that there is a dynamic component active. Users already know the "radio mode" concept so this way they dont have to learn something new.


@MarvinSchenkel implemented most of the current "radio mode" features which is working on based on similar tracks. Let's try to get this over the finish line with reusing/extending as much of our existing structure as possible. In the end the concept is the same: Our current generic radio mode dynamically fills the queue (but gets similar tracks based on the reference) and this new dynamic playlists just automatically enable radio mode and provide the dynamic tracks themselves.

music_assistant/controllers/player_queues.py
```diff
@@ -772,6 +772,8 @@ def clear(self, queue_id: str, skip_stop: bool = False) -> None:
         queue.radio_source = []
         if queue.state != PlaybackState.IDLE and not skip_stop:
             self.mass.create_task(self.stop(queue_id))
+        self.mass.cancel_timer(f"queue_play_index_{queue_id}")
+        self._transitioning_players.discard(queue_id)
```
marcelveldt: this seems an unrelated change

music_assistant/controllers/player_queues.py
```diff
@@ -924,7 +926,20 @@ async def next(self, queue_id: str) -> None:
...
+            current_item = self.get_item(queue_id, idx)
+            if current_item is not None and current_item.media_type == MediaType.RADIO:
+                current_item.streamdetails = None
+                self.mass.call_later(
+                    1,
+                    self.play_index,
+                    queue_id,
+                    idx,
+                    task_id=f"queue_play_index_{queue_id}",
+                )
```
marcelveldt: this can be dropped

music_assistant/controllers/player_queues.py
```diff
@@ -1395,6 +1412,16 @@ async def load_next_queue_item(
...
+                # For RADIO type items, treat them as a continuous source: clear the
+                # cached stream details and re-request the next track from the provider.
+                cur_item = self.get_item(queue_id, cur_index)
+                if cur_item is not None and cur_item.media_type == MediaType.RADIO:
+                    cur_item.streamdetails = None
+                    try:
+                        await self._load_item(cur_item, None)
+                    except (MediaNotFoundError, AudioError) as err:
+                        raise QueueEmpty("No more tracks available from radio station.") from err
+                    return cur_item
```
marcelveldt: this can be dropped

music_assistant/controllers/player_queues.py
```diff
@@ -1749,6 +1776,17 @@ async def player_media_from_queue_item(self, queue_item: QueueItem) -> PlayerMed
...
+            # stream_metadata carries the currently playing track's artist/title.
+            if (
+                queue_item.media_type == MediaType.RADIO
+                and queue_item.streamdetails
+                and (sm := queue_item.streamdetails.stream_metadata)
+            ):
+                if sm.title:
+                    media.title = sm.title
+                if sm.artist:
+                    media.artist = sm.artist
```
marcelveldt: this can be dropped

music_assistant/controllers/player_queues.py
```diff
@@ -1869,9 +1907,12 @@ async def get_playlist_tracks(
             "Fetching tracks to play for playlist %s",
             playlist.name,
         )
+        force_refresh = bool(getattr(playlist, "is_dynamic", False))
```
marcelveldt: why do we need the magic force refresh ?
isn't it true that we always get the dynamic tracks from a separate handler ?

music_assistant/controllers/player_queues.py
dmoo500: You're right, that's out of scope. Removed.

music_assistant/controllers/player_queues.py
dmoo500: Done, removed.

music_assistant/controllers/player_queues.py
dmoo500: Done, removed.

music_assistant/controllers/player_queues.py
dmoo500: Without `force_refresh`, `playlists.tracks()` returns cached rows from the local database — so every refill call would return the same tracks instead of asking the provider for a fresh batch. For dynamic playlists the whole point is to get new tracks each time the queue runs low.

If you prefer the provider to own this (e.g. by setting a short cache TTL on dynamic playlist tracks), I'm happy to drop it here and handle it there instead.

music_assistant/controllers/player_queues.py
dmoo500: Done, removed.

dmoo500:
Thanks, agreed on all four points — the PR is now rescoped accordingly:

- **`MediaType.RADIO`**: All RADIO-specific changes removed. No ordinary radio stream behavior is touched.
- **`is_dynamic` playlists**: This PR now only handles dynamic playlist refill using `is_dynamic=True`.
- **Reuse existing logic**: Dynamic playlists hook directly into the existing `radio_source` / `_fill_radio_tracks()` pipeline — no new end-of-queue logic added.
- **Radio items UX**: When a dynamic playlist is queued, it is automatically added to `queue.radio_source` so the user sees the familiar radio mode indicator.

music_assistant/controllers/player_queues.py
```diff
@@ -2042,9 +2066,38 @@ async def _fill_radio_tracks(self, queue_id: str) -> None:
...
+                        insert_at_index=len(self._queue_items[queue_id]) + 1,
+                    )
+            except MusicAssistantError as err:
+                self.logger.warning(
+                    "Failed to refill dynamic playlist %s for queue %s: %s",
+                    getattr(dynamic_playlist, "name", repr(dynamic_playlist)),
+                    queue.display_name,
+                    err,
+                )
+            return
```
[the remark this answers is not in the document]
dmoo500: Fixed — return is now only reached when tracks were actually loaded. On empty result or error, the code falls through to the existing `_get_radio_tracks` path.

music_assistant/controllers/player_queues.py
```diff
@@ -2736,6 +2793,60 @@ async def _clear_or_resume_delayed() -> None:
...
+                        ]
+                        if queue_items:
+                            await self.load(
+                                queue.queue_id,
+                                queue_items,
+                                insert_at_index=len(self._queue_items[queue.queue_id]),
+                                keep_remaining=False,
+                                keep_played=True,
+                                shuffle=False,
+                            )
```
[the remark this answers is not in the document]
dmoo500: Fixed — same pattern applied: `insert_at_index=current_index+1` with `keep_remaining=False so refreshed tracks immediately follow the current position and any stale tail is replaced.

music_assistant/controllers/player_queues.py
```diff
@@ -2042,9 +2066,38 @@ async def _fill_radio_tracks(self, queue_id: str) -> None:
...
+                dynamic_tracks = await self.get_playlist_tracks(dynamic_playlist, start_item=None)
+                queue_items = [
+                    QueueItem.from_media_item(queue_id, x) for x in dynamic_tracks if x.available
+                ]
+                if queue_items:
+                    await self.load(
+                        queue_id,
+                        queue_items,
+                        insert_at_index=len(self._queue_items[queue_id]) + 1,
+                    )
```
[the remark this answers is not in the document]
dmoo500: Fixed — now uses `insert_at_index=current_index+1` with `keep_remaining=False`, which replaces the stale tail of the queue with the fresh batch instead of appending. This prevents unbounded queue growth and avoids repeated tracks.

music_assistant/controllers/player_queues.py
```diff
@@ -558,6 +558,25 @@ async def play_media(
                     queue.enqueued_media_items.append(media_item)
                     if len(queue.enqueued_media_items) > 10:
                         queue.enqueued_media_items.pop(0)
+                    if isinstance(media_item, Playlist) and getattr(
+                        media_item, "is_dynamic", False
+                    ):
```
marcelveldt: you can skip the getattr here - a Playlist mediaitem has the is_dynamic attribute

music_assistant/controllers/player_queues.py
```diff
@@ -558,6 +558,25 @@ async def play_media(
...
+                    # existing radio_source refill behavior.
+                    try:
+                        mapped_playlist = await self.mass.music.playlists.get(
+                            media_item.item_id,
+                            media_item.provider,
+                        )
+                    except MusicAssistantError:
+                        mapped_playlist = None
+                    if mapped_playlist and getattr(mapped_playlist, "is_dynamic", False):
+                        radio_source.append(mapped_playlist)
```
marcelveldt: maybe do this call a bit above so you dont have to duplicate the logic ?
so if you first detect that the item is an itemmapping of type playlist, resolve the full mediaitem, overwriting the existing media_item

music_assistant/controllers/player_queues.py
dmoo500: You're right. After rebasing onto current dev (models 1.1.109), `Playlist.is_dynamic` is now typed, so I removed the `getattr(...)` usage and switched to direct `media_item.is_dynamic` checks.

music_assistant/controllers/player_queues.py
dmoo500: Good point. I moved playlist `ItemMapping` resolution earlier in the flow and overwrite `media_item` with the resolved playlist (when available), so the dynamic-playlist handling now goes through a single path without duplicate logic.

music_assistant/controllers/player_queues.py
```diff
@@ -568,6 +580,8 @@ async def play_media(
                     queue.enqueued_media_items.append(media_item)
                     if len(queue.enqueued_media_items) > 10:
                         queue.enqueued_media_items.pop(0)
+                    if isinstance(media_item, Playlist) and media_item.is_dynamic:
```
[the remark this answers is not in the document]
dmoo500: Fixed — great catch. Dynamic playlists were indeed getting added twice to `radio_source` when `radio_mode=True` (once in the dynamic-playlist branch and once in the radio-mode branch). I now only append in the dynamic-playlist branch when `not radio_mode`, so radio mode keeps a single source entry.

music_assistant/controllers/player_queues.py
```diff
@@ -554,6 +554,18 @@ async def play_media(
                     # item is MediaItemType | ItemMapping at this point
                     media_item = item
 
+                if (
+                    isinstance(media_item, ItemMapping)
+                    and media_item.media_type == MediaType.PLAYLIST
+                ):
+                    # Resolve mapping once so playlist-specific logic can use
+                    # a full Playlist object without duplicating branches.
```
```suggestion
                    # Resolve ItemMapping for a playlist so the full Playlist object
                    # so we have access to details such as 'is_dynamic'
```

music_assistant/controllers/player_queues.py
```diff
@@ -2796,6 +2855,61 @@ async def _clear_or_resume_delayed() -> None:
                         )
                         await self.play_index(queue.queue_id, next_index)
                     return
+            # If the queue was started from a dynamic playlist, fetch fresh tracks and continue.
```
marcelveldt, line 2858: in a future PR, let's try to clean up this code a bit as its kind of hard to read and there are many statements .
Not for now - can be done once we start refactoring the queue controller a bit because it has grown over time into a big beast.

music_assistant/controllers/player_queues.py
marcelveldt: Its fine for now - we can always optimize later

marcelveldt (approved):
Nice work @dmoo500 !

PR #3433, opened 2026-03-19
Apple Music: Add support for radio stations 

dmoo500:
## Overview

Adds Apple Music radio station playback and improved URL import support. Stations can be played directly via their share URL (e.g. `music.apple.com/…/station/…`) or by adding a station playlist (`ra.xxx`) to the queue.

> **Note:** PR #3527 (Dynamic playlist queue support) must be merged before this PR. The station playback implemented here relies on the queue controller in #3527 correctly handling `is_dynamic=True` playlists for continuous track refill.

### `providers/apple_music/__init__.py`

**Radio station playback:**
- `get_radio()`: new implementation that looks up a station in the catalog and falls back to a synthetic `Radio` object if the station is not indexed in the user's storefront.
- `get_playlist()`: `ra.xxx` station IDs now return a synthetic `Playlist` with `is_dynamic=True` instead of hitting a non-existent library endpoint (which previously raised `MediaNotFoundError`).
- `get_playlist_tracks()`: `ra.xxx` IDs are routed to the `me/stations/next-tracks/{id}` endpoint via a new `_get_station_tracks()` helper — the playlist tracks endpoint does not exist for stations.
- `_get_station_tracks()`: POSTs to `me/stations/next-tracks/{station_id}` and parses the returned tracks including ratings.

**Robustness:**
- Widevine CDM file loading is now wrapped in `try/except FileNotFoundError` with a clear warning — radio stations and unencrypted library tracks remain available without CDM files.
- `get_playlist` cache now uses `cache_checksum="stations-metadata-v2"` to bust stale cached metadata for station playlists.

### `helpers/uri.py`
- Parse Apple Music share URLs for radio stations and map them to the correct `media_type` / provider item ID.

### `tests/core/test_helpers.py`
- Tests for the new URI parsing logic covering station, playlist, album, artist, and track share URLs.

### DIFF
--- a/music_assistant/providers/apple_music/__init__.py
+++ b/music_assistant/providers/apple_music/__init__.py
@@ -58,7 +58,7 @@ from music_assistant_models.media_items import (
     Track,
     UniqueList,
 )
-from music_assistant_models.streamdetails import StreamDetails, StreamMetadata
+from music_assistant_models.streamdetails import StreamDetails
 from pywidevine import PSSH, Cdm, Device, DeviceTypes
 from pywidevine.license_protocol_pb2 import WidevinePsshData
 from shortuuid import uuid
@@ -641,14 +641,19 @@ class AppleMusicProvider(MusicProvider):
             tracks.append(track)
         return tracks
 
-    @use_cache(3600 * 3)  # cache for 3 hours
     async def get_playlist_tracks(self, prov_playlist_id, page: int = 0) -> list[Track]:
         """Get all playlist tracks for given playlist id."""
         if prov_playlist_id.startswith("ra."):
             # Radio stations don't support paging; each call to next-tracks yields a fresh batch.
+            # Must not be cached — every call should return the next set of tracks from Apple Music.
             if page > 0:
                 return []
             return await self._get_station_tracks(prov_playlist_id)
+        return await self._get_playlist_tracks_cached(prov_playlist_id, page)
+
+    @use_cache(3600 * 3)  # cache for 3 hours
+    async def _get_playlist_tracks_cached(self, prov_playlist_id, page: int = 0) -> list[Track]:
+        """Fetch and cache tracks for a regular (non-station) playlist."""
         if self._is_catalog_id(prov_playlist_id):
             endpoint = f"catalog/{self._storefront}/playlists/{prov_playlist_id}/tracks"
         else:
@@ -792,51 +797,7 @@ class AppleMusicProvider(MusicProvider):
 
     async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
         """Return the content details for the given track when it will be streamed."""
-        if media_type == MediaType.RADIO:
-            station_hash, is_live = await self._get_station_play_params(item_id)
-            if is_live:
-                try:
-                    stream_url = await self._fetch_radio_stream_url(item_id, station_hash)
-                    return StreamDetails(
-                        item_id=item_id,
-                        provider=self.instance_id,
-                        audio_format=AudioFormat(content_type=ContentType.UNKNOWN),
-                        media_type=MediaType.RADIO,
-                        stream_type=StreamType.HLS,
-                        path=stream_url,
-                        can_seek=False,
-                        allow_seek=False,
-                    )
-                except MediaNotFoundError:
-                    # webPlayback failed — station may be a curated track-based channel
-                    # (e.g. third-party stations like "Radio Bob Best of Rock") that are
-                    # served via next-tracks rather than a live HLS stream, so fall through.
-                    self.logger.debug(
-                        "Live stream unavailable for %s, trying track-based fallback", item_id
-                    )
-            # Track-based Artist Radio: fetch next track from the station and stream it
-            track = await self._fetch_next_radio_station_track(item_id)
-            streamdetails = await self._get_track_stream_details(track.item_id)
-            if track.name:
-                artist_name = track.artists[0].name if track.artists else None
-                streamdetails.stream_metadata = StreamMetadata(
-                    title=track.name,
-                    artist=artist_name,
-                )
-            if track.duration:
-                streamdetails.duration = track.duration
-            return streamdetails
         return await self._get_track_stream_details(item_id)
-
-    async def _fetch_next_radio_station_track(self, station_id: str) -> Track:
-        """Fetch the next track for a track-based Apple Music radio station."""
-        response = await self._post_data(f"me/stations/next-tracks/{station_id}")
-        tracks = response.get("data", [])
-        if not tracks:
-            raise MediaNotFoundError(f"No tracks available for station {station_id}")
-        track_obj = tracks[0]
-        rating_response = await self._get_ratings([track_obj["id"]], MediaType.TRACK)
-        return self._parse_track(track_obj, rating_response.get(track_obj["id"]))
 
     async def _get_track_stream_details(self, item_id: str) -> StreamDetails:
         """Return StreamDetails for a single catalog or library track."""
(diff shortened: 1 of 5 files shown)


dmoo500:
@marcelveldt, @MarvinSchenkel: 

I've opened a discussion to address this topic and gather input from the maintainers: [https://github.com/orgs/music-assistant/discussions/5129](https://github.com/orgs/music-assistant/discussions/5129)

dmoo500:
This PR is intentionally scoped to **Apple Music provider + URL import handling only**.

### Included in this PR
- Mapping supported Apple Music station/share URLs to the appropriate media resolution path.
- Provider-side handling for Apple Music station items that are represented as dynamic playlists.
- URL import improvements needed to make Apple Music station links resolve and play reliably.

### Not included in this PR
- Queue/controller-level endless refill behavior.

That queue behavior is handled separately in **PR #3432** to keep concerns split:
- **PR #3433** = provider/import behavior
- **PR #3432** = queue playback continuity behavior

MarvinSchenkel:
@dmoo500 just double checking with you before I do a deep dive review, this should be rewritten to dynamic playlists first right? Now that your other PR has been merged

dmoo500:
Need a bit more time to verify this properly.
I also need to open a separate PR because `is_dynamic` is currently not persisted. Fixing that requires changes in shared core components (`playlists.py`, `music.py`, and again on the `player_queues.py`).  

Before PR #3432 was merged, it was already requested that the `player_queues.py` refactor be handled in a separate PR. I’m planning to combine that refactor with the common-component changes in the follow-up PR.

So, I'm setting this PR back to draft.

dmoo500:
**Update:** The shared core changes mentioned above have been addressed in PR #3527, which handles:
- `is_dynamic` persistence in `playlists.py` and `music.py` (DB schema v36)
- The `player_queues.py` refactor (dynamic playlist refill, session guard, method extraction)

This PR (#3433) has been cleaned up to be **provider-only** — it now contains only changes to `apple_music/__init__.py`, `helpers/uri.py`, and the corresponding tests.

PR #3527 must be merged before this one, as the station playback here relies on the queue controller's `is_dynamic` handling introduced there.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -460,6 +545,29 @@ async def get_library_playlists(self) -> AsyncGenerator[Playlist, None]:
             elif item and item["id"]:
                 yield self._parse_playlist(item, is_favourite)
 
+    @use_cache(3600 * 24)
+    async def get_radio(self, prov_radio_id: str) -> Radio:
```
marcelveldt: Is this an actual Radio station in the context of MA ?
So a single stream with content, not a dynamic playlist ?

music_assistant/providers/apple_music/__init__.py
```diff
@@ -694,6 +929,106 @@ async def set_favorite(self, prov_item_id: str, media_type: MediaType, favorite:
             endpoint = f"me/ratings/library-{item_type}/{prov_item_id}"
         await self._put_data(endpoint, data=data)
 
+    def _parse_radio(self, station_obj: dict[str, Any]) -> Radio:
```
marcelveldt: I'm having the feeling that you try to squeeze dynamic playlists into a Radio model here. Correct me if I'm wrong. If these are not actual radio stations, I really think we should list them in the Playlists section and create some nice visual element that they're dynamic.

If these are actual radio stations (like a station on tunein etc) then ignore me, this is good

music_assistant/providers/apple_music/__init__.py
```diff
@@ -694,6 +929,106 @@ async def set_favorite(self, prov_item_id: str, media_type: MediaType, favorite:
...
+            # Also try with stationHash in case it is required for some stations.
+            attempts += [
+                {"stationId": station_id, "stationHash": station_hash},
+                {"salableAdamId": numeric_id, "stationHash": station_hash},
+            ]
+        last_error: str = ""
+        for data in attempts:
+            try:
+                async with self.mass.http_session.post(
+                    playback_url, headers=self._get_decryption_headers(), json=data, ssl=True
```
marcelveldt:
```suggestion
                    playback_url, headers=self._get_decryption_headers(), json=data
```
you can skip the ssl=True because mass.http_session is ssl by default

dmoo500: Good catch — removed. Pushed.

music_assistant/providers/apple_music/__init__.py
dmoo500: `get_radio` handles true live broadcast stations (isLive=true, e.g. Apple Music 1). However, after reviewing this more carefully: this method is effectively dead code in the current PR — `_browse_stations` filters out all isLive=true stations upfront, so `get_radio` is never invoked. Happy to remove both `get_radio` and `_parse_radio` if you prefer to keep the PR clean.

music_assistant/providers/apple_music/__init__.py
dmoo500: You're right to flag this. Apple Music's recommendations API returns a mixed list of station objects — both live broadcast stations (isLive=true, e.g. Apple Music 1) and artist stations (isLive=false, e.g. "Dead Sara Essentials"). In `_browse_stations`, live stations are filtered out because the Apple Music webPlayback endpoint returns error 3077 for live streams — this is an account/token-level restriction, not a missing CDM file issue. Only isLive=false artist stations are shown, mapped to `Playlist(is_dynamic=True)` via `_parse_station_as_playlist` — never as `Radio` items.

This means `_parse_radio` and `get_radio` are not used today. They were kept as a stub in case live broadcast streaming becomes viable in the future (e.g. if the webPlayback endpoint drops the error 3077 restriction). Happy to remove both if you prefer. The browse folder name "Radio Stations" follows Apple Music's own terminology — happy to rename it if that's confusing in the MA context.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -485,9 +593,50 @@ async def get_track(self, prov_track_id) -> Track:
         is_favourite = rating_response.get(prov_track_id)
         return self._parse_track(response["data"][0], is_favourite)
 
-    @use_cache()
+    @use_cache(cache_checksum="stations-metadata-v2")
     async def get_playlist(self, prov_playlist_id, is_favourite: bool = False) -> Playlist:
         """Get full playlist details by id."""
+        if prov_playlist_id.startswith("ra."):
+            # Try the catalog endpoint to get the real name and artwork.
+            name = prov_playlist_id
```
MarvinSchenkel: Can we put this logic in a separate private function called `_get_station_playlist` and call it from here?

music_assistant/providers/apple_music/__init__.py
```diff
@@ -644,6 +813,67 @@ async def get_similar_tracks(self, prov_track_id, limit=25) -> list[Track]:
...
+                streamdetails.stream_metadata = StreamMetadata(
+                    title=track_title,
+                    artist=artist_name,
+                )
+            if duration_secs:
+                streamdetails.duration = duration_secs
+            return streamdetails
+        return await self._get_track_stream_details(item_id)
+
+    async def _fetch_next_station_track(
```
MarvinSchenkel: `_fetch_next_radio_station_track`

music_assistant/providers/apple_music/__init__.py
```diff
@@ -644,6 +813,67 @@ async def get_similar_tracks(self, prov_track_id, limit=25) -> list[Track]:
...
+        self, station_id: str
+    ) -> tuple[str, str | None, str | None, int | None]:
```
MarvinSchenkel: Can we return a typed `Track` object here instead of this tuple?

music_assistant/providers/apple_music/__init__.py
```diff
@@ -485,9 +593,50 @@ async def get_track(self, prov_track_id) -> Track:
         is_favourite = rating_response.get(prov_track_id)
         return self._parse_track(response["data"][0], is_favourite)
 
-    @use_cache()
+    @use_cache(cache_checksum="stations-metadata-v2")
```
MarvinSchenkel: What is this checksum doing here?

MarvinSchenkel:
Just had a play with this, nice job so far 👍 . The stations seems to work nicely icw the dynamic playlists.
I left a few inline comments.

Furthermore, a few additional comments:
- I would prefer not to merge dead code. So for the radio's, can we implement `ProviderFeature.LIBRARY_RADIOS / get_library_radios` and actually sync them, or, if they cannot be obtained from the user's library, make them show up under browse so that users can add them to their MA library?
- For stations, it would be cool to also expose them as Recommendations, so they show up in the Discover page for the users. This way, they are not just available via browse but shown on the home page. An example can be found [here](https://github.com/music-assistant/server/blob/aff052481ff874655ff7cddbead7d90d99cfd6f2/music_assistant/providers/ytmusic/__init__.py#L662). Also happy to merge this without and do this (myself) in a follow up PR, but just wanted to let you know this exists

MarvinSchenkel (changes requested):
Please mark this as 'ready for review' once you want us to have another look.

dmoo500:
Good point on both.

**Dead code (`get_radio` / `_parse_radio`):**
Apple Music live broadcast stations (e.g. Apple Music 1) fail with error 3077 from the webPlayback endpoint — this is an account/token-level DRM restriction that we can't work around. So `get_radio` is never reachable in practice. I'll remove both `get_radio` and `_parse_radio` to keep the PR clean. If live broadcast support ever becomes possible, it can be added back then.

**Recommendations:**
Happy to defer that to a follow-up PR — the `_browse_stations()` method already fetches from `me/recommendations`, so adapting it for `get_recommendations()` should be straightforward. If you're happy to take that, feel free — otherwise I'll pick it up as a follow-up.

music_assistant/providers/apple_music/__init__.py
dmoo500: Done, extracted into `_get_station_playlist` and called from here.

music_assistant/providers/apple_music/__init__.py
dmoo500: Renamed.

music_assistant/providers/apple_music/__init__.py
dmoo500: Done, now calls `_parse_track` directly and returns a typed `Track`.

music_assistant/providers/apple_music/__init__.py
dmoo500: The checksum was added as a development-time safeguard to bust my local cached entries when `is_dynamic=True` was first introduced. Since `ra.` station support is entirely new in this PR, no existing installation has ever cached a station playlist — so there are no stale entries to invalidate. Removed.

dmoo500:
Rebased onto current `upstream/dev` (93 commits). Local testing is working again — dynamic station playlists now refill correctly and `is_dynamic` persists properly in the DB.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -488,6 +614,8 @@ async def get_track(self, prov_track_id) -> Track:
     @use_cache()
     async def get_playlist(self, prov_playlist_id, is_favourite: bool = False) -> Playlist:
         """Get full playlist details by id."""
+        if prov_playlist_id.startswith("ra."):
+            return await self._get_station_playlist(prov_playlist_id)
```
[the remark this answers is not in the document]
dmoo500: Fixed. `get_playlist` no longer has `@use_cache()` — `ra.` station IDs are routed directly to `_get_station_playlist` (no caching), while regular playlists are handled by the new `_get_regular_playlist` helper which carries the `@use_cache()` decorator. This way a transient 404 on station metadata never gets persisted in cache.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -360,15 +368,133 @@ async def search(
...
+                    )
+                image = MediaItemImage(
+                    provider=self.instance_id,
+                    type=ImageType.THUMB,
+                    path=url,
+                    remotely_accessible=True,
+                )
+        except (MediaNotFoundError, KeyError, IndexError):
+            pass
+        playlist = Playlist(
```
MarvinSchenkel: This is now duplicated logic with `_parse_station_as_playlist` right? I think this entire function can be rewritten as:

MarvinSchenkel:
```
try:
    station_response = await self._get_data(
        f"catalog/{self._storefront}/stations/{station_id}"
    )
    station_obj = station_response["data"][0]
    station_obj["id"] = station_id
    return self._parse_station_as_playlist(station_obj)
except (MediaNotFoundError, KeyError, IndexError):
    return self._parse_station_as_playlist({"id": station_id})
```

MarvinSchenkel (changes requested):
Nice improvements! One last questions and then this can be merged :)

music_assistant/providers/apple_music/__init__.py
dmoo500: Good catch — done. `_get_station_playlist` now delegates entirely to `_parse_station_as_playlist`, as suggested.

music_assistant/providers/apple_music/__init__.py
```diff
@@ -538,6 +651,21 @@ async def get_playlist_tracks(self, prov_playlist_id, page: int = 0) -> list[Tra
                 result.append(parsed_track)
         return result
 
+    async def _get_station_tracks(self, station_id: str) -> list[Track]:
+        """Fetch the next batch of tracks for a radio station."""
+        response = await self._post_data(f"me/stations/next-tracks/{station_id}")
```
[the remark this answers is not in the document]
dmoo500: Fixed — `_get_station_tracks` now passes `include="artists"` to the endpoint, consistent with how `get_similar_tracks` calls the same endpoint.

MarvinSchenkel (approved):
Looks great, thanks for this enhancement @dmoo500 🙏

PR #3435, opened 2026-03-19
Audiobookshelf: handle errors gracefully during provider unload

teancom:
Remove @handle_refresh_token decorator from unload() - during shutdown, retrying authentication is pointless and actively harmful when the server is unreachable (causes a LoginError chain through socketio reconnect).

Broaden `except` clause from ClientError to Exception because aioaudiobookshelf's exception hierarchy has no common base class - LoginError, RefreshTokenExpiredError, and others all inherit directly from Exception, not from ClientError or any shared base. The unload phase is the right place to silently accept connection failures and move on without escalating them (I think. As always, I could be wrong!).

The mass.py change was just ruff forcing a reformat, I believe.

(The connection from MA to my Audiobookshelf server is... flaky. So I've been discovering all of the ways that, as a provider, it dumps massive backtraces into the server log. This isn't Audiobookshelf's fault or the providers fault, per-se. Finding these is just a silver-lining to having some weird network issue going on.)


fmunkes:
I still don't want a bare Exception. If the provider has some bad handling of your connection errors, then we should handle them. We can adjust the lib as well, and introduce an AbsException to inherit from.

Also, if you have these network errors, then it could be a nice challenge to make reconnecting etc more robust.

If you have some kind of trace of these errors we can take a look at the log file too, e.g. in an abs thread on discord.

The reformat of mass.py should definitely not be part of this PR.

And just to mention it, I'm very happy if you are challenging the provider implementation :)

fmunkes:
@teancom I released a new version of the lib with your changes included.

After our next stable release, I might do some renaming of the ApiError in the lib itself, but would like to avoid that beforehand. However, your bugfix is certainly something we should include.

teancom:
Very happy to help! Rebased off of main, took out the mass.py formatting change, and bumped the library version we pull in.

fmunkes (approved):

fmunkes:
Thanks @teancom !

PR #3438, opened 2026-03-19
Gpodder: Handle None value for position when manually marking episodes as played

frenetic00:
When manually marking an episode as played in the web interface, the `on_played` function is called with  `position` set to  `None`.
The subsequent call to `update_progress` will throw an error when it tries to [cast](https://github.com/music-assistant/server/blob/dev/music_assistant/providers/gpodder/client.py#L297) the `None` value to an `int` and the progress won't be updated.

Tested with Nextcloud gpodder sync app.

### DIFF
--- a/music_assistant/providers/gpodder/__init__.py
+++ b/music_assistant/providers/gpodder/__init__.py
@@ -558,6 +558,10 @@ class GPodder(MusicProvider):
             return
         if time.time() - self.progress_guard_timestamp <= 5:
             return
+        if position is None and fully_played:
+            position = media_item.duration
+        elif position is None:
+            position = 0
         podcast_id, guid_or_stream_url = prov_item_id.split(" ")
         stream_url = await self._get_episode_stream_url(podcast_id, guid_or_stream_url)
         assert stream_url is not None


OzGav:
@frenetic00 there are mypy errors that need fixing. Ensure you run pre-commit before creating your PR

fmunkes:
Hi @frenetic00, thanks for wanting to contribute. Which is the MA version you encountered this Problem with? position must never be None according to this methods type hint, so we would have either a backend or frontend bug here.

fmunkes:
It is indeed a backend bug. I'm gonna mark you PR as draft for the time being, and we need to see, if this one here https://github.com/music-assistant/server/pull/3449 can be merged, which resolves the issue in core.

frenetic00:
@OzGav noted, I've seen the mypy errors now.
@fmunkes thanks for confirming, I'm using version 2.6.3 but happy to close this PR in favor of yours.

fmunkes:
We are doing a two-step approach now - the bug will be fixed in the next stable release, scheduled for tomorrow (and on beta already). This is via an assumption approach, which was easier to implement. However, after the new stable release, I'm gonna adjust the models themselves, and the providers accordingly, to have a more consistent fix.

I am gonna close this PR then, but thanks a lot for bringing that issue up, highly appreciated!

PR #3439, opened 2026-03-20
Optimize NFS provider and localfilesystem for faster import speed

apophisnow:
Fixes some issues causing MA to take a long time when importing songs from NFS.

AI fixes here, so need critical review. However I did test locally and it seems to work great.


OzGav:
Why are there changes to the party provider here as well?

apophisnow:
Messed up merge, will fix

apophisnow:
> Why are there changes to the party provider here as well?

Reverted uninintended changes.

MarvinSchenkel (approved):
Looks good, let's give it a test in beta :)

PR #3447, opened 2026-03-21
Fix provider rename not refreshing in frontend without browser reload

apophisnow:
I ran into a bug where if updating the name of a plugin the name is not immediately shown on the settings page. It required a manual page refresh.

Signal PROVIDERS_UPDATED event when a provider's name changes so frontends receive the update without requiring a full page refresh.


music_assistant/controllers/config.py
```diff
@@ -1540,6 +1540,9 @@ async def _update_provider_config(
                 self.set_provider_default_name(
                     prov_instance.instance_id, prov_instance.default_name
                 )
+            if "name" in changed_keys:
+                # signal providers updated so frontends refresh the provider name
+                self.mass.signal_event(EventType.PROVIDERS_UPDATED, data=self.mass.get_providers())
```
marcelveldt, line 1573: You can actually move this a bit below right before retuning the config.
In the end, if anything is changed, it will result in the provider changed.

OzGav:
@apophisnow Can you make the suggested tweak and we might get this into 2.8?

apophisnow:
> @apophisnow Can you make the suggested tweak and we might get this into 2.8?

Done.

MarvinSchenkel (approved):

PR #3448, opened 2026-03-21
Feature/party multi instance

apophisnow:
Companion to the frontend PR.

### DIFF
--- a/music_assistant/controllers/webserver/auth.py
+++ b/music_assistant/controllers/webserver/auth.py
@@ -49,7 +49,7 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(f"{MASS_LOGGER_NAME}.auth")
 
 # Database schema version
-DB_SCHEMA_VERSION = 5
+DB_SCHEMA_VERSION = 6
 
 # Token expiration constants (in days)
 TOKEN_SHORT_LIVED_EXPIRATION = 30  # Short-lived tokens (auto-renewing on use)
@@ -219,6 +219,7 @@ class AuthenticationManager:
                 use_count INTEGER DEFAULT 0,
                 last_used_at TEXT,
                 device_name TEXT,
+                instance_id TEXT,
                 FOREIGN KEY (user_id) REFERENCES users(user_id) ON DELETE CASCADE
             )
             """
@@ -299,6 +300,12 @@ class AuthenticationManager:
                 )
                 """
             )
+            await self.database.commit()
+
+        # Migration to version 6: Add instance_id column to join_codes
+        if from_version < 6:
+            with contextlib.suppress(OperationalError):
+                await self.database.execute("ALTER TABLE join_codes ADD COLUMN instance_id TEXT")
             await self.database.commit()
 
     async def _get_or_create_jwt_secret(self) -> str:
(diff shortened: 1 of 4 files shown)


music_assistant/controllers/webserver/auth.py
```diff
@@ -1623,6 +1654,7 @@ async def generate_join_code(
         :param expires_in_hours: Hours until code expires (default: 8).
         :param max_uses: Maximum number of uses (0 = unlimited).
         :param device_name: Device name for tokens created with this code.
+        :param instance_id: Optional provider instance ID to embed in the resulting JWT.
```
marcelveldt, line 1657:
```suggestion
        :param provider_id: Optional provider instance ID to embed in the resulting JWT.
```

marcelveldt: I fid it a bit weird to embed a provider instance ID into a JWT token but if we have no choice, then please opt for a recognizable name such as provider_id or provider_instance_id

marcelveldt:
Questions:

1) Why on earth would you want to host multiple parties at once in the same house ?
2) How would this work in the frontend ? Multiple tabs in the menu or a selection screen ?

apophisnow:
> Questions:
> 
> 1. Why on earth would you want to host multiple parties at once in the same house ?
> 2. How would this work in the frontend ? Multiple tabs in the menu or a selection screen ?

There is a one to one mapping of the party instance and the player assigned to it. Each instance has its own QR code (which ends up giving you the JWT scoped to that instance). The use case is having a dashboard with convenient QR code login for various rooms. So people hanging out in the Living Room can use the Living Room party instance, and other rooms can have their own instances with their own queues.

To be fair, I do feel some of this is compensating for the fact that we currently have no anonymous login (other than the Party plugin), and don't yet have the granular permissions required to pull something like this off.

If we did have a way for people to log in with QR code, not needing the admin to create them an account, and those users could get a view which doesn't include certain functionality (volume control, or the ability to pause or skip tracks as an example), then much of this could probably integrate better.

For addidtional context, there has been some discussion on discord around it. https://discord.com/channels/753947050995089438/1483038553360437382/1483163113720119377

marcelveldt:
What I read in that thread on discord is more a request for a more granular guest/role system than having multiple parties in one home.

If there is a real usecase and demand I agree we should look into it but it will make it more difficult to setup and maintain.

So my advice is await feedback on the current party mode first and focus on high priority/demand bugs and requests

apophisnow:
> What I read in that thread on discord is more a request for a more granular guest/role system than having multiple parties in one home.
> 
> If there is a real usecase and demand I agree we should look into it but it will make it more difficult to setup and maintain.
> 
> So my advice is await feedback on the current party mode first and focus on high priority/demand bugs and requests

I'm good with that. I'll convert this to a draft if that is OK, and we can wait and see.

MarvinSchenkel:
@apophisnow I think the way forward here is to introduce as guest mode rather than making party mode multi-instance if I recall correctly?

Hedda:
> Why on earth would you want to host multiple parties at once in the same house ?

> If there is a real usecase and demand I agree we should look into it but it will make it more difficult to setup and maintain.

@marcelveldt Having kids is the real usecase! Especially teenagagers who want to listen to their own music in their rooms.

* One use case scenario is having one teenage son and one teenage daughter each having friends over in the own rooms.

 * Another use case scenario is so adults can have party in the living/dining-room while children have their own part in one of the childrens room or if you have other kid-friendly area in your home that is better soundproofed 😜 

 I can think of several more child-related scenarios why would want to host multiple parties at once in the same house.

Have been in many real world scenarios when have friends/family for parties when the guests have brought their own children in similar ages to ours, and then the "kids" will go to their own rooms to chill and entertain themselves with music while parents stay in the living/dining-room area to in turn play other music there, (i.e. hosting multiple parties simultaneously in the same house at the time).

So I think it would still be a good idea to have multi-instance party-mode support in Music Assistant in some scenarios.

In the end you can only listen to teen's and children's music songs so many times when trying to have a party for grownups.

MarvinSchenkel:
Politely closing this PR as part of a clean up. We agreed this will eventually need a dedicated guest mode rather than a multi instance party mode

PR #3449, opened 2026-03-21
Fix 'mark item played' in music controller

fmunkes:
This one has been brought to my attention here https://github.com/music-assistant/server/pull/3438

The music provider model's on_played method is type hint as:
```python
async def on_played(self, ..., position: int, ... ) -> None:
```
but the music controller's mark_item_played method may put position = None. I went through all providers using the on_played method, and for which media_types. I had a look at the type hint expected of the respective client used in these providers:

```
  kion_music: TRACK, but doesn't care about the position
  yandex_music: TRACK, but doesn't care about the position
  spotify: checks for AUDIOBOOK & PODCAST_EPISODE , but only uses position in logging, and safe guards a None

  ard_audiothek: checks for PODCAST_EPISODE ; graphql doesn't state type hint
  audiobookshelf: checks for AUDIOBOOK & PODCAST_EPISODE # None not ok
  gpodder: checks for PODCAST_EPISODE # None not ok
  opensubsonic: checks for PODCAST_EPISODE # None not ok
  audible: AUDIOBOOK & PODCAST_EPISODE # None not ok ; needs return for PODCAST?
  bbc_sounds: TRACK, PODCAST, PODCAST_EPISODE # None not ok ; needs return for PODCAST?
```

The fix is to set position in accordance with the fully_played parameter. As far as I can tell, this doesn't break anything. I think (but I don't use these providers) that the audible and bbc_sounds providers should not care about bare PODCAST updates (only about PODCAST_EPISODES). We might want to ask the respective maintainer here.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -34,11 +34,15 @@ from music_assistant_models.errors import (
 )
 from music_assistant_models.helpers import get_global_cache_value
 from music_assistant_models.media_items import (
+    Album,
     Artist,
     AudioFormat,
     BrowseFolder,
+    Genre,
     ItemMapping,
     MediaItemType,
+    Playlist,
+    Podcast,
     ProviderMapping,
     RecommendationFolder,
     SearchResults,
@@ -1265,6 +1269,17 @@ class MusicController(CoreController):
                     params,
                     allow_replace=True,
                 )
+
+        # Set seconds_played in accordance with fully_played, if the media_item has
+        # a duration, before it is forwarded to music_providers
+        if seconds_played is None:
+            seconds_played = 0
+            if (
+                fully_played
+                and not isinstance(media_item, Album | Artist | Genre | Playlist | Podcast)
+                and isinstance(media_item.duration, int)  # for Radio duration can be None
+            ):
+                seconds_played = media_item.duration
 
         # forward to provider(s) to sync resume state (e.g. for audiobooks)
         for prov_mapping in media_item.provider_mappings:
--- a/music_assistant/providers/audible/__init__.py
+++ b/music_assistant/providers/audible/__init__.py
@@ -575,6 +575,9 @@ class Audibleprovider(MusicProvider):
 
         media_item is the full media item details of the played/playing track.
         """
+        if media_type == MediaType.PODCAST:
+            # We only care about position updates for AUDIOBOOK & PODCAST_EPISODE
+            return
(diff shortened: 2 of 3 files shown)


music_assistant/controllers/music.py
```diff
@@ -1266,6 +1270,17 @@ async def mark_item_played(
...
+        # Set seconds_played in accordance with fully_played, if the media_item has
+        # a duration, before it is forwarded to music_providers
+        if seconds_played is None:
+            seconds_played = 0
+            if (
+                fully_played
+                and not isinstance(media_item, Album | Artist | Genre | Playlist | Podcast)
+                and isinstance(media_item.duration, int)  # for Radio duration can be None
+            ):
+                seconds_played = media_item.duration
```
marcelveldt, line 1282: This is a bit cheating isnt it ?
Because you didn't actually played this but you want to manually mark the item as played

fmunkes: But this branch is taken, if we call the method with fully_played. Isn't that the same, as setting the position to the end? What would be the alternative - adjusting the model's method to allow a None for position? I was reluctant to do that, given the type hints of the respective client methods. Or fully played and position = 0, if it is not given?

marcelveldt: Yeah, dunno... maybe set position to 0 if its not given and fully_played marked is best here

fmunkes: Checked again. If I do that, then providers not taking the fully_played parameter into account, will see no difference between mark_played and mark_unplayed. This is true for audible, ard_audiothek, gpodder. Audiobookshelf needs some tweaking too. I can check the ard_audiothek provider, but I don't have audible. But honestly, if the provider doesn't take the fully_played into account, I would implement exactly what I did on the controller now on the provider level.

I see two options - either what I added to this PR, or a type hint change on the model. If we do the latter, then I will wait till after the next stable release. I really don't want to be the guy who introduced that famous last minute bug ;)

marcelveldt: Let's do a two step approach then.
Fix now with your fix here and a follow-up later with the type hint change in the models.

marcelveldt (approved):

music_assistant/controllers/music.py
fmunkes: Sounds good. I'll do the follow up next weekend.

PR #3450, opened 2026-03-21
Fix corrupt players config when player deleted

### LINKED ISSUE #5112: KeyError: 'player_id' in get_player_configs crashes Players settings page
## Bug Description

The `config/players` WebSocket command crashes with `KeyError: 'player_id'` when a player config entry exists in `settings.json` without a `player_id` field. This makes the entire Players settings page fail to load — no players are shown at all.

## Environment

- **Music Assistant Server**: 2.8.0b21 (beta)
- **Installation**: Home Assistant OS Add-on
- **Player Provider**: hass_players (Alexa Media Player integration)
- **Affected Device**: Alexa-enabled vehicle ("Justin's Ford") discovered via Alexa Media Player HA integration

## Error Log

```
ERROR (MainThread) [music_assistant.webserver] Error handling message: CommandMessage(message_id='...', command='config/players', args={'include_values': False, 'include_unavailable': False, 'include_disabled': True})
  File ".../music_assistant/controllers/webserver/websocket_client.py", line 244, in _run_handler
    result = await result
  File ".../music_assistant/controllers/config.py", line 560, in get_player_configs
    player = self.mass.players.get_player(raw_conf["player_id"], False)
KeyError: 'player_id'
```

## Steps to Reproduce

1. Have an Alexa Media Player integration in HA with a vehicle Alexa device (e.g., Ford with built-in Alexa)
2. Enable the HA Players provider in Music Assistant
3. MA discovers the vehicle Alexa device and creates a player config entry in `settings.json`
4. The created entry is missing the `player_id` field
5. Navigate to MA Settings → Players — the page shows no players

## Root Cause

In `controllers/config.py` line 560, `get_player_configs` iterates over raw configs and accesses `raw_conf["player_id"]` without checking if the key exists. If any single entry is missing `player_id`, the entire endpoint crashes and no players are returned.

## Impact

- The Players settings page is completely broken (shows zero players)
- All other player configs are inaccessible due to the crash
- The corrupted entry regenerates on restart even after manual removal from `settings.json`, because MA re-discovers the device from HA

## Workaround

Had to deregister the vehicle Alexa device entirely from the Alexa app, which removes it from HA and prevents MA from recreating the broken config. This is not ideal as it disables the vehicle's Alexa functionality.

## Suggested Fix

`get_player_configs` should gracefully skip entries missing `player_id` rather than crashing:

```python
# In config.py get_player_configs:
if "player_id" not in raw_conf:
    logger.warning("Skipping player config entry missing 'player_id': %s", key)
    continue
```

Additionally, the code that creates the player config entry from HA-discovered devices should ensure `player_id` is always set.

OzGav:
Fixes: https://github.com/music-assistant/support/issues/5112

After deleting an unavailable player from the UI, I was getting the following error and realised it was the same as the linked issue

```
ERROR (MainThread) [music_assistant.webserver] Error handling message: CommandMessage(message_id='...', command='config/players',
args={'include_values': False, 'include_unavailable': False, 'include_disabled': True})
  File ".../music_assistant/controllers/webserver/websocket_client.py", line 244, in _run_handler
    result = await result
  File ".../music_assistant/controllers/config.py", line 539, in get_player_configs
    player = self.mass.players.get_player(raw_conf["player_id"], False)
KeyError: 'player_id'
```

After some investigation, I found a corrupt entry in `settings.json` that looked like this:

```
"kodi_192.168.1.158_9090": {
  "values": {
    "protocol_parent_id": null
  }
}
```

I believe the root cause is in `_clear_protocol_parent_id()`. When a parent player is removed, this method is called to clear the `protocol_parent_id` from linked protocol players. However, if the protocol player's config has already been deleted, the `config.set()` call recreates a partial entry because `set()` uses `setdefault()` to create intermediate dictionaries.

So I did the following:

1. `protocol_linking.py`: Added a check in `_clear_protocol_parent_id()` to verify the player config exists before attempting to clear the value
2. `config.py`: Added a migration to clean up any existing corrupt player configs that are missing the `player_id` key (marked with a TODO to remove it after the 2.9 release - probably could do earlier?)

### DIFF
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -1349,6 +1349,20 @@ class ConfigController:
                 LOGGER.warning("Removed corrupt provider configuration: %s", instance_id)
                 changed = True
 
+        # Remove corrupt player configurations that are missing the required 'player_id' key
+        # This can happen when _clear_protocol_parent_id is called on an already-deleted player
+        # TODO: remove after 2.9 release
+        for player_id, player_config in list(self._data.get(CONF_PLAYERS, {}).items()):
+            if "player_id" not in player_config:
+                LOGGER.warning(
+                    "Removing corrupt player configuration (missing player_id): %s", player_id
+                )
+                self._data[CONF_PLAYERS].pop(player_id, None)
+                # Also remove any DSP config for this player
+                if CONF_PLAYER_DSP in self._data:
+                    self._data[CONF_PLAYER_DSP].pop(player_id, None)
+                changed = True
+
         # The background tasks controller originally persisted runtime state directly under
         # core/tasks, which could create a CoreConfig object without the required domain field.
         # Repair that single known corruption case on load.
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -1354,9 +1354,6 @@ class ConfigController:
         # TODO: remove after 2.9 release
         for player_id, player_config in list(self._data.get(CONF_PLAYERS, {}).items()):
             if "player_id" not in player_config:
-                LOGGER.warning(
-                    "Removing corrupt player configuration (missing player_id): %s", player_id
-                )
                 self._data[CONF_PLAYERS].pop(player_id, None)
                 # Also remove any DSP config for this player
                 if CONF_PLAYER_DSP in self._data:
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -1351,7 +1351,7 @@ class ConfigController:
 
         # Remove corrupt player configurations that are missing the required 'player_id' key
         # This can happen when _clear_protocol_parent_id is called on an already-deleted player
-        # TODO: remove after 2.9 release
+        # TODO: remove after 2.8 release
         for player_id, player_config in list(self._data.get(CONF_PLAYERS, {}).items()):
             if "player_id" not in player_config:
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/config.py
```diff
@@ -1349,6 +1349,20 @@ async def _migrate(self) -> None:  # noqa: PLR0915
                 LOGGER.warning("Removed corrupt provider configuration: %s", instance_id)
                 changed = True
 
+        # Remove corrupt player configurations that are missing the required 'player_id' key
+        # This can happen when _clear_protocol_parent_id is called on an already-deleted player
+        # TODO: remove after 2.9 release
```
marcelveldt:
```suggestion
        # TODO: remove after 2.8 release
```

music_assistant/controllers/config.py
```diff
@@ -1349,6 +1349,20 @@ async def _migrate(self) -> None:  # noqa: PLR0915
...
+        for player_id, player_config in list(self._data.get(CONF_PLAYERS, {}).items()):
+            if "player_id" not in player_config:
+                LOGGER.warning(
+                    "Removing corrupt player configuration (missing player_id): %s", player_id
+                )
```
marcelveldt:
```suggestion
```

marcelveldt (approved):
Thanks, @OzGav !

PR #3453, opened 2026-03-22
Global genre exclusion

jozefKruszynski:
Add global genre exclusion functionality to genre management page. Table with delet/exclude and restore functionality.

### DIFF
--- a/music_assistant/controllers/media/genres.py
+++ b/music_assistant/controllers/media/genres.py
@@ -1123,9 +1123,12 @@ class GenreController(MediaControllerBase[Genre]):
             DB_TABLE_GENRE_MEDIA_ITEM_EXCLUSION, {"genre_id": db_id}
         )
         if exclude_globally:
+            # Fetch the item while it is still visible (base_query filters is_excluded=1).
+            library_item = await self.get_library_item(db_id)
             await self.mass.music.database.update(
                 DB_TABLE_GENRES, {"item_id": db_id}, {"is_excluded": 1}
             )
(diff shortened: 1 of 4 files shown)


MarvinSchenkel:
Can't we just simplify this entire PR by adding a single column to the genre table `is_excluded`? We can remove a lot of redundant info + copy logic and excluding/including a genre just becomes a simple update statement?

jozefKruszynski:
I thought about that too and I can't remember why I decided against it.
I'll go back and reevaluate

music_assistant/controllers/media/genres.py
```diff
@@ -225,12 +240,22 @@ async def _add_library_item(self, item: Genre, overwrite_existing: bool = False)
...
+            )
+            self.logger.debug("restored soft-deleted genre %s (id: %s)", item.name, db_id)
+            return db_id
         db_id = await self.mass.music.database.insert(
             self.db_table,
             {
                 "name": item.name,
                 "sort_name": item.sort_name,
-                "translation_key": item.translation_key,
+                "translation_key": None,
```
MarvinSchenkel: Why is this change needed?

jozefKruszynski: Damnit, it very likely is not. I missed that this leaked through.

jozefKruszynski: Default genres are never created via _add_library_item — they go through the direct INSERT in restore_default_genres
_add_library_item is exclusively the path for custom/user-created genres
The Genre model auto-computes translation_key from the name as a convenience default, but for DB storage it must be None to mark non-default genres

MarvinSchenkel (approved):

PR #3455, opened 2026-03-23
Improve json_dumps handling for nested plain dataclasses

benklop:
## Summary
Isolates JSON helper improvements and a focused regression test so they can be reviewed independently of other work.

## Changes
- **`get_serializable_value`**: Recursively convert dataclass instances via `asdict`, walk plain dicts, fix `dict_valueiterator` detection (compare `type(obj).__name__` instead of class to string), remove duplicate `dict_values` in `isinstance`, and add a defensive `to_json()` fallback with debug logging on failure.
- **Test**: `test_multipartpath_list_serializes` ensures `json_dumps` can serialize a list of `MultiPartPath` (plain `@dataclass`) without hitting orjson default recursion when using `OPT_PASSTHROUGH_DATACLASS`.

## Context
With `OPT_PASSTHROUGH_DATACLASS`, orjson defers dataclasses to the default serializer. Plain dataclasses without mashumaro mixins (e.g. `MultiPartPath`) previously caused recursion if they appeared inside structures passed to `json_dumps`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/music_assistant/helpers/json.py
+++ b/music_assistant/helpers/json.py
@@ -31,10 +31,16 @@ def get_serializable_value(obj: Any, raise_unhandled: bool = False) -> Any:
     if getattr(obj, "do_not_serialize", None):
         return None
 
-    # Convert dataclass *instances* to dicts so nested dataclasses get handled recursively.
+    # Convert plain dataclass *instances* (without custom to_dict / mashumaro mixin)
+    # to dicts so nested dataclasses get handled recursively.
     # `is_dataclass` can also return True for dataclass *types*, so ensure we only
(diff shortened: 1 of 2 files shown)


marcelveldt:
Can you first show some evidence of the recursion because this code works fine for ages

benklop:
> Can you first show some evidence of the recursion because this code works fine for ages

to be honest, I can't really even remember why this came up. It was a few months ago, and I also no longer run into issues without it in place. I wasn't sure if it was valuable or not, but i'd just as soon forget about it if you don't think it's needed.

PR #3460, opened 2026-03-23
Schedule task to fix protocol ids > native ids in SyncGroups

MarvinSchenkel:
This PR will schedule a task that will resolve protocol ids to native ids for the use case where people have created sync groups in 2.7 with members such as 'Sonos Kitchen Airplay'. These will end up as `apxxxxxx` members after the user upgrades to 2.8. This task will auto-correct it to their native parent ids (e.g. RINCON_xxxxx).

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -159,10 +159,10 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
         """Async initialize of module."""
         self._cleanup_stale_protocol_parent_ids()
         self._poll_task = self.mass.create_task(self._poll_players())
-        self.mass.call_later(
-            300,
-            self._schedule_fix_group_member_configs,
+        self.mass.tasks.run_background_task(
             task_id="fix_group_member_configs",
+            name="Fix sync group member configurations",
+            handler=self._fix_group_member_configs,
         )
 
     async def close(self) -> None:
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/players/controller.py
```diff
@@ -158,6 +159,11 @@ async def setup(self, config: CoreConfig) -> None:
         """Async initialize of module."""
         self._cleanup_stale_protocol_parent_ids()
         self._poll_task = self.mass.create_task(self._poll_players())
+        self.mass.call_later(
+            300,
+            self._schedule_fix_group_member_configs,
+            task_id="fix_group_member_configs",
+        )
```
marcelveldt: This is kind of weird. Use call later to schedule a background task. Also (like CoPilot also suggests) the timer isnt cleared on teardown.

Either just fire a background task (which can even have a 30s sleep in it) OR just schedule manually with a timer.

music_assistant/controllers/players/controller.py
```diff
@@ -2170,6 +2176,65 @@ def _cleanup_stale_protocol_parent_ids(self) -> None:
...
+        """Fix stale protocol player IDs in sync group member configs.
+
+        When a sync group references a protocol player ID instead of
+        the parent player ID, correct it using the cached protocol parent mapping.
+        """
+        all_player_configs = self.mass.config.get(CONF_PLAYERS, {})
+        total_fixes = 0
+        fixed_groups: list[str] = []
+
+        for group_id, group_config in all_player_configs.items():
```
marcelveldt:
```suggestion
        for group_id, group_config in list(all_player_configs.items()):
```

guard against the collection being modified when iterating

music_assistant/controllers/tasks/controller.py
```diff
@@ -275,7 +277,14 @@ def run_background_task(  # noqa: PLR0913
             max_log_lines=max_log_lines,
         )
         self._tasks[task_info.id] = managed
-        self._queue_task(managed, reset_logs=True, run_user_id=user_id)
+        if delay is not None and delay > 0:
+            self.mass.call_later(
+                delay,
+                partial(self._queue_task, managed, reset_logs=True, run_user_id=user_id),
```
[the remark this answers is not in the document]
marcelveldt: this is not resolved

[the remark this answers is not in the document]
marcelveldt: the correct way of doing this is calling "register_scheduled_task" with a "one off" schedule.
this delay in the "run_background_task" feels like a hack.

You want to schedule a task but you are now abusing the "run background task now" logic with a delay so you now have two ways to schedule a task. That is confusing and hacky. Lets fix it properly

What we really need is extending TaskScheduleType with a ONE_TIME (and while at it, we could maybe also add AT_BOOT or something) so you can really schedule this task in the future.

If you want to avoid making all these changes now, you can also simply go with a DAILY schedule with the time set to now + 5 minute sin the future. And then do a follow-up PR later to fix this to use ON_BOOT instead or a one off / one time schedule.

music_assistant/controllers/players/controller.py
```diff
@@ -158,6 +160,17 @@ async def setup(self, config: CoreConfig) -> None:
         """Async initialize of module."""
         self._cleanup_stale_protocol_parent_ids()
         self._poll_task = self.mass.create_task(self._poll_players())
+        self.mass.tasks.register_scheduled_task(
+            task_id="fix_group_member_configs",
+            name="Fix sync group member configurations",
+            handler=self._fix_group_member_configs,
+            schedule=TaskSchedule.weekly(
```
marcelveldt, line 167: This doesn't make a lot of sense to me...
hour should be the current hour in UTC
minute should be the current minute in UTC + 5 min

marcelveldt (approved):
Approved for now but let's fix this properly in a follow-up PR

PR #3461, opened 2026-03-23
Fix audio stream memory leaks on playback cancellation

maximmaxim345:
Cancelling playback mid-stream could leak memory because several cleanup paths were missing.
Buffer consumers would hang after `clear()`, while producers kept running indefinitely, and upstream generators were never closed.


music_assistant/helpers/audio.py
```diff
@@ -435,6 +435,13 @@ async def get_buffered_media_stream(
     # checksum based on filter_params
     checksum = f"{filter_params}"
 
+    async def clear_abandoned_buffer(audio_buffer: AudioBuffer) -> None:
```
marcelveldt: How do you define if its abandoned ? After a certain timeout ?
Note that the queue controller also schedules a cleanup for current index -1 

We need to handle the situation where the playback is stopped and then quickly resumed afterwards, such as when seeking or transferring playback. So we should only destroy a buffer if its really out of context

marcelveldt:
@maximmaxim345 can you review the comments please and either solve or resolve them ?

marcelveldt (approved):
Nice work @maximmaxim345 !

PR #3464, opened 2026-03-24
Fix Cast autoplay

OzGav:
While investigating https://github.com/music-assistant/support/issues/4574 I wasn't able to recreate that problem but I discovered a different problem associated with the auto power on feature.

When the "Automatically play/resume on power on" setting is enabled for a Chromecast player, starting playback via `media_player.media_play` (or  from the MA interface) while the player is powered off causes playback to fail. The player shows a spinning circle indefinitely and never starts playing. Logs show two play_media commands being issued ~2 seconds apart for the same track.

Looking at the cast provider I believe the code path is:
1. cmd_play() → player_queues.resume() → play_index() → play_media() → _handle_play_media()
2. _handle_play_media() powers on the player by calling _handle_cmd_power()
3. At the end of _handle_cmd_power(), the auto-play feature checks if it should start playback - but it doesn't know that playback was already being initiated by the caller
4. This triggers a second player_queues.resume() call, causing two concurrent playback attempts that interfere with each other

I have fixed this by adding a `skip_auto_play` parameter to `_handle_cmd_power()`. When` _handle_play_media()` powers on the player as part of starting playback, it now passes `skip_auto_play=True` to prevent the redundant auto-play from triggering.

This may also resolve the intermittent "await wasn't used with future" errors reported in #4574, but as I said I couldnt recreate it to test...?

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -2677,11 +2677,17 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
         # fallback: just try to resume queue playback
         await self.mass.player_queues.resume(player.player_id)
 
-    async def _handle_cmd_power(self, player_id: str, powered: bool) -> None:
+    async def _handle_cmd_power(
+        self, player_id: str, powered: bool, skip_auto_play: bool = False
+    ) -> None:
         """
         Handle player power on/off command.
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
Nice find, great work

PR #3468, opened 2026-03-24
Fix single artist in ARTISTS tag with semi-colon in name

OzGav:
Further tweak required to the tag parsing as Picard always sets the ARTISTS tag even for single artist tracks


MarvinSchenkel (approved):
Thanks @OzGav !

PR #3476, opened 2026-03-25
Upgrade stages of various providers for 2.8

OzGav:


music_assistant/providers/roku_media_assistant/manifest.json
```diff
@@ -8,5 +8,5 @@
   "requirements": ["async-upnp-client==0.46.2","rokuecp==0.19.5"],
   "upnp_discovery": ["roku:ecp"],
   "documentation": "https://music-assistant.io/player-support/roku/",
-  "stage": "beta"
+  "stage": "stable"
```
marcelveldt, line 11: is it ? as far as I'm aware this is not super straight forward and we should mark it experimental ?

OzGav: I don’t use it but we have had no issues reported. It is not Alexa level of hard. It appears just an app needs to be added to the device? https://beta.music-assistant.io/player-support/roku/

marcelveldt (approved):

PR #3492, opened 2026-03-27
Twitch music provider

Drizzt321:
Created a Twitch Music Provider, to listen to `audio_only` streams. Primarily for DJ streams, but could probably be used for any stream.

I'm not a Python developer, and I heavily used Claude Code, so while I did look at things, I'm more a Java developer, rather than Python. 

Website docs update: https://github.com/music-assistant/music-assistant.io/pull/561


OzGav:
Replacing the ads with silence violates Twitch’s terms and conditions. This will not be allowed.

Drizzt321:
> Replacing the ads with silence violates Twitch’s terms and conditions. This will not be allowed.

Ah, I should have thought and considered. I'll go remove that part then.

Drizzt321:
@OzGav ok, silence removed for Ad breaks. Initial testing is fine, but I need to wait for a longer playback for seeing how it handles mid-stream, although I don't expect any issues.

OzGav:
Is there a clean Streamlink session option to guarantee ads are always passed through as audio, without any monkey-patching of internals?

If the answer is no — that there’s no clean option and patching internals is the only way — then the question becomes whether Streamlink is the right tool at all, given that direct HLS fetching via aiohttp would give full control over ad segment handling without any of this complexity.​​​​​​​​​​​​​​​​

Drizzt321:
So looking at Streamlink, without the monkey patch, it'd just no do any data at all (e.g. silence with no audio data, so interrupted stream) rather than pass through the audio. With the previous version, there was the option to pass no audio segments to keep the stream data intact, just with silence. Without that, it just passes through the Ad audio data. So yes, it's still needed unfortunately.

music_assistant/providers/twitch/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "music",
+  "domain": "twitch",
+  "stage": "beta",
+  "name": "Twitch Audio",
+  "description": "Audio-only streaming from Twitch channels with raid following.",
+  "codeowners": ["@Drizzt321"],
+  "requirements": ["streamlink>=8.0,<9"],
+  "documentation": "",
+  "multi_instance": false,
```
MarvinSchenkel: Is this set to `false` on purpose?

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+        if media_item is not None:
+            for pm in getattr(media_item, "provider_mappings", []):
+                if getattr(pm, "provider_domain", "") == self.domain:
+                    item_id = getattr(pm, "item_id", "")
+                    if item_id:
+                        return item_id
+
+        return None
+
+    async def _on_queue_updated(self, event: Any = None) -> None:
```
MarvinSchenkel: Why do we need to track the queue for raids? Can't we simply just call play_media when we receive a raid event (`on_raid`)?

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+            provider=self.instance_id,
+            item_id=item_id,
+            audio_format=AudioFormat(content_type=ContentType.UNKNOWN),
+            media_type=MediaType.RADIO,
+            stream_type=StreamType.CUSTOM,
+            allow_seek=False,
+            can_seek=False,
+        )
+
+    async def get_audio_stream(
```
MarvinSchenkel, line 716: Why can't we simply rely on streamlink resolving the urls and use streamurls rather than chunking ourselves? Are there no ads in those urls?

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+            prev_ad_state = False
+            try:
+                while True:
+                    chunk = await asyncio.to_thread(fd.read, STREAM_CHUNK_SIZE)
+                    if chunk:
+                        reconnects = 0
+                        if _ah.ad_break_active != prev_ad_state:
+                            prev_ad_state = _ah.ad_break_active
+                            if _ah.ad_break_active:
+                                streamdetails.stream_title = f"{item_id} - Ad Break"
```
MarvinSchenkel: I think you need to set the `streamdetails_stream_title` back to the previous title, otherwise it will remain to show 'Ad break' after the ad is done?

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+                return
+
+            await asyncio.sleep(RECONNECT_DELAY)
+
+    def _resolve_streams(self, channel: str) -> dict[str, Any] | None:
+        """Resolve Streamlink streams for a channel. Blocking — call via to_thread."""
+        from streamlink import Streamlink  # type: ignore[attr-defined]  # noqa: PLC0415
+
+        try:
+            if self._streamlink_session is None:
```
MarvinSchenkel: `self._streamlink_session` should be a local var. This function is called from to_thead and mutating `self` is not thread safe. Then again, it seems that `self._streamlink_session` is only used within this function anyway so you could just turn this into a local variable.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+        while True:
+            streams = await asyncio.to_thread(self._resolve_streams, item_id)
+            if not streams:
+                return
+
+            stream = self._select_quality(streams)
+            if not stream:
+                return
+
+            import music_assistant.providers.twitch.ad_handling as _ah  # noqa: PLC0415
```
MarvinSchenkel: Move this import to the top

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+            # Apply ad handling monkey-patch to the ACTUAL reader class from
+            # Streamlink's plugin system. Must be done after streams() because
+            # Streamlink loads plugins into a fresh module namespace — patching
+            # the imported class at startup patches a different class object.
+            result = dict(streams)
+            any_stream = next(iter(result.values()), None)
+            if any_stream is not None:
+                reader_cls = getattr(type(any_stream), "__reader__", None)
+                if reader_cls is not None:
+                    from music_assistant.providers.twitch.ad_handling import (  # noqa: PLC0415
```
MarvinSchenkel: Move this to the top

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+                        patch_ad_handling,
+                    )
+
+                    patch_ad_handling(reader_cls=reader_cls)
```
MarvinSchenkel: Why do we do this in a loop and not just once in the handle_async_init? This seems wasteful.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,1004 @@
...
+        if self._eventsub is not None:
+            await self._eventsub.stop()
+            self._eventsub = None
+
+        # Clear cache
+        self._clear_cache()
+
+    # --- Raid State Machine ---
+
+    def _extract_twitch_login(self, queue_item: Any) -> str | None:
```
MarvinSchenkel: Why are you using get_attr on typed MA objects? For example, you can just use `queue_item.uri`

music_assistant/providers/twitch/eventsub.py
```diff
@@ -0,0 +1,249 @@
+"""EventSub WebSocket client for Twitch raid following."""
+
+# mypy: disable-error-code="unreachable"
```
MarvinSchenkel: Why are you disabling this mypy error here?

MarvinSchenkel (changes requested):
Please have a look at my comments. Marking this PR as draft so we can keep track of things, please mark it as ready for review again when you want us to take another look 🙏 .

music_assistant/providers/twitch/__init__.py
Drizzt321: The original intent is that only subscribe to raids for the currently playing channel. Rather than subscribing to all Live Channels and trying to figure out which one is the one that is currently playing. Alternatively, if someone switches away to a different channel or even different type of track (Spotify, SoundCloud, etc) entirely, that's the only way to be notified in order to unsubscribe to that channels raids. Was the first solution that was used, but exploring further, not the best solution. Will be changing this to track for raids differently.

music_assistant/providers/twitch/manifest.json
Drizzt321: Misunderstood what this was, you're right, should be left to default or set to true.

music_assistant/providers/twitch/__init__.py
Drizzt321: Reworked the raid handling entirely, using the audio stream lifecycle better/properly now.

music_assistant/providers/twitch/__init__.py
Drizzt321: There are a few things in Streamlinks HLS processing that are needed, that we can't bypass by simply handing the HLS URL to MusicAssistant. 

1. Streamlink's default `TwitchHLSStreamWriter.should_filter_segment()` silently drops the embedded Ad audio segments (what the monkey-patch fixes), so no audio data flows. So this could end up being a dead stream/playback halts, which could cause problems beyond not passing through the Ad audio.
2. The HLS segment URLs are ephemeral and authenticated (if the user provides the logged in token, to get fewer/no Ads for subscribers), and this gets injected into the segment requests. 
3. As well, if there's a stream interruption, Streamlink properly/automatically re-connets/re-resolves things.

music_assistant/providers/twitch/__init__.py
Drizzt321: So in my actual observed experience, the Ad Break actually did disappear. The else block setting it to None actually ends up updating and resetting it. Added some comments to make it clearer.

music_assistant/providers/twitch/__init__.py
Drizzt321: This has been fixed, and Streamlink instances are explicitly **not** thread-safe, so doing one per queue session.

music_assistant/providers/twitch/__init__.py
Drizzt321: Since each Streamlink instance is not thread-safe, this now is created at the top of each `get_audio_stream` call, and reused within that call. So each queue/time that that gets called, we create the new instance as needed.

music_assistant/providers/twitch/__init__.py
Drizzt321: Fixed

music_assistant/providers/twitch/eventsub.py
Drizzt321: Was for a couple of specific false positives, shouldn't have been at the file level, should have been inline. Added comment as to why it's needed at those 2 inline spots.

Drizzt321:
Sorry for the delay, it's been a busy week. Should have your concerns handled now. Hopefully doesn't raise new ones :)

Drizzt321:
Updated as per Copilot reivew.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,946 @@
...
+                stream = live_by_login.get(login)
+                radio = self._channel_to_radio(ch, stream, profiles.get(login))
+                if not stream:
+                    radio.name = f"{ch['broadcaster_name']} (offline)"
+                result.append(radio)
+            return result
+
+        return []  # pragma: no cover
+
+    async def get_library_radios(self) -> AsyncGenerator[Radio, None]:
```
OzGav, line 866: The current approach with both `browse()` and `get_library_radios()` implemented is not correct. We recently standardised the radio provider pattern — if a provider declares `LIBRARY_RADIOS` and implements `get_library_radios()`, the base class handles browse automatically by calling the `get_library_*()` methods. You should not also be implementing `browse()` or declaring `BROWSE `in `SUPPORTED_FEATURES`.
The correct pattern is:

`get_library_radios()` yields all followed channels (all of them) and MA's library sync takes care of adding and removing from the library automatically
`recommendations()` is the right place to surface currently live channels since it is surfaced to the user on demand

Users therefore have all of their library from the provider in the MA library and use the search to find more.

Please remove `browse()` and remove `BROWSE` from `SUPPORTED_FEATURES`.
If you have a specific use case that genuinely requires a dedicated browse implementation that cannot be served by this pattern, please raise it and we can discuss, but our strong preference is to maintain the standardised approach across all providers."

OzGav: To be clear. You are filtering out non-live channels but this method is only called on each full sync of the library at the specified interval, so that could be hours old. A channel that was live when the sync ran will appear in the library, and one that wasn't live will be missing — even though the user follows both. The live status enrichment (viewer count in the name, live thumbnail) will also be stale by the time the user sees it, so it has no value here. The library should be a stable list of all followed channels with basic profile info and nothing more.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,946 @@
...
+            msg = f"Twitch API: rate limited ({status})"
+            raise ResourceTemporarilyUnavailable(msg)
+        if status in (401, 403):
+            msg = f"Twitch API: authentication failed ({status})"
+            raise LoginFailed(msg)
+        if status >= 500:
+            msg = f"Twitch API: server error ({status})"
+            raise ProviderUnavailableError(msg)
+        msg = f"Twitch API error ({status})"
+        raise MusicAssistantError(msg)
```
OzGav, line 517: The other thing here is that for unhandled status codes it raises MusicAssistantError directly — this is the base exception and should not be used directly. `ProviderUnavailableError` or `SetupFailedError` would be appropriate for an unexpected HTTP status I would think.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,946 @@
...
+
+    # Best-effort revoke — clear local state even if revoke fails
+    if access_token:
+        try:
+            async with mass.http_session.post(
+                TWITCH_REVOKE_URL,
+                data={"client_id": client_id, "token": access_token},
+            ):
+                pass
+        except Exception:
```
OzGav, line 167: There are nine of these. You need to be catching the specific errors you are looking for to ensure they are handled correctly.

music_assistant/providers/twitch/__init__.py
```diff
@@ -0,0 +1,946 @@
...
+        timer = self._unsubscribe_timers.pop(channel_login, None)
+        if timer is not None:
+            timer.cancel()
+
+        prev_count = self._active_streams.get(channel_login, 0)
+        self._active_streams[channel_login] = prev_count + 1
+        self.logger.debug("Stream started for %s (active: %d)", channel_login, prev_count + 1)
+
+        if prev_count == 0:
+            asyncio.create_task(self._subscribe_raids_for_channel(channel_login))
```
OzGav, line 404: When you call `asyncio.create_task()` without storing the result, any exceptions raised by that task are silently lost with no way to know something went wrong. Look at how how other providers do this — please apply the same pattern to the other `create_task()` calls in `eventsub.py`

music_assistant/providers/twitch/eventsub.py
```diff
@@ -0,0 +1,268 @@
...
+EVENTSUB_WS_URL = "wss://eventsub.wss.twitch.tv/ws"
+MAX_BACKOFF = 60.0
+
+
+class EventSubClient:
+    """Async EventSub WebSocket client for channel.raid subscriptions."""
+
+    def __init__(
+        self,
+        http_session: Any,
```
OzGav, line 26: Typing this as Any is defeating mypy type checking. I think this should be aiohttp.ClientSession and below ws should be aiohttp.ClientWebSocketResponse

music_assistant/providers/twitch/eventsub.py
```diff
@@ -0,0 +1,268 @@
...
+            url = self._reconnect_url or EVENTSUB_WS_URL
+            self._reconnect_url = None  # consume after use
+
+            try:
+                self._ws = await self._http_session.ws_connect(url)
+                async for msg in self._ws:
+                    if self._stopped:
+                        # mypy false positive (python/mypy#12784): async for + except
+                        # CancelledError makes mypy think the loop body is unreachable
+                        break  # type: ignore[unreachable]
```
OzGav, line 176: You need to fix these rather than ignore them. If you fix the types I mentioned before then somthing like this should work (so look at all the ignores again you will be able to fix them)
```
try:
    self._ws = await self._http_session.ws_connect(url)
except aiohttp.ClientError:
    logger.debug("EventSub: WebSocket connection failed", exc_info=True)
    continue

async for msg in self._ws:
    if self._stopped:
        break
    ...
```

tests/providers/twitch/test_twitch_api.py
```diff
@@ -0,0 +1,297 @@
...
+    # Expire the cache
+    provider._cache_time = time.monotonic() - 301  # past 5 min TTL
+
+    await provider._get_followed_live_status()
+    assert provider.mass.http_session.get.call_count > call_count_after_first
+
+
+async def test_cache_cleared_on_logout(provider: TwitchProvider) -> None:
+    """Logout invalidates the cache."""
+    provider._cached_channels = [{"broadcaster_id": "123"}]
```
OzGav, line 221: How come you aren't using MA's cache system?

OzGav:
General comment on the docstrings. They should provide clarity to the caller, not explain implementation. Several docstrings describe internal mechanics, e.g.: _create_streamlink_session: "Blocking — call via to_thread" is an implementation note, not caller-facing
The format needs to be Sphinx-style. e.g.
```python
def my_function(param1: str, param2: int, param3: bool = False) -> str:
    """
    Brief one-line description of the function.

    :param param1: Description of what param1 is used for.
    :param param2: Description of what param2 is used for.
    :param param3: Description of what param3 is used for.
    """
```

MarvinSchenkel:
So I just did some digging in the [TOS](https://legal.twitch.com/legal/developer-agreement/) and I am afraid we cannot merge this PR because of it :(. Specifically:

- D1 - _Embeddable Experiences: You must adopt and integrate our embeddable player for all Twitch video content. **You may not use any other player or system to display Twitch video content without Twitch’s prior written permission.**_
- X1.I: _You will not access undocumented Program Materials or otherwise attempt to derive or use the underlying source code of undocumented Program Materials without Twitch’s prior written permission. You may only access data from Program Materials consistent with the terms of this Agreement and, unless you have Twitch’s prior written permission, will only access Program Materials documented on the Twitch_. StreamLink seems to use undocumented GraphQL API's

What is your interpretation of this?

Drizzt321:
Sorry for how delayed getting back to you I have been, work has slammed me. I should be able to pick this back up in a few weeks.

Drizzt321:
@MarvinSchenkel so, I hadn't even considered the TOS, but you're almost certainly right. It's highly unlikely I'd ever be able to get Twitch to sign off on this, so I'll close this since can't be done to contribute :(

PR #3494, opened 2026-03-27
Fix podcasts from filesystem source not appearing in library

teancom:
Add in_library=True to the Podcast ProviderMapping in the filesystem provider. Without this flag, podcasts were saved to the database but invisible to library queries (which filter on in_library=1), preventing them from appearing in the UI and causing them to be re-processed on every restart.


marcelveldt (approved):

teancom:
Added another fix as per feedback on the first one via Discord.

PR #3495, opened 2026-03-27
bump bandcamp to 0.1.1

ALERTua:
via https://github.com/music-assistant/support/issues/5149


ALERTua:
Um... I did `uv sync --dev --upgrade`, so all packages are latest in the `uv.lock` and in the built PyPI package.
Should I do something with these warnings?

ALERTua:
superseded by https://github.com/music-assistant/server/pull/3496

PR #3506, opened 2026-03-29
Add bpm and key

apophisnow:
If this looks right to you feel free to use it, but if you had something else in mind fine by me if you scrap it. I have this working with frontend changes to support the Tidal key and bpm metadata.


MarvinSchenkel:
Bpm and Key are not part of `MediaItem` / `Track` level, but rather on the actual stream level (the audio). I am currently working on a new type of provider called `AudioAnalysisProvider` which will capture this information automatically by analyzing the pcm audio. 

These metrics will be stored in an [`AudioAnalysis` model](https://github.com/music-assistant/server/blob/8b34aef84398913e901b89d01bb97e2c192d535a/music_assistant/models/audio_analysis.py#L23), but this on the `Track's provider` level, not the `Track`.

However, I think it makes sense to leverage the information when music providers provide them, so I was thinking it makes sense that we introduce a new `ProviderFeature.AUDIO_ANALYSIS` that we call during a music provider sync in order to grab those metrics and store them in the `AudioAnalysis` object without having to do any pcm audio analysis. Then we can also read those metrics from ID3 tags on local files

apophisnow:
Sounds good, just going to close this them. Cheers

PR #3509, opened 2026-03-30
Add Audio Analysis controller and Audio Analysis provider

MarvinSchenkel:
Depends on https://github.com/music-assistant/models/pull/195


music_assistant/controllers/streams/audio_buffer.py
```diff
@@ -404,11 +418,15 @@ async def get_buffer(
...
         if seek_position_ms == 0:
+            # TODO: Remove loudness / smart fades after they have been implemented as
+            # audio analysis providers
             # loudness analysis for all streams (tracks and radio)
             mass.streams.audio.attach_loudness_analyzer(audio_buffer, streamdetails)
             # smart fades analysis only for music tracks (not podcasts/audiobooks)
             if streamdetails.media_type == MediaType.TRACK:
                 mass.streams.smart_fades_analyzer.attach_to_buffer(audio_buffer, streamdetails)
+            # audio analysis providers (beat tracking, key detection, etc.)
+            await mass.streams.audio_analysis.start_analysis(audio_buffer, streamdetails)
```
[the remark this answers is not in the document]
MarvinSchenkel: This won't guarantee that callbacks are registered at the time the first chunks are sent

music_assistant/controllers/streams/audio.py
```diff
@@ -875,22 +875,22 @@ def attach_loudness_analyzer(
...
-                chunk_queue.put_nowait(None)
+                await chunk_queue.put(None)
                 return
             chunks_received += 1
-            chunk_queue.put_nowait(pcm_data)
+            await chunk_queue.put(pcm_data)
             if chunks_received >= max_duration_seconds:
                 # signal we have enough data
-                chunk_queue.put_nowait(None)
+                await chunk_queue.put(None)
```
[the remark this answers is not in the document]
MarvinSchenkel: @marcelveldt  this was exactly my doubt with making this async. Thoughts?

music_assistant/controllers/streams/smart_fades/analyzer.py
```diff
@@ -65,10 +65,10 @@ def attach_to_buffer(
...
         intro_analyzed = False
 
-        def _on_chunk(position_seconds: int, pcm_data: bytes) -> None:  # noqa: ARG001
+        async def _on_chunk(position_seconds: int, pcm_data: bytes, is_last_chunk: bool) -> None:  # noqa: ARG001
             nonlocal intro_analyzed
 
-            if not pcm_data:
+            if is_last_chunk:
                 # EOF — trigger outro analysis
                 if outro_chunks:
```
marcelveldt, line 73: so, this can be optimized in the future by pushing chunks directly into the analyzer as they come in so you don't have to store (yet another) copy of the audio first before you can even do the analyze.

ofcourse this can only be done for the into analyse because of routro you need to wait until you have the last part complete

music_assistant/controllers/streams/audio.py
```diff
@@ -875,22 +875,22 @@ def attach_loudness_analyzer(
         media_type = streamdetails.media_type
         pcm_format = audio_buffer.pcm_format
 
-        chunk_queue: asyncio.Queue[bytes | None] = asyncio.Queue()
+        chunk_queue: asyncio.Queue[bytes | None] = asyncio.Queue(maxsize=10)
```
marcelveldt, line 885: we need to rewrite this so we can push the chunks directly into the ffmpeg process that is doing the analyse. with the -ss parameter we can limit to a certain duration

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+            for prov in self.mass.get_providers(ProviderType.AUDIO_ANALYSIS)
+            if isinstance(prov, AudioAnalysisProvider) and prov.available
+        ]
+
+    async def start_analysis(
+        self,
+        audio_buffer: AudioBuffer,
+        stream_details: StreamDetails,
+    ) -> None:
+        """Start analysis session for a track across all providers.
```
marcelveldt:
```suggestion
        """
        Start analysis session for a track across all providers.
```

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+
+        :param audio_buffer: The AudioBuffer to observe for PCM chunks.
+        :param stream_details: The stream details for the item being analyzed.
+        """
+        providers = self.providers
+        if not providers:
+            self.logger.debug("No audio analysis providers available")
+            return
+
+        session_key = (
```
marcelveldt: why not use stream_details.uri ?

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+        async def _on_chunk(position_seconds: int, pcm_data: bytes, is_last_chunk: bool) -> None:  # noqa: ARG001
+            nonlocal finalized
+            if finalized:
+                return
+            if is_last_chunk:
+                finalized = True
+                await queue.put(None)
+                self.mass.create_task(_finalize_session())
+                return
+            await queue.put(pcm_data)
```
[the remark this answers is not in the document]
marcelveldt: we just need to make sure that we catch slow handlers. the maxsize 10 gives us a 10 seconds headroom.
Its fine if the buffer filling is a bit slowdown due to analyzing - a long as we abort if its too slow

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+            if stored_version is not None and stored_version >= provider.analysis_version:
+                self.logger.debug(
+                    "Analysis already exists for provider %s (version %d >= %d), skipping",
+                    provider.name,
+                    stored_version,
+                    provider.analysis_version,
+                )
+                continue
+            try:
+                await provider.start_analysis(
```
marcelveldt: why doesnt this call return a boolean if its accepted or not ?
that way you can leave the logic to the provider

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+                    self.logger.warning(
+                        "Provider %s timed out processing chunk for %s, removing from session",
+                        pid,
+                        session_key,
+                    )
+                    timed_out.add(pid)
+                except Exception as err:
+                    self.logger.warning("Error processing PCM chunk on provider %s: %s", pid, err)
+
+            await asyncio.gather(*[_process(pid) for pid in provider_ids])
```
marcelveldt: why arent you collecting the exceptions here instead of that timed_out set ?

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,233 @@
...
+import contextlib
+import logging
+from typing import TYPE_CHECKING
+
+from music_assistant_models.enums import ProviderType
+
+from music_assistant.constants import MASS_LOGGER_NAME
+from music_assistant.models.audio_analysis_provider import AudioAnalysisProvider
+
+CHUNK_PROCESS_TIMEOUT = 5
```
marcelveldt:
```suggestion
CHUNK_PROCESS_TIMEOUT = 0.8
```

I would propose a timeout of 800ms per chunk so you stay below 1 second (as each chunk is 1 second)

music_assistant/models/audio_analysis.py
```diff
@@ -0,0 +1,44 @@
...
+from dataclasses import dataclass, fields
+
+import numpy as np
+import numpy.typing as npt
+from mashumaro import DataClassDictMixin
+from mashumaro.config import BaseConfig
+
+
+@dataclass(kw_only=True)
+class AudioAnalysisData(DataClassDictMixin):
```
marcelveldt, line 15: should we make loudless analyze also a provider ?

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+        into the running MusicAssistant instance. Use it to subscribe to events,
+        set up background tasks, or perform other initialization that requires
+        the full MA instance to be available.
+        """
+        self.logger.debug("Demo Audio Analysis Provider loaded in Music Assistant")
+
+    async def start_analysis(
+        self,
+        session_id: str,
+        stream_details: StreamDetails,
```
marcelveldt: I'm pretty sure in other parts of the code we use "streamdetails" (one word) instead of "stream_details"

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+            session_id,
+            stream_details.uri,
+            audio_format.sample_rate,
+            audio_format.bit_depth,
+            audio_format.channels,
+        )
+
+    async def process_pcm_chunk(
+        self,
+        session_id: str,
```
marcelveldt, line 144: why introduce a new concept - why not the uri or the streamdetails object itself

OzGav:
WIll need some docs for this before it hits the beta

MarvinSchenkel:
> WIll need some docs for this before it hits the beta

This is just internal stuff, the actual providers built on top of this will probably need docs? Or do we have generic docs on things like 'player providers', 'music sources' etc?

music_assistant/controllers/streams/smart_fades/analyzer.py
MarvinSchenkel: Smart fades will be rewritten as an AA provider, fully streaming. Already have a draft ready.

music_assistant/controllers/streams/audio.py
MarvinSchenkel: Agreed, let's do this as part of the rewrite of the loudness analyzer as an AA provider.

OzGav:
We do have generic pages for those things but only in so far as what affects the user directly. That is what they can see or what then need to know about the inner workings to understand what to expect.

music_assistant/controllers/streams/audio_analysis.py
MarvinSchenkel: Because we want:
- To log a warning about WHICH provider got evicted
- We also want to call `cancel()` on that specific provider for cleanup

That said, I did rewrite it a bit so we don't have to pass that set

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
MarvinSchenkel: Didn't know that was the same identifier, adjusted!

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
MarvinSchenkel: Renamed

music_assistant/controllers/streams/audio_analysis.py
MarvinSchenkel: Rewritten as a hook, so providers can also write their own logic that decides whether to accept/reject a session. Version check now is done in the base class.

music_assistant/controllers/streams/audio.py
```diff
@@ -948,6 +948,11 @@ async def _attach() -> None:
                 return
             self.logger.debug("Attached loudness analyzer to buffer for %s", streamdetails.uri)
             audio_buffer.register_chunk_callback(_on_chunk)
+
+            def _on_cancel() -> None:
+                chunk_queue.put_nowait(None)
```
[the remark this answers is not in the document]
marcelveldt: this is valid feedback

music_assistant/controllers/streams/audio.py
marcelveldt: yes please do that as soon as this foundational PR is in

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,234 @@
...
+        async def _on_chunk(position_seconds: int, pcm_data: bytes, is_last_chunk: bool) -> None:  # noqa: ARG001
+            nonlocal finalized
+            if finalized:
+                return
+            if is_last_chunk:
+                finalized = True
+                await queue.put(None)
+                self.mass.create_task(_finalize_session())
+                return
+            await queue.put(pcm_data)
```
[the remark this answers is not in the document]
marcelveldt: as long as we can guarantee that the chunk consumers do not block too long this is fine. dropping chunks is also not great as that renders the analyze unusable

music_assistant/models/audio_analysis.py
marcelveldt, line 15: I'm wondering why this should be stored as NDArray

marcelveldt, line 15: So, this makes it extremely inflexible to add new providers as the model needs to be adjusted and you run into schema mismatches.

Why is this not a simply dict[str, Any] or from my part a dict[str, SerializableType]
that way new analyse providers are more flexible in adding extra fields as they wish

you can then also drop the DataClassDictMixin if you only allow standard serializable objects in this dict

music_assistant/models/audio_analysis.py
```diff
@@ -0,0 +1,44 @@
...
+    """Audio analysis data using standard MIR (Music Information Retrieval) concepts.
```
marcelveldt: Is it ? because that makes it inflexible.
I want to be able to do any audio analysis with this new provider type, one of them being MIR compliant concepts but also silence detection, loudness measurement etc.

I think this current approach is too narrow, we should make it more flexible/universal

music_assistant/controllers/music.py
```diff
@@ -1180,6 +1182,114 @@ async def get_smart_fades_analysis(
...
+        :param aa_provider_domain: Domain of the AA provider that produced the data.
+        :param analysis: The analysis data to store.
+        :param analysis_version: Version of the AA provider's algorithm.
+        :param media_type: The media type of the item being analyzed.
+        """
+        if not (provider := self.mass.get_provider(provider_instance_id_or_domain)):
+            return
+        prov_key = provider.domain if provider.is_streaming_provider else provider.instance_id
+        data_json = json_dumps(analysis.to_dict())
+        await self.database.insert_or_replace(
```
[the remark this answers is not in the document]
marcelveldt: copilot is not wrong here. depending on the size of these blobs, doing the json dumps in an executor is safer

music_assistant/controllers/streams/audio.py
```diff
@@ -882,22 +882,22 @@ def attach_loudness_analyzer(
...
-                chunk_queue.put_nowait(None)
+                await chunk_queue.put(None)
                 return
             chunks_received += 1
-            chunk_queue.put_nowait(pcm_data)
+            await chunk_queue.put(pcm_data)
             if chunks_received >= max_duration_seconds:
                 # signal we have enough data
-                chunk_queue.put_nowait(None)
+                await chunk_queue.put(None)
```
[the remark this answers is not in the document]
marcelveldt: this is covered by a timeout on the chunk processing itself

music_assistant/controllers/streams/controller.py
```diff
@@ -102,6 +103,7 @@ def __init__(self, mass: MusicAssistant) -> None:
         self._bind_ip: str = "0.0.0.0"
         self.audio = StreamsAudio(mass)
         self._smart_fades_analyzer = SmartFadesAnalyzer(self)
+        self.audio_analysis = AudioAnalysisController(self)
```
marcelveldt: let's follow the pattern here of putting this in a private attribute and expose as property

music_assistant/controllers/music.py
```diff
@@ -1180,6 +1182,114 @@ async def get_smart_fades_analysis(
             )
         return None
 
+    async def set_audio_analysis(
```
marcelveldt: we should seriously consider letting the audio analysis controller have it's own database instead of (ab)using the music library db

music_assistant/controllers/music.py
```diff
@@ -1180,6 +1182,114 @@ async def get_smart_fades_analysis(
...
+        return merged if found else None
+
+    async def get_audio_analysis_version(
+        self,
+        item_id: str,
+        provider_instance_id_or_domain: str,
+        aa_provider_domain: str,
+        media_type: MediaType = MediaType.TRACK,
+    ) -> int | None:
+        """Get the stored analysis version for a specific AA provider and track.
```
marcelveldt: track only ?

music_assistant/controllers/music.py
```diff
@@ -3011,6 +3121,19 @@ async def __create_database_tables(self) -> None:
...
         )
 
+        await self.database.execute(
+            f"""CREATE TABLE IF NOT EXISTS {DB_TABLE_AUDIO_ANALYSIS}(
+                    [id] INTEGER PRIMARY KEY AUTOINCREMENT,
+                    [media_type] TEXT NOT NULL,
+                    [item_id] TEXT NOT NULL,
+                    [provider] TEXT NOT NULL,
+                    [aa_provider_domain] TEXT NOT NULL,
+                    [analysis_data] json NOT NULL,
```
marcelveldt, line 3047: I'm not sure I'm a fan of this concept of storing one big blob of json of the same model.
Maybe we should turn this into a key, value storage instead.

Then the analysis type could defined as simple a strEnum

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+        pcm_chunk: bytes,
+    ) -> None:
+        """Process a PCM audio chunk.
+
+        Called for each 1-second chunk of raw PCM audio during streaming.
+        A real provider would feed this data into its analysis algorithm
+        (e.g. beat tracker, key detector, loudness analyzer).
+
+        The audio format is available via self._sessions[session_id].audio_format.
```
marcelveldt, line 153: why ? its not set that way in the _start_analysis call above so this is false info

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+        self.logger.debug(
+            "Processing chunk for session %s (%d bytes)",
+            session_id,
+            len(pcm_chunk),
+        )
+
+    async def _finalize(self, session_id: str) -> None:
+        """Finalize analysis and store results.
+
+        Called when the track has finished streaming and all chunks have been
```
marcelveldt: strictly speaking this is called when it has finished buffering and not streaming.
the entire audio analysis runs ahead of time in the audiobuffer

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,217 @@
...
+    """Controller that distributes PCM chunks to all registered AudioAnalysisProviders."""
+
+    def __init__(self, streams: StreamsController) -> None:
+        """Initialize the AudioAnalysisController.
+
+        :param streams: Parent StreamsController instance.
+        """
+        self.streams = streams
+        self.mass = streams.mass
+        self.logger = logging.getLogger(MASS_LOGGER_NAME).getChild("audio_analysis")
```
marcelveldt: you can also replace this with mass.logger.getChild(

music_assistant/controllers/music.py
MarvinSchenkel: No, this is used to short circuit analysis sessions if a certain AA provider already has done analysis for track ABC

music_assistant/controllers/music.py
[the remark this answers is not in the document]
MarvinSchenkel: Let's take a 10 minute song as an example at 120bpm. That's 1200 beats, 300 downbeats, +- 600 rms / spectroid values and 800 waveform values.. total: +- 3500 floats (worse case). Not rocket science. Sure you want the overhead of a worker thread here?

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
MarvinSchenkel: Because _start_analysis is the hook. The base implementation start_analysis automatically sets this for each provider implementation and automatically does the version check etc.

music_assistant/controllers/music.py
marcelveldt: what I meant was that audio analysis is not limited to tracks. For example loudness measurements are for all media types

music_assistant/providers/_demo_audio_analysis_provider/__init__.py
marcelveldt: yeah but that was not clear from the demo code. It gave the impression that this private attribute would already exist

music_assistant/models/audio_analysis_provider.py
```diff
@@ -0,0 +1,152 @@
...
+            self.domain,
+        )
+        if stored_version is not None and stored_version >= self.analysis_version:
+            return False
+        self._sessions[session_id] = AnalysisSessionData(
+            streamdetails=streamdetails,
+            audio_format=audio_format,
+        )
+        if not await self._start_analysis(session_id, streamdetails, audio_format):
+            self._sessions.pop(session_id, None)
```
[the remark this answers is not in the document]
marcelveldt: this comment makes sense

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -0,0 +1,342 @@
...
+        :param item_id: Provider-native item ID from streamdetails.item_id.
+        :param provider_instance_id_or_domain: Music provider instance ID or domain.
+        :param aa_provider_domain: Domain of the AA provider.
+        :param media_type: The media type of the item.
+        """
+        if not (
+            provider := self.mass.get_provider(
+                provider_instance_id_or_domain, provider_type=MusicProvider
+            )
+        ):
```
marcelveldt, line 229: this is a bit ugly to read.
maybe do if (blah := blah) is None:

marcelveldt: can be done in a follow-up if you like.

marcelveldt (approved):
Nice!

PR #3510, opened 2026-03-30
Plex Connect: refactor and fixes plugin

anatosun:
This PR refactors the Plex Connect plugin and fixes several issues encountered during
   testing.                                                                            
                                                            
  Refactor                                                                             
   
  The monolithic player_remote.py has been split into focused modules for better       
  readability and easier contributions:                     
                                                                                       
  - server.py — HTTP server setup and shared state                                     
  - timeline.py — Plex timeline XML generation and broadcasting
  - playback.py — Playback control handlers (play, pause, seek, skip, etc.)            
  - queue_commands.py — Play queue command handlers (playMedia, createPlayQueue,       
  refreshPlayQueue)                                                                    
  - queue_sync.py — Queue synchronisation between MA and Plex                          
                                                                                       
  The plugin stage has also been updated from alpha to beta.                           
                                                                                       
  Bug fixes                                                                            
                                                            
  - Infinite loop on refreshPlayQueue: the refreshPlayQueue handler was missing the    
  _updating_from_plex guard, causing a loop when features like Guest DJ were used.
  - ~Play queue truncated at 50 items: the default Plex API window of 50 was used when  
  fetching play queues. The fetch now paginates through the full queue using           
  playQueueTotalCount.~ This incurs significant delays and throttling of the playback on large queues! The queue is capped to 100 tracks.
  - Queue ownership: own=True is now passed on the initial PlayQueue.get() call so MA  
  takes full control of the queue.                                                     
  - Shuffle not propagated to Plex: toggling shuffle from Plexamp now triggers a
  _create_plex_playqueue_from_ma call to sync the new order back.                      
  - Double-shuffle on shuffled Plex queues: when a shuffled PlayQueue is loaded, the
  original source is now fetched via playQueueSourceURI and loaded into MA in its      
  natural order. MA then applies its own shuffle and propagates the result back to
  Plex, avoiding double-shuffling an already-shuffled list.                            
  - Stale MA shuffle corrupting unshuffled queues: the shuffle state is now always
  explicitly synced from Plex on every playMedia command, preventing a previously      
  enabled MA shuffle from silently reordering a non-shuffled queue.


music_assistant/providers/plex_connect/queue_sync.py
```diff
@@ -0,0 +1,394 @@
...
+                if isinstance(result, Exception):
+                    LOGGER.debug(f"Error fetching track: {result}")
+                    continue
+                _plex_idx, track, play_queue_item_id = result  # type: ignore[misc]
+                if track:
+                    ma_idx = len(tracks_to_add) + 1  # +1 because first track is already queued
+                    tracks_to_add.append(track)
+                    if play_queue_item_id:
+                        self.play_queue_item_ids[ma_idx] = play_queue_item_id
+
```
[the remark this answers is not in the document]
anatosun: this is capped to 500, which, after some testing, seems not to impact performance in a noticeable fashion

music_assistant/providers/plex_connect/server.py
```diff
@@ -0,0 +1,371 @@
...
+        if self._ma_player_id:
+            player = self.provider.mass.players.get_player(self._ma_player_id)
+            if player:
+                player_name = player.display_name
+
+        state = "stopped"
+        if self._ma_player_id:
+            player = self.provider.mass.players.get_player(self._ma_player_id)
+            if player and player.state:
+                state_value = (
```
MarvinSchenkel: I think you want to use `player.playback_state` here?

OzGav:
@anatosun i think we are hoping to get this into the 2.9 release which is only a few weeks away. FYI.

anatosun:
It should be good to go... Sorry for the delay but I was gaslighted by Copilot's suggestions

OzGav:
@anatosun if you resolve the latest copilot comments we will keep moving this forward!

OzGav:
@anatosun there is just one open conversation from Marvin here https://github.com/music-assistant/server/pull/3510#discussion_r3264544899

music_assistant/providers/plex_connect/timeline.py
```diff
@@ -0,0 +1,312 @@
...
+        :param client_id: The client ID to send the timeline to.
+        """
+        subscription = self.subscriptions.get(client_id)
+        if not subscription:
+            return
+
+        timeline_xml = await self._build_timeline_xml()
+
+        try:
+            await self.provider.mass.http_session.post(
```
MarvinSchenkel: Can we use `async with` here to avoid connection leaking?

anatosun: That's a good point. Let me fix this.

MarvinSchenkel (approved):
Thanks @anatosun 🎉

PR #3512, opened 2026-03-30
Fix plugin volume feedback loop in group volume changes

theturtle32:
## Summary

Fixes a feedback loop where adjusting individual child player volumes within a group that has an active plugin source (e.g. Spotify Connect) caused oscillating volume changes. Each child volume change fired the plugin's `on_volume` callback with that child's individual volume, which Spotify echoed back as a group-level command, cascading unintended changes to all other group members.

### Root Cause

In `_handle_cmd_volume_set`, every volume change -- including those dispatched by `set_group_volume` to child players -- unconditionally called `plugin_source.on_volume(volume_level)` if a plugin source was active. For group volume changes, this meant N plugin callbacks (one per child, each with a different volume) instead of a single callback with the group volume. Spotify Connect's inbound `volume_changed` handler then fed each echo back as a new `cmd_volume_set`, creating the loop.

### Fix

**Plugin callback isolation via `from_group_volume` flag:**

- `_handle_cmd_volume_set` gains a `from_group_volume: bool = False` parameter
- When `True` (called from `set_group_volume`), plugin source callbacks are skipped entirely -- the caller handles the callback at the group level
- `set_group_volume` now fires exactly one plugin callback with the group volume value after dispatching all child volume commands
- The `from_group_volume` flag propagates through protocol player volume redirects

**Projected group average for child volume changes:**

- When a child player's volume changes individually (not via `set_group_volume`), the code computes the projected new group average by substituting the child's new volume into the average calculation
- If the integer average shifts, the plugin callback fires on the parent group with the new group volume
- If the average doesn't change (due to integer rounding), no callback fires
- This ensures Spotify and other plugins stay in sync when individual child volume changes cause the group volume slider to move in the MA UI

**Standalone player behavior preserved:**

- Standalone (non-group) player volume changes continue to fire plugin callbacks normally with the player's volume

**Spotify Connect echo suppression:**

- The inbound `volume_changed` handler now checks if the incoming volume matches `_last_volume_sent_to_spotify` before issuing `cmd_volume_set`, suppressing echoes of our own outbound volume commands

**Debug logging:**

- `[GroupVolume]`-prefixed debug logging added to `cmd_volume_set`, `cmd_group_volume`, `set_group_volume`, `_handle_cmd_volume_set`, `_handle_volume_plugin_callback`, and Spotify Connect volume handlers for production diagnostics

### Files Changed

- `music_assistant/controllers/players/controller.py` -- `from_group_volume` flag, plugin isolation logic, projected group average, group-level callback, `_handle_volume_plugin_callback` helper, debug logging
- `music_assistant/providers/spotify_connect/__init__.py` -- inbound echo suppression, debug logging
- `tests/core/test_plugin_volume_isolation.py` -- new test suite (9 tests)

## Test Plan

- [x] `set_group_volume` passes `from_group_volume=True` to all child dispatches
- [x] `set_group_volume` fires exactly one plugin callback with the group volume
- [x] Child volume change that does NOT shift group average: no plugin callback
- [x] Child volume change that DOES shift group average: plugin fires with new average
- [x] `from_group_volume=True` skips plugin callback entirely
- [x] Protocol player redirect preserves `from_group_volume` flag
- [x] Standalone player fires plugin callback normally
- [x] Pre-commit hooks pass
- [x] Full test suite passes (3 pre-existing unrelated failures in test_tags.py)

## Relationship to #3399

This PR extracts the critical feedback loop fix from #3399 (which also includes ratio-based group volume with max-tracking). The intent is to allow this targeted bug fix to be reviewed and merged independently, since the broader group volume algorithm in #3399 may require additional discussion.

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import asyncio
 import time
+from collections.abc import Awaitable
 from contextlib import suppress
 from contextvars import ContextVar
 from typing import TYPE_CHECKING, Any, cast
@@ -147,6 +148,11 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
         self._pending_protocol_evaluations: dict[str, asyncio.TimerHandle] = {}
         # Serialize delayed evaluations to prevent race conditions
         self._delayed_evaluation_lock = asyncio.Lock()
+        # Coalescing state for cmd_group_volume: during rapid slider drags,
+        # intermediate values are dropped so only the first and latest are
+        # executed.  See cmd_group_volume for the coalescing loop.
+        self._group_vol_in_flight: dict[str, bool] = {}
+        self._group_vol_target: dict[str, int] = {}
 
     async def get_config_entries(
         self,
@@ -622,6 +628,11 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
         :param player_id: player_id of the player to handle the command.
         :param volume_level: volume level (0..100) to set on the player.
         """
+        self.logger.debug(
+            "[GroupVolume] cmd_volume_set ENTRY: player_id=%s, volume_level=%d",
+            player_id,
+            volume_level,
+        )
         await self._handle_cmd_volume_set(player_id, volume_level)
 
     @api_command("players/cmd/volume_up")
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/players/controller.py
```diff
@@ -622,6 +623,11 @@ async def cmd_volume_set(self, player_id: str, volume_level: int) -> None:
         :param player_id: player_id of the player to handle the command.
         :param volume_level: volume level (0..100) to set on the player.
         """
+        self.logger.debug(
+            "[GroupVolume] cmd_volume_set ENTRY: player_id=%s, volume_level=%d",
+            player_id,
+            volume_level,
+        )
```
marcelveldt, line 635: do we still need this ?
or can we generalize it a bit ?
The groupvolume part may be confusing for some

music_assistant/controllers/players/controller.py
```diff
@@ -685,12 +691,20 @@ async def cmd_group_volume(
...
+        self.logger.debug(
+            "[GroupVolume] cmd_group_volume ENTRY: player=%s (%s), vol=%d, "
+            "type=%s, group_members=%s, synced_to=%s",
+            player.state.name,
+            player_id,
+            volume_level,
+            player.state.type,
+            player.state.group_members,
+            player.state.synced_to,
+        )
```
marcelveldt, line 714: same note - can we drop this ?
or otherwise change it to the verbose log level

music_assistant/controllers/players/controller.py
```diff
@@ -1714,26 +1728,60 @@ def get_active_queue(self, player: Player) -> PlayerQueue | None:
...
             group_player, only_powered=True, exclude_self=False
         ):
             if child_player.state.volume_control == PLAYER_CONTROL_NONE:
                 continue
             cur_child_volume = child_player.state.volume_level or 0
-            new_child_volume = int(cur_child_volume + volume_dif)
-            new_child_volume = max(0, new_child_volume)
-            new_child_volume = min(100, new_child_volume)
-            # Use private method to skip permission check - already validated on group
-            # ATTR_MUTE_LOCK on muted players prevents auto-unmute during group volume changes
```
marcelveldt, line 1745: important comment deleted ?

music_assistant/controllers/players/controller.py
```diff
@@ -2857,16 +2905,30 @@ async def _handle_cmd_power(
...
         assert player is not None  # for type checker
+        self.logger.debug(
+            "[GroupVolume] _handle_cmd_volume_set: player=%s (%s), vol=%d, "
+            "from_group_volume=%s, current_vol=%s",
+            player.state.name,
+            player_id,
+            volume_level,
+            from_group_volume,
+            player.state.volume_level,
+        )
```
marcelveldt, line 3250: drop the log or update so its correct - it mentions groupvolume while this is an individual player call

marcelveldt:
Yes, this is much better. Tackle one issue at a time.
Once this is in, we can have another look/discussion for the relative group volume changes

marcelveldt:
OK, all the AI slop is back again in this PR with an overengineered solution that nobody on the planet can follow anymore. Exactlty the thing I commented on in the previous PR, then you added a new PR with a more narrowed down solution for one issue at a time, yet the whole circus of unrelated changes and overly complex solutions is back.

I'm not disagreeing there is an issue but this is not the solution.
If you can give me a (human written!) clear explaination of the issue, I will provide the fix

theturtle32:
Apologies, again, @marcelveldt!

I did not intend this yet to be formally submitted and reviewed and taking up your time.

I'm still exploring the codebase and iterating on approaches to this and learning how everything fits together. My intent was for this to be a _**draft**_ PR, a work-in-progress, and then to make sure it was tidy, tight, clean, and polished before formally submitting it to you.

Moving forward, I will refrain from using draft PRs in this way, and only open a PR, even a draft PR, when I feel it is ready for you to look at.

PR #3513, opened 2026-03-30
Add Socks proxy option for Pandora

TermeHansen:
As far as I can see, so far no providers use a socks-proxy. With this PR I added my suggestion for Pandora based on aiohttp-socks.

I also had some 403 error issues in general with the Pandora provider that was fixed by changing the user-agent in the headers. I guess this needs to be tested also by some of the users were it already worked until now.


marcelveldt:
I think this proxy option is better to implement at a global level ?
So you can configure a proxy for all outbound requests from MA to the internet.
Does that make sense ? Or would this be more of a provider specific thing ?

TermeHansen:
I think it make sence to have it as an option for the global http_client helper, but I think it makes most sence per provider.

I can have a look and test and suggest on a more general approach?

TermeHansen:
@marcelveldt do I need to do something more here for now in this PR (I don't think I can add a label for instance)?

TermeHansen:
@OzGav do you have an opinion on this PR, and could you test if the user-agent change still works for you?

OzGav:
Unfortunately since I wrote this, Pandora has increased their geo blocking capability and I cant use the Pandora provider at the moment. I need to sort out a VPN…. In terms of the user agent I recall I went through a few iterations to get something acceptable. If your change is working reliably the I can only assume it is ok.

TermeHansen:
@OzGav I just sent you a mail directly with suggestion to how you can test my PR here...

music_assistant/providers/pandora/__init__.py
```diff
@@ -70,4 +70,12 @@ async def get_config_entries(
             description="Your Pandora password",
             required=True,
         ),
+        ConfigEntry(
+            key=CONF_SOCKS_URL,
+            type=ConfigEntryType.STRING,
+            label="Socks",
+            description="Socks proxy server",
```
marcelveldt: I think the label should be "Socks proxy server" and the description should be a bit more descriptive. Also I would mark it as advanced because most people will have no clue what a socks proxy server is

music_assistant/helpers/aiohttp_client.py
```diff
@@ -171,6 +174,15 @@ def _get_connector(
     else:
         ssl_context = ssl_util.client_context_no_verify(ssl_cipher)
 
+    if socks_url:
```
marcelveldt, line 177: This approach is fine for now but if we at some point see a need for a global config of the socks server instead of per-provider, we could consider making a global config option to configure a proxy server for all outbound traffic through the http client.

TermeHansen: I agree, which is why I put it like that so it should be easy to use also for the global mass.http_client. I'm fairly new to the whole MA infrastructure so I would like not do that part here with this PR.

music_assistant/providers/pandora/__init__.py
TermeHansen: image

music_assistant/helpers/aiohttp_client.py
```diff
@@ -158,6 +160,7 @@ class MusicAssistantTCPConnector(aiohttp.TCPConnector):
 def _get_connector(
     mass: MusicAssistant,
     verify_ssl: bool = True,
+    socks_url: str = "",
```
marcelveldt:
```suggestion
    socks_url: str | None = None,
```

marcelveldt (approved):
Thanks, @TermeHansen !

OzGav:
There is a mypy and test failure that needs resolving

MarvinSchenkel:
We also need to add aiohttp_socks to the dependencies

TermeHansen:
@marcelveldt it just needs to re-run the actions or is there something more to do?

MarvinSchenkel:
Nope looks like aiohttp released a new patch because the current version has some cves. Not related to this PR. This can be merged as is!

PR #3514, opened 2026-03-30
Apple Music: Add content rating check for explicit tracks

LosCV29:
The Apple Music provider parses contentRating on albums (line 841) but not on tracks, leaving track.metadata.explicit as None for all Apple Music tracks. The Apple Music API does return contentRating on song objects with values "explicit", "clean", or absent.

This prevents Music Assistant from distinguishing between explicit and clean versions of the same track, affecting compare_track() matching and any downstream integrations relying on the explicit flag.

Every other provider (Spotify, Tidal, Deezer, Qobuz, YouTube Music) already sets metadata.explicit on tracks.

Ref: https://developer.apple.com/documentation/applemusicapi/songs/attributes-data.dictionary


marcelveldt (approved):
Thanks, @LosCV29 !

PR #3515, opened 2026-03-31
Improve Qobuz API rate limiting, backoff, and sync efficiency

teancom:
Throttler rate: increase from 1 req/2sec (0.5/sec) to 2 req/sec. The old rate was overly conservative, causing every API call during sync to queue behind a ~1.6 second throttle delay. The new rate is still conservative but 4x faster, and the retry logic will properly back off if Qobuz actually sends 429 responses.

Page size: increase from 50 to 500 for paginated endpoints (_get_all_items). This is the biggest win — a library with 500 favorite albums previously required 10 sequential API calls (20+ seconds with throttling). Now it's a single call. The Qobuz API supports large limits (their own getUserPurchasesIds uses limit=5000). Added a guard using the response's `total` field so that if the API silently caps our limit, we continue paginating instead of truncating.

429 Retry-After handling: add WARNING-level logging when Qobuz actually rate-limits us (previously silent), making it easy to distinguish local throttle delays (DEBUG) from server-imposed backoff (WARNING). Parse the Retry-After header safely with try/except to handle non-integer values (HTTP-date format per RFC 9110, floats).

Backoff logic rewrite in throttle_retry.py:
- Server-provided backoff (Retry-After > 0): used exactly as-is on each retry. No doubling — if the server says wait 10s, we wait 10s.
- No server guidance (bare 429 or backoff_time=0): exponential backoff with ±25% jitter to decorrelate concurrent retries. Progression: initial_backoff, 2x, 4x, 8x... capped at MAX_BACKOFF (120s).
- These two paths are independent: the exponential counter keeps ticking during server-directed retries but doesn't affect them.

Also fixed the misleading comment on QobuzProvider.throttler (it's a class attribute shared across instances, not an instance attribute).

Tests: added tests/core/test_throttle_retry.py with 9 test cases covering basic behavior, server-provided backoff, exponential backoff with jitter, backoff cap, and mixed scenarios.

### DIFF
--- a/music_assistant/helpers/throttle_retry.py
+++ b/music_assistant/helpers/throttle_retry.py
@@ -1,13 +1,16 @@
 """Context manager using asyncio_throttle that catches and re-raises RetriesExhausted."""
 
 import asyncio
+import datetime
 import functools
 import logging
+import random
 import time
 from collections import deque
 from collections.abc import AsyncGenerator, Awaitable, Callable, Coroutine
 from contextlib import asynccontextmanager
 from contextvars import ContextVar
+from email.utils import parsedate_to_datetime
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Concatenate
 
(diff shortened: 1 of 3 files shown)


music_assistant/helpers/throttle_retry.py
```diff
@@ -123,14 +153,20 @@ async def wrapper(self: ProviderT, *args: P.args, **kwargs: P.kwargs) -> R:
...
                     if attempt < throttler.retry_attempts - 1:
-                        self.logger.info(f"Retrying in {backoff_time} seconds...")
-                        await asyncio.sleep(backoff_time)
-                        backoff_time *= 2
+                        if e.backoff_time > 0:
+                            # Server told us exactly how long to wait — respect it
+                            sleep_time = float(e.backoff_time)
+                        else:
+                            # No server guidance — exponential backoff with jitter
+                            sleep_time = min(exp_backoff, MAX_BACKOFF)
```
MarvinSchenkel: Shouldn't these two lines be `sleep_time = min(exp_backoff * random.uniform(0.75, 1.25), MAX_BACKOFF)` in order to not exceed the max backoff?

music_assistant/providers/qobuz/__init__.py
```diff
@@ -833,7 +837,13 @@ async def _get_all_items(
                 break
             for item in result[key]["items"]:
                 all_items.append(item)
-            if len(result[key]["items"]) < limit:
+            total = result[key].get("total", 0)
+            items_received = len(result[key]["items"])
+            if items_received < limit:
+                # If the API returned fewer items than requested but reports more exist,
+                # the server silently capped our limit. Continue paginating.
+                if total > len(all_items):
```
MarvinSchenkel: Let's change this to `if items_received > 0 and total > len(all_items)` to ensure we cannot enter infinite loops

MarvinSchenkel (dismissed):
Some minor changes, other than that looks good to go 👏 .

music_assistant/helpers/throttle_retry.py
teancom: I actually had it as a soft cap, to prevent a cluster of requests right at the 120s mark. But also I realized that's my background as an SRE coming through, dealing with herds of thundercats vs. something we realistically have to worry about here. And it would just be confusing to everyone else going forward. Changed. 😄

marcelveldt (approved):
Nice work @teancom !

PR #3516, opened 2026-03-31
Feature: sonic analysis provider

chrisuthe:
## Sonic Analysis Provider

This PR adds a **sonic analysis provider** that extracts semantic audio features from PCM audio during playback using librosa and stores them as standard `AudioAnalysisData` fields.

### What It Does

- Processes audio in **10-second blocks** with overlap to avoid STFT artifacts
- Derives **human-readable semantic descriptors** from raw spectral features:

| Field | Derivation |
|---|---|
| `bpm` | Tempo estimation from onset envelope |
| `key` / `mode` | Krumhansl-Kessler profile correlation against chroma |
| `energy` | Normalized mean RMS |
| `danceability` | Onset regularity + tempo suitability |
| `loudness_integrated` / `loudness_range` | RMS-derived dB approximations |
| `brightness` | Spectral centroid normalized against Nyquist |
| `harmonic_complexity` | Shannon entropy of mean chroma vector |
| `roughness` | Spectral contrast range + flatness |
| `rhythmic_regularity` | Inter-onset interval coefficient of variation |
| `rms_energy_per_second` / `spectral_centroid_per_second` | Per-second time series |

- Stores results via `set_audio_analysis()` as a plain `AudioAnalysisData` — no opaque blobs, no custom subclasses
- Proposes 4 new upstream fields: `brightness`, `harmonic_complexity`, `roughness`, `rhythmic_regularity`

### Architecture

The provider is a pure **feature extraction + distillation** layer. It does not store similarity vectors or compute distances — that responsibility belongs to the similarity plugin (separate stacked PR).

Both the provider and the similarity plugin depend on the shared `AudioAnalysisData` model contract. Any audio analysis provider that populates the same fields can feed the similarity plugin.

### Code Organization

- `helpers.py` — Pure feature extraction (`extract_block_features`, `merge_block_features`) and semantic derivation (`collapse_to_analysis` with private `_derive_*` helpers)
- `__init__.py` — MA integration: PCM streaming, block accumulation, session management, `_finalize()` stores the result

### Testing

- `test_helpers.py` — Tests for block extraction, merging, and `collapse_to_analysis` (scalar ranges, determinism, noise vs sine differentiation)
- `test_provider_units.py` — Tests for PCM byte conversion (16/24/32-bit, mono/stereo)

### Dependencies

- Builds on top of the upstream `audio_analysis_controller_provider` branch (PR #3509)
- The **sonic similarity plugin** (fork PR #1) stacks on this branch

### DIFF
--- a/music_assistant/providers/sonic_analysis/__init__.py
+++ b/music_assistant/providers/sonic_analysis/__init__.py
@@ -1,0 +1,269 @@
+"""Sonic Analysis provider for Music Assistant.
+
+Extracts audio features from PCM audio streams during playback and
+stores them as semantic AudioAnalysisData fields.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+from music_assistant.models.audio_analysis_provider import (
+    AnalysisSessionData,
+    AudioAnalysisProvider,
+)
+
+from .helpers import (
+    BlockFeatures,
+    collapse_to_analysis,
+    extract_block_features,
+    merge_block_features,
+)
+
+if TYPE_CHECKING:
+    from music_assistant_models.config_entries import ConfigEntry, ConfigValueType, ProviderConfig
+    from music_assistant_models.media_items import AudioFormat
+    from music_assistant_models.provider import ProviderManifest
+    from music_assistant_models.streamdetails import StreamDetails
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+
+
+BLOCK_SECONDS: int = 10
+OVERLAP_SAMPLES: int = 2048
+
+
+@dataclass
+class SonicSessionData(AnalysisSessionData):
+    """Per-session state: PCM block buffer and accumulated per-block features."""
+
+    pcm_buffer: bytearray = field(default_factory=bytearray)
+    block_samples: int = 0
+    accumulated: BlockFeatures = field(default_factory=BlockFeatures)
+    total_samples: int = 0
+    overlap: np.ndarray | None = None
+    start_time: float = 0.0
+    peak_absolute: float = 0.0
+    waveform_peaks: list[float] = field(default_factory=list)
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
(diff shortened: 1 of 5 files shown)


music_assistant/providers/sonic_analysis/helpers.py
```diff
@@ -0,0 +1,253 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+import librosa
+import numpy as np
+
+SIGNATURE_VERSION: int = 2
+
+# Feature group index ranges (start, end) within the 95-dim signature
```
MacgyverH: Forgive me, as this might be my misunderstanding of Python, but are these start and end indexes overlapping? Is that intentional?

chrisuthe: Good catch- these use Python's "half-open" slice convention: (0, 39) means indices 0-38, (39, 69) means 39-68, etc. No overlap since they exclude the last.

MacgyverH: Thank you for the explanation! I appreciate it.

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,1207 @@
...
+        pcm_chunk: bytes,
+    ) -> None:
+        """Append a PCM audio chunk to the session buffer.
+
+        :param session_id: The analysis session ID.
+        :param pcm_chunk: Raw PCM audio data.
+        """
+        session = self._sessions[session_id]
+        assert isinstance(session, SonicSessionData)
+        session.pcm_buffer.extend(pcm_chunk)
```
MarvinSchenkel, line 178: We cannot keep extending pcm audio into a buffer because that will kill memory. We will need to incrementally calculate features rather than keep collecting pcm audio and doing a 1 pass feature extraction. Have a look at my (wip ;-) ) smart fades provider [here](https://github.com/music-assistant/server/blob/eca8895f80715d0914954a247065d50eb766b9a1/music_assistant/providers/smart_fades/provider.py#L162)

chrisuthe: Makes total sense, buffering up an entire song times 5 people listening off a Raspberry Pi that's already running HAOS == disaster memory-wise. I can implement your 10 second block pattern without a problem. Thank you for the catch!

PR #3369, opened 2026-03-12
Rewrite tidal stream behaviour to avoid premature cutoff

jozefKruszynski:
Closes https://github.com/music-assistant/support/issues/5077

decode dash xml and serve as internal http endpoint. Cache manifest details for short term replays of a given track

PR #3518, opened 2026-03-31
Rewrite tidal streaming

jozefKruszynski:
DASH manifests are now parsed by the provider rather than being handed off to ffmpeg.

### DIFF
--- a/music_assistant/providers/tidal/provider.py
+++ b/music_assistant/providers/tidal/provider.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from collections.abc import AsyncGenerator
 from datetime import datetime
 from typing import TYPE_CHECKING, Any
 
@@ -187,6 +188,13 @@ class TidalProvider(MusicProvider):
     ) -> StreamDetails:
         """Return the content details for the given track when it will be streamed."""
         return await self.streaming.get_stream_details(item_id)
+
+    async def get_audio_stream(
+        self, streamdetails: StreamDetails, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """Return the audio stream for the provider item."""
+        async for chunk in self.streaming.get_audio_stream(streamdetails, seek_position):
+            yield chunk
 
     def get_item_mapping(self, media_type: MediaType, key: str, name: str) -> ItemMapping:
         """Create a generic item mapping."""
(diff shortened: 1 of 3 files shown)


marcelveldt:
Can you explain why this change is needed ? It would be more efficient to let ffmpeg deal with the heavy lifting.
If there's an issue with ffmpeg's DASH demuxer, we can explore if we should update to ffmpeg 8.

If this is about handling over the dash playlist/manifest as base64 string, why not host an extra endpoint on the streamserver that serves the dash manifest ?

jozefKruszynski:
> Can you explain why this change is needed ? It would be more efficient to let ffmpeg deal with the heavy lifting. If there's an issue with ffmpeg's DASH demuxer, we can explore if we should update to ffmpeg 8.

ffmpeg still does all the heavy lifting, it decodes the FLAC-in-fMP4, applies volume normalization, resampling, crossfade mixing, and encodes the output format. I'm only bypassing the demuxer, which appears to be what is the root cause if the initial bug.

The failure mode is that ffmpeg's DASH demuxer silently stops fetching segments partway through the track (return code 0, no error), producing a truncated stream. It's not a crash or timeout — it just decides it's done early. This happened repeatedly during testing with both data: URIs and HTTP-served manifests.

> If this is about handling over the dash playlist/manifest as base64 string, why not host an extra endpoint on the streamserver that serves the dash manifest ?

That's exactly what PR #3369 did, but I introduced a new bug that we saw in discord, dynamic routes get orphaned or expire before playback completes, causing 404s.

I fixed this bug, but then found that the exact same behaviour of streams prematurely stopping was still present.

marcelveldt:
maybe check the ffmpeg bugtracker if this was an issue they fixed ?

jozefKruszynski:
Will do.

jozefKruszynski:
There was a fix that was added to address a very similar issue to what we're seeing. Unfortunately, that fix was already added to the 7.1 branch, so our 7.1.2 version should already have it.

I'll look further into it.

Edit: looks like there is a relevant bugfix that might help.
I'll do some testing when I next get the opportunity.

MarvinSchenkel:
Can you mark this as ready for review when you want another review please? :)

jozefKruszynski:
Will do, I should hopefully have some time to test tomorrow again.

OzGav:
@jozefKruszynski How is this looking?

jozefKruszynski:
Good point, I'll close it as the pr that was recently merged should fix the original issue.

PR #3520, opened 2026-03-31
Fix MusicBrainz track lookup to aggregate release groups from all rec…

OzGav:
When looking up release groups for a track, the old code would return as soon as it found the first recording with any release groups. This meant that if Recording 1 only had a single but Recording 4 had the album, we'd never find the album.

Now we aggregate release groups from all matching recordings before returning.

Also filters singles to only include those where the release group title matches the track name. This prevents returning artwork for unrelated singles that happen to include the track (eg. a B-side)


marcelveldt (approved):
Nice work @OzGav !

PR #3523, opened 2026-03-31
Add Coverart Archive metadata provider

OzGav:


music_assistant/providers/coverartarchive/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+  "domain": "coverartarchive",
+  "stage": "stable",
+  "name": "Cover Art Archive",
+  "description": "Provides album artwork from the Cover Art Archive, the official artwork repository for MusicBrainz.",
+  "codeowners": ["@music-assistant"],
+  "credits": ["[Cover Art Archive](https://coverartarchive.org)"],
+  "requirements": [],
+  "documentation": "",
+  "multi_instance": false,
+  "builtin": false
```
OzGav:
```suggestion
  "builtin": true
```

music_assistant/providers/coverartarchive/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Fetches album artwork from the Cover Art Archive using MusicBrainz release group IDs.
+    """
+
+    async def get_album_metadata(self, album: Album) -> MediaItemMetadata | None:
+        """Retrieve metadata for an album.
+
+        :param album: Album to retrieve metadata for.
+        """
+        if not self.config.get_value(CONF_ENABLE_ALBUM_IMAGES):
```
marcelveldt: why is this configurable ?

music_assistant/providers/coverartarchive/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+    :param action: [optional] action key called from config entries UI.
+    :param values: the (intermediate) raw values for config entries sent with the action.
+    """
+    return (
+        ConfigEntry(
+            key=CONF_ENABLE_ALBUM_IMAGES,
+            type=ConfigEntryType.BOOLEAN,
+            label="Enable retrieval of album images",
+            default_value=True,
+        ),
```
marcelveldt: I'm not understanding this at all

music_assistant/providers/coverartarchive/__init__.py
OzGav: Just following the lead of fanart and theaudiodb?

marcelveldt: yeah but there it actually makes sense - in this case you have a provider that has a sole purpose of providing album images that has a config option to not provide album images. ?!
in that case you can maybe just disable the provider ?

marcelveldt: follow up question: does coverart archive also provide artist images ?

OzGav: Yes you can disable so I will remove the config option. 

The site only supplies cover art images https://musicbrainz.org/doc/Cover_Art_Archive/API

music_assistant/providers/coverartarchive/__init__.py
OzGav: Config option removed

marcelveldt (approved):
Thanks @OzGav !

PR #3525, opened 2026-04-01
Add DTZ006 mypy rule

OzGav:


music_assistant/controllers/media/base.py
```diff
@@ -1105,7 +1105,7 @@ def _parse_db_row(db_row: Mapping[str, Any]) -> dict[str, Any]:
         db_row_dict["favorite"] = bool(db_row_dict["favorite"])
         db_row_dict["item_id"] = str(db_row_dict["item_id"])
         db_row_dict["date_added"] = datetime.fromtimestamp(
```
marcelveldt, line 1107: we should consider adding a "from_utc_timestamp" helper in helpers/datetime

OzGav: I am not 100% sure I understand what you are after but I think maybe converting from the UTC in the database to local? I have added two helper that do that...

music_assistant/helpers/datetime.py
```diff
@@ -37,11 +37,21 @@ def from_utc_timestamp(timestamp: float) -> datetime.datetime:
...
 def iso_from_utc_timestamp(timestamp: float) -> str:
     """Return ISO 8601 datetime string from UTC timestamp."""
     return from_utc_timestamp(timestamp).isoformat()
 
 
+def iso_local_from_utc_timestamp(timestamp: float) -> str:
+    """Return ISO 8601 datetime string in local timezone from UTC timestamp."""
+    return from_utc_timestamp_to_local(timestamp).isoformat()
+
+
```
OzGav:
```suggestion
```

music_assistant/helpers/datetime.py
```diff
@@ -37,6 +37,11 @@ def from_utc_timestamp(timestamp: float) -> datetime.datetime:
     return datetime.datetime.fromtimestamp(timestamp, datetime.UTC)
 
 
+def from_utc_timestamp_to_local(timestamp: float) -> datetime.datetime:
+    """Convert UTC timestamp to datetime in local timezone."""
+    return from_utc_timestamp(timestamp).astimezone(LOCAL_TIMEZONE)
+
+
```
OzGav:
```suggestion
```

music_assistant/controllers/media/base.py
OzGav: Actually I think you meant a helper to replace the line "datetime.fromtimestamp(timestamp, tz=UTC)"

I looked at it but since it is only one line didn't think there was much to be saved so I will resolve this but let me know if you want it done

MarvinSchenkel:
Can you elaborate why you remove the utc timestamp function from the helpers file?

music_assistant/providers/ard_audiothek/__init__.py
MarvinSchenkel: `datetime.now())` would also need a `tz` here to avoid TypeErrors

OzGav: Actually that is covered by DTZ005 and was fixed in that other PR.

music_assistant/providers/audiobookshelf/parsers.py
```diff
@@ -182,7 +182,7 @@ def parse_podcast_episode(
     if episode.published_at is not None:
         position = -episode.published_at
         # abs published_at is ms epoch
-        release_date = datetime.fromtimestamp(episode.published_at / 1000)
+        release_date = datetime.fromtimestamp(episode.published_at / 1000, tz=UTC)
```
marcelveldt, line 185: future improvement: we should consider adding a small helper in our helpers/datetime.py to do this

marcelveldt (approved):
Thanks, @OzGav !

music_assistant/providers/audiobookshelf/parsers.py
OzGav: Actually the helper already exists! I will do a quick PR to actually use it!

PR #3527, opened 2026-04-01
Dynamic playlist queue support for is_dynamic playlists

dmoo500:
## Overview

Adds support for `is_dynamic` playlists in the queue controller, enabling providers to supply playlists where tracks are fetched on demand (e.g. radio stations, AI-generated playlists). Includes refactoring of `player_queues.py` to improve clarity and correctness.

### `controllers/player_queues.py`

**Dynamic playlist support:**
- When the end of a queue is reached, check if the queue was started from a dynamic playlist (`is_dynamic=True`). If so, fetch a fresh batch of tracks from the provider and continue playback — instead of stopping or falling back to generic radio mode.
- In `play_media`, detect dynamic playlists earlier in the for-loop and fetch the initial track batch directly from the provider.
- Restore stale `is_dynamic` guard: if a library playlist has `is_dynamic=False` but the provider item says otherwise, correct it in-memory before enqueueing.

**Refactoring:**
- Extract `_enqueue_with_option()` helper from `play_media` (~65 lines), reducing method size and improving readability.
- Extract `_update_current_index_from_player()` from `_update_queue_from_player()`, separating index/elapsed-time resolution from the state-comparison logic.
- Restore session-ID guard in `_enqueue_next_item`: bail out if the queue session changed between scheduling and executing the enqueue callback.
- Reduce `call_later` delay from 1s to 0.5s for snappier track transitions.

### `controllers/media/base.py`
- `MediaControllerBase.get()` now accepts provider share URLs as `item_id` (Spotify, Tidal, Apple Music, …). The URL is resolved via `parse_uri()` to the actual provider and item ID before lookup. This is a generic controller improvement that benefits all providers.

### `controllers/media/playlists.py`
- Propagate the `is_dynamic` flag from provider items to library items during sync.
- When `allow_dynamic_tracks=True`, always force a cache refresh so stale empty responses don't silently prevent refill.
- Guard: when `allow_dynamic_tracks=False`, return early if the playlist is dynamic — hiding station tracks from browse/library views.

### `controllers/music.py`
- DB schema v36: add `is_dynamic` column to the playlists table.
- Backfill migration for existing Apple Music station playlists (`ra.%` provider item IDs).

## Notes
- The previous PR #3432 was not completely correct to handle `is_dynamic`.
- Provider-specific changes (Apple Music URL import, station endpoint routing, URI helpers) are in a separate PR #3433, which depends on this PR being merged first.

### DIFF
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast, final
 from music_assistant_models.enums import EventType, ExternalID, MediaType, ProviderFeature
 from music_assistant_models.errors import (
     InsufficientPermissions,
+    InvalidProviderID,
     InvalidProviderURI,
     MediaNotFoundError,
     ProviderUnavailableError,
@@ -31,6 +32,7 @@ from music_assistant.constants import (
     DB_TABLE_PLAYLOG,
     DB_TABLE_PROVIDER_MAPPINGS,
     MASS_LOGGER_NAME,
+    PROVIDERS_WITH_SHAREABLE_URLS,
 )
 from music_assistant.controllers.webserver.helpers.auth_middleware import get_current_user
 from music_assistant.helpers.compare import compare_media_item, create_safe_string
@@ -341,17 +343,17 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
         :param allow_update_metadata: Schedule a metadata refresh on access.
             Set to False when fetching items in bulk (e.g. provider sync).
         """
-        # if item_id is a provider share URL, resolve the actual provider and item_id
+        # if item_id is a share URL from a known provider, resolve to actual provider + item_id
         if item_id.startswith(("http://", "https://")):
-            with suppress(InvalidProviderURI):
-                parsed_media_type, provider_instance_id_or_domain, item_id = await parse_uri(
-                    item_id
+            with suppress(InvalidProviderURI, InvalidProviderID):
+                parsed_media_type, parsed_provider, parsed_item_id = await parse_uri(
+                    item_id, validate_id=True
                 )
-                if parsed_media_type != self.media_type:
+                if parsed_provider in PROVIDERS_WITH_SHAREABLE_URLS:
                     return cast(
                         "ItemCls",
                         await self.mass.music.get_item(
-                            parsed_media_type, item_id, provider_instance_id_or_domain
+                            parsed_media_type, parsed_item_id, parsed_provider
                         ),
                     )
         # always prefer the full library item if we have it
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -13,8 +13,6 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast, final
 from music_assistant_models.enums import EventType, ExternalID, MediaType, ProviderFeature
 from music_assistant_models.errors import (
     InsufficientPermissions,
-    InvalidProviderID,
-    InvalidProviderURI,
     MediaNotFoundError,
     ProviderUnavailableError,
 )
@@ -32,13 +30,11 @@ from music_assistant.constants import (
     DB_TABLE_PLAYLOG,
     DB_TABLE_PROVIDER_MAPPINGS,
     MASS_LOGGER_NAME,
-    PROVIDERS_WITH_SHAREABLE_URLS,
 )
 from music_assistant.controllers.webserver.helpers.auth_middleware import get_current_user
 from music_assistant.helpers.compare import compare_media_item, create_safe_string
 from music_assistant.helpers.database import UNSET
 from music_assistant.helpers.json import json_loads, serialize_to_json
-from music_assistant.helpers.uri import parse_uri
 from music_assistant.helpers.util import guard_single_request, parse_optional_bool
 
 if TYPE_CHECKING:
@@ -343,19 +339,6 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
         :param allow_update_metadata: Schedule a metadata refresh on access.
             Set to False when fetching items in bulk (e.g. provider sync).
         """
-        # if item_id is a share URL from a known provider, resolve to actual provider + item_id
-        if item_id.startswith(("http://", "https://")):
-            with suppress(InvalidProviderURI, InvalidProviderID):
-                parsed_media_type, parsed_provider, parsed_item_id = await parse_uri(
-                    item_id, validate_id=True
-                )
-                if parsed_provider in PROVIDERS_WITH_SHAREABLE_URLS:
-                    return cast(
-                        "ItemCls",
-                        await self.mass.music.get_item(
-                            parsed_media_type, parsed_item_id, parsed_provider
-                        ),
-                    )
         # always prefer the full library item if we have it
         if library_item := await self.get_library_item_by_prov_id(
             item_id,
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast, final
 from music_assistant_models.enums import EventType, ExternalID, MediaType, ProviderFeature
 from music_assistant_models.errors import (
     InsufficientPermissions,
+    InvalidProviderURI,
     MediaNotFoundError,
     ProviderUnavailableError,
 )
@@ -35,6 +36,7 @@ from music_assistant.controllers.webserver.helpers.auth_middleware import get_cu
 from music_assistant.helpers.compare import compare_media_item, create_safe_string
 from music_assistant.helpers.database import UNSET
 from music_assistant.helpers.json import json_loads, serialize_to_json
+from music_assistant.helpers.uri import parse_uri
 from music_assistant.helpers.util import guard_single_request, parse_optional_bool
 
 if TYPE_CHECKING:
@@ -339,6 +341,19 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
         :param allow_update_metadata: Schedule a metadata refresh on access.
             Set to False when fetching items in bulk (e.g. provider sync).
         """
+        # if item_id is a provider share URL, resolve the actual provider and item_id
+        if item_id.startswith(("http://", "https://")):
+            with suppress(InvalidProviderURI):
+                parsed_media_type, provider_instance_id_or_domain, item_id = await parse_uri(
+                    item_id
+                )
+                if parsed_media_type != self.media_type:
+                    return cast(
+                        "ItemCls",
+                        await self.mass.music.get_item(
+                            parsed_media_type, item_id, provider_instance_id_or_domain
+                        ),
+                    )
         # always prefer the full library item if we have it
         if library_item := await self.get_library_item_by_prov_id(
             item_id,
--- a/music_assistant/controllers/media/playlists.py
+++ b/music_assistant/controllers/media/playlists.py
@@ -116,11 +116,33 @@ class PlaylistController(MediaControllerBase[Playlist]):
         item_id: str,
         provider_instance_id_or_domain: str,
         force_refresh: bool = False,
+        allow_dynamic_tracks: bool = False,
     ) -> AsyncGenerator[PlaylistPlayableItem, None]:
         """Return playlist tracks for the given provider playlist id."""
+        library_item: Playlist | None = None
         if provider_instance_id_or_domain == "library":
             library_item = await self.get_library_item(item_id)
             provider_instance_id_or_domain, item_id = self._select_provider_id(library_item)
+        else:
+            library_item = await self.get_library_item_by_prov_id(
+                item_id, provider_instance_id_or_domain
+            )
+
+        # Dynamic playlists should not expose a static track list in browse/refresh views.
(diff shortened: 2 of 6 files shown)


music_assistant/controllers/media/base.py
```diff
@@ -339,6 +341,19 @@ async def get(
...
+                parsed_media_type, provider_instance_id_or_domain, item_id = await parse_uri(
+                    item_id
+                )
+                if parsed_media_type != self.media_type:
+                    return cast(
+                        "ItemCls",
+                        await self.mass.music.get_item(
+                            parsed_media_type, item_id, provider_instance_id_or_domain
+                        ),
+                    )
```
marcelveldt: where is this coming from ? this should not happen
please remove this

music_assistant/controllers/media/playlists.py
```diff
@@ -116,11 +116,37 @@ async def tracks(
         item_id: str,
         provider_instance_id_or_domain: str,
         force_refresh: bool = False,
+        allow_dynamic_tracks: bool = False,
```
marcelveldt, line 119: why would this be optional ?

music_assistant/controllers/player_queues.py
```diff
@@ -566,6 +566,17 @@ async def play_media(
...
+                # stale False value if the item was added before is_dynamic tracking was introduced.
+                if isinstance(media_item, Playlist) and not media_item.is_dynamic:
+                    with suppress(MusicAssistantError):
+                        provider_item = await self.mass.music.playlists.get_provider_item(
+                            media_item.item_id,
+                            media_item.provider,
+                        )
+                        if provider_item.is_dynamic:
+                            media_item.is_dynamic = True
+
```
marcelveldt: how is that possible because you ran the db migration ?
maybe the cache db ?

music_assistant/controllers/media/base.py
dmoo500: The issue is that when a user imports an Apple Music share URL (e.g. https://music.apple.com/de/station/.../ra.xxx) via the `music/radios/get_radio`API endpoint  (on UI "Add Item from URL") , it goes directly through `base.get()` → `get_provider_item("builtin")` → ffprobe, bypassing `music.get_item()` and `search()` entirely. I tried moving the resolution to `music.get_item()` but it doesn't intercept that code path.

I've restricted it to PROVIDERS_WITH_SHAREABLE_URLS only (currently spotify, qobuz, `apple_music`) so it won't affect other providers. Without this, share URLs from those providers will always fall through to the builtin provider and cause an ffprobe crash. The change is not yet pushed — will add it to this PR.

music_assistant/controllers/media/playlists.py
dmoo500: It's optional with False as default so that browse/library views that call `tracks()` never expose tracks from dynamic playlists — dynamic playlists have no stable track list, and surfacing them in browse/library would return incomplete or stale results. Only the queue controller passes True to explicitly request a fresh batch for playback. This keeps all existing callers unchanged.

music_assistant/controllers/player_queues.py
dmoo500: This was likely added during cross-PR testing before the DB migration was in place — at that point `is_dynamic` was always False in library items. With migration v36 backfilling the value, it's redundant and I'll remove it.

music_assistant/controllers/media/base.py
marcelveldt: hmmm but then it is not caught by the code that parses these share URL's that should be in the helpers somewhere to parse an URI

also, are these real radio stations ? why would a user add an apple playlist that way ? it feels weird

music_assistant/controllers/media/playlists.py
marcelveldt: ah yeah, fair point.

music_assistant/controllers/media/base.py
dmoo500: This change was removed. Users can now discover algorithmic stations (artist radio, personal mixes) directly via `Browse` → `Radio Stations` in Apple Music, making URL import unnecessary for this use case. (see last commit on PR #3433)

music_assistant/controllers/player_queues.py
```diff
@@ -1916,6 +1928,7 @@ async def get_playlist_tracks(
             playlist.item_id,
             playlist.provider,
             force_refresh=force_refresh,
+            allow_dynamic_tracks=True,
```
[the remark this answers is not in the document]
dmoo500: The concern is valid in theory, but in practice `get_playlist_tracks()]` is only called from this path during active playback — never during bulk sync. The fix `playlist.is_dynamic` would also be correct, but the current code is safe because the allow_dynamic_tracks guard in `PlaylistController.tracks()` only forces `force_refresh=True`, which just bypasses the in-memory cache for one call. The overhead is negligible for a single queue refill operation.

music_assistant/controllers/media/playlists.py
```diff
@@ -116,11 +116,37 @@ async def tracks(
...
             library_item = await self.get_library_item(item_id)
             provider_instance_id_or_domain, item_id = self._select_provider_id(library_item)
+        else:
+            library_item = await self.get_library_item_by_prov_id(
+                item_id, provider_instance_id_or_domain
+            )
+
+        # Dynamic playlists always need fresh tracks from the provider.
+        if allow_dynamic_tracks:
+            force_refresh = True
```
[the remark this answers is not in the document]
dmoo500: This is addressed transitively: the only caller that passes `allow_dynamic_tracks=True` is `get_playlist_tracks() in the queue controller, which is only invoked for playlists that were already identified as dynamic. A non-dynamic playlist never reaches this code path with `allow_dynamic_tracks=True.

music_assistant/controllers/media/base.py
```diff
@@ -339,6 +343,19 @@ async def get(
...
+        if item_id.startswith(("http://", "https://")):
+            with suppress(InvalidProviderURI, InvalidProviderID):
+                parsed_media_type, parsed_provider, parsed_item_id = await parse_uri(
+                    item_id, validate_id=True
+                )
+                if parsed_provider in PROVIDERS_WITH_SHAREABLE_URLS:
+                    return cast(
+                        "ItemCls",
+                        await self.mass.music.get_item(
+                            parsed_media_type, parsed_item_id, parsed_provider
```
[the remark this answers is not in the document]
dmoo500: This code was removed in the latest commit — the URL resolution was moved to PR #3433 which handles it at the Apple Music provider level instead.

music_assistant/controllers/music.py
```diff
@@ -827,6 +827,18 @@ async def get_item(
...
+            if item_id.startswith(("http://", "https://")):
+                with suppress(InvalidProviderURI, InvalidProviderID):
+                    parsed_media_type, parsed_provider, parsed_item_id = await parse_uri(
+                        item_id, validate_id=True
+                    )
+                    if parsed_provider in PROVIDERS_WITH_SHAREABLE_URLS:
+                        return await self.get_item(
+                            media_type=parsed_media_type,
+                            item_id=parsed_item_id,
+                            provider_instance_id_or_domain=parsed_provider,
```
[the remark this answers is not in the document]
dmoo500: The `builtin` provider path is only hit when a user explicitly plays a share URL — never during bulk library sync. Not forwarding `allow_update_metadata` here is intentional since the resolved item should always be checked for fresh metadata on explicit user access.

music_assistant/controllers/player_queues.py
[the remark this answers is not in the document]
marcelveldt: not entirely true - force_refresh on large playlists (especially on severe rate limited providers like spotify) can make a difference that is huge. Like a large spotify playlist with 1000+ tracks will take up to a minute to load. Not a nice experience.

So we should avoid the "force refresh" at all times during actual playback or people will get very frustrated

music_assistant/controllers/media/playlists.py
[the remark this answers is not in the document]
marcelveldt: ah ok, that is better

music_assistant/controllers/media/playlists.py
```diff
@@ -116,11 +116,37 @@ async def tracks(
...
+
+        # Dynamic playlists should not expose a static track list in browse/refresh views.
+        # Only the playback queue may request tracks for dynamic refill.
+        if not allow_dynamic_tracks:
+            if library_item is not None and library_item.is_dynamic:
+                return
+            # Generic guard for requests where the library item may not carry is_dynamic yet.
```
marcelveldt: why may it not carry 'is_dynamic' yet ?

music_assistant/controllers/music.py
```diff
@@ -827,6 +827,18 @@ async def get_item(
             provider_instance_id_or_domain = "library"
         if provider_instance_id_or_domain == "builtin":
             # handle special case of 'builtin' MusicProvider which allows us to play regular url's
+            # but first check if the url is a share URL from a known provider
```
marcelveldt: I'm not understanding how we can have an item_id which is a share URL - that is a path that simply doenst exist. Only builtin provider can have actual HTTP url's

Aren't you conflating URI's with item id ?

music_assistant/controllers/player_queues.py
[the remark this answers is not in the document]
dmoo500: `allow_dynamic_tracks=True` is only ever passed for playlists that have already been verified as `is_dynamic=True` — either via the `media_item.is_dynamic` check in `play_media` or the `dynamic_playlist is not None` guard in `_fill_radio_tracks`. A regular Spotify playlist with 1000 tracks can never reach this path with `allow_dynamic_tracks=True`. The `force_refresh` only fires for dynamic stations, which by definition always need fresh tracks.

music_assistant/controllers/media/playlists.py
dmoo500: You're right — the fallback was redundant. The DB migration v36 backfills `is_dynamic` for all existing stations, so the library item always carries the correct value. Removed.

music_assistant/controllers/music.py
dmoo500: Agreed — `item_id` should never be a share URL at this point in the call chain. The URL resolution originally lived in `base.py` and was moved here during refactoring, but it no longer belongs in either place. Removed.

music_assistant/controllers/player_queues.py
```diff
@@ -603,6 +603,10 @@ async def play_media(
                     # Type guard for mypy - only add full MediaItemType to radio_source
                     if not isinstance(media_item, (ItemMapping, BrowseFolder)):
                         radio_source.append(media_item)
+                elif isinstance(media_item, Playlist) and media_item.is_dynamic:
+                    # Dynamic playlists supply their own tracks on demand; fetch first batch now.
+                    initial_tracks = await self.get_playlist_tracks(media_item, start_item=None)
+                    media_items += initial_tracks
```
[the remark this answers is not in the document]
dmoo500: `start_item=None` is intentional — dynamic playlists return a fresh provider-defined batch on every call, so a user-selected start position has no meaning. The `mark_item_played` point is valid — added.

music_assistant/controllers/media/playlists.py
```diff
@@ -116,11 +116,28 @@ async def tracks(
...
     ) -> AsyncGenerator[PlaylistPlayableItem, None]:
         """Return playlist tracks for the given provider playlist id."""
+        library_item: Playlist | None = None
         if provider_instance_id_or_domain == "library":
             library_item = await self.get_library_item(item_id)
             provider_instance_id_or_domain, item_id = self._select_provider_id(library_item)
+        else:
+            library_item = await self.get_library_item_by_prov_id(
+                item_id, provider_instance_id_or_domain
+            )
```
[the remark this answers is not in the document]
dmoo500: Good catch. The `get_library_item_by_prov_id` lookup is only needed for the `not allow_dynamic_tracks` early-return guard. When `allow_dynamic_tracks=True`, it adds an unnecessary DB query. Fixed — the lookup is now skipped in that path.

music_assistant/controllers/player_queues.py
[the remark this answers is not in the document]
dmoo500: You're right, and I apologise — my earlier reply claiming this was safe ("only called for dynamic playlists") was incorrect. `get_playlist_tracks()` itself had no guard, and `allow_dynamic_tracks=True` would have forced `force_refresh=True` inside `PlaylistController.tracks()` even for non-dynamic playlists if the call chain ever changed. The issue only became fully clear to me after marcelveldt's note about Spotify playlists. Fixed in the latest commit: `allow_dynamic_tracks=playlist.is_dynamic` is now passed, consistent with the `force_refresh` logic directly above it.

music_assistant/controllers/player_queues.py
```diff
@@ -1916,6 +1933,7 @@ async def get_playlist_tracks(
             playlist.item_id,
             playlist.provider,
             force_refresh=force_refresh,
+            allow_dynamic_tracks=True,
```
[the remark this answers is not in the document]
dmoo500: You're right, and I apologise — my earlier reply claiming this was safe ("only called for dynamic playlists") was incorrect. `get_playlist_tracks()` itself had no guard, and `allow_dynamic_tracks=True` would have forced `force_refresh=True` inside `PlaylistController.tracks()` even for non-dynamic playlists if the call chain ever changed. The issue only became fully clear to me after marcelveldt's note about Spotify playlists. Fixed in the latest commit: `allow_dynamic_tracks=playlist.is_dynamic` is now passed, consistent with the `force_refresh` logic directly above it.

marcelveldt (approved):
Looks good!
Nice work, @dmoo500

PR #3538, opened 2026-04-02
Mypy fixes for the Sonos provider

OzGav:


music_assistant/providers/sonos/player.py
```diff
@@ -156,6 +164,7 @@ async def setup(self) -> None:
         await self.mass.players.register_or_update(self)
 
         # register callback for state changed
+        # aiosonos typed event_filter as tuple[EventType] but accepts multiple elements
```
OzGav:
```suggestion
```

music_assistant/providers/sonos/player.py
```diff
@@ -86,9 +87,15 @@ def __init__(
...
         self.discovery_info = discovery_info
         self.connected: bool = False
-        self._listen_task: asyncio.Task | None = None
+        self._listen_task: asyncio.Task[None] | None = None
         self.sonos_queue: SonosQueue = SonosQueue()
 
+    @property
+    def group_controller(self) -> SonosGroup:
+        """Get the group controller, raising if unavailable."""
+        assert self.client.player.group is not None, "Group controller unavailable"
```
marcelveldt: assert may only be used for type hinting, not for runtime checks

music_assistant/providers/sonos/player.py
```diff
@@ -106,6 +113,7 @@ def synced_to(self) -> str | None:
 
     async def setup(self) -> None:
         """Handle setup of the player."""
+        assert self.device_info.ip_address is not None, "IP address required for setup"
```
marcelveldt: assert may only be used for type hinting, not for runtime checks

in case an assert is used, add a comment that it is for type hint

music_assistant/providers/sonos/player.py
OzGav: Fixed

music_assistant/providers/sonos/player.py
OzGav: comment added

music_assistant/providers/sonos/player.py
```diff
@@ -341,17 +351,21 @@ async def play_media(
...
-        await self.client.player.group.play_stream_url(
+        await self.group_controller.play_stream_url(
             stream_url,
-            {
-                "name": media.title,
-                "type": "track",
-                "imageUrl": media.image_url,
-                "id": {
-                    "objectId": object_id,
+            # Sonos API accepts this dict format but aiosonos Container TypedDict is stricter
```
marcelveldt: yikes, cant we fix this in aiosonos then ? this looks a bit awful tbh

music_assistant/providers/sonos/player.py
```diff
@@ -571,12 +587,12 @@ def update_attributes(self) -> None:  # noqa: PLR0915
...
             track_duration_millis = track.get("durationMillis")
             current_media = PlayerMedia(
-                uri=track.get("id", {}).get("objectId") or track.get("mediaUrl"),
+                uri=track.get("id", {}).get("objectId") or track.get("mediaUrl") or "",
                 media_type=MediaType.TRACK,
                 title=track["name"],
                 artist=track.get("artist", {}).get("name"),
                 album=track.get("album", {}).get("name"),
-                duration=track_duration_millis / 1000 if track_duration_millis else None,
+                duration=int(track_duration_millis / 1000) if track_duration_millis else None,
```
marcelveldt, line 597: isnt duration a float ?

music_assistant/providers/sonos/provider.py
```diff
@@ -72,17 +74,18 @@ async def on_mdns_service_state_change(
...
+        assert isinstance(player_id, str)  # for type checking
         # handle update for existing device
         if sonos_player := self.mass.players.get_player(player_id):
-            assert isinstance(sonos_player, SonosPlayer), (
-                "Player ID already exists but is not a SonosPlayer"
-            )
-            # if mass_player := sonos_player.mass_player:
+            if not isinstance(sonos_player, SonosPlayer):
+                msg = f"Player {player_id} already exists but is not a SonosPlayer"
+                raise TypeError(msg)
```
marcelveldt: I think you could also have left the assert as this is really for type checking only.
there is no way that we can get another player back at runtime

music_assistant/providers/sonos/player.py
OzGav: It is an int here https://github.com/music-assistant/models/blob/209dc630d0f73ce2ef9096a77f32ea2f667b937b/music_assistant_models/player.py#L111

music_assistant/providers/sonos/player.py
OzGav: OK I think what needs to be fixed is here https://github.com/music-assistant/aiosonos/pull/127

music_assistant/providers/sonos/provider.py
OzGav:
```suggestion
            assert isinstance(sonos_player, SonosPlayer), (
                "Player ID already exists but is not a SonosPlayer"
            )
            # if mass_player := sonos_player.mass_player:
```

OzGav: Reverted

music_assistant/providers/sonos/player.py
OzGav: Fixed with aiosonos 0.1.12

music_assistant/providers/sonos/player.py
marcelveldt: ugh

music_assistant/providers/sonos/player.py
marcelveldt: did you bump it in this PR?

marcelveldt (approved):
Thanks, @OzGav !

PR #3539, opened 2026-04-02
Guard against non-UTF-8 filenames in file system providers

OzGav:
Hopefully fixes: https://github.com/orgs/music-assistant/discussions/5198

IN that discussion recurring errors were reported during library sync and websocket event handling caused by a file with a non-UTF-8 character in its filename. The surrogate character propagated through the system causing multiple failure points:

- Library sync failed to process the file and logged an error that itself crashed the logging handler
- Every websocket event carrying the filename caused an unhandled exception in `_send_message_sync`, spamming the log
- A config save also failed when the surrogate reached orjson

Changes:
`music_assistant/helpers/tags.py` — `AudioTags.parse()` now checks the filename for non-UTF-8 characters before doing anything else and raises `InvalidDataError` with the offending filename, skipping the file cleanly. The error message construction in `parse_tags()` is also guarded so that logging the skip does not itself crash.

### DIFF
--- a/music_assistant/controllers/webserver/websocket_client.py
+++ b/music_assistant/controllers/webserver/websocket_client.py
@@ -297,7 +297,20 @@ class WebsocketClientHandler:
 
         Serializes inline without executor overhead since events are typically small.
         """
-        _message = message.to_json()
+        try:
+            _message = message.to_json()
+        except (UnicodeEncodeError, TypeError):
+            if isinstance(message, MassEvent):
+                self._logger.warning(
+                    "Dropping websocket event '%s' for '%s': contains non-UTF-8 characters.",
+                    message.event,
+                    message.object_id,
+                )
+            else:
+                self._logger.warning(
+                    "Dropping websocket message: contains non-UTF-8 characters.",
+                )
+            return
 
         try:
             self._to_write.put_nowait(_message)
--- a/music_assistant/controllers/webserver/websocket_client.py
+++ b/music_assistant/controllers/webserver/websocket_client.py
@@ -297,20 +297,7 @@ class WebsocketClientHandler:
 
         Serializes inline without executor overhead since events are typically small.
         """
-        try:
-            _message = message.to_json()
-        except (UnicodeEncodeError, TypeError):
-            if isinstance(message, MassEvent):
-                self._logger.warning(
-                    "Dropping websocket event '%s' for '%s': contains non-UTF-8 characters.",
-                    message.event,
-                    message.object_id,
-                )
-            else:
-                self._logger.warning(
-                    "Dropping websocket message: contains non-UTF-8 characters.",
-                )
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/webserver/websocket_client.py
```diff
@@ -297,7 +297,20 @@ def _send_message_sync(self, message: MessageType) -> None:
...
+                self._logger.warning(
+                    "Dropping websocket event '%s' for '%s': contains non-UTF-8 characters.",
+                    message.event,
+                    message.object_id,
+                )
+            else:
+                self._logger.warning(
+                    "Dropping websocket message: contains non-UTF-8 characters.",
+                )
+            return
```
marcelveldt: Let's leave out this change for now. I'd like this error to propagate instead of being swallowed.
If that happens, we have a more serious issue goin on.

music_assistant/helpers/tags.py
```diff
@@ -602,6 +602,13 @@ def track_album_loudness(self) -> float | None:
     @classmethod
     def parse(cls, raw: dict[str, Any]) -> AudioTags:
         """Parse instance from raw ffmpeg info output."""
+        filename = raw["format"]["filename"]
+        try:
+            filename.encode("utf-8")
+        except UnicodeEncodeError:
+            raise InvalidDataError(
+                f"File skipped, filename contains non-UTF-8 characters: {filename!r}"
```
marcelveldt, line 610: Can we find out how we got here in the first place ?
Can we end up in a situation where any of the filebased providers uses another encoding ?

music_assistant/controllers/webserver/websocket_client.py
OzGav: Reverted

music_assistant/helpers/tags.py
OzGav: The answer is that this isn't an encoding choice made by MA — it comes from the OS/filesystem layer.

I understand that when Python's os module (or ffprobe via subprocess) encounters a filename that contains bytes that aren't valid UTF-8, it decodes those bytes using surrogateescape error handler. This is Python's way of preserving the raw bytes without losing data, but it produces surrogate characters that are not valid Unicode and cannot be re-encoded as UTF-8.

In the specific case in that discussion the filename was written on Windows using a filesystem or tool that used Windows-1252 encoding. When that file was accessed on the Linux system (inside the Docker container) via a volume mount, the kernel handed the raw bytes to Python which applied surrogateescape, producing the \udc92 surrogate.
So it's not MA choosing a different encoding — it's the filesystem handing MA a filename that was never valid UTF-8 to begin with. The guard is there because MA has no control over what filenames users put on their music drives.

marcelveldt (approved):
Thanks, @OzGav !

PR #3550, opened 2026-04-03
Fix filesystem provider sync config checkboxes not being respected

teancom:
I saw Dawsnet in the discord mention these as issues, and @marcelveldt pointed me in the right direction. 

The sync config checkboxes (e.g. "Import tracks/files into the Music
Assistant library") allow users to control whether files are added to
the library database during sync, or only available via Browse. These
checkboxes were ignored — the provider imported everything on sync
regardless of settings.

Additionally, once a filesystem provider instance was saved and its
config page reopened, phantom "Sync Library Albums" and "Sync Library
Artists" checkboxes appeared. These were auto-injected by the config
controller because the provider advertised LIBRARY_ALBUMS and
LIBRARY_ARTISTS in supported_features. Those features are meaningless
for the filesystem provider — albums and artists are always derived
from track imports, never synced independently.

Changes:
- _process_item_async now checks the sync config value before
  importing each media type (tracks, playlists, audiobooks, podcasts).
- supported_features no longer advertises LIBRARY_ALBUMS or
  LIBRARY_ARTISTS, preventing the phantom checkboxes.
- Removed SUPPORTED_FEATURES constant from constants.py (dead since
  #2405 added a replacement in __init__.py).
- Added tests for supported_features across all content types.

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -190,6 +190,8 @@ class LocalFileSystemProvider(MusicProvider):
         self.base_path: str = base_path
         self.write_access: bool = False
         self.sync_running: bool = False
+        self._sync_tracks: bool = True
+        self._sync_playlists: bool = True
         self.media_content_type = cast("str", config.get_value(CONF_ENTRY_CONTENT_TYPE.key))
 
     @property
@@ -333,6 +335,22 @@ class LocalFileSystemProvider(MusicProvider):
         if media_type in (MediaType.ARTIST, MediaType.ALBUM):
             # artists and albums are synced as part of track sync
             return
+        # check if any sync options are enabled for this content type
+        # the filesystem provider processes all file types in one scan,
+        # so we can return early if nothing needs syncing
+        if self.media_content_type == "music":
+            self._sync_tracks = bool(self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_TRACKS.key))
+            self._sync_playlists = bool(
+                self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_PLAYLISTS.key)
+            )
+            if not self._sync_tracks and not self._sync_playlists:
+                return
+        elif self.media_content_type == "audiobooks":
+            if not self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_AUDIOBOOKS.key):
+                return
+        elif self.media_content_type == "podcasts":
+            if not self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_PODCASTS.key):
+                return
         assert self.mass.music.database
         start_time = time.time()
         if self.sync_running:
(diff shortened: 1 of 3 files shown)


music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -197,16 +197,20 @@ def supported_features(self) -> set[ProviderFeature]:
         """Return the features supported by this Provider."""
         base_features = {*SUPPORTED_FEATURES}
         if self.media_content_type == "audiobooks":
-            return {ProviderFeature.LIBRARY_AUDIOBOOKS, *base_features}
+            if self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_AUDIOBOOKS.key):
+                base_features.add(ProviderFeature.LIBRARY_AUDIOBOOKS)
```
marcelveldt: This is incorrect. The fact that you don't want to sync them doesn't mean it doesnt provide a library. Although that is not how its handled in any of the other providers.
So, or we do the logic at a global level or not at all but this is wrong.

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -197,16 +197,20 @@ def supported_features(self) -> set[ProviderFeature]:
...
+            if self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_PODCASTS.key):
+                base_features.add(ProviderFeature.LIBRARY_PODCASTS)
+            return base_features
+        # Albums and artists are derived from track imports; not advertised as
+        # independent features to prevent phantom sync checkboxes in the UI.
+        music_features = {*base_features}
+        if self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_TRACKS.key):
+            music_features.add(ProviderFeature.LIBRARY_TRACKS)
+        if self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_PLAYLISTS.key):
+            music_features.add(ProviderFeature.LIBRARY_PLAYLISTS)
```
marcelveldt: this should be set based on the media_type that is configured, not the sync preference.

marcelveldt (changes requested):
Sorry, this is not the right approach to fix this

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -444,6 +444,8 @@ async def _process_item_async(self, item: FileSystemItem, prev_checksum: str | N
             self.logger.log(VERBOSE_LOG_LEVEL, "Processing: %s", item.relative_path)
 
             if item.ext in TRACK_EXTENSIONS and self.media_content_type == "music":
+                if not self.config.get_value(CONF_ENTRY_LIBRARY_SYNC_TRACKS.key):
```
marcelveldt: should we do this on the sync_library method instead so we can return early ?

teancom: As per this (and our discussion in Discord), done, with the caveat that this approach means we do not remove items the local provider added until/unless the user removes the provider completely.

marcelveldt (approved):
Thanks, @teancom !

PR #3552, opened 2026-04-03
Bandcamp: fix Liskov substitution violation in get_artist signature

teancom:
The override accepted `str | int` while the base class declares `str`. Convert band_id to str at the call site instead of widening the parameter type.


marcelveldt (approved):
Thanks, @teancom !

PR #3554, opened 2026-04-03
Clean up leaked throttlers, command locks, and protocol evaluations on player unregister

teancom:
unregister() in PlayerController never cleans up _player_throttlers, _player_command_locks, or _pending_protocol_evaluations. In environments with frequent player discovery (Chromecast, AirPlay, DLNA), entries for departed players accumulate indefinitely, and stale timer handles may fire after the player is gone.


marcelveldt (approved):
Nice!

PR #3555, opened 2026-04-03
Add Infinite Mix dynamic builtin playlists

dmoo500:
## Summary

Adds two new built-in playlists that use \`Playlist.is_dynamic=True\` to deliver an endless, randomly-ordered playback experience from the local library — without pre-loading the full track list upfront.

### New playlists

| ID | Name | Source |
|---|---|---|
| \`infinite_mix\` | Infinite Mix (library) | 25 random tracks from the full library |
| \`infinite_mix_favorites\` | Infinite Mix (favorites) | 25 random tracks from favorited tracks only |

### How it works

- Both playlists set \`is_dynamic=True\` on the returned \`Playlist\` object.
- \`get_playlist_tracks()\` returns 25 tracks per call with \`order_by="random"\` — **no cache**, so every refill yields a fresh random batch.
- The queue controller (#3527, already merged) detects \`is_dynamic=True\` at end-of-queue and automatically calls \`get_playlist_tracks(force_refresh=True)\` to append the next 25 tracks.
- Memory usage stays constant regardless of library size — no 50k-track pre-expansion.
- UI descriptions are included as \`ConfigEntry.description\` (English fallback; translatable via the frontend repo using keys \`settings.infinite_mix\` / \`settings.infinite_mix_favorites\`).

### What this replaces / complements

The existing \`RANDOM_TRACKS\` playlist loads up to 500 tracks upfront and caches them for 2 minutes. The Infinite Mix playlists load 25 tracks on demand with no cap and no cache — more suitable for long listening sessions.

Closes / implements the "Infinite Mix" use case from [discussion #5129](https://github.com/orgs/music-assistant/discussions/5129).

### DIFF
--- a/music_assistant/providers/builtin/__init__.py
+++ b/music_assistant/providers/builtin/__init__.py
@@ -84,6 +84,9 @@ from .constants import (
     CONF_KEY_TRACKS,
     DEFAULT_FANART,
     DEFAULT_THUMB,
+    DYNAMIC_BUILTIN_PLAYLISTS,
+    INFINITE_MIX,
+    INFINITE_MIX_FAVORITES,
     RANDOM_ALBUM,
     RANDOM_ARTIST,
     RANDOM_TRACKS,
@@ -255,6 +258,7 @@ class BuiltinProvider(MusicProvider):
                 },
                 owner="Music Assistant",
                 is_editable=False,
+                is_dynamic=prov_playlist_id in DYNAMIC_BUILTIN_PLAYLISTS,
                 metadata=MediaItemMetadata(
                     images=UniqueList([DEFAULT_THUMB])
                     if prov_playlist_id in COLLAGE_IMAGE_PLAYLISTS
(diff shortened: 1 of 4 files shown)


dmoo500:
## Future work: seed-based Infinite Mixes

This PR intentionally keeps scope small (library-wide + favorites). During development we also evaluated genre-, album- and track-seeded variants:

- **Genre mix** — technically trivial (\`library_items(genre=id)\`), but would pollute the builtin playlist list proportionally to library size (users with 20+ genres end up with 20+ extra playlists). Skipped for now.
- **Album / track seeds** — e.g. "25 random tracks similar to this album/track". Not feasible without audio analysis or an external similarity source (Last.fm, ListenBrainz). Needs a dedicated PR once a similarity signal is available.

These will be tracked as separate issues/PRs.

marcelveldt (approved):
Looks good, nice QoL improvement!

PR #3561, opened 2026-04-03
Plex: fix streaming of newly added Plex tracks

anatosun:
Use ?download=1 URL for all track types instead of the plain part URL or the transcode streaming endpoint, both of which return HTTP 500 for tracks that haven't been fully processed by the Plex media scanner yet. The download endpoint bypasses Plex's streaming pipeline and serves the raw file directly, making newly added tracks immediately playable.

Also guard against empty audioStreams() for unanalyzed tracks.

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -1284,7 +1284,8 @@ class PlexProvider(MusicProvider):
             ContentType.try_parse(media.container) if media.container else ContentType.UNKNOWN
         )
         media_part: PlexMediaPart = media.parts[0]
-        audio_stream: PlexAudioStream = media_part.audioStreams()[0]
+        audio_streams = media_part.audioStreams()
+        audio_stream: PlexAudioStream | None = audio_streams[0] if audio_streams else None
 
         stream_details = StreamDetails(
             item_id=plex_track.key,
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):

campbellwray:
I see this fix has already been merged, but I made [this comment](https://github.com/orgs/music-assistant/discussions/3810#discussioncomment-16464644) a couple of days ago, thinking it was weird behaviour, but today I discovered that this was actually changed as part of this PR.

I think this should be considered a regression, since this isn't how Plex is intended to be used (and per the docs could actually lead to a worse experience). Additionally I'm not sure if this will work for non-Plex Pass users, isn't downloading a Plex Pass only feature?

The [documentation](https://github.com/orgs/music-assistant/discussions/3810#discussioncomment-16464644) for this method states:
> Bandwidth will be prioritized behind playbacks and will get a fair share of what remains.

Would it be possible to achieve this in a conditional manner? i.e. if the track is unprocessed it downloads it, otherwise it streams it?

anatosun:
Hi @campbellwray ,

That's a fair point, I could actually implement a conditional pipeline. However, the playback is not reported if we call the stream url. To report (and scrobble) the playback, please have a look at the Plex Connect plugin.

anatosun:
Also, for issues and demands with Plex, please make sure to tag me; I did not follow the thread you linked above.

campbellwray:
@anatosun 

Yeah, I figured simply changing this one thing wouldn’t magically fix the reporting, I just wanted to let you know that I had noticed the difference in how plays appear in the Plex dashboard. I assume there is some kind of Plex API endpoint that needs to be called to inform the server when a play occurs, but I haven’t looked into it, and I don’t really know how music assistant works to know where to hook that in. 

I think I’ve seen elsewhere that plays are recorded when the track is 80% listened, is this something that music assistant supports? Some kind of callback system as playback progresses? 

Btw, I’m not saying that you need to make these changes, I’m just curious about the internals and how this might be possible. Also I’m very grateful of the work you’re doing here, I’ve been using MA + Plex + AirPlay together for a couple of months now and it seems to be getting better all the time. 

Edit:
Possibly the `/timeline` endpoint or the `/scrobble` endpoint

anatosun:
The playback reporting is done through the Plex Connect plugin, have you given it a try?

PR #3563, opened 2026-04-03
Enable Sendspin bridge for Cast stereo pairs

marcelveldt:
## Problem

Cast stereo pairs were excluded from the Sendspin Cast bridge, meaning they couldn't benefit from Sendspin-based playback. Users with stereo pair setups had no way to use the Sendspin protocol on these devices.

## Changes

- Remove the blanket exclusion of all audio groups from the Sendspin bridge, and instead only skip non-stereo-pair groups (regular Cast groups still use their own playback mechanism)
- Allow `STEREO_PAIR` player type in protocol linking so the Sendspin bridge player gets linked to the Cast stereo pair player via `CAST_UUID` matching (previously both `GROUP` and `STEREO_PAIR` were excluded)

> [!NOTE]
> This should be tested in nightly/beta first — I don't have access to a Cast stereo pair to verify this works in practice.

### DIFF
--- a/music_assistant/controllers/players/protocol_linking.py
+++ b/music_assistant/controllers/players/protocol_linking.py
@@ -110,10 +110,10 @@ class ProtocolLinkingMixin:
         if player.state.type == PlayerType.PROTOCOL:
             # Protocol player: try to find a native parent
             self._try_link_protocol_to_native(player)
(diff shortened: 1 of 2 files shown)


MacgyverH:
I'll give this a test once nightly build releases and let you know how it goes!

tbrasser:
amazing (untill I gut them for esphome pcb's ;) )

PR #3566, opened 2026-04-04
Fix audio analysis version gating for non-track media

marcelveldt:
## Problem
Audio analysis version gating looked up stored results without `media_type`, so an analysis for one media type could incorrectly suppress analysis for another item with the same provider/item ID.

## Changes
- include `media_type` in the audio analysis version lookup
- add regression coverage for media-type-aware version gating
- update the demo provider example to pass `media_type` when storing analysis


marcelveldt:
@MarvinSchenkel FYI

PR #3569, opened 2026-04-04
Add support for audiobook collections

fmunkes:
Depends on
https://github.com/music-assistant/models/pull/196

I'm testing with abs, but I can split that off from this PR once I'm out of drafting.

I added just a single api endpoint to get series. So the idea would be, that, if the users selects that, the api calls library_items and filters series out, and then calls series.


music_assistant/controllers/media/audiobooks.py
```diff
@@ -84,9 +87,16 @@ async def library_items(
         :param order_by: Order by field (e.g. 'sort_name', 'timestamp_added').
         :param provider: Filter by provider instance ID (single string or list).
         :param genre: Filter by genre id(s).
+        :param without_series: Do not return audiobooks which are part of a series
         """
         extra_query_params: dict[str, Any] = {}
         extra_query_parts: list[str] = []
+        self.logger.error(without_series)
```
MarvinSchenkel, line 94: I think this still needs to be cleaned up?

music_assistant/controllers/media/audiobooks.py
```diff
@@ -84,9 +87,16 @@ async def library_items(
...
+        if without_series:
+            extra_query_parts = [
+                "WHERE json_extract(audiobooks.metadata, '$.series') IS NULL "
```
MarvinSchenkel, line 97: Let's be defensive here and already add `(` and `)` around the OR

music_assistant/controllers/media/audiobooks.py
```diff
@@ -362,3 +372,57 @@ async def _set_playlog(self, db_id: int, media_item: Audiobook) -> None:
...
+        offset: int = 0,
+    ) -> list[AudiobookSeries]:
+        """Get all available audiobook series.
+
+        :param limit: Maximum number of items to return.
+        :param offset: Number of items to skip.
+        """
+        # key is the series' title
+        series_dict: dict[str, list[Audiobook]] = {}
+        audiobooks_with_series = await self.get_library_items_by_query(
```
MarvinSchenkel, line 388: I think limit and offset are now applied on AudioBooks here rather than on the series itself?

fmunkes: Hmm, that's true. I removed it completely now, as I don't think that I can apply this offset by pure db methods? The default limit for items is 500, and having more than 500 series would be extraordinary I think. Not certain here.

music_assistant/controllers/media/audiobooks.py
```diff
@@ -371,3 +380,55 @@ async def _set_playlog(self, db_id: int, media_item: Audiobook) -> None:
                 },
                 allow_replace=True,
             )
+
+    async def collections(
+        self,
+    ) -> list[AudiobookCollection]:
+        """Get all available audiobook collections.
```
OzGav:
```suggestion
        """
        Get all available audiobook collections.
```

music_assistant/controllers/media/audiobooks.py
```diff
@@ -73,6 +75,7 @@ async def library_items(
         order_by: str = "sort_name",
         provider: str | list[str] | None = None,
         genre: int | list[int] | None = None,
+        without_collections: bool | None = None,
         **kwargs: Any,
     ) -> list[Audiobook]:
         """Get in-database audiobooks.
```
OzGav:
```suggestion
        """
        Get in-database audiobooks.
```

music_assistant/controllers/media/audiobooks.py
OzGav:
```suggestion
        """
        Try to find match on all (streaming) providers for the provided (database) audiobook.
```

OzGav (approved):
Just some new nit formatting things then I think this is good to go?

OzGav:
Oh and there is a minor conflict :-(

fmunkes:
Thanks a lot for your review @OzGav!

PR #3570, opened 2026-04-04
Adapt artist / audiobook controller for authors and narrators

fmunkes:
Adds endpoints for authors/ narrators. Is that sql stuff actually safe?

Same question in reverse to https://github.com/music-assistant/server/pull/3569 can be asked here - would it make sense to make authors/ narrators media items, so things like pictures, or more metadata can be added. Also, the provider could be asked directly for their items, instead of searching the db.

I'm willing to do that, but what do you think?

### update after discussion below
depends on https://github.com/music-assistant/models/pull/209


marcelveldt:
Not sure how common it is that the authors/narrators have actual metadata ?
We could also say that these are mapped to the artist mediaitem (and we add a artist_type?)

I think it may make sense to quickly browse by narrator or author but not sure if we need to create a whole media type for them, I'm also wondering if we even have metadata. When I checked spotify etc. they did not provide any metadata for authors or narrators except the name

fmunkes:
> Not sure how common it is that the authors/narrators have actual metadata ? We could also say that these are mapped to > I think it may make sense to quickly browse by narrator or author but not sure if we need to create a whole media type for them, I'm also wondering if we even have metadata. When I checked spotify etc. they did not provide any metadata for authors or narrators except the name

I like the idea of an artist_type, and then enhancing the artist controller a bit. You are right, the data is usually name and maybe a picture - but since we then have a provider id, we can also ask the provider for the audiobooks belonging to the author/ narrator. Shall I go with the artist_type then?

I could also enhance the albums controller for series and introduce another album_type?

marcelveldt:
I'm not sure how it would work in practice to count authors and narrators as artists but I guess we can say the same for composers (classical music) so it's worth the investigation. would be nice if we can fix it.

As for series; can you enlighten me what those are ?
I'm not into podcasts that much so I need a bit of help here and there understanding how its structured and used in practice ?

fmunkes:
Series would be for audiobooks, not podcast (but why not include it, although I don't know if that exists).

A book series would be something like Lord of the rings with multiple books. Attributes would be the series name, picture, and the overall progress. So it is similar to an album (or playlist?) but if I would hit play on a series, my expectation would be that it starts where you left off, (e.g. book no 2) and not with the first one.

I guess essentially you can map all of these types to existing MA types with a little bit of extra information. What I can't judge is if it is more future proof to just keep this extra branches on an artist or album in mind, or just separate them completely with their own controller. Or put a controller between base controller and implementation, where both types can inherit from. Maintenance of course being important too.

I don't know how well Spotify's integration of audiobooks is, but series is definitely something you find e.g. in audible.

Of course, we can also just add them via attributes, and search the db for them. I really don't know what's best - maybe I'm to influenced by abs...
I would add least add support for this to abs and the local filesystem.

marcelveldt:
For putting narrators, authors and composers into artists with a dedicated type I'd say yes.
But with a sidenote that we cant make this exactly foolproof as often all we got is a single string with the name, no id or anything.

For series I'm not sure. Definitely not Album. To me it's more a metadata field which we can use for extra filtering and such. I guess you can compare this to collections within movies

fmunkes:
Ok, sounds good. I'll adapt both PRs, and mark them as draft for the time being. For series, what I really would like to have, is, that you can collapse them into a single visual item, instead of multiple Audiobooks. But that's a frontend thing I think, so I'll think about something there.

fmunkes:
Ok, I think this is ready for a first review.

music_assistant/models/music_provider.py
```diff
@@ -185,6 +185,20 @@ async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
         """
         raise NotImplementedError
 
+    async def get_author_audiobooks(self, prov_artist_id: str) -> list[Audiobook]:
+        """Get a list of all albums for the given artist.
```
MarvinSchenkel, line 189: albums -> audiobooks

music_assistant/models/music_provider.py
```diff
@@ -185,6 +185,20 @@ async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
...
+
+        Only called if provider supports ProviderFeature.AUTHOR_AUDIOBOOKS.
+        """
+        raise NotImplementedError
+
+    async def get_narrator_audiobooks(self, prov_artist_id: str) -> list[Audiobook]:
+        """Get a list of all albums for the given artist.
```
MarvinSchenkel, line 196: albums -> audiobooks

MarvinSchenkel:
Before doing an in-depth review, let's chat about the controller first. It feels a bit strange to have all this mixed logic in the ArtistController (eg Type = ARTIST or Type IN (Narrator , Author) ).

@marcelveldt would it be an idea to implement this as a child controller or generic?

marcelveldt:
A composer, author or writer is a form of Artist. Currently we only support singers but maybe we should broaden it. Can still belong in the Artist controller, just like we already have multiple Album types. It's just a type...

fmunkes:
I renamed the default  artist type to "singer", that should make it clearer then, that we are talking about music artists and so on.

music_assistant/controllers/media/artists.py
```diff
@@ -223,9 +260,17 @@ async def remove_item_from_library(self, item_id: str | int, recursive: bool = T
             with contextlib.suppress(MediaNotFoundError):
                 await self.mass.music.tracks.remove_item_from_library(db_row["track_id"])
 
-        # delete the artist itself from db
-        # this will raise if the item still has references and recursive is false
-        await super().remove_item_from_library(db_id)
+    async def _remove_author_narrator_from_library(self, db_id: int, recursive: bool) -> None:
+        # recursively also remove author/ narrator audiobooks
+        for db_row in await self.mass.music.database.get_rows_from_query(
+            f"SELECT album_id FROM {DB_TABLE_AUDIOBOOK_ARTISTS} WHERE artist_id = :artist_id",
```
MarvinSchenkel: I think `album_id` does not exist in the table?

music_assistant/controllers/media/artists.py
```diff
@@ -535,3 +598,146 @@ def artist_from_item_mapping(self, item: ItemMapping) -> Artist:
...
+        item_id: str,
+        provider_instance_id_or_domain: str,
+    ) -> list[Audiobook]:
+        """Return audiobooks for an author on given provider."""
+        assert provider_instance_id_or_domain != "library"
+        if not (prov := self.mass.get_provider(provider_instance_id_or_domain)):
+            return []
+        prov = cast("MusicProvider", prov)
+        if ProviderFeature.NARRATOR_AUDIOBOOKS in prov.supported_features:
+            return await prov.get_author_audiobooks(item_id)
```
MarvinSchenkel: Shouldn't this be `prov.get_narrator_audiobooks(item_id)` ?

music_assistant/controllers/media/artists.py
```diff
@@ -535,3 +598,146 @@ def artist_from_item_mapping(self, item: ItemMapping) -> Artist:
...
+            # return in-library items only
+            return result
+        # return all (unique) items from all providers
+        # initialize unique_ids with db_items to prevent duplicates
+        unique_ids: set[str] = {f"{item.name}.{item.version}" for item in db_items}
+        unique_providers = self.mass.music.get_unique_providers()
+        for provider_mapping in library_artist.provider_mappings:
+            if provider_mapping.provider_instance not in unique_providers:
+                continue
+            provider_audiobooks = await self.get_provider_author_audiobooks(
```
MarvinSchenkel: Doesn't this call depend on the `artist_type`? Ie `get_provider_author_audiobooks` or `get_provider_narrator_audiobooks`

music_assistant/controllers/media/artists.py
fmunkes: yeah. Rename issue, I should have checked that more closely...

music_assistant/controllers/media/artists.py
fmunkes: You're right.

music_assistant/controllers/media/artists.py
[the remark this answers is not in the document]
fmunkes: fixed above.

music_assistant/controllers/media/artists.py
```diff
@@ -535,3 +598,146 @@ def artist_from_item_mapping(self, item: ItemMapping) -> Artist:
...
+                provider_mapping.item_id, provider_mapping.provider_instance
+            )
```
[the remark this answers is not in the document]
fmunkes: also fixed above.

music_assistant/controllers/media/artists.py
```diff
@@ -223,9 +260,17 @@ async def remove_item_from_library(self, item_id: str | int, recursive: bool = T
...
+        # recursively also remove author/ narrator audiobooks
+        for db_row in await self.mass.music.database.get_rows_from_query(
+            f"SELECT album_id FROM {DB_TABLE_AUDIOBOOK_ARTISTS} WHERE artist_id = :artist_id",
+            {"artist_id": db_id},
+            limit=5000,
+        ):
+            if not recursive:
+                raise MusicAssistantError("Artist still has albums linked")
+            with contextlib.suppress(MediaNotFoundError):
+                await self.mass.music.audiobooks.remove_item_from_library(db_row["album_id"])
```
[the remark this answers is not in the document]
fmunkes: I really should have checked more closely. Also, see above.

music_assistant/models/music_provider.py
```diff
@@ -999,6 +1041,27 @@ async def _sync_library_audiobooks(self) -> set[int]:
...
+                    if len(library_item.narrators) > 0:
+                        lib_narrator = library_item.narrators[0]
+                    if len(prov_item.narrators) > 0:
+                        prov_narrator = prov_item.narrators[0]
+                    if (isinstance(lib_author, Artist) and not isinstance(prov_author, Artist)) or (
+                        isinstance(lib_narrator, Artist) and not isinstance(prov_narrator, Artist)
+                    ):
+                        library_item = await self.mass.music.audiobooks.update_item_in_library(
+                            library_item.item_id, prov_item
+                        )
```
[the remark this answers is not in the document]
fmunkes: fixed

OzGav:
@fmunkes can you have a look at what I am working on here as there is some overlap in ideas I think. https://github.com/music-assistant/models/compare/main...add-classical

fmunkes:
> @fmunkes can you have a look at what I am working on here as there is some overlap in ideas I think. [music-assistant/models@main...add-classical](https://github.com/music-assistant/models/compare/main...add-classical)

Yeah, very similar, and much more extensive on your end :). I think your ArtistRole and the ArtistType do in some way describe the same, though I get that an artist can have multiple roles at the same (conductor, composer, etc). Then my question would be, which way we should go - have the same artist potentially multiple times in the database due to different types (e.g. one row for composer, one row for singer), or remove the ArtistType from the Artist completely and add the role of an artist via your credit approach.

I personally like the first approach a bit more, not because it's mine, but because I think the downside of having some duplicates, is compensated by the easier use of db methods via sqlite, i.e. by just filtering for the ID of the artist (get me all tracks composed by xy would be a db query, and not in python, which should be faster.

What do you think?

OzGav:
Thanks for taking a look. Looking at it more deeply I think these are actually compatible and we don’t need to pick one as they solve different problems.

Your Artist.type is the artist’s primary identity (what they mostly are — narrator, author, singer, composer). That’s what your audiobook PR needs, and it’s a useful field to keep.

Our Credit with ArtistRole is the per-relationship role — what this artist does on this specific track. Classical music (and could be extended to all music types) needs this because the same person can play multiple roles on the same recording (Karajan is both conductor and main artist on a Beethoven track) and different roles across recordings (some performers also compose). We can’t collapse that to a single per-Artist type without losing information.

On the performance point — both approaches are DB-side queries, not Python filtering. Ours filters on track_artists.role, yours filters on artist_id. Both are indexed; neither is slower.

Concretely I’d suggest: keep your Artist.type for primary type / audiobook browse, and if mine gets up then I’ll keep Credit for per-track-relationship role. They sit at different layers and serve different queries.

fmunkes:
Alright, sounds good

music_assistant/controllers/media/artists.py
```diff
@@ -886,3 +939,151 @@ def artist_from_item_mapping(self, item: ItemMapping) -> Artist:
...
         )
+
+    async def audiobooks(
+        self,
+        item_id: str,
+        provider_instance_id_or_domain: str,
+        artist_type: ArtistType = ArtistType.AUTHOR,
+        in_library_only: bool = False,
+    ) -> list[Audiobook]:
+        """Return audiobooks for an artist.
```
OzGav:
```suggestion
        """
        Return audiobooks for an artist.
```

music_assistant/models/music_provider.py
```diff
@@ -187,6 +187,20 @@ async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
         """Get full audiobook details by id."""
         raise NotImplementedError
 
+    async def get_author_audiobooks(self, prov_artist_id: str) -> list[Audiobook]:
+        """Get a list of all audiobooks for the given author.
```
OzGav:
```suggestion
        """
        Get a list of all audiobooks for the given author.
```

music_assistant/models/music_provider.py
```diff
@@ -187,6 +187,20 @@ async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
...
+
+        Only called if provider supports ProviderFeature.AUTHOR_AUDIOBOOKS.
+        """
+        raise NotImplementedError
+
+    async def get_narrator_audiobooks(self, prov_artist_id: str) -> list[Audiobook]:
+        """Get a list of all audiobooks for the given narrator.
```
OzGav:
```suggestion
        """
        Get a list of all audiobooks for the given narrator.
```

OzGav (commented):
I think Marvin's coments are done so they can marked as resolved?

Then is this ready to go as well?

music_assistant/controllers/media/artists.py
fmunkes: Sorry, I merged locally already, so it was easier now to copy your suggestions over.

fmunkes:
> I think Marvin's coments are done so they can marked as resolved?
> 
> Then is this ready to go as well?

I was almost ready to say yes, but just realized, that I forgot to add the authors from the db if they exist. I adjusted the base query for this, and added ItemMapping to the model. https://github.com/music-assistant/models/pull/261

The PR does work, however, the frontend needs some small adjustment, otherwise the we render plain json. Doesn't break anything though. I'll add the frontend label, but if you could review the changes I just made here, that would be appreciated!

Some afterthought - as there is no provider yet supporting the new feature, the frontend will actually be just fine.

marcelveldt:
can you rebase the PR and fix the merge conflicts @fmunkes ?
conceptually I think this is now fine

music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -192,8 +237,75 @@ async def _add_library_item(self, item: Audiobook, overwrite_existing: bool = Fa
...
+    async def _set_audiobook_authors_narrators(
+        self,
+        db_id: int,
+        artists: Iterable[Artist | ItemMapping],
+        overwrite: bool = False,
+    ) -> None:
+        """Write audiobook id and author/ narrator id to DB_TABLE_ALBUM_ARTISTS."""
+        if overwrite:
+            # on overwrite, clear the album_artists table first
+            await self.mass.music.database.delete(
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -221,10 +336,10 @@ async def _update_library_item(
                 ),
                 "publisher": cur_item.publisher or update.publisher,
                 "authors": serialize_to_json(
-                    update.authors if overwrite else cur_item.authors or update.authors
+                    _update_authors if overwrite else cur_item.authors or _update_authors
                 ),
                 "narrators": serialize_to_json(
-                    update.narrators if overwrite else cur_item.narrators or update.narrators
+                    _update_narrators if overwrite else cur_item.narrators or _update_narrators
```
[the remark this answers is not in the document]
fmunkes: fixed by calling the helper. The indicated part must remain a list of string.

fmunkes:
The rebase was more manual then usual, I hope I caught everything. At least seems to have worked in my abs test branch I based on top of this one. The copilot feedback made me add the set_artist_mappings to be used both in add and updating library items.
I have some strange serialization issue in abs test branch, will investigate that first.

Found the reason - external ids in the artists metadata when being part of the Audiobook are not correctly serialized, when part of the event data.

fmunkes:
I found the issue. https://github.com/music-assistant/models/pull/266

music_assistant/controllers/music/media/artists.py
```diff
@@ -57,14 +71,18 @@ def __init__(self, mass: MusicAssistant) -> None:
...
         self.mass.register_api_command(f"music/{api_base}/top_albums", self.top_albums)
+        self.mass.register_api_command(f"music/{api_base}/artist_audiobooks", self.audiobooks)
         self.mass.register_api_command(f"music/{api_base}/similar_artists", self.similar_artists)
 
     async def library_count(
-        self, favorite_only: bool = False, album_artists_only: bool = False
+        self,
+        favorite_only: bool = False,
+        album_artists_only: bool = False,
+        artist_type: ArtistType = ArtistType.SINGER,
```
marcelveldt: should we default to all artist types ?
so have that filter default to None for no explicit artisttype filter

music_assistant/controllers/music/media/artists.py
```diff
@@ -86,6 +106,7 @@ async def library_items(
         provider: str | list[str] | None = None,
         genre: int | list[int] | None = None,
         album_artists_only: bool = False,
+        artist_type: ArtistType = ArtistType.SINGER,
```
marcelveldt: same comment as before - should we default to NO filter ?

marcelveldt:
Marked as draft so you can fix these last changes and then press the "ready for review" button once you are ready. Thanks!

music_assistant/controllers/music/media/artists.py
fmunkes: I changed it. As long as we don't have a provider supporting that, we can still merge it without adapting the frontend. I'm not able to work on the frontend before Monday.

marcelveldt (approved):
Nice job, @fmunkes !

PR #3575, opened 2026-04-04
Handle Ogg FLAC in-band metadata in chained radio streams

a2hill:
## Summary
This adds Ogg-FLAC in-band metadata support to the chained Ogg stream handler used for internet radio streams.  Before this change the handler could extract metadata from Opus/Vorbis comment packets but would not surface metadata carried in Ogg-FLAC streams.

## Notes
This implementation only detects FLAC metadata blocks containing Vorbis comments when they begin at the start of an Ogg page, which is consistent with how the FLAC spec says streams "should" be implemented (see [RFC 9639 Section 10.1](https://www.rfc-editor.org/rfc/rfc9639.html#name-ogg-mapping)). Supporting non-typical streams that pack multiple FLAC metadata packets into a page and place the Vorbis comment block later in the same page payload would require packet-level parsing and reassembly which is outside the scope of this change.

## Testing
This implementation was tested with a local icecast stream. Additionally I've added unit tests that cover most of the original Opus/Vorbis path and all of the Ogg Flac path.

### DIFF
--- a/music_assistant/controllers/streams/ogg_handler.py
+++ b/music_assistant/controllers/streams/ogg_handler.py
@@ -32,6 +32,12 @@ OGG_FLAG_CONTINUATION: int = 0x01  # Continuation of previous page
 OGG_FLAG_BOS: int = 0x02  # Beginning of stream
 OGG_FLAG_EOS: int = 0x04  # End of stream
 
+# Ogg FLAC constants
+FLAC_METADATA_HEADER_SIZE: int = (
+    4  # Size of FLAC metadata block header (1 byte type + 3 bytes length)
+)
+FLAC_METADATA_BLOCK_VORBIS_COMMENT: int = 4  # FLAC metadata block type for Vorbis comments
+
 
 class OggPage:
     """Parsed OGG page with header fields and payload."""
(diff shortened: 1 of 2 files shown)


marcelveldt:
The change looks good to me but chiming in @OzGav here who did the initial implementation of the in band metadata, for the final approval

OzGav:
I tested against `streams.radiomast.io/ref-lossless-ogg-flac-stereo` and metadata extracted fine without these changes. @a2hill can you share or describe the stream you tested with that required the native FLAC block parsing? Also did you test against any other streams to check for any regressions?

I found this useful https://www.radiomast.io/reference-streams and I also tested against tested with [http://stream.tilos.hu:80/tilos.opus](http://stream.tilos.hu/tilos.opus) and http://srv.tsmod.net:8008/dnb

a2hill:
The stream I’m testing with is a local Icecast mount built with Liquidsoap from my record player line-in. Liquidsoap encodes the audio as `Ogg/FLAC` and publishes it to Icecast. Since the source audio has no metadata I use a music recognition service to identify the audio and then inject that metadata back into the stream as Vorbis comments via Liquidsoap. The relevant Liquidsoap setup is here:
https://github.com/a2hill/line-in-streamer/blob/main/liquidsoap/line-in.liq#L121

After you pointed it out, I looked at https://streams.radiomast.io/ref-lossless-ogg-flac-stereo because Music Assistant does display metadata for it without my change. When probing it with ffmpeg I'm not seeing any Vorbis comments in the `Ogg/FLAC` stream itself. For example the following probe command returns empty stream tags:
```
ffprobe -v error \
  -show_entries format_tags:stream_tags \
  -of json \
  https://streams.radiomast.io/ref-lossless-ogg-flac-stereo
```
What I found is that the Radiomast lossless stream is exposing metadata out-of-band via ICY metadata. When I requested the stream with `Icy-MetaData: 1`, the response included `icy-metaint`, and the interleaved ICY metadata block contained a changing `StreamTitle`.

For my stream I intentionally did not use ICY metadata. Liquidsoap’s docs describe ICY metadata as primarily intended for formats that do not support in-band metadata, and Ogg/FLAC does support in-band metadata. So I do not set the ICY metadata flag on my stream and inject the metadata directly into the Ogg/FLAC stream as Vorbis comments instead. That means Music Assistant falls back to the in-band metadata path here:
https://github.com/music-assistant/server/blob/d8db5aa82c19c79c6e515d2d2dd93ef07b2aa862/music_assistant/controllers/streams/audio.py#L530

For regression testing I validated the following:

https://stream.tilos.hu/tilos.opus uses in-band metadata, and Music Assistant displays the same metadata that ffprobe sees.
http://srv.tsmod.net:8008/dnb uses in-band metadata, and Music Assistant displays the same metadata that ffprobe sees.
https://streams.radiomast.io/ref-64k-heaacv2-stereo does not have in-band metadata, uses ICY metadata, and it displays correctly in Music Assistant.

OzGav:
Sounds good to me then

OzGav (approved):

PR #3576, opened 2026-04-04
Add duplicate track prevention and empty default for party name/QR text

apophisnow:
- Add "Prevent Duplicate Tracks" config option (enabled by default) that rejects guest requests for tracks already in the queue
- Change party name and QR text defaults to empty string so the fields are blank by default


marcelveldt (approved):
Thanks, @apophisnow !

PR #3582, opened 2026-04-05
YT Music: Fix syncing 'Episodes for later' in podcast library sync

teancom:
Found this while trying to figure out a different issue.

ytmusicapi.get_library_podcasts() returns special auto-playlists like "Episodes for Later" (podcastId: "SE") alongside real podcasts. These aren't actual podcasts, so fetching episodes via ytm.get_podcast() returns HTTP 400. Filter them using the existing YT_PERSONAL_PLAYLISTS tuple, matching how playlist sync already handles them.

### DIFF
--- a/music_assistant/providers/ytmusic/__init__.py
+++ b/music_assistant/providers/ytmusic/__init__.py
@@ -323,6 +323,8 @@ class YoutubeMusicProvider(MusicProvider):
             headers=self._headers, language=self.language, user=self._yt_user
         )
         for podcast in podcasts_obj:
+            if podcast.get("podcastId") in YT_PERSONAL_PLAYLISTS:
+                continue
             yield self._parse_podcast(podcast)
 
     @use_cache(3600 * 24 * 30)  # Cache for 30 days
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Not sure if I agree with this logic. I guess we just have to treat this like we do with Liked Music (LM) in playlists. Now we are blindly filtering 'Episodes for Later' which is definitely something nice to have. Options I see:
- Treat `SE` as a special podcast id just like we do for `LM`
- Filter `SE` from podcasts, but expose them as a RecommendationFolder.

I think I prefer the former as it is a nice way for people to jump in

teancom:
That makes sense. What about this approach? Also, I was going to add some tests for this and saw there weren't any. I don't want to add them to this PR but would you like a different PR to add some test structure and basic tests for this provider? I might be able to do that in the near-term.

MarvinSchenkel:
I just had a play with it and it turns out that `ytmusicapi` cannot parse those 'Personal podcast / playlists' as podcasts, neither as playlist. So let's just ignore those for now as we cannot parse them. 

This will make sure the podcast sync task succeeds again and people can listen to podcasts from YT Music

MarvinSchenkel (approved):
Thanks @teancom 👏

PR #3595, opened 2026-04-06
Include missing description in automatic artist metadata scan

OzGav:
The scheduled metadata scan task for artists that have never had a metadata lookup only checked for missing artwork (images). This PR also includes missing description in the scan criteria, so artists without a biography are also picked up during the initial metadata lookup.

### DIFF
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -133,7 +133,7 @@ LOCALES = {
 DEFAULT_LANGUAGE = "en_US"
 REFRESH_INTERVAL = 60 * 60 * 24 * 90  # 90 days
 CONF_ENABLE_ONLINE_METADATA = "enable_online_metadata"
-MISSING_ARTIST_ARTWORK_SCAN_TASK_ID = "metadata_missing_artist_artwork_scan"
+MISSING_ARTIST_METADATA_SCAN_TASK_ID = "metadata_missing_artist_metadata_scan"
 PLAYLIST_METADATA_SCAN_TASK_ID = "metadata_playlist_metadata_scan"
 THUMB_CACHE_CLEANUP_TASK_ID = "metadata_thumb_cache_cleanup"
 METADATA_LOOKUP_TASK_ID_PREFIX = "metadata_lookup"
@@ -1004,12 +1004,12 @@ class MetaDataController(CoreController):
         utc_hour, utc_minute = local_clock_time_to_utc(4, 0)
         desired_schedule = TaskSchedule.daily(hour=utc_hour, minute=utc_minute)
         self.mass.tasks.register_scheduled_task(
-            task_id=MISSING_ARTIST_ARTWORK_SCAN_TASK_ID,
-            name="Scan missing artist artwork",
-            handler=self._scan_missing_artist_artwork,
+            task_id=MISSING_ARTIST_METADATA_SCAN_TASK_ID,
+            name="Scan missing artist metadata",
+            handler=self._scan_missing_artist_metadata,
             schedule=desired_schedule,
-            translation_key="background_task.scan_missing_artist_artwork",
-            metadata={"task_domain": "metadata_missing_artist_artwork_scan"},
+            translation_key="background_task.scan_missing_artist_metadata",
+            metadata={"task_domain": "metadata_missing_artist_metadata_scan"},
             allow_retry=True,
         )
         self.mass.tasks.register_scheduled_task(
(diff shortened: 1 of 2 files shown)


marcelveldt:
Lets not do this please until we have a metadata proxy

music_assistant/controllers/metadata.py
```diff
@@ -1036,23 +1036,31 @@ def _get_metadata_lookup_task_id(uri: str) -> str:
...
+        # find artists missing images or description that haven't been refreshed recently
+        missing_images = (
             f"(json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') ISNULL "
-            f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') = '[]') "
-            f"AND (json_extract({DB_TABLE_ARTISTS}.metadata,'$.last_refresh') ISNULL "
+            f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') = '[]')"
+        )
+        missing_description = (
+            f"(json_extract({DB_TABLE_ARTISTS}.metadata,'$.description') ISNULL "
+            f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.description') = '')"
```
marcelveldt: let's only test the NULL value where it has never been retrieved at all

music_assistant/controllers/metadata.py
```diff
@@ -1036,23 +1036,31 @@ def _get_metadata_lookup_task_id(uri: str) -> str:
...
+        )
+        stale_refresh = (
+            f"(json_extract({DB_TABLE_ARTISTS}.metadata,'$.last_refresh') ISNULL "
             f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.last_refresh') < {refresh_before})"
         )
+        query = f"({missing_images} OR {missing_description}) AND {stale_refresh}"
```
marcelveldt: OK, this is good. We ONLY do this when images or description are completely missing AND the item has either never been looked up or is stale. When reading the PR on my phone earlier today I though we were going to fetch metadata for any artist without description, which would be bad for api calls. This is better.

That said, we can even remove the whole last refresh call as that is already handled on access.
This background task is to make sure that we fetch at least the images for the artists listing

music_assistant/controllers/metadata.py
OzGav: Ok no empty string check

music_assistant/controllers/metadata.py
OzGav: Removed the stale_refresh / last_refresh SQL filter

music_assistant/controllers/metadata.py
```diff
@@ -1036,23 +1036,22 @@ def _get_metadata_lookup_task_id(uri: str) -> str:
...
-        update_current_task_progress_text("Searching for artists with missing artwork")
-        refresh_before = int(time() - REFRESH_INTERVAL)
-        query = (
+    async def _scan_missing_artist_metadata(self) -> None:
+        """Refresh metadata for a small batch of artists missing artwork or description."""
+        update_current_task_progress_text("Searching for artists with missing metadata")
+        missing_images = (
             f"(json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') ISNULL "
-            f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') = '[]') "
-            f"AND (json_extract({DB_TABLE_ARTISTS}.metadata,'$.last_refresh') ISNULL "
```
marcelveldt, line 1046: this still needs to be there;

we-re looking for artists that never had a metadata lookup
and do not have images or description

music_assistant/controllers/metadata.py
```diff
@@ -1036,23 +1036,22 @@ def _get_metadata_lookup_task_id(uri: str) -> str:
         """Return deterministic task id for a metadata lookup."""
         return f"{METADATA_LOOKUP_TASK_ID_PREFIX}_{uuid5(NAMESPACE_URL, uri).hex}"
 
-    async def _scan_missing_artist_artwork(self) -> None:
-        """Refresh metadata for a small batch of artists missing artwork."""
-        update_current_task_progress_text("Searching for artists with missing artwork")
-        refresh_before = int(time() - REFRESH_INTERVAL)
-        query = (
+    async def _scan_missing_artist_metadata(self) -> None:
+        """Refresh metadata for a small batch of artists missing artwork or description."""
```
marcelveldt: The docstring should be more like "Scan for artists with missing metadata"

music_assistant/controllers/metadata.py
OzGav: Added back: now checks `last_refresh ISNULL` to only target artists that never had a metadata lookup. The stale/interval part stays removed since that's handled in `_update_artist_metadata`.

---
_Generated by [Claude Code](https://claude.ai/code)_

music_assistant/controllers/metadata.py
OzGav: Updated to "Scan for artists with missing metadata."

---
_Generated by [Claude Code](https://claude.ai/code)_

OzGav: Done

music_assistant/controllers/metadata.py
OzGav: Claude is getting new features and responding for me! I have turned that off!

Additionally, I have removed the last_refresh check entirely as you suggested. Keeping `last_refresh ISNULL` here would reintroduce the original bug — if the initial metadata lookup ran and set `last_refresh` but failed to retrieve the description for whatever reason, the scan would never retry that artist. The `REFRESH_INTERVAL `guard in `_update_artist_metadata` already prevents excessive API calls, so I think this task can safely just find artists with missing images or description without any refresh timestamp filtering.

marcelveldt: Not really - this was actually a bug in the previous implementation and a good explanaition for many api calls
we can really only do the background scan for items that never had a metadata lookup.
and then to narrow it down even further we look for artists that have no images or descriptions and skip the ones that do.

metadata is checked on ACCESS of an item. That is when the last refresh is evaluated and metadata refreshed if needed. If people already having artists that went through the scan and have the info missing, they either need to re-setup their db, wait for 90 days or just run the metadata lookup manually.

It is what is is as long as we have the issues with too many api calls

OzGav: Makes sense — the background scan is for the listing view (images), and bios get refreshed on access when you actually open the artist page. Restored the last_refresh ISNULL check.

marcelveldt (approved):
Thanks, @OzGav !

PPPDUD:
Did we _really_ need to use Claude for this PR? Why couldn't you have paid a human $5 to do the same thing?

OzGav:
@PPPDUD What is the point of your post. I was away from home on my iPad and spent some of my free time to fix this.

PPPDUD:
> @PPPDUD What is the point of your post. I was away from home on my iPad and spent some of my free time to fix this.

As you are (as far as I can tell) a human developer, I'm sure that you can understand that the usage of LLMs to write code threatens projects like Music Assistant in several different ways:
- whenever major LLM services go down (and this happens _a lot_), development quickly grinds to a halt
- it takes away opportunities from new developers who want to learn their way around a codebase
- it contributes to an ecosystem where the outputs of zero-accountability machines are treated equally to those of real humans with real thoughts and effort
- etc.

I would also argue that there's an intrinsic value in 100%-manmade code, but that's a matter of philosophy and not practical conversation. For me, the inclusion of LLM-written code in a project that I use is equivalent to one of my favorite foods having Red 40 in it.

Using LLMs to debug code is fine with me, but I don't think that it's healthy to have them write the final code or anything else that would warrant listing Claude as a contributor.

PPPDUD:
See also the discussion at https://github.com/orgs/music-assistant/discussions/5338.

OzGav:
OK this is not the place for you to grind your axe. I see you other post and will continue there.

PPPDUD:
> OK this is not the place for you to grind your axe. I see you other post and will continue there.

Understood; my apologies.

PR #3602, opened 2026-04-07
Add manual item management API and on_item_updated provider callback

OzGav:
Satisifies https://github.com/orgs/music-assistant/discussions/3279

FRONTEND: https://github.com/music-assistant/frontend/pull/1685

This adds API endpoints for manually adding radio stations and tracks (builtin/add_radio, builtin/add_track), plus the ability to edit them via the standard music/*/update endpoints.

The backend now has a generic on_item_updated callback that providers can implement to sync library edits back to their own storage. The builtin provider implements this to update:
  - Radio/track name and image in config storage
  - Playlist name in M3U files
  - Playlist images in config storage (new playlist_images key)

URL changes aren't supported through this path (the URL is the identifier), so changing a stream URL requires deleting and re-adding the item.

### DIFF
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -203,6 +203,11 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
             library_item.uri,
             library_item,
         )
+        # notify providers of the update so they can sync their own storage
+        for prov_mapping in library_item.provider_mappings:
+            if provider := self.mass.get_provider(prov_mapping.provider_instance):
+                provider = cast("MusicProvider", provider)
+                await provider.on_item_updated(library_item)
         return library_item
 
     async def remove_item_from_library(self, item_id: str | int, recursive: bool = True) -> None:
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -205,7 +205,7 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
         )
         # notify providers of the update so they can sync their own storage
         for prov_mapping in library_item.provider_mappings:
-            if provider := self.mass.get_provider(prov_mapping.provider_instance):
+            if provider := self.mass.music.get_provider(prov_mapping.provider_instance):
                 provider = cast("MusicProvider", provider)
                 await provider.on_item_updated(library_item)
         return library_item
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -205,7 +205,7 @@ class MediaControllerBase[ItemCls: "MediaItemType"](metaclass=ABCMeta):
         )
         # notify providers of the update so they can sync their own storage
         for prov_mapping in library_item.provider_mappings:
-            if provider := self.mass.music.get_provider(prov_mapping.provider_instance):
+            if provider := self.mass.get_provider(prov_mapping.provider_instance):
                 provider = cast("MusicProvider", provider)
                 await provider.on_item_updated(library_item)
         return library_item
--- a/music_assistant/controllers/media/radio.py
+++ b/music_assistant/controllers/media/radio.py
@@ -5,8 +5,8 @@ from __future__ import annotations
 import asyncio
 from typing import TYPE_CHECKING, cast
 
-from music_assistant_models.enums import MediaType, ProviderFeature
-from music_assistant_models.errors import ProviderUnavailableError
+from music_assistant_models.enums import EventType, MediaType, ProviderFeature
+from music_assistant_models.errors import InvalidDataError, ProviderUnavailableError
 from music_assistant_models.media_items import ProviderMapping, Radio, Track
 
 from music_assistant.constants import DB_TABLE_RADIOS
@@ -43,6 +43,7 @@ class RadioController(MediaControllerBase[Radio]):
         self.mass.register_api_command(f"music/{api_base}/radio_versions", self.versions)
         self.mass.register_api_command(f"music/{api_base}/export_radios", self.export_radios)
         self.mass.register_api_command(f"music/{api_base}/import_radios", self.import_radios)
+        self.mass.register_api_command(f"music/{api_base}/edit_radio", self.edit_radio)
 
     async def export_radios(self) -> str:
         """Export all library radio stations to M3U8 format."""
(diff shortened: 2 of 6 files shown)


music_assistant/controllers/media/radio.py
```diff
@@ -61,6 +62,54 @@ async def import_radios(self, m3u_data: str) -> int:
         builtin_prov = cast("BuiltinProvider", provider)
         return await builtin_prov.import_radios(m3u_data)
 
+    async def edit_radio(
```
marcelveldt: why aren't you using the existing methods for this ?
why does the radio's need special treatment ?
the plumbing for this already exists in update_item, it just needs UI

OzGav: The reason is that I learnt that manually-added radio stations are stored in two places:

  1. Config storage
  2. The database

The existing `update_item_in_library` only updates the database. For builtin radios, I believe we also need to update the config storage, otherwise:
  - The next sync would overwrite user changes with the old config values
  - URL changes wouldn't persist (the URL is the item_id in config)

Additionally, this PR supports URL changes which require updating the provider mapping's item_id, which the standard update flow doesn't handle.

If you'd prefer, we could extend the existing `_update_library_item` in `RadioController `to detect builtin provider mappings and update config there instead? What do you think?

marcelveldt: OK, got it.
Since the builtin provider is the only one that allows manually adding items (and now editing them) I think it makes sense that we add some dedicated handlers to the builtin provider itself and register the api commands from in there.

That way we can also undo the "add to library" workaround that is currently in place and call the handler on the builtin provider directly, which will take care of adding the item (or editing) in its internal logic and then will issue a focused sync if needed.

these methods should either be "add_item" and "edit_item" which take an optional media type or we create dedicated handlers for track and track (because track can currently also be manually registered)

OzGav: As discussed. Callback added instead.

music_assistant/controllers/media/base.py
```diff
@@ -203,6 +203,11 @@ async def update_item_in_library(
             library_item.uri,
             library_item,
         )
+        # notify providers of the update so they can sync their own storage
+        for prov_mapping in library_item.provider_mappings:
+            if provider := self.mass.get_provider(prov_mapping.provider_instance):
```
marcelveldt, line 208:
```suggestion
            if provider := self.mass.music.get_provider(prov_mapping.provider_instance):
```
this will do the cast for you

music_assistant/models/music_provider.py
```diff
@@ -474,6 +474,16 @@ async def on_played(
         is_playing is True when the track is currently playing.
         """
 
+    async def on_item_updated(self, item: MediaItemType) -> None:
+        """
+        Handle callback when a library item's metadata has been updated.
+
+        Providers can override this to sync changes to their own storage
```
marcelveldt:
```suggestion
        Providers can implement this to sync changes to their own storage
```

music_assistant/providers/builtin/__init__.py
```diff
@@ -176,6 +176,8 @@ async def loaded_in_mass(self) -> None:
             schedule=TaskSchedule.hourly(every=24),
             initial_delay=60,
         )
+        # register API commands for manual item management
+        self.mass.register_api_command("builtin/add_radio", self.add_radio)
```
marcelveldt, line 181: what about the add track ?

music_assistant/providers/builtin/__init__.py
```diff
@@ -400,6 +402,62 @@ async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool
         self.mass.config.set(key, stored_items)
         return True
 
+    async def on_item_updated(self, item: MediaItemType) -> None:
+        """Update stored item config when a library item is edited.
```
marcelveldt:
```suggestion
        """
        Update stored item config when a library item is edited.
```

music_assistant/providers/builtin/__init__.py
```diff
@@ -400,6 +402,62 @@ async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool
...
+            None,
+        )
+        if not builtin_mapping:
+            return
+
+        stored_items: list[StoredItem] = self.mass.config.get(key, [])
+        for stored_item in stored_items:
+            if stored_item["item_id"] == builtin_mapping.item_id:
+                stored_item["name"] = item.name
+                if item.image:
```
marcelveldt, line 451: future improvement: also allow updating description and other image types

music_assistant/controllers/media/base.py
OzGav: It seems `self.mass.music.get_provider()` doesn't exist? - The MusicController has a providers property that returns a list, but no `get_provider()` method. The existing pattern throughout the codebase (albums.py, artists.py, etc.) uses `self.mass.get_provider()` with a cast to MusicProvider, so I've kept it for now unless I have missed something?

music_assistant/providers/builtin/__init__.py
OzGav: Yeah of course I was focussed solely on the radio station feature. Track functionality added

music_assistant/providers/builtin/__init__.py
OzGav: TODO added

music_assistant/providers/builtin/__init__.py
```diff
@@ -400,6 +403,89 @@ async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool
...
+        Update stored item config when a library item is edited.
+
+        :param item: The updated media item with new metadata.
+        """
+        if item.media_type == MediaType.RADIO:
+            key = CONF_KEY_RADIOS
+        elif item.media_type == MediaType.TRACK:
+            key = CONF_KEY_TRACKS
+        else:
+            return
```
marcelveldt: what about playlists ?

OzGav: Yes didn't think about that either. Also added playlist support - name updates go to the M3U file and images are stored in a new playlist_images config key.

marcelveldt: I think we can also add support for a generic playlist image to the m3u itself - I'm actually pretty sure of that!

marcelveldt: In any case lets not introduce a new key in config for that - we just got rid of that for playlists to make them exchangeable. maybe do playlist image support in a follow-up PR or adjust this one

marcelveldt:
marked as enhencement instead of feature - I propose to mark the frontend PR as feature

music_assistant/providers/builtin/__init__.py
```diff
@@ -400,6 +419,117 @@ async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool
...
+        # update name in M3U file
+        m3u_data = await self._read_m3u_file(playlist_id)
+        if m3u_data:
+            existing_items = parse_m3u(m3u_data)
+            try:
+                await self._write_m3u_file(playlist_id, new_name, list(existing_items))
+            except OSError as err:
+                self.logger.warning("Failed to update playlist name: %s", err)
+        # update image in config
+        playlist_images: dict[str, str] = self.mass.config.get(CONF_KEY_PLAYLIST_IMAGES, {})
```
marcelveldt: let's not do this

music_assistant/providers/builtin/__init__.py
OzGav: Removed the config key approach entirely. Switched to storing playlist images in the M3U file using` #EXTIMG:url` directly after `#EXTM3U`. There is no official standard for image URLs in M3U files but I noted we support this tag on a track level basis so adding another at the top for the playlist seems appropriate.

 Also fixed the frontend to check `is_editable` before showing the edit option for playlists.

music_assistant/providers/builtin/__init__.py
OzGav: As above

music_assistant/controllers/media/base.py
marcelveldt: Ok I must have dreamed that we added that then :-)
Makes sense to add some day to the controllers...

music_assistant/helpers/playlists.py
```diff
@@ -233,6 +233,27 @@ def parse_m3u_playlist_name(m3u_data: str) -> str | None:
     return None
 
 
+def parse_m3u_playlist_image(m3u_data: str) -> str | None:
+    """Extract the playlist cover image from an M3U #EXTIMG directive.
```
marcelveldt:
```suggestion
    """
    Extract the playlist cover image from an M3U #EXTIMG directive.
```

marcelveldt (approved):
Thanks, @OzGav !

PR #3604, opened 2026-04-07
Subsonic: Fix structured lyrics yet again

khers:
Gonic handles this differently from Navidrome and fixing this requires another bump to the py-opensonic library.


MarvinSchenkel:
Should this be backported in the next patch release?

khers:
If lyrics were part of that release then yes

MarvinSchenkel (approved):

staticdev:
@khers @MarvinSchenkel I am a bit confused, in the official docs of [Subsonic](https://www.music-assistant.io/music-providers/subsonic/) provider says that lyrics are not supported but we have this PR here. Are maybe the docs outdated? Or something else?

PR #3607, opened 2026-04-07
Add AI_QUERY and TTS to HA Plugin

MarvinSchenkel:
This PR will allow plugins to use the HA plugin to:
- Query LLMs
- Generate TTS

Plugins can call these features by:
```
plugins = mass.get_plugins_by_feature(ProviderFeature.AI_QUERY)
plugins[0].ai_query("What are the top 3 jazz albums of all time?")
```

### DIFF
--- a/music_assistant/models/plugin.py
+++ b/music_assistant/models/plugin.py
@@ -175,7 +175,7 @@ class PluginProvider(Provider):
         yield b""
         raise NotImplementedError
 
-    async def tts(self, message: str, language: str | None = None) -> str:
+    async def get_tts_message_url(self, message: str, language: str | None = None) -> str:
         """
         Convert text to speech audio.
 
--- a/music_assistant/models/plugin.py
+++ b/music_assistant/models/plugin.py
@@ -175,6 +175,29 @@ class PluginProvider(Provider):
         yield b""
         raise NotImplementedError
 
+    async def tts(self, message: str, language: str | None = None) -> str:
+        """
+        Convert text to speech audio.
+
+        Will only be called if ProviderFeature.TTS is declared.
+
+        :param message: The text to convert to speech.
+        :param language: Optional language code.
+        :return: URL to the generated audio.
+        """
+        raise NotImplementedError
+
+    async def ai_query(self, query: str) -> str:
+        """
+        Handle an AI query.
+
+        Will only be called if ProviderFeature.AI_QUERY is declared.
+
+        :param query: The query/prompt to send.
+        :return: The AI response as a string.
+        """
+        raise NotImplementedError
+
     async def resolve_image(self, path: str) -> str | bytes:
         """
         Resolve an image from an image path.
--- a/music_assistant/models/plugin.py
+++ b/music_assistant/models/plugin.py
@@ -15,7 +15,7 @@ from music_assistant.models.player import PlayerSource
 from .provider import Provider
 
 if TYPE_CHECKING:
-    from music_assistant_models.streamdetails import StreamMetadata
+    from music_assistant_models.streamdetails import StreamDetails, StreamMetadata
 
 
 @dataclass
@@ -175,7 +175,7 @@ class PluginProvider(Provider):
         yield b""
         raise NotImplementedError
 
-    async def get_tts_message_url(self, message: str, language: str | None = None) -> str:
+    async def get_tts_message(self, message: str, language: str | None = None) -> StreamDetails:
         """
         Convert text to speech audio.
 
@@ -183,7 +183,7 @@ class PluginProvider(Provider):
 
         :param message: The text to convert to speech.
         :param language: Optional language code.
-        :return: URL to the generated audio.
+        :return: StreamDetails for the generated audio.
         """
         raise NotImplementedError
 
--- a/music_assistant/providers/hass/__init__.py
+++ b/music_assistant/providers/hass/__init__.py
@@ -26,9 +26,22 @@ from hass_client.utils import (
     get_websocket_url,
 )
 from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption, ConfigValueType
-from music_assistant_models.enums import ConfigEntryType, ProviderFeature
-from music_assistant_models.errors import LoginFailed, MusicAssistantError, SetupFailedError
+from music_assistant_models.enums import (
+    ConfigEntryType,
+    ContentType,
+    MediaType,
+    ProviderFeature,
+    StreamType,
+)
+from music_assistant_models.errors import (
+    LoginFailed,
+    MusicAssistantError,
+    SetupFailedError,
+    UnsupportedFeaturedException,
+)
+from music_assistant_models.media_items.audio_format import AudioFormat
 from music_assistant_models.player_control import PlayerControl
+from music_assistant_models.streamdetails import StreamDetails
 
 from music_assistant.constants import MASS_LOGO_ONLINE, VERBOSE_LOG_LEVEL
 from music_assistant.helpers.auth import AuthenticationHelper
@@ -190,7 +203,7 @@ async def get_config_entries(
         hass_prov = cast("HomeAssistantProvider", hass_prov)
         return (
             *base_entries,
-            *(await _get_ha_config_entries(hass_prov.hass)),
+            *(await _get_config_entries(hass_prov.hass)),
         )
 
     return (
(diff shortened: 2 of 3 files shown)


music_assistant/providers/hass/__init__.py
```diff
@@ -293,7 +321,32 @@ async def _get_player_control_config_entries(hass: HomeAssistantClient) -> tuple
...
+            ConfigEntry(
+                key=CONF_AI_TASK_ENTITY,
+                type=ConfigEntryType.STRING,
+                label="AI Task entity",
+                required=False,
+                options=ai_task_entities,
+                description="Select which Home Assistant AI Task entity to use for AI queries.",
+                category="Features",
+            )
+        )
```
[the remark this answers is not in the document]
MarvinSchenkel: Fallback path will show these options

music_assistant/providers/hass/__init__.py
```diff
@@ -585,18 +638,52 @@ async def get_user_details(self, ha_user_id: str) -> tuple[str | None, str | Non
...
+        result = await self.hass.send_command(
+            "call_service",
+            domain="ai_task",
+            service="generate_data",
+            service_data={
+                "task_name": "music_assistant",
+                "instructions": query,
+                "entity_id": str(entity_id),
+            },
+            return_response=True,
```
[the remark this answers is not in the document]
MarvinSchenkel: ai_query is only advertised when an ai_task entity is configured.

music_assistant/providers/hass/__init__.py
```diff
@@ -585,18 +638,52 @@ async def get_user_details(self, ha_user_id: str) -> tuple[str | None, str | Non
...
+    async def tts(self, message: str, language: str | None = None) -> str:
+        """Handle text-to-speech via Home Assistant's REST API."""
+        entity_id = self.config.get_value(CONF_TTS_ENTITY)
+        ha_url, headers, http_session = self._get_ha_http()
+        payload: dict[str, str] = {"engine_id": str(entity_id), "message": message}
+        if language:
+            payload["language"] = language
+        async with http_session.post(
+            f"{ha_url}/api/tts_get_url", headers=headers, json=payload
+        ) as response:
```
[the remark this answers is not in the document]
MarvinSchenkel: tts is only advertised when the entity is configured

music_assistant/models/plugin.py
```diff
@@ -175,6 +175,29 @@ async def get_audio_stream(self, player_id: str) -> AsyncGenerator[bytes, None]:
         yield b""
         raise NotImplementedError
 
+    async def tts(self, message: str, language: str | None = None) -> str:
```
marcelveldt:
```suggestion
    async def get_tts_message_url(self, message: str, language: str | None = None) -> str:
```

marcelveldt: or something along those lines to make the intention more clear

music_assistant/models/plugin.py
```diff
@@ -175,6 +175,29 @@ async def get_audio_stream(self, player_id: str) -> AsyncGenerator[bytes, None]:
...
+
+        Will only be called if ProviderFeature.TTS is declared.
+
+        :param message: The text to convert to speech.
+        :param language: Optional language code.
+        :return: URL to the generated audio.
+        """
+        raise NotImplementedError
+
+    async def ai_query(self, query: str) -> str:
```
marcelveldt, line 190: what about response type ? Is it always a string ? Not structured data such as json ?

music_assistant/providers/hass/__init__.py
```diff
@@ -293,7 +321,32 @@ async def _get_player_control_config_entries(hass: HomeAssistantClient) -> tuple
             "like to import as player Mute controls in Music Assistant.",
             category="player_controls",
         ),
-    )
+    ]
+    if tts_entities:
```
marcelveldt: why only append the configentry if there are existing entities ? why not always show it so users have an idea that something is missing

music_assistant/models/plugin.py
MarvinSchenkel: Just did some experimenting. It looks like HA is always returning a string. When I try to force it to return JSON, it returns a JSON serialized string. So I think `-> str` is fine here. Callers will need to deserialize JSON results themselves.

image

music_assistant/providers/hass/__init__.py
```diff
@@ -209,15 +216,28 @@ async def get_config_entries(
             label=CONF_MUTE_CONTROLS,
             default_value=[],
         ),
+        ConfigEntry(
+            key=CONF_TTS_ENTITY,
+            type=ConfigEntryType.STRING,
+            label="Text-to-Speech entity",
```
marcelveldt: please make sure that the frontend at least a better label and description for these options as they are very unclear this way

music_assistant/providers/hass/__init__.py
```diff
@@ -209,15 +216,28 @@ async def get_config_entries(
...
+            type=ConfigEntryType.STRING,
+            label="AI Task entity",
+            required=False,
+        ),
     )
 
 
-async def _get_player_control_config_entries(hass: HomeAssistantClient) -> tuple[ConfigEntry, ...]:
-    """Return all HA state objects for (valid) media_player entities."""
+async def _get_ha_config_entries(hass: HomeAssistantClient) -> tuple[ConfigEntry, ...]:
```
marcelveldt:
```suggestion
async def _get_config_entries(hass: HomeAssistantClient) -> tuple[ConfigEntry, ...]:
```

music_assistant/providers/hass/__init__.py
```diff
@@ -209,15 +216,28 @@ async def get_config_entries(
...
+    """Return config entries for player controls, TTS and AI task entity selection."""
```
marcelveldt:
```suggestion
    """Return the (entity based) config entries."""
```

music_assistant/providers/hass/__init__.py
```diff
@@ -293,7 +321,30 @@ async def _get_player_control_config_entries(hass: HomeAssistantClient) -> tuple
...
+    ]
+    entries.append(
+        ConfigEntry(
+            key=CONF_TTS_ENTITY,
+            type=ConfigEntryType.STRING,
+            label="Text-to-Speech entity",
+            required=False,
+            options=tts_entities,
+            description="Select which Home Assistant TTS entity to use for text-to-speech.",
+            category="Features",
```
marcelveldt, line 345: should we add a default ?
default = tts_entities[0].value if tts_entities else None

music_assistant/providers/hass/__init__.py
```diff
@@ -293,7 +321,30 @@ async def _get_player_control_config_entries(hass: HomeAssistantClient) -> tuple
...
     )
+    entries.append(
+        ConfigEntry(
+            key=CONF_AI_TASK_ENTITY,
+            type=ConfigEntryType.STRING,
+            label="AI Task entity",
+            required=False,
+            options=ai_task_entities,
+            description="Select which Home Assistant AI Task entity to use for AI queries.",
+            category="Features",
```
marcelveldt, line 355: should we add a default ?

music_assistant/providers/hass/__init__.py
```diff
@@ -585,18 +636,52 @@ async def get_user_details(self, ha_user_id: str) -> tuple[str | None, str | Non
...
+        response = result.get("response", {}) if isinstance(result, dict) else {}
+        data = response.get("data") if isinstance(response, dict) else None
+        if not data:
+            msg = f"AI Task returned no data in response: {result}"
+            raise MusicAssistantError(msg)
+        return str(data)
+
+    async def get_tts_message_url(self, message: str, language: str | None = None) -> str:
+        """Handle text-to-speech via Home Assistant's REST API."""
+        entity_id = self.config.get_value(CONF_TTS_ENTITY)
```
marcelveldt: Add a guard here

```suggestion
        if (entity_id := self.config.get_value(CONF_TTS_ENTITY)) is None:
            raise UnsupportedFeature("TTS entity is not configured")
```

music_assistant/models/plugin.py
```diff
@@ -175,6 +175,29 @@ async def get_audio_stream(self, player_id: str) -> AsyncGenerator[bytes, None]:
         yield b""
         raise NotImplementedError
 
+    async def get_tts_message_url(self, message: str, language: str | None = None) -> str:
```
marcelveldt: So, what we should consider is letting this return a StreamDetails object because then it is flexible to be used for streaming and you can support more than just an URL. You can also support a local file, a custom stream etc

marcelveldt (approved):
LGTM, Thanks @MarvinSchenkel !

PR #3613, opened 2026-04-08
Fix local audio seg fault and hide unavailable players

OzGav:
Encountered a seg fault crashing MA when transferring queues between local audio players, Also noticed many players discovered didn't actually work. Fixes therefore are:

  1. Segfault prevention - graceful shutdown of the audio writer task
  2. Device filtering - unavailable devices no longer appear in the UI

### DIFF
--- a/music_assistant/providers/local_audio/sendspin_bridge.py
+++ b/music_assistant/providers/local_audio/sendspin_bridge.py
@@ -394,6 +394,7 @@ class LocalAudioBridgeManager:
                 continue
             # Test if the device can actually be opened - many ALSA virtual
             # devices will fail if PipeWire or another audio server is active
+            # TODO: query supported sample rates per device and default to 48kHz
             try:
                 test_stream = sd.RawOutputStream(
                     device=idx,
--- a/music_assistant/providers/local_audio/sendspin_bridge.py
+++ b/music_assistant/providers/local_audio/sendspin_bridge.py
@@ -266,19 +266,29 @@ class SendspinLocalAudioBridge:
     async def _stop_streaming(self) -> None:
         """Stop streaming (internal, called with lock held)."""
         self._is_streaming = False
-        if self._writer_task:
-            self._writer_task.cancel()
-            with suppress(asyncio.CancelledError, Exception):
-                await self._writer_task
+        if self._writer_task and not self._writer_task.done():
+            # Signal the writer to stop gracefully by sending None.
+            # Don't cancel the task directly as it may be blocked in
+            # run_in_executor writing to the audio device - cancelling
+            # while a write is in progress can cause a segfault.
+            while not self._write_queue.empty():
+                self._write_queue.get_nowait()
+            self._write_queue.put_nowait(None)
+            try:
+                await asyncio.wait_for(self._writer_task, timeout=2.0)
+            except TimeoutError:
+                # Writer didn't stop in time - it may be stuck in a blocking write.
+                # Cancel it as a last resort but don't close the stream until it's done.
+                self._writer_task.cancel()
+                with suppress(asyncio.CancelledError, Exception):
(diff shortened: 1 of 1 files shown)


music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -367,17 +377,36 @@ async def discover_and_register(self) -> None:
 
     @staticmethod
     def _enumerate_output_devices() -> list[dict[str, Any]]:
-        """Enumerate available audio output devices via sounddevice."""
+        """Enumerate available audio output devices via sounddevice.
```
marcelveldt:
```suggestion
        """
        Enumerate available audio output devices via sounddevice.
```

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -367,17 +377,36 @@ async def discover_and_register(self) -> None:
...
+            if max_output_channels < 2:
+                continue
+            # Test if the device can actually be opened - many ALSA virtual
+            # devices will fail if PipeWire or another audio server is active
+            try:
+                test_stream = sd.RawOutputStream(
+                    device=idx,
+                    samplerate=BRIDGE_SAMPLE_RATE,
+                    channels=BRIDGE_CHANNELS,
+                    dtype="int16",
```
marcelveldt, line 403: Follow-up improvement: we need to be able to specify the specify the supported sample rates per output and default to 48khz

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -367,17 +377,37 @@ async def discover_and_register(self) -> None:
...
         for idx, dev in enumerate(all_devices):
             max_output_channels: int = dev.get("max_output_channels", 0)
-            if max_output_channels >= 2:
-                dev_with_index = dict(dev)
-                dev_with_index["index"] = idx
-                devices.append(dev_with_index)
+            if max_output_channels < 2:
+                continue
+            # Test if the device can actually be opened - many ALSA virtual
+            # devices will fail if PipeWire or another audio server is active
```
OzGav, line 396:
```suggestion
            # devices will fail if PipeWire or another audio server is active
            # TODO: query supported sample rates per device and default to 48kHz
```

music_assistant/providers/local_audio/sendspin_bridge.py
OzGav: Agreed, that's a good improvement. I've added a TODO

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -266,19 +266,29 @@ async def _stop_streaming_locked(self) -> None:
...
+            try:
+                await asyncio.wait_for(self._writer_task, timeout=2.0)
+            except TimeoutError:
+                # Writer didn't stop in time - it may be stuck in a blocking write.
+                # Cancel it as a last resort but don't close the stream until it's done.
+                self._writer_task.cancel()
+                with suppress(asyncio.CancelledError, Exception):
+                    await self._writer_task
+            except (asyncio.CancelledError, Exception):  # noqa: S110
+                pass
```
MarvinSchenkel, line 286: Why are we swallowing all Exceptions here?

OzGav: I think this was just me being overly cautious to make sure the cleanup never raised during shutdown, but you're right — there's no good reason to swallow everything here. The only expected exception is CancelledError, anything else we'd want to see. Fixed to only catch CancelledError in both places. 

Looking at the original code I was probably just following that example...?

marcelveldt (approved):
Thanks, @OzGav !

PR #3619, opened 2026-04-09
Add secondary sort by year for album artist sorting

OzGav:
When sorting albums by artist name, add year as a secondary sort key so albums by the same artist appear in descending chronological order instead of arbitrary database order.

### DIFF
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -81,8 +81,8 @@ SORT_KEYS = {
     "year_desc": "year DESC",
     "position": "position ASC",
     "position_desc": "position DESC",
-    "artist_name": "artists.search_name ASC",
-    "artist_name_desc": "artists.search_name DESC",
+    "artist_name": "artists.search_name ASC, year ASC",
+    "artist_name_desc": "artists.search_name DESC, year ASC",
     "random": "RANDOM()",
     "random_play_count": "RANDOM(), play_count ASC",
 }
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -81,8 +81,8 @@ SORT_KEYS = {
     "year_desc": "year DESC",
     "position": "position ASC",
     "position_desc": "position DESC",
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
Just playing devil's advocate here , wouldn't you want to see the most recent releases first? So `year DESC` in that case?

OzGav:
It would not occur to me to look at them that way. In my experience all discographies are in chronological order https://www.discogs.com/artist/137418-Billy-Joel

MarvinSchenkel:
Apple Music:
image

YT Music
image

Spotify
image
(Their terrible UX shows all songs but it starts at the most recent album)

OzGav:
I feel this is ging to be one of those contentious UX choices with a roughly 50/50 split. Reference-style sources like Discogs list albums in ascending order to show an artist’s evolution and support exploration, while the streaming services you showed prefer descending, probably to surface the most recent content for quicker discovery. They reflect different user intents—understanding a discography vs. finding something to play. Since we seem to focussing on discovery lets go with descending but I expect feedback and we may need to add another sort option for ascending. I could do that now but it will require a minor frontend change to surface it. What do you think?

MarvinSchenkel:
> I feel this is ging to be one of those contentious UX choices with a roughly 50/50 split. Reference-style sources like Discogs list albums in ascending order to show an artist’s evolution and support exploration, while the streaming services you showed prefer descending, probably to surface the most recent content for quicker discovery. They reflect different user intents—understanding a discography vs. finding something to play. Since we seem to focussing on discovery lets go with descending but I expect feedback and we may need to add another sort option for ascending. I could do that now but it will require a minor frontend change to surface it. What do you think?

Let's go with DESC as it's better than the current approach. I would wait with changing the frontend until we actually get feedback about this.

OzGav:
As an option to declutter the options somewhat is to drop the last two. Do people want to reverse sort by album artist?

image

OzGav:
As discussed back to plan A but with year descending

MarvinSchenkel (approved):

PR #3621, opened 2026-04-09
Propagate now-playing info to Chromecast media_player

robtpaton:
Since 2.8.0, only limited now-playing metadata is propagated to the native Chromecast media_player entity, while full metadata (title, artist, artwork...) is only propagated to the Music Assistant media_player entity.

This change restores the pre-2.8.0 behaviour.

**Root Cause**

Two properties exist on the Player model:
* `self.current_media` → `self._attr_current_media` — raw provider-level metadata. For Chromecast playing an MA radio stream, this contains the stream URL and `MediaType.TRACK`, with no ICY/HLS song info.
* `self.state.current_media` — final calculated state from `__final_current_media()`, which overlays `streamdetails.stream_metadata` (title, artist, album, artwork from ICY/HLS tags). This is what the MA UI displays.

The Chromecast's `_on_player_media_updated()` callback reads `self.current_media` (line 300), so it never sees the stream metadata.

**Changes**

Change the metadata source from `self.current_media` to `self.state.current_media`:
 
* Line 300: `self.current_media` → `self.state.current_media` (so current_media local var has stream metadata)
* Line 303: Keep the /flow/ URI check but use `self._attr_current_media.uri `since the stream URL (containing /flow/) is only in the raw provider state; `self.state.current_media.uri` contains the queue item URI (e.g. library://radio/123)
* Line 304: `self.current_media.media_type` → `current_media.media_type `(use local var, which now points to `state.current_media` with correct `MediaType.RADIO`)

Also add a guard for `self._attr_current_media` being None (for the URI check).

Tested successfully using the Dev Server addon:

Screenshot 2026-04-09 at 11 50 24


marcelveldt:
We should already have a dedicated callback for this in the Player model

robtpaton:
> We should already have a dedicated callback for this in the Player model

Thanks, changed!

marcelveldt (approved):
Thanks, @robtpaton !

robtpaton:
Hi @marcelveldt would this change be suitable to be backported to stable?

PR #3623, opened 2026-04-09
Implement metadata provider priority

OzGav:
Per discussion here https://discord.com/channels/753947050995089438/1470389250993422509

Added a priority property to the MetadataProvider base model (default 50, lower = preferred). The metadata controller now iterates providers in priority order so images from preferred sources appear first in the list.                                                                                             

Fix fanart.tv album lookups — A related problem that was preventing this from working as designed was the fanart.tv provider was using artist image keys (artistthumb, hdmusiclogo, etc.) for album responses too, so album cover art was silently ignored. Split into ARTIST_IMG_MAPPING and ALBUM_IMG_MAPPING (albumcover, cdart).

### DIFF
--- a/music_assistant/constants.py
+++ b/music_assistant/constants.py
@@ -1034,3 +1034,7 @@ EXTERNAL_SOURCES: Final[set[str]] = {
     # external (hass_players)
     "external",
 }
+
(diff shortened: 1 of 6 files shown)


marcelveldt (approved):
Thanks, @OzGav !

PR #3632, opened 2026-04-10
Fix Sendspin bridge race causing Chromecast player to disappear

maximmaxim345:
Snapshot the bridge hello to prevent the Cast Receiver JS client from overwriting it on reconnect, which sometimes caused the native player to be deleted.

Related issues:  music-assistant/support#5244, music-assistant/support#5137


marcelveldt (approved):
Thanks, @maximmaxim345 !

PR #3635, opened 2026-04-10
Add controller startup retry to HEOS

Tommatheussen:
reported on Discord, if devices are set to power saving (ie turning of the internal networking when the devices is in standby), Music Assistant won't find the device after it does start up.
Upon debugging, it was found out those devices send out mDNS requests before the internal CLI is fully ready, specifically pyheos sends some initialization commands to fully populate the Heos class. These commands can time out and result in a failed setup.

This PR adds a retrying functionality with backoff strategy. We close the connection between failed attempts, as pyheos does keep some internal heartbeating active, even though the full init process failed.

Some additional changes to the mDNS I made:
- Explicitly discover players after the mDNS controller is created, if MA is already running when the first HEOS device comes online, won't discover players. The `discover_players` only gets called by MA itself when the provider gets loaded, not on every mDNS update
- Clear the existing/partial controller if we fail to connect. This allows the provider to retry a setup if another mDNS attempt failed for whatever reason.


music_assistant/providers/heos/provider.py
```diff
@@ -205,6 +249,14 @@ async def on_mdns_service_state_change(
...
+                "Failed to set up HEOS controller at %s discovered via mDNS", device_ip
+            )
+            if self._heos:
+                with suppress(Exception):
+                    await self._heos.disconnect()
+                self._heos = None
         finally:
             self._controller_discovery_running = False
+
+        await self.discover_players()
```
marcelveldt, line 262: why is this needed ? That should already be handled by default

Tommatheussen: Should the default also handle when the first mDNS is received after the provider was already active? It looked like it wasn't triggering it for me in that case. 
If that should work I might have to check whether the controller setup really gets finished in that case

marcelveldt: Yes, it will push the mdns events from cache

Tommatheussen: I've been trying to get this case to work, but it never triggers the player discovery. As far as I can see from the code, the player discovery is run when mDNS events are served from cache, when the provider gets loaded, but not when the provider was already active. This is my case:
- MA is already running, the provider is already active and no HEOS devices are on the network.
- Plugging in a new HEOS device, sends out mDNS events, which we handle and set up the controller.
- Player discovery is never triggered
- Reloading the provider does trigger the player discovery after mDNS are served (from cache then)

marcelveldt: So you either do manual discovery or you implement mdns (or upnp) discovery.
Do you fully implement the mdns callback ? Then you should get notifications of mdns events and this callback also gets called when the provider is (re)loaded.

So I'm not fully understanding why you need to manually call discovery ? Or are you using a manual discovery path ? In that case I didnt pay enough attention

Tommatheussen: It's kind of a special situation, as HEOS implements mDNS, but we only use 1 speaker for the main controller. It will create an internal conneciton to the other speakers.

So if we configure a specific IP to use, we use a fully manual discovery, creating the controller at that specific IP.
If we configure mDNS, we grab the first mDNS entry and set up the controller from that one.

The tricky part here is, the mDNS callback doesn't create the players, they are only created on the discover_players call. But only if we have a controller setup.

So in the case where the provider is loaded, with no controller, it will trigger player discover, but skipped because no controller is present yet.
On subsequent mDNS events, we create the controller, but the discover_players isn't fired from MA, only when the provider gets (re)loaded

marcelveldt: ok get it

marcelveldt (approved):

PR #3647, opened 2026-04-11
Prevent filesystem unavailability from removing all associated library items

### LINKED ISSUE #5243: Repeated synchronization deletes existing libraries

2.8.1 / 2.8.2

Home Assistant OS Addon

- [ ] I have tried a wired connection for issues related to interrupted or poor playback quality
- [ ] If the problem relates to a device then I have checked the device settings
- [ ] If it is a frontend issue, I have tried a different widely used browser
- [ ] For voice problems, I have sought help elsewhere before returning here
- [ ] For playback problems, I have recycled power to the physical device

### The problem

Since version 2.8.1, my music collection on a Synology NAS starts synchronizing every 12 hours, but before that, MA deletes everything that has already been scanned. This means the indexing always starts from scratch and can take days because it is so unbelievable slow, so it never actually finishes.

The indexing seems to completely overload MA, making it unusable unless I manually stop the synchronization. However, this then means I can't find albums/artists/etc. that haven't been indexed by that point.

Music Assistant is running under HA in Proxmox, and the NAS has always been mounted as a network share. I had no problems before the update to version 2.8.1.

### How to reproduce

It reproduces itself ...

### Music Providers

none

### Player Providers

none

### Full log output

none

### Additional information

none

2026.4.1

Home Assistant OS

Proxmox

- [x] Yes

OzGav:
Fixes https://github.com/music-assistant/support/issues/5243

When a filesystem provider could not scan its base path — for example because a NAS share was offline — `recursive_iter()` silently returned an empty iterator. `sync_library()` then computed `deleted_files = prev_filenames - cur_filenames` against that empty set and removed every previously indexed track, album and artist from the library.

A closely related case is an empty-but-reachable mount: if the wrong share is mounted, a backup job has swapped the export, or a stale mount point presents as an empty directory, `recursive_iter()` returns cleanly with zero results and the same mass deletion happens.

This PR adds two guards to `sync_library()` and wires the provider's own available flag into the first one so the rest of the system reacts accordingly.

### DIFF
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -448,17 +448,12 @@ class LocalFileSystemProvider(MusicProvider):
         base_path_reachable = await self._check_base_path_reachable()
         if scan_errors or not base_path_reachable:
             self.logger.error(
-                "Aborting sync for %s before the deletion phase to prevent data loss: "
-                "%d scan error(s) encountered, base path reachable=%s. "
-                "The previously indexed library items will be kept untouched.",
+                "Aborting sync for %s: %d scan error(s), base path reachable=%s",
                 self.name,
                 len(scan_errors),
                 base_path_reachable,
             )
-            report_current_task_failure(
-                "Sync aborted: filesystem became unavailable during the scan. "
-                "Library items were preserved."
-            )
+            report_current_task_failure("Sync aborted: filesystem unavailable during scan")
             return
 
         # work out deletions
--- a/music_assistant/providers/filesystem_local/__init__.py
+++ b/music_assistant/providers/filesystem_local/__init__.py
@@ -20,6 +20,7 @@ import xmltodict
 from aiofiles.os import wrap
 from music_assistant_models.enums import (
     ContentType,
+    EventType,
     ExternalID,
     ImageType,
     MediaType,
@@ -449,6 +450,7 @@ class LocalFileSystemProvider(MusicProvider):
                 base_path_reachable,
             )
             report_current_task_failure("Sync aborted: filesystem unavailable during scan")
+            self._set_available(False)
             return
 
         # work out deletions
@@ -457,6 +459,9 @@ class LocalFileSystemProvider(MusicProvider):
 
         # process orphaned albums and artists
         await self._process_orphaned_albums_and_artists()
+
+        # flag provider as available again if an earlier sync had marked it down
+        self._set_available(True)
 
     async def _check_base_path_reachable(self) -> bool:
         """Return whether the provider's base path is currently reachable."""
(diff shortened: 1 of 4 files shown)


music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -376,12 +377,18 @@ async def sync_library(self, media_type: MediaType) -> None:
...
         def enumerate_files() -> None:
             """Enumerate all files, collecting changed items for processing."""
             scanned = 0
             for item in recursive_iter(
-                self.base_path, self.base_path, SUPPORTED_EXTENSIONS, self.logger
+                self.base_path,
+                self.base_path,
+                SUPPORTED_EXTENSIONS,
+                self.logger,
+                scan_errors,
```
marcelveldt: can we make this named parameters?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -433,13 +440,43 @@ async def _process(item: FileSystemItem, prev_checksum: str | None) -> None:
...
+            return
+
         deleted_files = prev_filenames - cur_filenames
         await self._process_deletions(deleted_files)
-
-        # process orphaned albums and artists
         await self._process_orphaned_albums_and_artists()
 
+        # flag provider as available again if an earlier sync had marked it down
+        self._set_available(True)
```
marcelveldt, line 474: How would this work ? Isn't the scan skipped when the provider is unavailable?

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -433,13 +440,43 @@ async def _process(item: FileSystemItem, prev_checksum: str | None) -> None:
         finally:
             self.sync_running = False
 
-        # work out deletions
+        # do not run deletions if the root base path could not be scanned
+        if scan_errors:
```
marcelveldt: so if any error occurs we would fail ? What about some directories that are not traversable ?

music_assistant/providers/filesystem_local/__init__.py
OzGav: The scheduled sync task is registered once when the provider is loaded, and its handler holds a direct reference to the provider and just calls `provider.sync_library(media_type)` with no availability check. Therefore, if the share is back online it will be marked as available and the sync will complete. Outside of the auto sync interval the menu for the provider has the "synchronise now" menu item removed but the user can still click on "Reload" and that will mark the provider as available again (assuming it is actually available) and then they can sync manually

music_assistant/providers/filesystem_local/__init__.py
OzGav: Done

music_assistant/providers/filesystem_local/__init__.py
OzGav: The abort only fires on root-level failures, not on "any" error. The code has been changed to make that more clear.

marcelveldt (approved):
Thanks, @OzGav !

PR #3650, opened 2026-04-11
Use original stream for ibroadcast

robsonke:
- Use the original file format that was uploaded into ibroadcast, let ffmpeg decide the rest.
- Some small stability fixes.


marcelveldt (approved):
LGTM, thanks @robsonke !

PR #3651, opened 2026-04-11
Correct link to liblistenbrainz

ijc:
Param Singh was the original author but maintenance is now under the musicbrainz umbrella.

---

Noticed in passing by @dassystem in https://github.com/orgs/music-assistant/discussions/4446#discussioncomment-16508312.


marcelveldt (approved):
Thanks, @ijc !

PR #3662, opened 2026-04-12
Subsonic: Cache requests

khers:
Using the internal cache is simple and it's silly to reach out to the subsonic server with each page change. Let's cache the returns from most server interactions.


marcelveldt (approved):
Thanks, @khers !

PR #3674, opened 2026-04-13
Change Mother Earth Radio icon URL and corrupt SVGs

OzGav:
Current URL is throwing a 403. Move to the MA GitHub repo as per other providers

SVGs weren't rendering properly so new versions added


marcelveldt (approved):
Thanks, @OzGav !

PR #3681, opened 2026-04-13
Sendspin: guard against negative track_progress in metadata

marcelveldt:
## Problem

`SendspinPlayer.send_current_media_metadata` occasionally crashes with:

```
ValueError: track_progress must be non-negative, got -1000
```

`current_media.corrected_elapsed_time` can briefly extrapolate to a small negative value around track transitions / seek / clock jitter. Multiplied by 1000 it fails the strict non-negative validator in aiosendspin's `Progress` model and spams tracebacks in the log.

## Changes

- Clamp `track_progress` to `max(0, ...)` before passing it to the `Metadata` payload in `providers/sendspin/player.py`.


apophisnow (approved):
This looks good to me.

PR #3695, opened 2026-04-14
Update base image versions to 1.4.14

apophisnow:


.github/workflows/release.yml
```diff
@@ -42,8 +42,8 @@ on:
 env:
   PYTHON_VERSION: "3.12"
   BASE_IMAGE_VERSION_STABLE: "1.4.13"
```
marcelveldt, line 44: don't forget to update the base image when we do our next stable release
if the changes in the base image are rather harmless, consider bumping sooner

marcelveldt (approved):

.github/workflows/release.yml
apophisnow: Sounds good. Shouldn't be an impactful change, so I'll probably bump it soon.

PR #3701, opened 2026-04-15
Fix Mother Earth indicated audio quality

OzGav:


music_assistant/providers/motherearthradio/provider.py
```diff
@@ -91,6 +91,8 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
             provider=self.instance_id,
             audio_format=AudioFormat(
                 content_type=channel_info["content_type"],
+                sample_rate=MER_SAMPLE_RATE,
+                bit_depth=MER_BIT_DEPTH,
```
marcelveldt, line 95: Is this really this format all the time ?
I think we can also fix our detection logic (where detect the codec) to also detect sample rate and bit depth from the source - if its not provided

OzGav: Yes I believe so. Yes I thought it would be better to detect the quality on the fly but looked around and we dont seem to be doing that in the other providers

marcelveldt: yeah we can add it later

marcelveldt (approved):
Thanks, @OzGav !

PR #3704, opened 2026-04-15
Add setting to replace Pause with Stop for DLNA

peteS-UK:
Some older uPnP players are unable to pause streamed music - they tend to just ignore a pause command when the source is uPnP/DLNA.  This means that MA sends a pause, thinks the music is paused and the player is in a paused state, but the stream continues to play.  This is made worse because most players seem to report they are able to pause in their CTA if they're able to pause for non-streams - i.e. because they're capable of pausing for some source types, they don't differentiate and reports they can pause for all types.

To overcome this, this PR adds a per-player advanced setting to always replace pause with stop.


marcelveldt (approved):
Thanks, @peteS-UK !

PR #3705, opened 2026-04-15
Subsonic: always request raw source file and let MA handle seeking

marcelveldt:
## Problem

Server-side seeking via the OpenSubsonic `transcodeOffset` extension is unreliable across implementations. Navidrome in particular ignores `timeOffset` for direct-passthrough codecs like **opus**, which causes playback to restart from the beginning when the user seeks (reported in 5210, earlier in 3464).

Additionally, any server-side transcoding was being thrown away and re-decoded by ffmpeg on MA's side — wasteful and lossy for no benefit.

## Changes

- Always pass `format=raw` to the Subsonic stream endpoint so MA receives the original source file
- Set `can_seek=False` on stream details so MA's core pipeline handles seeking for all codecs
- Drop the `transcodeOffset` extension probe and the `_seek_support` flag
- Remove the now-obsolete "Force Player Provider Seek" config option

Fixes  support issues 5210, 3464.

### DIFF
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -77,7 +77,6 @@ if TYPE_CHECKING:
 CONF_BASE_URL = "baseURL"
 CONF_ENABLE_PODCASTS = "enable_podcasts"
 CONF_ENABLE_LEGACY_AUTH = "enable_legacy_auth"
-CONF_OVERRIDE_OFFSET = "override_transcode_offest"
 CONF_RECO_FAVES = "recommend_favorites"
 CONF_NEW_ALBUMS = "recommend_new"
 CONF_PLAYED_ALBUMS = "recommend_played"
@@ -96,8 +95,6 @@ class OpenSonicProvider(MusicProvider):
 
     conn: SonicConnection
     _enable_podcasts: bool = True
-    _seek_support: bool = False
-    _ignore_offset: bool = False
     _show_faves: bool = True
     _show_new: bool = True
     _show_played: bool = True
@@ -131,16 +128,13 @@ class OpenSonicProvider(MusicProvider):
             )
             raise LoginFailed(msg) from e
         self._enable_podcasts = bool(self.config.get_value(CONF_ENABLE_PODCASTS))
-        self._ignore_offset = bool(self.config.get_value(CONF_OVERRIDE_OFFSET))
         try:
             extensions: list[OpenSubsonicExtension] = await self.conn.get_open_subsonic_extensions()
             for entry in extensions:
-                if entry.name == "transcodeOffset" and not self._ignore_offset:
-                    self._seek_support = True
                 if entry.name == "songLyrics":
                     self._id_lyrics = True
         except OSError:
-            self.logger.info("Server does not support transcodeOffset, seeking in player provider")
+            self.logger.debug("Server does not advertise OpenSubsonic extensions")
         self._show_faves = bool(self.config.get_value(CONF_RECO_FAVES))
         self._show_new = bool(self.config.get_value(CONF_NEW_ALBUMS))
         self._show_played = bool(self.config.get_value(CONF_PLAYED_ALBUMS))
@@ -677,7 +671,7 @@ class OpenSonicProvider(MusicProvider):
             item_id=item.id,
             provider=self.instance_id,
             allow_seek=True,
-            can_seek=self._seek_support,
+            can_seek=False,
             media_type=media_type,
             audio_format=AudioFormat(
                 content_type=ContentType.try_parse(mime_type),
@@ -776,16 +770,13 @@ class OpenSonicProvider(MusicProvider):
         self, streamdetails: StreamDetails, seek_position: int = 0
     ) -> AsyncGenerator[bytes, None]:
         """Provide a generator for the stream data."""
-        # ignore seek position if the server does not support it
-        # in that case we let the core handle seeking
-        if not self._seek_support:
-            seek_position = 0
-
+        # Always request the raw (untranscoded) source file and let MA's core stream
+        # pipeline handle decoding, seeking and any optional re-encoding. Server-side
+        # transcoding with timeOffset is unreliable across OpenSubsonic implementations
+        # (e.g. Navidrome ignores timeOffset for direct-passthrough codecs like opus).
         self.logger.debug("Streaming %s", streamdetails.item_id)
         try:
-            resp = await self.conn.stream(
-                streamdetails.item_id, time_offset=seek_position, estimate_length=True
-            )
+            resp = await self.conn.stream(streamdetails.item_id, tformat="raw")
         except DataNotFoundError as err:
             msg = f"Item '{streamdetails.item_id}' not found"
             raise MediaNotFoundError(msg) from err
--- a/music_assistant/providers/opensubsonic/sonic_provider.py
+++ b/music_assistant/providers/opensubsonic/sonic_provider.py
@@ -2,9 +2,12 @@
 
 from __future__ import annotations
 
+import os
 from asyncio import TaskGroup
 from datetime import datetime
+from hashlib import md5
 from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar
+from urllib.parse import urlencode
 
 from libopensonic import AsyncConnection as SonicConnection
 from libopensonic.errors import (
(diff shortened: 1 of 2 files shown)


marcelveldt:
@khers can you review this one and test it ?

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -677,7 +671,7 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
             item_id=item.id,
             provider=self.instance_id,
             allow_seek=True,
```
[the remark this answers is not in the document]
marcelveldt: this seems unavoidable with subsonic - it doesn't expose a HTTP URL as far as I'm aware.
it's also less of an issue since the AudioBuffer handles this pretty well in most occasions.

But agreed, a raw stream URL and no custom audio stream would be much bettter

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -776,16 +770,13 @@ async def get_audio_stream(
...
+        # Always request the raw (untranscoded) source file and let MA's core stream
+        # pipeline handle decoding, seeking and any optional re-encoding. Server-side
+        # transcoding with timeOffset is unreliable across OpenSubsonic implementations
+        # (e.g. Navidrome ignores timeOffset for direct-passthrough codecs like opus).
         self.logger.debug("Streaming %s", streamdetails.item_id)
         try:
-            resp = await self.conn.stream(
-                streamdetails.item_id, time_offset=seek_position, estimate_length=True
-            )
+            resp = await self.conn.stream(streamdetails.item_id, tformat="raw")
```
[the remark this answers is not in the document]
khers: This is correct, I left a comment where the change needs to be made.

music_assistant/providers/opensubsonic/sonic_provider.py
khers: If we are going to force raw streaming always, then we should not use `item.transcoded_content_type` ever.

khers:
I disagree with the assertion that transcoding on the server was wasteful or pointless. It gives the user control over streaming bandwidth for folks with high quality, large recordings.

It is unclear to me why this commit is different from the user setting Force Player Seek.

I am not opposed to merging this but we need to be clear in the documentation for the provider that we do not honor transcoding setups and will always request the raw file.

khers (commented):
I will try and test this tonight.

marcelveldt:
> I disagree with the assertion that transcoding on the server was wasteful or pointless. It gives the user control over streaming bandwidth for folks with high quality, large recordings.

Isn't it true that the subsonic server is always hosted locally so data transfer is cheap ?
Transcoding on the subsonic server means a loss of audio quality and waste of cpu cycles.

One tradeoff versus the other.
In general we just request the original/raw audio file from provider sand let MA handle the transcoding if/when needed to keep the chain as pure as possible. 

I saw a lot of issue reports about seeking with subsonic and how it's kind of a hot mess in the various implementations so I've let Claude Code poke into it a bit base don my instructions. But I have no inner knowledge of Subsonic at all.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -627,6 +624,28 @@
...
+        if conn.api_key:
+            qdict["apiKey"] = conn.api_key
+        else:
+            qdict["u"] = conn.username
+            if conn.legacy_auth:
+                qdict["p"] = "enc:" + "".join(f"{ord(c):02X}" for c in conn.password)
+            else:
+                salt = md5(os.urandom(100)).hexdigest()[:16]
+                qdict["s"] = salt
+                qdict["t"] = md5((conn.password + salt).encode("utf-8")).hexdigest()
```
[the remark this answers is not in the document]
marcelveldt: unavoidable - this is how subsonic does this, not our choice

music_assistant/providers/opensubsonic/sonic_provider.py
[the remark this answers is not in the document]
marcelveldt: avoided with fetching as HTTP directly

khers:
> > I disagree with the assertion that transcoding on the server was wasteful or pointless. It gives the user control over streaming bandwidth for folks with high quality, large recordings.
> 
> Isn't it true that the subsonic server is always hosted locally so data transfer is cheap ? Transcoding on the subsonic server means a loss of audio quality and waste of cpu cycles.

I don't think we can make this statement at all. My Gonic instance lives on a VM in a data center so that my family and I can use it when we are not home (I live in a backwards place with a single ISP that offers up to 10Mb upstream which means I see ~3-5 normally and prohibits any hosting activity from the only kind of account I can have at my address).

> 
> One tradeoff versus the other. In general we just request the original/raw audio file from provider sand let MA handle the transcoding if/when needed to keep the chain as pure as possible.
> 
> I saw a lot of issue reports about seeking with subsonic and how it's kind of a hot mess in the various implementations so I've let Claude Code poke into it a bit base don my instructions. But I have no inner knowledge of Subsonic at all.

The force provider seek question has nothing to do with Subsonic, rather this is forcing the user into a single configuration rather than letting them decide. Gonic and Navidrome don't honor transcodeOffset always, but I believe that Ampache does for instance.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -621,6 +624,28 @@ async def remove_playlist_tracks(
             msg = f"Failed to remove songs from {prov_playlist_id}, check your permissions."
             raise ProviderPermissionDenied(msg) from ex
 
+    def _build_stream_url(self, item_id: str) -> str:
```
khers, line 627: This feels really bad. Let me add a method to the AsyncConnection object that builds a URL and returns it instead so at least we can keep that wierdness contained to the py-opensonic library.

khers: Alternatively, we could simply return the aiohttp.ClientResponse object rather than chunking it in the async with

khers (changes requested):

music_assistant/providers/opensubsonic/sonic_provider.py
khers: The more I think about this, the less I like it. When using the library, arguments are passed in the request body via POST and therefore encrypted when HTTPS is in use. This moves them all to the URL and always in the clear. Not a problem for API key users but would surprise u+p users that rely on POST.

Returning the ClientResponse is the best option here, IMO.

marcelveldt: the best thing would be if you can return the full URL so we can let ffmpeg deal with it.
pretty sure seeking will then also work out of the box with plain range requests.

So some helper function to return the stream url would be very nice.
If that really is not possible, then revert to the custom stream approach again, which can be the ClientResponse if you want. 

But in general passing a file or url and let our internal ffmpeg deal with it is a lot better.

marcelveldt:
> I don't think we can make this statement at all. My Gonic instance lives on a VM in a data center so that my family and I can use it when we are not home (I live in a backwards place with a single ISP that offers up to 10Mb upstream which means I see ~3-5 normally and prohibits any hosting activity from the only kind of account I can have at my address).

OK, my bad. I made assumptions about subsonic servers always being home/self hosted.
In that case maybe offer some toggle of what format to request from the server ?
The nicest would be to just return the file as-is without any transcoding or whatsoever

music_assistant/providers/opensubsonic/sonic_provider.py
khers: Please pardon my technical ignorance here but what is the difference between returning a URL and having ffmpeg make the request (range requests are still going to be ignored by the server) and handing the ClientResponse Object and giving ffmpeg the underlying socket fd to pull from?

To the best of my knowledge, all of the subsonic client applications work by actually downloading the entire media file locally and then handling seeking from that. I don't think any of them play exclusively from stream.

Returning a URL is always going to have the problem of exposing authentication details in the URL parameters.

marcelveldt: My answer was range requests but if that is not supported anyways, then the current approach is fine enough. If the user has a more ore less descent setup with enough memory, the file will be loaded in memory and seeking performed from there.

So if the server will not support range requests anyways, there is zero benefit of changing this

marcelveldt:
I closed this in favor of your PR to add an option to request the raw file

PR #3706, opened 2026-04-15
Fix smart fades mixer sometimes choking up the flow stream + Smart Fades provider not starting on ARM

MarvinSchenkel:
When Smart Fades falls back to a standard crossfade, it yield the entire pre and post sections as a single chunk, causing the entire audio pipeline to stall.

Solution: yield 1s chunks.


marcelveldt:
This should have been 2 PR's because now the fix can't be backported anymore.

marcelveldt (approved):

PR #3710, opened 2026-04-16
Subsonic: Add option to always request stream of raw file

khers:
In the case that either Subsonic is home hosted on the same network as MA or hosted in a place with enough outgoing bandwidth we want to stream the raw files rather than transcode. Make this a user configurable option that defaults to requesting raw.


marcelveldt (approved):
Thanks, @khers !

PR #3719, opened 2026-04-16
Fix audiobook controller not using userid in library_items call

fmunkes:
Small fix which ensures, that the in_progress/finished icons in the frontend refer to the correct user.

Shall we backport this one?


marcelveldt (approved):
Thanks, @fmunkes !

PR #3723, opened 2026-04-17
Bump auntie-sounds to 1.1.8

### LINKED ISSUE #8: NameError: name 'PlayableSoundsTypes' is not defined.
I have Music Assistant Server running in a Docker container on a Raspberry Pi with an image from ghcr.io/music-assistant/server:beta

Music Assistant Server 2.9.0b6
auntie-sounds             1.1.7

I can add  BBC Sounds as a Music Source but when I try  browsing the BBC Sounds folder I get a message pop up "name 'PlayableSoundsTypes' is not defined" error.

I think the relevant part of the music assistant log is 

`File "/app/venv/lib/python3.13/site-packages/sounds/models.py", line 636, in model_factory
    elif schema_type in PlayableSoundsTypes:
                        ^^^^^^^^^^^^^^^^^^^
NameError: name 'PlayableSoundsTypes' is not defined. Did you mean: 'BaseSoundsTypes'?`

I notice that in models.py PlayableSoundsTypes is conditionally imported

`if TYPE_CHECKING:
    from .constants import PlayableSoundsTypes`

If I import PlayableSoundsTypes unconditionally (with BaseSoundsTypes) then the error doesn't occur. (I have a local installation where I can edit the .venv files)

Does PlayableSoundsTypes need to be conditionally imported or can it be imported alongside BaseSoundsTypes?

And many thanks for the great plugin - I've been relying on expectingtofly's LMS one for many years, but Music Assistant looks promising.

kieranhogg:
Fixes https://github.com/kieranhogg/auntie-sounds/issues/8

### DIFF
--- a/music_assistant/providers/bbc_sounds/manifest.json
+++ b/music_assistant/providers/bbc_sounds/manifest.json
@@ -11,7 +11,7 @@
     "[auntie-sounds](https://github.com/kieranhogg/auntie-sounds)"
   ],
   "requirements": [
-    "auntie-sounds==1.1.7",
+    "auntie-sounds==1.1.8",
     "pytz==2025.2"
   ],
   "multi_instance": false
--- a/requirements_all.txt
+++ b/requirements_all.txt
@@ -21,7 +21,7 @@ aiovban==0.6.3
 alexapy==1.29.17
 async-upnp-client==0.46.2
 audible==0.10.0
-auntie-sounds==1.1.7
+auntie-sounds==1.1.8
 av==16.1.0
 awesomeversion>=24.6.0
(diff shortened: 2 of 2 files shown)


OzGav:
@kieranhogg any reason why this should not be backported? If no I will add the appropriate label,

kieranhogg:
> @kieranhogg any reason why this should not be backported? If no I will add the appropriate label,

No — it was reported in 2.8 and 2.9 and it's a one line bug fix, so that would be appreciated :)

https://github.com/kieranhogg/auntie-sounds/issues/8

MarvinSchenkel (approved):
Thanks @kieranhogg !

PR #3683, opened 2026-04-13
Add Pulse Audio Out player provider

iVolt1:
<html><body>
<html><head></head><body><h2>Add Pulse Audio Out player provider</h2>
<h3>Summary</h3>
<p>Adds a new <code>pulse_audio</code> player provider that exposes PulseAudio sinks as Music Assistant sendspin players. Each sink is registered as an external Sendspin bridge client, enabling synchronized multi-room playback alongside existing Sendspin players. Audio is written directly to PulseAudio via a minimal <code>libpulse-simple</code> ctypes wrapper — no additional Python dependencies are required beyond <code>numpy</code> for software volume scaling.</p>
<h3>Motivation</h3>
<p>Users running Music Assistant on Home Assistant OS with multi-channel sound cards have no way to use those cards as MA players. The existing local audio provider uses PortAudio/sounddevice which cannot enumerate PulseAudio virtual sinks such as <code>module-remap-sink</code> stereo pairs. This provider targets PulseAudio directly via <code>pactl</code> and <code>libpulse-simple</code>, correctly discovering and playing to all sinks including remap sinks, combined sinks, S/PDIF, and HDMI outputs.</p>
<h3>Changes</h3>
<h4>New provider — <code>music_assistant/providers/pulse_audio/</code></h4>

File | Description
-- | --
__init__.py | Provider entry point, config entries, and setup
provider.py | LocalPulseAudioProvider — thin shell delegating to bridge manager
player.py | LocalPulseAudioPlayer — MA player model per sink with software volume and mute
sendspin_bridge.py | SendspinPulseAudioBridge and LocalPulseAudioBridgeManager — sink enumeration via pactl, Sendspin bridge registration, audio writer loop
pa_simple.py | Minimal ctypes wrapper around libpulse-simple for direct PCM playback to a named PA sink
helpers.py | find_pactl() and pactl_env() utilities
constants.py | Shared constants — UUID namespace, config keys, volume control mode
manifest.json | Provider manifest declaring depends_on: sendspin, type player, stage alpha

<h3>Dependencies</h3>
<ul>
<li><code>libpulse-simple.so.0</code> — must be present on the host (standard PulseAudio installation)</li>
<li><code>pactl</code> — required at startup for sink enumeration (<code>pulseaudio-utils</code> on Debian/Ubuntu, <code>pulseaudio</code> on Alpine)</li>
<li><code>numpy</code> — used for PCM volume scaling (already a MA dependency)</li>
[...]

PR #3724, opened 2026-04-17
Extend Local Audio Out provider with PulseAudio support

iVolt1:
<html><body>
<html><head></head><body><h1>Extend Local Audio Out provider with PulseAudio support</h1>
<blockquote>
<p><strong>Note to reviewers:</strong> This PR supersedes #3683 which proposed a separate <code>pulse_audio</code> provider. Based on feedback from that review, this PR instead extends the existing <code>local_audio</code> provider to support PulseAudio on Linux. Key changes addressing the review comments:</p>
<ul>
<li><strong><code>pulseaudio-utils</code> added to <code>Dockerfile.base</code></strong>: Added as a standard apt dependency alongside existing runtime packages — a single line addition rather than a separate build stage</li>
<li><strong>Bundled <code>pactl</code> binary</strong>: Still included in the provider <code>bin/</code> directory to facilitate testing with the MA DEV SERVER app, which does not build from the modified <code>Dockerfile.base</code>. This will be removed in a follow-up PR once validated against MA Nightly</li>
<li><strong>PulseAudio output via <code>libpulse-simple</code></strong>: As suggested, <code>pa_simple.py</code> uses <code>libpulse-simple</code> directly via ctypes rather than PortAudio, avoiding the PortAudio/PulseAudio build-from-source issue entirely</li>
<li><strong>AI assistance disclosure</strong>: Claude was substantially involved in the development of this provider</li>
</ul>
</blockquote>
<h2>Summary</h2>
<p>Extends the existing <code>local_audio</code> player provider to support PulseAudio on Linux, while preserving the existing PortAudio/CoreAudio path on macOS. Each PulseAudio sink is registered as an external Sendspin bridge client, enabling synchronized multi-room playback alongside existing Sendspin players. Audio is written directly to PulseAudio via a minimal <code>libpulse-simple</code> ctypes wrapper. The separate <code>pulse_audio</code> provider is superseded by this change and removed.</p>
<h2>Motivation</h2>
<p>The existing <code>local_audio</code> provider uses PortAudio/sounddevice which cannot enumerate PulseAudio virtual sinks such as <code>module-remap-sink</code> stereo pairs. On Linux, PulseAudio sits on top of ALSA and owns the hardware devices, so PortAudio can only see physical ALSA devices — not virtual sinks. This change targets PulseAudio directly on Linux via <code>pactl</code> and <code>libpulse-simple</code>, correctly discovering and playing to all sinks including remap sinks, combined sinks, S/PDIF, and HDMI outputs, while preserving full macOS functionality.</p>
<h2>Changes</h2>
<h3>Modified — <code>music_assistant/providers/local_audio/</code></h3>

File | Description
-- | --
__init__.py | Added hardware volume ceiling config entry (Linux only); updated volume control mode description
provider.py | Added libpulse-simple presence check on Linux at init
player.py | Added pa_sink_name and is_remap params; Linux uses software volume mode with hardware ceiling set once at startup; apply_hardware_ceiling sets physical ALSA sinks to ceiling value and remap sinks to 100%; pulsectl helpers for volume and mute
sendspin_bridge.py | Unified Linux (PASimpleStream) and Darwin (sounddevice) audio output paths; native format negotiation per sink; initial_volume=25 on bridge registration; apply_hardware_ceiling call per player
pa_simple.py | New file — ctypes wrapper around libpulse-simple for direct PCM playback; enumerate_pa_sinks via pactl --format=json
constants.py | Added CONF_HARDWARE_VOLUME_CEILING, DEFAULT_HARDWARE_VOLUME_CEILING
manifest.json | Added pulsectl; sys_platform == 'linux' to requirements
README.md | Updated for unified Linux/macOS provider
bin/pactl | Bundled pactl binary (fallback if pulseaudio-utils not installed)
bin/lib/libpulsecommon-16.1.so | Bundled library required by bundled pactl binary

<h2>Dependencies</h2>
<ul>
<li><strong>libpulse-simple.so.0</strong> <em>(Linux)</em>: Must be present on the host — standard PulseAudio installation</li>
<li><strong>pactl</strong> <em>(Linux)</em>: Required for sink enumeration — provided by <code>pulseaudio-utils</code> in <code>Dockerfile.base</code>, with bundled binary as fallback</li>
<li><strong>pulsectl</strong> <em>(Linux)</em>: Python PulseAudio bindings for volume and mute control — installed via provider manifest requirements</li>
<li><strong>numpy</strong>: Used for PCM volume scaling (already a MA dependency)</li>
<li><strong>Sendspin provider</strong> (<code>depends_on: sendspin</code>)</li>
</ul>
<h2>Expanding Outputs with Stereo Pair Remap Sinks</h2>
<p>Multi-channel sound cards (5.1, 7.1 surround) expose a single multi-channel PulseAudio sink by default. To use each channel pair as an independent MA player, PulseAudio <code>module-remap-sink</code> can split a multi-channel sink into individual stereo sinks — one per channel pair (front, rear, side, center/LFE). The Local Audio Out provider discovers and registers all remap sinks automatically alongside physical sinks, so no additional configuration is needed in MA once the remap sinks exist.</p>
<p>For Home Assistant OS users, the companion addon <a href="https://github.com/iVolt1/hassio-apps">Pulse Audio Stereo Pairs</a> automates this setup. It runs as a lightweight HA addon that creates the remap sinks on startup and reacts to audio device hot-plug and unplug events, removing the need to configure remap sinks manually via pactl. Once both the addon and this provider are running, each channel pair of every multi-channel card appears as a separate player in Music Assistant.</p></body></html>
</body>
</html># Extend Local Audio Out provider with PulseAudio support


music_assistant/providers/local_audio/__init__.py
```diff
@@ -47,11 +50,33 @@ async def get_config_entries(
...
+                "Disabled passes audio at full volume."
+            ),
         ),
-    )
+    ]
+
+    if sys.platform == "linux":
+        entries.append(
+            ConfigEntry(
+                key=CONF_HARDWARE_VOLUME_CEILING,
```
marcelveldt, line 65: I think we can leave this out ? It seems to be a volume limiter which is already being added to MA core

music_assistant/providers/local_audio/bin/lib/libpulsecommon-16.1.so
marcelveldt, line 1: leave out the binaries 
we should just build a new base image first, so that can be used by the dev addon as well.

Maybe do a quick PR to update the base dockerfile first so I can kickoff a new base image build ?

music_assistant/providers/local_audio/manifest.json
```diff
@@ -3,9 +3,14 @@
   "domain": "local_audio",
   "name": "Local Audio Out",
   "description": "Play audio through locally attached soundcards (USB DACs, built-in speakers, etc.).",
-  "codeowners": ["@music-assistant"],
+  "codeowners": [
+    "@music-assistant"
```
marcelveldt: feel free to add your own name here as codeowner

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -0,0 +1,258 @@
...
+    PA_SAMPLE_S32LE: 32,
+}
+
+
+def _pa_sample_format(bit_depth: int) -> int:
+    """Return PA sample format constant for given bit depth."""
+    if bit_depth == 32:
+        return PA_SAMPLE_S32LE
+    if bit_depth == 24:
+        # MA delivers in 32-bit containers; _apply_software_volume repacks to
```
marcelveldt, line 28: this is actually configurable. MA (and sendspin) can deliver what the player wants

music_assistant/providers/local_audio/player.py
```diff
@@ -71,8 +82,10 @@ def __init__(
         device_uuid = get_device_uuid(device_name, hostapi_index)
         self._attr_device_info.add_identifier(IdentifierType.UUID, device_uuid)
         self._attr_can_group_with = set()
-        self._attr_volume_level = 100
+        self._attr_volume_level = 25 if (sys.platform == "linux" and pa_sink_name) else 100
```
marcelveldt, line 85: we should consider caching this so we can restore it at startup (see squeezelite provider for an example)

music_assistant/providers/local_audio/player.py
```diff
@@ -142,15 +169,94 @@ async def _set_hardware_mute(self, muted: bool) -> None:
...
-                    "amixer", "-c", str(self._device_index), "sset", "Master", toggle
-                )
-                if rc != 0:
-                    self.logger.warning("amixer mute failed for card %d", self._device_index)
+                if _PULSECTL_AVAILABLE and self._pa_sink_name:
+                    loop = asyncio.get_running_loop()
+                    ok = await loop.run_in_executor(
+                        None, self._set_pulse_mute, self._pa_sink_name, muted
+                    )
+                    if not ok:
```
marcelveldt, line 177: consider combining these

music_assistant/providers/local_audio/README.md
```diff
@@ -2,14 +2,17 @@
...
+The Local Audio Out provider exposes locally attached soundcards as players in Music Assistant. On Linux it enumerates PulseAudio sinks (USB DACs, built-in audio, HDMI, remap sinks, virtual sinks, etc.); on macOS it enumerates PortAudio/CoreAudio devices. It leverages the Sendspin provider for synchronization and timing, registering each device as an external Sendspin bridge client.
 
 ### Key Features
 
-- **Automatic Device Discovery**: Enumerates all audio output devices via PortAudio/sounddevice
+- **Automatic Device Discovery**: On Linux, enumerates all PulseAudio output sinks via `pactl --format=json` — returns native sample rate and format regardless of active stream state. On macOS, enumerates via PortAudio/sounddevice
+- **Native Format Negotiation** *(Linux)*: Each PA sink advertises its native sample rate and bit depth (16, 24, or 32-bit) so Music Assistant transcodes to the correct format — no unnecessary resampling
```
marcelveldt, line 10: ah ok - this is indeed what I expected

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -124,17 +139,31 @@ async def start(self) -> None:
...
+            self.logger.error("No BridgePlayerRole found for %s", self.device_name)
+            return
+
+        self._bridge_role.set_callbacks(
+            on_audio_chunk=self._on_audio_chunk,
+            on_volume_change=self._on_volume_change,
+            on_mute_change=self._on_mute_change,
+            on_stream_start=self._on_bridge_stream_start,
+            on_stream_end=self._on_bridge_stream_end,
+            initial_volume=25,
```
marcelveldt, line 160: we should maybe cache this so we can restore last volume (see squeezelite provider)

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -200,52 +221,130 @@ def _on_audio_chunk(self, chunk: AudioChunk) -> None:
         """Handle an incoming audio chunk."""
         if not self._is_streaming:
             return
+        if not hasattr(self, "_logged_chunk_fmt"):
```
marcelveldt, line 224: why didnt we initialize this in the init ?

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -200,52 +221,130 @@ def _on_audio_chunk(self, chunk: AudioChunk) -> None:
...
-        """Write queued audio data to the soundcard."""
+        """Write queued audio to the output device."""
+        if sys.platform == "linux":
+            await self._audio_writer_pulse()
+        else:
+            await self._audio_writer_sounddevice()
+
+    async def _audio_writer_pulse(self) -> None:
+        """Write queued audio to a PA sink via PASimpleStream (Linux)."""
+        loop = asyncio.get_running_loop()
```
marcelveldt, line 277: consider mass.loop here

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -267,39 +366,34 @@ async def _stop_streaming(self) -> None:
...
 
 class LocalAudioBridgeManager:
     """Manages Sendspin bridges for all local audio output devices."""
 
     def __init__(self, provider: LocalAudioProvider) -> None:
-        """
-        Initialize the bridge manager.
-
-        :param provider: The Local Audio provider instance.
-        """
```
marcelveldt, line 302: keep the docstring

music_assistant/providers/local_audio/__init__.py
iVolt1: Will do.

music_assistant/providers/local_audio/bin/lib/libpulsecommon-16.1.so
iVolt1: Sounds good.  Will do.

iVolt1: Do you want the other pulseaudio-utils removed and just leave pactl?

marcelveldt: if those are available as os packages, I'd remove them here and use os installed once

music_assistant/providers/local_audio/pa_simple.py
iVolt1: Could you please clarify what you mean here?

music_assistant/providers/local_audio/bin/lib/libpulsecommon-16.1.so
iVolt1: I was referring to the Dockerfile.base install of pulseaudio-utils.  Something like this: 

        # PortAudio for local audio output (sounddevice)
        libportaudio2 \
        # Allows pipewire devices to be detected when running in docker container. (local audio)
        pipewire-alsa \
        # Allows pulse audio devices to be detected using pactl. (local audio)
        pulseaudio-utils \
        libvorbisidec1 \
        libflac12 \
        libavahi-client3 \
        libavahi-common3 \
        # AirPlay receiver dependencies (shairport-sync)
        libconfig9 \
        libpopt0 \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/* \
    && rm -f \
        /usr/bin/parecord \
        /usr/bin/paplay \
        /usr/bin/parec \
        /usr/bin/pamon \
        /usr/bin/pacat \
        /usr/bin/pasuspender \
        /usr/bin/pacmd

music_assistant/providers/local_audio/pa_simple.py
iVolt1: Not really related to these comments but  I meant to share this earlier.  Pretty cool.

Screenshot 2026-04-17 134721

music_assistant/providers/local_audio/bin/lib/libpulsecommon-16.1.so
marcelveldt: Yes, open a PR just to adjust the docker basefile. I'll merge that and issue a build of the base image

OzGav:
Marking as draft. There are number of mypy/lint problems as well as the review comments above. Select ready to review again when you want another review.

music_assistant/providers/local_audio/bin/lib/libpulsecommon-16.1.so
marcelveldt: New base image was built and dev add-on version 1.5.2 released. Please confirm if this PR works fine with the dev add-on

iVolt1: The PR is working fine in the dev add-on.

iVolt1:
The Dependency Security Check / security-check (pull_request_target)](https://github.com/music-assistant/server/actions/runs/24722720402/job/72315484752?pr=3724) is from the added pulsectl as a Linux-only dependency for PulseAudio volume control. It's a pure Python wrapper around libpulse with no native code of its own. Requesting dependencies-reviewed label.

vingerha:
I did some tests with following feedback
- ubuntu 24.04 machine, pipewire, this machine is purely used to play music
- MA as a container
```
sudo docker run -d \
  --name music_assistant_beta \
  --network host \
  --group-add audio \
  --privileged \
  -v /home/vingerha/docker/music_assistant_beta:/data \
  -v /mnt/qnapmedia/music:/media \
  -v /home/vingerha/docker/music_assistant_beta/local_audio:/app/venv/lib/python3.14/site-packages/music_assistant/providers/local_audio \
  -v /home/vingerha/docker/music_assistant_beta/sendspin:/app/venv/lib/python3.14/site-packages/music_assistant/providers/sendspin \
  -v /run/user/1000/pulse:/run/user/1000/pulse \
  -e PULSE_SERVER=unix:/run/user/1000/pulse/native \
  --cap-add=DAC_READ_SEARCH \
  --cap-add=SYS_ADMIN \
  --security-opt apparmor:unconfined \
  ghcr.io/music-assistant/server:beta
```
After that I installed pulseaudio-utils in the container and restarted.
MA shows to play at 48k/32b, I tested with source files from 32b to 332(DSD) and it 
In order to go 'higher' I modified pipewire.conf to use 96k default, I restarted pipewire service
I restarted MA container so it identifies the new setting. When playing MA does show 96k/32b but after 1-2s play no sound is heared, regardless of the bitrate of the sourcefile. Used a bit oa AI (as this goes beyond my basic knowledge) and this does not really come any further

EDIT: I just noticed you updated the list of files, I was still on yesterday's state

iVolt1:
> I did some tests with following feedback
> 
>     * ubuntu 24.04 machine, pipewire, this machine is purely used to play music
> 
>     * MA as a container
> 
> 
> ```
> sudo docker run -d \
>   --name music_assistant_beta \
>   --network host \
>   --group-add audio \
>   --privileged \
>   -v /home/vingerha/docker/music_assistant_beta:/data \
>   -v /mnt/qnapmedia/music:/media \
>   -v /home/vingerha/docker/music_assistant_beta/local_audio:/app/venv/lib/python3.14/site-packages/music_assistant/providers/local_audio \
>   -v /home/vingerha/docker/music_assistant_beta/sendspin:/app/venv/lib/python3.14/site-packages/music_assistant/providers/sendspin \
>   -v /run/user/1000/pulse:/run/user/1000/pulse \
>   -e PULSE_SERVER=unix:/run/user/1000/pulse/native \
>   --cap-add=DAC_READ_SEARCH \
>   --cap-add=SYS_ADMIN \
>   --security-opt apparmor:unconfined \
>   ghcr.io/music-assistant/server:beta
> ```
> 
> After that I installed pulseaudio-utils in the container and restarted. MA shows to play at 48k/32b, I tested with source files from 32b to 332(DSD) and it In order to go 'higher' I modified pipewire.conf to use 96k default, I restarted pipewire service I restarted MA container so it identifies the new setting. When playing MA does show 96k/32b but after 1-2s play no sound is heared, regardless of the bitrate of the sourcefile. Used a bit oa AI (as this goes beyond my basic knowledge) and this does not really come any further
> 
> EDIT: I just noticed you updated the list of files, I was still on yesterday's state

It looks like you are working off of beta branch.  Possibly the dev branch is required.  What does 'pactl list sinks short' say?

vingerha:
Yeah true, I use beta, should it be dev, there is just a few days diff between them
This is on the host:
```
 pactl list sinks short
50      alsa_output.pci-0000_00_1f.3.hdmi-surround      PipeWire        s32le 6ch 96000Hz       IDLE
```
and this inside the container
```
root@hplaptop-ubuntu:/app/venv#  pactl list sinks short
50      alsa_output.pci-0000_00_1f.3.hdmi-surround      PipeWire        s32le 6ch 96000Hz       IDLE
50      alsa_output.pci-0000_00_1f.3.hdmi-surround      PipeWire        s32le 6ch 96000Hz       IDLE
bash: 50: command not found
```

The error in MA (whenin 96k):
```
[music_assistant.Local Audio Out.bridge.Built-in Audio Digital Surround 5.1 (HDMI)] PA stream error for alsa_output.pci-0000_00_1f.3.hdmi-surround: pa_simple_write failed (pa_error=3)
```

vingerha:
Update, claude suggested this; 
In this case the fix would be replacing PASimpleStream in pa_simple.py with a pa_mainloop-based stream that sets prebuf=0 (start playing immediately without waiting to fill a prebuffer) and a large tlength matching MA's chunk size of ~59000 bytes. That alone would likely fix the 3s drop.

It then generated a whole new pa_main.py which now indeed allows me to play at 92kHz....but the changes are substantial, beyond my knowledge and I just never trust AI to be correct and concise, so without the proper knowledge at my end I cannot judge this.  

[pa_simple.py](https://github.com/user-attachments/files/26968261/pa_simple.py)

EDIT: it is not perfect as I still have ticks at various moments, seemingly random and not a lot but still ....

iVolt1:
> [music_assistant.Local Audio Out.bridge.Built-in Audio Digital Surround 5.1 (HDMI)] PA stream error for alsa_output.pci-0000_00_1f.3.hdmi-surround: pa_simple_write failed (pa_error=3)
> ```

Thanks for testing.

I assume /app/venv/lib/python3.14/site-packages/music_assistant/providers/local_audio and /app/venv/lib/python3.14/site-packages/music_assistant/providers/sendspin came from the latest in MA_pulse_audio in my repo.

Can you run the local_audio provider with debug logging on in the provider settings and share some more log output?

vingerha:
Just to be sure I overwrote the whole local_audio and sendspin files... it still needs the 'fix' as suggested by Claude, else sound dies in 2-3s with the same error in MA log. I will play a while today/tomorrow and add updates if/where needed  

Update: pw top showed a mismatch in frames between MA and PW host....aligned now. There is quite a bit of stuff required on the host to get this local audio stuff (also for my PR) to work.... I wonder how it would work on windows container :)

iVolt1:
> Just to be sure I overwrote the whole local_audio and sendspin files... it still needs the 'fix' as suggested by Claude, else sound dies in 2-3s with the same error in MA log. I will play a while today/tomorrow and add updates if/where needed
> 
> Update: pw top showed a mismatch in frames between MA and PW host....aligned now. There is quite a bit of stuff required on the host to get this local audio stuff (also for my PR) to work.... I wonder how it would work on windows container :)

It sounds like you got 96k working in the container with the alignment? How was that accomplished?  I have been working in the MA dev server app environment so the pulse audio stuff is already configured.

vingerha:
With pipewire correctly configured, when starting MA container it will find the 'best' rate, since I configured my device default on 96k ...it took that. the pa_simple adaptation is still needed to be able to play more than 2-3s though. I will setup dev as well now as else this testing makes less sense...likely back tomorrow.

vingerha:
>  I have been working in the MA dev server app environment so the pulse audio stuff is already configured.

Hmm... whatever I use, nightly or beta-latest, both dockerbase state to have pulseaudio but both fail on pactl, only when I add pulseaudio-utils (again??) then the container is OK and discovers the pulse device.
The downside is that I need to change quant with increasing rate and for now I can only play pw-top error-less at 48k. with 96 and 192 there are infrequent dropouts

image

.vscode/settings.json
```diff
@@ -16,5 +16,8 @@
       "name": "music_assistant_models",
       "depth": 2
     }
-  ]
+  ],
+  "files.associations": {
+    "*.yaml": "home-assistant"
```
MarvinSchenkel: Accidental commit?

iVolt1: Yes.  I reverted the code back to the MA version in my repo so I hope that resolves the issue.  

Anything else besides the security review that needs attention?

Thanks.

vingerha:
With quite a bit of help from AI, I now have this code modified so that it takes the sourcetrack bit/depth and use that to send at the closest (identical or next highest) bitrate. i.e. it switches bitrate when the next track has another bitrate. It works now on my hdmi but this pulseaudio PR never showed my local speaker or line out so I cannot test it.
Questions, assuming you are interested to take this one level further
- since AI often goes rabbit hunting and I am still not the expert.... are you able to review proposed changes on usability/quality? 
- even if the quality looks (more or less) OK, I am not even sure if this approach is correct as changes to local_audio or sendpsin should not just be targetting pulse, would you know how to verify this on other OS/devices?

EDIT: the key reason for changing bits originally was that your ootb code would not play 96k on my setup, so pa_simple.py was modified. Later on I then wanted to be dynamic on the rates..... All in all it may be too much as well for MA to accept, small(er) steps needed. My suggestion would then be to first be able to play on 'any' device target rate. Dynamic changing I may add later as a separate PR, after your PR gets accepted. If you want to stick woith 48k playing only , fine too but I guess there will be bug/issue reports afterwards.

iVolt1:
> EDIT: the key reason for changing bits originally was that your ootb code would not play 96k on my setup, so pa_simple.py was modified. Later on I then wanted to be dynamic on the rates..... All in all it may be too much as well for MA to accept, small(er) steps needed. My suggestion would then be to first be able to play on 'any' device target rate. Dynamic changing I may add later as a separate PR, after your PR gets accepted. If you want to stick woith 48k playing only , fine too but I guess there will be bug/issue reports afterwards.

Hi Arjan, thanks for all the testing effort — it's really helpful to know how this behaves on PipeWire/Docker.
For this PR I'd like to keep the scope focused on the core local audio output feature running on a native MA PulseAudio setup, which is the primary target for the initial version. The pa_simple approach works well in that environment and lets PulseAudio's daemon.conf control sample rate and bit depth globally, which I think is the right design for now — it keeps MA out of the business of managing hardware audio negotiation.

The PipeWire buffer underrun issue you're hitting is better addressed separately once the core PR is merged, either as a follow-up PR or by the maintainers deciding on the right approach for PipeWire compatibility. The threaded mainloop changes are substantial and introduce complexity that I'm not comfortable merging without broader review.
For your immediate needs, setting the default rate in pipewire.conf and using prebuf=0 in a local override of pa_simple.py sounds like a workable solution while this matures.

vingerha:
OK, understood and kips is a good approach, I will take your code as-is and see if I can get it to work on other bitrates, it did not before but I may have deviated to quickly to focus on pa_simple, keep you posted

vingerha:
So, feedback...and the same as it was a few days ago. The current setup does not allow to play at 96k on pipewire, the buffer underrun is there (as expected). Let's see if this is OK enough then so a next PR can take that up :)

EDIT: I still wonder why the container is not picking up on pulseaudio-utils, pactl issue remains with nightly/latest-dev

iVolt1:
I have no problem outputting to multiple pulse audio sinks with higher sample rates per device capabilities on pipewire-pulse using the conf below.

```
system@system-MacPro:~$ cat ~/.config/pipewire/pipewire.conf.d/99-high-samplerate.conf
context.properties = {
    default.clock.rate = 192000
    default.clock.allowed-rates = [ 44100 48000 88200 96000 176400 192000 ]
    default.clock.quantum = 1024
    default.clock.min-quantum = 512
}
system@system-MacPro:~$ pactl list sinks short
50	alsa_output.pci-0000_00_1b.0.analog-stereo	PipeWire	s32le 2ch 192000Hz	RUNNING
377	nx_voice_out	PipeWire	float32le 2ch 192000Hz	RUNNING
483	alsa_output.usb-C-Media_Electronics_Inc._USB_PnP_Sound_Device-00.analog-stereo	PipeWire	s16le 2ch 48000Hz	RUNNING
system@system-MacPro:~$ 
```

```
S   ID  QUANT   RATE    WAIT    BUSY   W/Q   B/Q  ERR FORMAT           NAME 
S   29      0      0    ---     ---   ---   ---     0                  Dummy-Driver
S   30      0      0    ---     ---   ---   ---     0                  Freewheel-Driver
S   41      0      0    ---     ---   ---   ---     0                  Midi-Bridge
R   50   2048 192000 532.2us  55.0us  0.05  0.01    1   S32LE 2 192000 alsa_output.pci-0000_00_1b.0.analog-stereo
R   90    900  48000 115.5us 130.8us  0.01  0.01    0    F32LE 2 48000  + Firefox
R  104   8192 192000 248.1us  10.9us  0.02  0.00    0   S32LE 2 192000  + music-assistant
R  126   8192 192000  95.0us  18.0us  0.01  0.00    0   S32LE 2 192000  + music-assistant
R  106   8192 192000 260.8us  11.2us  0.02  0.00    0   S32LE 2 192000  + music-assistant
R   74   3600  48000 273.7us 127.5us  0.03  0.01    0    F32LE 1 48000  + Firefox
R   66   3600  48000 403.2us  93.7us  0.04  0.01    0    F32LE 1 48000  + Firefox
S   51      0      0    ---     ---   ---   ---     0                  alsa_input.pci-0000_00_1b.0.iec958-stereo
R  122   2048 192000 321.6us   9.3us  0.03  0.00    0    F32P 2 192000 nx_voice_out
R  123      0      0  82.3us 153.7us  0.01  0.01    0    S16LE 2 22050  + NoMachine Stream
R  141   8192 192000  30.5us  23.7us  0.00  0.00    0   S32LE 2 192000  + music-assistant
R  114    512  48000  63.3us   9.1us  0.01  0.00    0    S16LE 2 48000 alsa_output.usb-C-Media_Electronics_Inc._USB_PnP_Sound_Device-00.analog-stereo
R   82   5040  48000  28.1us  11.9us  0.00  0.00    0    S16LE 2 48000  + music-assistant
R  121   2048 192000 184.5us   0.7us  0.02  0.00    0    S16LE 1 48000 alsa_input.usb-C-Media_Electronics_Inc._USB_PnP_Sound_Device-00.mono-fallback
R  107      0      0   6.5us   4.8us  0.00  0.00    0         F32P 1 0  + nx_remapped_out
R  110      0      0  81.6us   8.8us  0.01  0.00    0         F32P 2 0  + input./bin/pw-loopback-16144
R   85      1     25  32.4us  12.8us  0.00  0.00    2       F32LE 1 25  + cinnamon
R   56      1     25  47.6us   6.5us  0.00  0.00    0       F32LE 1 25  + cinnamon
R   60      1     25  56.1us   6.2us  0.01  0.00    0       F32LE 1 25  + cinnamon
system@system-MacPro:~$ 
```

image

vingerha:
> I have no problem outputting to multiple pulse audio sinks with higher sample rates per device capabilities on pipewire-pulse using the conf below.
> 
Re-edit: with only these config settings, my MA takes that value by default for all (!) devices. For hdmi this then presents buffer underruns on anything above 48k
I also wonder how you get to multiple output bitrates, have you tweaked wireplumber?

vingerha:
A related issue I ran into: when playing via the Local Audio / PipeWire output, multiple corked sink-inputs remain active long after playback stops and only restarting the MA container clears them. As a side effect, any USB output devices stay powered on unnecessarily. Do you see the same behaviour?

I traced it to sendspin_bridge.py: the PASimpleStream is not reliably closed on stop because the CancelledError handling interrupts the finally block before stream.close() is called. A fix involving asyncio.shield() around the close call resolves it on my end. 

If this is a known issue and/or you can reproduce it, I will/can open a separate PR for sendspin_bridge once yours lands.

EDIT: context: I have a DAC at the other end of the usb out of my laptop that remains 'on' needlessly

iVolt1:
> If this is a known issue and/or you can reproduce it, I will/can open a separate PR for sendspin_bridge once yours lands.
> 
> EDIT: context: I have a DAC at the other end of the usb out of my laptop that remains 'on' needlessly

In HA's hassio_audio environment module-suspend-on-idle isn't loaded so DACs stay powered once played to regardless of stream state. Maybe your pipewire-pulse loads the module.  

The asyncio.shield fix is included in my PR so feel free to test it.

vingerha:
I will verify tomorrow. With me it was partially the local host config of the usb .... but when corked processes stayed open it never shut down too, only with a restart of pipewire (after playing ended) or a shutdown of MA it went down after x seconds. Thanks for the effort so far, as said back tomorrow :)

vingerha:
update: the fix is not working for me, the processes remain active even after they stopped playing minutes ago and (with me) this prevents to suspend the usb, which then continues to provide 5v to the dac and that stays on.
The sendspin_bridge 'fix' (call it a draft version) that I worked on terminates the corked/no processes after a few seconds.


image

```
 pactl list sink-inputs | grep -E "Sink Input #|Corked:"
Sink Input #68
        Corked: no
Sink Input #82
        Corked: no
Sink Input #88
        Corked: no
```

iVolt1:
> update: the fix is not working for me, the processes remain active even after they stopped playing minutes ago and (with me) this prevents to suspend the usb, which then continues to provide 5v to the dac and that stays on. The sendspin_bridge 'fix' (call it a draft version) that I worked on terminates the corked/no processes after a few seconds.

The MA hassio_audio pulse audio plugin does not include suspend_on_idle so that is likely why devices are not going to Idle or Suspend.  It may depend on your linux distribution whether that is included or not.

iVolt1:
@marcelveldt, @MarvinSchenkel Is there anything else besides the Security Check holding back this PR from being merged to dev?  I would really like to see it move forward since it is how I play audio here every day and it seems really solid.  Also local audio is a much requested feature for whole house audio using line outputs with multiple DACS or 5.1/7.1 stereo pairs so I think many users will be happy to see this implemented.  Thanks, iVolt1.

vingerha:
> > update: the fix is not working for me, the processes remain active even after they stopped playing minutes ago and (with me) this prevents to suspend the usb, which then continues to provide 5v to the dac and that stays on. The sendspin_bridge 'fix' (call it a draft version) that I worked on terminates the corked/no processes after a few seconds.
> 
> The MA hassio_audio pulse audio plugin does not include suspend_on_idle so that is likely why devices are not going to Idle or Suspend. It may depend on your linux distribution whether that is included or not.

Nope...it is the processes that stay open, see my screenshot above, as soon as I reset pw (i.e. processes gone) or close MA container (ditto), USB goes down in seconds and after that also my dac suspends. Pipe/pulse is not proprely closing processes so some code needed to push to do so . Alsa devices no problem

vingerha:
> @marcelveldt, @MarvinSchenkel Is there anything else besides the Security Check holding back this PR from being merged to dev? I would really like to see it move forward since it is how I play audio here every day and it seems really solid. Also local audio is a much requested feature for whole house audio using line outputs with multiple DACS or 5.1/7.1 stereo pairs so I think many users will be happy to see this implemented. Thanks, iVolt1.

Ditto, I would like to see this go live too so next steps can be planned.
Wrt to hdmi....on intel there remains an issue playing above 48k, this seems to be a intel<>pw issue (not related to this PR)

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -337,26 +444,31 @@ async def discover_and_register(self) -> None:
...
+                    self.provider,
+                    player_id=device_uuid,
+                    device_name=device_name,
+                    hostapi_index=hostapi_index,
+                    device_index=device.get("index", 0),
+                    pa_sink_name=pa_sink_name,
                 )
                 await self.mass.players.register_or_update(player)
+                # Restore cached volume/mute state from previous session
+                await player.restore_state()
```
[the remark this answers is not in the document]
iVolt1: restore_state() before register_or_update(). Done.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -0,0 +1,269 @@
...
+        desc: str = sink.get("description", name)
+        spec_str: str = sink.get("sample_specification", "")
+        driver: str = sink.get("driver", "")
+        try:
+            parts = spec_str.split()
+            fmt = parts[0]  # e.g. 's32le'
+            channels = int(parts[1].replace("ch", ""))
+            sample_rate = int(parts[2].replace("Hz", ""))
+            bit_depth = int("".join(filter(str.isdigit, fmt.split("le")[0].split("be")[0])))
+        except (IndexError, ValueError):
```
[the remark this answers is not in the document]
iVolt1: The problem is s24-32le — splitting on le gives s24-32, then extracting all digits gives 2432. The correct approach is to parse known PA format strings explicitly.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -0,0 +1,269 @@
...
+        check=False,
+    )
+    if result.returncode != 0:
+        raise RuntimeError(f"pactl exited {result.returncode}: {result.stderr.strip()}")
+
+    sinks = []
+    for sink in json.loads(result.stdout):
+        name: str = sink.get("name", "")
+        desc: str = sink.get("description", name)
+        spec_str: str = sink.get("sample_specification", "")
```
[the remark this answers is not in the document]
iVolt1: device_info["name"] is used as device_name which feeds into get_device_uuid — that's the collision problem. The UUID is generated from desc (human-readable description) but should be generated from name (the stable PA sink name). Fix is to swap name and desc in the dict so "name" holds the PA sink name for UUID generation, and add a separate "description" key for display.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -0,0 +1,269 @@
...
+    else:
+        raise FileNotFoundError(
+            "pactl not found — bundled binary missing and pulseaudio-utils not installed"
+        )
+
+    # Build environment with bundled lib dir and PULSE_SERVER
+    lib_dir = os.path.join(os.path.dirname(__file__), "bin", "lib")
+    existing_ld = os.environ.get("LD_LIBRARY_PATH", "")
+    ld_path = f"{lib_dir}:{existing_ld}" if existing_ld else lib_dir
+    env = {**os.environ, "LD_LIBRARY_PATH": ld_path}
```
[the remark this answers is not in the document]
iVolt1: the logic should prefer system pactl and only fall back to bundled if system isn't available.

music_assistant/providers/local_audio/player.py
```diff
@@ -71,86 +77,79 @@ def __init__(
...
-
-        :param muted: Whether to mute or unmute.
+        if sys.platform != "linux" or not self._pa_sink_name:
+            return
+        loop = asyncio.get_running_loop()
+        ok = await loop.run_in_executor(None, self._set_pulse_volume, self._pa_sink_name, 100)
+        if ok:
+            self.logger.debug("PA sink %s set to 100%% hardware volume", self._pa_sink_name)
+        else:
+            self.logger.warning("Failed to set hardware volume for sink %s", self._pa_sink_name)
```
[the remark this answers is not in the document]
iVolt1: Copilot's concern was that setting PA sinks to 100% would affect other applications using the same sinks. That's a valid point — if something else is playing through PA at the same time, its volume would jump to 100% hardware. But in practice:

In a dedicated MA/HA audio setup other apps aren't typically sharing PA sinks
The HA audio stack is designed around MA owning the audio hardware
100% hardware is actually the correct state for software-volume-only operation — you want full hardware range available for the software scaler

So Copilot's concern is theoretically valid but practically irrelevant for the target deployment. Worth a brief note in your PR response acknowledging it but explaining the design intent — hardware at 100% gives software volume the full dynamic range, and the 25% default init keeps actual output at a safe level.

music_assistant/providers/local_audio/__init__.py
```diff
@@ -29,29 +22,13 @@
...
-                ConfigValueOption(title="Software", value=VOLUME_CONTROL_SOFTWARE),
-                ConfigValueOption(title="Disabled", value=VOLUME_CONTROL_DISABLED),
-            ],
-            default_value=VOLUME_CONTROL_HARDWARE,
-            description="Hardware uses OS/ALSA-level volume control. "
-            "Software applies volume scaling to PCM audio data. "
-            "Disabled passes audio at full volume.",
-        ),
-    )
+    return ()
```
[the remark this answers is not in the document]
iVolt1: The hardware ceiling and volume-control mode config entries have been intentionally removed. Volume control is now software-only — the PA sink is set to 100% hardware volume at startup and all day-to-day volume adjustment is done via PCM scaling in the bridge. This gives users a single consistent volume control in the MA UI without exposing audio stack implementation details. The README will be updated to reflect this.

music_assistant/providers/local_audio/player.py
```diff
@@ -71,86 +77,79 @@ def __init__(
...
+        await self._save_state()
         self.update_state()
 
     async def volume_mute(self, muted: bool) -> None:
         """Handle VOLUME_MUTE command."""
         self._attr_volume_muted = muted
-        mode = self.volume_control_mode
-        if mode == VOLUME_CONTROL_HARDWARE:
-            await self._set_hardware_mute(muted)
+        await self._save_state()
```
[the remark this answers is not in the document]
iVolt1: This is intentional for this PR. The provider has been simplified to software-only volume control — all volume and mute changes are applied via PCM scaling in the bridge. The CoreAudio hardware volume/mute path has been removed. macOS support via CoreAudio hardware control can be addressed in a follow-up PR if needed.

music_assistant/providers/local_audio/player.py
```diff
@@ -71,86 +77,79 @@ def __init__(
...
+            provider=self._provider.instance_id,
+            category=CACHE_CATEGORY_PREV_STATE,
+        )
 
     async def volume_set(self, volume_level: int) -> None:
         """Handle VOLUME_SET command."""
         self._attr_volume_level = volume_level
-        if self.volume_control_mode == VOLUME_CONTROL_HARDWARE:
-            await self._set_hardware_volume(volume_level)
+        await self._save_state()
```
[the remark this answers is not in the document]
iVolt1: Intentional for this PR. Volume control has been simplified to software-only — PCM scaling in the bridge handles all volume changes. The CoreAudio hardware volume path has been removed. macOS hardware volume control can be addressed in a follow-up PR if needed.

music_assistant/providers/local_audio/constants.py
```diff
@@ -4,13 +4,16 @@
 
 import uuid
 
-# UUID namespace for generating stable player IDs from device name + host API
-DEVICE_UUID_NAMESPACE = uuid.UUID("a1b2c3d4-e5f6-7890-abcd-ef1234567890")
+# UUID namespace for generating stable player IDs from device name + host API.
```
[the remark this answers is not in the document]
iVolt1: This provider has never shipped PulseAudio support to users, so no existing installations have player IDs derived from any UUID namespace. The namespace is being established here for the first time and will not change after merge. I've added a comment to the constant making this contract explicit as suggested.

vingerha:
I did not update for a week or two but with recent changes on the pactl side, I wanted to test this again.
Issues:
- the container still does not show pactl, I have to amnually install this inside the container. I have the impression that you only test this on HAOS / app.... it needs to also work in a standalone container....and with me it does not
- Maybe the biggest issue is that this PR no longer allows to play via alsa, i.e. the alsa devices are not shown and that is a real issue (for me at least). Context: I cannot use the pw hdmi due to hardward/kernel issues on an pw intel driver, spent many hours on it and it just does not work properly with a 6-channel setup. The hours were spent outside MA scope, trying to play to 6-channel pw hdmi.  I however can make it work using alsa, inside and outside MA ....so that would remain my hdmi go-to player from MA, hence the request to make sure that yours is not filtering them out.

iVolt1:
> I did not update for a week or two but with recent changes on the pactl side, I wanted to test this again. Issues:
> 
>     * the container still does not show pactl, I have to amnually install this inside the container. I have the impression that you only test this on HAOS / app.... it needs to also work in a standalone container....and with me it does not

pactl is installed in the MA dockerfile.base and not by the plugin.  It is up to the docker environment to make pactl available.

> 
>     * Maybe the biggest issue is that this PR no longer allows to play via alsa, i.e. the alsa devices are not shown and that is a real issue (for me at least). Context: I cannot use the pw hdmi due to hardward/kernel issues on an pw intel driver, spent many hours on it and it just does not work properly with a 6-channel setup. The hours were spent outside MA scope, trying to play to 6-channel pw hdmi.  I however can make it work using alsa, inside and outside MA ....so that would remain my hdmi go-to player from MA, hence the request to make sure that yours is not filtering them out.

HA audio does not allow access to the the alsa layer, everything goes through pulse audio.  I have investigated Multichannel digital audio and currently HA audio only allows two channel audio to be passed.  I have a multichannel player in the works that can play 5.1 and 7.1 audio through a surround sound multichannel 5.1 or 7.1 card and possibly a mix of cards.

vingerha:
> pactl is installed in the MA dockerfile.base and not by the plugin. It is up to the docker environment to make pactl available.

Yeah, clear... I will try to simulate this a bit better

> HA audio does not allow access to the the alsa layer

That is incorrect, in its current form (current dev/beta) it does provide the alsa devices (and only alsa devices). Regardless of my hdmi issues, a PR should not take away existing capabilities. I suggest you verify things on container-only as well but I am happy to try and have a look where this can be improved, let me know. 
Wrt to multi-channel, this is a next step for HA, not sure when but in any case, I tried multiple ways to play to PW 6-channel, MPD, Kodi and MA, and they all have issues with the intel driver vs. my end-device (which is apparently a known issue for multi-channel). AMD is supposedly better with the driver but I am not switching machine for that :)

vingerha:
These are a draft version on how it may work, I can play on all devices, alsa is a bit slow to respond but that is also due some local settings....t.b.d. 
[sendspin_bridge.py](https://github.com/user-attachments/files/27631957/sendspin_bridge.py)
[pa_simple.py](https://github.com/user-attachments/files/27631960/pa_simple.py)
image


This is a schematic of the use-case. Over time it is the idea (I hope) that MA will output 6ch ...that would then still need alsa on this device and this would remove the need to run MPD on the side (currently for multi-channel only).

image

iVolt1:
@vingerha The scope of this PR is specifically to expand the existing local_audio provider from exposing a single configured pulse audio output to automatically discovering and registering all PulseAudio sinks as individual MA players including stereo pair sinks that can optionally be automatically created in real-time using my Pulse Audio Stereo Pairs app/addon so that one 8 or 6 channel surround sound cards can play to multiple amplifiers as individual sendspin players. That's already a substantial change. ALSA direct output is a different output path with different conflict/UX implications and deserves separate treatment.

I don't recall if you tried PipeWire device rules to force the correct rate/channel config for you hdmi device as a first step. Getting an alsa bug fixed can be a challenge, but not impossible.

I'd suggest submitting MA ALSA support as a separate PR or even as a new  ALSA player provider once this PR lands so that it could get a proper focused review on its own merits.

@marcelveldt and @MarvinSchenkel I am sure you two are busy, but I feel this PR is solid and ready to move to dev pending any other comments you may have.  As far as I know, there is just the security issue to sign off on and maybe get that plex lint issue that crept into dev resolved.

vingerha:
> ... specifically to expand the existing local_audio provider  ... 

This is exactly what your PR is not (!) doing, it is reducing (!) by taking away the access to alsa devices, the fact that you donot see these does not mean this is non existing. 

> ALSA direct output is a different output path

yes... an existing one

iVolt1:
I had forgotten that the original provider enumerated ALSA devices via sounddevice on Linux and that this PR removes that. However on HA OS with hassio_audio, PA owns all ALSA devices exclusively so ALSA direct access was never very functional there anyway — the primary target environment is unaffected, as well as most Linux distributions, except in cases of hardware driver bugs.  

I still recommend the PulseAudio-only approach for this PR. The core value here is expanding from a single configured audio output to automatically discovering and registering all PulseAudio sinks as individual MA players — that's already a significant capability gain for HA OS and most Linux deployments where PulseAudio or PipeWire-pulse is the standard audio stack.

For ALSA support and broader backend flexibility I'd suggest the MA team consider one of these approaches as a follow-up:

1. A separate local_audio_alsa provider for direct ALSA output and a separate local_audio_pipewire provider — clean separation, users choose explicitly, no duplicate player confusion

2. One local_audio provider with a backend selector config (PulseAudio/PipeWire-pulse, ALSA, or PipeWire native) defaulting to PulseAudio for MA on HA OS, PipeWire if available, else ALSA.

The auto-detection or backend selection question also naturally connects to PipeWire native support which would be a better fit for standalone Linux server deployments than going through the PipeWire-pulse compatibility layer. 

@marcelveldt and @MarvinSchenkel please provide direction on which approach fits the MA architecture best. I am happy to contribute to that follow-up work once this PR lands

vingerha:
Yeah... some architectural decision by the pro's would be great :)  At the moment this only supports linux so alsa/pw are sort the 'standard' options but if it goes to say windows (would it?) or mac (could it?) then it may need a proper basis to work off from.

iVolt1:
Yes, the provider does support macOS — just without hardware volume control now.  I don't know that anyone has tested it.

marcelveldt:
imo the only viable option is option 2. one single "local audio out" provider with a (automatic?) backend selector;

- on Linux choose between pulse audio and alsa
- on macos only portaudio (but macos is not a supported install target but used for dev, so easy for testing)
windows is not supported by MA core

vingerha:
> imo the only viable option is option 2. one single "local audio out" provider with a (automatic?) backend selector;
> 
> * on Linux choose between pulse audio and alsa
> * on macos only portaudio (but macos is not a supported install target but used for dev, so easy for testing)
>   windows is not supported by MA core

If automatic = at-activation .... then fine as long as it is possible to manually override this. Repeating that with my intel machine pipe/pulse is not working fine to mutli-c hdmi, regardless of the streaming software. My hdmi must remain multi-channel as I stream a lot on multi-channel (mpd, previous kodi)

vingerha:
Hey, I see you are progressing...and I am of course happy to help testing. Quick question... in the base setup (alsa-only) the config 'Expose this player to Home Assistant' does not do anything with me, i.e. nothing to do with your PR. Are you aiming to cover this gap too (for pw/pule then) ?

EDIT: @marcelveldt is this a known issue or possibly (likely?) just with me

iVolt1:
I have used these players in HA with no issues and that is really my
preferred use case for controlling players.  The work I did today is
incomplete in regards to some garbled sound so consider it a work in
progress.  I may get back to fixing it tonight or tomorrow.  Thanks for
checking in.

vingerha:
Odd... maybe @marcelveldt will know why this fails in a container only setup. Anyhow.... let me know when I should have a look :) ... I know I an gnagging but I do want this to go live too, despite my hdmi issues :)

iVolt1:
> Odd... maybe @marcelveldt will know why this fails in a container only setup. Anyhow.... let me know when I should have a look :) ... I know I an gnagging but I do want this to go live too, despite my hdmi issues :)

@vingerha I think the alsa local ouput is in pretty good shape now and the HA MA pulse audio support is working well too so please give it a try and provide any feedback you may have.

Recent highlights:

Adds ALSA backend via sounddevice/PortAudio for non-HA Linux and macOS
Adds backend config dropdown (Auto/PulseAudio/ALSA) on Linux
Auto mode tries PA first, falls back to ALSA
Removes direct LocalAudioPlayer registration — players now appear under Universal Player, attributed to Local Audio Out provider via a lightweight protocol player registration
player.py removed (functionality absorbed into bridge)
Software volume scaling handles 16/24/32-bit paths
Single failing test (test_late_join_empty_buffer) is a pre-existing AirPlay timing flake unrelated to this PR

Thanks.

vingerha:
I just spent about an hour or so on it and it looks good, **nice job** !

Setup:
- container based on newest dev/nightly image 2026060407 (4th June)
- shared volume for local_audio and sendspin in which I use your PR local_audio and sendspin files
- verified USB (dac), HDMI and local speaker

Few findings, imo none of these are a showstopper to got dev/beta
 
1. During initial setup I had one oddity during alsa-only, see below. It is not persisten, first I had it for USB and HDMI, then a restart of container seemed to have helped ... no clue why...., then it showed for for local speaker and that disappeared too.
This may very well be my host so not sure if this should get attention but just reporting this, maybe you have a clue.

image

2.  On pw I only see USB and HDMI, no local speaker but this is due to profile/hardware limitations for this(my) host.

iVolt1:
>     1. During initial setup I had one oddity during alsa-only, see below. It is not persisten, first I had it for USB and HDMI, then a restart of container seemed to have helped ... no clue why...., then it showed for for local speaker and that disappeared too.
>        This may very well be my host so not sure if this should get attention but just reporting this, maybe you have a clue.
> 
> 
> image

IIRC, I think I have seen those errors before, usually like you said, when restarting and the prior output was to a player that is missing or changed. Picking the desired player and playing again should work.

iVolt1:
@marcelveldt and @MarvinSchenkel - I think this PR is ready for review and hopefully gets moved to dev soon for feedback from a greater audience.  The two check failures are out of my control and should be easy to resolve.  

Thanks,
iVolt1

OzGav:
If you have resolved the review comments then add info about the resolution to the conversation. I think you might need to REBASE to fix all of the CI problems.

music_assistant/providers/local_audio/player.py
iVolt1: Caching is implemented and ourlined in the README.md.

music_assistant/providers/local_audio/player.py
iVolt1: player.py is no longer used.

music_assistant/providers/local_audio/README.md
iVolt1: README.md explains the current device discovery and format negotiation in depth.

music_assistant/providers/local_audio/sendspin_bridge.py
iVolt1: Caching is implemented. See README.md for details.

music_assistant/providers/local_audio/sendspin_bridge.py
iVolt1: mass.loop is used in multiple places in sendspin_bridge.py and there are no uses of asyncio.get_running_loop.

music_assistant/providers/local_audio/sendspin_bridge.py
iVolt1: The docstring is restored.

music_assistant/providers/local_audio/sendspin_bridge.py
iVolt1: This code was removed.

iVolt1:
> If you have resolved the review comments then add info about the resolution to the conversation. I think you might need to REBASE to fix all of the CI problems.

Review comments are resolved. I've also rebased the branch onto dev and force-pushed. CI is showing the same two issues I don't have access to reolve.

OzGav:
It looks like you have manually modified requirements_all.txt which you should not have done. You need to adjust the manifest to something like the following and run `pre-commit run -—all-files`

`"requirements": ["sounddevice==0.5.5", "pulsectl==<version>; sys_platform == 'linux'"],`

iVolt1:
@OzGav - Thanks for the feedback. To clarify — pulsectl was removed intentionally. As part of this PR's refactor, player.py was deleted (the LocalAudioPlayer class was removed and volume/mute state is now managed directly in the Sendspin bridge). pulsectl was only used in player.py to set PA sink hardware volume to 100% — that functionality was replaced by pure software volume scaling. So pulsectl is no longer a dependency and was correctly removed from manifest.json.
The requirements_all.txt commit (7278de4) was a mistake — I have it reverted now.

vingerha:
FYI, not sure if interested but ... once every 2 days I update too and see if it still works, I try to play as much as possible through the nigtly/dev. Till now all still fine

Issues I have and very (!) likely due to my own device,..so as a FYI only
Switching on alsa, either album or device, sometimes this kills the container. Quite hard to reproduce as if I do the same after container restart then all works fine (of course...). 
My usb dac takes time to start and it needs 3 clicks for the first track to play (after being off)

iVolt1:
> FYI, not sure if interested but ... once every 2 days I update too and see if it still works, I try to play as much as possible through the nigtly/dev. Till now all still fine
> 
> Issues I have and very (!) likely due to my own device,..so as a FYI only Switching on alsa, either album or device, sometimes this kills the container. Quite hard to reproduce as if I do the same after container restart then all works fine (of course...). My usb dac takes time to start and it needs 3 clicks for the first track to play (after being off)

@vingerha  Thanks for the testing.  I have not seen a container self kill before.  The only recent material change I have made is today where the synchronization between players better conforms to the Sendspin standard which made a noticeable improvement in syncing between various local audio devices on the same box.  So please give that a chance if you can.  Thanks again.

vingerha:
> @vingerha Thanks for the testing. I have not seen a container self kill before. The only recent material change I have made is today where the synchronization between players better conforms to the Sendspin standard which made a noticeable improvement in syncing between various local audio devices on the same box. So please give that a chance if you can. Thanks again.

Will try later today, was pre-occupied to make it work with hdmi multi-channel and got pretty close ...no cigar yet :)

iVolt1:
@marcelveldt and @MarvinSchenkel I would like to request another review or update on anything else you think I need to do in order to move this PR forward.  Thanks - iVolt1.

music_assistant/providers/local_audio/constants.py
```diff
@@ -4,13 +4,25 @@
...
+# Audio backend selector
+CONF_AUDIO_BACKEND = "audio_backend"
+AUDIO_BACKEND_AUTO = "auto"
+AUDIO_BACKEND_PULSEAUDIO = "pulseaudio"
+AUDIO_BACKEND_ALSA = "alsa"
+
+# Volume control — software only
 VOLUME_CONTROL_SOFTWARE = "software"
-VOLUME_CONTROL_HARDWARE = "hardware"
-VOLUME_CONTROL_DISABLED = "disabled"
```
marcelveldt, line 16: So, why got the hardware volume control got removed ?
hardware volume control should at all time be preferred over software control

vingerha:
I did some tests yesterday, all still fine but as you are updating again (I have been there :) ) ...let me know if / when I should retest

music_assistant/providers/local_audio/constants.py
iVolt1: Hardware volume control was switched to software volume control when it appeared that it would not work correctly with master sinks and remap sinks.  A new approach has since been taken in order to enable hardware volume control the master sinks in which they are are presented as _multichannel_stereo remap sinks along with the stereo pair remap sinks.  The README.md has a detailed explanation of this.

OzGav: Per the [AI POLICY](https://github.com/music-assistant/.github/blob/51c584e009c4600aecc4845961f4374eb5b962f2/AI_POLICY.md) do not answer questions with AI content

iVolt1:
> I did some tests yesterday, all still fine but as you are updating again (I have been there :) ) ...let me know if / when I should retest

@vingerha  Now would be a good time to test.  Thanks.

vingerha:
PW: there seems to be a mapping issue with the naming
pactl shows my hdmi but MA shows it as Built in Analog Stereo.... which does (!) play to hdmi

Alsa: still crashing issues playing to USB, seems a race condition in _on_bridge_stream_start (sendpsin bridge)...someting for other PR. 

Also on alsa, not (!) on PW, the system cannot handle clicking restart of a track/album well, I did this to provoke the above condition with some small code changes. The behaviour is already there with last PR version (and also with my race-related changes)
```
2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 1104 ms behind schedule
2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 1004 ms behind schedule
2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 923 ms behind schedule
```
Unclear at this moment if this is related to MA code or not but I suspect so bcause when using MPD on alsa I donot have this behaviour.

GUI side, the vol slider showed volume as much lower so I moved it to max, without any effect but for a log statement
````
2026-06-15 11:02:10.849 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.Built-in Audio Analog Stereo] Failed to set hardware volume to 100% for alsa_output.pci-0000_00_1f.3.analog-stereo
```

vingerha:
I have another question / food-for-thought. Playing to USB via PW is smooth so this would be my primary choice...However, HDMI and PW are (on my device) not good when the rates increase, for this alsa is the go-to. The question; have you thought about making both available at the same time? Local/sendspin works differently than (say) wiim when one can select a protocol per device, here it is the device that sets/collects the devices. 
Adding... why not hve 1 device may you ask, this is because for stereo my dac is noticeably better then the receiver but my dac cannot play multi-channel :)

iVolt1:
> I have another question / food-for-thought. Playing to USB via PW is smooth so this would be my primary choice...However, HDMI and PW are (on my device) not good when the rates increase, for this alsa is the go-to. The question; have you thought about making both available at the same time? Local/sendspin works differently than (say) wiim when one can select a protocol per device, here it is the device that sets/collects the devices. Adding... why not hve 1 device may you ask, this is because for stereo my dac is noticeably better then the receiver but my dac cannot play multi-channel :)

That is an interesting use case for a future PR.  Wireplumber would need to configured to ignore devices of your choosing and the local_audio provider would need some work on it to enumerate, deduplicate, and run per-device backends of pulse audio or alsa.

iVolt1:
> PW: there seems to be a mapping issue with the naming pactl shows my hdmi but MA shows it as Built in Analog Stereo.... which does (!) play to hdmi
> 

See the README.md for more information on PA and alsa sink enumeration.

> Alsa: still crashing issues playing to USB, seems a race condition in _on_bridge_stream_start (sendpsin bridge)...someting for other PR.
> 

I did some hot plugging of a USB device on my linux mint box and played and did not crash anything.

> Also on alsa, not (!) on PW, the system cannot handle clicking restart of a track/album well, I did this to provoke the above condition with some small code changes. The behaviour is already there with last PR version (and also with my race-related changes)
> 

I have no issues with play/pause/prev/next and seek  playing to an alsa device on my linux mint box.

> ```
> 2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 1104 ms behind schedule
> 2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 1004 ms behind schedule
> 2026-06-15 09:58:59.230 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.D90 III SABRE: USB Audio] Dropping late chunk for D90 III SABRE: USB Audio (hw:1,0): 923 ms behind schedule
> ```

This is warning of  a catch-up flush as each successive chunk is one chunk-duration closer to real-time until the bridge is in sync.

> 
> Unclear at this moment if this is related to MA code or not but I suspect so bcause when using MPD on alsa I donot have this behaviour.
> 
> GUI side, the vol slider showed volume as much lower so I moved it to max, without any effect but for a log statement
> 

The previous linear volume taper was too aggressively loud at the low end so a different taper has been applied.  The README.md has more information.

> ```
> 2026-06-15 11:02:10.849 WARNING (MainThread) [music_assistant.Local Audio Out.bridge.Built-in Audio Analog Stereo] Failed to set hardware volume to 100% for alsa_output.pci-0000_00_1f.3.analog-stereo
> ```

This warning was from a timeout on the half second time allowed when applying a hardware volume change, possibly due to the extra time this can take on pipewire-pulse.  It likely succeeded on the next attempt.

> ```

Thank you for the feedback.

music_assistant/providers/local_audio/constants.py
```diff
@@ -2,15 +2,72 @@
...
+AUDIO_BACKEND_AUTO = "auto"
+AUDIO_BACKEND_PULSEAUDIO = "pulseaudio"
+AUDIO_BACKEND_ALSA = "alsa"
+
+# Volume control mode.
+# HARDWARE: PA sink hardware volume via PAVolumeController (libpulse direct
+#   calls, no subprocess) — used automatically for the PulseAudio/PipeWire
+#   backend, including module-remap-sink.c sinks (each has an independent
+#   hardware volume that does not affect its master sink or siblings).
+# SOFTWARE: numpy-based PCM scaling — used for the ALSA/sounddevice backend,
```
marcelveldt, line 30: no hardware volume control for alsa backend ?

marcelveldt:
@iVolt1 can you rebase your branch and fix the merge conflict ?

vingerha:
Just to try and keep the spirits up ... with your code I have today been able to stream multi-channel over pa and alsa, already added a PR for aisosendspin. Not sure if this helps you (!? do you use mc?) but knowing the mc-scene and the power of MA, this is a great pth ... until my PR gets all above comments lol :)

music_assistant/providers/local_audio/constants.py
iVolt1: The current local_audio provider on nightly only checks for a Master mixer control, defaulting to software volume if it's not found — and it does this lookup on every volume change, which isn't very efficient. It also misses cards whose primary hardware control is named something else (PCM, Speaker, Headphone, etc.), and even where Master does exist, there's no guarantee it's the actual pre-DAC hardware control rather than another digital gain stage.

For this PR I'd like to keep ALSA software-only volume control, and follow up with a separate PR to properly address ALSA hardware volume: a per-player advanced config option letting the user select the actual hardware mixer from a list of mixers enumerated for that player's specific device, rather than guessing a name.

marcelveldt:
Note that @deCodeIt also has some thoughts in #4383 on how to add PA support to the local audio provider.
Maybe the both of you can settle on a final implementation and we can get this over the finish line.

I'm OK if you change the maintainers list to your name(s) because I just added this once as a small experiment but I'm clearly missing a lot of context (and need) to maintain it so this would be better fit with people that are actually using it and experienced with alsa/pulseaudio.

iVolt1:
> Note that @deCodeIt also has some thoughts in #4383 on how to add PA support to the local audio provider. Maybe the both of you can settle on a final implementation and we can get this over the finish line.
> 
> I'm OK if you change the maintainers list to your name(s) because I just added this once as a small experiment but I'm clearly missing a lot of context (and need) to maintain it so this would be better fit with people that are actually using it and experienced with alsa/pulseaudio.

I think my PR is ready to move to the dev branch for more exposure.  I have done all my checks on HA/MA and Linux mint pipewire-pulse and ALSA on both direct line audio and HDMI and everything looks and sounds good.  The primary change of late was to improve the volume taper for an overall maximum volume increase.  All checks have passed except the security checks.

I have updated codeowners to iVolt1 in manifest.json.  Thanks to your team for your patience and guidance with this PR.  It has been a learning curve to get up to speed with MA guidelines, git, vscode, and Claude.  I feel I now have a good workflow for this and future PRs.

deCodeIt:
> Note that @deCodeIt also has some thoughts in #4383 on how to add PA support to the local audio provider. Maybe the both of you can settle on a final implementation and we can get this over the finish line.
> 
> I'm OK if you change the maintainers list to your name(s) because I just added this once as a small experiment but I'm clearly missing a lot of context (and need) to maintain it so this would be better fit with people that are actually using it and experienced with alsa/pulseaudio.

I'll most likely close that PR. Even with the fallback, the playback experience wasn't that of local audio. I experienced failures to start playing media for a couple of times only to work on next retry.

music_assistant/providers/local_audio/provider.py
```diff
@@ -14,6 +22,24 @@ class LocalAudioProvider(PlayerProvider):
...
+            needs_pulse = configured_backend in (AUDIO_BACKEND_PULSEAUDIO, AUDIO_BACKEND_AUTO)
+            if needs_pulse:
+                # Verify libpulse-simple is present before attempting PA output.
+                # On AUTO we probe but don't hard-fail — bridge manager will fall
+                # back to ALSA if pactl returns no sinks at enumeration time.
+                try:
+                    ctypes.CDLL("libpulse-simple.so.0")
+                except OSError:
+                    if configured_backend == AUDIO_BACKEND_PULSEAUDIO:
+                        raise RuntimeError(
```
OzGav: Use an MA error model (SetupFailedError / UnsupportedSystemError) so the provider loader handles it correctly

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -173,80 +325,326 @@ def _on_bridge_stream_start(self) -> None:
...
+                        self.mass.create_task(_reapply_hardware_volume_once_active())
+
+        except asyncio.CancelledError:
+            pass
+        except OSError as err:
+            self.logger.error("PA stream error for %s: %s", self.pa_sink_name, err)
+        finally:
+            self._is_streaming = False
+            if write_future is not None:
+                with suppress(Exception):
```
OzGav: Comment for here and below at L613 we don't generally want broad exception handling like this. I appreciate you may see this in the legacy code but we want to move away so actual bugs aren't hidden. OSError would probably be appropriate.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -0,0 +1,723 @@
...
+                "index": idx,
+                "hostapi": dev.get("hostapi", 0),
+            }
+        )
+    return devices
+
+
+def enumerate_pa_sinks() -> list[dict[str, Any]]:
+    """
+    Enumerate stereo-capable PulseAudio sinks via pactl JSON output.
```
OzGav: This is a general comment thoughout but is particualrly notable here. Use Sphinx-style docstrings with :param: syntax. For simple functions, a single-line docstring is fine. Multi line docstrings should have the first line on its own line to maintain indentation with the following. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally

OzGav (commented):
Just a few points to tidy up from me.

music_assistant/providers/local_audio/provider.py
iVolt1: Fixed by Claude and pushed.

music_assistant/providers/local_audio/sendspin_bridge.py
iVolt1: Fixed by Claude and pushed.  Thanks for the review.

music_assistant/providers/local_audio/pa_simple.py
iVolt1: Fixed by Claude and pushed.

music_assistant/providers/local_audio/constants.py
marcelveldt: Yes, please do so as I'd rather not introduce a software volume control for no reason

marcelveldt (approved):
Thanks, @iVolt1 !

iVolt1:
@marcelveldt Should I be the one to contribute documentation should this PR get moved to Beta?

marcelveldt:
> @marcelveldt Should I be the one to contribute documentation should this PR get moved to Beta?

Yes

OzGav:
@iVolt1 We didn't have any docs for this so I have generated a page based on the code but would appreciate you having a look and advising on any changes required. Thanks
https://beta.music-assistant.io/player-support/local-audio/

iVolt1:
@OzGav I read the doc and had my Claude instance compare it to the one I was working on and your version is quite good, more user focused, and I see no issues worth mentioning.   Thanks for putting that together and moving the code to beta.

PR #3727, opened 2026-04-17
Migrate loudness analyzer to audio analysis provider

marcelveldt:
## Problem
Volume-loudness detection was the last analyzer still hand-rolled in the streams and music controllers, storing measurements in its own `loudness_measurements` table. Every other analyzer already uses the `AudioAnalysisProvider` pattern. Keeping loudness separate meant duplicated plumbing, a bespoke DB table, and no reuse of the PCM chunk fan-out the new controller already does.

## Changes
- New builtin `loudness_analysis` audio analysis provider (always enabled) - ebur128 on live PCM via the shared audio-analysis fan-out
- Removes `attach_loudness_analyzer` from the streams controller and `set_loudness` / `get_loudness` from the music controller; `filesystem_local` and `opensubsonic` now use `mass.streams.audio_analysis.set_track_loudness(...)`
- Adds `loudness_album` field to `AudioAnalysisData` so album-level loudness from tags survives the unified format
- DB migration 38→39: copies valid rows (`> -50 LUFS`) from `loudness_measurements` into `audio_analysis` under `aa_provider_domain='loudness_analysis'`; drops the old table
- Loudness hydration on `streamdetails` moved to just-in-time in `get_queue_item_stream`, so a measurement completed during a previous play is picked up without restart
- Provider skips analysis when `streamdetails.volume_normalization_mode == DISABLED` (respects per-player opt-out); the base-class version gating already skips tracks that already have a measurement
- Optional nightly background task (default on): batches 250 local-filesystem tracks without a loudness row, runs ebur128 directly on the file path, 2s spacing, aborts if the storage provider goes offline mid-run
- Opt-in `write_replaygain_tags` config (default off): writes `REPLAYGAIN_TRACK_GAIN` back to the file (ID3/MP4/Vorbis/APEv2) via a new `write_replaygain_track_gain` mutagen helper

### DIFF
--- a/music_assistant/controllers/streams/audio_analysis.py
+++ b/music_assistant/controllers/streams/audio_analysis.py
@@ -7,9 +7,15 @@ import contextlib
 from math import inf
 from typing import TYPE_CHECKING
 
-from music_assistant_models.enums import MediaType, ProviderType
+from music_assistant_models.background_task import TaskSchedule
+from music_assistant_models.enums import MediaType, ProviderType, StreamType
 
-from music_assistant.constants import DB_TABLE_AUDIO_ANALYSIS, LOUDNESS_MEASUREMENT_MIN_LUFS
+from music_assistant.constants import (
+    DB_TABLE_AUDIO_ANALYSIS,
+    DB_TABLE_PROVIDER_MAPPINGS,
+    LOUDNESS_MEASUREMENT_MIN_LUFS,
+)
+from music_assistant.helpers.datetime import local_clock_time_to_utc
 from music_assistant.helpers.json import json_dumps, json_loads
 from music_assistant.models.audio_analysis import AudioAnalysisData
 from music_assistant.models.audio_analysis_provider import AudioAnalysisProvider
@@ -17,6 +23,15 @@ from music_assistant.models.music_provider import MusicProvider
 
 CHUNK_PROCESS_TIMEOUT = 0.8
 LOUDNESS_ANALYSIS_DOMAIN = "loudness_analysis"
+BACKGROUND_SCAN_TASK_ID = "audio_analysis_background_scan"
+BACKGROUND_SCAN_BATCH_SIZE = 250
+BACKGROUND_SCAN_SLEEP_BETWEEN_ITEMS = 2.0
+# providers whose tracks can be analyzed from their local filesystem path
+FILESYSTEM_PROVIDER_DOMAINS: tuple[str, ...] = (
+    "filesystem_local",
+    "filesystem_smb",
+    "filesystem_nfs",
+)
 
 if TYPE_CHECKING:
     from music_assistant_models.media_items import AudioFormat
@@ -39,6 +54,17 @@ class AudioAnalysisController:
         self.logger = self.mass.logger.getChild("audio_analysis")
         self._active_sessions: dict[str, set[str]] = {}
         self._workers: dict[str, asyncio.Task[None]] = {}
+
+    def setup(self) -> None:
+        """Register the nightly background scan task."""
+        utc_hour, utc_minute = local_clock_time_to_utc(0, 0)
+        self.mass.tasks.register_scheduled_task(
+            task_id=BACKGROUND_SCAN_TASK_ID,
+            name="Audio analysis — background scan of local files",
+            handler=self._run_background_scan,
+            schedule=TaskSchedule.daily(hour=utc_hour, minute=utc_minute),
+            metadata={"task_domain": "audio_analysis"},
+        )
 
     @property
     def providers(self) -> list[AudioAnalysisProvider]:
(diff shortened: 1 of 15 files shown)


music_assistant/providers/loudness_analysis/provider.py
```diff
@@ -0,0 +1,360 @@
...
+            self.logger.debug(
+                "Loudness measurement for %s discarded: "
+                "%s LUFS is below the reliability threshold (%s LUFS)",
+                session.streamdetails.uri,
+                loudness,
+                LOUDNESS_MEASUREMENT_MIN_LUFS,
+            )
+            return
+
+        analysis = AudioAnalysisData(loudness_integrated=loudness)
```
MarvinSchenkel: Any way we can extract all the loudness related features in this provider?
```
# EBU R128 integrated loudness in LUFS (typical range: -70.0 to 0.0).
loudness_integrated: float | None = None
# EBU R128 loudness range in LU (typical range: 1.0 to 30.0).
loudness_range: float | None = None
# ITU-R BS.1770-4 true peak in dBTP (typical range: -20.0 to +3.0).
true_peak: float | None = None
```

music_assistant/providers/loudness_analysis/provider.py
```diff
@@ -0,0 +1,360 @@
...
+
+    async def _send_eof(self, data: LoudnessSessionData) -> None:
+        """Signal end-of-input to the session's ffmpeg process (idempotent)."""
+        if data.eof_sent:
+            return
+        data.eof_sent = True
+        with contextlib.suppress(Exception):
+            await data.ffmpeg.write_eof()
+
+    async def _analyze_local_files_background(self) -> None:
```
MarvinSchenkel: My idea for this was to move this logic to the `AudioAnalysisController` instead of each individual provider. So the AA Controller can schedule a background task that runs let's say between 00:00-06:00 to do background scanning. That way we can tackle multiple analyses at once. (think beat analysis on local files, loudness extraction etc).

music_assistant/providers/loudness_analysis/provider.py
```diff
@@ -0,0 +1,360 @@
...
+            utc_hour, utc_minute = local_clock_time_to_utc(0, 0)
+            self.mass.tasks.register_scheduled_task(
+                task_id=BACKGROUND_TASK_ID,
+                name="Loudness analysis for local files",
+                handler=self._analyze_local_files_background,
+                schedule=TaskSchedule.daily(hour=utc_hour, minute=utc_minute),
+                metadata={"task_domain": "loudness_analysis"},
+            )
+
+    async def process_pcm_chunk(
```
MarvinSchenkel, line 63: Can we use the function to replace the dynamic normalization? So something like:
- Analyze x- seconds of audio > get a dynamic value
- Set on StreamDetails so we can use it straight away
- ONLY in the finalize(), we will calculate the final values for loudness and only store it there. This will also automatically prevent faulty values from being persisted in the db

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1075,7 +1075,7 @@ async def _parse_track(
         # handle (optional) loudness measurement tag(s)
         if tags.track_loudness is not None:
             self.mass.create_task(
-                self.mass.music.set_loudness(
+                self.mass.streams.audio_analysis.set_track_loudness(
```
MarvinSchenkel, line 1078: Can we introduce a new `ProviderFeature.AUDIO_ANALYSIS` for this? I know some music providers (eg Tidal) also have bpm/key info in their API, so it would be nice if we could automatically do this and move the responsibiliy of saving the data to the controller.

music_assistant/providers/loudness_analysis/provider.py
marcelveldt: sounds good

music_assistant/providers/loudness_analysis/provider.py
marcelveldt: No that is not possible

music_assistant/providers/filesystem_local/__init__.py
marcelveldt: No that is not possible

music_assistant/providers/loudness_analysis/provider.py
```diff
@@ -0,0 +1,360 @@
...
+        domains_sql = ", ".join(f"'{d}'" for d in filesystem_domains)
+        query = (
+            f"SELECT pm.item_id AS item_id, pm.provider_instance AS provider_instance "
+            f"FROM {DB_TABLE_PROVIDER_MAPPINGS} pm "
+            f"LEFT JOIN {DB_TABLE_AUDIO_ANALYSIS} aa "
+            f"  ON aa.item_id = pm.item_id "
+            f"  AND aa.provider = pm.provider_instance "
+            f"  AND aa.aa_provider_domain = 'loudness_analysis' "
+            f"  AND aa.media_type = 'track' "
+            f"WHERE pm.media_type = 'track' "
```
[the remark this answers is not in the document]
marcelveldt: Nice catch, you're right — pm.item_id is the library-row integer id while audio_analysis.item_id holds the provider-native id. Fixed + switched to MediaType.TRACK.value.

music_assistant/providers/loudness_analysis/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "audio_analysis",
+  "domain": "loudness_analysis",
+  "name": "Loudness Analysis",
+  "description": "EBU R128 integrated loudness measurement used for volume normalization.",
+  "codeowners": ["@music-assistant"],
+  "requirements": [],
+  "multi_instance": false,
+  "builtin": true,
+  "allow_disable": false,
+  "icon": "volume-variant-high"
```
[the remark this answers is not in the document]
marcelveldt: Good catch, added.

music_assistant/helpers/tags.py
```diff
@@ -1385,3 +1386,64 @@ async def get_embedded_image(input_file: str) -> bytes | None:
...
+        else:
+            return False
+        audio.save()
+        return True
+    except (OSError, PermissionError) as err:
+        LOGGER.debug("could not write replaygain tag to %s: %s", path, err)
+        return False
+    except Exception as err:
+        LOGGER.warning("unexpected failure writing replaygain tag to %s: %s", path, err)
+        return False
```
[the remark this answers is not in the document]
marcelveldt: Added — 4 format roundtrips (MP3/M4A/FLAC/WV) plus missing-file and read-only cases.

music_assistant/providers/filesystem_local/__init__.py
MarvinSchenkel: Care to explain why?

music_assistant/providers/loudness_analysis/provider.py
MarvinSchenkel: Care to explain why?

marcelveldt: dynamic normalization runs at runtime on the live audio, its not an analysis. Its a fallback of the streams controller when there is no measurement (or user wants to use dynamic on purpose such as with radio stations)

music_assistant/providers/filesystem_local/__init__.py
marcelveldt: It will be very hard to standardize on this so I think the best thing for now is to expose some specific methods until we find out how to commonly do this.

music_assistant/providers/loudness_analysis/provider.py
marcelveldt: Good point — now parsing LRA and true peak from the same ebur128 summary and storing them on AudioAnalysisData (both live and file-based path).

music_assistant/providers/loudness_analysis/provider.py
marcelveldt: Agreed, refactored. Background scan now lives on AudioAnalysisController and iterates every AA provider that overrides a new analyze_file(streamdetails) hook. smart_fades etc. can opt in by implementing it — no changes needed yet since its default returns None. Removed the provider-level schedule config.

music_assistant/providers/filesystem_local/__init__.py
marcelveldt: Good call for Tidal/Spotify BPM/key — that's a separate direction (music provider declares AUDIO_ANALYSIS, hands structured data to the controller). I'll tackle it in a follow-up PR so we don't bloat this one; the analyze_file hook I just added is a useful first step toward the same controller-owned pipeline.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -210,6 +238,46 @@ async def get_audio_analysis(
             found = True
         return merged if found else None
 
+    async def set_track_loudness(
```
MarvinSchenkel, line 241: This does not belong here. An AA Provider should return an AudioAnalysisData object with the loudness a music provider (that implement ProviderFeature.AUDIO_ANALYSIS) should return an AudioAnalysisData with the loudness filled in.

marcelveldt: Yes, in a follow up. This is legacy code from filesystem provider.
Also this can be sent in a sync, while already reading the file.

Doing this later at stream time means yet another tag scan so a lot of overhead.

Fine to rethink this but after this PR in a follow up

PR #3734, opened 2026-04-18
Add pactl to base Docker file

iVolt1:
This a prerequisite to Extend Local Audio Out provider with PulseAudio support#3724.  With this PR the pactl binary will be available in the MA container for enumerating pulse audio sinks for the local_audio player provider. With pactl available in the container the pactl binary and libraries will not need to be included with the player provider.


marcelveldt (approved):
Thanks, @iVolt1 !

PR #3736, opened 2026-04-19
Plex: Add extended recommendations with "Mixes For You" support

ajacobson:
I'm a heavy user of Plexamp and using the Music Assistant Plex integration I noticed that a lot of the default recommendations are not available in Music Assistant. Specifically this adds **Mixes For You** (personalized mixes Plex generates from your listening history), **"On This Day"** anniversary releases, and shortcuts to **library playlists** — matching the hub set the official Plex clients show on their home screens.

## Summary
- Add a configurable "extended recommendations" mode that widens the Plex hubs query to include Mixes For You, library playlists, and On This Day — matching the hub set Plex's own clients surface.
- Mixes For You are fetched as raw JSON (bypassing plexapi's broken partial-item parser) and exposed as shuffled playlists.
- A `mix:` item prefix distinguishes mix playlists from regular ones, following the existing `collection:` convention.
- The feature defaults to enabled and can be toggled in advanced config.

## Notes

This feature is default-enabled, but the toggle lives in advanced config for anyone who wants the narrower, pre-existing hub set.

### Why the raw JSON fetch for mixes?

plexapi can't currently materialize Mixes For You hub items — the `<Metadata>` type isn't in its `PLEXOBJECTS` registry, so `_buildItem()` silently drops them and `server.hubs(...)` returns an empty hub.

The provider issues one JSON `GET` against `/hubs/sections/<key>` for this hub and parses the response directly, routing token refresh through the existing helper. Registering the mix type upstream in plexapi is the right long-term fix but is out of scope here — once that lands, the raw fetch can be replaced with a plain `server.hubs()` call.

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import logging
+import random
 import warnings
 from asyncio import Task, TaskGroup
 from collections.abc import Awaitable
@@ -12,6 +13,7 @@ from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar, cast
 
 import plexapi.exceptions
+import plexapi.utils
 import requests
 import urllib3.exceptions
 from music_assistant_models.config_entries import (
@@ -23,6 +25,7 @@ from music_assistant_models.config_entries import (
 from music_assistant_models.enums import (
     ConfigEntryType,
     ContentType,
+    ImageType,
     MediaType,
     ProviderFeature,
     StreamType,
@@ -41,6 +44,7 @@ from music_assistant_models.media_items import (
     ItemMapping,
     MediaItem,
     MediaItemChapter,
+    MediaItemImage,
     MediaItemType,
     Playlist,
     Podcast,
@@ -75,6 +79,7 @@ from music_assistant.providers.plex.constants import (
     CONF_ACTION_GDM,
     CONF_AUTH_TOKEN,
     CONF_COLLECTION_PREFIX,
+    CONF_EXTENDED_RECOMMENDATIONS,
     CONF_HUB_ITEMS_LIMIT,
     CONF_IMPORT_COLLECTIONS,
     CONF_LIBRARY_ID,
@@ -95,6 +100,9 @@ from music_assistant.providers.plex.constants import (
     ERR_NO_ARTIST_FOR_TRACK,
     ERR_TRACK_NOT_FOUND,
     FAKE_ARTIST_PREFIX,
+    MIX_CACHE_EXPIRATION,
+    MIX_ITEM_PREFIX,
+    RECOMMENDATIONS_HUB_PARAMS,
 )
 from music_assistant.providers.plex.helpers import (
     AUDIOBOOK_FEATURES,
@@ -480,6 +488,14 @@ async def get_config_entries(  # noqa: PLR0915
             default_value=10,
             advanced=True,
             range=(1, 100),
+        )
+    )
+    entries.append(
+        ConfigEntry(
+            key=CONF_EXTENDED_RECOMMENDATIONS,
+            type=ConfigEntryType.BOOLEAN,
+            default_value=True,
+            advanced=True,
         )
     )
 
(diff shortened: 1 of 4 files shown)


OzGav:
@anatosun

anatosun:
I will have a look at this PR later this week. Have you had a chance to test the performance of this feature @ajacobson ?

ajacobson:
Performance impact has been negligible in my testing. The net effect is one additional request (the Mixes For You fetch) plus a slightly wider primary hubs response, both gated by the 3-hour recommendations cache. `CONF_HUB_ITEMS_LIMIT` is honored for both paths.

I did notice an issue that trying to play a "mix" from "Recently played" after that mix has been rotated out of the Plex hub would lead to playback errors. I just amended the commit with a fix: when a mix is opened, its raw JSON is cached for 90 days (matching `DB_TABLE_PLAYLOG` retention), so replay works independently of hub state.

Here is what my Discover page looks like with the extended recommendations:
Plex Extended Recommendations

OzGav:
@anatosun Any update on this one?

anatosun:
It looks good and seems to work on my side.

The only thing that makes me a bit hesitant is the raw JSON request + manual parsing for the Mixes. I'm wondering why plexapi doesn't handle music.mixes items properly. Have you (or anyone) opened an issue upstream about this? It might be worth a quick upstream improvement so we don't have to maintain this workaround long-term.

OzGav:
@ajacobson Can you answer the question above?

ajacobson:
I agree an upstream plexapi improvement would be the right long-term fix.

I looked into this while building the PR, but the scope was larger than I wanted to take on here.  plexapi’s hub parser currently materializes hub items through its registered object factory, and the "Mixes For You" hub returns `<Metadata>` entries that are not currently registered/handled.  Fixing it properly would mean teaching plexapi how to parse those "Mixes For You" entries, including their nested centroid artist/Directory children, so callers can get the correct title, key, and artwork.

OzGav:
@anatosun What are your thoughts now on this one?

ajacobson:
I'll take a look at making the upstream changes to plexapi.

ajacobson:
I've opened the upstream plexapi PR: pushingkarmaorg/python-plexapi#1621

It adds the two pieces plexapi was missing for these items:

- `LibrarySection.hubs()` now accepts query parameters (e.g. `includeMyMixes=1`, `count=10`), matching the signature `Library.hubs()` already has — so the mixes hub can be requested through the public API instead of a hand-built key.
- A new `Playlist.centroid` property exposing the nested centroid artist a "Mix For You" is built around. This was the real gap: mix playlists have no `thumb`/`composite` of their own, so their artwork was previously unreachable without touching plexapi internals. The property is guarded so accessing it on partial objects never triggers an auto-reload.

Validated against a live PMS (mixes parse natively as `Playlist` objects with centroid + artwork, existing `hubs()` callers unchanged), with tests included — the mixes hub can't be generated by plexapi's CI bootstrap server, so the test replays a captured response.

Once that's merged and released, the manual parsing here collapses to `library.hubs(includeMyMixes=1, count=n)` + `mix.centroid.thumb`. Happy to handle that however you prefer — as a small follow-up after this merges (pin bump + swapping the call sites), or by holding this PR until the plexapi release and updating it in place.

ajacobson:
The raw-JSON workaround is gone as of 688044c — mixes now parse through plexapi's normal `fetchItems` path on the pinned plexapi (4.18.1), which also drops the second HTTP request per recommendations refresh. The branch is rebased onto current `dev` (constants/strings restructure, `recommendations()` rename).

Remaining work, planned as a small follow-up PR rather than here:

- Upstream [pushingkarmaorg/python-plexapi#1621](https://github.com/pushingkarmaorg/python-plexapi/pull/1621) (merged, not yet released) adds `LibrarySection.hubs(**kwargs)` and `Playlist.centroid`. Once a plexapi release ships with it and the pin bumps, the manual hub-key construction and the `._data` read in `_mix_playlist_fields` collapse to public API: `hubs(count=..., includeMyMixes=True, ...)` and `mix.centroid.thumb`.

anatosun:
Solid work! Thanks for implementing this upstream :)

MarvinSchenkel:
> Solid work! Thanks for implementing this upstream :)

@anatosun The code looks good, is this good to merge?

MarvinSchenkel (approved):
Thanks @ajacobson , looks good 👍

OzGav:
@ajacobson can you also raise a PR against the BETA branch of the docs with any changes needed for the Plex page

ajacobson:
Docs PR is up against `beta`: music-assistant/music-assistant.io#784 — adds an Advanced Settings section covering the `Extended recommendations` toggle.

PR #3739, opened 2026-04-19
Last FM: Add one-click auth

duanyutong:
### Background

https://discord.com/channels/753947050995089438/1487828488277983413

### Changes

- Use built-in Last.fm API credentials (`app_var(12)`, `app_var(13)`) so users no longer need to register their own API application. For Last.fm, the "Authorize" button now works out of the box with zero configuration.
- Move the API key and shared secret config fields to advanced settings. They remain available for Libre.fm users (who must provide their own) or anyone who wants to override the built-in key.
- Add `_resolve_credentials()` helper that transparently falls back to built-in defaults when no custom credentials are provided. Validates that both key and secret are provided together.
- Introduce `_NetworkProvider` Enum for type-safe provider matching, using `.value` at serialization boundaries to remain compatible with the config system.
- Type `network` as `pylast._Network | None`: because built-in credentials now let the provider load without a session key, `handle_async_init` can return early without setting `self.network`. `loaded_in_mass` checks for `None` and returns early, letting the plugin load gracefully while unauthenticated.
- Fix incorrect "listenbrainz client" comment (copy-paste from ListenBrainz scrobbler) and add proper Sphinx-style docstrings throughout.

No breaking changes.
Existing configurations with user-provided API keys continue to work -- `_resolve_credentials` prefers user-provided values when present.

### Manual testing performed

Commands:

```sh
uv run pytest tests/core/test_scrobbler.py -v
uv run python -m music_assistant --log-level debug
```

Screenshot 2026-04-19 at 1 13 21 pm

Steps:

1. Open `http://localhost:8095` in a browser
2. Go to Settings > System > Webserver and ensure **Base URL** is set to `http://localhost:8095` (the URL must match how your browser reaches the server, so the auth callback is reachable)
3. Go to Settings > Providers > Add Provider > LastFM Scrobbler
4. Leave the default provider "Last.FM" selected
5. Do not expand advanced settings or enter any API key/secret
6. Click "Authorize with lastfm"
7. Verify you are redirected to the Last.fm website authorization page
8. Authorize the application on Last.fm
9. Verify you are redirected back and see "Successfully logged in as \<username\>"
10. Click Save
11. Play a track and verify it appears in your Last.fm listening history
12. Check your Last.fm account under Settings > Applications -- "Music Assistant" should appear as an authorized application
13. Check that Last.fm Now Playing status is updated
14. Check that Last.fm scrobbling history is updated

#### Backward Compatibility with existing user-provided API key

1. If you have an existing Last.fm scrobbler instance with a custom API key/secret, reload the provider
2. Verify the existing session key still works and scrobbling continues
3. Verify the API key/secret fields appear under advanced settings with your stored values

#### Libre.fm with custom credentials

1. Add a new LastFM Scrobbler instance
2. Change the Provider dropdown to "LibreFM"
3. Verify that the API Key and Shared Secret fields are now marked as required
4. Attempt to click "Authorize with librefm" without entering credentials
5. Verify it fails with a clear error about credentials being required for Libre.fm
6. Enter valid Libre.fm API credentials and authorize
7. Verify scrobbling works against the Libre.fm endpoint

### DIFF
--- a/music_assistant/providers/lastfm_scrobble/__init__.py
+++ b/music_assistant/providers/lastfm_scrobble/__init__.py
@@ -1,10 +1,11 @@
 """Allows scrobbling of tracks with the help of PyLast."""
 
 import asyncio
+import enum
 import logging
 import time
-from collections.abc import Callable
-from typing import TYPE_CHECKING, cast
+from collections.abc import Callable, Mapping
+from typing import Final, cast
 
 import pylast
 from music_assistant_models.config_entries import (
@@ -20,6 +21,7 @@ from music_assistant_models.playback_progress_report import MediaItemPlaybackPro
 from music_assistant_models.provider import ProviderManifest
 
 from music_assistant.constants import MASS_LOGGER_NAME
+from music_assistant.helpers.app_vars import app_var  # type: ignore[attr-defined]
 from music_assistant.helpers.auth import AuthenticationHelper
 from music_assistant.helpers.scrobbler import (
     ScrobblerConfig,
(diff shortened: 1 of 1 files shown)


music_assistant/providers/lastfm_scrobble/__init__.py
```diff
@@ -30,15 +32,81 @@
...
 
-SUPPORTED_FEATURES: set[ProviderFeature] = (
-    set()
-)  # we don't have any special supported features (yet)
+# Built-in Last.fm API credentials (not available for Libre.fm)
+_DEFAULT_API_KEY: str = app_var(12)
+_DEFAULT_API_SECRET: str = app_var(13)
+
+
+# This plugin is purely event-driven (listens for MEDIA_ITEM_PLAYED)
```
MarvinSchenkel: Please remove this AI comment as it is too verbose

music_assistant/providers/lastfm_scrobble/__init__.py
```diff
@@ -54,16 +122,20 @@ async def setup(
 class LastFMScrobbleProvider(PluginProvider):
     """Plugin provider to support scrobbling of tracks."""
 
-    network: pylast._Network
+    # The ``network`` attribute is None until
```
MarvinSchenkel: Please remove this AI comment as it is too verbose

MarvinSchenkel (changes requested):
Nice job! Just a few small nits to make the code more readable. Should be good to merge after that 👏

music_assistant/providers/lastfm_scrobble/__init__.py
duanyutong: Done.

music_assistant/providers/lastfm_scrobble/__init__.py
duanyutong: I actually wrote this by hand. Because this is a deferred mutable classvar in the absence of a init method that guarantees instance methods, it is changed to an optional, union type, which calls into question if and how it is populated and where we do and do not need the null guard.

Are you sure you want this removed? I thought it was confusing without these comments.

MarvinSchenkel: Check, if it confused you then let's leave it like this for clarity purposes.

duanyutong: I changed the style here a bit, following the existing `_on_unload` which is an instance method guaranteed by the init hook appearing like a ClassVar. Removed the comment since the logic in the init hook is readable enough now.
The only thing that needs some doc would be the init hook being guaranteed to get called for any subclass.

MarvinSchenkel (approved):
Looks good, thanks @duanyutong !

OzGav:
@duanyutong Please raise a PR against the beta branch of our docs repo capturing this new functionality. Thanks!

duanyutong:
@OzGav Done: https://github.com/music-assistant/music-assistant.io/pull/667. Thanks for the note.

PR #3740, opened 2026-04-20
Add iTunes artwork metadata provider

OzGav:
Adds a new metadata provider that can fetch high-resolution (1500x1500) album artwork from the iTunes catalog using UPC barcode lookup.

Captures `barcode` on MusicBrainzReleaseGroup so release-group lookups (including the new radio-stream artwork flow) can propagate the barcode to this provider. Adds the barcode external ID onto the temp album in MetaDataController._get_release_group_artwork so iTunes is checked for radio streams.


MarvinSchenkel (approved):
Looks good to me, thanks @OzGav

PR #3743, opened 2026-04-20
Preserve multi-value album type across all tag parsers

OzGav:
The MusicBrainz Album Type tag can carry multiple values (primary plus secondary types, e.g. "album" + "live"). The Vorbis, APEv2, and MP4 parsers were each reading only the first value, so secondary types like "live" were silently dropped and albums never received AlbumType.LIVE. The ID3 parser wasn't reading the tag at all.

### DIFF
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -1118,12 +1118,8 @@ def _parse_vorbis_tags(tags: VCommentDict) -> dict[str, Any]:
         result["compilation"] = compilation
 
     # album type may be multi-value (repeated fields); join them
-    if albumtypes := _vorbis_get_multi(tags, "MUSICBRAINZ_ALBUMTYPE"):
-        result["musicbrainzalbumtype"] = ";".join(albumtypes)
-    # RELEASETYPE is an alternative tag name for the same field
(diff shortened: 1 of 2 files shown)


music_assistant/helpers/tags.py
```diff
@@ -971,6 +971,9 @@ def _parse_id3_tags(tags: dict[str, Any]) -> dict[str, Any]:
         result["musicbrainzalbumartistid"] = tags["TXXX:MusicBrainz Album Artist Id"].text
     if "TXXX:MusicBrainz Artist Id" in tags:
         result["musicbrainzartistid"] = tags["TXXX:MusicBrainz Artist Id"].text
+    # album type may be multi-value; join them
+    if "TXXX:MusicBrainz Album Type" in tags:
```
MarvinSchenkel: We should guard against empty list here:
```
if text := tags["TXXX:MusicBrainz Album Type"].text:
    result["musicbrainzalbumtype"] = ";".join(text)
```

OzGav: That wasnt the only one upon review so guards added on all tags

MarvinSchenkel (approved):
Thanks @OzGav !

PR #3745, opened 2026-04-20
[Soundcloud]: improving search

fionn-r:
Fixes an issue where the total_duration and duration are almost the same but not quite - but are still not previews.

E.g. for NTS Radio Breakfast w/ Flo, all the epsiodes are very long and the total duration is often different to the duration by 5ms so they never appear in search results (if you need me to add more details to explain this issue I can, but it's definitely a problem with the response from the `soundcloudpy` library.

The URLs aren't url-encoded by the `soundcloudpy` library so I've also wrapped them in an encoding call.

I have also removed the unused searchtypes, this functionality at some point was replaced with the if statement lower down but never removed.


music_assistant/providers/soundcloud/__init__.py
```diff
@@ -127,29 +127,25 @@ async def search(
...
         for item in searchresult["collection"]:
             media_type = item["kind"]
             if media_type == "user" and MediaType.ARTIST in media_types:
                 result.artists = [*result.artists, await self._parse_artist(item)]
             elif media_type == "track" and MediaType.TRACK in media_types:
-                if item.get("duration") == item.get("full_duration"):
-                    # skip if it's a preview track (e.g. in case of free accounts)
+                duration = item.get("duration", 0)
+                full_duration = item.get("full_duration", 0)
+                if full_duration == 0 or abs(duration - full_duration) < 5000:
```
MarvinSchenkel: In your PR description you mentioned `5ms`, this is `5s`. Is that intentional? If so, why 5s? It seems very generous. Also, this should be added to a constant with a description and then used here.

MarvinSchenkel: Why would `full_duration` ever be 0? Please add a comment why we need this check

MarvinSchenkel (changes requested):
Few questions about the implementation

music_assistant/providers/soundcloud/__init__.py
fionn-r: Understood about the constant, will make the change.

The reason I made it more generous is because I don't know what other Soundcloud tracks this affects. It's possible that the difference scales with how long the track is.

The previous for comparing the durations was "# skip if it's a preview track (e.g. in case of free accounts)", I guess the preview track thing is like iTunes back in the day, though I'm not sure.

My guess was that if it's something like that, the difference in length is going to be longer than 5 seconds and if it's less than 5 seconds, then include it in the search results.

That being said, I think I'll drop it to 1s since then it's a rounding amount.

fionn-r: If there's anything that goes wrong with getting the `full_duration` value from the `item` dict, it will be returning 0 so I was thinking it was just better to include that in the results anyway. If the ethos is more "better safe than sorry" I'm happy to drop it

MarvinSchenkel:
Just a bit of CI, other than that this should be good to merge now :)

fionn-r:
Great, thanks for being supportive and thorough!

MarvinSchenkel (approved):
Thanks @fionn-r 🎉

PR #3750, opened 2026-04-20
AirPlay 2 provider now supports sync

bradkeifer:
cliap2 now functions similarly to cliraop by dropping audio packets that are received too late to meet the playback timing requirements.
Additionally, the excessive log messages from the cliap2 mass_timer_cb() function have been migrated to SPAM/VERBOSE level, so this specific case mapping can be removed from MA code.


MarvinSchenkel (approved):
AWESOME JOB! 🙌 

Looks good to me.

PR #3765, opened 2026-04-21
Refactor Pandora provider to use dynamic playlists

TermeHansen:
This re write of the Pandora provider will make use of the new dynamic playlist features to simplify some of the code.

Pandora pulls 4 tracks in a fragment, that can be played as much as one wants and in any order, but only those 4 tracks, as soon as one pull a new fragment of 4 tracks they can not be played anymore.

This refactor gives some nice new user experiences:
- Seeking Support 
- Track Visibility
  -  ~4 upcoming tracks are constantly shown in the dynamic playlist, given the knowledge of what comes next and the option to skip.
  - History of played tracks can be seen


TermeHansen:
@dmoo500 please have a look here, it is still somewhat work in progress with many warning prints.

I added the reading of the authtoken from a local file, as I get flagged by pandora from all my logins and can not login again.

music_assistant/providers/pandora/provider.py
```diff
@@ -21,20 +23,26 @@
...
 )
 from music_assistant_models.media_items import (
+    Album,
+    Artist,
     AudioFormat,
+    BrowseFolder,
+    ItemMapping,
     MediaItemImage,
-    MediaItemMetadata,
+    MediaItemType,
```
dmoo500, line 32: [PROBLEM] If the queue contains an old-style Radio ID (e.g. `"12345"` without an underscore), `split("_")` raises a `ValueError` and the entire cleanup loop aborts — so stale queues are never cleared. Add a guard before calling `_find_track`:
```python
if "_" not in item.media_item.item_id:
    self.mass.player_queues.clear(item.queue_id)
    continue
```

music_assistant/providers/pandora/provider.py
```diff
@@ -114,34 +115,30 @@ async def handle_async_init(self) -> None:
...
         await self.close()
         await super().unload(is_removed)
 
     async def _authenticate(self, username: str, password: str) -> None:
         """Authenticate with Pandora and get auth token."""
         try:
             self._csrf_token = await get_csrf_token(self.http_session)
+            if self._auth_token:
+                self.logger.warning(f"Using existing auth token: {self._auth_token}")
+                return
```
dmoo500, line 134: [CRITICAL] The auth token is written to the log in plain text — this is a security issue. Remove the token value from the log message:
```python
self.logger.warning("Using existing auth token from file")
```

music_assistant/providers/pandora/provider.py
```diff
@@ -297,83 +306,226 @@ async def get_library_radios(self) -> AsyncGenerator[Radio, None]:
...
+
+    async def get_playlist(self, prov_playlist_id: str) -> Playlist:
+        """Get full playlist details by id."""
+        self.logger.warning(f"get playlist {prov_playlist_id}")
+        async for station in self._get_stations():
+            if station.item_id == prov_playlist_id:
+                self.logger.warning(f"returning playlist station {station.name}")
+                return station
+        return self._parse_station({"stationId": prov_playlist_id})
+
```
dmoo500, line 351: [CRITICAL] `_parse_station` accesses `station["name"]` unconditionally, but the fallback dict `{"stationId": prov_playlist_id}` has no `"name"` key — this raises a `KeyError`. Either raise `MediaNotFoundError` instead, or provide a default:
```python
raise MediaNotFoundError(f"Playlist {prov_playlist_id} not found")
```

music_assistant/providers/pandora/provider.py
TermeHansen: this will not get in, this is just something I have now for sideloading the authtoken so I don't have to do login so many times and get flagged by Pandora servers...

music_assistant/providers/pandora/provider.py
TermeHansen: I resolved it by making a guard for "_" in _find_track() function.

music_assistant/providers/pandora/provider.py
TermeHansen: I removed this self._auth_token print now

music_assistant/providers/pandora/provider.py
TermeHansen: done

MarvinSchenkel:
If this is ready for review, please mark the PR as such and we will have a look

TermeHansen:
No I'm still testing based on inputs from Discord

TermeHansen:
@MarvinSchenkel I think it is ready now for review :)

TermeHansen:
@OzGav can you have a look?

OzGav:
Unfortunately I cant test it but having a quick look all I do is ask how these things perform:

* You have indicated seek is supported so seeking mid track to further along in the track?
* Pause for a couple of minutes, does it resume OK?
* Can you skip back to the previous track?
* Did you check memory usage after a long listening session?
* Did you play 2 or more simultaneous streams from the same station?

I am assuming you have stress tested it but just thought I'd ask...

TermeHansen:
@OzGav, sorry I didn't respond, but I ended up looking alot into the memory usage by following the music_assistant python process in linux, and it seems to me that alot of the memory build up comes from the playlist usage.... But lets start with your list of things:

- You have indicated seek is supported so seeking mid track to further along in the track?
-- Yes, works perfectly fine.
- Pause for a couple of minutes, does it resume OK?
-- Yes, also works fine.
- Can you skip back to the previous track?
-- In the code we pop the song from the cache as soon as the last chunk has been taken, so no.
- Did you play 2 or more simultaneous streams from the same station?
-- Yes also working fine.
- Did you check memory usage after a long listening session?
First of I made the code so we only cache up to 10 songs (could be an advanced setting), and we will pop the oldest when adding a new "11th" song.

I added some debug output to follow just the cache locally:
```
2026-05-20 14:49:16.216 WARNING (MainThread) [music_assistant.Pandora] Popped song: 1.64MB,  cache size: 18.44MB (8 songs)
2026-05-20 14:52:49.088 WARNING (MainThread) [music_assistant.Pandora] Popped song: 1.67MB,  cache size: 20.56MB (9 songs)
2026-05-20 14:53:14.145 WARNING (MainThread) [music_assistant.Pandora] Popped song: 1.42MB,  cache size: 19.14MB (8 songs)
2026-05-20 14:56:50.303 WARNING (MainThread) [music_assistant.Pandora] Popped song: 2.90MB,  cache size: 12.75MB (6 songs)
2026-05-20 14:56:58.065 WARNING (MainThread) [music_assistant.Pandora] Popped song: 3.49MB,  cache size: 12.75MB (6 songs) 

```

So that seems to work, and not that much extra space used, but then I monitored the python process and I'm not sure why so much memory builds up.
image
Using dynamic playlist with local mp3 seems fine. "mp3" I played local mp3's and then started playing Pandora. So now I'm not so sure if it actually building up memory because of my code, or because of my usage of MA.

Is this expected, and if not how can I debug this further?

TermeHansen:
@OzGav when starting playback again after a longer pause I see this error in my log:

```
2026-05-22 09:48:51.350 ERROR (MainThread) [music_assistant] Error doing task: Task exception was never retrieved
  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/players/helpers.py", line 132, in execute
    await fn(self, *args, **kwargs)
  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/players/controller.py", line 1129, in enqueue_next_media
    await self._handle_enqueue_next_media(player_id, media)
  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/players/controller.py", line 3411, in _handle_enqueue_next_media
    raise UnsupportedFeaturedException(
        f"Player {player.state.name} does not support enqueueing"
    )
music_assistant_models.errors.UnsupportedFeaturedException: Player osmc4_sendspin does not support enqueueing

  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/player_queues.py", line 2269, in _enqueue_next_item_on_player
    await self.mass.players.enqueue_next_media(
    ...<2 lines>...
    )
  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/players/helpers.py", line 138, in wrapper
    await execute()
  File "/home/dietpi/HA/repos/MA/music_assistant/controllers/players/helpers.py", line 134, in execute
    raise PlayerCommandFailed(str(err)) from err
music_assistant_models.errors.PlayerCommandFailed: Player osmc4_sendspin does not support enqueueing

```

Is this a bug I have run into?

OzGav:
I am away for a a little bit so hopefully one of the others will be able to look

TermeHansen:
no worries @OzGav 

I just let a player play pandora continuously over the last 24 hours, and the increase again around 14 is due to that I started listening as well on another player simultaneously (on same station, so that for sure works fine).
image
But this usage of soon 3Gb memory seem crazy, especially when the cache dict holds less than 20Mb data. It would be nice to hear if this is a know thing or it is something to do with my implementation.

TermeHansen:
pandora

OzGav:
So are you saying you starting playing Pandora at time 0 and that is the result. Did you try with the existing stable server to see what happens with that Pandora provider?

OzGav:
Also can you try setting `can_seek=False` and `allow_seek=False` and see if that improves anything

TermeHansen:
@OzGav I tried without seek (red line) with no substantial difference, and the old pandora without a playlist and a history of played numbers is not having this memory build up. Interestingly it seems that turning off "volume normalization" for the player I use seems to reduce this memory build up to close to what is seen with the old provider.
image

I will let it run a day or two more, and then switch normalization on again and see if that indeed was the thing that changed now for my long runs build up memory usage...

OzGav:
Very interested in what you find. I will loop in the experts if you confirm it is the audio analysis (loudness) pipeline

chrisuthe:
I can add to the discussion a bit @TermeHansen...

The old pandora provider never enqueued discrete tracks, so it never tripped the per-track analysis trigger. (streams are treated differently, we can't/don't want to analyze them for loudness, etc)

The new rewrite turns every track into a normal MA queue item (way better, IMHO) but that means that we fire start_analysis on that track, with volume normalization on that means an additonal ffmpeg subprocess per track (plus we'll launch and do the pytorch based work if smart_fades and sonic_analysis are on). 

That accounts for... some... of what you're seeing but not an uncontrolled rise, and not 3gb of worth of memory growth imho.

Can you watch the number of ebur128 sessions (volume checking) with something like 
`while sleep 60; do printf '%s ebur128=%s ffmpeg=%s ma_rss_mb=%s\n' "$(date +%H:%M:%S)" "$(pgrep -fc ebur128)" "$(pgrep -fc ffmpeg)" "$(ps -o rss= -p "$(pgrep -f music_assistant | head -1)" 2>/dev/null | awk '{print int($1/1024)}')"; done | tee -a /tmp/ma_aa_watch.log`

That should help narrow down if it's the AA bits.

We can also consider behavior based misses on cleanup like rapid skipping, or something like that which would start sessiosn and perhaps not clean them up properly....

TermeHansen:
OK getting closer now. Volume normalization didn't change anything more or less, but I had also  disabled "Smart fades" in "Audio Analysis" even though it actually also stands as "disabled [default]" in the player I saw after so I thought that didn't make any difference - but that make this big memory usage difference...

Now I enabled it again in "Audio analysis" but kept it disabled in the player and now it runs up high in memory usage (yellow line).
Screenshot_2026-06-04_21-45-55

TermeHansen:
@chrisuthe 

```
22:58:16 ebur128=1 ffmpeg=3 ma_rss_mb=1711
22:59:16 ebur128=0 ffmpeg=1 ma_rss_mb=1917
23:00:16 ebur128=0 ffmpeg=1 ma_rss_mb=1917
23:01:16 ebur128=0 ffmpeg=1 ma_rss_mb=1917
23:02:16 ebur128=0 ffmpeg=1 ma_rss_mb=2081
23:03:16 ebur128=0 ffmpeg=1 ma_rss_mb=2121
23:04:16 ebur128=0 ffmpeg=1 ma_rss_mb=2121
23:05:17 ebur128=0 ffmpeg=1 ma_rss_mb=2121
23:06:17 ebur128=1 ffmpeg=3 ma_rss_mb=2144
23:07:17 ebur128=0 ffmpeg=1 ma_rss_mb=2233
23:08:17 ebur128=0 ffmpeg=1 ma_rss_mb=2233
23:09:17 ebur128=0 ffmpeg=1 ma_rss_mb=2233
23:10:17 ebur128=0 ffmpeg=1 ma_rss_mb=2234
23:11:17 ebur128=0 ffmpeg=1 ma_rss_mb=2363
23:12:17 ebur128=0 ffmpeg=1 ma_rss_mb=2363
23:13:17 ebur128=0 ffmpeg=1 ma_rss_mb=2363
23:14:17 ebur128=0 ffmpeg=1 ma_rss_mb=2496
23:15:18 ebur128=0 ffmpeg=1 ma_rss_mb=2496
23:16:18 ebur128=0 ffmpeg=1 ma_rss_mb=2496
23:17:18 ebur128=1 ffmpeg=3 ma_rss_mb=2517
23:18:18 ebur128=0 ffmpeg=1 ma_rss_mb=2690
23:19:18 ebur128=0 ffmpeg=1 ma_rss_mb=2711
23:20:18 ebur128=0 ffmpeg=1 ma_rss_mb=2711
23:21:18 ebur128=0 ffmpeg=1 ma_rss_mb=2711
23:22:18 ebur128=0 ffmpeg=1 ma_rss_mb=2799
23:23:18 ebur128=0 ffmpeg=1 ma_rss_mb=2768
23:24:18 ebur128=0 ffmpeg=1 ma_rss_mb=2768
23:25:18 ebur128=0 ffmpeg=1 ma_rss_mb=2874
23:26:19 ebur128=0 ffmpeg=1 ma_rss_mb=2874
23:27:19 ebur128=0 ffmpeg=1 ma_rss_mb=2874
23:28:19 ebur128=0 ffmpeg=1 ma_rss_mb=2920
23:29:19 ebur128=0 ffmpeg=1 ma_rss_mb=2936
23:30:19 ebur128=0 ffmpeg=1 ma_rss_mb=2936
23:31:19 ebur128=0 ffmpeg=1 ma_rss_mb=2936
23:32:19 ebur128=0 ffmpeg=1 ma_rss_mb=2936
23:33:19 ebur128=1 ffmpeg=3 ma_rss_mb=2936
23:34:19 ebur128=0 ffmpeg=1 ma_rss_mb=2964
23:35:19 ebur128=0 ffmpeg=1 ma_rss_mb=2964
23:36:19 ebur128=0 ffmpeg=1 ma_rss_mb=2964
23:37:20 ebur128=0 ffmpeg=1 ma_rss_mb=2964
23:38:20 ebur128=1 ffmpeg=3 ma_rss_mb=2964
23:39:20 ebur128=0 ffmpeg=1 ma_rss_mb=2967
23:40:20 ebur128=0 ffmpeg=1 ma_rss_mb=2967
```

TermeHansen:
just killed the debug logging now this morning, but wow it is just on going upwards...

```
10:22:17 ebur128=0 ffmpeg=1 ma_rss_mb=4293
10:23:17 ebur128=1 ffmpeg=3 ma_rss_mb=4293
10:24:17 ebur128=0 ffmpeg=1 ma_rss_mb=4320
10:25:18 ebur128=0 ffmpeg=1 ma_rss_mb=4320
10:26:18 ebur128=1 ffmpeg=3 ma_rss_mb=4320
10:27:18 ebur128=0 ffmpeg=1 ma_rss_mb=4286
10:28:18 ebur128=0 ffmpeg=1 ma_rss_mb=4287
10:29:18 ebur128=0 ffmpeg=1 ma_rss_mb=4287
10:30:18 ebur128=0 ffmpeg=1 ma_rss_mb=4287
10:31:18 ebur128=0 ffmpeg=1 ma_rss_mb=4277
10:32:18 ebur128=0 ffmpeg=1 ma_rss_mb=4288
10:33:18 ebur128=0 ffmpeg=2 ma_rss_mb=4294
10:34:18 ebur128=0 ffmpeg=1 ma_rss_mb=4315
10:35:18 ebur128=0 ffmpeg=1 ma_rss_mb=4315
10:36:19 ebur128=0 ffmpeg=1 ma_rss_mb=4315
10:37:19 ebur128=1 ffmpeg=3 ma_rss_mb=4315
10:38:19 ebur128=0 ffmpeg=1 ma_rss_mb=4315
10:39:19 ebur128=0 ffmpeg=1 ma_rss_mb=4315
10:40:19 ebur128=1 ffmpeg=3 ma_rss_mb=4319
10:41:19 ebur128=0 ffmpeg=1 ma_rss_mb=4354
10:42:19 ebur128=0 ffmpeg=1 ma_rss_mb=4393
10:43:19 ebur128=0 ffmpeg=1 ma_rss_mb=4446
10:44:19 ebur128=0 ffmpeg=1 ma_rss_mb=4450
10:45:19 ebur128=0 ffmpeg=1 ma_rss_mb=4452
10:46:19 ebur128=0 ffmpeg=1 ma_rss_mb=4451
```

TermeHansen:
@chrisuthe do you have any idea where this large memory usage comes from and how we can try to mitigate it?

Also I don't feel that this memory issue actually has a lot to do with this PR, so maybe it was better to take this part to a separate issue and then focus here on the actual Pandora provider upgrade?

chrisuthe:
Apologies @TermeHansen for the delay.  Full disclosure I am beyond my depth here. I would agree however that this is not a problem with your integration changes but rather PROBABLY related how ram is allocated on instances that are built, not running a production docker image which uses a different allocator.... 

The default allocator (glibc) makes this worse in two ways: it gives each thread its own memory pool, and it almost never returns freed memory to the OS. Smart Fades spreads its number-crunching across 32 worker threads, so you end up with ~32 pools each bloating to a full song's worth of buffers and never shrinking → the steady climb you're seeing.

The fix the project already uses is jemalloc (it pulls it in within the dockerfile), a drop-in replacement allocator that uses few pools and returns memory to the OS aggressively. The Docker image / HA add-on already loads it automatically — but you're running from source, which doesn't, so you're seeing the unmasked behavior.

Can you try this (no code changes)?


sudo apt-get install -y libjemalloc2
LD_PRELOAD="$(ls /usr/lib/*/libjemalloc.so.2 | head -n1)" mass    # in front of however you start it


Let it run for a few hours and watch the same graph. If the memory curve flattens out, that confirms it's allocator retention rather than a leak — which means it's a deployment/docs thing for source installs, and separable from this PR (agreeing with your instinct to split it out).


Again- I'm learning as I go here so I could be completely wrong.  My hope is using jemalloc will show the ram freed back up more aggressively.... but either way I don't think it's the pandora rewrite.

TermeHansen:
@chrisuthe I belive you are right. Might be a good idea to add to the [DEVELOPMENT.md](https://github.com/music-assistant/server/blob/dev/DEVELOPMENT.md) at least when looking into memory usage :)

```
09:29:45 ebur128=1 ffmpeg=3 ma_rss_mb=1449
09:30:45 ebur128=0 ffmpeg=1 ma_rss_mb=1535
09:31:45 ebur128=0 ffmpeg=1 ma_rss_mb=1517
09:32:45 ebur128=0 ffmpeg=1 ma_rss_mb=1518
09:33:45 ebur128=0 ffmpeg=1 ma_rss_mb=1518
09:34:45 ebur128=0 ffmpeg=1 ma_rss_mb=1517
09:35:46 ebur128=0 ffmpeg=1 ma_rss_mb=1559
09:36:46 ebur128=0 ffmpeg=1 ma_rss_mb=1522
09:37:46 ebur128=0 ffmpeg=1 ma_rss_mb=1610
09:38:46 ebur128=0 ffmpeg=1 ma_rss_mb=1530
09:39:46 ebur128=1 ffmpeg=3 ma_rss_mb=1535
09:40:46 ebur128=0 ffmpeg=1 ma_rss_mb=1548
09:41:46 ebur128=0 ffmpeg=1 ma_rss_mb=1536
09:42:46 ebur128=0 ffmpeg=1 ma_rss_mb=1536
09:43:46 ebur128=1 ffmpeg=3 ma_rss_mb=1606
09:44:46 ebur128=0 ffmpeg=1 ma_rss_mb=1571
09:45:46 ebur128=0 ffmpeg=1 ma_rss_mb=1566
09:46:47 ebur128=0 ffmpeg=1 ma_rss_mb=1564
09:47:47 ebur128=0 ffmpeg=1 ma_rss_mb=1564
09:48:47 ebur128=1 ffmpeg=3 ma_rss_mb=1642
09:49:47 ebur128=0 ffmpeg=1 ma_rss_mb=1595
09:50:47 ebur128=0 ffmpeg=1 ma_rss_mb=1592
09:51:47 ebur128=0 ffmpeg=1 ma_rss_mb=1639
09:52:47 ebur128=0 ffmpeg=1 ma_rss_mb=1582
09:53:47 ebur128=0 ffmpeg=1 ma_rss_mb=1575
09:54:47 ebur128=0 ffmpeg=1 ma_rss_mb=1575
09:55:47 ebur128=0 ffmpeg=1 ma_rss_mb=1576
09:56:47 ebur128=0 ffmpeg=1 ma_rss_mb=1667
09:57:48 ebur128=0 ffmpeg=1 ma_rss_mb=1618
09:58:48 ebur128=0 ffmpeg=1 ma_rss_mb=1620
09:59:48 ebur128=0 ffmpeg=1 ma_rss_mb=1619
10:00:48 ebur128=0 ffmpeg=2 ma_rss_mb=1660

```

I suggest this PR returns to normal review then, who would that be then?

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +288,141 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
+        raise MediaNotFoundError(f"Artist {artist_name} not found")
+
+    async def get_audio_stream(
+        self, streamdetails: StreamDetails, seek_position: int = 0
+    ) -> AsyncGenerator[bytes, None]:
+        """Return bytes of track from local _audio_cache."""
+        track_id = streamdetails.item_id.split("_")[-1]
+        audio = self._audio_cache.pop(track_id, b"")
+        for i in range(0, len(audio), 65536):
+            yield audio[i : i + 65536]
```
chrisuthe:
```suggestion
    async def get_audio_stream(
        self, streamdetails: StreamDetails, seek_position: int = 0
    ) -> AsyncGenerator[bytes, None]:
        """Return bytes of track from local _audio_cache."""
        track_id = streamdetails.item_id.split("_")[-1]
        if not (audio := self._audio_cache.get(track_id)):
            raise MediaNotFoundError(f"No cached audio for track {streamdetails.item_id}")
        start = 0
        if seek_position and streamdetails.duration:
            start = int(len(audio) / streamdetails.duration * seek_position)
        for i in range(start, len(audio), 65536):
            yield audio[i : i + 65536]
```
A few intertwined issues in this method that one change can fix:

**1. `seek_position` is ignored, but `get_stream_details` advertises `can_seek=True`.** For `StreamType.CUSTOM` streams the streams controller skips its own ffmpeg `-ss` seek and trusts the provider to start at the right offset (`controllers/streams/audio.py`). So dragging the position slider currently restarts audio at 0:00 while the UI shows the seeked position.

**2. The `pop` makes any second read fail.** A seek (and some buffer-recovery paths) triggers a fresh `get_audio_stream` call for the same track — but the first call already popped the bytes, so the retry gets `b""`. Switching to `.get()` keeps the entry alive; memory stays bounded exactly as before since eviction already happens in `_add_audio` via `MAX_CACHE_SIZE`.

**3. A cache miss currently yields nothing, silently.** ffmpeg then fails with a generic "No audio was received" that points nowhere near the real cause. Raising `MediaNotFoundError` instead lets the queue controller skip the track with a log line that actually names it.

The seek math is the same linear approximation the deezer provider uses (seconds → byte offset via `size / duration`) — MP3/AAC frames are self-syncing, so the decoder locks onto the next frame boundary after the cut.

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +288,141 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
+            fragment = await self._get_fragment_data(session, fragment_index + i)
+            for track in fragment:
+                if not track["cached"]:
+                    async with self.mass.http_session.get(track["audioURL"]) as resp:
+                        self._add_audio(track["musicId"], await resp.read())
+                    track["cached"] = True
+                    tracks.append(track)
+                    if len(tracks) >= 2:
+                        return [self._parse_track(track) for track in tracks]
+        return [self._parse_track(track) for track in tracks]
```
chrisuthe:
```suggestion
                if not track["cached"]:
                    try:
                        async with self.http_session.get(
                            track["audioURL"], timeout=aiohttp.ClientTimeout(total=30)
                        ) as resp:
                            if resp.status != 200:
                                self.logger.warning(
                                    "Failed to download audio for %s (HTTP %s) - skipping",
                                    track.get("songTitle"),
                                    resp.status,
                                )
                                continue
                            self._add_audio(track["musicId"], await resp.read())
                    except (TimeoutError, aiohttp.ClientError) as err:
                        self.logger.warning(
                            "Error downloading audio for %s: %s", track.get("songTitle"), err
                        )
                        continue
                    track["cached"] = True
```
I think you meant to use the existing session to respect the SOCKS proxy if enabled? 

Pandora's CDN URLS expire, would be good to raise an error if it does..

music_assistant/providers/pandora/provider.py
```diff
@@ -114,34 +109,38 @@ async def handle_async_init(self) -> None:
...
 
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await super().loaded_in_mass()
+        for player in self.mass.players.all_players(return_disabled=True):
+            items = self.mass.player_queues.items(player.player_id)
+            if items:
+                if track := items[-1].media_item:
+                    if track.provider == self.domain and self._find_track(track.item_id) == {}:
+                        self.mass.player_queues.clear(items[-1].queue_id)
```
chrisuthe:
```suggestion
    async def loaded_in_mass(self) -> None:
        """Call after the provider has been loaded."""
        await super().loaded_in_mass()

        for player in self.mass.players.all_players(return_disabled=True):
            try:
                items = self.mass.player_queues.items(player.player_id)
                if not items or not (track := items[-1].media_item):
                    continue
                if track.provider != self.domain:
                    continue
                self.logger.info(
                    "Clearing stale Pandora queue for player %s", player.player_id
                )
                self.mass.player_queues.clear(items[-1].queue_id)
            except Exception as err:
                self.logger.warning(
                    "Failed to check/clear queue for player %s: %s", player.player_id, err
                )
```

cached fragments/audio do not survive a restart or reload, so any Pandora tracks left in player queues are no longer playable: clear those queues.  The _find_track check is dead weight I think, it should always return TRUE since _sessions at load time are always empty...

music_assistant/providers/pandora/provider.py
```diff
@@ -96,10 +91,10 @@ class PandoraProvider(MusicProvider):
     _sessions: dict[str, PandoraStationSession]
     _socks_proxy: bool = False
     _high_quality_available: bool = False
+    _audio_cache: OrderedDict[str, bytes] = OrderedDict()  # musicId → raw audio bytes
 
     async def handle_async_init(self) -> None:
         """Handle async initialization of the provider."""
-        self._on_unload_callbacks = []
         self._sessions = {}
```
chrisuthe:
```suggestion
    _audio_cache: OrderedDict[str, bytes]  # musicId → raw audio bytes

    async def handle_async_init(self) -> None:
        """Handle async initialization of the provider."""
        self._sessions = {}
        self._audio_cache = OrderedDict()
```
Change up so every pandora provider doesn't share the same dict, if you want to have multiple accounts configured for multiple streams at once you'd want one per.

chrisuthe (changes requested):
Great work on this rewrite — the dynamic playlist approach is a much better fit for how Pandora actually works👍

I left a handful of inline suggestions (please play with them and sanity-check me — I wrote them against the code, not against a running station):

Seek/replay: keep cached audio readable instead of popping it on first play, and actually honor seek_position since the streamdetails advertise can_seek=True
Proxy + download hardening: route the audio fetch through self.http_session so SOCKS users stay proxied, plus status/timeout/exception guards so a bad CDN response can't get cached as "audio" or silently kill queue refill
loaded_in_mass: per-player error isolation so one bad queue can't break provider load, and a log line when we clear someone's queue
_audio_cache: move to instance state like _sessions so two Pandora instances don't share/evict one cache

None of these change your design — they're all hardening around the edges of it. The big one to verify on a real station is the seek change, since I couldn't test against live Pandora.

music_assistant/providers/pandora/provider.py
TermeHansen: thanks added in my last commit

music_assistant/providers/pandora/provider.py
TermeHansen: I agree with the dead weight, and remember putting a mental note of changing it later, that became now :)  I find it is difficult to read with the not ifs to continue and have chosen to stay with the multiple ifs as I think this gives the same in the end right?

music_assistant/providers/pandora/provider.py
TermeHansen: thank you I have implemented in my last commit

music_assistant/providers/pandora/provider.py
TermeHansen: great. I have implemented and tested it a bit with the seeking and it seems to work. I will test it some more the coming days...

TermeHansen:
@chrisuthe 
So I have played a bit around with it now and found that the 10 spot audio_cache was to small when playing two players at the same time, as I could see in logs that some times it failed to find a track in the cache. This wasn't an issue for me before, and I can see that this is because of the suggestion of not cleaning out streamed tracks when last part of a track has been yielded. We could of course just have a bigger audio_cache, but I don't see a reason to fill memory up also with tracks that are not needed anymore.

So I added the cleaning up again in my last commit, please share thoughts or comments on this.

I see one still standing issue tough. Players that are not there when we start up MA and `loaded_in_mass` is called. I have the issue with for instance browser players. They will show the old playlist of tracks, but when trying to play it just fails with:

```
2026-06-12 09:58:44.571 WARNING (MainThread) [music_assistant.player_queues] Skipping unplayable item Tord Gustavsen Quartet - Circling
2026-06-12 09:58:44.571 WARNING (MainThread) [music_assistant.player_queues] Skipping unplayable item Pat Metheny - Goin' Ahead
2026-06-12 09:58:44.571 WARNING (MainThread) [music_assistant.player_queues] Skipping unplayable item Bobo Stenson - Alfonsina
2026-06-12 09:58:44.571 WARNING (MainThread) [music_assistant.player_queues] Skipping unplayable item Jim Hall - I Can't Get Started
2026-06-12 09:58:44.571 WARNING (MainThread) [music_assistant.player_queues] Skipping unplayable item Lars Danielsson - Forever You
2026-06-12 09:58:44.572 WARNING (MainThread) [music_assistant.webserver] players/cmd/play_pause: No playable item found to start playback
```

Why is it completely failing and not just pulling in new tracks? If I clear queue manually in UI and push play then it work fine...

OzGav:
There is a conflict to resolve

TermeHansen:
Is it for me or reviewer to decide and click resolve?

OzGav:
You need to resolve it

music_assistant/providers/pandora/provider.py
```diff
@@ -50,8 +59,7 @@
 )
 from .helpers import create_auth_headers, get_csrf_token, handle_pandora_error
 
-if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator
+MAX_CACHE_SIZE = 10
```
chrisuthe: Keep audio in the bounded LRU instead of freeing it on stream completion, and raise the cap so two players streaming the same account don't evict each other's queued tracks. Pairs with the `get_audio_stream` change below.

```suggestion
# Bounded LRU of recently played audio; freeing here (rather than on stream
# completion) keeps just-played tracks seekable and survives two parallel players.
MAX_CACHE_SIZE = 30
```

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +297,162 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
+        track_id = streamdetails.item_id.split("_")[-1]
+        if not (audio := self._audio_cache.get(track_id)):
+            raise MediaNotFoundError(f"No cached audio for track {streamdetails.item_id}")
+        start = 0
+        if seek_position and streamdetails.duration:
+            start = int(len(audio) / streamdetails.duration * seek_position)
+        for i in range(start, len(audio), 65536):
+            yield audio[i : i + 65536]
+        # clean up cache after streaming
+        del self._audio_cache[track_id]
```
chrisuthe: The trailing `del` breaks seeking and can `KeyError`:
- For `StreamType.CUSTOM` with `can_seek=True`, the streams controller re-calls `get_audio_stream` with the new offset (it skips its own ffmpeg `-ss`). After the first play the bytes are already deleted, so a seek hits an empty cache → `MediaNotFoundError`.
- Under two parallel players, `_add_audio` can evict this entry mid-stream, so the final `del` raises `KeyError`.

Drop the `del`, refresh LRU recency on read, and rely on `MAX_CACHE_SIZE` eviction (bumped to 30 above):

```suggestion
        track_id = streamdetails.item_id.split("_", 1)[-1]
        if not (audio := self._audio_cache.get(track_id)):
            raise MediaNotFoundError(f"No cached audio for track {streamdetails.item_id}")
        self._audio_cache.move_to_end(track_id)
        start = 0
        if seek_position and streamdetails.duration:
            start = int(len(audio) / streamdetails.duration * seek_position)
        for i in range(start, len(audio), 65536):
            yield audio[i : i + 65536]
```

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +297,162 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
+                        continue
+                    track["cached"] = True
+                    tracks.append(track)
+                    if len(tracks) >= 2:
+                        return [self._parse_track(track) for track in tracks]
+        return [self._parse_track(track) for track in tracks]
+
+    async def get_stream_details(self, prov_item_id: str, media_type: MediaType) -> StreamDetails:
+        """Get streamdetails for a radio station."""
+        station_id, track_id = prov_item_id.split("_")
```
chrisuthe: Split on the first underscore so this stays consistent with the cache key (`musicId`) and survives an underscore in the id. `get_audio_stream` above and `_find_track` below use the same form.

```suggestion
        station_id, track_id = prov_item_id.split("_", 1)
```

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +297,162 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
+    def _add_audio(self, key: str, value: bytes) -> None:
+        # If the cache is full, pop the oldest item
+        if len(self._audio_cache) >= MAX_CACHE_SIZE:
+            self._audio_cache.popitem(last=False)
+        self._audio_cache[key] = value
+
+    def _find_track(self, prov_track_id: str) -> dict[str, Any]:
+        """Find track in all station fragments from provider track_id."""
+        if "_" in prov_track_id:
+            station_id, track_id = prov_track_id.split("_")
```
chrisuthe: Same first-underscore split for consistency.

```suggestion
            station_id, track_id = prov_track_id.split("_", 1)
```

music_assistant/providers/pandora/provider.py
```diff
@@ -400,35 +479,18 @@ async def _get_fragment_data(
...
-                    session.track_map.append((fragment_index, track_idx))
-                    session.cumulative_times.append(current_cumulative)
-
-                    duration = track.get("trackLength", 0)
-                    current_cumulative += duration
-
-            return result
+                session.fragments.append([])
+            session.fragments[fragment_index] = tracks
+            return tracks
```
chrisuthe, line 530: `session.fragments` grows unbounded — one entry per fragment for the whole session, each holding track-metadata dicts. A likely contributor to the memory growth reported in this thread. Drop metadata once a fragment ages out of the cache window:

```suggestion
            # Store in session cache
            while len(session.fragments) <= fragment_index:
                session.fragments.append([])
            session.fragments[fragment_index] = tracks
            # Drop metadata for a fragment that has aged out of the audio cache
            # window so memory stays bounded over a long listening session.
            prune_index = fragment_index - MAX_CACHE_SIZE
            if prune_index >= 0 and session.fragments[prune_index]:
                session.fragments[prune_index] = []
            return tracks
```

music_assistant/providers/pandora/provider.py
```diff
@@ -288,92 +297,162 @@ async def get_radio(self, prov_radio_id: str) -> Radio:
...
-                "pageSize": 250,
-            },
-        )
+    async def browse(self, path: str) -> Sequence[MediaItemType | ItemMapping | BrowseFolder]:
+        """Browse Pandora radio stations."""
+        sub_path = path.split("://", 1)[1] if "://" in path else ""
+        if not sub_path:
+            result: list[MediaItemType | ItemMapping | BrowseFolder] = []
+            async for station in self._get_stations():
+                self.logger.info(f"Retrieved {station.name}, {station.is_dynamic}")
```
chrisuthe: Consider if the Per-station retrieval logging is noisy at INFO level, maybe not remove but move it up?

```suggestion
                self.logger.debug(f"Retrieved {station.name}, {station.is_dynamic}")
```

music_assistant/providers/pandora/provider.py
```diff
@@ -524,99 +536,120 @@ def _get_or_create_session(self, station_id: str) -> PandoraStationSession:
...
+                        provider=self.instance_id,
+                        type=ImageType.THUMB,
+                        path=album_art_url,
+                        remotely_accessible=True,
+                    )
+                )
+        if artist_name := obj.get("artistName"):
+            track.artists = UniqueList([self._parse_artist(artist_name)])
+        track.album = self._parse_album(obj, track_id)
+        self.logger.info(name)
```
chrisuthe: Leftover debug line logs every parsed track title at INFO — remove it or move it up to debug

```suggestion
        track.album = self._parse_album(obj, track_id)
```

music_assistant/providers/pandora/provider.py
```diff
@@ -114,34 +110,46 @@ async def handle_async_init(self) -> None:
...
+                    if track := items[-1].media_item:
+                        if track.provider == self.domain:
+                            self.logger.info(
+                                f"Clearing stale Pandora queue for player {player.player_id}"
+                            )
+                            self.mass.player_queues.clear(items[-1].queue_id)
+            except Exception as err:
+                self.logger.warning(
+                    f"Failed to check/clear queue for player {player.player_id}: {err}"
+                )
```
chrisuthe: This sweep only sees players present at load. Browser/cast players that connect *after* load keep a queue of now-unplayable Pandora tracks (cache/session are empty after a restart), so play fails on stale items instead of refilling — matching the "Skipping unplayable item" report in the thread.

This fix spans several locations so it can't be a one-click suggestion (it's on branch `pr3765-review-fixes` if you'd rather pull it). Refactor the clear into a helper and run it both at load and on `PLAYER_ADDED`:

```python
# imports
from collections.abc import AsyncGenerator, Callable, Sequence
from typing import TYPE_CHECKING, Any
from music_assistant_models.enums import ContentType, EventType, ImageType, MediaType, StreamType

if TYPE_CHECKING:
    from music_assistant_models.event import MassEvent

# class attribute
_unsub_player_added: Callable[[], None] | None = None

async def unload(self, is_removed: bool = False) -> None:
    """Handle unload/close of the provider."""
    if self._unsub_player_added:
        self._unsub_player_added()
        self._unsub_player_added = None
    await self.close()
    await super().unload(is_removed)

async def loaded_in_mass(self) -> None:
    """Call after the provider has been loaded."""
    await super().loaded_in_mass()
    for player in self.mass.players.all_players(return_disabled=True):
        self._clear_stale_queue(player.player_id)
    if self._unsub_player_added is None:
        self._unsub_player_added = self.mass.subscribe(
            self._on_player_added, EventType.PLAYER_ADDED
        )

def _on_player_added(self, event: MassEvent) -> None:
    """Clear a stale Pandora queue for a player that connected after load."""
    if event.object_id:
        self._clear_stale_queue(event.object_id)

def _clear_stale_queue(self, player_id: str) -> None:
    """Clear a player's queue if it ends on a now-unplayable Pandora track."""
    try:
        items = self.mass.player_queues.items(player_id)
        if items and (track := items[-1].media_item) and track.provider == self.domain:
            self.logger.info("Clearing stale Pandora queue for player %s", player_id)
            self.mass.player_queues.clear(items[-1].queue_id)
    except Exception as err:
        self.logger.warning("Failed to check/clear queue for player %s: %s", player_id, err)
```

This *clears* stale queues (safe, minimal); repopulating from the dynamic playlist would be nicer UX but is a design call for you.

chrisuthe (changes requested):
Follow-up on my earlier review. The dynamic-list rewrite is solid — these are edge-hardening fixes, several of which line up with symptoms reported in this thread (seeking, memory growth, late-joining players failing to play).

7 of these are one-click applyable suggestions. The late-joining-player fix spans several locations, so it's posted as a comment with the full code (also on branch `pr3765-review-fixes`).

Caveat: I verified lint/format/mypy locally but could **not** test against a live Pandora station — please sanity-check the seek and `PLAYER_ADDED` behaviour on a real station before merging.

music_assistant/providers/pandora/provider.py
TermeHansen: @chrisuthe Thank you for the hints in this direction. I'm testing it out right now, and it seems like the queue is not always filled when the `PLAYER_ADDED` fires, actually I think I have only seen it ones. The `QUEUE_ADDED` event seems to do exactly what I want. Now I just want to understand when this fires if it only happens in this case after player_added or also at other points. I have tried in ui to clear queue and and add new items, but it does not fire from any of that. For now I safe guard that the current_item is NOT to be found by `_find_tracks()`

Here you can see from my debug logs that first `PLAYER_ADDED` fires with no queue items, and right after `QUEUE_ADDED` with the old pandora items...

```
2026-06-17 10:59:25.058 WARNING (MainThread) [music_assistant.Pandora] Player added...
2026-06-17 10:59:25.058 WARNING (MainThread) [music_assistant.Pandora] trying to clear stale queue for player ma_jrisg0js88
2026-06-17 10:59:25.058 WARNING (MainThread) [music_assistant.Pandora] Checking for stale Pandora queue for player ma_jrisg0js88, queue items: []
2026-06-17 10:59:25.064 WARNING (MainThread) [music_assistant.Pandora] Queue added...
2026-06-17 10:59:25.064 WARNING (MainThread) [music_assistant.Pandora] Trying to clear stale Pandora queue for player ma_jrisg0js88 from queue added event
```

TermeHansen: ok leaving the browser closed for a while and the player leaves, and opening it will come again and both events fire, so it is needed to check if tracks already exists (or of course clear no matter what every time :D

Here later (no queues cleared as tracks exist in session.fragments)
```
2026-06-17 14:44:00.826 WARNING (MainThread) [music_assistant.Pandora] Player added...
2026-06-17 14:44:00.826 WARNING (MainThread) [music_assistant.Pandora] trying to clear stale queue for player ma_jrisg0js88
2026-06-17 14:44:00.826 WARNING (MainThread) [music_assistant.Pandora] Checking for stale Pandora queue for player ma_jrisg0js88, queue items: []
2026-06-17 14:44:00.833 WARNING (MainThread) [music_assistant.Pandora] Queue added...
```

TermeHansen:
@chrisuthe I have tested the seek feature (I think you asked here), many times and it works perfectly fine for me at least :smiley_cat: 

The new cleanup of queues also seem to work as wanted (clearing only if pandora tracks no more exist in the fragments).

Last thing to me on the usage of Pandora like this with dynamic playlists, is how to see what "radio station"  currently is playing (or potentially has a cleared queue) for a player in the ui...

mcaulifn:
I was randomly working on the same thing today and ran into a few bugs during testing. Hopefully these help.

**1. Background metadata-enrichment task can corrupt a live Pandora session.** MA runs a recurring background task (`_refresh_playlist_metadata_batch` in `metadata/controller.py`) that calls `.tracks()` on library playlists to scan for genre/collage metadata. Since `is_dynamic` playlists now include real station sessions, and Pandora only permits one active stream per account, this task hitting `getFragment` on 5 different stations back-to-back (batch size 5) caused real content corruption for me - wrong tracks loading just from background scanning, no playback involved. Fix was to skip `is_dynamic` playlists in both the query that selects playlists due for refresh (`metadata/controller.py`) and in `_update_playlist_metadata` itself (`metadata/enrichment.py`) as a belt-and-suspenders. This isn't Pandora-specific - same flag is used by Apple Music stations / Yandex wave / NetEase personal FM, so it's a generic fix worth landing regardless of this PR's approach.

**2. No migration for existing library Pandora stations.** Before this rework, Pandora stations were stored as `Radio` library rows. Since the provider now reports everything as `Playlist`, existing users upgrading would be left with stale/orphaned `radios` rows and broken provider_mappings pointing at a media_type the provider no longer returns. We added a one-time, schema-gated migration (`DB_SCHEMA_VERSION` 41->42) that moves existing Pandora `radios` rows into `playlists` and repoints their `provider_mappings` accordingly, preserving favorites/play history. Verified against a real backup DB.


I am happy to raise a PR against this branch if you are interested.

TermeHansen:
At 1. It is solved in this PR by how it is implemented. I have played for long periods on 2 players simultaneously with no problem. Maybe you misunderstood this that only urls from latest pulled fragment works / are active, which is why we in this PR cache the audio streams...

For 2. I don't know how to handle such updates for a provider in a nice way for the user.

OzGav:
It sounds like we do need to add some migration logic.

mcaulifn:
Re # 1: I was able to reproduce this directly against your branch to check, since I wanted to make sure I wasn't conflating it with the URL-staleness issue. 

Steps: 
- pulled `pandora_dynamic_list` (merged with current `dev`)
- favorited ~8 real Pandora stations
- started playback on one
- then manually triggered the "Refresh playlist metadata" scheduled task (`tasks/run`) while still playing. 
 
Here's the relevant log excerpt:

```
09:12:12.424 Updating metadata for Playlist Bob Marley & The Wailers Radio
09:12:13.817 Updating metadata for Playlist 90s Alternative Radio
09:12:13.820 [Pandora] Evicting session for station 196020935147391677
09:12:16.240 Updating metadata for Playlist Relaxing Radio
09:12:16.243 [Pandora] Evicting session for station 3856267052821519452
09:12:18.596 Updating metadata for Playlist The Beatles Radio
09:12:18.606 [Pandora] Evicting session for station 43
...
09:12:29.557 Preparing audio buffer for next track Billie Eilish - bad guy on queue Web (Edge on Windows)
09:12:29.578 [in#0 @ 0x...] Error opening input: Invalid data found when processing input
09:12:29.578 Process stdin feeder task ended with error: No cached audio for track 3383384990253849692_S17647864
09:12:29.582 Failed to prepare next audio buffer: Failed to stream audio
```

What happened: the scan touched 4 stations back-to-back. Each fetch downloads new track audio into `_audio_cache` (the shared `OrderedDict`, `MAX_CACHE_SIZE = 24`). The track that failed (`...S17647864`) had already been downloaded earlier when the active queue preloaded it as the next-up track - it was sitting in `_audio_cache` waiting to be played. By the time the queue went to actually use it, the scan's fetches for the 4 other stations had inserted enough new entries that this one got evicted via plain `popitem(last=False)`, and the queue failed outright with "No cached audio for track."

The root issue: eviction in both `_audio_cache` (24-slot) and `_sessions` (10-slot) is pure insertion/access-order LRU, with no concept of "this entry is currently in use or about to be played." A track that's queued up and waiting to play looks identical to eviction logic as a track nobody cares about anymore - it just hasn't been "touched" recently, so it reads as the oldest entry once a handful of unrelated stations get fetched in between. That's what happened here: the actively-relevant entry wasn't touched again after being preloaded, while the scan's fetches for other stations kept inserting newer entries, so it aged out and got evicted out from under the queue that needed it.

So I don't think this is the same failure as the stale-signed-URL problem the audio caching addresses - it's a separate issue in how the cache is shared and evicted across stations regardless of what's actively in use.

Happy to share the full log or the test setup if useful.

TermeHansen:
ahh ok now I get what you refer to I think, so by adding to favorite (I don't they are right there in browse->Pandora anyhow) the internals of MA will pull tracks from them even though they are not going to play. So by flavoring 8 stations it is actually similar to starting 8 players at the same time playing the 8 stations....

That is for sure a border case where I have not been :smile:

mcaulifn:
Kinda edge case but I encountered it almost immediately when testing my implementation. I got around it by labeling Pandora (and some Apple music) playlists as dynamic and skipping the refresh. This avoids refreshing for something that will end up changing anyway and then it also doesn't impact active streaming.

TermeHansen:
@OzGav can this approach still handle the `ProviderFeature.LIBRARY_RADIOS` and a related question. I actually left the `get_radio()` around in the hope that I could at some point later make it work from the Radio side as well. I guess I could just remove that function if not possible, right?

TermeHansen:
> Kinda edge case but I encountered it almost immediately when testing my implementation. I got around it by labeling Pandora (and some Apple music) playlists as dynamic and skipping the refresh. This avoids refreshing for something that will end up changing anyway and then it also doesn't impact active streaming.

@mcaulifn In my opinion this is more a bug/issue for the use of dynamic playlist for streaming services where there can be a limit to how often and much to pull tracks from the service.

mcaulifn:
@TermeHansen I think this is a function of how the service (Pandora/Apple Music) works and for a good user experience, MA should do its best to accommodate. My take is that there is no need to refresh metadata because each dynamic playlist changes depending on when you refresh. The data fetched now is different than the data fetched in 24 hours. And neither is really relevant because both will be different from when the user actually presses play.

TermeHansen:
I agree which is why I think it should be taken as a seperate issue followed by a seperate PR 🤗

mcaulifn:
Sorry, I misunderstood :oops: I'll look at a small PR to establish the framework then.

OzGav:
@TermeHansen I think you are right to remove all the radio related methods, since the provider now uses playlists to surface the Pandora content.

One thing I would reconsider though. With only BROWSE declared, stations only reach the library when the user favourites them from browse. I think auto adding them would be fine, the user has already made a decision to create those stations in Pandora. You could declare ProviderFeature.LIBRARY_PLAYLISTS and add a get_library_playlists() (basically the old get_library_radios logic, yielding playlists) so they get pulled one way into MA on each sync. That also keeps parity with how the provider works today, where all stations are auto added to the library as radios.

It is an unusual setup now (dynamic playlists rather than radios) but I think it holds together. The one thing to pair it with is skipping is_dynamic playlists in the metadata refresh, otherwise the background scan will pull fragments from every synced station. That is the separate fix you and @mcaulifn are already lining up.

music_assistant/providers/pandora/provider.py
```diff
@@ -96,11 +96,13 @@ class PandoraProvider(MusicProvider):
...
     _socks_proxy: bool = False
     _high_quality_available: bool = False
+    _audio_cache: OrderedDict[str, bytes]  # musicId → raw audio bytes
+    _unsub_queue_added: Callable[[], None] | None = None
 
     async def handle_async_init(self) -> None:
         """Handle async initialization of the provider."""
-        self._on_unload_callbacks = []
         self._sessions = {}
+        self._audio_cache = OrderedDict()
```
marcelveldt, line 105: why do we need a audio cache in the provider code ?

TermeHansen: Because Pandora stream links only work for the latest 4 songs fetched from the api at any time, so when building the playlist we easy make links unplayable, and playing more than one station at a time basically impossible. Caching makes this window of playable tracks bigger.

chrisuthe (commented):
Woops, missed a revision, not applicable!

music_assistant/providers/pandora/provider.py
```diff
@@ -114,34 +116,194 @@ async def handle_async_init(self) -> None:
...
+                            self._add_audio(track["musicId"], await resp.read())
+                    except (TimeoutError, aiohttp.ClientError) as err:
+                        self.logger.warning(
+                            "Error downloading audio for %s: %s", track.get("songTitle"), err
+                        )
+                        continue
+                    track["cached"] = True
+                    tracks.append(track)
+                    if len(tracks) >= 2:
+                        return [self._parse_track(track) for track in tracks]
```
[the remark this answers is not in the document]
TermeHansen: I don't agree. get_playlist_tracks() is here made so it always return 2 freshly cached songs. When a song is cached, the 'cached' boolean on the track is set so it will not be passed again to a playlist. In sum this function iterate over first songs in current fragment to find two uncached tracks and then if still running pull a new fragment until it has found 2 uncached tracks to cache and return.

TermeHansen:
> It sounds like we do need to add some migration logic.

@OzGav Do you know how this can and should be handled?

OzGav:
All migration logic is handled in `music_assistant/controllers/music/migrations.py`. I'd add a step there that moves any existing Pandora stations from the old radio table over to playlists, and updates their links to match.

TermeHansen:
@mcaulifn do you have the SQL string you used for the `one-time, schema-gated migration` you tested?

wouldn't it be something like:

```
DELETE FROM radios
WHERE item_id IN (
    SELECT item_id FROM provider_mappings
    WHERE provider_domain = 'pandora' AND media_type = 'radio'
);
```

and 

```
DELETE FROM provider_mappings
WHERE provider_domain = 'pandora' AND media_type = 'radio';
```

I don't see the point in migration to the playlists table as the provider now will add them anyhow next time in syncs I believe.

I tested it locally and seems to work as expected.

TermeHansen:
@marcelveldt  is this an ok way to add the migration rule needed to remove the old pandora radio items in database?

https://github.com/music-assistant/server/pull/3765/commits/b36ab3fd06a2ab6b15f6541eb91e33a7bdfd2178

mcaulifn:
I can pull out what I had. I think it would be helpful to migrate favorites if nothing else.

Do you want a PR, commit to this branch, or just some code pasted here?

TermeHansen:
I think I figured out what is needed in the last commit, I'll ping you again if I need a pointer to what you did specifically

music_assistant/providers/pandora/provider.py
marcelveldt: OK so this not actual audio stored here but links to the audio then ?

marcelveldt:
Marked as draft so you can fix these last changes and then press the "ready for review" button once you are ready. Thanks!

PS: can you also make the PR title a bit more descriptive ?

music_assistant/providers/pandora/provider.py
TermeHansen: It is indeed the audio stream that we download and cache for up to 24 songs at a time in total.

marcelveldt (changes requested):
Thanks for this @TermeHansen — the dynamic-playlist UX (upcoming-track visibility + seeking) is a really nice step up from the old radio model. A few things to sort out before merge:

1. **Drop the `PANDORA_AUTHTOKEN` env-var login path** (`handle_async_init` + `_authenticate`). It looks like a leftover testing shortcut — it bypasses the configured credentials, is shared across all instances, and logs on the normal path. Let's remove it (and the `os` import if it goes unused).

2. **No local audio caching in the provider.** Pandora is a streaming provider, and for those the provider must stay stateless and return a stream URL from `get_stream_details` rather than buffering audio (see the note in `_demo_music_provider`). Holding ~24 full tracks in RAM is something we specifically want to avoid. Can we go back to serving the fragment's `audioURL` directly (as the previous implementation did) and keep only the lightweight fragment/track metadata in the session? Keep `can_seek` on if Pandora's CDN honors range requests, otherwise it's fine to leave it off.

3. **Provider shouldn't reach into players/queues.** The `loaded_in_mass` / `QUEUE_ADDED` subscription / `_clear_stale_queue` code drives the players + queue controllers from a music source. I think it only exists to clean up queues pointing at evicted cached audio — so once the in-memory cache is gone (point 2) and tracks stay resolvable, this part can all be removed; the core dynamic-playlist refill (`_fill_radio_tracks`) already handles topping up the queue.

Happy to pair on the streaming side if it helps.

TermeHansen:
@marcelveldt 

At point 1 you are right that this is a develop thing that I needed to add, as it seems that many restarts and by that fresh logins to Pandora will flag my IP as a scrapper and make it impossible to login for several days.

On point 2 (and 3) I totally get the wish of not having to cache streamed audio locally, but we ended at this solution out of necessity to be able to get the dynamic playlist working. Problem in this case is:
- When pulling a fragment from Pandora you get 4 links that works for streaming. As soon as you pull a new fragment (for more songs or another station) the first 4 links will not work anymore.
- `get_playlist_tracks`  is called by MA to top up the playlist until at least 5 tracks in the playlist. Initially I made an implementation where `get_playlist_tracks` would only return new tracks if the last of the previous 4 tracks had played. This however was still having many issues to get working correctly and stable. Together with the good help of @OzGav we made this current approach with a custom stream_type.

Long story short I suggest looking into a solution where by default Pandora will use that initial approach without caching and I can see again what will be known not to work well. Then adding an advanced Pandora setting where one can set the maximum number of tracks to cache. 8 (2 fragments) makes it possible to play on one player without any issues.

Again I get the point on not wanting to cache audio in general, but I also as a user would like to have the option to choose to use a bit of extra memory for some niceness that I appreciate.

MarvinSchenkel:
@TermeHansen just a heads up, we will never accept PRs that will cache audio from streaming providers, because it either eats up RAM or it is a really (dark) gray area in terms of violating the service's TOS if you want to cache it on disk.

TermeHansen:
@MarvinSchenkel Yes I get that now, honestly right now I feel a bit discouraged about all the hours I have spent on this PR. Months ago I asked for help and guidance to get this upgrade to work without caching both here and at discord. I was then guided towards making all this work to make this local caching work with custom_audio, and a lot of hours for testing during this development to get it to work fully with that approach.

Right now I fell like scratching this whole PR and start all over again with a new branch and PR sticking to not doing any local caching. It would be nice if there would be someway to know of the organization behind MA and who are experienced co-developers, who are reviewers and who are final decision makers of what is accepted for the project.

TermeHansen:
I close the PR as it is now superseded by #5557

PR #3776, opened 2026-04-23
Handle syncing for cases where session establishment is both early and late for adhering to required ntpstart value

bradkeifer:
cliap2 version bumped.
cliap2 logging reduced
MA logging of cliap2 output improved to eliminate duplication of timestamp and log level
Config entry for pairing latency
cliap2 handles syncing for scenarios where session establishment is too early or too late for adhere to ntpstart


bradkeifer:
Next iteration ready for your testing @MarvinSchenkel 
I suggest doing the initial test with all default values for both output buffer duration and pairing latency.
If that does not work, please try increasing output buffer duration to 2000ms for both ap2 devices.
If you do get them playing in sync, then I suggest tuning the pairing latency to a lower number, but a time which is longer than the time taken to establish the sessions. You devices seems to do the session establishment between 600-850ms, so perhaps 1000ms for each is a nice safe number.
If they remain playing in sync, then try adjusting the output buffer duration down. It would be great if we can get them playing in sync with this set at it's minimum value of 500ms.
I have tested this again with my configuration of Yamaha AV Receiver using RAOP and AirPort Express using AirPlay2 with pairing latency of 2500ms and output buffer duration of 1000ms and it is playing in sync.

bradkeifer:
Hi @MarvinSchenkel 
Here is next iteration of cliap2 with the changes as describedin our Discord chat. It is ready for your next round of testing. Please do an initial ap2-ap2 test with pairing_latency set to 1000ms for both your devices and try 1000ms, then 500ms for the output_buffer_duration for both devices and let's see how that goes.
It would also be appreciated if you could do a raop-ap2 test with the pairing_latency set for 1000ms and output_buffer_duration set at each of 1000 and 500ms for the ap2 device.
Once we have those results, I would like to work with you to try some different values for the new input_write_ms hidden command line argument. I am not sure yet if this should be a tuneable parameter or if it should be something fixed, but testing will ultimately reveal the answer to that.

bradkeifer:
I've made another set of changes to cliap2 @MarvinSchenkel. They are included in this PR. The changes improve the logic for determining the earliest playback commencement time possible when the binary is working out if it needs to ignore some of the audio before initial playback. Since we burst up to 5secs of audion, and cliap2 reads it early, then we don't need to consider the duration of the output buffer in working our what is the earliest possible time we could commence playback - we only need to consider the pairing latency in that scenario.

bradkeifer:
@MarvinSchenkel - updated binaries to fix the issue with idle queue still showing. Sorry about that - it was my mistake.

music_assistant/providers/airplay/constants.py
```diff
@@ -24,6 +24,7 @@ class StreamingProtocol(IntEnum):
 CONF_ENCRYPTION: Final[str] = "encryption"
 CONF_ALAC_ENCODE: Final[str] = "alac_encode"
 CONF_VOLUME_START: Final[str] = "volume_start"
+CONF_PAIRING_LATENCY: Final[str] = "pairing_latency"
```
marcelveldt: Why would all this be needed ? Isn't the beauty of AirPlay that this is all worked out automatically by adhering to the ntp start ? 

Like, have you ever had to configure the pairing latency or buffer duration on your iphone when you wanted to stream ? Or is this just for testing ?

bradkeifer: Once testing with @MarvinSchenkel is finished, we will know for sure. But at this stage, I expect it to be required as a means to tune the ntpstart time to prevent initial audio loss.

bradkeifer:
@MarvinSchenkel - Next iteration ready for your testing and for discussion on config items. I believe we can eliminate the sync adjust and output buffer duration configuration items, but we need the session establishment latency item in order to deal with the material differences in this time across devices. With my AirPort Express's, if I set the session establishment latency below 2500, then there is no playback at all. With 2500ms it is not 100% reliable and I have landed on 3200ms as the ideal setting for those devices. Seems like 500ms is a good setting for some of yours and 1000ms for others. I have made improvements to logging so people can easily see the performance of their device and there is potential that we could show them this info in the settings screen if desired - or even auto configure the value. Best for you & I to chat about this so I can properly explain to you what I've done and why and then we can bounce some ideas about what the best and most reliable way forward is.

In terms of sync'ed playback, my test environment is AirPort Express running AirPlay2 and Yamaha RX-V477 AV Receiver running RAOP. They always play in sync and I have not been able to find a way to make them play out of sync, even when I need to drop some initial audio for the AirPort Express.

There is something different in the handling of interactions with the binaries when using sync'ed players and when using just a single player. The sync'ed handling is more responsive in handling the end of playback and in pausing playback - not sure why and not sure if it makes any difference - both methods seem to function ok.

I remain of the view that we should provide a mechanism to allow people to have an installation that does not drop audio just because they have a slow performing device. It basically becomes a trade-off between latency and completeness of playback. I would appreciate discussing with you how we can do that and maybe there are some improvement opportunities we can make once we talk it through. For example, I found that (in my test environment), there was 800-900ms of time between when the ntpstart time is calculated and the binary is spawned - that becomes material when trying to prevent audio loss and keep the latency to the minimum possible.

In terms of your testing regime, I would appreciate if you can please test with the same 500ms and 1000ms values you used for pairing latency before (now named session establishment latency to better reflect what it actually represents) and please try 250ms for the output buffer duration. I believe we can move to making the output buffer duration fixed for all devices and remove this as a configurable parameter. It would be helpful if you could send me debug level logs from your testing and I will look for anything anomalous.

I think there is another area where we can make some improvements too. It's in the handling of end of playback/STOP command processing. We are currently sending a kill signal to both binaries. I made changes to cliap2 a while ago so that it gracefully exists on receipt of either a STOP command on command named pipe or on receipt of EOF on stdin. But, MA never seems to take advantage of that and immediately goes for the heavy handed kill signal approach. I am not sure if there is a compelling case to change it, but I would appreciate talking it through with you to make sure that you're all good with the approach being used and are aware of alternatives in case you want to change it in the future.

I will make myself available Monday evening my time to work with you on this i that works for you?

bradkeifer:
Hi @MarvinSchenkel This PR is now updated with the changes we discussed on Monday.
I have tested it with my configuration of devices and it works as expected. However, I did experience a different outcome than previously. This change now implements a default session establishment latency of 500ms and I tried using that for my old and slow airport express. Previously this would result in no audio being played at all. Now the device does play, but it is out of sync with the raop device (Yamaha RX-V477 AV Receiver). The only other change that might be influencing this is that I have made the output buffer duration a default of 1000ms for both raop and ap2. I did this because we know 1000ms works well for cliraop and I didn't want to risk making unnecessary changes to raop that have unintended consequences and it did not seem sensible to have different defaults for the two protocols. I would have been using a lower value like 250 or 500ms for my prior testing, so perhaps such a combination would result in inability of device to play at all?
I have implemented the "big fat disclaimer" as requested - pls let me know if you're ok with how I have worded it.
This is now ready for your testing with however many devices you'd like to try playing in sync. It would be great if you could include at least one raop device in the set of players that should be playing in sync.
If you are happy with progressing this into dev and onwards into nightly, beta etc., I would be keen to start working with you on how we implement the PTP master clock daemon into MA and then we can get the Samsung devices working with airplay and also Shairport Sync devices with AirPlay2 (so long as they are not on same host as MA).

bradkeifer:
A couple of small changes applied here @MarvinSchenkel related to the parsing of the stderr output from the binary. Does not impact sync functionality, but worthwhile change to improve robustness and likely small, but immaterial, performance gain.

music_assistant/providers/airplay/protocols/airplay2.py
```diff
@@ -28,10 +30,46 @@ class AirPlay2Stream(AirPlayProtocol):
...
+        binary and once done, there will be no need for the complexity of a session establishment
+        latency concept to exist in MA.
+        """
+        if self._session_establishment_ms is None:
+            return None
+        high = min(
+            ((self._session_establishment_ms + 99) // 100) * 100,
+            AIRPLAY_SESSION_ESTABLISHMENT_LATENCY_MAX_MS,
+        )
+        low = max(high - 600, AIRPLAY_SESSION_ESTABLISHMENT_LATENCY_MIN_MS)
```
MarvinSchenkel, line 71: Why are we using 600 here while the docstring mentions 700?

music_assistant/providers/airplay/constants.py
```diff
@@ -35,14 +36,23 @@ class StreamingProtocol(IntEnum):
 RAOP_DISCOVERY_TYPE: Final[str] = "_raop._tcp.local."
 DACP_DISCOVERY_TYPE: Final[str] = "_dacp._tcp.local."
 
+# Time allowance for MA processing and OS spawning of the binary. Helps reduce initial audio loss.
+AIRPLAY_DEFAULT_SESSION_DELAY_MS: Final[int] = 900
 # Read ahead buffer for cliraop. Default output buffer duration.
 AIRPLAY_OUTPUT_BUFFER_DEFAULT_DURATION_MS: Final[int] = 1000
 # Minimum output buffer duration permitted.
-AIRPLAY_OUTPUT_BUFFER_MIN_DURATION_MS: Final[int] = 500
+AIRPLAY_OUTPUT_BUFFER_MIN_DURATION_MS: Final[int] = 250
```
MarvinSchenkel: Both MIN and MAX output buffer duration seem unused now after we dropped the config. We can remove the constants

music_assistant/providers/airplay/protocols/airplay2.py
```diff
@@ -55,23 +93,55 @@ def _cli_loglevel(self) -> int:
...
+        Streaming session is now established.
+
+        Log the actual session establishment duration and check against configured value.
+        """
+        if self._connected_ts is not None:
+            """We have previously connected, so we ignore this call"""
+            return
+        self._connected_ts = time.time()
+        assert self._session_establishment_ms is not None  # for type checker
+        self.logger.info(
```
MarvinSchenkel: Do we really need this at info level?

MarvinSchenkel (changes requested):
Few minor things, other than that, I think we are good to merge 🎉

music_assistant/providers/airplay/protocols/airplay2.py
bradkeifer: Because when you did the testing, you thought 700 was the limit and 600 gives us some protection. Will add comment to that effect.

music_assistant/providers/airplay/protocols/airplay2.py
bradkeifer: No. Now that we have the warning if outside desired range, it's not required at info. I will move to debug.

MarvinSchenkel (approved):
Amazing stuff @bradkeifer 👏 . I think this is ready for beta :)

PR #3784, opened 2026-04-24
Fix YTMusic stream format selection

greenmansuperhero:
### This fixes a YTMusic provider bug in stream format selection.

**Problem:**
The current code uses an invalid next(...) expression when selecting the best audio format. In failure paths this can leak StopIteration through the async setup path and break provider setup instead of raising the intended UnplayableMediaError.

**Change:**
- replace the broken next(...) expression with a proper next(generator, None) call
- explicitly check for a missing result before raising No stream formats found

**Why:**
This is a logic bug in the provider itself and is independent of deployment environment. It was reproducible during YTMusic setup and caused setup failures instead of the expected provider error handling.


MarvinSchenkel (approved):
Thanks @greenmansuperhero !

PR #3788, opened 2026-04-25
Enrich Sendspin metadata with track number, year, album artist, and artist artwork

### LINKED ISSUE #5369: Track number and year aren't sent to SendSpin clients

2.8.5

Home Assistant OS Addon

### The problem

MA currently [hard codes track and year](https://github.com/music-assistant/server/blob/28c66f17c5a87bc4d1da7de41ef0ca0eac38e177/music_assistant/providers/sendspin/player.py#L843) to `None` and thus they're aren't sent to SendSpin clients.
Fixing this might also require changing the `PlayerMedia` model since it [doesn't contain those fields](https://github.com/music-assistant/models/blob/main/music_assistant_models/player.py#L102).

### How to reproduce

* Connect a SendSpin client with a metadata role
* Play a track
* The `metadata` object in `server/state` has `track` and `year` set to `null`.

### Music Providers

I've only tried this with a file system music provider, but the source code shows that this problem applies to all sources.

### Player Providers

This is only relevant to SendSpin providers.

### Full log output

[d5369777_music_assistant_2026-04-22T15-04-11.127Z.log](https://github.com/user-attachments/files/26974913/d5369777_music_assistant_2026-04-22T15-04-11.127Z.log)

### Additional information

N/A

_No response_

Home Assistant OS

- [x] Yes

OnFreund:
## Summary

Adds missing metadata fields to the Sendspin player and fixes artist artwork not being sent.

Fixes https://github.com/music-assistant/support/issues/5369
Replaces #2825

### Missing metadata fields (track number, year, album artist)

The `Metadata` object sent to Sendspin clients had `track`, `year`, and `album_artist` hardcoded to `None`. This PR populates them:

- **`track_number`** and **`year`**: read directly from the `QueueItem` — `Track.track_number` is already on the queue item, and `ItemMapping.year` is preserved from the full `Album` when the queue item is created. Zero extra cost.
- **`album_artist`**: fetched via `mass.music.get_library_item_by_prov_id(MediaType.ALBUM, ...)` — a pure SQLite library lookup with no external API call. Gracefully absent (`None`) for tracks not in the local library.

This is the approach suggested in #2825 as an alternative to the heavy `get_item_by_uri` call that caused that PR to be shelved.

### Artist artwork fix

`_send_artist_artwork` was not sending artwork in practice because artists in a `QueueItem` are simplified to `ItemMapping` objects, which often have `image=None` (the image is only preserved if the artist's THUMB was already in `metadata.images` at queue-creation time). Additionally, the fallback code path was dead — `artist_artwork_url` was set but the subsequent guard skipped the actual send.

The fix uses the same library DB lookup to retrieve the full `Artist` object (with reliable, up-to-date artwork), falling back to the `ItemMapping` if the artist is not in the library.

### Clearing metadata on media unload

When `current_media` becomes `None` (e.g. queue ends), the metadata role was cleared but the artwork roles were not — leaving stale artwork on the client. Both album and artist artwork are now cleared, and the URL cache is reset so the next track's artwork is always re-sent.

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -50,7 +50,7 @@ from music_assistant_models.enums import (
     PlayerType,
     RepeatMode,
 )
-from music_assistant_models.media_items import Album, Artist, is_track
+from music_assistant_models.media_items import Album, Artist, ItemMapping, is_track
 from music_assistant_models.player import DeviceInfo
 from PIL import Image
 
@@ -751,7 +751,7 @@ class SendspinPlayer(SendspinBasePlayer):
     async def _send_artist_artwork(self, current_item: QueueItem) -> None:
         """Send artist artwork to the sendspin group."""
         artist_artwork_url: str | None = None
-        fetch_target: Artist | None = None
+        fetch_target: Artist | ItemMapping | None = None
 
         if current_item.media_item is not None and is_track(current_item.media_item):
             artists = current_item.media_item.artists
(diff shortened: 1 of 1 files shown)


maximmaxim345 (approved):
LGTM.
Thanks @OnFreund !

PR #3793, opened 2026-04-27
Fully implement hide in the UI

OzGav:
Following on from this https://github.com/music-assistant/frontend/pull/1711

The "Hide in UI" setting was only loosely applied: a few server-rendered config pickers (sync-group / universal-group members, members filter, plus several provider connectors) still listed hidden players as candidates. This PR finishes the coverage so hidden players are filtered from every player picker, while still appearing if they're already the saved selection (so existing group members / connector targets remain visible and changeable). The filter is centralized as two new kwargs on `PlayerController.all_players` (`exclude_hidden, keep_player_ids`) so each call site is just one line.

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -114,7 +114,7 @@ from .helpers import AnnounceData, handle_player_command, wait_for_power_on
 from .protocol_linking import ProtocolLinkingMixin
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterator
+    from collections.abc import Callable, Iterable, Iterator
 
     from music_assistant_models.config_entries import (
         ConfigEntry,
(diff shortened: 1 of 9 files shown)


marcelveldt:
I am not agreeing on this. The setting is called "hide in UI" and not "hide in backend" - if you want that, you need to disable the player.

music_assistant/controllers/players/controller.py
```diff
@@ -259,6 +262,12 @@ def all_players(
         :param return_disabled [bool]: Include disabled players.
         :param provider_filter [str]: Optional filter by provider lookup key.
         :param return_protocol_players [bool]: Include protocol players (hidden by default).
+        :param exclude_hidden [bool]: When True, exclude players with hide_in_ui=True
+            from the result. Use this for UI candidate pickers.
+        :param keep_player_ids: Player id(s) that must always be returned even when
```
marcelveldt, line 267: this is a mess and this alone is why we should not do this at all

marcelveldt (changes requested):

music_assistant/controllers/players/controller.py
OzGav: OK closing this PR and will open one to adjust the description for this setting

PR #3795, opened 2026-04-27
Add Sonic Analysis audio-analysis provider (CLAP-driven scalars + embedding)

chrisuthe:
This adds a new audio analysis provider, `sonic_analysis`, that runs Microsoft's CLAP model locally on the host CPU to populate the `audio_analysis` tables for both the background scan and live playback. Alongside the usual measurement features (BPM, key, loudness, brightness, etc.) it derives soft perceptual scalars (danceability, valence, arousal, instrumentalness, acousticness) from CLAP zero-shot inference, and persists the raw 1024-dim CLAP audio embedding to `extra_data["clap_embedding"]` so downstream plugins can build their own search/similarity indexes from one place in SQLite.

Everything is on-device. No external services required.

> **Scope note:** The three `sonic_analysis/*` API commands and the read-side `AudioAnalysisController` helpers they originally relied on have moved to PR #3851, where they are generalized to `audio_analysis/*` for use by all AA providers. This PR is now provider-only — no central controller surface area.

## What it does

- Full `AudioAnalysisData` for files and live sessions, off a single decode per track.
- Measurements (always populated): `energy`, `brightness`, `harmonic_complexity`, `roughness`, `rhythmic_regularity`, `loudness_integrated`, `loudness_range`, `true_peak`, plus `rms_energy` / `spectral_centroid` time series.
- Soft perceptual scalars (Platt-calibrated to a 0–1 probability from CLAP zero-shot logits): `danceability`, `instrumentalness`, `valence`, `arousal`, `acousticness`. 5-fold CV accuracy on a 50-track validation set ranges 0.71 to 0.91 depending on attribute.
- Raw 1024-dim CLAP embedding written to `audio_analysis.extra_data["clap_embedding"]` as L2-normalised f32 JSON. Reuses the embedding already produced for scalar inference, so no extra model cost. Roughly 10KB per row (≈10MB per 1k tracks).
- Configurable sampling preset: `fast` (1 window, default), `balanced` (3), `thorough` (8).

## Vendored vs Actually reviewable

The PR is large (~6k lines, 35 files), but most of that is vendored model code. Suggested reading order:

1. **`music_assistant/providers/sonic_analysis/__init__.py`** (713 lines) — the actual provider. Implements the `AudioAnalysisProvider` contract and the live-PCM dispatch path.
2. **`music_assistant/providers/sonic_analysis/helpers.py`** (404 lines) — pure helpers (window selection, resampling, feature extraction). Heavily unit-tested.
3. **`music_assistant/providers/sonic_analysis/clap_prompts.py`** (147 lines) — calibrated prompt set + Platt coefficients used to derive the soft scalars.
4. **`music_assistant/providers/sonic_analysis/manifest.json`** — config schema.
5. **`tests/providers/sonic_analysis/`** — 13 test files covering helpers, the live dispatch path, finalize/integration, prompt loading, background model load, etc.
6. **`music_assistant/providers/sonic_analysis/vendored_clap/`** — copy of [microsoft/CLAP](https://github.com/microsoft/CLAP). I do **not** expect a line-by-line review here; modifications are flagged with `# MA MOD:` and explained below.

`NOTICE`, `pyproject.toml`, `requirements_all.txt`, and `scripts/precompute_clap_prompt_embeddings.py` round out the change.

## Vendored CLAP — what I changed and why

`music_assistant/providers/sonic_analysis/vendored_clap/` is a copy of [microsoft/CLAP](https://github.com/microsoft/CLAP) (MIT). All MA-side modifications are flagged with a `# MA MOD:` comment so a re-vendor stays mechanical:

- `clap_wrapper.py`: replace `torchaudio.load` with `librosa.load` (avoids the `torchcodec` / ffmpeg shared-lib coupling introduced in torch 2.11+); accept pre-decoded tensors via `preprocess_audio_from_tensor` so we can share the live PCM buffer; add a `text_enabled` flag to skip the GPT2 download; migrate `tokenizer.encode_plus(...)` to `tokenizer(...)` for transformers v5.
- `models/clap.py`: `skip_text_encoder` on `CLAP` and `skip_text_model` on `TextEncoder`, so we never instantiate the text head when text search is disabled.
- Pruned: `clapcap` (CLAP captioning) and the 2022 audio model are removed since neither is used. Only the 2023 audio config is shipped.

The third-party `LICENSE` is consolidated into the repo's root `NOTICE` per maintainer feedback. `vendored_clap/README.md` documents every modification for future re-vendoring audits.

- First time the provider is enabled it downloads ~300MB of CLAP audio weights into the HuggingFace cache.

## Added requirements

```
+ huggingface-hub==1.12.0       # pulls CLAP weights from HF Hub
+ PyYAML==6.0.3                 # vendored CLAP config parsing
+ torchlibrosa==0.1.0           # HTSAT audio frontend
+ transformers==5.6.2           # vendored CLAP imports (CVE-2026-1839 fix; encode_plus -> __call__ migrated)
```


music_assistant/providers/sonic_analysis/vendored_clap/LICENSE
```diff
@@ -0,0 +1,21 @@
+    MIT License
```
MarvinSchenkel: There's a NOTICE.md in the root of the repo where we can put third party licenses.

chrisuthe: Cleaned this up

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,864 @@
...
+            return
+        self.logger.info(
+            "CLAP model loaded; %d prompt pairs ready",
+            len(self._clap_prompt_order),
+        )
+
+    async def loaded_in_mass(self) -> None:
+        """Register API commands once the provider is live."""
+        self._unregister_handles = [
+            self.mass.register_api_command("sonic_analysis/status", self._handle_status),
```
MarvinSchenkel: These API endpoints sounds like they do something that could/should be centralized in the AA Controller so it's generic for all AA providers?

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,864 @@
...
+            try:
+                unregister()
+            except Exception as err:
+                self.logger.debug("API command unregister failed: %s", err)
+        self._unregister_handles.clear()
+        self._clap_model = None
+        self._clap_text_embeddings = None
+        await super().unload(is_removed)
+
+    async def _handle_status(self) -> dict[str, Any]:
```
MarvinSchenkel: Let's do this centrally in the AA controller in a follow up PR

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,864 @@
...
+            self.domain
+        )
+        return {
+            "provider_loaded": True,
+            "clap_model_loaded": self._clap_model is not None,
+            "analyzed_tracks_count": analyzed_tracks_count,
+            "analysis_version": self.analysis_version,
+        }
+
+    async def _handle_analyzed_tracks(
```
MarvinSchenkel: Let's do this centrally in the AA controller in a follow up PR

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,864 @@
...
+                artists = ", ".join(a.name for a in getattr(t, "artists", []) or [])
+                return {"item_id": item_id, "name": t.name, "artist": artists}
+            except Exception as err:
+                self.logger.debug("Failed to resolve track %s/%s: %s", provider, item_id, err)
+                return {"item_id": item_id, "name": "(unknown)", "artist": ""}
+
+        items = list(await asyncio.gather(*[_resolve(iid, prov) for iid, prov in entries]))
+        return {"total": total, "offset": offset, "limit": limit, "items": items}
+
+    async def _handle_export_analysis(
```
MarvinSchenkel: Let's do this centrally in the AA controller in a follow up PR

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,864 @@
...
+        :param session_id: The analysis session ID.
+        :param pcm_chunk: Raw PCM audio data.
+        """
+        if session_id not in self._sessions:
+            return
+        session = self._sessions[session_id]
+        assert isinstance(session, SonicSessionData)
+        session.pcm_buffer.extend(pcm_chunk)
+        af = session.audio_format
+        while len(session.pcm_buffer) >= session.block_bytes:
```
MarvinSchenkel: Can you explain why we need a while loop here? Why not just accumulate pcm chunks until you have gathered 10s and then process the entire block at once?

chrisuthe: We don't, because pcm_chunk will never be > 10 seconds (it'll always be 1 second) so while is doing exactly the same thing as an if, which tracks how you're thinking of it. This was AI coding being overly defensive and ignoring the contract, trying to guard in case pcm_chunk got big (which, it won't). I'll shift this to an if statement that makes it more clear what we're expecting.

chrisuthe: see e871a08

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Going to use https://github.com/music-assistant/server/pull/3851 to do that.

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Going to use https://github.com/music-assistant/server/pull/3851 to do that.

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Going to use https://github.com/music-assistant/server/pull/3851 to do that.

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,713 @@
...
+                "Invalid audio format for session %s (sample_rate=%d, bit_depth=%d, channels=%d)"
+                " — skipping analysis",
+                session_id,
+                audio_format.sample_rate,
+                audio_format.bit_depth,
+                audio_format.channels,
+            )
+            return False
+        target_starts: list[int] = []
+        if self._clap_model is not None and streamdetails.duration:
```
MarvinSchenkel: Should we short circuit on tracks without streamdetails.duration? We now run the analysis anyway and store an AudioAnalysisData object without any useful info. Next time the analysis runs, it does not run the analysis because it already exists in the database.

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,713 @@
...
+                ConfigValueOption("Fast (1 window)", CLAP_SAMPLING_FAST),
+                ConfigValueOption("Balanced (3 windows, 2.4x CPU)", CLAP_SAMPLING_BALANCED),
+                ConfigValueOption("Thorough (8 windows, 6.6x CPU)", CLAP_SAMPLING_THOROUGH),
+            ],
+            required=False,
+        ),
+    )
+
+
+def select_clap_window(audio: np.ndarray, source_sr: int) -> np.ndarray | None:
```
MarvinSchenkel: This seems unused (dead code).

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -0,0 +1,713 @@
...
+        return audio[skip_n : skip_n + window_n]
+    if n >= window_n:
+        start = (n - window_n) // 2
+        return audio[start : start + window_n]
+    if n >= source_sr:
+        return audio
+    return None
+
+
+def select_clap_windows(audio: np.ndarray, source_sr: int, n_windows: int) -> list[np.ndarray]:
```
MarvinSchenkel: This seems unused (dead code).

music_assistant/providers/sonic_analysis/clap_prompts.py
```diff
@@ -0,0 +1,147 @@
...
+            "SCALAR_PROMPT_PAIRS hash (%s). The Platt calibration "
+            "coefficients were fit against different prompts; CLAP "
+            "scalar outputs will be systematically biased. Re-fit "
+            "CALIBRATION and update CALIBRATION_PROMPTS_HASH.",
+            CALIBRATION_PROMPTS_HASH,
+            actual,
+        )
+
+
+validate_calibration_freshness()
```
MarvinSchenkel: Can we remove this call? Importing this module will run the function automatically in this case.

MarvinSchenkel (changes requested):
Few minor things, almost there 🙏

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: cleaned up! TY!

music_assistant/providers/sonic_analysis/clap_prompts.py
chrisuthe: Yes we can, Cleaned up, TY!

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: cleaned up, TY!

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Added a guard earlier, right after check for model. We shouldn't bother trying any part of it, even if it comes back around trying to stamp the audio of a radio stream etc is bad news bears for actually being useful.

MarvinSchenkel (approved):
Amazing job @chrisuthe. I CLAP my hands for you 👏  ;-)

chrisuthe:
> Amazing job @chrisuthe. I CLAP my hands for you 👏 ;-)

Pun of the Month award!

OzGav:
I think this https://github.com/music-assistant/server/pull/2153 has been superseded now hasn't it?

OzGav:
Also we need some docs to explain this. I have an audio analysis section in the beta docs now. Here is an example https://beta.music-assistant.io/audio-analysis/loudness-analysis/

PR #3798, opened 2026-04-27
WiiM: Set the default max sample rate to 96kHz

teancom:
I was seeing weird bugs with 192kHz where tracks would restart from the beginning every ~52 seconds. After doing a bunch of debugging, dropping the sample rate "fixed" it. By not dropping the max_sample_rate we allow for testing (presumably against a newer version of either the WiiM python library or against newer firmware versions of the players themselves) and for people to use if they want to. 96kHz is a reasonable default anyway, for the vast majority of people.

### DIFF
--- a/music_assistant/providers/wiim/player.py
+++ b/music_assistant/providers/wiim/player.py
@@ -126,7 +126,7 @@ class WiimPlayer(Player):
         return [
             create_sample_rates_config_entry(
                 max_sample_rate=192000,
-                safe_max_sample_rate=192000,
+                safe_max_sample_rate=96000,
                 max_bit_depth=24,
                 safe_max_bit_depth=24,
             ),


teancom:
Hmm. For clarification, it looks like MA isn't letting me re-enable higher-than-safe. At least, I can enable them on a per-player basis but it is still only picking 96kHz on playback start. Which, it *works* (which is better than before) it just doesn't match my mental model of how the max vs. safe max is supposed to work.

MarvinSchenkel:
I was under the impression all WiiM devices support up to 192/24? Which WiiM device is causing issues? 

Oh, and:
`max_xxx` -> Max selectable
`safe_max_xx` -> Default values enabled

Might be a pre-existing config in your setup that is causing this. I just tested this PR on a clean install and it works as expected.

teancom:
I have a WiiM Pro, firmware is 4.8.805729 (latest, according to the app).

And the *gui* looks great with that config - does exactly what you'd expect. But selecting the higher values didn't make it actually *use* them the way it did before I changed the safe_max. Even with 172 and 196 enabled, it was still only connecting at 96kHz. I haven't nailed down exactly what's going on there though. :-/

MarvinSchenkel:
Could be the content itself. Anyway, this PR looks good to me. Thanks :)

MarvinSchenkel (approved):

PR #3801, opened 2026-04-27
Bump wiim SDK to 0.1.4 to fix track transition tracking

teancom:
The 0.1.1 SDK silently drops AVTransport metadata events whose payload is JSON rather than escaped XML DIDL — it logs "is raw XML in event", fails to parse, and writes current_track_info["uri"] = None. This makes device.current_media.uri stale after the WiiM auto-advances on SetNextAVTransportURI, which in turn prevents Music Assistant's queue tracker from advancing current_index past the first track.

0.1.4 includes 0bfa84c ("Add grouping support and fix metadata parsing"), which detects JSON payloads and adds a trackURIs[0] -> res fallback, so transitions populate current_track_info correctly.

0.1.4 also drops the public current_track_uri / current_position / current_track_duration / current_track_info attributes (12d058f) and changes async_join_group to take list[str] instead of a single UDN. This commit updates the provider for the new API:

- Read URI/position via device.current_media.uri / .position
- Batch follower UDNs into a single async_join_group call
- Drop the now-irrelevant attribute fixtures from the test mock

### DIFF
--- a/music_assistant/providers/wiim/manifest.json
+++ b/music_assistant/providers/wiim/manifest.json
@@ -5,7 +5,7 @@
   "name": "WiiM",
   "description": "Stream music to WiiM devices.",
   "codeowners": ["@davidanthoff", "@music-assistant"],
-  "requirements": ["wiim==0.1.1"],
+  "requirements": ["wiim==0.1.4"],
   "documentation": "https://music-assistant.io/player-support/wiim/",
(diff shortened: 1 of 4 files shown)


MarvinSchenkel:
Thanks! I dropped the backport label as the WiiM provider is not in stable yet ;-)

MarvinSchenkel (approved):

PR #3805, opened 2026-04-28
Emby Music Provider: add on played event handler

hatharry:


OzGav:
Thanks @hatharry . Any changes required for the docs as a result of this addition?

MarvinSchenkel (approved):

PR #3814, opened 2026-04-30
Decrease buffer warmup duration to 8s

MarvinSchenkel:
This PR decreases the warmup duration from 45s (crossfade duration) to 8s. This should be enough for slow sources to fill the crossfade buffer.

This prevents smart fades from being skipped when seeking towards the end of the song. Note, a smart fade is still cut short when seeking within 45s (crossfade buffer) + 8s (warmup duration) before the track ends.


OzGav (approved):
LGTM!

PR #3815, opened 2026-04-30
Add option to use local genre metadata only when available

OzGav:
Adds a new "Use local genre metadata only when available" toggle under the metadata controller config (default off, so existing behaviour is preserved). When enabled, online metadata providers will not add genres to items that already have a genre from a local source such as a file tag or NFO file. Items with no local genre still receive genres from online providers as usual.

Implemented by masking the genres field on each online metadata response with dataclasses.replace before merging, so the existing set-union behaviour for other fields is unchanged.

### DIFF
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -10,6 +10,7 @@ import random
 import urllib.parse
 from base64 import b64encode
 from contextlib import suppress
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
Thanks @OzGav , looks good to me

PR #3436, opened 2026-03-19
Use /playlists/{id}/items endpoint (Spotify Feb 2026 API change)

Yipsh:
## Summary

Spotify removed `GET/POST/DELETE /playlists/{id}/tracks` in their [February 2026 API changes](https://developer.spotify.com/documentation/web-api/references/changes/february-2026) and replaced them with `/playlists/{id}/items`.

This causes 403 Forbidden errors on all playlist operations (fetch tracks, add to playlist, remove from playlist). Individual track playback is unaffected.

## Changes

4 endpoint references updated in `providers/spotify/provider.py`:
- `get_playlist_tracks` (line 591): GET `/tracks` → `/items`
- `add_playlist_tracks` (line 685): POST `/tracks` → `/items`
- `remove_playlist_tracks` (line 693): GET `/tracks` → `/items` (position lookup)
- `remove_playlist_tracks` (line 700): DELETE `/tracks` → `/items`

The response format is unchanged — the new `/items` endpoint still returns objects with a `track` field, so no parser changes needed.

## Note

There are additional deprecated endpoints in the Spotify provider that will need migration in a follow-up (`/playlists/{id}/followers`, `/users/{id}/playlists`, `me/albums`, `me/tracks`, etc. → consolidated `me/library`). This PR focuses only on the playlist fix since it's the most impactful breakage.

Fixes music-assistant/support#4978

PR #3818, opened 2026-04-30
Fix Spotify playlists failing when track count is a multiple of 50

gitviola:
## What's broken

Playlists with exactly N*50 tracks (notably all Daily Mix variants at 50 tracks) fail to play, with "No playable items found" in the UI. Discover Weekly and Release Radar (30 tracks each) still work, which is what masked the bug for a while.

## Why

`PlaylistController._get_provider_playlist_tracks` pages through `provider.get_playlist_tracks(page=N)` until it gets an empty page back. For a 50-track playlist, page 0 returns a full 50 items, so the controller asks for page 1 (offset=50). Spotify's API has started responding 5xx for `offset == total` instead of returning the documented empty items list. `@throttle_with_retries` then sleeps 30s, 5 times, before raising `RetriesExhausted`, which surfaces as "No playable items found".

Smaller playlists don't trip this because page 0 already returns < 50 items, so the controller loop stops on its own.

The failing log looks like:

```
DEBUG handling get data .../playlists/.../items with kwargs {'limit': 50, 'offset': 50, ...}
INFO  Attempt 1/5 failed: 
INFO  Retrying in 30.0 seconds...
```

(empty error body because the 5xx path raises `ResourceTemporarilyUnavailable(backoff_time=30)` with no message.)

## What the PR does

A single `_get_paginated_meta` probe (`limit=1`) returns both the etag (cache checksum) and the playlist total in one cached request. `get_playlist_tracks` now returns `[]` early when `offset >= total`, avoiding the page request that triggers Spotify's 5xx. The same guard is applied in `_get_all_items` so other paginated endpoints (`me/tracks`, `me/albums`, …) are covered when their total is an exact multiple of 50.

## How this was tested

Verified live by bind-mounting the patched `provider.py` into a running Music Assistant container, restarting, and playing different playlists from the UI:

- Daily Mix (50 tracks): plays, queue populates with all 50 tracks, no retry storm in the log, no `offset=50` request made.
- Larger playlists (more than 50 tracks): play fully, all pages fetched correctly, no retry storm.
- Discover Weekly / Release Radar (30 tracks): still works as before.

Static checks pass locally:

```
ruff check  -> All checks passed!
ruff format -> 1 file already formatted
mypy        -> Success: no issues found in 1 source file
```

This bug is also present in `stable`, so it's a candidate for `backport-to-stable`.

## Result

Playlist is loading properly and songs are playing after applying the fix.

image

## Related

- Reproduces the symptom reported in https://github.com/orgs/music-assistant/discussions/5304.
- The 5xx-at-`offset==total` behaviour matches what other Spotify API users have observed: https://community.spotify.com/t5/Spotify-for-Developers/Web-API-502-Bad-Gateway-for-top-items-endpoints-when-offset-and/td-p/5897842

## Not the same as

To avoid confusion with other recent Spotify-provider fixes:

- music-assistant/support#5114 / #4978 — the Feb 2026 `/playlists/{id}/tracks` → `/items` endpoint rename. Already fixed by #3436 and merged in `dev`. This PR sits on top of that change; without #3436, the failing request would be a 403 instead of a 5xx.
- music-assistant/support#5360 — Spotify catalog endpoints (artist albums, top tracks, search) returning empty for dev apps without Extended Quota. Different code paths (`get_artist_albums`, `get_artist_toptracks`), addressed in PR #3762.

### DIFF
--- a/music_assistant/providers/spotify/provider.py
+++ b/music_assistant/providers/spotify/provider.py
@@ -606,20 +606,27 @@ class SpotifyProvider(MusicProvider):
         page_size = 50
         offset = page * page_size
 
-        # Get etag for caching
-        cache_checksum = await self._get_etag(uri, limit=1, offset=0, use_global_session=use_global)
+        meta = await self._get_paginated_meta(uri, limit=1, offset=0, use_global_session=use_global)
+        cache_checksum = meta["etag"]
+        total = meta["total"]
+
+        # Spotify has started returning 5xx for offset >= total on some
+        # playlists (notably algorithmic ones like Daily Mix). The retry
+        # storm that follows surfaces as "No playable items found".
+        if total and offset >= total:
+            return result
 
         spotify_result = await self._get_data_with_caching(
             uri, cache_checksum, limit=page_size, offset=offset, use_global_session=use_global
         )
-        total = spotify_result.get("total", 0)
+        page_total = spotify_result.get("total", total)
         items = spotify_result.get("items", [])
         # playlists/{id}/items is transitioning from item["track"] to item["item"]
         # during Spotify's Feb 2026 rollout, so accept either shape.
         for index, item in enumerate(items, 1):
             # Spotify wraps/recycles items for offsets beyond the playlist size,
             # so we need to break when we've reached the total.
-            if (offset + index) > total:
+            if page_total and (offset + index) > page_total:
                 break
             track_data = item and (item.get("item") or item.get("track"))
             if not (track_data and track_data.get("id")):
@@ -1126,9 +1133,15 @@ class SpotifyProvider(MusicProvider):
         """Get all items from a paged list."""
         limit = 50
         offset = 0
-        # do single request to get the etag (which we use as checksum for caching)
-        cache_checksum = await self._get_etag(endpoint, limit=1, offset=0, **kwargs)
+        # single request to fetch the etag (used as cache checksum) and total
+        meta = await self._get_paginated_meta(endpoint, limit=1, offset=0, **kwargs)
+        cache_checksum = meta["etag"]
+        total = meta["total"]
         while True:
+            # Avoid requesting beyond the known end. Spotify can return 5xx
+            # for offset >= total on some endpoints (e.g. algorithmic playlists).
+            if total and offset >= total:
+                break
             result = await self._get_data_with_caching(
                 endpoint, cache_checksum=cache_checksum, limit=limit, offset=offset, **kwargs
             )
@@ -1158,11 +1171,11 @@ class SpotifyProvider(MusicProvider):
         )
         return result
 
-    @use_cache(120, allow_bypass=False)  # short cache for etags (subsequent calls use cached data)
-    async def _get_etag(self, endpoint: str, **kwargs: Any) -> str | None:
-        """Get etag for api endpoint."""
+    @use_cache(120, allow_bypass=False)  # short cache: subsequent calls reuse cached data
+    async def _get_paginated_meta(self, endpoint: str, **kwargs: Any) -> dict[str, Any]:
+        """Get etag and total item count for a paginated api endpoint."""
         _res = await self._get_data(endpoint, **kwargs)
-        return _res.get("etag")
+        return {"etag": _res.get("etag"), "total": _res.get("total", 0)}
 
     @throttle_with_retries
     async def _get_data(self, endpoint: str, **kwargs: Any) -> dict[str, Any]:
--- a/music_assistant/providers/spotify/provider.py
+++ b/music_assistant/providers/spotify/provider.py
@@ -606,9 +606,9 @@ class SpotifyProvider(MusicProvider):
         page_size = 50
         offset = page * page_size
 
-        meta = await self._get_etag(uri, limit=1, offset=0, use_global_session=use_global)
-        cache_checksum = meta["etag"]
-        total = meta["total"]
+        # Get etag for caching
+        cache_checksum = await self._get_etag(uri, limit=1, offset=0, use_global_session=use_global)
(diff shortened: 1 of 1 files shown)


music_assistant/providers/spotify/provider.py
```diff
@@ -606,20 +606,27 @@ async def get_playlist_tracks(self, prov_playlist_id: str, page: int = 0) -> lis
...
+        # playlists (notably algorithmic ones like Daily Mix). The retry
+        # storm that follows surfaces as "No playable items found".
+        if total and offset >= total:
+            return result
 
         spotify_result = await self._get_data_with_caching(
             uri, cache_checksum, limit=page_size, offset=offset, use_global_session=use_global
         )
-        total = spotify_result.get("total", 0)
+        page_total = spotify_result.get("total", total)
```
OzGav: Any reason for the rename here?

OzGav:
> I tried a tuple first but @use_cache deserializes via parse_value, which mishandles heterogeneous tuples (it would silently turn (etag, 50) into (etag, None) on cache hits, defeating the bounds check after the first call). A dict round-trips cleanly via parse_value's Any fast path.

Can you point me at the parse_value behaviour you’re describing, because if it’s real it affects more than this PR

music_assistant/providers/spotify/provider.py
gitviola: No particular reason. I just renamed this one and also the method name back to what it was before :)

gitviola:
@OzGav and this is what happens when someone isn't native to the language and thinks their $100/m claude code subscription can help fix a bug and improve something.. 🥺😅

I reviewed the code before committing and tested it and the changes looked reasonable enough to raise this PR. But now going deeper, I'm not that sure anymore. If you want I can remove that specific change. It insists that it is a bug but I'm honestly not that python hands-on that I can properly explain it.


> The parse_value bug is real: yes, verified by running it against [`parse_value`](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/helpers/api.py#L469-L478). The PR's actual code (dict return) works correctly. The tuple alternative I described in the PR body really does misbehave. The [deezer](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/deezer/__init__.py#L251)/[yandex](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/yandex_music/provider.py#L1649)/[kion](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/kion_music/provider.py#L1093) callsites I named really would crash on a cache hit if they ever produced a `None` in slot 3, because their callers do `for a, b, c in result:` (I checked: [deezer:409](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/deezer/__init__.py#L409), [yandex:1606](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/yandex_music/provider.py#L1606), [kion:1050](https://github.com/music-assistant/server/blob/33dd73ee95f1eff375a0fd6902eba39c86c4ab1c/music_assistant/providers/kion_music/provider.py#L1050)). But those are pre-existing latent bugs in those providers, completely separate from this PR.

OzGav:
Thanks for being upfront about it. For clarity, the tuple/parse_value reasoning shouldn’t be in this PR’s description at all. Nothing in the diff touches parse_value or the cache layer, so it reads like justification for a change that isn’t here. The dict return stands on its own; etag + total seems a reasonable fix without needing a “tuples are broken” rationale.
I’d suggest just dropping that paragraph so the PR scope is clean.
Separately — if parse_value really does have a problem,  then that’s worth a standalone issue so Marcel and Marvin can see it on its own merits.

Please also note the new [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md)

gitviola:
@OzGav perfect, just pushed and re-tested by again copying the new code into the container (`cp /root/provider.py {my container id}:/app/venv/lib/python3.13/site-packages/music_assistant/providers/spotify/provider.py`) and restarting.

### Before:
image


### After:
image

OzGav:
Looks like the split into `_get_etag` and `_get_total` means we now hit `_get_data` twice with the same `limit=1, offset=0` kwargs on the first call — once for each helper — where the previous version did it in a single call. Can you confirm that's the case? If so, please combine them back into a single fetch so we're not doubling the probe requests against Spotify.

gitviola:
@OzGav I missed that I re-introduced the duplicated call when trying to avoid introducing new namings/variables. But here it made sense. I was choosing between the original `_get_paginated_meta` naming vs. just a simple `_get_meta`. But I kept the longer name because I found it more descriptive (we get meta data of a paginated endpoint, so the expectation is that we return something related to that). Does that sound reasonable or do you have a preference?

gitviola:
Saw that #3762 got merged, so I just updated my branch and re-tested. Fix is still working as expected.

OzGav:
To answer your question we dont want any unnecessary API calls. Naming I dont think I have a preference

OzGav:
Minor nit. Description still says "fetched via a small _get_total helper, cached the same way as the existing _get_etag". Looks like leftover wording from the earlier two-helper iteration. Worth tweaking so the description matches what the code does.

Otherwise LGTM!

OzGav (approved):

OzGav:
Also you need to run pre-commit as there is some formatting thing to fix.

gitviola:
@OzGav formatting fix pushed and PR description updated 😊

PR #3821, opened 2026-04-30
Stream PCM to audio analysis providers during background scan

chrisuthe:
## Summary

Shifts the audio-analysis background scan from per-provider `analyze_file()` (each provider opens its own decoder) to a single decode-once-fan-out streaming PCM path, so providers only need to implement live-streaming hooks. Adds an optional `post_analysis` lifecycle hook for filesystem side effects (replaces loudness's previous in-`analyze_file` ReplayGain tag-write), and adds a configurable `background_scan_concurrency` knob.

- Controller now decodes a track once via ffmpeg in the file's source-native PCM format and fans chunks to every AA provider that needs that track
- Per-track 5-minute wall-clock budget replaces the per-chunk timeout for background; live playback's 1.0s per-chunk timeout is unchanged
- `analyze_file()` removed from the base class and from `loudness_analysis`; `loudness_analysis` now writes ReplayGain tags via the new `post_analysis` hook (which fires for both background scans AND live filesystem playback)

## design decisions

| Decision | Setting |
|---|---|
| Background PCM rate | Flat-out (slowest provider naturally caps via the existing chunk-distribution barrier) |
| Per-track timeout | 300 s wall-clock budget |
| PCM format | Source-native (whatever the file's native format is) |
| Concurrency | Configurable `background_scan_concurrency`, default 1 |
| `_finalize` return type | `AudioAnalysisData \| None` — analysis (or None to skip), used by base class to drive `post_analysis` |
| Provider contract | Codified in `AudioAnalysisProvider` class docstring — \"`process_pcm_chunk` MUST `await` its work; per-track-bounded background tasks OK; no cross-track work\" |

## changes

Two intentional behavior changes worth flagging for reviewers:

1. **Provider exceptions in `process_pcm_chunk` now evict the provider** from the session (previously logged but the provider stayed in). Spec rationale: a provider that raises can't keep processing; this also makes the live and background paths consistent.

2. **`_finalize` exceptions are now caught and logged ERROR** by the base class wrapper instead of propagating. Spec rationale: a misbehaving provider must not leave session state behind or take down the controller. Companion test in `tests/core/test_audio_analysis_controller.py` updated to match.

3. **\"Storage provider unavailable\" no longer aborts the whole scan batch** — only that track is skipped now. Slightly more granular than before.

## tested locally

- 51/51 audio-analysis-related tests pass (`tests/core/test_audio_analysis_controller.py` + `tests/controllers/streams/` + `tests/models/` + `tests/providers/loudness_analysis/` + `tests/providers/smart_fades/` + `tests/integration/`)
- `pre-commit run --all-files` clean across the whole worktree
- `mypy` clean on all touched modules
- Integration test produced `loudness_integrated = -14.1 LUFS` against a 0.3-amplitude 440 Hz mono sine — plausible


tests/fixtures/audio/short_test.wav
MarvinSchenkel, line 1: This file seems unused?

chrisuthe: You are correct sir. No longer needed after testing formats.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -437,24 +544,41 @@ async def _find_tracks_missing_analysis(
...
-        # audio_analysis.item_id holds the provider-native item id,
-        # so join against provider_mappings.provider_item_id (not pm.item_id,
-        # which is the integer library-row id)
+        aa_domains_in = ", ".join(f"'{d}'" for d in aa_provider_domains)
+
+        # Find every (item_id, provider_instance) for filesystem tracks, then
+        # left-join the existing analysis rows to figure out which AA domains
+        # are already covered. The resulting "missing_domains" is the diff.
         query = (
             f"SELECT pm.provider_item_id AS item_id, "
```
MarvinSchenkel, line 535: I think we can get rid of the loop below the query if we inverse the logic. So if we GROUP_CONCAT all the data for which we DONT have a row. So:
```
SELECT ...
FROM pm
CROSS JOIN <possible_aa_provider_domains> possible
WHERE pm.provider_domain in <file system provider domains>
AND NOT EXISTS (
    SELECT 0
    FROM aa
    WHERE aa.aa_provider_domain = possible.aa_provider_domain
    AND aa.provider_instance = pm.provider
    ..etc
)

File: music_assistant/controllers/streams/audio_analysis.py [hunk 2]
```diff
@@ -26,8 +29,14 @@
 CHUNK_PROCESS_TIMEOUT = 1.0
 LOUDNESS_ANALYSIS_DOMAIN = "loudness_analysis"
 BACKGROUND_SCAN_TASK_ID = "audio_analysis_background_scan"
-BACKGROUND_SCAN_BATCH_SIZE = 250
-BACKGROUND_SCAN_SLEEP_BETWEEN_ITEMS = 2.0
+BACKGROUND_PER_TRACK_TIMEOUT_SECONDS = 300
```
MarvinSchenkel, line 33: Can we move the shared constants (by controller.py) into constants.py? e.g. DEFAULT_BACKGROUND_SCAN_CONCURRENCY and CONF_BACKGROUND_SCAN_CONCURRENCY

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -330,100 +344,193 @@ async def get_audio_analysis_version(
...
+        accepted = await self._start_analysis_on_providers(
+            session_key, streamdetails, pcm_format, providers
+        )
+        if not accepted:
+            self.logger.debug("No providers accepted background analysis for %s", session_key)
+            return
+        self._active_sessions[session_key] = accepted
+
+        # 1 second of PCM per chunk
+        chunk_size = pcm_format.sample_rate * (pcm_format.bit_depth // 8) * pcm_format.channels
```
MarvinSchenkel: `AudioFormat` has a property called `pcm_sample_size` you can use.

music_assistant/controllers/streams/controller.py
```diff
@@ -252,6 +254,17 @@ async def get_config_entries(
                 category="generic",
```
MarvinSchenkel: Should we also set this to `audio_analysis`?

MarvinSchenkel (approved):
Awesome job @chrisuthe , thanks 👏

PR #3528, opened 2026-04-01
Remaintain jellyfin

staticdev:
I propose to be added as maintainer to jellyfin provider and get it back under stable status.

PR #3822, opened 2026-04-30
Revert "Remaintain jellyfin (#3528)"

staticdev:
Reverts music-assistant/server#3528

After trying a month I had little to no progress improving the flaky behavior of Jellyfin with Music Assistant and decided to switch to Subsonic Media Server Library. Then I would like to step down from the maintenance of Jellyfin.

I underestimated the amount of work just to make it work properly, adding to the facts discussed in https://github.com/Jc2k/aiojellyfin/issues/8


OzGav:
All good. Thanks for offering.

OzGav (approved):

PR #3823, opened 2026-04-30
Add player filter to scrobble providers

Tommatheussen:
Similar to https://github.com/music-assistant/server/pull/2822, adds player filtering to the scrobble providers.

Needs model update: https://github.com/music-assistant/models/pull/217

ProgressReport playerid is set to the queueid, as this is set to the playerid that owns the queue.


music_assistant/helpers/scrobbler.py
```diff
@@ -168,3 +183,23 @@ async def create_scrobble_users_config_entry(mass: MusicAssistant) -> ConfigEntr
         multi_value=True,
         default_value=[],
     )
+
+
+async def create_scrobble_players_config_entry(mass: MusicAssistant) -> ConfigEntry:
+    """Create a reusable configentry to specify a player list for scrobbling providers."""
+    ma_player_list = mass.players.all_players(return_unavailable=True, return_disabled=False)
+    player_options = [
```
MarvinSchenkel, line 199: Can we sort this by player name?

MarvinSchenkel (changes requested):
Looks good overall, small comment about the configs! Could you please also name everything `player_id` instead of `playerid` so that it is consistent?

MarvinSchenkel:
Just noticed that `subsonic_scrobble` should also be updated with the player filter. I think we have then covered all scrobble providers

Tommatheussen:
> Just noticed that `subsonic_scrobble` should also be updated with the player filter. I think we have then covered all scrobble providers

Huh, I see `subsonic_scrobble` doesn't have any of the config options set. I'm wondering if we should actually add the user and player filtering to the [`get_shared_config_entries`](https://github.com/music-assistant/server/blob/dev/music_assistant/helpers/scrobbler.py#L127) method, so it get added to all scrobblers by default?

MarvinSchenkel:
> > Just noticed that `subsonic_scrobble` should also be updated with the player filter. I think we have then covered all scrobble providers
> 
> Huh, I see `subsonic_scrobble` doesn't have any of the config options set. I'm wondering if we should actually add the user and player filtering to the [`get_shared_config_entries`](https://github.com/music-assistant/server/blob/dev/music_assistant/helpers/scrobbler.py#L127) method, so it get added to all scrobblers by default?

That might even be a better solution indeed.

Tommatheussen:
> > > Just noticed that `subsonic_scrobble` should also be updated with the player filter. I think we have then covered all scrobble providers
> > 
> > 
> > Huh, I see `subsonic_scrobble` doesn't have any of the config options set. I'm wondering if we should actually add the user and player filtering to the [`get_shared_config_entries`](https://github.com/music-assistant/server/blob/dev/music_assistant/helpers/scrobbler.py#L127) method, so it get added to all scrobblers by default?
> 
> That might even be a better solution indeed.

Moved the user and player filtering to the shared scrobble config. Also added that config to subsonic.
We might want to look at how the options are ordered right now, maybe move some of the specific config options (like user ID and keys for last.FM and listenbrainz) to dedicated sections or something?

MarvinSchenkel:
> > > > Just noticed that `subsonic_scrobble` should also be updated with the player filter. I think we have then covered all scrobble providers
> > > 
> > > 
> > > Huh, I see `subsonic_scrobble` doesn't have any of the config options set. I'm wondering if we should actually add the user and player filtering to the [`get_shared_config_entries`](https://github.com/music-assistant/server/blob/dev/music_assistant/helpers/scrobbler.py#L127) method, so it get added to all scrobblers by default?
> > 
> > 
> > That might even be a better solution indeed.
> 
> Moved the user and player filtering to the shared scrobble config. Also added that config to subsonic. We might want to look at how the options are ordered right now, maybe move some of the specific config options (like user ID and keys for last.FM and listenbrainz) to dedicated sections or something?

@OzGav might have some ideas about this :)

Tommatheussen:
For visual reference on the current (with this PR) situation:
image
image

Can't show subsonic as that needs a server first, but it only the suffix, user and player options

OzGav:
I have nothing intelligent to add! This looks good. It will also address

https://github.com/orgs/music-assistant/discussions/4002

MarvinSchenkel:
@Tommatheussen Do you still want to implement a separate section for this PR or would you want to merge this as-is? I am fine either way

Tommatheussen:
I think it's fine as-is for now, we can revisit if there are more options getting added

MarvinSchenkel (approved):
Thanks @Tommatheussen

PR #3824, opened 2026-04-30
Surface Bandcamp label-released performers as their own artists

### LINKED ISSUE #5389: Bandcamp: Artist search returns label/store as artist
[music-assistant.log](https://github.com/user-attachments/files/27043091/music-assistant.log)

2.8.6

Home Assistant OS Addon

### The problem

When searching in the Artist section the Bandcamp provider returns the label/store name and not the actual artist. For example when looking up a release of an artist within the Bandcamp provider, you have to know which label/store the release is located within Bandcamp.

When playing a track from the Bandcamp provider the artist and track are correctly scrobbled to Last.fm

### How to reproduce

- search within Artist and filtered only music provider Bandcamp on Mortaja -> no results
- repeat above with searching on Audiophob and you'll see all the releases on this label including Mortaja

### Music Providers

Bandcamp

### Player Providers

Not a problem related to a player provider

### Full log output

added

### Additional information

no additional info

2026.4.3

Home Assistant OS

- [x] Yes

teancom:

Albums published on a band's page by a different performer (Mortaja on audiophob, The Green Arrows on Analog Africa) used to be filed under the page owner. They now get a synthetic artist ID `{band_id}:{performer_slug}` — scoped per band page so cross-page same-name performers stay distinct. Synthetic IDs that match the page-owner's own slug short-circuit back to the real band on resolve, collapsing any search-time drift before it can persist.

- Closes music-assistant/support#5389

## Types of changes

## Checklist

Depends on https://github.com/ALERTua/bandcamp_async_api/pull/23


teancom:
@ALERTua I've been testing this over the past few days, I think this solves most of the label-as-band issues?

ALERTua:
Search by "Night Moves" album https://hipdozer.bandcamp.com/album/night-moves, which is Apollo Brown, but hosted on the label Hip Dozer

On dev:

image

Clicking the artist name:

image

On this PR:

image

Clicking the artist name:

image

I'm not sure this is consistent.
_____

Search by "the green arrows" artist

On dev:

image


On this PR:

image

image


This is definitely good!

____
In this PR, it is possible to search for an artist who is hosted only on a label. It's impossible on the dev.

So the only issue I see is that within such an album (e.g., Night Moves), there's no way to go to the label page. But is it bad? Only for the music exploration. And I doubt MA is a music exploration-focused app.

Looks good to me. Could be better, but I can't think of a way to make it more comprehensive in the reality of the Bandcamp structure.

The number of Bandcamp workarounds is increasing. I hope they are not planning to change any of the core structures, as this code is becoming less and less readable, alas. Maybe, at some point, we will have to rewrite this integration from scratch. But it's OK.

teancom:
@ALERTua I was focusing only on Mortaja with the first commit, completely forgot to test the other bands. This is now, I believe, an approach that handles all of the various edge cases.

I agree it's getting a bit convoluted, hopefully we're nearing the end of the edge-case road? All my fingers are crossed...

ALERTua:
In this commit, searching for "Night Moves" brings me the correct album

image

But clicking the performer title I am redirected to an unexpected artist page

image

which does not have the album I was redirected from, as the album is published on the Hip Dozer label.
While the artist is correct in this case, it is just one of their pages, similar only by the performer's name.
I cannot say I am a fan of this behavior, as it can produce false redirections, and I wish the performer title redirected to the label page, not an assumed artist page.
While all this is irrelevant for the users who know what they are looking for, it has impact on the music exploration user journey, which, as I imagine, is not the goal of MA.

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -180,7 +183,18 @@ def is_streaming_provider(self) -> bool:
     async def search(
         self, search_query: str, media_types: list[MediaType], limit: int = 50
     ) -> SearchResults:
-        """Perform search on music provider."""
+        """Perform search on music provider.
+
+        Bandcamp's autocomplete returns three result kinds: bands/labels (b),
```
OzGav: Standardisation point here. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed).

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -196,20 +210,193 @@ async def search(
...
+            expiration=CACHE_METADATA,
+            provider=self.instance_id,
+        )
+        return band_id
+
+    @throttle_with_retries
+    async def _fetch_performer_band_id(self, performer_name: str, target_slug: str) -> int | None:
+        """Autocomplete-search for ``performer_name``, return the first matching band_id.
+
+        ``is_label`` results are skipped so a same-named label doesn't
```
OzGav: This is really implementation as well

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -196,20 +210,193 @@ async def search(
...
+        masquerade as the performer's band page.
+        """
+        try:
+            results = await self._client.search(performer_name)
+        except BandcampRateLimitError as error:
+            raise ResourceTemporarilyUnavailable(
+                "Bandcamp rate limit reached", backoff_time=error.retry_after
+            ) from error
+        except BandcampAPIError:
```
OzGav: Everywhere else in the code you logged or raised on this error?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -338,10 +525,76 @@ async def get_library_tracks(self) -> AsyncGenerator[Track, None]:
...
+        # band when the slug actually matches the band's own name (e.g.
+        # cached IDs constructed before disambiguation was reliable).
+        return await self._get_synthetic_artist(prov_artist_id, band_id, performer_slug)
+
+    async def _get_synthetic_artist(
+        self, prov_artist_id: str, band_id: int, performer_slug: str
+    ) -> Artist:
+        """Resolve a synthetic ID to either the real band or a per-page performer.
+
+        Order matters. Bandcamp's autocomplete sometimes returns ``a``/``t``
```
OzGav: Implementation detail

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -351,14 +604,72 @@ async def get_artist(self, prov_artist_id: str) -> Artist:
...
+            url=api_artist.url,
+            image_url=image_url,
+        )
+
+    @use_cache(CACHE_METADATA)
+    @throttle_with_retries
+    async def _fetch_discography(self, band_id: int) -> list[dict[str, Any]]:
+        """Fetch a band's discography keyed by band_id (cached).
+
+        Real artist (``"{band_id}"``) and synthetic performer
```
OzGav: Implementation detail and parameters not in Sphinx style which is the standard

music_assistant/providers/bandcamp/_ids.py
```diff
@@ -0,0 +1,76 @@
...
+# Slugs are lowercase alphanumerics joined by single hyphens.
+# The colon separator between band_id and slug is unambiguous because
+# band_ids are pure digits and slugs contain no colons.
+_SLUG_NON_ALNUM = re.compile(r"[^a-z0-9]+")
+
+
+def slugify_performer(name: str) -> str:
+    """Reduce a performer name to a stable lowercase slug.
+
+    Two performer names compare equal as artists when their slugs match.
```
OzGav, line 17: Implementation detail

music_assistant/providers/bandcamp/converters.py
```diff
@@ -374,3 +432,74 @@ def album_from_api(self, album: APIAlbum) -> MAAlbum:
...
+
+    def synthetic_artist(
+        self,
+        band_id: int,
+        performer_name: str,
+        *,
+        url: str | None = None,
+        image_url: str | None = None,
+    ) -> MAArtist:
+        """Build an Artist for a per-album performer that has no band page.
```
OzGav:
```suggestion
        """
        Build an Artist for a per-album performer that has no band page.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -374,3 +432,74 @@ def album_from_api(self, album: APIAlbum) -> MAAlbum:
...
+        return output
+
+
+def _resolve_artist_id(
+    *,
+    band_id: int | str,
+    performer: str | None,
+    band_name: str | None,
+) -> str:
+    """Choose between a real or synthetic artist item_id.
```
OzGav:
```suggestion
    """
    Choose between a real or synthetic artist item_id.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -331,9 +380,18 @@ def album_from_discography_item(self, item: DiscographyItem) -> MAAlbum:
             )
         return output
 
-    def album_from_api(self, album: APIAlbum) -> MAAlbum:
-        """Convert an API Album object to MA Album format."""
+    def album_from_api(self, album: APIAlbum, *, artist_item_id: str | None = None) -> MAAlbum:
+        """Convert an API Album object to MA Album format.
```
OzGav:
```suggestion
        """
        Convert an API Album object to MA Album format.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -272,18 +309,30 @@ def artist_from_api(self, artist: APIArtist) -> MAArtist:
         )
         return output
 
-    def album_from_discography_item(self, item: DiscographyItem) -> MAAlbum:
+    def album_from_discography_item(
+        self, item: DiscographyItem, *, artist_item_id: str | None = None
+    ) -> MAAlbum:
         """Convert a raw discography dict to MA Album format.
```
OzGav:
```suggestion
        """
        Convert a raw discography dict to MA Album format.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -182,10 +202,27 @@ def track_from_api(
         album_id: str | int | None = None,
         album_name: str = "",
         album_image_url: str = "",
+        *,
+        tralbum_artist: str | None = None,
+        artist_item_id: str | None = None,
     ) -> MATrack:
-        """Convert a Track object from the API to MA Track format."""
+        """Convert a Track object from the API to MA Track format.
```
OzGav:
```suggestion
        """
        Convert a Track object from the API to MA Track format.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -109,9 +121,17 @@ def track_from_search(self, item: SearchResultTrack) -> MATrack:
             },
         )
 
-    def album_from_search(self, item: SearchResultAlbum) -> MAAlbum:
-        """Create an Album from new API SearchResultAlbum."""
+    def album_from_search(
+        self, item: SearchResultAlbum, *, artist_item_id: str | None = None
+    ) -> MAAlbum:
+        """Create an Album from new API SearchResultAlbum.
```
OzGav:
```suggestion
        """
        Create an Album from new API SearchResultAlbum.
```

music_assistant/providers/bandcamp/converters.py
```diff
@@ -72,9 +74,19 @@ def streaming_url_from_api(
             content_type = ContentType.UNKNOWN
         return streaming_url, bitrate, content_type
 
-    def track_from_search(self, item: SearchResultTrack) -> MATrack:
-        """Create a Track from new API SearchResultTrack."""
+    def track_from_search(
+        self, item: SearchResultTrack, *, artist_item_id: str | None = None
+    ) -> MATrack:
+        """Create a Track from new API SearchResultTrack.
```
OzGav:
```suggestion
        """
        Create a Track from new API SearchResultTrack.
```

OzGav (commented):
There are a number of other multi-line docstrings to have the first line dropped down to match the indentation of the following text that I didn't put a suggestion on but they all need doing just as a point of standardisation.

teancom:
@ALERTua I think I've tightened this up as much as I can, without doing full discography o^n-style searches (which would make us bad API users and have a not-great - meaning slow - user experience). If how this branch works overall is - in your eyes - a regression or even just a 'no-op', then let's drop it, no harm, no foul, no use changing behavior and adding code without making some sort of progress. I will truly not be offended. I think this works better for more use-cases, but it's definitely not perfect.

OzGav:
No at the end of the day I think it should work how the users expect, as best as you can. 

On the false-redirect behaviour you identified, if it can’t be fixed then I think a note in the docs would be required.

Will see what Marvin thinks when he reveiws.

MarvinSchenkel (commented):
This looks good to me code wise, but I cannot judge this functionally as I am not a Bandcamp user. I have one question before I merge it. Is it possible to (partly) reuse the helper logic we already have in `helpers/compare.py` for comparing band and performers?

ALERTua:
Okay, trying to look at the label-published artists' behavior from the user perspective.
I will omit the "All" search mode, as I don't like it and want it gone.

1. I'm looking for all Apollo Brown albums (they have their own page and label-published albums)
I'm entering "Apollo Brown" in the "Album" mode search.
I'm seeing all Apollo Brown albums from their pages and the label pages.
Good.

2. I'm looking for the "The Green Arrows" artist.
I'm entering "The Green Arrows" in the "Artists" mode search.
Got their virtual artist, got their one album, got the link to their label.
Good.

3. I'm looking for the "Apollo Brown" artist in the "Artists" search mode.
Got all their pages and collaboration artists that mention them. All of the Apollo Brown pages. This way, I cannot find their album that was published on a label. That's expected. Can we do better without overloading the API? No.
OK.

4. I'm looking for a song from an album that is published on a label, from this song I want its album, from its album I want all albums.
I'm entering "[Moon at Midnight](https://hipdozer.bandcamp.com/track/moon-at-midnight)" song into the "Tracks" search mode.
I got "Moon at Midnight" from "Night Moves" by "Apollo Brown". Good.
Navigating to the song, I can navigate to the album. Night Moves by Apollo Brown. Got the URL for the album on the label's page. Clicking on the author's name I was navigated to

image

which is a wild guess that is a coincidence.
We are navigating the user from an album to the artist who does not have this album. Not OK.
Another try: with a song from "The Green Arrows".
Searching for [Madzangara Dzimu](https://analogafrica.bandcamp.com/track/madzangara-dzimu) in the "Tracks" search mode. Got the track. Navigating to the album. All good. Navigating to the artist. Got the virtual artist's page.
Good.

Let's say, we are navigating the user to the wild-guessed artist page, and it is effective in 75% cases. Good enough. We will do better if the users provide us the edge-cases we can fix.

What user scenarios do we have more? "Browse".

I'm following Analog Africa and I want to brows my "Following" to search for "The Green Arrows".
Navigating to Browse→Bandcamp→Following
The page is stuck. Exception:
```
2026-05-04 09:39:50.428 ERROR (MainThread) [music_assistant.webserver] Error handling message: music/browse: 'int' object has no attribute 'partition'
```
`_ids.py:72`
https://github.com/music-assistant/server/pull/3824/changes#diff-748edd462bf55e5abb80bfa3e6733b576285d42fb762b4f521afdb617f5f853dR72

Let's fix this one, make sure the Browse works correctly with both artist pages and virtual artists and all this **looks good to me**.

teancom:
I, just last night, left on a 3 week vacation with no laptop (I'm very excited). I can address those when I get back or you can push into the branch, I'm happy either way. ❤️

MarvinSchenkel:
Gonna set this to 'draft' just to be able to keep track of things that need my attention. Please mark it as ready for review again when you want me to have a look

ALERTua:
Sorry, got no energy to spend even an hour on this now. Things got quite emotionally destabilizing for me here in Kyiv, Ukraine.
I will try to find strength before Teancom arrives. Otherwise, no rush, this waits.

MarvinSchenkel:
> Sorry, got no energy to spend even an hour on this now. Things got quite emotionally destabilizing for me here in Kyiv, Ukraine. I will try to find strength before Teancom arrives. Otherwise, no rush, this waits.

Sorry to hear that. All the best and of course, no rush!

tests/providers/bandcamp/test_init.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pytest.fixture
-async def bandcamp_provider(mass: MusicAssistant) -> AsyncGenerator[ProviderConfig]:
+async def bandcamp_provider(mass: MusicAssistant) -> AsyncGenerator[ProviderConfig]:   # noqa: PLR0915
```
OzGav:
```suggestion
async def bandcamp_provider(mass: MusicAssistant) -> AsyncGenerator[ProviderConfig]:  # noqa: PLR0915
```

OzGav:
@teancom are you able to progress this one?

teancom:
I just rebased off of origin/dev, cleaned up the merge conflicts, and am taking another look, to see if I can fixup ALERTua's concerns.

teancom:
> 2026-05-04 09:39:50.428 ERROR (MainThread) [music_assistant.webserver] Error handling message: music/browse: 'int' object has no attribute 'partition'

The [newest commit](https://github.com/music-assistant/server/pull/3824/commits/286732e8d11f328df45706d785d0a056000a56e4) fixes this, good catch.

As for the other concerns, I've gone through the API again and I just can't see how to get both "always resolve to the canonical artist if there is one" and "all albums of an artists's discography are there" at the same time, without hammering Bandcamp's API and (in effect) replicating a good portion of their database locally in cache. I think this is as good as it gets*, until/unless Bandcamp adds globally unique artist IDs. Until then, using heuristics with the possibility of same-name-different-artist results is what we have.

Which isn't to say that means we *should* merge this! We just need to decide if the drawbacks are worth the upsides. I'm going to move this out of draft and then we can merge or close.

*As good as *I* can get it, to be sure. It's always possible I'm missing something still.

teancom:
@MarvinSchenkel and @OzGav I'm really looking for your feedback on my previous message. I don't think we should bother alertUA. ❤️

ALERTua:
Two independent reviews from Claude Fable and Opus stated:
## Review findings

Reviewed the full diff plus the post-PR files, the relevant Music Assistant core code (`UniqueList`, `sort_search_result`, `use_cache`, `throttle_with_retries`), and the `bandcamp-async-api` 0.2.2 sources. The composite-ID design itself is coherent: the three resolution paths (`_resolve_search_artist_ids`, `_resolve_artist_item_id`, `_discography_artist_item_id`) agree on precedence, drift collapse in `get_artist` is correct, and `_ids.py` is clean. That said, there is one blocker and a number of smaller issues.

### 🔴 Blocker

**Synthetic artists share a `uri` and get deduplicated away** — `converters.py`, `synthetic_artist()`

`synthetic_artist` sets `uri=url`, where `url` is the *hosting band page* URL. `_MediaItemBase.__hash__`/`__eq__` compare by `uri`, and the search controller runs results through `sort_search_result()` → `UniqueList`. So every synthetic performer on one label page — and the label itself — compare equal, and all but one are silently dropped. Reproduced: `UniqueList([audiophob, Mortaja, Spherical Disrupted])` yields just `['audiophob']`. This defeats the exact scenario the PR exists to fix, whenever a label page contributes more than one performer.

The same collision exists on the `_get_synthetic_artist` path (`url=api_artist.url`) and the search-emission path (`item.artist_url`).

**Fix:** don't set `uri` on synthetic artists (let MA auto-generate `bandcamp://artist/{band_id}:{slug}`); keep the band-page URL only in `provider_mappings.url` / `metadata.description`. The current tests miss this because the search mocks set `artist_url=""`, which happens to fall back to the auto-generated uri — please make the mocks carry a realistic `artist_url` too.

### 🟠 Medium

1. **Cold-cache search fan-out is genuinely parallel** — `__init__.py`, `_lookup_performer_band_ids_parallel`. `throttle_with_retries` uses a sliding-window counter, not a lock, so the `asyncio.gather` fires every unique-slug autocomplete concurrently. Since the upstream parser populates `artist_name` for every `a`/`t` row, any row whose `b` row didn't make the same response triggers a lookup — a cold search can consume most of the 50/10s rate-limit window in one shot (N+1 amplification).
2. **Sequential `get_artist` calls inside the search loop** — `__init__.py`, search artist-emission. One extra round trip per real-page performer, awaited sequentially on the search path; unbounded latency, and the `suppress(...)` swallows rate-limit signals that would otherwise reach core.
3. **Doubled `band_details` request on synthetic resolve** — `_get_synthetic_artist` calls `_client.get_artist(band_id)` and then `_fetch_discography(band_id)`; both POST the same `mobile/24/band_details` endpoint with the same payload. `@use_cache` only dedups across calls, not this pair, so every cold synthetic resolve doubles the traffic.
4. **Rate-limit error type changed and is now inconsistent** — `get_artist` used to raise `RateLimited` on `BandcampRateLimitError`; it now raises the parent `ResourceTemporarilyUnavailable` (which `throttle_with_retries` treats differently — no escalating backoff floor), while the discography path still raises `RateLimited`. Should be one type, and the behavior change deserves a mention in the PR description.

### 🟡 Low

1. **List/detail asymmetry after drift collapse** — `get_artist("{id}:{band-own-slug}")` collapses to the real band, but `get_artist_albums` with the same ID filters the discography by that slug instead, so the collapsed artist's album list drops items credited to collaboration names.
2. **Synthetic artist can display the band's own name** — in `_get_synthetic_artist`, `first.get("artist_name") or first.get("band_name")` can yield the band's name when a discography row has `artist_name` null and its `band_name` slugs to `performer_slug` but misses the earlier short-circuit.
3. **Incorrect docstring claim in `_ids.py`** — it says `"&"` vs `"and"` "collapse together"; they don't: `slugify_performer("A & B") == "a-b"` but `"A and B" == "a-and-b"`. Worth fixing so future maintainers don't rely on it.
4. **`track_from_feed` not updated** — feed tracks still credit `str(track.band_id)` / `track.band_name`, so the same track reached via the feed and via `get_track` gets two different artist item_ids.
5. **`CancelledError` swallowed** — `isinstance(result, BaseException)` after `gather(..., return_exceptions=True)` logs a child cancellation as a warning instead of propagating it.
6. **`results.artists` can exceed the caller's `limit`** — synthetics are appended after the capped loop. Core re-caps, so cosmetic.
7. **Test mocks hide the blocker** — `_search_artist_mock` / `_search_album_mock` set empty URLs, which is exactly the case where the uri collision doesn't manifest; otherwise the mocks are faithful to the 0.2.2 dataclasses.
8. **`test_get_artist_albums_synthetic_id_filters_discography` passes accidentally** — it never stubs `_client.search`; the `AsyncMock` result iterates empty, so "synthetic IDs are kept" is not actually asserted and the unintended lookup call goes unchecked.
9. **`str(item.band_id)` in the following-artists path** — confirmed this is *required by this PR* (not a pre-existing bug): `parse_artist_id` calls `.partition(":")`, which would raise `AttributeError` on an int. Fine as is.

### Verified non-issues

- `@use_cache` key derivation works for `_fetch_discography(band_id: int)`, and `parse_value(..., list[dict[str, Any]])` round-trips the discography payload unchanged.
- `parse_artist_id("123:")` returns `(123, None)` and both callers treat it identically.
- Slug collisions between different performers on the same page merge silently (documented behavior) but never crash.
- `get_artist_toptracks` handles synthetic IDs transitively via `get_artist_albums`.

### Verdict

The data model achieves the stated goal and the drift-collapse logic is sound, but the PR is **not mergeable as-is**: the shared-`uri` collision makes MA's own search dedup discard every synthetic performer beyond the first on a given label page. The fix is small (drop `uri=url` from `synthetic_artist`, realistic `artist_url` in test mocks). After that, the cold-cache search cost (parallel autocomplete fan-out + sequential `get_artist` calls) is the main thing worth revisiting.

teancom:
Alright, I think I've addressed the feedback, @ALERTua. The exception is `Doubled band_details request on synthetic resolve` which to do cleanly will require a small update in bandcamp-async-api. I'll open that PR and a follow-up PR here once it's merged / released.

Unrelated to all of that - I'm glad that you're okay. ❤️

OzGav:
Please restore the PR template so our bot can do its work

teancom:
@OzGav This PR actually predates the template! I only say that so you know I would never be so uncouth as to _remove_ it. I grabbed a copy from the .github directory and filled it out. 🙇

OzGav:
Of course. I should have written my note to say I understood that. Thanks!

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -170,7 +174,19 @@ def is_streaming_provider(self) -> bool:
...
+
+        Bandcamp's autocomplete returns three result kinds: bands/labels (b),
+        albums (a), and tracks (t). For album/track results, ``band_id`` is
+        the page owner (could be a label) and ``band_name`` is the *performer*
+        credit — which for label-released albums differs from the page
+        owner's own name. We map performer-without-a-band-page entries onto
+        synthetic artists keyed by ``{band_id}:{slug}`` and emit them in
+        artist results too, so searching for e.g. "Mortaja" surfaces
+        Mortaja-on-audiophob even though the performer has no band page of
+        their own. See :mod:`._ids` for ID semantics.
```
OzGav: I'm still not feeling that this is caller facing?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -239,6 +311,122 @@ async def get_recommendation_items(
...
+        )
+        return band_id
+
+    @throttle_with_retries
+    async def _fetch_performer_band_id(self, performer_name: str, target_slug: str) -> int | None:
+        """Autocomplete-search for ``performer_name``; return the first non-label band match."""
+        try:
+            results = await self._client.search(performer_name)
+        except BandcampRateLimitError as error:
+            raise ResourceTemporarilyUnavailable(
```
OzGav: RateLimited?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -369,10 +557,82 @@ async def get_library_tracks(self) -> AsyncGenerator[Track]:
...
+        band's name). The search-time path mints a synthetic
+        ``{band_id}:slug-of-band-own-name`` in that case because it can't
+        tell band-by-itself from label-release without the ``b`` row. We
+        collapse that drift here, on the navigation/persistence path, by
+        resolving the synthetic to the real band whenever the slug equals
+        the band's own slug — BEFORE consulting the discography. Otherwise
+        an album whose performer-name happens to match the band's own name
+        (band-by-itself items with ``artist_name=null``) would build a
+        shadow synthetic that lives parallel to the real band in MA's
+        library and produces duplicate artist entries.
```
OzGav: Again this is not caller facing?

music_assistant/providers/bandcamp/__init__.py
```diff
@@ -382,14 +642,76 @@ async def get_artist(self, prov_artist_id: str) -> Artist:
...
+    @use_cache(CACHE_METADATA)
+    @throttle_with_retries
+    async def _fetch_discography(self, band_id: int) -> list[dict[str, Any]]:
+        """
+        Fetch a band's discography keyed by band_id (cached).
+
+        Real artist (``"{band_id}"``) and synthetic performer
+        (``"{band_id}:{slug}"``) lookups both go through this so the
+        underlying ``mobile/24/band_details`` call hits once per band per
+        cache window, not once per ``prov_artist_id``.
```
OzGav: Should be moved to an inline comment

music_assistant/providers/bandcamp/__init__.py
OzGav: I know this is existing but it would be good to tidy this up by narrowing. `_browse_person_people` can raise `(LoginFailed, RateLimited, MediaNotFoundError, RetriesExhausted)`

music_assistant/providers/bandcamp/_ids.py
```diff
@@ -0,0 +1,80 @@
...
+def slugify_performer(name: str) -> str:
+    """
+    Reduce a performer name to a stable lowercase slug.
+
+    Two performer names compare equal as artists when their slugs match.
+    The transform is intentionally lossy — punctuation and case are
+    discarded — and ``"&"`` is normalized to ``"and"``. Bandcamp's
+    per-album performer field is consistent within a band page in practice,
+    so within-band collisions are rare; cross-band collisions are not a
+    concern because ``band_id`` is part of the key.
```
OzGav, line 22: There is some text here that is useful as an inline comment but shouldn't be here in the docstring

OzGav (commented):
This looks good but some comments added to tidy up the docstrings plus two code points.

music_assistant/providers/bandcamp/__init__.py
teancom: Oh man, I try to be good about not letting the LLM slop pollute comments this much. I'm frankly embarrassed it got past me. Cleaned up this and the rest of the things you found. 🙇

music_assistant/providers/bandcamp/__init__.py
teancom: Done!

music_assistant/providers/bandcamp/__init__.py
teancom: Good catch, fixed.

OzGav (approved):
Ok let’s get this in for testing! Thanks @teancom and @ALERTua

music_assistant/providers/bandcamp/__init__.py
ALERTua: I found a good way to get rid of long comments by:
1. Forcing the agent to run a subagent with the instructions to conduct a code review after any *meaningful* change
2. One of the said subagent instructions is specifically to shorten the comments :)

```
## Self-review after implementing (mandatory, every project)

After finishing a meaningful chunk of code — a plan step, a feature slice, anything
substantial enough to report as "done" — and once its tests pass, but BEFORE telling
the user it is done: order an independent review of that diff from a fresh sub-agent.

- **Trigger:** a meaningful unit of work, not every edit. A one-line fix or a typo
  does not need it; a new module, a reworked flow, or a plan step does.
- **Brief the reviewer to reach its own conclusions.** Give it the relevant plan/spec
  file, the changed paths (`git diff --stat`), and the legacy/reference code being
  reimplemented if there is any. Ask it to find what a careful reviewer would object
  to — correctness, misleading output, naming, extensibility, missing test coverage.
  Never tell it what you expect it to find or that the code is fine.
- **The reviewer never re-runs checks.** Tests, linters and type checkers are run
  (and green) BEFORE the review is ordered; state their results in the brief as
  facts and forbid re-running them — the reviewer's budget goes into reading code.
- **Comment-length audit.** The reviewer must also assess every comment/docstring
  ADDED by the diff: can it be shortened without losing the constraint it states?
  If yes, it proposes the shorter wording. Norm: comment ≈ one line, docstring 1–3
  (human MR reviewers repeatedly object to long comments).
- **Judge the findings, do not execute them.** Some will be wrong or premature. Apply
  the ones that hold, and tell the user plainly which ones you rejected and why.
- **Report the review happened** as part of reporting the step: what was found, what
  was fixed, what was declined.
- Choose the subagent model according to the complexity of the reviewed code change
```

PR #3826, opened 2026-05-01
Radio Paradise small cleanup

teancom:
I've been really enjoying Radio Paradise lately, and decided to do a bit of a code-sweep just because. Absolutely no behavioral change, just some dedupe, taking out some dead code, etc.

Robot sidekick's detailed explanation:

- Add RadioParadiseChannel TypedDict (replacing dict[str, dict[str, Any]])
- Extract _fetch_json helper so play/now_playing methods stop duplicating the same try/except + ClientTimeout + http_session.get scaffolding
- Remove unreachable try/except in _update_stream_metadata (inner _get_channel_metadata already swallows aiohttp.ClientError)
- Pull magic values (API timeout, update interval, cover URL) into constants
- Drop unreachable defensive fallbacks in parse_radio; callers validate ids
- Switch logger calls to %-style to match codebase convention

PR #3828, opened 2026-05-01
Nicovideo: fix get_user_history kwargs to match niconico API

teancom:
I noticed when opening https://github.com/music-assistant/server/pull/3826 that mypy wasn't passing. Tracked it down to this.

The niconico library's video.get_history takes (*, page_size, page) — not `limit`. The previous call passed `limit=...`, which would TypeError at runtime and was failing repo-wide mypy.

Mirror the page_size/page conversion that get_like_history above already uses (limit -> page_size capped at 100, page=1).


teancom:
False alarm, I just hadn't run the ./scripts/setup.sh recently enough.

PR #3836, opened 2026-05-04
Add Mamma Mi Radio music provider

florianhorner:

Adds a **Mamma Mi Radio** music provider: a single Radio entry for the self-hosted radio station that runs as a Home Assistant addon (same author). The Italian hosts notice the home and work it into the show; live typed now-playing metadata (track/segment, hosts, artwork, up-next) is read from the addon's versioned consumer contract `GET /api/integrations/v1/now-playing`.

The provider is v1-only and requires **addon 2.13+**. It exposes browse and search, adds its single station to the library on load, and polls the now-playing contract conditionally (`If-None-Match`, 304 reuses the cached payload) while a stream is active.

- n/a — no tracking issue; discussed inline on this PR

## Types of changes

## Checklist

The two unticked boxes do not apply: this PR touches no shared models and no UI. The diff is confined to `music_assistant/providers/mammamiradio/**`, `tests/providers/mammamiradio/**` and the generated `translations/en.json` entries.

## Review rounds

Per review, the legacy `/public-status` fallback was dropped entirely: the provider is v1-only and requires **addon 2.13+**.

- `34eef4471`: drop legacy fallback; v1-only init; all style items (docstrings, plain-English comments, log prefixes, TypeError removal, class defaults, URL-at-init); v1 hardening (album only from payload, http(s)-only artwork, poisoned-ETag recovery); test suite rewritten for v1-only behavior
- `0243c7c0e`: manifest documentation URL; add `icon_monochrome.svg` (single-tone currentColor, 522 bytes)
- `67483efa2`: shared cross-repo contract fixture (see governance below)
- `12a8e7143`: junk `schema_version` values get the 2.13 hint (never echoed), tests to 100% provider line coverage
- `5a749269c`: malformed `relative_url` can no longer leak ValueError past init, unhashable `segment_class` renders the station frame, invalid host characters fail setup with the localized error
- `8310f8382`: sharpened description copy (manifest, strings, radio metadata)
- `c90d178d5`: `setup_flow.py` per @MarvinSchenkel (the provider cannot be configured on current `dev` without it, and the URL is now read via `get_setup_value`); the single station is added to the library on load per @OzGav; and the `schema_version` probe is reduced to the two checks @OzGav described, with the bool/int coercion removed from `_supports_v1_schema`

Docs PR against `beta`: music-assistant/music-assistant.io#799.

## Behavior

- Init validates the configured base URL first (localized setup error, no HTTP), then probes the v1 endpoint once. A missing endpoint, non-JSON body, or missing `schema_version` fails with "requires addon 2.13 or newer"; an unsupported version is named in the error; a 5xx reports the HTTP status; a connection failure reports unreachable. The probe doubles as format discovery (the contract publishes the stream's audio format) and as the liveness check, since the addon is a single process serving both the stream and the metadata endpoint.
- Failed loads self-heal: these errors are `MusicAssistantError` subclasses, which MA auto-retries every 120s.
- Metadata polling is one conditional request per active stream every 12s. Mid-stream failures keep the prior frame and never raise into MA.

## Contract stability

The v1 contract is governed on the addon side, merged and running: florianhorner/mammamiradio#889. `CONTRACT.md` freezes the wire surface as additive-only, and a drift CI check byte-compares the addon serializer's output against a golden fixture on every addon PR. The release-ordering rule (the provider side lands here before the addon ships any contract change) is enforced by the same CI: it checksum-compares the addon's fixture against the copy on this repo's `dev`. That fixture lives in this PR at `tests/providers/mammamiradio/fixtures/v1_now_playing.json` (byte-identical) and is exercised by `test_golden_fixture_maps_cleanly`. [proof: fixture]

## Proof

- [x] build: https://github.com/music-assistant/server/actions/runs/31439247832 [proof: build]
- [x] tests: https://github.com/music-assistant/server/actions/runs/31439247832/job/93620279722 (green on `b0463093c`; 142 provider tests passed, 1 opt-in live smoke skipped, 100% provider line coverage) [proof: tests]
- [x] lint: https://github.com/music-assistant/server/actions/runs/31439247832/job/93620234104 (pre-commit run --all-files clean: ruff, mypy, manifests, icons, method order, translations, config entries) [proof: lint]
- [x] runtime: https://gist.github.com/florianhorner/92276d370f8610484e025f191af1dbfd (provider run against the production addon: v1 payload capture with headers, live ETag/304 conditional check, and the provider's opt-in live smoke test passing) [proof: runtime]
- [x] schema: no MA core/model changes; the diff touches only `music_assistant/providers/mammamiradio/**`, `tests/providers/mammamiradio/**` and generated `translations/en.json` entries — the `provider-scope` check confirms this independently: https://github.com/music-assistant/server/actions/runs/31439247832/job/93620234157 [proof: schema]
- [x] fixture: sha256 `5ff694ae8bb32ad00c2924b1c7a1f3f272f98f7cdce5b93fce17963bc5400067` identical in this PR and the addon repo's `tests/integrations/golden/` [proof: fixture]


florianhorner:
Marking this as draft while I finalize a few things internally before review:

- swap placeholder icon for the real mammamiradio logo
- add a live-integration smoke test against a running mammamiradio addon (current tests mock the HTTP layer)
- run an internal code-review pass on the diff
- add a sample-audio link in the description so reviewers can hear what mammamiradio actually sounds like before reviewing

Will mark ready-for-review once those land. Sorry for the noise — should not have hit ready on the first push. Appreciate your patience.

OzGav:
Instead of a new provider why not just add this one station manually?

OzGav:
Again why this provider instead of adding the station manually?

florianhorner:
Thank you for the check-in, sorry for he lag - was finalizing the cleanup before reply.

Manual URL works for basic playback, fair point.

The reason for a provider is segment-aware metadata:
mammamiradio exposes typed segments (banter / song / ad as distinct typed segments with timing, host attribution, upcoming queue, and HA context) on /public-status that media_player.play_media(url) can't surface. v1 is intentionally minimal; the metadata hook is what I'd see as a separate RFC. Happy to close in favor of a Discussion if that's the cleaner path.

OzGav:
OK sounds fair. Currently though this isn't doing much so I would keep going and propose your final  version with all the metadata included.

florianhorner:
Helpful. Two ways I could go:

(a) Close this and refile as one bundled PR with the metadata extension RFC + reference impl + provider together.
(b) Keep this in draft and add the metadata extension as additional commits here before flipping ready.

Which feels cleaner from your side?

OzGav:
Personally I think B is fine.

MarvinSchenkel:
Hey @florianhorner , what's the status for this? If you want us to have a look at the code, please mark it as 'ready for review'.

florianhorner:
PR is now ready for review as invited. The Copilot-flagged fallback issue in `_probe_now_playing` is fixed in 164d0dd6d: `aiohttp.ContentTypeError` (non-JSON content-type) and non-dict JSON payloads now demote to the legacy `/healthz` probe instead of raising `ProviderUnavailableError`, and unsupported `schema_version` values fall back the same way. Each path has a regression test.

Validated locally:
- `pytest tests/providers/mammamiradio` — 100 passed
- `ruff check` + `ruff format --check` — clean
- `mypy music_assistant/providers/mammamiradio tests/providers/mammamiradio` — clean

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+        # Verify the stream endpoint is reachable; raise a clean
+        # ProviderUnavailableError so MA reports "source unavailable"
+        # rather than crashing inside ffmpeg.
+        try:
+            timeout = aiohttp.ClientTimeout(total=REACHABILITY_TIMEOUT)
+            async with self.mass.http_session.head(
+                stream_path, timeout=timeout, allow_redirects=True
+            ) as response:
+                if response.status >= 400:
+                    msg = (
```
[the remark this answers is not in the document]
florianhorner: Fixed in e84cc0a46: the stream-time HEAD probe was removed entirely — liveness is checked once at init with a GET, and get_stream_details passes through without any per-play probe.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,726 @@
...
+                    return None
+                payload = await response.json()
+                return payload if isinstance(payload, dict) else {}
+        except (aiohttp.ClientError, TimeoutError) as err:
+            msg = f"mammamiradio addon unreachable at {url}: {err}"
+            raise ProviderUnavailableError(msg) from err
+        except (ValueError, TypeError) as err:
+            # Endpoint answered but the body is not JSON — treat as no-v1, fall back.
+            self.logger.debug("mammamiradio v1 now-playing probe returned non-JSON: %s", err)
+            return None
```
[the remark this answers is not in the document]
florianhorner: Fixed in 164d0dd6d: a non-dict JSON payload returns None (no v1 selection) and aiohttp.ContentTypeError is caught before ClientError, so both fall back to /healthz instead of raising ProviderUnavailableError. Each path has a regression test.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,775 @@
...
+                parts.append(f"Up next: {up_label}")
+        casa = _clean_str(ha.get("mood")) or _clean_str(ha.get("weather"))
+        if casa:
+            parts.append(f"A casa: {casa}")
+        description = " · ".join(parts) or None
+
+    return StreamMetadata(
+        title=title,
+        artist=_clean_str(artist),
+        album=station_name,
```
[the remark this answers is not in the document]
florianhorner: Fixed in c9126315b: the legacy mapper now sets album only for music segments, mirroring the v1 mapper (covered by test_legacy_album_only_on_music_segments).

tests/providers/mammamiradio/test_mammamiradio.py
```diff
@@ -0,0 +1,1612 @@
...
+        # Browse returns exactly one Radio entry.
+        items = await prov.browse("mammamiradio://")
+        assert len(items) == 1
+        assert isinstance(items[0], Radio)
+        # Stream details succeed against the live addon.
+        details = await prov.get_stream_details(RADIO_ITEM_ID, MediaType.RADIO)
+        assert isinstance(details.path, str)
+        assert details.path.endswith("/stream")
+        assert details.audio_format.content_type == ContentType.MP3
+        # The live-metadata callback fetches a real /public-status payload.
```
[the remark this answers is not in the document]
florianhorner: Fixed in c9126315b: the comments now describe the v1-first init probe and the endpoint selected at init.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,775 @@
...
+        Return the configured mammamiradio URL stripped to a public base URL.
+
+        Query strings, fragments, and userinfo are intentionally discarded so an
+        admin token pasted into the setup field never reaches MA stream URLs,
+        probe URLs, logs, or provider-unavailable errors.
+        """
+        raw = str(self.config.get_value(CONF_MAMMAMIRADIO_URL) or DEFAULT_URL).strip()
+        parts = urlsplit(raw)
+        netloc = parts.netloc.rsplit("@", 1)[-1]
+        return urlunsplit((parts.scheme, netloc, parts.path.rstrip("/"), "", ""))
```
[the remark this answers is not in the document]
florianhorner: Fixed in c9126315b: the configured base URL is validated before any request — a full http(s) URL with a hostname is required, so schemeless values like localhost:8000 now fail setup immediately with a provider-localized error instead of a late aiohttp failure. Covered by the URL accept/reject matrix and test_invalid_base_url_fails_setup_before_http.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,828 @@
...
+                    self.logger.debug(
+                        "mammamiradio v1 now-playing returned unsupported schema_version %r",
+                        payload.get("schema_version"),
+                    )
+                    return None
+                new_etag = response.headers.get("ETag")
+                if isinstance(new_etag, str):
+                    data["v1_etag"] = new_etag
+                data["v1_last"] = payload
+                return payload
```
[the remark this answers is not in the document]
florianhorner: Fixed in aad859799: a 200 response without an ETag header now clears the stored validator, so polling actually becomes unconditional as documented (covered by test_v1_callback_drops_stale_etag_when_header_disappears).

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,832 @@
...
+            return
+
+        # Detect segment changes via a stable per-segment identity, NOT the
+        # contract's ``changed_at`` clock: the addon advances ``changed_at`` on any
+        # state change (e.g. a queue-append that fills the lookahead buffer
+        # mid-segment), which would otherwise snap the alternation back to "Now"
+        # mid-segment. ``changed_at`` is also structurally optional in the contract.
+        now = payload.get("now_playing")
+        now = now if isinstance(now, dict) else {}
+        seg_key = (now.get("segment_type"), now.get("title"), now.get("started_at"))
```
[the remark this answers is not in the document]
florianhorner: Fixed in 5494ba8ba: the segment key now uses the fields the v1 contract actually emits (segment_type, title, artist, host); started_at stays in the tuple so an addon that later adds it additively gains true per-segment identity. Covered by test_v1_callback_resets_on_artist_change_same_title.

florianhorner:
@MarvinSchenkel ready for review (and this time Copilot agrees)

All review-bot threads are resolved with the fixing commits referenced; Copilot’s pass on the current head generated no new comments. The fork workflows on this head are pending approval. Once the Test workflow runs, the required lint/test contexts will report.

What a reviewer can re-run:

- pytest tests/providers/mammamiradio — 131 passed, 1 opt-in live smoke skipped
- pre-commit run --all-files — clean (ruff, ruff-format, mypy, codespell, provider hooks)
- full suite with coverage — 5,824 passed

Validated against a live addon (2.14.x): init selects the v1 contract, the stream serves MP3 192 kbps / 48 kHz matching the contract-reported format, and each segment class renders correctly in live metadata. I maintain the HA app (add-on) side as well and will keep this provider maintained here.

OzGav:
Do we want the legacy /public-status fallback at all? It looks like requiring addon ≥ 2.13 would delete roughly a quarter of the file?

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+        raise TypeError("base URL must be a string")
+    raw = value.strip()
+    if not raw:
+        raise ValueError("base URL is empty")
+    if any(ch.isspace() or not ch.isprintable() for ch in raw):
+        raise ValueError("base URL contains whitespace or control characters")
+    try:
+        parts = urlsplit(raw)
+        hostname = parts.hostname
+        _ = parts.port  # a nonnumeric or out-of-range port raises ValueError
```
OzGav, line 135: Won't an out of range port already have failed at connect time?

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+
+def _host_display_names(hosts: Any) -> str | None:
+    """
+    Join host display names from a list of host objects.
+
+    The addon serialises ``brand.hosts`` / ``station.hosts`` as objects
+    (``{engine_host, display_name, description}``); earlier code assumed a list of
+    plain strings and silently dropped every host (so banter showed the station
+    name instead of the hosts). This reads ``display_name`` (falling back to
+    ``engine_host``) and tolerates the legacy bare-string shape.
```
OzGav: Use Sphinx-style docstrings with :param: syntax. For simple functions, a single-line docstring is fine. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

This is a general comment for throughout this file.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+    seg_class: Any = None
+
+    if now is not None:
+        seg_class = now.get("segment_class")
+        np_title = _clean_str(now.get("title"))
+        if seg_class == "music":
+            title = np_title
+            artist = _clean_str(now.get("artist"))
+            image_url = _clean_str(now.get("artwork"))
+            album = _clean_str(now.get("album")) or station_name
```
OzGav: The comment above says album should be None as the fallback?

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+            title = np_title or station_name
+            artist = station_name
+        else:
+            # "unavailable" or any future/unknown class -> idle station frame.
+            title = station_name
+    else:
+        # session_state stopped / empty_queue -> nothing playing.
+        title = station_name
+
+    # Terminal title clamp — the invariant: title is always a non-empty str.
```
OzGav: Did AI write this? Either way please change to plain english or delete. Same coment for line 364

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+            default_value=DEFAULT_URL,
+        ),
+    )
+
+
+class MammamiradioProvider(MusicProvider):
+    """Provider implementation for mammamiradio."""
+
+    # Selected in handle_async_init. Class-level defaults keep the accessors safe
+    # if a method is somehow reached before init runs.
```
OzGav: This is impossible as MA only exposes the provider after `handle_async_init` succeeds. Initialise these in `handle_async_init` and drop the comment.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+    _base_url: str | None = None
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        # mammamiradio is an external (self-hosted) audio source whose catalog
+        # is fixed (one Radio entry); treat it like other internet-radio
+        # providers (SomaFM, RadioBrowser).
+        return True
+
```
OzGav:
```suggestion
```

This is just returning what the base class returns.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+        url = self._stream_url_root()
+        payload = await self._probe_now_playing(url)
+        if payload is not None:
+            self._use_v1 = True
+            stream = payload.get("stream")
+            stream = stream if isinstance(stream, dict) else {}
+            audio_format = stream.get("audio_format")
+            self._audio_format_dict = audio_format if isinstance(audio_format, dict) else None
+            self._stream_path = _stream_path_from_contract(stream.get("relative_url"))
+            self.logger.info("mammamiradio v1 now-playing contract reachable at %s", url)
```
OzGav:
```suggestion
            self.logger.info("now-playing contract reachable at %s", url)
```
Log messages already show the provider. Remove this from all log messages.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+        if seg_key != data.get("v1_segment"):
+            data["v1_segment"] = seg_key
+            data["show_upcoming"] = False
+
+        show_upcoming = data.get("show_upcoming", False)
+        try:
+            stream_details.stream_metadata = _v1_to_stream_metadata(
+                payload, show_upcoming=show_upcoming
+            )
+        except Exception as err:
```
OzGav: We can't have broad catchalls. Please narrow this to the expected errors. This is a general comment for throughout the code.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+                        "mammamiradio v1 now-playing probe returned unsupported "
+                        "schema_version %r; falling back to /healthz",
+                        payload.get("schema_version"),
+                    )
+                    return None
+                return payload
+        # ContentTypeError subclasses ClientError but means the endpoint answered
+        # with a non-JSON content-type, so it must be caught first: it demotes to
+        # the /healthz fallback rather than raising ProviderUnavailableError.
+        except (aiohttp.ContentTypeError, ValueError, TypeError) as err:
```
OzGav: I don't think response.json() can raise TypeError. Thus this should be removed in the three places in the code where this appears.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,842 @@
...
+            try:
+                self._base_url = _normalize_base_url(DEFAULT_URL if raw is None else raw)
+            except (TypeError, ValueError) as err:
+                msg = "mammamiradio: invalid base URL configured; enter a full http(s):// URL"
+                raise SetupFailedError(
+                    msg,
+                    translation_key="invalid_base_url",
+                    translation_owner=self.translation_owner,
+                ) from err
+        return self._base_url
```
OzGav: MA reloads the provider on  a config change so you should just do the url building in handle_async_init and remove this

music_assistant/providers/mammamiradio/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "music",
+  "domain": "mammamiradio",
+  "name": "Mamma Mi Radio",
+  "description": "AI-generated Italian radio station — continuous music, banter, and ads. Self-hosted via the Mamma Mi Radio HA addon.",
+  "codeowners": ["@florianhorner"],
+  "requirements": [],
+  "documentation": "https://github.com/florianhorner/mammamiradio",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/music-providers/mamma-mi-radio/",
```

And you need to raise a PR for our docs against the BETA branch

OzGav (dismissed):
Overall looks pretty good. A few comments above and please also supply an icon_monochrome.svg

MarvinSchenkel:

florianhorner:
Thanks for the detailed review @OzGav, all points taken and I’m working on them now. I’ll push everything together as one and re-request review, aiming for this week.

The main change coming out of it:
1/ I’m dropping the legacy /public-status fallback entirely and making the provider v1-only, by that a good chunk of code goes away for the deliberate trade-off of making addon/app v2.13+ the minimum.
2/ Monochrome icon and the docs PR against beta are in the same batch, using the URL and structure you pointed to.

Meanwhile, on the addon/app side, I’m hardening the contract itself with mechanical, verifiable guardrails: freezing the v1 payload behind a byte-authoritative golden fixture and a dedicated contract-drift CI workflow with documented ceremony for proposing and landing future additive changes.

florianhorner:
> Do we want the legacy /public-status fallback at all? It looks like requiring addon ≥ 2.13 would delete roughly a quarter of the file?

Agreed and dropped in 34eef4471. The provider now requires app v2.13+ and fails setup with an error telling the user to update the app when the endpoint is missing. Because of that the whole legacy path went with it: the provider shrinked from 842 to 543 lines, the tests from 1844 to 1511 (rewritten for v1-only, then extended to full line coverage).

music_assistant/providers/mammamiradio/__init__.py
florianhorner: Kept deliberately: urlsplit defers port parsing, so without this line an out-of-range port surfaces later as a (confusing) aiohttp error instead of the localized invalid-URL setup error. Covered by the URL accept/reject matrix tests.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > Use Sphinx-style docstrings with :param: syntax. For simple functions, a single-line docstring is fine. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.
> 
> This is a general comment for throughout this file.

Done in 34eef4471: docstrings now matched to the project style, single-line where simple, :param: only where it helps the caller. Explanations of internals moved to inline comments or got deleted.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > The comment above says album should be None as the fallback?

Fixed in 34eef4471: the fallback is now None as the comment says. The station name is no longer substituted when the app doesn’t send an album.

florianhorner:
> Overall looks pretty good. A few comments above and please also supply an icon_monochrome.svg

Thank you. Added in 0243c7c0e: single-tone, currentColor, under the 5KB budget.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > MA reloads the provider on a config change so you should just do the url building in handle_async_init and remove this

Done in 34eef4471: the URL is validated and stored once in handle_async_init; the caching accessor is gone.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > I don't think response.json() can raise TypeError. Thus this should be removed in the three places in the code where this appears.

Done in 34eef4471: the URL is validated and stored once in handle_async_init; the caching accessor is gone.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > We can't have broad catchalls. Please narrow this to the expected errors. This is a general comment for throughout the code.

Fair one to narrow down. Removed both in 34eef4471. Every field access in the mapper is type-guarded (the malformed-payload tests pin that), so there were no expected errors left to catch.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > Log messages already show the provider. Remove this from all log messages.

Done in 34eef4471: prefix dropped from all log messages.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: Removed in 34eef4471.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > This is impossible as MA only exposes the provider after `handle_async_init` succeeds. Initialise these in `handle_async_init` and drop the comment.

Done in 34eef4471: initialised in handle_async_init, class-level defaults and the comment removed.

music_assistant/providers/mammamiradio/manifest.json
florianhorner: > And you need to raise a PR for our docs against the BETA branch

Changed in 0243c7c0e. Docs PR: music-assistant/music-assistant.io#799 against beta.

music_assistant/providers/mammamiradio/__init__.py
florianhorner: > Did AI write this? Either way please change to plain english or delete. Same coment for line 364

Done in 34eef4471: one reworded in plain English, the other went away with the legacy mapper. I also swept the rest of the file and removed that style.

And yes, AI-assisted development workflows. The review, testing, maintenance and answering for it all are on me though. Plus I run this thing daily, so I'm usually the first to notice when it breaks.

florianhorner:
I addressed all items from the review and every thread is resolved. CI is green on the current head and the Copilot re-review came back with no new findings. The docs PR is open at music-assistant/music-assistant.io#799. Back to you @OzGav, thanks for the thorough round.

music_assistant/providers/mammamiradio/__init__.py
```diff
@@ -0,0 +1,565 @@
...
+                        or not isinstance(schema, (str, int))
+                    ):
+                        # No usable version field: treat as a pre-2.13 addon (or
+                        # some other service answering on this port).
+                        raise ProviderUnavailableError(requires_msg)
+                    msg = (
+                        f"Mamma Mi Radio addon at {self._base_url} publishes unsupported "
+                        f"now-playing schema_version {str(schema)[:32]!r}; this provider "
+                        "supports v1 (addon 2.13+)"
+                    )
```
OzGav, line 451: There is a lot of complexity here. Your addon always sends `schema_version` as a string, so the probe only needs to answer two questions. Is it a string? Is it a supported version? If it isn't a string, raise "requires addon 2.13+". If it is a string but not a supported version, raise "unsupported version". The None and bool and int checks only exist because `_supports_v1_schema` (back on line 110) converts ints to strings, which is defending against a payload your own addon can never send. If you remove the int and bool handling from `_supports_v1_schema` then this whole block becomes two simple checks.

OzGav:
Also as an aside I tried to install the stable addon in HA but it is looking for the unreleased v2.18?

music_assistant/translations/en.json
```diff
@@ -1467,6 +1467,11 @@
...
   "provider.lrclib.config_entries.api_url.label": "API URL",
   "provider.lrclib.manifest.description": "Primary lyrics provider, delivering synced and unsynced lyrics from the LrcLib database.",
+  "provider.mammamiradio.config_entries.mammamiradio_url.description": "Full HTTP(S) base URL of your Mamma Mi Radio HA addon. The default works when Music Assistant and the addon run on the same Home Assistant host; remote or custom deployments need a reachable URL. Do not include an admin token.",
+  "provider.mammamiradio.config_entries.mammamiradio_url.label": "Mamma Mi Radio URL",
+  "provider.mammamiradio.errors.invalid_base_url": "Invalid Mamma Mi Radio URL. Enter a full HTTP(S) base URL with a hostname, for example http://localhost:8000.",
+  "provider.mammamiradio.manifest.description": "Two Italian hosts. One very opinionated smart home. Self-hosted radio for Home Assistant: music, banter, and ads, with your home's moments worked into the show.",
+  "provider.motherearthradio.manifest.description": "Stream audiophile-quality internet radio — human-curated, ad-free, and listener supported.",
```
OzGav:
```suggestion
```

OzGav:
Just one thing left @florianhorner and we will get this merged!

florianhorner:
Thanks @OzGav, and sorry for the lag. I will finish the leftover wok this week and then re-request review.

In detail:
The Mother Earth Radio translation line is already gone on the current head. I’ll also take your Jul 26 note on simplifying the `schema_version` probe (drop the bool/int special-casing and keep the two-question check).

Stable add-on advertising unreleased v2.18 is fixed on our side, thank you for flagging. The store advertises the stable release again (currently: v2.17), as it should.

OzGav:
Awesome. Looking forward to getting this in.

OzGav:
I was just looking at the docs and saw this "After setup, find Mamma Mi Radio via browse or search and add it to your library" 

Normally with radio providers we only allow auto-adding to the MA library if the provider has a library to sync. So what you are doing is correct. However, since this only exposes one station and the user has installed this to get that station it will be better UX if we just add the station to the library automatically.

MarvinSchenkel:
There's 1 other thing we need to do which is introducing a `setup_flow.py`. This is a recent addition in `dev` and I think the provider won't be able to be configured without it. Other than that this looks good to go 🎉

OzGav:
Just letting you know feature freeze for 2.10.0 is next week. No stress if you can’t make it but just letting you know.

MarvinSchenkel:

music_assistant/providers/mammamiradio/__init__.py
florianhorner: Done in c90d178d5: `_supports_v1_schema` is now one single string check. The probe is the two questions you described: not a string, so pre-2.13 addon; a string but not a supported version, so unsupported version.

I removed the None, bool and int branches together with the int-to-string coercion.

florianhorner:
> Also as an aside I tried to install the stable addon in HA but it is looking for the unreleased v2.18?
https://github.com/music-assistant/server/pull/3836#issuecomment-5173132336

Answered in the other thread but want to give an update to that. Answer there was:
> Stable add-on advertising unreleased v2.18 is fixed on our side, thank you for flagging. The store advertises the stable release again (currently: v2.17), as it should.

Which was correct when I made the comment. Meanwhile [v2.18.0 is the current stable release](https://github.com/florianhorner/mammamiradio/releases/tag/v2.18.0), since Aug 8th.

OzGav (approved):
Thanks for your hard work on this @florianhorner !

PR #3837, opened 2026-05-04
Enhance warning for Spotify top tracks fetch failure

OzGav:
The user can’t do anything about this and it is expected behaviour for some accounts so a warning is not appropriate.

### DIFF
--- a/music_assistant/providers/spotify/provider.py
+++ b/music_assistant/providers/spotify/provider.py
@@ -666,7 +666,7 @@ class SpotifyProvider(MusicProvider):
                 if (item and item["id"])
             ]
         except MediaNotFoundError:
-            self.logger.info(
+            self.logger.warning(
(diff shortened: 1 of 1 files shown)


MarvinSchenkel:
I agree with the textual change here, but why not leave it a warning? It warns the user that top tracks are not available for the account.

OzGav:
My view is that a warning is something has gone wrong and the user should do something about it? In this case we know Spotify iis removing the endpoint so this is normal but info this is why you see no tracks?

MarvinSchenkel:
No an ERROR is used when something goes wrong (unexpectedly). We can use a warning to, well, **warn** the user about unexpected behaviour

OzGav:
I guess the unusual thing here is this is kind of expected (eventually)? However, I will change back to a warning.

MarvinSchenkel (approved):
Thanks! :)

PR #3841, opened 2026-05-05
Fix Sendspin Cast bridge silently failing to set up

maximmaxim345:
The Cast bridge tried to write the model-specific static delay default into the player config DB on first connect. When the player config row did not yet exist the write and read raised, and the Sendspin Cast bridge silently failed to set up.


MarvinSchenkel (approved):
LGTM!

PR #3844, opened 2026-05-06
Add Rainy Mood provider

jlpouffier:

Adds a **Rainy Mood** provider that offers a looping rain ambience from [rainymood.com](https://rainymood.com) as a sound effect. You can play it directly, or select it as the audio overlay on a queue to gently mix rain on top of your music.

This is now a small, isolated content provider that builds on the sound-effect and audio-overlay support already in `dev`. The provider only hands the rain stream to the player pipeline — the core overlay engine takes care of the looping, format conversion and volume mixing.

**What's in this PR:**

- New `rain_mood` music provider exposing a single "Rain" sound effect
- Streams the rain audio straight to the player over HTTP (no local caching)
- Translations for the sound effect name/description
- Tests for the provider

- Frontend companion (queue overlay toggle): music-assistant/frontend#1753

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/rain_mood/__init__.py
+++ b/music_assistant/providers/rain_mood/__init__.py
@@ -1,0 +1,108 @@
+"""
+Rainy Mood provider for Music Assistant.
+
+Serves a looping rain ambience from rainymood.com as a sound effect, suitable for
+direct playback or as the source for a queue's audio overlay. The remote stream is
+handed straight to the player pipeline - the audio overlay engine takes care of the
+looping, format conversion and volume mixing.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from music_assistant_models.enums import ContentType, MediaType, ProviderFeature, StreamType
+from music_assistant_models.errors import MediaNotFoundError
+from music_assistant_models.media_items import AudioFormat, ProviderMapping, SoundEffect
+from music_assistant_models.streamdetails import StreamDetails
+
+from music_assistant.models.music_provider import MusicProvider
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncGenerator
+
+    from music_assistant_models.config_entries import ConfigEntry, ConfigValueType, ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.BROWSE,
+    ProviderFeature.SOUND_EFFECTS,
+}
+
+RAIN_ITEM_ID = "rain"
+RAIN_URL = "https://media.rainymood.com/0.mp3"
+
+
+async def setup(
+    mass: MusicAssistant, manifest: ProviderManifest, config: ProviderConfig
+) -> ProviderInstanceType:
+    """Initialize provider(instance) with given configuration."""
+    return RainyMoodProvider(mass, manifest, config, SUPPORTED_FEATURES)
+
+
+async def get_config_entries(
+    mass: MusicAssistant,  # noqa: ARG001
+    instance_id: str | None = None,  # noqa: ARG001
+    action: str | None = None,  # noqa: ARG001
+    values: dict[str, ConfigValueType] | None = None,  # noqa: ARG001
+) -> tuple[ConfigEntry, ...]:
+    """Return Config entries to setup this provider (none needed)."""
+    return ()
+
+
+class RainyMoodProvider(MusicProvider):
+    """Music provider serving a looping rain ambience from rainymood.com."""
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        return True
+
+    async def get_sound_effect(self, prov_sound_effect_id: str) -> SoundEffect:
+        """Get full sound effect details by id."""
+        if prov_sound_effect_id != RAIN_ITEM_ID:
+            raise MediaNotFoundError(f"Unknown sound effect: {prov_sound_effect_id}")
+        return self._build_sound_effect()
+
+    async def get_sound_effects(self) -> AsyncGenerator[SoundEffect]:
+        """Get all sound effect items this provider offers."""
+        yield self._build_sound_effect()
+
+    async def get_stream_details(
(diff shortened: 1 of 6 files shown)


music_assistant/controllers/streams/audio.py
```diff
@@ -2247,3 +2246,62 @@ async def _update_hls_radio_metadata(
...
+        """
+        for prov in self.mass.providers:
+            if prov.type != ProviderType.PLUGIN:
+                continue
+            if callable(get_overlay := getattr(prov, "get_player_overlay", None)):
+                if overlay := get_overlay(player_id):
+                    return cast("tuple[Any, float]", overlay)
+        return None
+
+    async def apply_rain_overlay(
```
marcelveldt: this should be more generic / reusable

rain is the actual implementation (covered in the plugin) so the global implementation would be optional overlay support

global function that gets the active overlay (returned as string for an ID)
global function that routes the request to the right plugin based on that id 

ProviderFeature.AUDIO_OVERLAY added to models

music_assistant/controllers/streams/audio.py
```diff
@@ -2247,3 +2246,62 @@ async def _update_hls_radio_metadata(
...
+        self,
+        music_gen: AsyncGenerator[bytes, None],
+        rain_reader: Any,
+        rain_vol: float,
+        pcm_format: AudioFormat,
```
marcelveldt: this assumes that the pcm format of the audio and audio overlay is the same which it wont be

music_assistant/controllers/streams/audio.py
```diff
@@ -2247,3 +2247,88 @@ async def _update_hls_radio_metadata(
 
         except Exception as err:
             self.logger.debug("Error fetching HLS metadata: %s", err)
+
+    # ------------------------------------------------------------------
+    # Audio overlay support (used by Rain Mood and similar plugins)
+    # ------------------------------------------------------------------
+
+    def get_active_overlay_provider(self, player_id: str) -> PluginProvider | None:
```
MarvinSchenkel: What should happen if multiple overlays are active?

music_assistant/controllers/streams/audio.py
```diff
@@ -2247,3 +2247,88 @@ async def _update_hls_radio_metadata(
...
+        """
+        Return the first plugin with AUDIO_OVERLAY feature that is active for the given player.
+
+        :param player_id: The player to check.
+        :returns: PluginProvider instance, or None.
+        """
+        for prov in self.mass.providers:
```
MarvinSchenkel: You can use `mass.get_providers_supporting_feature` here instead

music_assistant/controllers/streams/audio.py
MarvinSchenkel: @marcelveldt We should consider using ffmpeg filters here instead of calculating this ourselves. I think something like `amix` might be able to do what we want.

jlpouffier:
## Transparency note: seek is not seamless for the rain overlay

During testing I confirmed an audible artifact specifically on **seek** (skipping inside the currently playing track). It is **not** present on track transitions, which remain seamless. Wanted to flag it openly rather than have it surprise a reviewer.

### Symptom

On Sendspin (and to a lesser extent on Sonos), seeking inside a track causes the rain content position to jump forward — up to ~30 seconds on Sendspin's pipeline, less on Sonos. Music position is correct; only the rain "phase" appears to skip.

### Root cause

It's an architectural consequence of how live overlays interact with downstream pre-buffering rather than a bug in the mixer:

```
RainBuffer ffmpeg (continuous) ──► mix ffmpeg ──► player pre-buffer (up to ~30s on Sendspin) ──► speaker
```

The mix process consumes rain bytes as fast as the downstream player buffer wants them. Sendspin in particular pre-caches up to 30s of mixed audio (per `_PRODUCER_BUFFER_LIMIT_US`). The rain ffmpeg has already produced — and the mix has already consumed — that "future" rain content. On a **track transition** the player drains its buffer naturally, so every pre-consumed rain byte is played. On a **seek** the player discards the buffer, so the pre-consumed rain bytes are lost; the new mix process resumes reading rain from where the RainBuffer pipe is *now*, which is ahead of where playback resumes.

### What I tried

Pacing the rain ffmpeg at real time with `-re`: this prevents over-consumption (limits the skip to ≤ player's minimum start buffer) but introduces a 5+ second pause after every seek while the player waits for buffer to fill at 1×. Pause was clearly worse UX than the skip, so I reverted it (rationale: the rain content is non-rhythmic ambient noise; a forward jump is barely noticeable unless one is listening for a specific feature, whereas a 5s pause is immediately obvious).

marcelveldt:
Hi @jlpouffier, heads-up: #3938 just landed a refactor of the plugin sources architecture. Plugin sources are now first-class `AudioSource` MediaItems and the `PluginProvider` contract has changed — Rainy Mood will need a small migration to land on top of `dev`:

- `get_source()` → `async def get_audio_sources() -> list[AudioSource]`
- `get_audio_stream(player_id)` → `get_audio_stream(streamdetails, seek_position)` (matches `MusicProvider`)
- Add `async def get_stream_details(source_id, queue_id) -> StreamDetails` (raises `ResourceBusyError` on exclusive collisions)
- Optional `async def on_source_control(source_id, action, value)` if you want play/pause/seek controls
- Live metadata updates go through `mass.streams.update_stream_metadata(queue_id, StreamMetadata(...))`

Canonical example: [`vban_receiver`](https://github.com/music-assistant/server/blob/feat/audio-sources/music_assistant/providers/vban_receiver/provider.py); the demo template at `_demo_plugin_provider` shows the recommended shape.

marcelveldt:
@jlpouffier I took your idea and extracted it into a core functionality to provide sound effects and overlays and refactored your PR into a plugin dedicated to rainy mood. Thanks for kickstarting this!

marcelveldt (approved):
Thanks, @jlpouffier !

OzGav:
Do we need any documentation for this?

Hedda:
For reference, started feature request discussion based on this idea/concept, but a generic one with cache + local-file support:

* [[Feature Request] Create a audio loops overlay provider using locally cached sound effects files to play relaxing sounds like waves hitting a beach/shore and breezy winds ](https://github.com/orgs/music-assistant/discussions/6206)

PS: There is still no documentation available for the Rainy Mood provider -> https://www.music-assistant.io/music-providers/

PR #3847, opened 2026-05-06
Allow AirPlay2 devices to be selected for synchronised playback

bradkeifer:
Fix prior merge error and correct description for Session Establishment Latency config item


MarvinSchenkel (approved):
It's party time 🎉

PR #3848, opened 2026-05-07
AirPlay 2: Add PTP timing support

bradkeifer:
Adds support for PTP timing to be used for AirPlay 2 players. 
AirPlay 2 implementation of Shairport Sync only supports PTP timing, and does not work with NTP timing.
Samsung devices appear to only support PTP timing for their AirPlay 2 implementation also.
PTP timing with AirPlay 2 is also a prerequisite to future implementation of buffer RTP and hi-res audio support into the AirPlay 2 provider.


bradkeifer:
Over to you @MarvinSchenkel for the python changes.

dantelope:
Hi @bradkeifer @MarvinSchenkel — happy to help test on Samsung Frame whenever this is closer to a testable state.

Test environment:
- Samsung Frame TV (post-2018, AirPlay 2 confirmed working from native iOS/macOS clients)
- HA + Music Assistant on Linux x86_64 (Beelink, HA OS amd64)
- Frame currently auto-discovered by MA's AirPlay 2 provider and flagged via `BROKEN_AIRPLAY_MODELS` — silent playback as expected
- A couple of working AirPlay 2 targets (Mac, Kitchen Display hardware via Google Cast) for sync-group comparison

I can verify: basic playback (does audio come out), volume control, sync grouping, and behavior with `BROKEN_AIRPLAY_MODELS` bypassed. Can capture mDNS/RTSP packet traces and MA logs if useful.

Two questions:
1. Is there a branch I can try now, or should I wait for the Python side?
2. Will `airptpd` run as a sidecar alongside `cliap2`, or replace AirPlay 2 entirely in that path?

Thanks for the work on this — Samsung Frame in particular is a common-enough device that even a "known issue resolved" landing would be a big quality-of-life win.

dhl84:
**Confirmed working on a Samsung Music Frame HW-LS60D** — it now plays AirPlay 2 with sound, tightly PTP-synced with HomePods, driven entirely by Music Assistant (2.9.5, `cliap2` @ owntone `3801bcc`).

Sharing the one non-obvious gotcha in case it helps close this out: **`airptpd` must be launched in the foreground (`-f`).** If it self-daemonizes (the default), the double-fork abandons the libevent thread running `shm_update_cb` (the `DAEMON_INTERVAL_SECS_SHM_UPDATE = 5s` heartbeat that refreshes `daemon_info.ts`). The shm timestamp is then written once at startup and never updated, so after `AIRPTP_STALE_SECS` (15s) `airptp_daemon_find()` treats it as stale, and `cliap2` logs `AirPlay PTP daemon unavailable, only NTP will be available` → NTP fallback → Samsung AP2 receivers silently drop it. Symptom: session establishes, device shows "playing", **zero audio**. `airptpd -f` fixes it (ts refreshes every 5s, `cliap2` attaches, `use_ptp` flips on, Samsung plays).

So the missing server-side piece is just: ship `airptpd` (the prebuilt binaries in this PR work — same struct `version_major`) and **launch one shared instance with `-f`** (host-net, binds UDP 319/320 once per host) before any stream; that shared daemon is what lets the per-player `cliap2` processes share one clock domain and sync.

For the proof I baked it into a forked add-on whose entrypoint runs the daemon alongside the server:

```sh
( while true; do /usr/local/bin/airptpd -f; sleep 2; done ) &
exec /usr/local/bin/entrypoint.sh --data-dir /data --cache-dir /data/.cache "$@"
```

A Python `AsyncProcess` launch in `AirPlayProvider` setup (before streams) would slot into this PR cleanly, plus dropping `('Samsung', '*')` from `BROKEN_AIRPLAY_MODELS`.

@dantelope happy to compare notes — mine's a Music Frame HW-LS60D and it's rock solid so far; glad to share the add-on build for testing.

MarvinSchenkel:
@dhl84 Thanks for testing! I just added the necessary changes to this PR that should automatically make this work as part of Music Assistant. Are you able to take this for another test spin? 🙏

DerPicknicker:
Just to comment as on discord:
- Bose SoundTouch are playing fine within MA 
- Samsung speaker is not starting to play (metadata is visible) but I don’t hear any sound

DerPicknicker:
With my Samsung music frames and my low end host I have to adjust the ms for etablished connection up to 4000ms.

I tested 1000-2500 without a success and then directly 4000ms. Lower values than 4000ms aren't tested by me.

chelecuche:
Just wanted to add another data point — this isn't limited to Samsung devices. I've been hitting the same issue on a **HomePod mini** (native Apple AirPlay 2 device) with automation-triggered TTS.

Same error signature:
```
AirPlay PTP daemon unavailable, only NTP will be available
cliap2: ntpstart time too soon
Unable to obtain feasible playback start time
```

Here's the weird part — **manual playback works fine** through Developer Tools or the MA UI. It's specifically automation-triggered playback that fails. The race condition seems to be that automations fire faster than the NTP fallback can stabilize, so by the time `cliap2` asks for a start time, NTP hasn't synced yet.

Workaround that's been solid for weeks:
- **3 seconds of silence** before the announcement
- **4000 ms session-establishment latency**

RAOP (AirPlay 1) works too but has choppy audio at the start.

Figured I'd mention it since it proves the PTP/NTP issue isn't just a Samsung quirk — it affects **any** AirPlay 2 device when automation timing exposes the NTP startup race. Looking forward to seeing the PTP daemon land. Happy to test the PR branch on a HomePod if that helps.

marcelveldt:
Superseded by https://github.com/music-assistant/server/pull/4879

PR #3849, opened 2026-05-07
Refactor Fully Kiosk to single-instance

OzGav:
BREAKING CHANGE!

Refactored Fully Kiosk to single-instance, mirroring the MPD provider. Devices live in the provider config (host or host:port); password and SSL options are per-player so they show up in `needs_setup` until the password is filled in.
                                                                                                                                                                 
One-off migration in `controllers/config.py` collapses any existing `fully_kiosk--*` instances into the new shape and copies passwords/SSL across. Tagged for removal in 2.10.                                                                                                                                               
   
BREAKING CHANGE: player IDs change from device hardware ID to `fully_kiosk_<host>_<port>`, so existing universal group membership won't be migrated and will have to be setup again. 


marcelveldt:
why is this a breaking change if you can simply migrate the player configs ?

OzGav:
Just because I think any groups will need to be recreated. Probably very small chance but I wanted to be cautious as we are getting heat from some users when we break stuff and dont tell them...

music_assistant/providers/fully_kiosk/provider.py
```diff
@@ -2,116 +2,100 @@
...
-            f"{scheme}://{self.config.get_value(CONF_IP_ADDRESS)}:"
-            f"{self.config.get_value(CONF_PORT)}"
-        )
-        player = FullyKioskPlayer(self, player_id, fully_kiosk, address)
-        player.set_attributes()
-        await self.mass.players.register(player)
+    async def loaded_in_mass(self) -> None:
+        """Sync registered players against the current hosts config."""
+        entries = cast("list[str]", self.config.get_value(CONF_MANUAL_IPS) or [])
+        new_ids = {f"fully_kiosk_{h}_{p}" for h, p in (_parse_host_entry(e) for e in entries)}
```
MarvinSchenkel, line 59: Why are we using new ids here rather than the old `fully_kiosk.deviceInfo["deviceID"]` ? I think using that would save you the migration logic + makes this change non breaking?

OzGav: The reason I went with host:port IDs is a bit of a chicken/egg thing. In the new setup the password lives on the player config rather than the provider, so by the time we register the player we haven't actually talked to the device yet and don't know its deviceID.
  
I guess we could preserve the old deviceID for existing installs by reading it out of the legacy player config during migration and keeping it around but it would mean two ID shapes coexisting (legacy = deviceID, new = host:port), which is doable but adds a bit of complexity to carry forward.

Given the only thing that I think would actually break is universal groups containing a Fully Kiosk player, and my guess is that is not that likely that many users would have that, I figured a one-time re-link was the lighter trade-off. Happy to do the preserve-old-ID approach though if you think it's worth it!

marcelveldt: I would just go with the new ID's but if you can find a way to do an auto migration that would always be useful ofcourse!

Maybe do a quick migration run in the provider's init code that grabs existing players from config and adjusts them

OzGav: OK I looked into what would be required in addition to the existing migration and it's three things:

  1. controllers/config.py — extends the existing migration to build an old→new player_id map, rewrite references in group_members, members_filter, allowed_members, linked_protocol_ids and the party player target, then delete the orphan device-id player configs.
  2. controllers/player_queues.py — add a small public method to rename persisted queue cache entries. Needed because the cache isn't up at config-load time, and I dont think we want the provider reaching into player_queues' cache layout directly.
  3. providers/fully_kiosk/provider.py — consume an id_map handoff field on first load and call the player_queues rename method.

All this would be deleted alongside the existing migration in 2.10. I've tested it end-to-end and it works as far as I can tell but I was looking at all the changes and thought I would quickly ask if the overhead is desired? Happy to push it, but also happy to leave it if you'd rather not

marcelveldt: 1. Not needed - if there is any migration it should live in the probvider itself. The core config controller should not contain provider specific migration code.

2. same here, too much overhead - its fine if the cache isnt restored at all

3. see above

OzGav: After looking into it I believe I can move everything into the provider except one thing. The collapse migration deletes the legacy per-device provider instances, and that instance was the only link between a device ID (the old player_id) and its host. The legacy player config only stores password/SSL, so afterwards nothing can tell the provider which old device-id maps to the new `fully_kiosk_{host}_{port}` id.

So the only bit left in `config.py` is capturing that old to new map and handing it to the provider, which does all the actual rewriting. If you'd rather keep `config.py` completely clean, the alternative is to drop the remap entirely and just accept that the rare user with a universal group containing a Fully Kiosk player will need to re-add it.

marcelveldt: yes --> just accept that the rare user with a universal group containing a Fully Kiosk player will need to re-add it

OzGav: OK then I think this PR is good as is. It does a basic migration by collapsing the legacy per-device provider instances into the single instance, preserving each device with its password/SSL settings. It doesn't carry over `player_id` references, so it logs a warning that any universal group referencing the old IDs needs re-linking.

marcelveldt (approved):

PR #3851, opened 2026-05-07
audio_analysis: add coverage endpoint + bulk merged accessor for sonic_similarity

chrisuthe:
Adds backend support for the upcoming frontend coverage panel and the sonic_similarity index rebuild ([#3943](https://github.com/music-assistant/server/pull/3943)).

- `audio_analysis/coverage` API command — returns `analyzed` / `pending` / `stale_version` / `analysis_version` for a given AA provider, typed via [`AudioAnalysisCoverage`](https://github.com/music-assistant/models/pull/231).
- `iter_merged_audio_analysis_rows` — streams one merged `AudioAnalysisData` per track across available AA providers. Memory stays proportional to one track, not the whole library.

Requires `music-assistant-models==1.1.122`.

## Types of changes


music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -608,17 +608,102 @@ async def export(
...
 
         return {"total": total, "offset": offset, "limit": limit, "items": items}
 
+    @api_command("audio_analysis/track")
+    async def track(
+        self,
+        aa_domain: str,
+        item_id: str,
+        provider_instance_id_or_domain: str,
+    ) -> dict[str, Any] | None:
```
MarvinSchenkel: I would expect this to return a typed `AudioAnalysisData` object?

chrisuthe: Yea it absolutely should, I thought I caught that once before and fixed it but I think I didn't push my changes from laptop before I started working on the desktop again.

chrisuthe: I'm going to return a list of AudioAnalaysisData objects here, it may be empty if nothing was found, or a single analysis object if the domain was passed in.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -355,6 +358,355 @@ async def get_audio_analysis_version(
...
+            DB_TABLE_AUDIO_ANALYSIS,
+            {
+                "aa_provider_domain": aa_provider_domain,
+                "media_type": media_type.value,
+            },
+            limit=limit,
+            offset=offset,
+        )
+
+    async def get_merged_audio_analysis_rows(
```
MarvinSchenkel: This seems unused.

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -138,6 +137,51 @@ async def get_config_entries(
     )
 
 
+def select_clap_window(audio: np.ndarray, source_sr: int) -> np.ndarray | None:
```
MarvinSchenkel: Did you mean to add these select_clap_window + window functions? I thought we explicitly removed them in a previous PR as they are unusued.

music_assistant/providers/sonic_analysis/clap_prompts.py
```diff
@@ -142,3 +142,6 @@ def validate_calibration_freshness(
             CALIBRATION_PROMPTS_HASH,
             actual,
         )
+
+
+validate_calibration_freshness()
```
MarvinSchenkel: Same here, this was explicitly removed in the last PR

music_assistant/providers/sonic_analysis/__init__.py
```diff
@@ -228,6 +272,8 @@ def _pcm_bytes_to_audio(audio_format: AudioFormat, pcm_chunk: bytes) -> np.ndarr
     :param audio_format: The audio format describing the PCM data.
     :param pcm_chunk: Raw PCM audio data.
     """
+    import torch  # noqa: PLC0415
```
MarvinSchenkel, line 228: What is the use case for an inline import in this provider? Why not a top level one?

music_assistant/models/audio_analysis_provider.py
```diff
@@ -160,6 +160,17 @@ async def finalize(self, session_id: str) -> None:
                     )
         self._sessions.pop(session_id, None)
 
+    async def get_provider_status(self) -> dict[str, Any]:
```
MarvinSchenkel: Do we _really_ need this? This feels like an unnecessary addition?

music_assistant/controllers/streams/audio_analysis.py
MarvinSchenkel, line 1: This one is for the entire controller :) :

**API command function names**
We should keep the naming of api_commands consistent with the rest of the codebase, so `get_<thing>` or `list_<thing>`. So instead of
```
@api_command("audio_analysis/analyzed_tracks")   
async def analyzed_tracks(
```
we should use
```
@api_command("audio_analysis/analyzed_tracks")
async def get_analyzed_tracks(
```

**AA models**
Let's add a couple of dataclasses to the [models package](https://github.com/music-assistant/models). This will allow us to easily work with them and create/generate a frontend interface to populate on the frontend side. These models should really represent what we want to show in the frontend, not necessarily the database row. So for example, I feel we could drop unnecessary info such as `provider_loaded: True` from the `status` function. Have a look at the DSP models in the models package for some inspiration.

**Pagination**
Pagination is handled automatically by the websocket client if we use an AsyncGenerator. Have a look at [`playlist.tracks`](https://github.com/music-assistant/server/blob/75c6c18f0b42f88370dcbd0e801744cb095d72c5/music_assistant/controllers/media/playlists.py#L151) or [`iter_library_items`](https://github.com/music-assistant/server/blob/75c6c18f0b42f88370dcbd0e801744cb095d72c5/music_assistant/controllers/media/base.py#L308) how to do this

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -355,6 +358,355 @@ async def get_audio_analysis_version(
...
+                artists = ", ".join(a.name for a in getattr(t, "artists", []) or [])
+                return {"item_id": item_id, "name": t.name, "artist": artists}
+            except Exception as err:
+                self.logger.debug("Failed to resolve track %s/%s: %s", provider_id, item_id, err)
+                return {"item_id": item_id, "name": "(unknown)", "artist": ""}
+
+        items = list(await asyncio.gather(*[_resolve(iid, prov) for iid, prov in entries]))
+        return {"total": total, "offset": offset, "limit": limit, "items": items}
+
+    _EXPORT_SCALAR_FIELDS: tuple[str, ...] = (
```
MarvinSchenkel: Why a subset of the database table? I feel export/import should be able to recreate _entire_ database rows?

MarvinSchenkel (changes requested):

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: It was top level, I followed the ytmusic provider's lead on moving the import inline to try to avoid the hang prior to the save button. My understanding was it was two things primarily that hung that, this import and the provider dependency installer in helpers/util.py (it runs with no cache) - I didn't think that was flexible so I moved this instead.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Going to hit these with a few replies, bear with me.
Naming, updated
status == get_status
analyzed_tracks == list_analyzed_tracks
export == get_analysis_export
track == get_track
coverage == get_coverage

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Yea, this was for the Analysis plugin, I should add this with that if it's really needed, this is speculative. I'll slice it out.l

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Indeed, good catch, we do not need these here.

music_assistant/providers/sonic_analysis/clap_prompts.py
chrisuthe: So this I think needs to stay, it's a tripwire that tell us if the prompts that generate the clap scores have changed against the calibrated values.  We have modification scalars in place that were tuned on a 50 track sample set, if the prompt changes, those are possibly also in need of adjustment because we have two inputs there (prompt + calibration co-efficient) - if the prompt gets adjusted we should be reminded that the calibration numbers should be too.

music_assistant/models/audio_analysis_provider.py
chrisuthe: I'm overriding with information on if the clap model is loaded (since we're downloading 300+mb of model and then another 200+mb of the text model when needed)  so we know when the analysis runs if we should bother asking CLAP to participate, and if we can do free text search.  Should I approach differently?

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: RE: Pagination. Yes- I thought I needed something unique here but really I don't (Thanks Claude for showing me how I don't) conversion also fixed a bug I missed on a search I'm doing that only hit a single page prior. Shifting to stream paging.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: This has no consumer, I am going to remove it for now, if we want an actual export we should do what you're suggesting, but for now there isn't a reason to guess at what would be useful.

music_assistant/models/audio_analysis_provider.py
MarvinSchenkel: This is what `provider.available` is for. This is `False` by default and will be set to true once the `handle_async_init` finsishes. So if you make sure to download the model there, no analysis work will be scheduled to your provider during that period as the AA controller already checks on provider availability

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -81,6 +83,32 @@ def setup(self) -> None:
             allow_retry=True,
         )
 
+    def _register_api_commands(self) -> None:
```
MarvinSchenkel: Why not just add `self.streams.audio_analysis,` to the list in mass.py?

music_assistant/models/audio_analysis_provider.py
MarvinSchenkel: So instead of using a task you just await it in the async init and then we can just remove this

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -231,6 +262,35 @@ async def set_audio_analysis(
             },
         )
 
+    @staticmethod
```
MarvinSchenkel: Why a static method? Think we can just move this to the bottom with the other helper functions. Or define as a helper function outside the class

music_assistant/providers/sonic_analysis/clap_prompts.py
MarvinSchenkel: Since this is all based on static code. Let's just add a unit test instead of using a sideeffect on module import. E.g. something like 
```
def test_clap_calibration_hash_matches_prompts() -> None:
    """Tripwire: if CLAP prompts change, calibration coefficients must be re-tuned."""
    assert _compute_prompts_hash() == CALIBRATION_PROMPTS_HASH, (
        "CLAP prompts changed but CALIBRATION_PROMPTS_HASH was not updated. "
        "Re-run calibration on the 50-track sample set and update the hash + scalars."
    )
```

chrisuthe: Agreed, and done!

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Actually I want to make it clear, this isn't sliced out and it isn't unused, I left it here for the plugin which uses this to get bulk data that has all the most recent data across multiple AA rows.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: that is indeed the correct way of doing this, added and cleaned up.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Makes sense on review, I moved it down to above the class, is that the expected place?

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Resolved in commit c507a67

music_assistant/providers/sonic_analysis/__init__.py
chrisuthe: Resolved in commit 38de7bb

music_assistant/providers/sonic_analysis/clap_prompts.py
chrisuthe: Resolved in commit 6c42bbd

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Resolved in commit 869a892

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Resolved in commit 23a0eed

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Resolved in commit 7761993

music_assistant/providers/sonic_analysis/clap_prompts.py
chrisuthe: Resolved in commit 6c42bbd

music_assistant/models/audio_analysis_provider.py
chrisuthe: Resolved in commit a00599f

MarvinSchenkel (changes requested):
Looks good, just remove your personal fork for the models and we can merge this

MarvinSchenkel (approved):
Look good, thanks @chrisuthe 🙏

PR #3853, opened 2026-05-08
Detect Sendspin Cast Receiver failures and show them in the frontend

maximmaxim345:
Sendspin Cast players previously had no mechanism to signal failures to users. So, in case the Cast App immediately crashes, there was no feedback indicating that. Specifically newer Cast devices that are built on GC4A 2.0 that can't run Sendspin due to missing APIs would similarly show no feedback to the user.

This PR addresses this by:
- Wiring the Sendspin provider to wait for Cast devices when playing media or grouping speakers, raising an exception that surfaces to the frontend in case anything fails.
- Showing a "Sendspin isn't supported on <device>. Use the standard Cast protocol instead." if a missing `AudioContext` API is detected (on GC4A 2.0 devices). this is shown when starting playback or grouping the speaker.
- Showing a persistent alert in the Sendspin protocol settings of GC4A 2.0 devices: "Sendspin isn't supported on this Cast device. Use the standard Cast protocol instead."
- Detecting if the Sendspin Cast App stopped (e.g. the user casts through their phone to the speaker mid-playback), stopping/ungrouping the player from the Sendspin provider

Notes:
- In addition to this frontend error, a TTS announcement will also play on the speaker.
- The Cast protocol for this speaker is not auto-disabled. Auto-disabling would leave users wondering why their speaker stopped appearing in the grouping list if they missed the initial error message and announcement.

### DIFF
--- a/music_assistant/providers/chromecast/player.py
+++ b/music_assistant/providers/chromecast/player.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import time
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, cast
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


MarvinSchenkel (approved):
Nice, looks good to me @maximmaxim345 👍

PR #3866, opened 2026-05-11
Apple Music: add station recommendations to Discover page

dmoo500:
Adds personalized station recommendations from Apple Music to the Discover page.

## Changes

- Adds `ProviderFeature.RECOMMENDATIONS` to supported features
- Extracts `_get_personal_recommendations()` helper (with `@use_cache(3600)`) shared by both the browse page and the new `recommendations()` method
- Groups stations by recommendation section title into `RecommendationFolder`s (section titles come from Apple's API and are already localized)
- Fixes a bug where `seen` was updated before the live-station check, causing stations to be silently dropped when a live and non-live station shared an ID
- `_browse_stations()` now delegates to the shared helper instead of duplicating logic

### DIFF
--- a/music_assistant/providers/apple_music/__init__.py
+++ b/music_assistant/providers/apple_music/__init__.py
@@ -54,6 +54,7 @@ from music_assistant_models.media_items import (
     MediaItemType,
     Playlist,
     ProviderMapping,
+    RecommendationFolder,
     SearchResults,
     Track,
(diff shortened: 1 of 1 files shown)


dmoo500:
Closing — this functionality is already included in dev via #3715 (Split Apple Music provider into modular structure).

PR #3870, opened 2026-05-11
Use faster `stream/clear` for Sendspin track changes, behind a legacy toggle

maximmaxim345:

Replaces the Sendspin "Allow unencrypted connections" option with a single "Allow legacy clients" compatibility switch, and uses it to gate correct `stream/clear` behavior for track changes.

## `stream/clear` for skips and seeks

The Sendspin spec has a `stream/clear` message meant for skips and seeks, but Music Assistant never used it: it sent `stream/end` immediately followed by `stream/start` instead. Both are valid per the spec, but `stream/clear` lets clients handle seeks and skips more efficiently, and it keeps `stream/end` for what it is meant to signal, an empty queue.

Since some clients might still mishandle `stream/clear`, the new path only runs when legacy clients are disallowed.

## The new "Allow legacy clients" option

Unencrypted transport is itself a deviation from the spec, so the former encryption toggle is folded into one compatibility switch that covers every tolerated deviation and is also passed to aiosendspin's own compliance handling. It defaults to enabled, and the goal is to flip that default and eventually drop the option once clients have caught up.

Turning it off means: no unencrypted clients (which currently includes the built-in web player), clients that deviate from the spec are rejected instead of tolerated, and the server uses `stream/clear` for track changes.

Tested with the latest version of `sendspin-js`. `stream/clear` is used instead of stopping the stream, everything else also keeps working since it implements the latest version.

- 

## Types of changes

## Checklist


kahrendt:
VPE stable release 26.6.0 has the proper bugfix to make this work.

music_assistant/providers/sendspin/provider.py
```diff
@@ -374,13 +374,15 @@ async def handle_async_init(self) -> None:
                 "file-access problem and reload; do not delete the file or all pairings will be "
                 "lost."
             ) from err
+        allow_legacy_clients = cast("bool", self.config.get_value(CONF_ALLOW_LEGACY_CLIENTS, True))
```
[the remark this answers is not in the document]
maximmaxim345: I don't think this is an issue since it's likely that nobody (apart from developers) disabled the option - the majority of clients break with unencrypted clients enabled.
Also porting it will still result in a breaking change if a client was implementing encryption, but still was using the older version of the Spec for the actual roles.

marcelveldt:
Can we first also adjust our web player ?

music_assistant/providers/sendspin/constants.py
```diff
@@ -10,7 +10,7 @@
 CONF_VIRTUAL_PLAYER_OWNER = "virtual_player_owner"
 DEFAULT_SENDSPIN_STATIC_DELAY = 0
 
-CONF_ALLOW_UNENCRYPTED = "allow_unencrypted"
+CONF_ALLOW_LEGACY_CLIENTS = "allow_legacy_clients"
```
MarvinSchenkel, line 13: Do we need a migration for this?

maximmaxim345: IMO no, but Copilot also raised this: https://github.com/music-assistant/server/pull/3870#discussion_r3751080088

MarvinSchenkel (approved):
LGTM

PR #3873, opened 2026-05-11
Fix AA Background Processing to not Monopolize CPU

chrisuthe:
Currently background processing goes As Fast As Possible, using as much CPU as it can while it's running (default midnight to 4 AM). This PR shifts the AFAP to be throttled to at MOST 4x real time, also enforcing that for real-time analysis during playback.  

The background process will run one song at a time unless 8+ cores of CPU are detected, then it will run 2 songs until 16+ are detected then it will run 4.


marcelveldt (approved):
Thanks, @chrisuthe !

PR #3879, opened 2026-05-12
Fix media progress retrieval for open sessions in Audiobookshelf

fmunkes:
If there is an open session, the sync time might not be correct.
See: https://github.com/music-assistant/support/issues/4698


MarvinSchenkel (approved):
🎉

PR #3881, opened 2026-05-12
Support German radio station metadata

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/5462


music_assistant/helpers/util.py
```diff
@@ -410,6 +412,14 @@ def clean_stream_title(line: str) -> str:
     title: str = ""
     artist: str = ""
 
+    if german_match := german_von_pattern.match(line.strip()):
```
MarvinSchenkel: Should we consider moving this down to where we also do the other unstructured checks? I think we are less likely to hit this case so moving it down would save unnecessary checks

OzGav: Yeah good idea. Moved.

MarvinSchenkel (approved):
Thanks @OzGav

PR #3883, opened 2026-05-13
Preserve propagated artist / album genres across refreshes

OzGav:
Follow-up to the prefer-local-genres setting. A user reported that propagated genres on artists/albums (from "Propagate track genres") got wiped when they clicked Refresh Item or Update Metadata. https://github.com/orgs/music-assistant/discussions/5417#discussioncomment-16890419

This happens because propagation writes to the genre mapping table, not to metadata.genres on the artist row. The setting's "is local available" check only looked at `metadata.genres`, so for an artist without a `.NFO` file this evaluated False, online genres merged in, and the next genre scan dropped the propagated entries.

So this PR changes things so that propagation-derived mappings count as "local available". When they exist, online genres get skipped just as they would if genres were found in a `.NFO` file. Online artwork and bio still come through as normal.

### DIFF
--- a/music_assistant/controllers/metadata.py
+++ b/music_assistant/controllers/metadata.py
@@ -55,6 +55,7 @@ from music_assistant_models.unique_list import UniqueList
 from music_assistant.constants import (
     CONF_LANGUAGE,
     DB_TABLE_ARTISTS,
+    DB_TABLE_GENRE_MEDIA_ITEM_MAPPING,
     DB_TABLE_PLAYLISTS,
     VARIOUS_ARTISTS_MBID,
     VARIOUS_ARTISTS_NAME,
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/metadata.py
```diff
@@ -1214,6 +1215,14 @@ async def update_radio_stream_artwork(self, streamdetails: StreamDetails) -> Non
         except MusicAssistantError:
             pass
 
+    async def _has_derived_genre_mappings(self, media_type: MediaType, item_id: str | int) -> bool:
```
MarvinSchenkel: This belongs in the genre controller. Then you can query it here via `mass.music.genres.xxx`

OzGav: Thanks. Moved.

MarvinSchenkel (approved):
Thanks @OzGav !

PR #3885, opened 2026-05-13
Fall back to protocol-scoped key in get_player_config_value

flomotlik:
## Problem

Universal Group (UGP) streams produce **no audio** for any DLNA-backed member (Teufel/Raumfeld, generic UPnP) because every member's HTTP fetch from the UGP stream server raises `KeyError`. Reproducible against a Universal Group whose members are universal_players wrapping DLNA protocol players.

```
File "music_assistant/providers/universal_group/player.py", line 394, in _serve_ugp_stream
    output_format = await self.mass.streams.audio.get_output_format(
File "music_assistant/controllers/streams/audio.py", line 1151
    player.player_id, CONF_SAMPLE_RATES, unpack_splitted_values=True
File "music_assistant/controllers/config.py", line 733, in get_player_config_value
    raise KeyError(msg)
KeyError: 'Config key sample_rates not found for player up501e2d1d53ca'
```

The same KeyError fires for `CONF_HTTP_PROFILE` immediately afterwards in `_serve_ugp_stream` itself. Both lookups go through `ConfigController.get_player_config_value()`.

## Root cause

Since 2.8.0 (`CONF_PROTOCOL_KEY_SPLITTER = "||protocol||"`), universal_player wrappers store per-protocol settings (`sample_rates`, `http_profile`, etc.) under `<protocol_id>||protocol||<key>` rather than at the top level of the player config. Every caller still asks for the bare key, so the top-level lookup fails. The UGP stream handler is the most visible victim, but the latent shape is general.

Live config of an affected universal_player (Teufel One M wrapping DLNA):

```
uuid:640e08e3-…||protocol||sample_rates      = ['44100||16', '48000||16']
uuid:640e08e3-…||protocol||http_profile      = 'no_content_length'
```

No top-level `sample_rates` or `http_profile` exists, so `get_player_config_value(player_id, 'sample_rates')` raises.

## Fix

In `ConfigController.get_player_config_value`, when the bare key is missing and the player has an `active_output_protocol`, retry with `f"{active_protocol}{CONF_PROTOCOL_KEY_SPLITTER}{key}"`. Original `default` / `KeyError` semantics are preserved when no protocol is active.

A single-point fix at the config layer covers every caller (`get_output_format`, `select_flow_format`, `select_pcm_format`, `_serve_ugp_stream`'s `http_profile` lookup, and any future callers).

## Verification

Applied locally on 2.8.7, then created a UGP with two DLNA-wrapped members and played a Spotify playlist via `music_assistant.play_media`. Pre-patch: KeyError stack trace per member fetch, group reports `state=playing` but speakers silent. Post-patch: no errors in logs, both speakers audible. Same result for a single-member UGP.

Happy to add a test if maintainers want one — `get_player_config_value` doesn't currently have direct coverage, so I'd add a fixture rather than extending an existing test.

### DIFF
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -727,10 +727,33 @@ class ConfigController:
                 return raw_value
         conf = await self.get_player_config(player_id)
         if key not in conf.values:
-            if default is not None:
+            # Universal players that wrap a protocol player (DLNA, Sonos, etc.) store
+            # per-protocol settings such as ``sample_rates`` and ``http_profile`` under
+            # ``<protocol_id>||protocol||<key>`` rather than at the top level of the
+            # player config. When such a key is requested without the prefix (as the
+            # universal_group flow stream handler does), the bare lookup raises here
+            # and no audio is served to any DLNA-backed group member. Fall back to
+            # the active output protocol's namespaced key when present.
+            scoped_player = self.mass.players.get_player(player_id)
+            active_protocol = (
+                getattr(scoped_player, "active_output_protocol", None)
+                if scoped_player is not None
+                else None
+            )
+            if active_protocol:
+                scoped_key = f"{active_protocol}{CONF_PROTOCOL_KEY_SPLITTER}{key}"
+                if scoped_key in conf.values:
+                    key = scoped_key
+                elif default is not None:
+                    return default
+                else:
+                    msg = f"Config key {key} not found for player {player_id}"
+                    raise KeyError(msg)
+            elif default is not None:
                 return default
-            msg = f"Config key {key} not found for player {player_id}"
-            raise KeyError(msg)
+            else:
+                msg = f"Config key {key} not found for player {player_id}"
+                raise KeyError(msg)
         if unpack_splitted_values:
             return conf.values[key].get_splitted_values()
         return (


flomotlik:
**Alternative we tested — UGP-localized fix** (rejected after testing)

Before settling on the config-layer fallback, I explored the more surgical option: change the UGP stream URL to address the member's `active_output_protocol` player ID instead of the universal_player ID, so `_serve_ugp_stream`'s `sample_rates` / `http_profile` lookups resolve against the protocol player's config (which has those keys at the top level) without needing a fallback at all.

Two-line patch in `music_assistant/providers/universal_group/player.py`:

```diff
-uri=f"{base_url}?player_id={member.player_id}",
+uri=f"{base_url}?player_id={member.active_output_protocol or member.player_id}",
```

(same pattern in `set_members`).

**Tested locally; here's what I observed**

| Scenario | Result |
|---|---|
| Steady-state play, `active_output_protocol` already populated | Works, audible. |
| **First play after MA restart** | `active_output_protocol` is `None` on every member at the time the URL is built — it gets populated as a *side effect* of the subsequent `_handle_play_media` call. The patch silently falls back to `member.player_id`, the lookup raises `KeyError`, no audio. |
| Any subsequent play in the same session | Works. |

So this approach makes the bug intermittent rather than fixing it: every container restart / first-play-after-boot is silent, the second press works. From a user's seat: "press Styrbar, no music, press again, music plays" after every MA update — not acceptable for the user-visible Styrbar paths that motivated the fix.

The config-layer fallback in this PR doesn't have the race because it runs at lookup time (when the DLNA member fetches the stream URL), by which point `_handle_play_media` has already selected the protocol.

A heuristic-strengthened UGP-localized version (fall through to `preferred_output_protocol` config or the first-available `output_protocols[]`) might cover the cold-start case but starts to duplicate the protocol-selection logic that lives in `_handle_play_media`. Happy to switch to that approach if you'd prefer the smaller blast radius — let me know, or push back if there's a cleaner location for the fix that I'm not seeing.

marcelveldt:
The fix should be happening elsewehere.
Streams are always requested JIT for the active protocol player ID.

Issue is that this step is forgotten for the Universal Player Group (which is in general in very experimental state but that aside) - its missing a translation step to have the protocol-specific stream URL. Something that can only be resolved at the moment the player itself gets its play_media call (because only then the output protocol was decided).

UGP should send the play_media downstream using the playercontroller and not call the underlying players directly.
This way the translation will work properly.

If you want, I can create a fix later this week.

flomotlik:
Thanks. Want to confirm I've got the right spot before either of us touches it. The UGP already forwards via `mass.players._handle_play_media`, so protocol selection does run. What's missing is that `resolve_stream_url` short-circuits `MediaType.FLOW_STREAM` with `return media.uri`, so the `?player_id=` the UGP baked in (the *universal* player id) is never rewritten to the active protocol player id that's requesting the stream. Is the intended fix to do that `?player_id=` rewrite JIT in `resolve_stream_url`, and have `universal_group/player.py` stop pre-baking the id in `play_media`/`set_members`? Happy to take that on, or leave it to you if you'd rather. Just let me know.

flomotlik:
Follow-up: tried the `resolve_stream_url` translation locally on 2.8.7 (config-layer fallback removed) and it works. Whole change is in `resolve_stream_url`:

```python
if media.media_type in (MediaType.ANNOUNCEMENT, MediaType.FLOW_STREAM):
    # translate the baked-in player_id to the protocol player requesting the stream
    if media.media_type == MediaType.FLOW_STREAM and "/ugp/" in media.uri:
        parsed = urlparse(media.uri)
        query = parse_qs(parsed.query)
        if "player_id" in query:
            query["player_id"] = [player_id]
            return urlunparse(parsed._replace(query=urlencode(query, doseq=True)))
    return media.uri
```

The UGP stream endpoint now returns 200 + audio for the translated `?player_id=<protocol_id>` URL (was 500, `KeyError: sample_rates not found`). Speaker plays fine, stable across stop/replay.

If that's the approach you'd recommend, I'll update the PR to it.

marcelveldt:
superseded by https://github.com/music-assistant/server/pull/3951

marcelveldt:
@flomotlik I was working on some related stuff around this and I fixed your case in the same flow.
Hope that is OK. Anyways thanks for your PR and brining this to our attention!

flomotlik:
Yes thank you very much for the quick fix and the project in general!

PR #3887, opened 2026-05-13
Improve Listenbrainz multi-artist track scrobbles

Tommatheussen:
Currently we send artists as a single string to Listenbrainz, this works fine for tracks that only have 1 artist. 
However, when a track has multiple (feats or collabs), these are send as a single artists with `/` divider.
This result in Listenbrainz not being able to recognize these listens at all: 
image

Since we are already construct the `MediaItemPlaybackProgressReport` with the artist mbids (if they are populated), and we add those to the listen event of Listenbrainz, we can simply add them in the query to grab the track metadata.

This result in Listenbrainz somewhat linking the artists correctly:
image

The problem with that one is that the name is still incorrect, and it's assuming the full (concatenated) artists name belong to the first entry in the mbids list we've send. By sending the artists as a single string, separated with `,` instead, gives Listenbrainz a better chance at actually linking the track correctly.
image

**NOTE**: There still is an open issue with this, in the cases where the artist string is not in the correct order. As we don't have the concept of a 'main' performer, we don't always send the correct order of artists. 
Ideally we populate more data (like the track's mbid or album mbid), but I don't think these are fully implemented yet, at least not for online sources if I'm correct. The recording ID would be the main identifier to completely link the track correctly.
image

### DIFF
--- a/music_assistant/controllers/media/tracks.py
+++ b/music_assistant/controllers/media/tracks.py
@@ -78,7 +78,8 @@ class TracksController(MediaControllerBase[Track]):
                 'provider', 'library',
                     'name', artists.name,
                     'sort_name', artists.sort_name,
-                    'media_type', 'artist'
+                    'media_type', 'artist',
+                    'external_ids', json(artists.external_ids)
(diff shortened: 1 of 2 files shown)


MarvinSchenkel (approved):
Thanks @Tommatheussen , LGTM :)

PR #3891, opened 2026-05-15
Add Bose SoundTouch player provider

Odn0:

Adds a **Bose SoundTouch** player provider. SoundTouch speakers are detected on the network and owned by Music Assistant as native players, with control over power, volume, mute, transport, source selection and multiroom grouping. The physical preset buttons (1–6) can be mapped to Music Assistant content from each player's settings.

Following the Bose SoundTouch end of life, the speakers lost several connected-service features (presets via the product buttons, the Bose app, etc.). This provider restores that workflow and keeps the speakers fully usable in Music Assistant.

Reference: https://www.bose.com/soundtouch-end-of-life

**How it works**

- Discovery via mDNS (`_soundtouch._tcp.local.`) with a manual-IP fallback.
- Native control over the local SoundTouch API (HTTP `:8090` + the `gabbo` notification websocket `:8080`): power, volume, mute, transport, source selection and multiroom (zone) grouping.
- `PLAY_MEDIA` is intentionally not implemented — the SoundTouch API cannot play an arbitrary stream — so audio playback is delegated to a linked playback protocol (e.g. DLNA) through protocol linking. The provider therefore `depends_on` the DLNA provider.
- The physical preset buttons map to Music Assistant content per player (configured in the player settings); a press plays the configured item.
- Announcements are supported natively when a Bose developer app key is set in the provider settings — played as an overlay that ducks and resumes the current playback — otherwise they fall back to the linked playback protocol.

**Hardware**

- Bose SoundTouch 10 / 20 / 30

- n/a

## Types of changes

## Checklist


music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import logging
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+import websockets
```
OzGav: Why can't this use aiohttp's WebSocket support?

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+                        if isinstance(message, bytes):
+                            message = message.decode()
+                        preset_id = extract_preset_id(message)
+
+                        if preset_id is not None:
+                            await self._handle_preset(preset_id)
+
+            except asyncio.CancelledError:
+                raise
+            except Exception as err:
```
OzGav: Broad exception handling should be avoided and instead more narrow exceptions should be used for the errors you'd like to catch. This comment is applicable to all places this appears in the code.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+            label=f"{label}: {value if value not in (None, '') else 'unknown'}",
+            required=False,
+            category=CATEGORY_TARGET_PLAYER,
+            advanced=True,
+        )
+        for index, (label, value) in enumerate(details, start=1)
+    )
+
+
+async def _resolve_maybe_awaitable(value: Any) -> Any:
```
OzGav: What is this about?

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+            getattr(player, "display_name", None)
+            or getattr(player, "name", None)
+            or str(target_player)
+        )
+
+    async def loaded_in_mass(self) -> None:
+        """Start listening after Music Assistant loads the provider."""
+        target_player = self.config.get_value("target_player")
+        if not target_player:
+            raise RuntimeError("No Bose SoundTouch speaker has been configured")
```
OzGav: You need to be raising the MA specific errors not python generic

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+    """Extract identifier value from MA IdentifierType dict."""
+    for identifier_type, value in identifiers.items():
+        if getattr(identifier_type, "value", None) == key:
+            return value
+        if key in str(identifier_type).lower():
+            return value
+    return None
+
+
+def build_bose_player_info(player: Any) -> BosePlayerInfo | None:
```
OzGav: `Player` has properly typed properties (eg. `player.device_info`, `player.display_name`, `player.available`, `player.volume_level`) . Use them directly instead of getattr with fallbacks.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+
+if TYPE_CHECKING:
+    from music_assistant_models.config_entries import ConfigValueType, ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+
+LOGGER = logging.getLogger("music_assistant.Bose SoundTouch Favorites")
+PLUGIN_VERSION = "0.2.42"
```
OzGav: Why is this necessary?

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+SEARCH_RESULT_LIMIT = 25
+SEARCH_TIMEOUT = 10
+RECONNECT_DELAY = 10
+NO_SEARCH_RESULTS_VALUE = "__bose_soundtouch_favorites_no_results__"
+CATEGORY_GENERIC = "generic"
+CATEGORY_TARGET_PLAYER = "Target player"
+
+BOSE_SUBPROTOCOLS = [Subprotocol("gabbo")]
+
+MEDIA_TYPE_OPTIONS = [
```
OzGav: Why isn't the mediatype enum being used instead of this?

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,795 @@
...
+    from music_assistant.models import ProviderInstanceType
+
+LOGGER = logging.getLogger("music_assistant.Bose SoundTouch Favorites")
+PLUGIN_VERSION = "0.2.42"
+PRESET_IDS = range(1, 7)
+SEARCH_RESULT_LIMIT = 25
+SEARCH_TIMEOUT = 10
+RECONNECT_DELAY = 10
+NO_SEARCH_RESULTS_VALUE = "__bose_soundtouch_favorites_no_results__"
+CATEGORY_GENERIC = "generic"
```
OzGav: This is not usual for the MA codebase where categories are just defined in the config entry. Also there are a lot of categories used in this plugin which would seem to make the UI pretty busy?

OzGav (dismissed):

Odn0:
Thanks for the review.

I pushed updates addressing the WebSocket session, narrower exceptions, MA-specific setup errors, typed Player properties, MediaType values, config categories, and removed the hardcoded plugin version.

For the categories subject, I consolidated the per-button categories into a single Favorites category and added DIVIDER entries for each favorite. This keeps the UI clean and readable without the need to create six separate categories.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
Odn0: I switched to aiohttp's

music_assistant/providers/bose_soundtouch_favorites/__init__.py
Odn0: Sorry, I needed it to follow my tests on the Ubuntu VM + Docker setup and make sure the code was working correctly. I’ve deleted it now.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
Odn0: good point

marcelveldt:
Hi @Odn0, heads-up: #3938 just landed a refactor of the plugin sources architecture. The old `PluginProvider.get_source()` only allowed one source per provider instance — that limit is gone in the new model, so a SoundTouch Favorites provider can now naturally return one `AudioSource` per favorite.

The new contract on the provider:

- `async def get_audio_sources() -> list[AudioSource]` — return one AudioSource per favorite
- `async def get_stream_details(source_id, queue_id) -> StreamDetails` per session (raises `ResourceBusyError` on exclusive collisions if applicable)
- `async def get_audio_stream(streamdetails, seek_position)` for CUSTOM streams (matches the `MusicProvider` signature)
- `async def on_source_control(source_id, action, value)` for proxying play/pause/next/prev/seek/volume to the device
- Optional `async def on_source_selected(source_id, player_id, queue_id)` hook for reacting to selection

AudioSources are browseable under the new top-level "Live Inputs" node, are favoritable, and play through the standard `play_media` flow. Canonical example: [`vban_receiver`](https://github.com/music-assistant/server/blob/feat/audio-sources/music_assistant/providers/vban_receiver/provider.py); the demo template at `_demo_plugin_provider` shows the recommended shape.

Odn0:
> Hi @Odn0, heads-up: #3938 just landed a refactor of the plugin sources architecture. The old `PluginProvider.get_source()` only allowed one source per provider instance — that limit is gone in the new model, so a SoundTouch Favorites provider can now naturally return one `AudioSource` per favorite.
> 
> The new contract on the provider:
> 
> * `async def get_audio_sources() -> list[AudioSource]` — return one AudioSource per favorite
> * `async def get_stream_details(source_id, queue_id) -> StreamDetails` per session (raises `ResourceBusyError` on exclusive collisions if applicable)
> * `async def get_audio_stream(streamdetails, seek_position)` for CUSTOM streams (matches the `MusicProvider` signature)
> * `async def on_source_control(source_id, action, value)` for proxying play/pause/next/prev/seek/volume to the device
> * Optional `async def on_source_selected(source_id, player_id, queue_id)` hook for reacting to selection
> 
> AudioSources are browseable under the new top-level "Live Inputs" node, are favoritable, and play through the standard `play_media` flow. Canonical example: [`vban_receiver`](https://github.com/music-assistant/server/blob/feat/audio-sources/music_assistant/providers/vban_receiver/provider.py); the demo template at `_demo_plugin_provider` shows the recommended shape.

Thanks for the heads-up. I’m trying to make sure I understand the intended model before changing the plugin.

In this provider, a SoundTouch favorite button is currently only a physical trigger: when preset 1-6 is pressed, the plugin maps that preset to a configured MA media item and calls play_media for the configured target player. That target player is one Bose speaker; if the user wants to handle multiple Bose speakers, they would configure multiple instances of the plugin.

Do you expect each configured SoundTouch favorite to be exposed as its own AudioSource, even though it is not really an external/live audio source by itself, but rather a shortcut to an existing MA media item?

Or should the physical button handling remain separate, and only true live/external inputs be modeled as AudioSources?

Hedda:
@Odn0 not sure but believe that you might want to consider maybe renaming this plugin provider from "_**bose_soundtouch_favorites**_" to just "_**soundtouch_favorites**_" as that name chance removes that company name which would better follow the naming scheme /  naming convention of other providers in MA, or if @marcelveldt have better suggestion for the best name to use for this and other providers for long-term standardization of provider names(?).

Maybe its just my OCD, however does not look to be an a single standard naming scheme /  naming convention guideline for providers, like for example the Apple AirPlay Receiver is only called "AirPlay Receiver", and the Amazon Alexa Player Provider is called only Alexa, while the Google Cast Player Provider is only called "Chromecast" in the manifest but refered to as "Google Cast" in the end-user documentation, as well as there are other providers which uses the manufacturer name when they use it in the brand, such as AriaCast Receiver, Spotify Connect, LastFM Scrobbler, and Plex Connect.

IMHO naming conventions that establish a unified standardized rules for file names and code elements are vitally importnat because they establish a predictable framework, ensuring your team and external developers can instantly find and understand critical directories/files without relying on guesswork. And recently having and using a single naming scheme has also become important to make different AI agent tools work properly.

marcelveldt:
@Odn0 you can ignore my previous message. I was trying to be smart/lazy by tagging all pluginsoure related PR's that a change was upcoming to PluginSources but your feature is not about that.

Odn0:
No worries @marcelveldt, I understand. Thanks for clarifying.

@Hedda, regarding the provider name, I’m happy to follow the project’s preferred naming convention if there is one, and I can rename it if needed.

I originally included `bose` in `bose_soundtouch_favorites` to make the provider more explicit and easier to find for non-technical users. I initially built this plugin for my father, who uses a Home Assistant Green and has Bose speakers, but does not have a strong technical background.

If the maintainers prefer `soundtouch_favorites` or another name for consistency, no problem, I’ll update it.

OzGav:
`mass.music.search` has a stable, known signature — `search_query`, `media_types`, `limit`, `library_only` — and returns a typed `SearchResults `object. The entire `_search_media_items` function with its multiple kwargs/positional arg attempts, and `_iter_search_result_items` with its attribute probing, should be replaced with a single direct call. Similarly` _media_item_title` and `_media_item_value` should use the typed attributes on the SearchResults member models rather than getattr chains.

SearchResults has typed attributes,  `artists`, `albums`, `genres`, `tracks`, `playlists`, `radio`, `audiobooks`, `podcasts `. So `_media_item_title, _media_item_value`, and `_iter_search_result_items` should all be rewritten against the actual typed models (Artist, Album, Track, Radio etc.) which all have a name attribute and uri from their base class. Additionally, `MEDIA_TYPE_OPTIONS` includes types like FOLDER, ANNOUNCEMENT, FLOW_STREAM that have no corresponding field on SearchResults and can never return search results so these should either be removed or handled explicitly

music_assistant/providers/bose_soundtouch_favorites/__init__.py
OzGav: You should not resolve this until you have actually fixed the problem.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,748 @@
...
+
+if TYPE_CHECKING:
+    from music_assistant_models.config_entries import ConfigValueType, ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+    from music_assistant.models.player import Player
+
+LOGGER = logging.getLogger("music_assistant.Bose SoundTouch Favorites")
```
OzGav: Use self.logger from the base Provider class instead

Odn0:
> `mass.music.search` has a stable, known signature — `search_query`, `media_types`, `limit`, `library_only` — and returns a typed `SearchResults `object. The entire `_search_media_items` function with its multiple kwargs/positional arg attempts, and `_iter_search_result_items` with its attribute probing, should be replaced with a single direct call. Similarly` _media_item_title` and `_media_item_value` should use the typed attributes on the SearchResults member models rather than getattr chains.
> 
> SearchResults has typed attributes, `artists`, `albums`, `genres`, `tracks`, `playlists`, `radio`, `audiobooks`, `podcasts `. So `_media_item_title, _media_item_value`, and `_iter_search_result_items` should all be rewritten against the actual typed models (Artist, Album, Track, Radio etc.) which all have a name attribute and uri from their base class. Additionally, `MEDIA_TYPE_OPTIONS` includes types like FOLDER, ANNOUNCEMENT, FLOW_STREAM that have no corresponding field on SearchResults and can never return search results so these should either be removed or handled explicitly

Thanks for the review. I updated the provider to use the stable `mass.music.search` signature directly and to work against the typed `SearchResults` fields instead of probing runtime shapes.

I also removed unsupported media types from the search options, switched the provider logging to `self.logger`, and cleaned up a few overly broad helpers/casts around config values and media result handling.

`pre-commit run --all-files` passes locally.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,664 @@
...
+
+    from music_assistant.mass import MusicAssistant
+    from music_assistant.models import ProviderInstanceType
+    from music_assistant.models.player import Player
+
+PRESET_IDS = range(1, 7)
+SEARCH_RESULT_LIMIT = 25
+SEARCH_TIMEOUT = 10
+RECONNECT_DELAY = 10
+NO_SEARCH_RESULTS_VALUE = "__bose_soundtouch_favorites_no_results__"
```
OzGav: I am pretty sure when search returns no results, `build_media_options` will return an empty list rather than a sentinel value. The `get_config_entries` caller already handles an empty list correctly. This lets you remove NO_SEARCH_RESULTS_VALUE and simplify `_is_real_media_selection` and `show_media_selection`.

OzGav:
Just one last comment from me. One of the others will have a look as well soon.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
Odn0: Removed the no-results sentinel and now rely on the empty options list from `build_media_options`. This also removes `_is_real_media_selection` and simplifies `show_media_selection`. Thanks!

Odn0:
Hi all, just checking in on this PR.

I believe I addressed the last review comments, including the no-results sentinel cleanup, and `lint` / `test` are passing. The remaining failing `Dependency Security Check / report` appears to be waiting for maintainer dependency approval for `defusedxml`.

Is there anything else expected from my side, or should I just wait for another maintainer review / dependency approval?

Thanks!

OzGav:
We are preparing for the release of 2.9 so you can expect this will be looked at again in two weeks. Sorry for the delay.

Odn0:
No problem at all, thanks for the update.

I’ll wait for the next review after the 2.9 release prep. Good luck with the release!

Odn0:
Hi,

I was asked this weekend about adding a new feature to this plugin. Before implementing it in this PR, I wanted to check whether this is something you would like to support.

The idea is to allow the plugin to map favorite buttons from multiple speakers to the same set of favorites. On the UX/UI side, this would likely mean adding a checkbox list where users can select which speakers should be listened to for button presses.

Would you be okay with me adding this to the current PR?

OzGav:
> The idea is to allow the plugin to map favorite buttons from multiple speakers to the same set of favorites. On the UX/UI side, this would likely mean adding a checkbox list where users can select which speakers should be listened to for button presses.

Or maybe can you have a dropdown for each button and the user then selects from a list of speakers. Just like we do for say the creation of a group?

Odn0:
I pushed an update for this.

I went with a multi-select dropdown for the target speakers, so the user can select which Bose SoundTouch speakers should be listened to for favorite button presses. This seemed like a good fit here, since the same favorite mappings are shared across the selected speakers.

OzGav:
Please rebase to resolve the dependency security issues.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,690 @@
...
+    """Listen to Bose SoundTouch favorite button events."""
+
+    _listener_tasks: dict[str, asyncio.Task[None]] | None = None
+    _stop_event: asyncio.Event | None = None
+    _bose_player: BosePlayerInfo | None = None
+
+    @property
+    def instance_name_postfix(self) -> str | None:
+        """Return the target player name to identify multi-instance configs."""
+        if self._bose_player:
```
OzGav: I don't think the use of this attribute is meaningful now. You have the code below which will set the instance name from the first configured player anyway, so _bose_player can be removed entirely.

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,690 @@
...
+                bose_player.model,
+                bose_player.bose_uuid,
+                bose_player.mac_address,
+            )
+
+        self._bose_player = bose_players[0]
+
+        self._stop_event = asyncio.Event()
+        for bose_player in bose_players:
+            self._listener_tasks[bose_player.player_id] = asyncio.create_task(
```
OzGav: Validation and task creation should be separated into two loops because if a later player fails validation, any listener tasks already created in this loop will be left running with nothing to clean them up

music_assistant/providers/bose_soundtouch_favorites/__init__.py
```diff
@@ -0,0 +1,690 @@
...
+        )
+        if not target_player_ids:
+            return None
+
+        target_player_id = target_player_ids[0]
+        player = self.mass.players.get_player(target_player_id)
+        if player is None:
+            return target_player_id
+
+        return player.display_name
```
OzGav: With multiple players now supported, returning only the first player's name as the instance postfix may be misleading. Can you clarify the intended use case? Is this provider meant to be configured once for all speakers, or as one instance per speaker?

Odn0:
Thanks @OzGav, I agree with your points.

I updated the PR accordingly:
- removed the single `_bose_player` state since it no longer makes sense with multiple targets
- split target validation from listener task creation so no listener is started before all configured targets are valid
- updated the instance postfix to represent the configured target set instead of only the first player
- simplified the target detail helper now that the provider is multi-target only

Thanks again for the review!

OzGav:
Yeah looks good. Looks there is one mypy issue and Marcel has put the framework together to have all strings translatable so please have a look at the lint errors and then refer to some of the other providers for how to setup the keys and strings.

Odn0:
I fixed the mypy issue and migrated the provider’s configuration labels, descriptions, categories, and options to the new translation framework using `strings.json`, translation keys, and parameters for the dynamic preset entries.

The PR has been updated and the lint and mypy checks now pass locally. Thanks @OzGav for pointing me in the right direction!

marcelveldt:
Sorry for chiming in so late but I'm not entirely sure about this.
Where are the players coming from ? Are they present as generic protocol players such as DLNA ?

If the API provides some more options to control them, such as volume etc. why not add this as a Player provider ?
You can simply omit the play_media playerfeature if the api doesn't provide this.
The benefit is that these players will then be seen as native BOSE SoundTouch players in MA and the protocol linking will automagically link the supported playback protocols.

This makes the provider responsible for detecting the soundtouch players on the network and "own" them.
You can then still offer the presets as option but then the config will show up in the player settings of each player (like squeezelite has for jivelite based players).

marcelveldt:
Hi @Odn0 — first, thank you for the work here and for patiently working through all the review rounds. 🙏

I really wanted to get this one over the finish line, so rather than leave it in limbo I reshaped it into the form we can accept and merge: a native **Bose SoundTouch player provider**. To be clear — this stays **your** contribution: you're kept as code owner, your commits are preserved, and the preset-button feature you built is fully intact. I've pushed the changes on top of your branch.

What changed and why:
- The speakers are now **detected on the network** (mDNS + manual IP) and owned by Music Assistant as native Bose SoundTouch players, instead of layering on top of whatever provider happened to discover them.
- They expose native control over the local SoundTouch API: **power, volume, mute, transport, source selection and multiroom (zone) grouping**.
- `PLAY_MEDIA` is deliberately **omitted** — SoundTouch has no usable API to play an arbitrary stream — so actual audio is delegated to a **linked playback protocol (DLNA)** through our protocol-linking mechanism (the provider `depends_on` dlna). That's why the speakers "just play" without us reinventing streaming.
- Your **preset buttons** live on as a per-player setting: each speaker's settings has the same search-and-assign flow you built, and a physical press plays the configured item.
- **Announcements** are supported natively when you set a Bose developer app key in the provider settings: they're sent as an overlay that ducks the current music and resumes it afterwards (like Sonos). Without a key, announcements simply fall back through the linked protocol. (A shared MA-wide key isn't viable — the free tier is rate-limited per key — so it's opt-in per install.)

Because I don't have SoundTouch hardware here, the parts that need your eyes on real devices are: DLNA-linked playback, the multiroom zone grouping, source selection, now-playing metadata, native announcements, and the preset-press detection (I kept your tested detection logic). You know these devices far better than I do — please give it a spin and tweak anything that needs it.

Thanks again for sticking with it! 🙌

marcelveldt:
Marked as draft so you can test the changes locally with the actual hardware.
If it works fine for you, we're ready to merge. Let me know if you still find anything not working

Hedda:
> Sorry for chiming in so late but I'm not entirely sure about this. Where are the players coming from ? Are they present as generic protocol players such as DLNA ?
> 
> If the API provides some more options to control them, such as volume etc. why not add this as a Player provider ?

> I really wanted to get this one over the finish line, so rather than leave it in limbo I reshaped it into the form we can accept and merge: a native **Bose SoundTouch player provider**.

@marcelveldt To clearify, this pull request as-is does by definition today not (yet) provide all the full player provider capability which end-users would expect from a "_**SoundTouch Player Provider**_" as per the linked feature request, or does it meet what you feel could be set as a minimum requirements for it to be called a "player provider" now so can initiate playing of media on a such device? Please refer to this feature request for that here -> https://github.com/orgs/music-assistant/discussions/4766

* https://github.com/orgs/music-assistant/discussions/4766

That is, I would expect that if and when publicly introducing a native built-in "SoundTouch Player Provider" inside Music Assistent server codebase then it should be a relatively fully featured player provider component supporting offering not only full local media control commands for SoundTouch client devices now that Bose has openened its API documentation for, but also offer streaming audio playback, or is that too much to ask when something is called a "player provider"? -> https://github.com/orgs/music-assistant/discussions/4766

PS: As mentioned in that feature request discussion, don't have any SoundTouch hardware myself any longer.

Odn0:
Hi @marcelveldt,

Thanks a lot for reshaping the plugin. I’m really happy with this refactor and agree that a native Bose SoundTouch player provider is a much better fit for Music Assistant.

I don’t personally own SoundTouch hardware. I originally started this for my father, who does. I’m glad this may get close to production soon, as users have been blocked since May 6, and Music Assistant feels like a great local, open-source solution for this.

I should be able to test on real hardware around mid-July, but if someone else can test it before then, please feel free to do so.

Thanks again for pushing this forward and keeping the preset-button feature intact.

marcelveldt:
In that case I propose I just merge this so people on the nightly/beta channel can test this already. If there may still be an issue here or there, it can be reported and either you or one of the maintainers of MA will have a look.
In the end this is just one extra provider, worst that case that can happen is that something does not yet work properly on this provider, which is already better than no Bose player provider at all so to say.

marcelveldt (approved):

OzGav:
@Odn0 I had Claude create a docs page. If you could let me know if any changes need to be made https://github.com/music-assistant/music-assistant.io/pull/749

reinhard33:
Hello @Odn0 @Hedda @marcelveldt  thirst of all thanks for driving and developing this for the SoundTouch User Community. Maybe we get here a first step for a native player and important the Preset Button Feature. Many user love this simply press a button in the kitchen or bathroom and hear music (so my woman).

Hello @OzGav  and Good Morning,
i have activated the beta channel in my VM, currently i can see 2.10.0b2, maybe i can see later today or tomorrow 2.10.0b3. 
I hope to find then this in new Player Provider as SoundTouch or Bose SoundTouch ? Is this right?
I can test what i understand with 4*ST20 and one ST10 from maybe October 2015 from the first generation of this devices. I have not understand until now "Announcements". I'm a end user and newcomer, but i have the ST Hardware available so that i can start to test something.
Greetings from Reinhard

OzGav:
The new beta will come out later this week I expect. We look forward to you testing it thoroughly!

reinhard33:
Hello @marcelveldt ,
the beta3 was not coming until now, so i have activated the nightly Music assistant. 
In this version i can found a player Provider Bose Sound Touch, but i can not find any possibility to configure the Sound Touch favorite.
Maybe i am in the wrong Player ?
I have many Player with similar names and different protocol.

The Name of the speaker is the name of the room for example -- Arbeiterkammer

now i see the following Player:
Arbeiterkammer              with DLNA
Arbeiterkammer+             with AirPlay / airupnp
Arbeiterkammer+(AirPlay)    with AirPlay / airupnp

My Airplay container on the nas i have deactivated, otherwise i see 4 Player.

In the configuration of all three speaker is see no possibility to configure a Favorite?
How can i recognize the new --> native Bose Sound Touch Player?

Maybe this nightly Music Assistant has not the final functionality or i am not able to find the functionality.
Please help
Greetings from Reinhard

Hedda:
@reinhard33 since this pull request is now closed it would probably be better to move the discussion about changes, enhancements, and improvements to the this thread instead for continued discussion and follow-up questions or other feedback/input -> https://github.com/orgs/music-assistant/discussions/4766

* https://github.com/orgs/music-assistant/discussions/4766

PR #3892, opened 2026-05-15
Add Acoustid audio analysis provider

OzGav:
This audio analysis provider fingerprints audio with Chromaprint and looks the result up against AcoustID when a library track has no MusicBrainz recording ID attached. Local files are always analysed; streaming-provider tracks (Spotify, Tidal, Qobuz, …) are analysed when they exist in the library and the "Analyse tracks from streaming providers" option is enabled (default on). Podcasts and audiobooks are never analysed.

The lookup returns a recording MBID, and from there a MusicBrainz query reliably yields the ISRC and the artist MBIDs. Getting the release-group is trickier because there can be many candidates per track, so a cross-track consensus vote is run across the album, biased by album-title matches, and the release-group that the most tracks agree on wins. Even with that, sometimes the consensus is empty or wrong, so the release-group ID is best-effort. The consensus winner is further constrained to release-groups credited to the track's artist (defends against AcoustID fingerprint collisions that surface a different artist's recording), and falls back to a direct MusicBrainz artist:"X" AND release:"Y" search when no AcoustID-returned release-group fits — covers the case where MB has multiple recording entities for the same audio and the user's release uses a different one than AcoustID matched.

For these reasons, the writeback is split. The recording MBID, AcoustID and ISRC are always written to the database, and to the file tags if the user opts in. Artist MBIDs are written to the file tag only when the user opts in, and the filesystem provider's normal tag parse picks them up on the next sync. That keeps artist name-matching out of audio analysis where it doesn't belong. The release-group MBID only goes to the database and never to the file, since the consensus is best-effort and we don't want to pollute user files with a potentially wrong ID. If consensus is wrong, a REFRESH ITEM clears it. For streaming-provider tracks there's no local file to write to, so all writeback for those is DB-only regardless of the tag-write option.

Ideally a user will have tagged their files comprehensively, but if not, and whether those tracks live on local files or a streaming provider, the identifiers this provider gathers will let the metadata providers supply the rest and enable cross-provider matching. This improves the experience for those with poorly tagged tracks or streaming providers that supply minimal metadata. Many of MA's features will now "just work" for them. 

Note: from AcoustID website: Let us know — If you are deploying an application that you expect to generate significant traffic to this service, please [let us know](mailto:info@acoustid.org) in advance.


music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -0,0 +1,914 @@
...
+        Accept or decline an analysis session for the given track.
+
+        :param session_id: Session ID assigned by the AudioAnalysisController.
+        :param streamdetails: Stream details for the track being analysed.
+        :param audio_format: PCM format of the incoming audio stream.
+        """
+        if streamdetails.media_type != MediaType.TRACK:
+            return False
+        if streamdetails.stream_type != StreamType.LOCAL_FILE:
+            return False
```
MarvinSchenkel, line 139: Why not include streaming providers that are known to have terrible metadata, such as YT Music?

music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -0,0 +1,914 @@
...
+                    "acoustid: session %s pcm conversion failed (sample_width=%d): %s",
+                    session_id,
+                    data.sample_width,
+                    err,
+                )
+                data.error = f"pcm conversion failed: {err}"
+                return
+
+        try:
+            data.fingerprinter.feed(pcm_chunk)
```
MarvinSchenkel, line 210: Do you know if the downstream library keeps these chunks in memory? Or are these pushed down to the binary that extracts the fingerprint? We need to be sure we do not keep 120s of PCM audio in memory, as this will cause a lot of memory pressure during the analysis. This is very important to verify

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -320,6 +327,148 @@ async def set_track_loudness(
             media_type=media_type,
         )
 
+    async def set_track_identifiers(
```
MarvinSchenkel: I don't really like these purpose specific functions in the AA controller. Gotta think about this a little bit to see how we can do this properly. This feels more like it belongs in a metadata controller. So if the metadata controller exposes these functions, we can call them in the post_analysis

music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -0,0 +1,914 @@
...
+        try:
+            fingerprint = bytes(fingerprint_raw).decode("ascii")
+        except UnicodeDecodeError:
+            self.logger.debug("acoustid: fingerprint not ASCII")
+            return None
+        if not fingerprint:
+            self.logger.debug("acoustid: empty fingerprint")
+            return None
+
+        api_key = self.config.get_value(CONF_API_KEY)
```
MarvinSchenkel, line 264: Let's verify this pre requisite in the start_analysis so we can return early when absent

music_assistant/providers/acoustid_lookup/provider.py
OzGav: I did think of this initially but then had these thoughts:

Bandwidth cost: analysing a streaming track pulls full audio from the user's streaming account. For a large library, that's a lot of streaming-provider traffic the user didn't ask for and may have quota implications on their account. I guess you could do the playback only check instead of overnight?

Most streaming providers already expose ok metadata so the upside is maybe just for YT Music?

TOS uncertainty: re-streaming audio from a paid provider for fingerprinting I was unsure about?

music_assistant/providers/acoustid_lookup/provider.py
OzGav: This is outside my area of expertise. Claude says:

> Chromaprint's C library is streaming-designed — it computes a compressed feature representation per frame and doesn't retain raw PCM. pyacoustid wraps chromaprint_feed() via ctypes without buffering on the Python side. Memory per active session should be ~1 MB (one chunk in flight + chromaprint's internal FFT buffer + accumulated hashes)

I am not sure how I could test to confirm though?

music_assistant/controllers/streams/audio_analysis.py
OzGav: Agreed — `set_track_loudness` lives there because loudness is an AA output, but identifiers aren't. I will move `set_track_identifiers` and `set_album_release_group` to `controllers/metadata.py` (or the track/album controllers directly as you think best), and generalise `get_acoustid_extra_data_for_album_tracks` to take an aa-provider-domain parameter so it stays AA-generic.

music_assistant/providers/acoustid_lookup/provider.py
OzGav: OK will do

music_assistant/controllers/streams/audio_analysis.py
OzGav: `set_track_identifiers` and `set_album_release_group` moved to metadata controller

music_assistant/providers/acoustid_lookup/provider.py
OzGav: API-key prerequisite check moved into `start_analysis`

music_assistant/providers/acoustid_lookup/provider.py
MarvinSchenkel: We don't actively auto re-stream. We only analyze when the user is already streaming a track from a streaming service. So AA providers just hook into whatever is playing in MA.

The only exception here is the background task that runs at midnight, but this scans `filesystem` based providers only.

music_assistant/providers/acoustid_lookup/provider.py
MarvinSchenkel: This sounds promising!
Easiest way to double check is to run this in a DEV addon and keep the 'hardware' tab of HAOS open and monitor memory usage during the analysis. Make sure to only enable this provider to avoid false positives.

music_assistant/providers/acoustid_lookup/provider.py
OzGav: Yep sounds good. I am looking at it!

music_assistant/providers/acoustid_lookup/provider.py
OzGav: I analysed seven files of various formats (MP3, MP4, FLAC) and saw no movement in memory

image

music_assistant/providers/acoustid_lookup/provider.py
OzGav: OK now supports analysis of streaming provider tracks. Identifying the correct album was tricky but I believe the solution I have is workable.

music_assistant/providers/acoustid_lookup/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "audio_analysis",
+  "domain": "acoustid_lookup",
```
OzGav, line 3:
```suggestion
  "domain": "acoustid_lookup",
  "stage": "beta",
```

music_assistant/controllers/metadata.py
```diff
@@ -373,6 +373,98 @@ async def update_metadata(
                 )
         return item
 
+    async def set_track_identifiers(
```
marcelveldt: why are these methods on the metadata controller and not the music controller ?

music_assistant/helpers/tags.py
```diff
@@ -1467,3 +1470,223 @@ def _write_replaygain_track_gain_sync(path: str, track_gain_db: float) -> bool:
...
+    """
+    Write the AcoustID identifier tag to an audio file.
+
+    Returns True when the tag was successfully written and saved, False
+    when the file format is unsupported or the file cannot be written to.
+
+    :param path: Absolute path to the audio file.
+    :param acoustid: AcoustID UUID string to persist.
+    """
+    return await asyncio.to_thread(_write_acoustid_tag_sync, path, acoustid)
```
marcelveldt: why not inline the sync method ?

music_assistant/helpers/tags.py
```diff
@@ -1467,3 +1470,223 @@ def _write_replaygain_track_gain_sync(path: str, track_gain_db: float) -> bool:
...
+    Returns True when the tag was successfully written and saved, False
+    when the file format is unsupported or the file cannot be written to.
+
+    :param path: Absolute path to the audio file.
+    :param recording_id: MusicBrainz recording UUID to persist.
+    """
+    return await asyncio.to_thread(_write_musicbrainz_recording_id_tag_sync, path, recording_id)
+
+
+async def write_musicbrainz_artist_id_tag(path: str, artist_ids: list[str]) -> bool:
```
marcelveldt: If we're going to do 3 writes to a file to add these tags, we should consider a combined method to set multiple at once

music_assistant/providers/acoustid_lookup/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+    return (
+        ConfigEntry(
+            key=CONF_API_KEY,
+            type=ConfigEntryType.SECURE_STRING,
+            label="AcoustID API key",
+            description=(
+                "Free API key from https://acoustid.org/api-key. "
+                "AcoustID requests that high-traffic deployments notify the project "
+                "in advance — see https://acoustid.org/webservice for details."
+            ),
```
marcelveldt, line 54: why is this API key configurable ? why not request a key for MA ?

music_assistant/providers/acoustid_lookup/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+            required=True,
+            default_value=None,
+        ),
+        ConfigEntry(
```
marcelveldt, line 58: can we mark this ad advance if the default covers most cases ?

music_assistant/providers/acoustid_lookup/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+            ),
+            default_value=True,
+            required=False,
+        ),
+        ConfigEntry(
+            key=CONF_WRITE_TAGS_BACK,
+            type=ConfigEntryType.BOOLEAN,
+            label="Write AcoustID/MusicBrainz tags back to files",
+            description=(
+                "When the audio-analysis scan identifies a track via AcoustID, also "
```
marcelveldt, line 89: should the description make clear that this can only work for local files providers and when we have write access to the path ?

OzGav: A bit further down the write access requirement was mentioned but I have expanded it to explicitly state the Filesystem requirement as well.

music_assistant/providers/acoustid_lookup/__init__.py
OzGav: Yes good idea. Done.

music_assistant/providers/acoustid_lookup/__init__.py
OzGav: I am writing to them shortly so will advise when I hear back what their preference is

music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -0,0 +1,1203 @@
...
+        streamdetails: StreamDetails,
+        audio_format: AudioFormat,
+    ) -> bool:
+        """
+        Accept or decline an analysis session for the given track.
+
+        :param session_id: Session ID assigned by the AudioAnalysisController.
+        :param streamdetails: Stream details for the track being analysed.
+        :param audio_format: PCM format of the incoming audio stream.
+        """
```
marcelveldt: why is the public handler copied here ? why not just pass in the library track

music_assistant/helpers/tags.py
OzGav: The async wrapper means anyone using the tag-writing functions in helpers/tags.py can call them safely without thinking about it. The underlying _write_identifier_tags_sync functions call mutagen.File(...).save() to write to disk. That's a blocking operation.

I think that inlining would be ok for the current single call site in the AcoustID provider but the risk would be the next person to add a tag-write call method, who would have to (a) know that the _sync function exists, (b) know it's blocking, and (c) remember to wrap it. Nothing in the API hints at any of that.

I based this on the same wrapper pattern used by write_replaygain_track_gain in the same file.

music_assistant/helpers/tags.py
OzGav: Yes. Thats what happens when I incrementally add stuff as I think of it! This is now one function

music_assistant/controllers/metadata.py
OzGav: Yeah it was probably late when I moved them out of the audio analysis controller to the wrong place. `set_identifiers` is now in the tracks controller and `set_release_group` is now in the albums controller.

music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -0,0 +1,1203 @@
...
+                streamdetails.item_id,
+                err,
+            )
+            track = None
+
+        fingerprinter = self._create_fingerprinter(audio_format.sample_rate, audio_format.channels)
+        if fingerprinter is None:
+            return False
+
+        bit_depth = audio_format.bit_depth or 16
```
marcelveldt:
```suggestion
        bit_depth = audio_format.bit_depth
        if bit_depth is None:
            return False
```

music_assistant/providers/acoustid_lookup/provider.py
OzGav: The public `start_analysis` override was a leftover from when I changed how this provider decides to skip already analysed tracks. I wanted the MBID check to run before the base class's version gate, but this provider doesn't store an analysis version anyway, so the override was guarding nothing. Moved everything back into `_start_analysis` so there's a single library-track fetch.

music_assistant/providers/acoustid_lookup/provider.py
OzGav: Actually mypy complained as `bit_depth` can't be `None `as it is typed as just `int`. So adjusted slightly to check `if not bit_depth`

MarvinSchenkel:
Let's give them until after the weekend to reply on our query about a shared API key. Otherwise this looks good to merge with a per-user API key

MarvinSchenkel (approved):

music_assistant/providers/acoustid_lookup/__init__.py
OzGav: No response after a week. Plan is to go with individual API keys for now and see what happens.

PR #3893, opened 2026-05-15
feat(audible): resume playback position from Audible Whispersync

scootaash:
## Summary

Implements full bidirectional progress sync between Music Assistant and the
Audible app via Whispersync. My first PR so feedback appreciated

**Before:** MA always resumed from 0:00 after listening on the Audible phone app
or any other Whispersync-connected device. Write-back (MA → Audible) existed but
was unreliable due to ACR cache lifecycle and a feedback loop that fired redundant
PUT calls every 30 s.

**After:** Position is shared correctly in both directions.

### Read direction (Audible app → MA) — was completely missing

- `sync_progress_from_audible()` batch-fetches last-heard positions via
  `GET /1.0/annotations/lastpositions` for all library audiobooks and writes
  them into MA's internal playlog via `mark_item_played()`.
- Fires once at cold-start and then hourly via `TaskSchedule.hourly(every=1)`.
- Registered/unregistered in `handle_async_init` / `unload`.
- `get_resume_position()` does a live single-ASIN fetch immediately before
  playback so MA always starts from the freshest position; raises
  `NotImplementedError` on failure so MA falls back to its own playlog.

### Write direction (MA → Audible) — reliability fixes

- Timestamp comparison in `_sync_progress_chunk`: parses Audible's
  `last_updated` and compares against MA's playlog row before writing —
  skips the write when MA is already as current or newer.
- Per-ASIN suppression set (`_sync_suppressed_asins`): prevents the
  `mark_item_played` → `on_played` → `set_last_position` feedback loop
  without a global flag or `asyncio.sleep(0)` race.
- Audible API chunking: enforces both the 500-char query-string limit
  (`_MAX_ASINS_PARAM_LEN = 480`) and the 25-ASIN per-request limit
  (`_MAX_ASINS_PER_REQUEST = 25`).
- `get_audible_resume_position()` parses Audible's actual `last_updated`
  timestamp so MA's playlog tiebreaker picks the fresher source correctly.
  Raises `NotImplementedError` on transport failure for graceful fallback.

### Notes

- Pattern mirrors the Audiobookshelf provider's `_update_playlog_book` /
  `on_played` approach.
- `fully_played` is always `False` in sync — the `lastpositions` endpoint
  does not expose finished-state.
- `_get_user_for_provider` is not called; `userid=None` lets `mark_item_played`
  resolve the user via its own existing fallback.

## Test plan [all completed by scootaash]

- [ ] Listen on Audible phone app, restart MA — verify resume position is correct
- [ ] Play in MA, open Audible app — verify position updated
- [ ] Library of 200+ books — verify sync completes without 400 errors
- [ ] Provider reload — verify no duplicate scheduled tasks in logs


OzGav:
@ztripez

OzGav:
Why are there changes to the music controller in this PR? https://github.com/music-assistant/.github/blob/main/AI_POLICY.md

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
...
+
+        Raises NotImplementedError when data is unavailable or on transport failure so
+        MA's music controller falls back to its own internal playlog instead.
+        """
+        if not asin or asin == "error":
+            raise NotImplementedError
+
+        try:
+            response = await self._call_api("annotations/lastpositions", asins=asin)
+        except Exception as exc:
```
OzGav: Broad exception handling should be avoided and instead more narrow exceptions should be used for the errors you'd like to catch. This occurs in various places in this PR

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
...
+        position_ms = int(last_position.get("position_ms", 0))
+        if position_ms == 0:
+            # No meaningful position on Audible's side — let MA use its own playlog.
+            raise NotImplementedError
+
+        # Parse the Audible-supplied timestamp so MA's playlog tiebreaker works
+        # correctly. The field is "last_updated" in the Audible annotation payload;
+        # additional names are tried as a fallback against future API variation.
+        timestamp: datetime | None = None
+        for field in ("last_updated", "reported_time", "last_updated_time", "timestamp"):
```
OzGav: This is duplicated in sync_progress_chunk. Please extract to a helper

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
...
+                )
+                if mass_audiobook is None:
+                    # Book not yet in MA DB (library sync still running); skip
+                    continue
+
+                # Skip if MA's existing record is as recent as Audible's position.
+                # mark_item_played writes with media_item.provider which resolves to
+                # "library" for library items — use mass_audiobook.provider to match.
+                if audible_ts is not None:
+                    existing = await self.mass.music.database.get_row(
```
OzGav: A provider must not access the database directly `mass.music.database` is the music controller's internal concern. The controller already handles playlog timestamp comparison in `get_resume_position`

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
...
+        mark_item_played. This lets MA resume at the correct position even
+        when the previous session happened in the Audible phone app or another
+        Whispersync-connected device.
+
+        Called as a background task from handle_async_init so it does not
+        block provider startup. Books not yet in MA's DB (library sync still
+        in progress) are skipped silently.
+        """
+        # Prevent concurrent runs (cold-start task + first scheduled fire can overlap).
+        if self._syncing_from_audible:
```
OzGav: Consider asyncio.Lock instead

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
...
+            self.logger.debug("Audible progress sync already in progress, skipping")
+            return
+        try:
+            self._syncing_from_audible = True
+            # Pass userid=None so mark_item_played uses its own user-resolution
+            # fallback (via provider_mappings on the media item). This avoids
+            # calling a private API and is equivalent for all supported setups.
+            userid: str | None = None
+            asins: list[str] = []
+            async for item in self._fetch_library_items(
```
OzGav: This looks like you are fetching every book just to get a list of ASINs? You can get this from `mass.music.audiobooks.library_items()` filtered by this provider instance

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +538,227 @@ async def set_last_position(
         except Exception as exc:
             self.logger.error(f"Unexpected error reporting position for ASIN {asin}: {exc}")
 
+    async def get_audible_resume_position(
```
OzGav: Why was the music controller's `get_last_postion` not extended or replaced instead of adding a new method?

music_assistant/providers/audible/__init__.py
```diff
@@ -575,14 +590,33 @@ async def on_played(
 
         media_item is the full media item details of the played/playing track.
         """
+        if prov_item_id in self.helper._sync_suppressed_asins:
+            self.helper._sync_suppressed_asins.discard(prov_item_id)
```
OzGav: You are accessing a private attribute here?

music_assistant/providers/audible/__init__.py
```diff
@@ -575,14 +590,33 @@ async def on_played(
...
+            return
         await self.helper.set_last_position(prov_item_id, position, media_type)
 
+    async def get_resume_position(
+        self, prov_item_id: str, media_type: MediaType
+    ) -> tuple[bool, int, datetime | None]:
+        """Return resume position from Audible for the given item.
```
OzGav:
```suggestion
        """
        Return resume position from Audible for the given item.
```
For standardisation please ensure all multi line docstrings start on their own line

music_assistant/providers/audible/__init__.py
```diff
@@ -575,14 +590,33 @@ async def on_played(
...
+
+        Called by MA just before queuing an audiobook for playback.
+        MA compares the returned timestamp against its internal playlog and
+        uses whichever source is more recent.  Raises NotImplementedError on
+        transport failure so MA falls back to its own playlog rather than
+        presenting a stale or missing position as authoritative.
```
OzGav: Dosctrings may not contain implementation detail

OzGav (changes requested):

OzGav:
Also lint is failing. Marking as draft so we can keep track of what needs attention. Please mark as ready for review again when you want another look.

scootaash:
Many thanks for the detailed comments. Will take a look. Cheers

music_assistant/providers/audible/audible_helper.py
MaZZly: @OzGav I think this could be automated with ruff/lint checks? :) 

Either one of these? :thinking: 
https://docs.astral.sh/ruff/rules/bare-except/
https://docs.astral.sh/ruff/rules/blind-except/

OzGav: Thanks I'll pass the suggestion onto the lint gurus!

MarvinSchenkel:

moelsaied88:
Is this PR still a work in progress? Very excited for this feature, considering taking a stab if OP is no longer interested.

scootaash:
Hi there. Thanks for the positive comment! Yes sorry been a busy few weeks! Will review in light of the earlier comments over the weekend.

scootaash:
Thanks again for the detailed review. I've reworked this significantly:

- I dropped the background full-library sync entirely. That was the source of the
  direct-DB-access, full-library-fetch, private-attribute and locking concerns.
  The provider now only implements get_resume_position — a single fetch
  before playback — and lets the controller have the playlog and timestamp
  comparison. Surfacing externally-progressed books in the in-progress shelf is
  a core concern rather than a provider one, so I'll raise that separately.
- No music controller changes should remain in this PR.
- I narrowed the exception handling to audible.exceptions.AudibleError plus the
  relevant parse errors.
- De-duplicated the timestamp parsing into _parse_audible_timestamp /
  _fetch_last_position; get_last_postion now reuses them instead of a parallel
  implementation.
- Fixed the docstrings (own-line summaries, no implementation detail) and lint
  now passes.

scootaash:
CI appears to be waiting for workflow approval — could a maintainer approve the workflows to run?

music_assistant/providers/audible/audible_helper.py
```diff
@@ -453,39 +454,68 @@ async def _fetch_chapters(self, asin: str) -> list[dict[str, Any]]:
 
         return chapters_data
 
-    async def get_last_postion(self, asin: str) -> int:
-        """Fetch last position of asin."""
-        if not asin or asin == "error":
-            return 0
+    @staticmethod
+    def _parse_audible_timestamp(raw_ts: Any) -> datetime | None:
+        """Parse an Audible timestamp value into a timezone-aware datetime.
```
OzGav:
```suggestion
        """
        Parse an Audible timestamp value into a timezone-aware datetime.
```

music_assistant/providers/audible/audible_helper.py
```diff
@@ -453,39 +454,68 @@ async def _fetch_chapters(self, asin: str) -> list[dict[str, Any]]:
...
+            return None
+        if parsed.tzinfo is None:
+            parsed = parsed.replace(tzinfo=UTC)
+        return parsed
 
-            if not response:
-                self.logger.debug(f"No last position data available for ASIN {asin}")
-                return 0
+    async def _fetch_last_position(self, asin: str) -> tuple[int, datetime | None] | None:
+        """Fetch the last-heard position for a single ASIN from Audible.
```
OzGav:
```suggestion
        """
        Fetch the last-heard position for a single ASIN from Audible.
```

music_assistant/providers/audible/audible_helper.py
```diff
@@ -453,39 +454,68 @@ async def _fetch_chapters(self, asin: str) -> list[dict[str, Any]]:
...
+            timestamp = self._parse_audible_timestamp(
+                last_position.get(field) or annotation.get(field)
+            )
+            if timestamp is not None:
+                break
+
+        return position_ms, timestamp
+
+    async def get_last_postion(self, asin: str) -> int:
+        """Fetch the last-heard position in milliseconds for the given ASIN.
```
OzGav:
```suggestion
        """
        Fetch the last-heard position in milliseconds for the given ASIN.
```

music_assistant/providers/audible/audible_helper.py
```diff
@@ -529,6 +559,23 @@ async def set_last_position(
         except Exception as exc:
             self.logger.error(f"Unexpected error reporting position for ASIN {asin}: {exc}")
 
+    async def get_audible_resume_position(self, asin: str) -> tuple[bool, int, datetime | None]:
+        """Return resume state for the given ASIN from Audible.
```
OzGav:
```suggestion
        """
        Return resume state for the given ASIN from Audible.
```

music_assistant/providers/audible/audible_helper.py
OzGav: Whilst this is existing code your new methods are calling here and this line could raise an uncaught exception. Suggest catching `audible.exceptions.AudibleError` and re-raising as `ProviderUnavailableError`

music_assistant/providers/audible/audible_helper.py
```diff
@@ -453,39 +454,68 @@ async def _fetch_chapters(self, asin: str) -> list[dict[str, Any]]:
...
+        for field in ("last_updated", "reported_time", "last_updated_time", "timestamp"):
+            timestamp = self._parse_audible_timestamp(
+                last_position.get(field) or annotation.get(field)
+            )
+            if timestamp is not None:
+                break
+
+        return position_ms, timestamp
+
+    async def get_last_postion(self, asin: str) -> int:
```
OzGav: typo

OzGav (commented):
There are some mypy errors and can you mark as resolved the comments that you have done

scootaash:
Thanks @OzGav, I think I've addressed your comments. I've pushed the changes and the lint job is now passing on my fork's CI; hopefully this follows here too.

- For the mypy error: I think audible.exceptions.AudibleError doesn't actually exist in the library — the base transport exception looks to be RequestError, so I've changed _call_api to catch that and re-raise as ProviderUnavailableError.  Please let me know if you'd prefer a different exception there.
- Fixed the get_last_postion typo → get_last_position.
- Moved the docstrings to the multi-line style.
- Dropped the redundant except (...) parentheses so ruff format is happy.

I've pushed these to the branch — I'd like to let the CI run and see if it's green now

OzGav:
`AudibleError` does exist in the library as the base exception class. `RequestError` is a subclass of it. Catching `AudibleError` is preferable I think as it also covers auth-related errors (`AuthFlowError`, `NoRefreshToken`) that could arise during the client call.

https://github.com/mkb79/Audible/blob/458131b4702cca48a8a6eb68c19c21b91b276d37/src/audible/exceptions.py#L8

scootaash:
@OzGav looks like we are both having a day of it, thanks!  :-)

Am I right that `AudibleError` is the base class and that catching it (covering `AuthFlowError`/`NoRefreshToken` too) is the nicer design.

If so I think it's only on the library's `main` branch and hasn't been released yet,  `0.10.0` is the latest version and that's what `requirements_all.txt` pins. 

In `0.10.0` there's no `AudibleError`; `RequestError` is the base, and `AuthFlowError`/`NoRefreshToken` inherit straight from `Exception`. So catching `AudibleError` against the pinned version fails mypy (`attr-defined`) and would raise `AttributeError` at runtime - that was the CI failure.

I used `RequestError` because it's the transport base in `0.10.0` and is still valid on `main` (where it subclasses `AudibleError`). If you'd like to also cover the auth errors on the pinned version I guess we could:
1. Catch them explicitly: `except (RequestError, AuthFlowError, NoRefreshToken) as exc:` — works on `0.10.0` and `main`.
2. Keep `RequestError` for now and switch to `AudibleError` once a release includes it (and the pin is bumped).

Thoughts?

OzGav:
Yeah sorry you are right. Looking at the released code only `RequestError` is raised so all good.

OzGav:
Last confirmation as I can't test this as I don't use Audible. I assume you have been using this and there are no issues with how it works in practice?

scootaash:
Plenty of earlier testing but let me re-verify against the current build before I confirm, rather than rely on my earlier testing (the implementation changed quite a bit since then). I'll install this branch and report back shortly.

scootaash:
Result - re-verified against my live Audible account on the current build. 

Confirmed the read direction works: pressing play in MA fetches the current Audible position and resumes from there, as long as Audible is the most recently-updated source — MA's existing conflict-resolution keeps the freshest record (same behaviour as the other providers). 

Tested with a book last heard on the Audible app and MA resumed from the Audible position correctly. Happy with it from my side. Boom.

OzGav (approved):
Thanks @scootaash

OzGav:
@scootaash are any changes to the docs required for this? If therte are you can PR the BETA branch of the docs repo or just tell me and I will do it.

scootaash:
> @scootaash are any changes to the docs required for this? If therte are you can PR the BETA branch of the docs repo or just tell me and I will do it.

Hi

I've added to 
src/content/docs/music-providers/audible.md
in the beta branch
Hope that's the right place and the below makes sense

---
> ### Whispersync Progress Sync
>
> Music Assistant syncs your listening position with Audible's Whispersync service automatically. This means:
>
> - **Resume anywhere**: If you've been listening on your phone, Echo, or any other Audible device, MA will pick up from exactly where you left off when you press play.
> - **Keep your phone in sync**: When you stop or pause in MA, your position is reported back to Audible so you can continue on any other device.
>
> No configuration is needed — this works automatically once the Audible provider is set up.
---

OzGav:
Looks good. Just open the PR and I can have a proper look.

moelsaied88:
Looking forward to this one! Will the last played time of the book also update so that sort by recently played will also respect cross device sync?

scootaash:
oops, sorry didn't commit from the fork. should be done now. Have tagged you

scootaash:
> Looking forward to this one! Will the last played time of the book also update so that sort by recently played will also respect cross device sync?

Good question — not quite. The Audible last_updated timestamp is used at play-time to determine which position is fresher, so MA resumes from the right place across devices. But MA's own "recently played" sort is driven by its internal playlog, which only updates when you play in MA. So cross-device activity won't reorder the recently played shelf — that would require a separate sync step and is a bit more involved I think, and needs more thought about the impact on other things, was trying to make this one small and targeted.

PR #3894, opened 2026-05-15
Fix Continue Listening / Recently Played shelves when MA is opened via HA ingress

scootaash:
**Observed behaviour**
When I open the MA web UI through Home Assistant's sidebar (HA ingress) and start an audiobook or podcast playing on an mass_* player. The track plays fine and progress is tracked, but:

- The Continue Listening shelf on the home view does not show the audiobook/podcast that's currently playing, even mid-chapter.
- After it finishes (or you stop it), it does not appear in the Recently Played shelf either.
- I then have to navigate to the audiobook/podcast in the library each time to resume — there's no shortcut from the home view.
- Inspecting library.db confirms a row exists in playlog with correct seconds_played / fully_played data, but its userid is the homeassistant_system service-account id, not the human user logged in to the UI.

**Problem**
When the MA web UI is opened via Home Assistant ingress, the WebSocket session authenticates as the homeassistant_system service user. Playback initiated through HA-owned mass_* players writes playlog.userid = <homeassistant_system user id>. The playlog uniqueness key (item_id, provider, media_type, userid) pins the row to that identity. Both in_progress_items() and recently_played() filter strictly by the viewing user's id, so the playlog row never surfaces on the Continue Listening / Recently Played shelves.

**Fix**
Treat homeassistant_system as a shared household identity and OR it into the userid filter alongside the viewing user. The constant HOMEASSISTANT_SYSTEM_USER already exists and is already excluded from list_users() for the same conceptual reason; this aligns the shelf queries with that convention.

There is no schema change or attribution refactor. Per-user scoping for any non-system user remains intact (a different human user's rows are still hidden from the first).

**Notes / alternatives**
A deeper fix would thread the originating HA user through the play_media service call so attribution lands on the right human in the first place, but that's a cross-repo change (HA integration ↔ MA) with much larger scope.

**Code changes**
All edits are to [music_assistant/controllers/music.py](https://github.com/scootaash/MAserver/blob/claude/fix-continue-listening-shelf-8R6P4/music_assistant/controllers/music.py), plus one new import.

Import the existing constant. HOMEASSISTANT_SYSTEM_USER is already defined in /music_assistant/constants.py and used by the auth controller to identify the HA-ingress service user. Imported here so the music controller can look up that user by username.

**recently_played() (parameterised SQL path)**
Previously, when no explicit userid was passed in, the query scoped to the current user only:
```
elif user := get_current_user():
    query += "AND userid = :userid "
    params["userid"] = user.user_id
```
Now, if the current user is not the HA system user, the query expands to include rows owned by the HA system user as well:
```
elif user := get_current_user():
    system_user = await self.mass.webserver.auth.get_user_by_username(
        HOMEASSISTANT_SYSTEM_USER
    )
    if system_user and system_user.user_id != user.user_id:
        query += "AND userid IN (:userid, :system_userid) "
        params["userid"] = user.user_id
        params["system_userid"] = system_user.user_id
    else:
        query += "AND userid = :userid "
        params["userid"] = user.user_id
```
The else branch preserves the original behavior for standalone MA installs (no HA system user exists) and for the unlikely case that the viewer is the system user itself.
in_progress_items() (string-interpolated SQL path)
This function builds its SQL via f-strings rather than bound parameters, so the change matches that style: build a list of allowed user ids, format it as a SQL IN (…) clause, and replace the single-user equality check.
Before:
```
f"AND p.userid = '{user.user_id}' "
After:
allowed_user_ids = [user.user_id]
system_user = await self.mass.webserver.auth.get_user_by_username(
    HOMEASSISTANT_SYSTEM_USER
)
if system_user and system_user.user_id != user.user_id:
    allowed_user_ids.append(system_user.user_id)
allowed_userids_str = "(" + ",".join(f"'{uid}'" for uid in allowed_user_ids) + ")"
# …
f"AND p.userid IN {allowed_userids_str} "
```
User ids are MA-generated UUID-like strings (not user input), so this matches the existing interpolation pattern in this function without introducing new injection surface.

**What is intentionally not changed**
The userid argument-override path on recently_played() is untouched: explicit callers asking for a specific user still get strict per-user scoping.
The all_users=True path on in_progress_items() is untouched.
No changes to write paths (mark_item_played, scrobblers, queue ownership) — attribution still goes to whichever user owns the queue. This PR only widens the read filter on the two shelf queries.

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -69,6 +69,7 @@ from music_assistant.constants import (
     DB_TABLE_TRACK_ARTISTS,
     DB_TABLE_TRACKS,
     DEFAULT_GENRE_MAPPING,
+    HOMEASSISTANT_SYSTEM_USER,
     LOUDNESS_MEASUREMENT_MIN_LUFS,
     PROVIDERS_WITH_SHAREABLE_URLS,
 )
@@ -632,8 +633,20 @@ class MusicController(CoreController):
             query += "AND userid = :userid "
             params["userid"] = userid
         elif user := get_current_user():
-            query += "AND userid = :userid "
-            params["userid"] = user.user_id
+            # Also include rows owned by the HA system user: when the MA web UI
+            # is opened via HA ingress, playback initiated through HA-owned
+            # players scrobbles under that shared system identity rather than
+            # the viewing user.
+            system_user = await self.mass.webserver.auth.get_user_by_username(
+                HOMEASSISTANT_SYSTEM_USER
+            )
+            if system_user and system_user.user_id != user.user_id:
+                query += "AND userid IN (:userid, :system_userid) "
+                params["userid"] = user.user_id
+                params["system_userid"] = system_user.user_id
+            else:
+                query += "AND userid = :userid "
+                params["userid"] = user.user_id
         if queue_id:
             query += "AND queue_id = :queue_id "
             params["queue_id"] = queue_id
@@ -701,6 +714,17 @@ class MusicController(CoreController):
             filter_for_str = available_providers_str
             if user.provider_filter:
                 filter_for_str = "(" + ",".join(f'"{x}"' for x in user.provider_filter) + ")"
+            # Also include rows owned by the HA system user: when the MA web UI
+            # is opened via HA ingress, playback initiated through HA-owned
+            # players scrobbles under that shared system identity rather than
+            # the viewing user.
+            allowed_user_ids = [user.user_id]
+            system_user = await self.mass.webserver.auth.get_user_by_username(
+                HOMEASSISTANT_SYSTEM_USER
+            )
+            if system_user and system_user.user_id != user.user_id:
+                allowed_user_ids.append(system_user.user_id)
+            allowed_userids_str = "(" + ",".join(f"'{uid}'" for uid in allowed_user_ids) + ")"
             query += (
                 f"AND m.provider_instance IN {filter_for_str} "
                 f"AND m.provider_instance IN {available_providers_str} "
@@ -708,7 +732,7 @@ class MusicController(CoreController):
                 f"ELSE (p.provider IN {filter_for_str} AND p.provider IN {available_providers_str})"
                 "END "
                 ") "
-                f"AND p.userid = '{user.user_id}' "
+                f"AND p.userid IN {allowed_userids_str} "
             )
         else:
             # for a library item, we still have to verify via the provider mapping table


OzGav:
I am not sure we want to do this. You have in the PR description the proper fix which is to pass the user via the HA action. Why didn't you do that?

scootaash:
> I am not sure we want to do this. You have in the PR description the proper fix which is to pass the user via the HA action. Why didn't you do that?

Hey there. I was going for the smallest change that restores the broken behaviour to avoid it growing into something with a bigger wider effect, and to get something landed while the larger change sits in the backlog.

The bigger fix, I think, touches the HA integration, the MA WebSocket API, and requires a call on how an HA user maps to an MA user. That mapping is a design decision I didn’t think an outside contributor should be making unilaterally without your input.

If you’d rather see no fix than this read-side approach, I’ll close the PR — totally understandable. Or I can raise an issue instead so the bug is at least documented.

I don’t think I’m the person to take on the cross-repo change, but I wanted to at least surface the bug and a possible low-risk mitigation which resolved something I found personally irritating in my own implementation and thought was probably impacting others too

marcelveldt:
This has come up a few times and I think the real fix is if we make HA pass along the user.
For automations, we should have an optional username/userid parameter on the play_media call.
For play actions triggered from the HA interface, we can fetch the current HA user from the context and translate that to the MA user.

Also, I think this PR description is a bit misleading because when MA is opened through HA ingress, the user IS authenticated correctly. It's the case where a play action is started from HA itself where we have a blind spot.
Those actions currently get executed with the user attached to the API token used by the HA integration.


> When the MA web UI is opened via Home Assistant ingress, the WebSocket session authenticates as the homeassistant_system service user

That is not true. When the HA Ingress session is used, the HA user is sent to MA

OzGav:
I think this is getting superseded by https://github.com/music-assistant/server/pull/4015

scootaash:
Thanks @marcelveldt — you're right, and I had the mechanism wrong in the description. Going back through the auth flow:
- Ingress requests do authenticate as the real HA user (`X-Remote-User-ID` / `X-Remote-User-Name` → resolved to the MA user), so the viewing user is correct. My "authenticates as the `homeassistant_system` user" line was wrong — apologies for the misleading description.
- The actual blind spot is the one you described: play actions started from HA itself (automations, HA media controls, voice) run over the integration's API token, which is the `homeassistant_system` user, so the playlog row gets attributed to that system identity rather than the human who triggered it.

 So my read-side change just papers over the symptom — and it'd leak those rows across all users, which isn't right. Passing the real user through the action (optional `username`/`userid` on `play_media` for automations, and resolving the HA context user for UI-triggered actions) is clearly the correct fix.

Looks like #4015 is already heading that way, so I'm happy to close this in favour of it. I've got a setup that reproduces the original symptom, so if it's useful I can test #4015 against it once it's ready. Thanks for the steer.

PR #3896, opened 2026-05-16
Fix podcast seek for RSS episodes without duration

n-IA-hane:
## Summary

- Lazily probe podcast stream metadata when the RSS episode duration is missing or zero.
- Cache probed duration and stream size for later playback and seek requests.
- Serve podcast feed playback through a custom HTTP range stream so byte-range seek works for MP3 podcast episodes.
- Preserve the original duration and size in `StreamDetails.data`, so repeated seeks keep using the full episode length instead of a remaining-duration value.
- Clear the current audio buffer before replaying the same queue item at a new seek position.
- Add focused unit coverage for podcast byte-range calculation and media size parsing.

## Root Cause

Some podcast feeds expose playable MP3 URLs but report `total_time` as `0`. In that case Music Assistant has no useful item duration, so the frontend cannot expose a usable timeline and `player_queues/seek` rejects the item.

For these streams, `ffprobe` can still read the actual duration and size from the remote MP3. Once that metadata is known, the provider can calculate a byte-range offset for the requested seek position and request `Range: bytes=start-end` from the podcast host.

## Validation

- `python -m py_compile music_assistant/providers/podcastfeed/__init__.py music_assistant/controllers/player_queues.py tests/providers/test_podcastfeed.py`
- `uv run --extra test ruff check music_assistant/providers/podcastfeed/__init__.py music_assistant/controllers/player_queues.py tests/providers/test_podcastfeed.py`
- `uv run --extra test --with hass-client mypy music_assistant/providers/podcastfeed/__init__.py music_assistant/controllers/player_queues.py tests/providers/test_podcastfeed.py`
- `uv run --extra test --with hass-client pytest tests/providers/test_podcastfeed.py tests/core/test_audio_buffer.py tests/core/test_player_controls.py`

Manual validation on a Music Assistant 2.8.7 instance with Radio24 podcast episodes that report missing RSS duration:

- `La Zanzara del 10 aprile 2026`: play, seek near quarter, middle, and near the end.
- `La Zanzara del 15 maggio 2026`: play, seek 10, 5, 3, and 1 minutes from the end.
- Playback continued from the requested position after each seek and the queue stopped cleanly afterward.

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -618,7 +618,9 @@ class PlayerQueuesController(CoreController):
         queue.elapsed_time = 0
         queue.elapsed_time_last_updated = time.time()
         queue.index_in_buffer = None
-        self.mass.create_task(self._cleanup_queue_audio_data(queue_id))
+        self.mass.create_task(
+            self._cleanup_queue_audio_data(queue_id, self._queue_items.get(queue_id, []).copy())
+        )
         self.update_items(queue_id, [])
 
     @api_command("player_queues/save_as_playlist")
@@ -839,13 +841,36 @@ class PlayerQueuesController(CoreController):
         if not queue.current_item:
             raise InvalidCommand(f"Queue {queue_player.state.name} has no item(s) loaded.")
         if not queue.current_item.duration:
+            await self._ensure_current_item_duration(queue_id, queue.current_item)
+        if not queue.current_item.duration:
             raise InvalidCommand("Can not seek items without duration.")
         position = max(0, int(position))
         if position > queue.current_item.duration:
             raise InvalidCommand("Can not seek outside of duration range.")
         if queue.current_index is None:
             raise InvalidCommand(f"Queue {queue_player.state.name} has no current index.")
+        if queue.current_item.streamdetails and queue.current_item.streamdetails.buffer:
+            await queue.current_item.streamdetails.buffer.clear()
+            queue.current_item.streamdetails.buffer = None
         await self.play_index(queue_id, queue.current_index, seek_position=position)
+
+    async def _ensure_current_item_duration(self, queue_id: str, queue_item: QueueItem) -> None:
+        """Try to populate a missing duration before rejecting queue seek."""
+        if queue_item.duration:
+            return
+        try:
+            queue_item.streamdetails = await self.mass.streams.audio.get_stream_details(
+                queue_item=queue_item,
+                seek_position=0,
+            )
+        except MusicAssistantError:
+            return
+        if queue_item.streamdetails.duration:
+            if not queue_item.duration:
+                queue_item.duration = queue_item.streamdetails.duration
+            if queue_item.media_item and not queue_item.media_item.duration:
+                queue_item.media_item.duration = queue_item.streamdetails.duration
+            self.signal_update(queue_id, items_changed=True)
 
     @api_command("player_queues/resume")
     @handle_play_action
@@ -1605,8 +1630,11 @@ class PlayerQueuesController(CoreController):
             prefer_album_loudness=bool(playing_album_tracks),
         )
         # update queue_item.duration from streamdetails if we got a better value
-        if queue_item.streamdetails.duration and not queue_item.duration:
-            queue_item.duration = queue_item.streamdetails.duration
+        if queue_item.streamdetails.duration:
+            if not queue_item.duration:
+                queue_item.duration = queue_item.streamdetails.duration
+            if queue_item.media_item and not queue_item.media_item.duration:
+                queue_item.media_item.duration = queue_item.streamdetails.duration
             self.signal_update(queue_id, items_changed=True)
 
         # pre-initialize the AudioBuffer so audio is ready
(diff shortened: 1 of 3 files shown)


n-IA-hane:
This PR is ready for review. The current failing check is the repository label gate only: I do not have permission to add labels on this upstream repo, so a maintainer needs to add one of the accepted labels. The bugfix label should fit this change.

music_assistant/controllers/player_queues.py
```diff
@@ -618,7 +618,9 @@ def clear(self, queue_id: str, skip_stop: bool = False) -> None:
         queue.elapsed_time = 0
         queue.elapsed_time_last_updated = time.time()
         queue.index_in_buffer = None
-        self.mass.create_task(self._cleanup_queue_audio_data(queue_id))
+        self.mass.create_task(
+            self._cleanup_queue_audio_data(queue_id, self._queue_items.get(queue_id, []).copy())
+        )
```
marcelveldt, line 623: what is this ? seems unrelated and unneeded to copy all queue items

music_assistant/controllers/player_queues.py
```diff
@@ -838,15 +840,38 @@ async def seek(self, queue_id: str, position: int = 10) -> None:
...
         if not queue.current_item.duration:
             raise InvalidCommand("Can not seek items without duration.")
         position = max(0, int(position))
         if position > queue.current_item.duration:
             raise InvalidCommand("Can not seek outside of duration range.")
         if queue.current_index is None:
             raise InvalidCommand(f"Queue {queue_player.state.name} has no current index.")
+        if queue.current_item.streamdetails and queue.current_item.streamdetails.buffer:
+            await queue.current_item.streamdetails.buffer.clear()
+            queue.current_item.streamdetails.buffer = None
```
marcelveldt, line 854: This is just plain wrong. It throws away the buffer so a seek needs to refetch the audio.

music_assistant/controllers/player_queues.py
```diff
@@ -838,15 +840,38 @@ async def seek(self, queue_id: str, position: int = 10) -> None:
             raise PlayerUnavailableError(f"Player {queue_id} is not available")
         if not queue.current_item:
             raise InvalidCommand(f"Queue {queue_player.state.name} has no item(s) loaded.")
+        if not queue.current_item.duration:
+            await self._ensure_current_item_duration(queue_id, queue.current_item)
```
marcelveldt, line 844: This is weird. You can not even reach seek if an item has no duration. Please drop this

music_assistant/controllers/player_queues.py
```diff
@@ -838,15 +840,38 @@ async def seek(self, queue_id: str, position: int = 10) -> None:
...
+            )
+        except MusicAssistantError:
+            return
+        if queue_item.streamdetails.duration:
+            if not queue_item.duration:
+                queue_item.duration = queue_item.streamdetails.duration
+            if queue_item.media_item and not queue_item.media_item.duration:
+                queue_item.media_item.duration = queue_item.streamdetails.duration
+            self.signal_update(queue_id, items_changed=True)
+
```
marcelveldt, line 874: Why is this being hacked into the player queues controller ?
If this is an issue with the podcasts RSS provider, why not do the resolve of duration in the "get stream details"

music_assistant/controllers/player_queues.py
```diff
@@ -1605,8 +1630,11 @@ async def _load_item(
             prefer_album_loudness=bool(playing_album_tracks),
         )
         # update queue_item.duration from streamdetails if we got a better value
-        if queue_item.streamdetails.duration and not queue_item.duration:
-            queue_item.duration = queue_item.streamdetails.duration
+        if queue_item.streamdetails.duration:
+            if not queue_item.duration:
+                queue_item.duration = queue_item.streamdetails.duration
+            if queue_item.media_item and not queue_item.media_item.duration:
+                queue_item.media_item.duration = queue_item.streamdetails.duration
```
marcelveldt, line 1637: why ? this does not have any effect

music_assistant/providers/podcastfeed/__init__.py
```diff
@@ -175,24 +183,260 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
...
                     item_id=item_id,
                     audio_format=AudioFormat(
-                        content_type=ContentType.try_parse(stream_url),
+                        content_type=ContentType.try_parse(
+                            media_info.format if media_info else stream_url.split("?", 1)[0]
+                        ),
                     ),
                     media_type=MediaType.PODCAST_EPISODE,
-                    stream_type=StreamType.HTTP,
+                    stream_type=StreamType.CUSTOM,
```
marcelveldt, line 205: so we're trying to fix seek for -some- episodes missing a duration and then we decide to do all the streaming manually ?

ffmpeg is much more efficient in reading a url or file directly than python so we always try to prefer letting ffmpeg deal with teh audio sources directly and we only use custom streams if there is no other option

marcelveldt:
marked this PR as draft as there just too many red flags.
In the meanwhile I will look if we can maybe do something smart with ffmpeg itself while reading the URL. We already parse the codec dynamically so maybe we can also get the duration on the fly

marcelveldt:
This PR should fix it in a more elegant way: https://github.com/music-assistant/server/pull/3950

PR #3897, opened 2026-05-16
Improve UX of Sync groups

MarvinSchenkel:
This PR adds some UX improvements to the Sync groups based on community feedback

- Dynamic groups now act as a preset, meaning that whenever you have group that consists of players A, B and C and play music to the group AND B was already playing on its own, it will auto-form the group and start playing music as a group
- Dynamic members can now be ungrouped during playback / played to individually
- The member filter has been inverted and is now *inclusive*. The dropdown allows users to limit the options of speakers that show up as options when trying to join speakers to the group
- A bug that would not resolve player ids to their user friendly name during playback has been resolved.
- Pre 2.9 code has been cleaned up

### DIFF
--- a/music_assistant/providers/sync_group/player.py
+++ b/music_assistant/providers/sync_group/player.py
@@ -70,10 +70,7 @@ class SyncGroupPlayer(Player):
         preset_members = cast("list[str]", self.config.get_value(CONF_GROUP_MEMBERS, []))
         if self.is_dynamic:
             # In dynamic mode the configured members act as a preset: they are
-            # pulled in when the group plays (see _form_syncgroup) but the user
-            # can temporarily unjoin them via the OSD. Keeping
-            # _attr_static_group_members empty signals "no locked members" so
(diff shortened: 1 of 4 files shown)


music_assistant/providers/sync_group/player.py
```diff
@@ -67,30 +67,16 @@ async def on_config_updated(self) -> None:
...
-        if changes_made:
-            self.mass.config.set_raw_player_config_value(
-                self.player_id, CONF_GROUP_MEMBERS, static_members
-            )
-
-        self._attr_static_group_members = static_members.copy()
+        preset_members = cast("list[str]", self.config.get_value(CONF_GROUP_MEMBERS, []))
         if self.is_dynamic:
+            # In dynamic mode the configured members act as a preset: they are
+            # pulled in when the group plays
```
marcelveldt: this should be changed to "when the group is powered on"

marcelveldt (approved):

PR #3899, opened 2026-05-17
Surface MusicBrainz artist URL relations as MediaItemLinks

OzGav:
The MB provider already fetches url-rels on every artist details call but discards them. Parse the relations and expose them via ARTIST_METADATA so Wikipedia / official site / Last.fm / AllMusic / social URLs reach the UI without any extra API traffic.

Apart from satisfying an existing TODO the main purpose of this PR is to primarily get the Wikipedia links so that we can better support artist bios in languages other than English.


MarvinSchenkel (approved):
Thanks Gav, LGTM

PR #3900, opened 2026-05-17
Rewrite Deezer provider with GraphQL client

jdaberkow:
## Summary

Full rewrite of the Deezer provider, replacing the REST-based `deezer-python` dependency with [`deezer-python-gql`](https://github.com/music-assistant/deezer-python-gql) - a typed async GraphQL client for Deezer's Pipe API.

### Core architecture

- All metadata (tracks, albums, artists, playlists, podcasts, audiobooks) now fetched via typed GraphQL queries with Pydantic response models
- Shared GraphQL fragments provide consistent field coverage across all item types
- Cursor-based pagination for nested collections (album tracks, playlist tracks, artist albums, audiobook chapters)
- Shares MA's `aiohttp.ClientSession` (`self.mass.http_session`) for connection pooling with JWT auto-refresh

### New capabilities

- **Simplified auth**: Only the ARL cookie token is needed - removed the OAuth browser flow that previously required both an access token and ARL
- **Podcasts**: full library sync, episode browsing, bookmark/resume state sync (read + write via `on_played`)
- **Audiobooks**: library sync, chapter navigation with cumulative position calculation
- **Livestreams (radio)**: search and playback via external stream URLs
- **Lyrics**: synchronized (LRC) and plain text from GraphQL
- **Similar artists**: discover related artists via `ProviderFeature.SIMILAR_ARTISTS`
- **Music Together (Shaker)**: group discovery, suggested and curated playlists in browse/recommendations
- **Flow variants**: mood and genre Flows discovered dynamically via flow config queries
- **Smart Tracklists**: "Made for Me" mixes surfaced in browse and recommendations
- **Dynamic playlists** (`is_dynamic`): Flow, FlowConfig, and Shaker playlists return fresh tracks on each playback - enables endless radio-style queue refill
- **Share URL parsing**: Deezer URLs (`deezer.com/{type}/{id}`) are now resolved from search/paste, with Deezer added to `PROVIDERS_WITH_SHAREABLE_URLS`
- **Personal songs**: user-uploaded tracks accessible via "My Uploads" virtual playlist (GW API `personal_song.getList`)
- **Browse folders**: Made For You, Explore (charts, new releases, editorial playlists), Recently Played (including SmartTracklist items), Shaker, Discover Audiobooks

### Dependency change

- **Removed**: `deezer-python` (REST, unmaintained)
- **Added**: `deezer-python-gql` (async GraphQL, Pydantic models, auto-generated typed client, aiohttp backend)

### GW client changes

- Retained for: track streaming (URL + Blowfish decryption), listen logging, audiobook channel browsing, country code, personal songs
- Non-streaming REST calls removed
- Streaming size calculation made robust for personal tracks (`FILESIZE_MP3_MISC` fallback)

### Breaking changes

None. Drop-in replacement. Existing ARL token configuration is unchanged.

## Related

- Library: https://github.com/music-assistant/deezer-python-gql


music_assistant/providers/deezer/helpers.py
```diff
@@ -0,0 +1,154 @@
...
+    audiobook_id: str,
+    page_size: int = 200,
+) -> list[GetAudiobookAudiobookChaptersEdges]:
+    """Paginate through all chapters of an audiobook and return the full edge list.
+
+    :param gql_client: The Deezer GQL client to use.
+    :param audiobook_id: The audiobook ID to fetch chapters for.
+    :param page_size: Number of chapters per page.
+    """
+    result = await gql_client.get_audiobook(audiobook_id=audiobook_id, chapters_first=page_size)
```
OzGav: `get_audiobook` in `media.py` already fetches with `chapters_first=200` before calling this helper. So the first page is fetched twice.

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+# Deezer CDN image URL pattern
+DEEZER_CDN_IMAGE = "https://e-cdns-images.dzcdn.net/images"
+
+
+def _cover_image(provider: DeezerProvider, cover: object | None) -> MediaItemImage | None:
+    """Create a THUMB image from a GQL cover/picture object with a `.urls` list.
+
+    Returns None when the cover is absent or has no URLs.
+    """
+    if cover and hasattr(cover, "urls") and cover.urls:
```
OzGav: Why is cover typed as object here instead of a proper type that expresses whether .urls is always present or not?

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+    :param playlist: A GQL playlist model (fragment-based or slim search result).
+    :param is_editable: Whether the current user owns this playlist.
+    """
+    images: UniqueList[MediaItemImage] = UniqueList()
+    if img := _cover_image(provider, playlist.picture):
+        images.append(img)
+    owner_name = "Unknown"
+    if playlist.owner:
+        owner_name = playlist.owner.name
+        if not is_editable and playlist.owner.id == provider._user_id:
```
OzGav: Why is a private attribute being accessed directly?

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+        case _:
+            return AlbumType.UNKNOWN
+
+
+def parse_date(date_value: str | None) -> datetime:
+    """Parse a date value from the GQL API to a timezone-aware datetime."""
+    try:
+        return datetime.fromisoformat(str(date_value))
+    except (ValueError, TypeError):
+        return datetime.now(tz=UTC)
```
OzGav: Not sure about this. There is a problem with the date but rather than surfacing it you are just over writing it with now?

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+    if item_type == "playlist" and data.get("PLAYLIST_ID"):
+        return parse_gw_playlist(provider, data)
+    if item_type == "artist" and data.get("ART_ID"):
+        return parse_gw_artist(provider, data)
+    return None
+
+
+def parse_gw_audiobook(provider: DeezerProvider, data: dict[str, Any]) -> Audiobook:
+    """Parse a GW page album item to Music Assistant Audiobook."""
+    album_id = str(data["ALB_ID"])
```
OzGav, line 672: Are the API responses robust enough that these key values will always be available?

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+    )
+    if audiobook.is_favorite:
+        item.favorite = True
+    return item
+
+
+def parse_audiobook_from_album(
+    provider: DeezerProvider, album: AlbumFields | SearchSearchResultsAlbumsEdgesNode
+) -> Audiobook:
+    """Create an Audiobook from an AlbumFields result (search context).
```
OzGav:
```suggestion
    """
    Create an Audiobook from an AlbumFields result (search context).
```
General comment for docstrings throughout. Multi line docstrings should have the first line on its own line to maintain indentation with the following. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,870 @@
...
+            path=deezer_cover_url(md5, img_type),
+            provider=provider.instance_id,
+            remotely_accessible=True,
+        )
+    return None
+
+
+def apply_web_url(item: Artist | Album, gql_result: object) -> None:
+    """Set web URL on provider mappings if available in the GQL result."""
+    if hasattr(gql_result, "url") and hasattr(gql_result.url, "web_url"):
```
OzGav: The parameter type is object, which defeats typing. Should accept a typed union of the GQL models that have .url.web_url.

music_assistant/providers/deezer/provider.py
```diff
@@ -0,0 +1,288 @@
...
+    streaming_manager: DeezerStreamingManager
+    _user_id: str
+
+    async def handle_async_init(self) -> None:
+        """Handle async init of the Deezer provider."""
+        arl_token = str(self.config.get_value(CONF_ARL_TOKEN))
+
+        self.gql_client = DeezerGQLClient(arl=arl_token)
+        me = await self.gql_client.get_me()
+        if me is None:
```
OzGav: What happens if `get_me()` raises rather than returning None? And what exception does the caller see if `GWClient.setup()` raises `DeezerGWError`?

music_assistant/providers/deezer/provider.py
```diff
@@ -0,0 +1,288 @@
...
+        self.gw_client = GWClient(self.mass.http_session, arl_token)
+        await self.gw_client.setup()
+
+        self.media_manager = DeezerMediaManager(self)
+        self.browse_manager = DeezerBrowseManager(self)
+        self.streaming_manager = DeezerStreamingManager(self)
+
+    async def unload(self, is_removed: bool = False) -> None:
+        """Handle unload/close of the provider."""
+        await self.gql_client.close()
```
OzGav: No `super().unload()`?

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+            if ab_id in seen_ids:
+                continue
+            try:
+                yield await self.get_audiobook(ab_id)
+            except MediaNotFoundError:
+                continue
+
+    # -- Search --
+
+    @use_cache(3600 * 24 * 7)  # type: ignore[type-var]
```
OzGav: Why would you cache a search for a week?

What is the typing issue with `@use_cache` here and why is it suppressed rather than fixed?

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        self.logger = provider.logger
+        self._audiobook_ids_cache: set[str] | None = None
+
+    # -- Pagination helper --
+
+    async def _iter_paged(
+        self,
+        fetch: Callable[..., Awaitable[Any]],
+        extract: Callable[..., Any],
+    ) -> AsyncGenerator[Any, None]:
```
OzGav: This is losing type safety. Can this be done better?

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        async for edge in self._iter_paged(
+            self.provider.gql_client.get_favorite_artists,
+            lambda r: r.user_favorites.artists,
+        ):
+            item = parse_artist(self.provider, edge.node)
+            if edge.favorited_at:
+                item.date_added = parse_date(edge.favorited_at)
+            yield item
+        # Also include artists from user-uploaded personal songs
+        results = await self.provider.gw_client.get_personal_songs(start=0, nb=500)
```
OzGav: Personal songs are fetched separately in `get_library_artists`, `get_library_albums`, and `get_library_tracks` This should be cached or fetched once and shared to avoid unnecessary API calls.

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+            )
+        return self._audiobook_ids_cache
+
+    async def get_library_albums(self) -> AsyncGenerator[Album, None]:
+        """Retrieve all library albums from Deezer."""
+        # Deezer returns audiobooks mixed into the favorite albums list.
+        # Use check_audiobook_ids to identify and exclude them.
+        audiobook_ids = await self._get_audiobook_ids_in_albums()
+        async for edge in self._iter_paged(
+            self.provider.gql_client.get_favorite_albums,
```
OzGav, line 188: `_get_audiobook_ids_in_albums` loops through all the user's favourite albums page by page just to collect their IDs. Then `get_favorite_albums` loops through all the same favourite albums page by page again to yield them. So the full album list is fetched from Deezer twice in a row. Can this be done in a single pass?

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        for i in range(0, len(uncached_ids), batch_size):
+            batch = uncached_ids[i : i + batch_size]
+            fetched = await self.provider.gql_client.get_podcast_episodes_by_ids(ids=batch)
+            for ep in fetched:
+                if ep is not None:
+                    parsed = parse_podcast_episode(
+                        self.provider, ep, podcast_mapping, 0, podcast_image_url
+                    )
+                    cached_episodes[ep.id] = parsed
+                    self.mass.create_task(
```
OzGav, line 639: This method is cached and now you are also writing to an additional cache?

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        elif item.media_type == MediaType.TRACK:
+            await self.provider.gql_client.add_track_to_favorite(track_id=item.item_id)
+        elif item.media_type == MediaType.PLAYLIST:
+            await self.provider.gql_client.add_playlist_to_favorite(playlist_id=item.item_id)
+        elif item.media_type == MediaType.PODCAST:
+            await self.provider.gql_client.add_podcast_to_favorite(podcast_id=item.item_id)
+        elif item.media_type == MediaType.AUDIOBOOK:
+            await self.provider.gql_client.add_album_to_favorite(album_id=item.item_id)
+        else:
+            raise NotImplementedError
```
OzGav: Should be UnsupportedFeaturedException from MA errors, not the Python built-in

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        elif media_type == MediaType.TRACK:
+            await self.provider.gql_client.remove_track_from_favorite(track_id=prov_item_id)
+        elif media_type == MediaType.PLAYLIST:
+            await self.provider.gql_client.remove_playlist_from_favorite(playlist_id=prov_item_id)
+        elif media_type == MediaType.PODCAST:
+            await self.provider.gql_client.remove_podcast_from_favorite(podcast_id=prov_item_id)
+        elif media_type == MediaType.AUDIOBOOK:
+            await self.provider.gql_client.remove_album_from_favorite(album_id=prov_item_id)
+        else:
+            raise NotImplementedError
```
OzGav: Should be UnsupportedFeaturedException from MA errors, not the Python built-in

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,769 @@
...
+        """Get (full) podcast episode details by id."""
+        result = await self.provider.gql_client.get_podcast_episode(
+            podcast_episode_id=prov_episode_id,
+        )
+        if result is None:
+            raise MediaNotFoundError(f"Podcast episode {prov_episode_id} not found on Deezer")
+        podcast_mapping = ItemMapping(
+            media_type=MediaType.PODCAST,
+            item_id=result.podcast.id,
+            provider=self.instance_id,
```
OzGav, line 517: Elsewhere `self.provider.instance_id` is used?

music_assistant/providers/deezer/browse.py
```diff
@@ -0,0 +1,1092 @@
...
+        if recently_played:
+            result.append(
+                RecommendationFolder(
+                    item_id="recently_played",
+                    provider=self.instance_id,
+                    name="Recently Played",
+                    items=UniqueList(recently_played),
+                )
+            )
+        # Convert all items to ItemMapping to work around a core serialization bug
```
OzGav: I think you will need to fix this rather than work around it.

music_assistant/providers/deezer/browse.py
```diff
@@ -0,0 +1,1092 @@
...
+        """Browse Deezer content.
+
+        :param path: The browse path.
+        :param base_browse: Coroutine for the base class browse method.
+        """
+        path_parts = path.split("://")[1].split("/") if "://" in path else []
+        subpath = path_parts[0] if path_parts else None
+        sub_subpath = path_parts[1] if len(path_parts) > 1 else None
+
+        if subpath == "Made For You":
```
OzGav: These should all be provider level constants

music_assistant/providers/deezer/browse.py
```diff
@@ -0,0 +1,1092 @@
...
+        )
+        if charts is None or charts.country is None or charts.country.tracks is None:
+            return []
+        return [
+            parse_track(self.provider, edge.node)
+            for edge in charts.country.tracks.edges
+            if edge.node is not None
+        ]
+
+    @use_cache(3600)  # type: ignore[type-var]
```
OzGav: Should this be cached since `is_dynamic` is true?

music_assistant/providers/deezer/streaming.py
```diff
@@ -0,0 +1,378 @@
...
+        """Initialize streaming manager."""
+        self.provider = provider
+        self.mass = provider.mass
+        self.instance_id = provider.instance_id
+        self.domain = provider.domain
+        self.logger = provider.logger
+
+    # -- Resume position --
+
+    async def get_resume_position(
```
OzGav, line 47: This seems to be duplicating `_fetch_all_bookmarks` in `media.py`

music_assistant/providers/deezer/streaming.py
```diff
@@ -0,0 +1,378 @@
...
+            chunk_count = ceil(streamdetails.size / 2048)
+            skip_chunks = int(chunk_count / streamdetails.duration) * seek_position
+        else:
+            skip_chunks = 0
+
+        buffer = bytearray()
+        streamdetails.data["start_ts"] = utc_timestamp()
+        streamdetails.data["stream_id"] = uuid.uuid1()
+        self.mass.create_task(self.provider.gw_client.log_listen(next_track=streamdetails.item_id))
+        async with self.mass.http_session.get(
```
OzGav, line 333: You need to add error handling here.

OzGav (dismissed):

MarvinSchenkel:
Awesome stuff @jdaberkow . Just a heads up, we are almost in a feature freeze for the 2.9 release and are currently focusing on wrapping up new features. So this rewrite/refactor will be merged as soon as we start the 2.10 beta cycle 🙏

marcelveldt:
> Pydantic response models
Any chance you could use standard python dataclasses (icw mushumaro maybe) insteads of bloated pydantic models ?

> Self-managed httpx.AsyncClient pool 
Not sure if I'm a fan of this. Why introduce another http client for one provider ? We use a shared aiohttp session

marcelveldt:
Marked as draft considering there are some open review comments and my questions above.
Press "ready for review" once this is ready for another look

music_assistant/providers/deezer/browse.py
jdaberkow: @OzGav 

So from my observation the thing is the following:
`RecommendationFolder.items` is typed as `UniqueList[MediaItemType | ItemMapping | BrowseFolder]`. When `@use_cache` hits and reconstructs the object via `parse_value` -> `RecommendationFolder.from_dict()`, mashumaro has no discriminator for the Union, so it tries types in declaration order. `Artist` is first in `MediaItemType`, and since its required fields overlap enough, mashumaro deserializes everything as Artist. Playlist-specific fields like `is_dynamic` are lost.

`to_dict()` (what the frontend receives) is fine. The corruption only happens on the cache roundtrip through `from_dict()`. I assume this also affects YTMusic and Apple Music, they both `@use_cache(3600)` their `recommendations()` with full objects in items. 

Fix options as I see them:

1. Proper fix in `music_assistant_models`: add a mashumaro `Discriminator(field="media_type")` on the items Union.

2. Keep the ItemMapping workaround: ItemMapping objects survive the roundtrip correctly (they fail Artist deserialization due to missing `provider_mappings`, mashumaro falls through to the correct type). Works, just ugly.

I assume we should rather go with option 1 then? Would you rather have this in a seperate PR or is the preferred way to simply fix this within this PR?

music_assistant/providers/deezer/media.py
jdaberkow: @OzGav 

The caching here is in fact wrong. Will fix that.

But regarding the typing issue (I am suppressing this in many places):

The `@use_cache` decorator has `ProviderT = TypeVar("ProviderT", bound="Provider | CoreController")`. The manager classes (`DeezerBrowseManager`, `DeezerMediaManager`) aren't subclasses of either. They're plain classes with `self.mass`, `self.instance_id`, and `self.domain` copied from the provider. The decorator works fine at runtime (it only accesses `self.mass.cache` and `getattr(self, "instance_id", self.domain)`), but mypy can't verify the bound.

Apple Music has the same pattern (`AppleMusicRecommendationManager`) but sidesteps it because `music_assistant/providers/apple_music/.*` is excluded from mypy entirely in `pyproject.toml`.

A possible fix would be changing the TypeVar bound to a Protocol in the core:

```python
# in music_assistant/controllers/cache/helpers.py

class _Cacheable(Protocol):
    mass: MusicAssistant
    domain: str

ProviderT = TypeVar("ProviderT", bound="_Cacheable")
```

That's all `@use_cache` structurally needs. Provider, CoreController, and any manager class would satisfy it without changes.

Would this be something we could include here, or should it be a separate small PR?

OzGav: ~~I removed the exclude from Apple Music and I see no mypy errors on appleMusicRecommendationManager.~~

~~What actual error are you ignoring?~~

Actually I had been working on mypy fixes for the Apple Music provider a while ago and forgot I was going to PR a change and it was in the branch I was looking at. So yes soemthing like what you have suggested above as a separate small PR I think would be good because as you say it will benefit a couple of places in the code.

music_assistant/providers/deezer/browse.py
OzGav: Yes I think fix this properly in the models. It will need to be a separate PR as its a different repo. but I will ask @MarvinSchenkel to confirm

jdaberkow:
@marcelveldt 
###  Pydantic response models
> Any chance you could use standard python dataclasses (icw mushumaro maybe) insteads of bloated pydantic models ?

Unfortunately this is not possible when using the ariadne-codegen tool that I use for graphql code generation (which is a real game changer when developing type-safe towards a graphql API).
ariadne-codegen only generates Pydantic models. No configuration or plugin exists for dataclass/mashumaro output. But the Pydantic models are confined to the `deezer-python-gql` library boundary: they're parsed into MA's own dataclass-based `MediaItem` types in `parsers.py` and immediately discarded. They never flow through MA's core. The memory/CPU overhead is negligible compared to network I/O.


###  Self-managed httpx.AsyncClient pool
> Not sure if I'm a fan of this. Why introduce another http client for one provider ? We use a shared aiohttp session

Agreed. ariadne uses httpx as a default. Thats why I simply applied that as well. But since the base client is customized either way, I was able to switch to aiohttp and pass the already existing session. Fixed. 👍

music_assistant/providers/deezer/media.py
jdaberkow: Here it is:

https://github.com/music-assistant/server/pull/3965

music_assistant/providers/deezer/helpers.py
jdaberkow: Right. Fixed by passing the first set of already fetched edges towards the helper in case we need more.

music_assistant/providers/deezer/media.py
jdaberkow: Yes. Thats right. The two layers serve different purposes:

Outer cache prevents repeated work during interactive browsing. Without it, every navigation to a 500-episode podcast triggers 1 API call + 500 individual cache lookups. With it, repeat visits within the hour are a single cache.get() returning the full list.

Inner per-episode cache (30 days) prevents re-fetching episode details that rarely change. When the outer cache expires after 1h, the method re-runs but only batch-fetches genuinely new episodes (everything else hits the 30-day individual cache).

I've reduced the outer TTL from 24h to 1h (new episodes now appear within an hour instead of a day) and added a docstring explaining the two-layer strategy so the intent is clear to future readers.

OzGav: LGTM!

music_assistant/providers/deezer/streaming.py
OzGav: I see this has been fixed.

music_assistant/providers/deezer/media.py
jdaberkow: I improved it slightly. But I fear its not possible to return the exact type here, due to the generic nature of this helper function which uses the fetch Callable internally. The alternative would be to implement the iteration logic 7 times. :/

music_assistant/providers/deezer/provider.py
jdaberkow: Improved the error handling here. 👍

music_assistant/providers/deezer/media.py
OzGav: Seems like a good solution to me.

music_assistant/providers/deezer/parsers.py
jdaberkow: Defined `_CoverLike` Protocol with a `urls: list[str]` property.

music_assistant/providers/deezer/parsers.py
jdaberkow: Renamed to `self.user_id`. Makes sense to treat this as a public value.

music_assistant/providers/deezer/parsers.py
jdaberkow: Yeah. That was in fact a bad idea. Return type changed to `datetime | None`. Now returns `None` on parse failure.

music_assistant/providers/deezer/parsers.py
jdaberkow: Hard to say. So I hardened this place a little bit.
Entry point `parse_gw_item` validates required keys with `.get()` guards before calling inner parsers, and wraps all inner parser calls in `try/except KeyError` with a debug log.

music_assistant/providers/deezer/parsers.py
jdaberkow: Fixed. All multi-line docstrings across the provider should now follow these conventions.

music_assistant/providers/deezer/parsers.py
jdaberkow: Defined `_HasUrl` Protocol with a `url` property, to allow this method to be a little more generic, than being only suitable for specific GQL types.

music_assistant/providers/deezer/provider.py
jdaberkow: `await super().unload(is_removed)` is now called in the `unload` method.

music_assistant/providers/deezer/media.py
jdaberkow: Added `_get_personal_songs()` method with `@use_cache(3600 * 24)`.

music_assistant/providers/deezer/media.py
jdaberkow: Fixed. Single-pass in `get_library_albums`: collect all edges into memory, call `check_audiobook_ids` once, yield non-audiobooks from the in-memory list. Renamed `_audiobook_ids_cache` -> `_audiobook_ids_in_favorites` for clarity. `_get_audiobook_ids_in_albums()` retained as fallback for standalone `get_library_audiobooks()` calls.

music_assistant/providers/deezer/media.py
jdaberkow: Now raises `UnsupportedFeaturedException(f"Unsupported media type for ...: {media_type}")` with a descriptive message.

music_assistant/providers/deezer/media.py
jdaberkow: Now raises `UnsupportedFeaturedException(f"Unsupported media type for ...: {media_type}")` with a descriptive message.

music_assistant/providers/deezer/media.py
jdaberkow: Now alll manager classes store `self.instance_id = provider.instance_id` in `__init__` and use `self.instance_id` consistently throughout.

music_assistant/providers/deezer/browse.py
jdaberkow: @OzGav @MarvinSchenkel Shall I proceed in making a PR in the models repo for that?

music_assistant/providers/deezer/browse.py
jdaberkow: Fixed. All browse path strings defined as module-level constants in `helpers.py`.

music_assistant/providers/deezer/browse.py
jdaberkow: Definitely not. :) Fixed.

music_assistant/providers/deezer/streaming.py
jdaberkow: Added `if resp.status != 200: raise MediaNotFoundError(...)` check immediately after opening the HTTP response context manager, before iterating chunks.

music_assistant/providers/deezer/parsers.py
OzGav: LGTM now.

music_assistant/providers/deezer/helpers.py
```diff
@@ -0,0 +1,210 @@
...
+BROWSE_RECENTLY_PLAYED = "Recently Played"
+BROWSE_SHAKER = "Shaker"
+BROWSE_AUDIOBOOKS = "Discover Audiobooks"
+BROWSE_MOODS = "Moods"
+BROWSE_GENRES = "Genres"
+BROWSE_YOUR_TOP_ARTISTS = "Your Top Artists"
+BROWSE_YOUR_TOP_ALBUMS = "Your Top Albums"
+BROWSE_RECOMMENDED_PLAYLISTS = "Recommended Playlists"
+BROWSE_RECOMMENDED_ARTIST_PLAYLISTS = "Recommended Artist Playlists"
+BROWSE_PERSONALIZED_PLAYLISTS = "Personalized Playlists"
```
OzGav: I think these should be in constants.py

music_assistant/providers/deezer/media.py
```diff
@@ -0,0 +1,806 @@
...
+        return parse_radio(self.provider, result)
+
+    @use_cache(3600 * 24 * 30, allow_expired_cache=True)
+    async def get_podcast(self, prov_podcast_id: str) -> Podcast:
+        """Get full podcast details by id."""
+        result = await self.provider.gql_client.get_podcast(podcast_id=prov_podcast_id)
+        if result is None:
+            raise MediaNotFoundError(f"Podcast {prov_podcast_id} not found on Deezer")
+        podcast = parse_podcast(self.provider, result)
+        if hasattr(result, "raw_episodes"):
```
OzGav: Minor nit. You fixed a similar thing with cover_image so it would be nice to do the same here

music_assistant/providers/deezer/parsers.py
```diff
@@ -0,0 +1,919 @@
...
+
+
+def parse_audiobook_from_album(
+    provider: DeezerProvider, album: AlbumFields | SearchSearchResultsAlbumsEdgesNode
+) -> Audiobook:
+    """
+    Create an Audiobook from an AlbumFields result (search context).
+
+    Used when an album search result is identified as an audiobook via
+    check_audiobook_ids. Provides basic metadata from the album fields.
```
OzGav: Nit. This explains internal workings so should be in an inline comment if at all.

OzGav (commented):
Just a couple of other small items to consider. Otherwise LGTM. One of the other guys will have a look at this and we have the models PR question to go. Thanks for all your work on this re-write!

music_assistant/providers/deezer/helpers.py
jdaberkow: Done. 👍

music_assistant/providers/deezer/media.py
jdaberkow: Actually thats not even necessary, since the existing typing is already sufficient. Fixed.

music_assistant/providers/deezer/parsers.py
jdaberkow: Removed. 👍

OzGav:
Thanks. If you are happy with everything then press the ready for review button and one of the others will have a look soon.

jdaberkow:
> Thanks. If you are happy with everything then press the ready for review button and one of the others will have a look soon.

Thank you very much for your detailed reviews! Alright. From my side everything is good to go.

music_assistant/providers/deezer/browse.py
jdaberkow: https://github.com/music-assistant/models/pull/239

Created a bugfix pr in the models repo so that there is a baseline to discuss this topic on. :)

jdaberkow: @OzGav 
Fix is merged into models. So I removed this workaround here. Solved. :)

music_assistant/providers/deezer/browse.py
```diff
@@ -0,0 +1,1090 @@
...
+                    break
+        return [
+            create_virtual_playlist(
+                self.provider, TOP_CHARTS_PLAYLIST_ID, "Top Charts", image_url=charts_cover
+            ),
+            BrowseFolder(
+                item_id="top_albums",
+                provider=self.instance_id,
+                path=f"{base}Top Albums",
+                name="Top Albums",
```
OzGav: You could make these constants like you did above?

music_assistant/providers/deezer/browse.py
OzGav: Nice

music_assistant/providers/deezer/browse.py
jdaberkow: Good point! Introduced even more constants while I was on it. 👍

OzGav:
Just while you wait. Can you keep an eye on this https://github.com/music-assistant/server/pull/4048 to make sure your provider will not regress any fixes.

jdaberkow:
@OzGav 
> Just while you wait. Can you keep an eye on this #4048 to make sure your provider will not regress any fixes.

Good call! Implemented the fix from #4048 into this PR as well. 👍

marcelveldt:
make sure to add your name to the maintainers list within the manifest @jdaberkow you basically rewrote the entire provider!

jdaberkow:
> make sure to add your name to the maintainers list within the manifest @jdaberkow you basically rewrote the entire provider!

Thats done. :)

The docs do still need an update though, but I deferred that to when this is merged.

OzGav:
Please do put any update in to the beta branch of the docs soon so it is ready to go,

jdaberkow:
> Please do put any update in to the beta branch of the docs soon so it is ready to go,

Opened a PR :+1: 

https://github.com/music-assistant/music-assistant.io/pull/746

marcelveldt (approved):
Amazing job, @jdaberkow !

PR #3902, opened 2026-05-18
fixed mypy errors

anatosun:


MarvinSchenkel:
Could you maybe bump the plex api to 4.18.1 in this PR as well so we can check everything in one go and merge it?

anatosun:
Sure, I thought that since I was merging against the dependabot branch, it was not necessary. Do you confirm that I should modify `requirements_all.txt`?

MarvinSchenkel:
Ah I totally missed that, I thought it was a PR against dev. In that case, just fix the merge conflicts and I will merge both PRs 👍

PR #3903, opened 2026-05-18
Fix Airplay not stopping stream on some devices.

MarvinSchenkel:
Fixes a regression that was introduced in an attempt to fix this issue: https://github.com/music-assistant/support/issues/5033


music_assistant/providers/airplay/provider.py
```diff
@@ -364,45 +364,33 @@ async def _handle_dacp_request(  # noqa: PLR0915
                     # Already handling a prevent-playback for this stream
                     # (duplicate message while ungroup/stop is still in progress)
                     self.logger.debug("Ignoring duplicate prevent-playback for %s", player.name)
+                elif not player.stream._connected.is_set():
```
marcelveldt: accessing a private variable here - consider creating a public property for this

MarvinSchenkel: Ah missed that one, fixed 👍

marcelveldt (approved):

PR #3904, opened 2026-05-18
Fix Hue sync bridge not being able to join to players with a sendspin bridge

MarvinSchenkel:


music_assistant/providers/sendspin/player.py
```diff
@@ -1056,6 +1057,20 @@ def __init__(
         """
         super().__init__(provider, player_id, initial_hello)
         self._attr_supported_features = {PlayerFeature.SET_MEMBERS}
+        # Self-advertise the sendspin protocol so cross-provider grouping can match
```
marcelveldt: I think we should consider changing the output protocols a bit and replace the current "native" virtual option with the actual domain. There is already a property on the OutputProtocol model that indicates that the protocol is native.

MarvinSchenkel: So does that match what I suggested as an alternative on Discord? i.e.

In the ABC `Player.output_protocols`:
```
if (
    self.provider.domain in PROTOCOL_PRIORITY
    and PlayerFeature.SET_MEMBERS in self.supported_features
):
...add OutputProtocol
```

marcelveldt: Yes and a step beyond that even

MarvinSchenkel: Implemented the change to player.py. Could you elaborate a bit more on the 'step beyond'? Not really sure what needs changing.

music_assistant/models/player.py
```diff
@@ -1035,6 +1035,21 @@ def output_protocols(self) -> list[OutputProtocol]:
...
+            result.append(
+                OutputProtocol(
+                    output_protocol_id=self.player_id,
+                    name=self.provider.name,
+                    protocol_domain=self.provider.domain,
+                    priority=PROTOCOL_PRIORITY[self.provider.domain],
+                    available=self.available,
+                    is_native=True,
+                )
+            )
```
marcelveldt, line 1052: I was more talking about merging these together so we don't have this magic "native" value anymore

MarvinSchenkel: Is this something we want to do in this PR though? I feel that's quite a big change?

PR #3906, opened 2026-05-18
Subsonic: Provide close implementation that cleans up

khers:
The connection object from py-opensonic now provides and explicit way to cleanup the underlying async http session as it seems that the standard destructor does not handle this. Make sure we cleanly close the session when the proivder is unloaded.


MarvinSchenkel (approved):
Thanks @khers !

PR #3914, opened 2026-05-19
Remove Sonos from mypy excludes

OzGav:
This appears to have been accidentally added in this PR https://github.com/music-assistant/server/pull/3902


MarvinSchenkel (approved):
Sharp eyes

PR #3915, opened 2026-05-19
Compute the artwork-derived color palette in the backend

maximmaxim345:
This PR moves the color palette extraction from the frontend to the backend to prepare for the `color@v1` Sendspin role.

Extract a color palette from cover art with WCAG-AA contrast guarantees. The players controller attaches the palette to `PlayerMedia.palette`.

The next queue item's palette is also prefetched so clients see colors without flickering at track transitions.

Adds a python port of `colorthief` as a dependency. I also tried out `Image.quantize` but that looked more washed-out from some testing.

Depends on:
- https://github.com/music-assistant/models/pull/228

### DIFF
--- a/music_assistant/models/player.py
+++ b/music_assistant/models/player.py
@@ -1473,14 +1473,6 @@ class Player(ABC):
             needs_setup=self.needs_setup,
         )
 
-        # Peek the cached palette so the media checksum below sees palette
-        # transitions. Lazy import to avoid the helpers.colors -> helpers.images
-        # -> models.plugin cycle.
-        if (cm := self._state.current_media) and cm.image_url:
-            from music_assistant.helpers.colors import peek_palette_for_url  # noqa: PLC0415
-
-            cm.palette = peek_palette_for_url(cm.image_url)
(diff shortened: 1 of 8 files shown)


music_assistant/models/player.py
```diff
@@ -1473,6 +1473,14 @@ def __calculate_player_state(
...
         )
 
+        # Peek the cached palette so the media checksum below sees palette
+        # transitions. Lazy import to avoid the helpers.colors -> helpers.images
+        # -> models.plugin cycle.
+        if (cm := self._state.current_media) and cm.image_url:
+            from music_assistant.helpers.colors import peek_palette_for_url  # noqa: PLC0415
+
+            cm.palette = peek_palette_for_url(cm.image_url)
+
```
maximmaxim345: this is really ugly, any idea where this should go? has to run before update_state's
checksum compare so sendspin re-sends when palette flips from none to populated, but this isnt really a player state.

MarvinSchenkel (approved):
LGTM!

PR #3922, opened 2026-05-20
Improve lastfm recommendations

OzGav:
When using this against my full library I noticed that the "Discover Similar Tracks" and "Discover {Genre} Artists/Albums/Tracks" rows often returned only 9 items instead of the target 10. Digging into the debug logs showed two unrelated causes, plus a related issue where tracks I already owned were being linked to Spotify instead of my local copy.

Link resolved tracks back to the library when the streaming provider exposes a matching ISRC. Last.fm's similar-tracks API frequently omits MBIDs, so the upfront library check by external ID misses and we fall through to a streaming provider search. The Spotify result reliably carries an ISRC though, so after the streaming search returns, we now re-check the library against the resolved item's external IDs and prefer the library copy when one exists. No extra API calls — we just use the IDs we already have in hand.

Bump `SIMILAR_ITEMS_PER_SEED` from 3 to 5. With 5 seed tracks/artists and 3 items per seed, a single Last.fm "Track not found" error (which happens when Last.fm doesn't recognise an MBID it gave us) drops the ceiling below 10 before deduplication even runs.

Bump `RESOLUTION_BUFFER_LARGE` from 30 to 40. For the genre rows we filter out items already in the library before resolving (so users discover something new). In my case with mainly rock music, 21 of the 30 fetched items were already in the library, leaving only 9 survivors — not enough to fill the row. 40 gives roughly 12 survivors at a 30% novelty rate, which consistently hits the target of 10.

Add debug logging to `_is_in_library`. Each filter decision now logs which item was dropped and which check matched (MBID vs. name search), including the library item's name on a name match. This makes it easier to see which items are being filtered and why a row may come up short.

### DIFF
--- a/music_assistant/providers/lastfm_recommendations/constants.py
+++ b/music_assistant/providers/lastfm_recommendations/constants.py
@@ -32,7 +32,7 @@ TARGET_ITEM_COUNT = 10
 RESOLUTION_BUFFER_SMALL = 15
 
 # Number of items to fetch when we expect many resolution failures (large buffer)
-RESOLUTION_BUFFER_LARGE = 30
+RESOLUTION_BUFFER_LARGE = 40
 
 # Number of top items to always include when sampling (before random selection)
 TOP_ITEMS_TO_TAKE = 3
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
LGTM

PR #3923, opened 2026-05-20
Fix genre movements when genres are promoted or deleted

OzGav:
When a user promoted an alias to its own genre in the UI, the result was inconsistent. Two symptoms (reported [here](https://github.com/orgs/music-assistant/discussions/5507#discussioncomment-16983417)):

Albums and artists kept the old genre. A user with files tagged "Rap" (an alias of "Hip Hop") promoted "Rap" to its own genre. Their tracks moved to "Rap" as expected, but their albums and most artists were still listed under "Hip Hop".
Tracks ended up split across genres. A user with files tagged "Manele", which Music Assistant had assigned as an alias of both "Folk" and "Pop", promoted "Manele". Tracks ended up tagged as both "Folk" and "Manele" instead of just "Manele".

I believe this happens because `promote_alias_to_genre` was doing too little:

It only touched the one source genre the user clicked from, so if the alias was shared by multiple genres (n:n is supported elsewhere in the system), the other genres kept claiming the alias and kept their mappings.

It only moved rows in the mapping table whose alias column matched the promoted string. Rows produced by track-to-parent propagation have alias = NULL, so albums/artists derived from those tracks were never re-pointed at the new genre.
`remove_alias` and `merge_genres` had the same blind spot: they changed the track-level mappings without regenerating the derived album/artist rows that depended on them.

So this PR adjusts `promote_alias_to_genre` to now find every genre that claims the alias, moves all matching mappings to the new genre in one statement (UPDATE OR REPLACE handles the rare uniqueness collision), strips the alias from each owner, and re-runs propagation so derived album/artist rows follow the moved tracks.

`remove_alias` and `merge_genres` now also re-run propagation for the same reason.

The original report mentioned that album-artists moved with the promotion while featured artists did not. From the code, both go through the same propagation path, so I couldn't pin down what makes them behave differently without seeing the user's data. The fix corrects the broader stale-mapping issue; if the problem persists, the next step is to inspect the actual rows in `genre_media_item_mapping` for one of each kind of artist.

### DIFF
--- a/music_assistant/controllers/media/genres.py
+++ b/music_assistant/controllers/media/genres.py
@@ -1222,6 +1222,9 @@ class GenreController(MediaControllerBase[Genre]):
             "WHERE genre_id = :genre_id AND LOWER(alias) = LOWER(:alias)",
             {"genre_id": db_id, "alias": alias},
         )
+        # Rebuild propagated album/artist rows: the track rows we just removed may have
+        # fed them, so a derived row on this genre would otherwise linger past the alias.
+        await self._propagate_genre_mappings_to_parents()
         updated = await self.get_library_item(db_id)
         self.mass.signal_event(EventType.MEDIA_ITEM_UPDATED, updated.uri, updated)
         return updated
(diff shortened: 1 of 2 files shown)


MarvinSchenkel (approved):
Thanks @OzGav ! 🙏

PR #3928, opened 2026-05-20
Fix HEOS showing incorrect Now Playing

### LINKED ISSUE #5311: HEOS players: Artwork and Artist/Track Info not visible after upgrading to 2.8.x

2.8.5

Home Assistant OS Addon

- [ ] I have tried a wired connection for issues related to interrupted or poor playback quality
- [ ] For voice problems, I have sought help elsewhere before returning here
- [ ] For playback problems, I have recycled power to the physical device

### The problem

[I edited this comprehensively since I found out what the issue seems to be]
Since the upgrade to MA 2.8.x, there is no more artwork visible in MA when playing through HEOS speakers, beacuse the protocol is switched automatically (from Airplay) to HEOS. When switching back to default Airplay protocol of the HEOS speaker, it works.

In fact, whats happening, in MA when the player protocol of the HEOS player is set to HEOS, the artwork and artist info is visible for 0.5s to 4s (its random), and then it switches to "Url stream".

I have read that this is a known issue for HEOS displays and the HEOS App itself (https://www.music-assistant.io/player-support/heos/#known-issues--notes), however I'm referring to the Music Assistant Player Section: 

Image

But my understanding is, that at least in Music Assistant, the artist info and artwork should be visible in the "now playing" section?

I tried to use the feature "Try to inject metadata into stream (ICY)" in the different flavours, without luck unfortunately.

### How to reproduce

Play anything on a HEOS speaker through MA with default protocol HEOS.

### Music Providers

Local Files

### Player Providers

HEOS

### Full log output

[music-assistant.log](https://github.com/user-attachments/files/26718313/music-assistant.log)

Let me know if I need to provide more detailed logs from a different level and I will attach it.

2026.4.3

Home Assistant OS

- [x] Yes

Tommatheussen:
HEOS sends commands when it changes playing something. It looks like it also sends that command when the player state is changed (stopped, playing etc). 
When something is playing externally and we start playback from MA, there is an out of order sequence happening, causing the HEOS provider to think the player is playing something external (very limited check possible due to limited HEOS data available). This then marks the player as externally controller and we lose the internal MA metadata to be shown.

This PR makes sure we are not processing any of those commands when the player is not actually playing. We also update the now playing info when we change state, as the order of the events is not 100% guaranteed to be correct, causing the player is first signal it's now playing content has changed, and then signal it's playing.

Fixes https://github.com/music-assistant/support/issues/5311

### DIFF
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -11,6 +11,7 @@ from music_assistant_models.enums import MediaType, PlaybackState, PlayerFeature
 from music_assistant_models.errors import PlayerCommandFailed, SetupFailedError
 from music_assistant_models.player import DeviceInfo, PlayerSource
 from pyheos import Heos, HeosError, const
+from pyheos import PlayState as HeosPlayState
 
 from music_assistant.constants import (
     VERBOSE_LOG_LEVEL,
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):

MarvinSchenkel:
Nice one @Tommatheussen

PR #3930, opened 2026-05-20
Fix issues with M4B audiobooks

OzGav:
Two related issues affected single-file audiobooks (typically .m4b) in the local filesystem provider:

1. Embedded chapters were duplicated on every re-sync. ffprobe emits chapter start_time/end_time as JSON strings, which flowed unconverted into `MediaItemChapter.start/end`. After persistence and mashumaro deserialisation those fields came back as floats, so the dataclass __eq__ used by merge_lists in `MediaItemMetadata.update` considered the old and new chapter lists unequal and concatenated them.

2. Two standalone m4b files in the same folder caused the second one (alphabetically) to be dropped, because `_parse_audiobook` assumed any audiobook file without a track tag was part of a multi-file book and only kept the first file. A file carrying its own embedded chapter markers is by definition a complete audiobook, so skip the multi-file fallback in that case (and likewise don't treat a track>1 file with embedded chapters as a chapter file).

### DIFF
--- a/music_assistant/helpers/tags.py
+++ b/music_assistant/helpers/tags.py
@@ -555,8 +555,8 @@ class AudioTags:
                 chapters.append(
                     AudioTagsChapter(
                         chapter_id=chapter_data["id"],
-                        position_start=chapter_data["start_time"],
-                        position_end=chapter_data["end_time"],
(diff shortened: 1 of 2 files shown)


MarvinSchenkel (approved):
Thanks @OzGav

PR #3932, opened 2026-05-20
Fix HEOS queue cleanup slowing down other commands

Tommatheussen:
I've done a number of potential fixes in previous PR to the HEOS internal queuing logic, to fix the initial issue of 'ghost' items playing after MA's queue was finished.
Long story short, the latest fix works, but has the downside of, in some situations (mainly multiple wireless devices), blocking other commands, such as volume up/down while the queue operations are running.

This PR introduces a secondary connection to the main HEOS device (the HEOS specification also advises using 2 connections, 1 for event handling, 1 for sending commands) to operate all the queue commands. This secondary connection is only set up after the rest of the setup process for the HEOS provider is finished, so we know the default connection works for sure.
This secondary connection is only used for queue commands, all other commands and logic still flows through the main connection.
In the process I've also unified the scheduling checks, moved them all to the scheduling method where they belong.

### DIFF
--- a/music_assistant/providers/heos/provider.py
+++ b/music_assistant/providers/heos/provider.py
@@ -49,12 +49,19 @@ class HeosPlayerProvider(PlayerProvider):
         if ip_address := self.config.get_value(CONF_IP_ADDRESS):
             # Manual IP path
             ip_address = cast("str", ip_address)
-            await self._setup_controller(ip_address)
+            try:
+                await self._setup_controllers(ip_address)
+            except SetupFailedError:
+                self.logger.error(
+                    "Failed to set up HEOS controller at configured IP %s", ip_address
+                )
+                await self._disconnect_controllers()
+                raise
 
             # Explicitly discover players now
             await self.discover_players()
 
-    async def _setup_controller(self, controller_ip: str, connect_preferred: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


music_assistant/providers/heos/provider.py
```diff
@@ -96,6 +97,22 @@ async def _setup_controller(self, controller_ip: str, connect_preferred: bool =
...
+                    controller_ip,
+                    timeout=cast("int", self.config.get_value(CONF_TIMEOUT)),
+                    auto_reconnect=True,
+                    auto_failover=True,
+                    events=False,
+                )
+            )
+            await self._heos_queue.connect()
+        except HeosError as err:
+            self.logger.error("Failed to set up HEOS queue controller: %s", err)
```
MarvinSchenkel, line 120: Let's add explicit queue cleanup here in case `connect()` raises:
```
if self._heos_queue is not None:
        with suppress(Exception):
            await self._heos_queue.disconnect()
        self._heos_queue = None
```

MarvinSchenkel: I think you'd also need this in the exception handling of the `on_mdns_service_state_change` ?

Tommatheussen: Good call, I noticed we weren't actually cleaning up connections during the manual path either

MarvinSchenkel (approved):
Thanks @Tommatheussen !

PR #3924, opened 2026-05-20
Squeezelite: Honor per-player output_codec in multi-client sync URL

MarvinSchenkel:
## Summary

The multi-client sync URL in `SqueezelitePlayer.play_media` hardcoded `fmt=flac`, ignoring each player's configured `output_codec`. Classic Squeezebox firmware (Boom, Radio, Touch) can't decode FLAC in sync mode, so those devices played back silently when grouped.

This change reads the leader's `output_codec` and, when it is `flac`, walks the sync clients and downgrades to a non-FLAC codec if any member needs one — so classic hardware in a sync group gets audio it can actually decode.

Fixes [#5506](https://github.com/music-assistant/support/issues/5506)

## Test plan

- [ ] Group a classic Squeezebox (Boom / Radio / Touch) with a modern Squeezelite player and verify audio plays on both
- [ ] Group two FLAC-capable players and verify the URL still uses `fmt=flac`
- [ ] Single-player playback unaffected

---
_Generated by [Claude Code](https://claude.ai/code/session_01UqzeCqV4Z16gdbmSrD71Bf)_

PR #3933, opened 2026-05-20
Squeezelite: Honor per-child output codec in sync group URL

digodigo:
# Squeezelite: Honor per-child output codec in sync group URL

## Summary

Fixes silent audio on classic Logitech Squeezebox devices (Boom / Radio /
Touch, MAC-OUI `00:04:20`) when grouped for synchronised playback. The
multi-client sync URL was hardcoded to `fmt=flac`, so every child got handed
a FLAC stream regardless of its slimproto-advertised codec capabilities.
Classic Squeezeboxes accept the FLAC stream syntactically (STAT frames
advance, MA shows "playing"), but the hardware decoder produces no audible
output.

Each child's URL now appends a per-child `fmt={resolved}` value computed
from the child's own `output_codec` config + slimproto HELO capabilities:

1. Explicit `output_codec` config wins (symmetric with the solo-path resolver
   in `StreamsController.resolve_stream_url`).
2. Otherwise pick the highest-preference codec from the child's HELO
   capabilities: `flac > mp3 > wav`.
3. Otherwise fall back to `mp3` (slimproto baseline).

`MultiClientStream` already transcodes per child via a separate ffmpeg
process per HTTP subscriber, so no shared-stream/encoder reorganisation was
required. Sample rate and bit depth continue to flow through the existing
`get_output_format` per-child path.

## Behaviour change

- Sync groups where every member supports FLAC: **no change** — each child
  still receives FLAC.
- Sync groups containing at least one non-FLAC-capable member (the bug
  scenario): that child receives MP3, FLAC-capable members keep FLAC. Each
  child gets its best-case codec individually — no group-wide downgrade.
- Solo playback: untouched.

## Fixes

`music-assistant/support#5506`

## Tests

`tests/providers/squeezelite/test_player.py` adds 6 parametrised unit tests
for `_resolve_child_codec`:

- Config override wins over HELO (both for sensible and explicitly-against-advice
  choices, symmetric with the solo path).
- HELO picks FLAC when the player advertises it.
- HELO picks MP3 when FLAC isn't advertised (the Boom case — this is the
  silent-audio fix).
- Empty HELO + no config → MP3 baseline.
- Unknown HELO codec codes are ignored, baseline applies.

## Hardware verification

Verified on a live MA 2.8.7 Docker container with the fix backported and
deployed via [`apply-fork-patch.sh`](https://github.com/digodigo/ma-flac-sync-patch/blob/main/apply-fork-patch.sh)
(companion workaround repo). All four scenarios audible on the speakers,
no anomalies observed:

- ✅ Classic Squeezebox sync group (2× Boom or more): audible synchronised
  playback — the bug scenario, was silent before the fix.
- ✅ Mixed group (classic Squeezebox + Touch / modern Squeezelite): Boom
  gets MP3, Touch keeps FLAC — confirms per-child resolution.
- ✅ Modern-only sync group (2× Muse Luxe): FLAC retained, no regression.
- ✅ Solo playback on each device model: no regression from the call-site
  refactor.

Hardware in the verification fleet:
- 4× Squeezebox Boom (original firmware)
- 5× Squeezebox Radio (original firmware)
- 1× Squeezebox Touch (original firmware)
- 2× Muse Luxe (SqueezeESP32, philippe44 firmware)

## Notes for reviewers

- The branch was based on `upstream/dev` @ `50283614`. On that tip, seven
  unrelated tests already fail (`tests/controllers/streams/test_audio_analysis.py`,
  `tests/core/test_discovery.py`, `tests/core/test_helpers.py`,
  `tests/core/test_tags.py`) — they pre-existed this change and are
  untouched by it.
- Diagnostic background and a DEBUG-level "before" log are linked from the
  upstream issue.

### DIFF
--- a/music_assistant/providers/squeezelite/player.py
+++ b/music_assistant/providers/squeezelite/player.py
@@ -7,8 +7,8 @@ import statistics
 import struct
 import time
 from collections import deque
-from collections.abc import Iterator
-from typing import TYPE_CHECKING, cast
+from collections.abc import Iterable, Iterator
+from typing import TYPE_CHECKING, Final, cast
 
 from aioslimproto.models import EventType as SlimEventType
 from aioslimproto.models import PlayerState as SlimPlayerState
@@ -31,6 +31,7 @@ from music_assistant_models.media_items import AudioFormat
 from music_assistant.constants import (
     CONF_ENTRY_HTTP_PROFILE_FORCED_2,
     CONF_ENTRY_SYNC_ADJUST,
+    CONF_OUTPUT_CODEC,
     INTERNAL_PCM_FORMAT,
     VERBOSE_LOG_LEVEL,
     create_sample_rates_config_entry,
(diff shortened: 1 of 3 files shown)


MarvinSchenkel:
I had already opened a PR this morning with a fix for this one. Could you please test this one and see if it fixes your use case? https://github.com/music-assistant/server/pull/3924

PR #3945, opened 2026-05-22
Fix AirPlay receiver album artwork stuck after first track

MarvinSchenkel:
## Summary

Fixes #5514 (support repo): AirPlay Receiver plugin album artwork permanently stuck on the first track's image regardless of source, device, or session.

**Root cause:** The thumbnail cache in `helpers/images.py` is keyed on `sha256(provider + "/" + path)`. The `airplay_receiver` always used the static path `"cover_art"` for every `MediaItemImage`, so every track produced the same cache key. The first artwork was written to disk and served forever for all subsequent image requests. The `&t=timestamp` suffix that was appended to the outer URL had no effect because the imageproxy handler extracts only the `path` query parameter before calling `get_image_thumb`, discarding the timestamp before the cache lookup.

**Fix:** Embed a short MD5 content-hash of the image bytes in the path (e.g. `cover_art_a1b2c3d4`), identical to the approach used by `ariacast_receiver`. Each unique piece of artwork now gets its own cache entry. `resolve_image` is updated to accept any path starting with `"cover_art"`.

## Changes

- `music_assistant/providers/airplay_receiver/__init__.py`
  - Add `import hashlib`
  - `_update_cover_art`: compute `md5(cover_art_bytes)[:8]` and use `path=f"cover_art_{img_hash}"` — removes the ineffective `&t=timestamp` suffix
  - `resolve_image`: change `path == "cover_art"` → `path.startswith("cover_art")`

## Test plan

- [ ] Play track A via AirPlay → artwork for A appears
- [ ] Play track B (different album/artwork) → artwork updates to B
- [ ] Stop and start a new AirPlay session with different artwork → artwork updates correctly
- [ ] Two tracks with identical artwork (same album) share one cache entry (same MD5 hash → same path)

---
_Generated by [Claude Code](https://claude.ai/code/session_01JtqsKeQkbKdn47Mj4e1rNL)_


MarvinSchenkel:
Seting to Draft, this might already be fixed with the plugin refactor

PR #3952, opened 2026-05-23
Resolve universal_player wrappers in UGP stream handler

OzGav:
Fixes https://github.com/music-assistant/support/issues/5517

PR #3956, opened 2026-05-23
Revert "Resolve universal_player wrappers in UGP stream handler"

OzGav:
Reverts music-assistant/server#3952


MarvinSchenkel (approved):
Automated review: clean revert of #3952, restoring prior working behavior in the UGP stream handler. All CI checks pass and the change is scoped to a single provider file. Thanks @OzGav :tada:

PR #3965, opened 2026-05-24
Use Protocol-bounded TypeVar for @use_cache decorator

jdaberkow:

The `@use_cache` decorator's `ProviderT` TypeVar is bounded to `Provider | CoreController`, but several providers use the decorator on helper/manager classes that aren't subclasses of either. This forces every call site to add `# type: ignore[type-var]`. 26 suppressions in the upcoming Deezer provider (#3900) alone.

The decorator only accesses `self.mass` and `self.domain` (via `getattr(self, "instance_id", self.domain)`), so the type bound is stricter than necessary.

This PR replaces the union bound with a `_Cacheable` Protocol that captures exactly what `@use_cache` needs:

```python
class _Cacheable(Protocol):
    @property
    def domain(self) -> str: ...

    @property
    def mass(self) -> MusicAssistant: ...
```

Using `@property` in the Protocol means read-only properties, class attributes, and instance attributes all satisfy the constraint.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/cache/helpers.py
+++ b/music_assistant/controllers/cache/helpers.py
@@ -4,7 +4,16 @@ from __future__ import annotations
 
 import functools
 from collections.abc import Awaitable, Callable, Coroutine
-from typing import TYPE_CHECKING, Any, Concatenate, ParamSpec, Protocol, TypeVar, cast, get_type_hints
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Concatenate,
+    ParamSpec,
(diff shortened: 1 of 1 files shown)


jdaberkow:
Did the force push since a rebase was necessary due to the split-up of `cache.py`.

marcelveldt (approved):
LGTM, thanks @jdaberkow !

marcelveldt:
CI task seems to be hanging. I'll merge anyways as the code looks good.

PR #3972, opened 2026-05-25
Add Wikipedia provider and associated plumbing

OzGav:

A new Wikipedia metadata provider that fetches artist biographies in the user's preferred language, falling back to English. It finds the right Wikipedia article via the artist's MusicBrainz links (and Wikidata where MusicBrainz doesn't list one), then pulls the article's intro text.

As there are now two biography-supplying metadata providers it is necessary to populate the database with the best one and in the right language. So this also:

- Updates TheAudioDB to report which language its bio is in. Previously it returned text with no language label; now it tags it (e.g. en, nl) so the system can compare it against the user's preference.
- Updates Qobuz the same way. Its API returns a language code with the bio, so we now tag it rather than guess.
- Adds bio-selection logic to the metadata controller. When several providers can supply a bio, it now picks the best one by a clear rule:
  - prefer the first bio in the user's language, in provider priority order (music, then TheAudioDB, then Wikipedia);
  - fall back to English only if no provider has the user's language and we don't already have a bio in that language stored;
  - a music provider's bio counts as the user's language or English only when the provider reports the language with certainty; if it leaves the language unknown, the bio is used only as a last resort;
  - a bio already stored in the user's language is never replaced by one not in the users preferred language;
  - the bio refreshes on each metadata update, so edits to the source article are picked up over time.
- Adds unit tests covering each of those selection outcomes, so the behaviour can't silently break.  
  
## Types of changes

## Checklist


marcelveldt:
Also don't forget about a Music provider itself providing a bio/description. For example Qobuz provides (localized) artist bio's

music_assistant/providers/theaudiodb/__init__.py
```diff
@@ -83,6 +83,25 @@
 CONF_ENABLE_ALBUM_METADATA = "enable_album_metadata"
 CONF_ENABLE_TRACK_METADATA = "enable_track_metadata"
 
+# TheAudioDB field suffix -> ISO 639-1 language code. CN/JP/SE/NO/IL use country-style
+# codes that don't match the ISO language code, so the mapping is explicit.
+TADB_SUFFIX_TO_ISO: dict[str, str] = {
```
MarvinSchenkel, line 88: Is `EN` supposed to be missing here?

OzGav: Yes because it uses an unsuffixed field for English `strBiography` vs `strBiographyNL`, for example.

OzGav:
I wrestled with this. I wasn't sure if a music provider returns a language code with the bio text so we know for certain what language it's in. I just checked Qobuz and it does, so I've changed Qobuz to populate the language field and adjusted the selection to favour music provider bios.

The order is now: first bio in the user's preferred language wins (music, then TADB, then Wikipedia), otherwise keep an existing preferred language bio, otherwise fall back to English in that same order, and only use an unknown language bio as a last resort.

I'll ask the other devs who populate artist.description to populate the language field too, but only when they know it for certain. If they leave it as None, that bio won't count as preferred language or English, so anything with a confident match will be used ahead of it.

MarvinSchenkel (approved):
Thanks @OzGav !

PR #3980, opened 2026-05-25
Avoid syncing native parent volume to AirPlay protocols

### LINKED ISSUE #5529: Sendspin/AirPlay track changes reset Yamaha MusicCast parent volume
### The problem

When using a Yamaha MusicCast receiver through Music Assistant with Sendspin/AirPlay as the preferred output protocol, advancing to the next track changes the parent MusicCast player's volume unexpectedly.

In this setup, the Music Assistant player `Living Room` is a Yamaha RX-V6A MusicCast player whose preferred output protocol is the Sendspin bridge for its AirPlay endpoint. On a normal `next` command, the Living Room volume dropped from 36% to 26% without any explicit volume command from Home Assistant or the user.

This looks like an AirPlay/Sendspin volume scale mismatch with the linked MusicCast parent:

- Music Assistant restarts the Sendspin/AirPlay bridge on track change.
- The bridge starts `cliraop` with `-volume 36`, copied from the parent MusicCast player's Music Assistant volume.
- The Yamaha AirPlay endpoint maps that to native MusicCast raw volume `43` out of `161`.
- Music Assistant's MusicCast provider reads that back as `int(43 / 161 * 100) == 26`.
- The parent player is then exposed as 26% volume.

Expected behavior: changing tracks through the Music Assistant queue should not change the native MusicCast parent player's volume when no explicit volume command was sent.

### How to reproduce

1. Configure a Yamaha MusicCast player in Music Assistant.
2. Link its AirPlay endpoint and use Sendspin as the preferred output protocol.
3. Start playback on the MusicCast parent player through the Music Assistant queue.
4. Set the parent player volume to 36%.
5. Send `player_queues/next` for the parent player's queue.
6. Observe the parent player volume change to 26% shortly after the AirPlay bridge restarts.

Controlled test data before the `next` command:

```json
{
  "player_id": "C4E5323097D0___main",
  "name": "Living Room",
  "provider": "musiccast",
  "playback_state": "playing",
  "volume_level": 36,
  "group_members": [
    "C4E5323097D0___main",
    "A8A237EA732E___main",
    "C4E5323097D0___zone2"
  ]
}
```

After the controlled `next` command:

```json
{
  "player_id": "C4E5323097D0___main",
  "name": "Living Room",
  "provider": "musiccast",
  "playback_state": "playing",
  "volume_level": 26,
  "group_members": [
    "C4E5323097D0___main",
    "A8A237EA732E___main",
    "C4E5323097D0___zone2"
  ]
}
```

Home Assistant recorder rows around the test window show the same transition with no Home Assistant context/user id on the volume-changing state update:

```json
{"ts":"2026-05-25 12:54:36","volume_level":0.36,"source":"Music Assistant Queue","media_title":"One Last Game"}
{"ts":"2026-05-25 12:54:38","volume_level":0.36,"source":"Music Assistant Queue","media_title":"One Last Game"}
{"ts":"2026-05-25 12:54:40","volume_level":0.26,"source":"Music Assistant Queue","media_title":"One Last Game"}
```

### Music Providers

- Plex
- Built-in radio provider is also configured, but this reproduction used local/library queue playback from Plex.

### Player Providers

- MusicCast
- AirPlay
- Sendspin

Environment:

- Music Assistant server: `2.8.8`
- Home Assistant: `2026.5.4`
- Receiver: Yamaha RX-V6A, firmware/software `1.77`
- Music Assistant native player provider: `musiccast`
- Preferred output protocol for the MusicCast player: Sendspin bridge
- Linked protocol ids include:
  - native MusicCast: `C4E5323097D0___main`
  - AirPlay: `apc4e5323097d1`
  - Sendspin: `spb_c4e5323097d1`

Relevant Music Assistant player settings:

```json
{
  "player_id": "C4E5323097D0___main",
  "provider": "musiccast",
  "values": {
    "preferred_output_protocol": "spb_c4e5323097d1",
    "linked_protocol_ids": [
      "uuid:9ab0c000-f668-11de-9976-c4e5323097d0",
      "apc4e5323097d1",
      "spb_c4e5323097d1"
    ]
  }
}
```

### Full log output

```text
2026-05-25 12:54:37.714 DEBUG [music_assistant.player_queues] (pre)loading (next) item for queue Living Room...
2026-05-25 12:54:37.719 DEBUG [music_assistant.streams.audio] Getting streamdetails for library://track/373
2026-05-25 12:54:37.726 DEBUG [music_assistant.streams.audio] Retrieved streamdetails for library://track/373 in 7 milliseconds
2026-05-25 12:54:37.726 DEBUG [music_assistant.audio_buffer] get_buffer[prepare]: Reusing buffer for plex--a2DdCPbG://track//library/metadata/16712 - available: 80s, seek_ms: 0, discarded: 0
2026-05-25 12:54:37.726 DEBUG [music_assistant.players] Handling command play_media for player Living Room (by user jyundt)
2026-05-25 12:54:37.728 DEBUG [music_assistant.players] Starting playback on Living Room via protocol Sendspin (target=Living Room (AirPlay)), group_members=[]
2026-05-25 12:54:37.728 DEBUG [music_assistant.Sendspin.spb_c4e5323097d1] Received PLAY_MEDIA command on player Living Room (AirPlay) with uri library://track/373
2026-05-25 12:54:37.728 DEBUG [music_assistant.Sendspin.spb_c4e5323097d1] Cancelling playback task (new media requested)
2026-05-25 12:54:37.731 DEBUG [music_assistant] BridgePlayerRole stream ended for client spb_c4e5323097d1
2026-05-25 12:54:37.738 DEBUG [music_assistant.AirPlay.bridge.apc4e5323097d1] Sendspin stream start request for Living Room (reason=ConnectionReason.PLAYBACK)
2026-05-25 12:54:37.743 INFO [music_assistant.streams.audio] Start Queue Flow stream for Queue Living Room - crossfade: disabled
2026-05-25 12:54:37.743 DEBUG [music_assistant.streams.audio] Start Streaming queue track: plex--a2DdCPbG://track//library/metadata/16712 (C418 - One Last Game) for queue Living Room
2026-05-25 12:54:37.751 DEBUG [music_assistant.helpers.process.cliraop] Killing process cliraop with PID 90
2026-05-25 12:54:37.782 DEBUG [music_assistant] BridgePlayerRole stream started for client spb_c4e5323097d1
2026-05-25 12:54:37.783 INFO [music_assistant.AirPlay.bridge.apc4e5323097d1] Bridge writer started for Living Room, awaiting first chunk
2026-05-25 12:54:37.824 DEBUG [music_assistant.AirPlay] Starting cliraop process for player apc4e5323097d1 with args: ['/app/venv/lib/python3.13/site-packages/music_assistant/providers/airplay/bin/cliraop-linux-x86_64', '-ntpstart', '17131346706292868359', '-port', '7000', '-latency', '1000', '-volume', '36', '-if', '192.168.67.12', '-encrypt', '-alac', '-et', '0,4', '-md', '0,1,2', '-am', 'RX-V6A', '-pk', 'a6a7c37e3b044d65fb07ca1ce7aec33f14c5f543213d5e97305e2a15e118d99d', '-debug', '5', '-dacp', 'B6ECCB14FBD249FF', '-activeremote', '842045393', '-cmdpipe', '/tmp/1-apc4e5323097d1-842045393-cmd', '-udn', 'C4E5323097D1@Living Room._raop._tcp.local.', '192.168.67.19', '-']
2026-05-25 12:54:37.832 DEBUG [music_assistant.AirPlay.bridge.apc4e5323097d1] Trimming 1869 frames from first chunk for Living Room
2026-05-25 12:54:37.847 INFO [music_assistant.AirPlay.bridge.apc4e5323097d1] Bridge protocol started for Living Room (NTP=17131346706292868359, lookahead=2500ms)
2026-05-25 12:54:38.370 INFO [music_assistant.AirPlay.bridge.apc4e5323097d1] AirPlay connection established for Living Room
2026-05-25 12:54:38.419 DEBUG [music_assistant.AirPlay] DACP request for Living Room (842045393): /ctrl-int/1/setproperty?dmcp.device-volume=-15.600000 --
2026-05-25 12:54:39.375 DEBUG [music_assistant.AirPlay] DACP request for Living Room (842045393): /ctrl-int/1/setproperty?dmcp.device-volume=-15.600000 --
```

Immediately after that, the Yamaha receiver's own MusicCast API reports raw volume 43:

```json
{
  "response_code": 0,
  "power": "on",
  "volume": 43,
  "mute": false,
  "max_volume": 161,
  "input": "airplay",
  "input_text": "AirPlay",
  "sound_program": "all_ch_stereo",
  "actual_volume": {
    "mode": "numeric",
    "value": 21.5,
    "unit": ""
  }
}
```

### Additional information

The suspected code path is that the AirPlay/Sendspin path syncs volume from the protocol parent before starting the stream:

```python
# providers/airplay/sendspin_bridge.py
self.airplay_player.sync_volume_level()
```

and:

```python
# providers/airplay/player.py
self._attr_volume_level = parent_player.state.volume_level
self.mass.config.set_raw_player_config_value(self.player_id, CONF_STORED_VOLUME, self._attr_volume_level)
```

The bridge then starts `cliraop` with that value as AirPlay volume. For this Yamaha receiver, AirPlay/RAOP volume and MusicCast native volume do not appear to use the same effective scale. So a parent MusicCast volume of 36% becomes Yamaha native raw volume 43/161, which Music Assistant correctly reads back as 26%.

Possible fixes might be one of:

- Avoid syncing the parent MusicCast volume into the linked AirPlay/Sendspin protocol volume when starting/restarting a stream.
- Preserve the parent MusicCast volume and reapply/restore it after AirPlay startup if the AirPlay endpoint changes the device volume.
- Add a config option to disable parent-to-AirPlay volume sync for linked protocol players where volume scales are known to differ.
- Treat `ignore_volume` or a similar setting as also preventing startup volume reassertion, not only DACP feedback handling.

jyundt:

Avoid copying a protocol-linked parent player's volume into an AirPlay protocol player when the parent has native volume control.

For native receivers/amplifiers, the parent volume represents device volume, while AirPlay/RAOP volume is stream/protocol volume. Some devices, including Yamaha MusicCast receivers, map those scales differently. Reusing the native parent volume as AirPlay startup volume can cause track changes to unexpectedly alter the receiver volume.

- fixes music-assistant/support#5529

## Types of changes

## Testing

Verified with:

- `python3 -m py_compile music_assistant/providers/airplay/player.py tests/providers/airplay/test_player.py`
- `git diff --check`
- Python 3.14 container: `SKIP=no-commit-to-branch pre-commit run --files music_assistant/providers/airplay/player.py tests/providers/airplay/test_player.py`
- Python 3.14 container: `pytest -o addopts="" tests/providers/airplay/test_player.py`


MarvinSchenkel (dismissed):
Automated review (no CRITICAL/PROBLEM findings).

The guard in sync_volume_level cleanly addresses the bug: a parent with native volume control reports device/amplifier volume (different scale than AirPlay/RAOP protocol volume), so copying it onto the AirPlay protocol volume on each track change caused unexpected receiver volume changes on devices like Yamaha MusicCast. Docstring and inline comment explain the why, and the two new tests cover both branches.

Thanks @jyundt :tada:

jyundt:
I updated my PR message to include the missing PR label and included the test fixes in 22b22a5.

@MarvinSchenkel could this also be backported to 2.8.x?

MarvinSchenkel:
@marcelveldt could you sanity check this? 


I remember you mentioned airplay forces us to send a volume level when the stream starts. 

I think this could still result in volume jumps when the receiver uses a different scale. Did you test this with slipping tracks or just playing a different track when something is already playing?

jyundt:
> I think this could still result in volume jumps when the receiver uses a different scale. Did you test this with slipping tracks or just playing a different track when something is already playing?

I reproduced this behavior by slipping (scrubbing?) through an actively playing track and manually advancing to the next track in the queue.

marcelveldt:
Yes, AirPlay REQUIRES to send the volume, that is part of the protocol. 
So you actually need to sync the volume from the parent player to AirPlay to prevent volume spikes, otherwise the volume will jump to the last known airplay volume.

I also notice that the volume differs on some devices, e.g. Sonos, so this most likely means that our scaling just is flawed and we need to fix that instead.

This PR is not the right fix as it will fix issue A but causes issue B instead.

jyundt:
> Yes, AirPlay REQUIRES to send the volume, that is part of the protocol. 
> So you actually need to sync the volume from the parent player to AirPlay to prevent volume spikes, otherwise the volume will jump to the last known airplay volume.
> 
> I also notice that the volume differs on some devices, e.g. Sonos, so this most likely means that our scaling just is flawed and we need to fix that instead.
> 
> This PR is not the right fix as it will fix issue A but causes issue B instead.
> 

Ah got it, thanks for the clarification. Lemme take another swing at this.

MarvinSchenkel:

jyundt:
@MarvinSchenkel @marcelveldt I threw some Codex tokens at this with 27e48d0. I tested with my MusicCast devices and everything appears to be working as expected.

marcelveldt:
thanks for the detailed investigation here @jyundt, the root cause analysis in the linked issue is spot on 👍

I'm a bit hesitant about the `volume_map` approach though, it feels like it's doing more work than needed and I'd rather not introduce a learned calibration table for this.

The thing is: for RAOP we always have to start cliraop with a `-volume`, so we can't just skip sending one. But the value we want isn't the parent's native volume - that's a different scale, which is exactly what causes the drift. It's the volume the device itself reports back to us over DACP (`dmcp.device-volume=`). That one is already on the airplay scale, so re-sending it on a stream restart is basically a no-op and the receiver stays put. In other words the device already tells us the right number, so we don't need to calculate/translate it.

So I think this can be reduced to:

- keep the DACP volume feedback enabled for players with a native parent, so the child keeps mirroring the device's own volume into `stored_volume` like it already does today
- in `sync_volume_level`, for a native parent, don't overwrite the child with the parent volume but just keep the child's own (stored) value

That drops the need for `CONF_VOLUME_MAP`, the seeding logic and the linear translation entirely.

The only thing we can't fully avoid is a cold start (or a volume change while nothing is streaming), since RAOP doesn't let us read the device volume before opening a session - so the first stream nudges the volume once and then the feedback corrects + stores it. That's a one-time, self-correcting nudge instead of drifting on every track change.

I'll set this to draft for now, let me know what you think and happy to help if anything's unclear!

jyundt:
@marcelveldt thanks for the feedback. I took another stab at this with [7618ff2](https://github.com/music-assistant/server/pull/3980/commits/7618ff26f760c5cd0bbdf887156348274ac67237). 

I tested with a local build and it worked as expected.

marcelveldt:
this is exactly what I had in mind, much cleaner now 👍

only thing left is the `lint` check failing on `check_method_order`: the new `_has_native_protocol_parent` property needs to move down to the bottom of the class (we keep private methods at the bottom, public on top). Once that's green I'm happy to merge.

jyundt:
Ah got it, fixed with 9ff511a2117edf513e2a44f18325af758554ce72

marcelveldt (approved):
Nice work, @jyundt !

PR #3997, opened 2026-05-26
Scale background audio-analysis timeout to track duration

chrisuthe:

The background-scan audio analysis used a fixed 300-second per-track
budget. Long mix tracks (hour-plus DJ sets, long podcasts) consistently
exceeded it, so analysis re-timed-out on every nightly run and never
completed — wasting cycles on the same files indefinitely.

This derives the per-track timeout from the known track duration:
`int(streamdetails.duration * 1.5)`. When duration is unknown or zero,
behaviour falls back to the existing 300-second constant. The same
`timeout_seconds` value flows into the `asyncio.wait_for` call, the
warning log, and the task-failure message, so the recorded budget
matches what was actually waited for.

Only the background path (`_run_background_streaming_for_track`) is
affected. Real-time playback analysis (`start_analysis`) consumes
chunks as the track plays, so wall-clock is already bounded by the
track itself — no timeout protection is applied there.

- https://github.com/orgs/music-assistant/discussions/5520

## Types of changes

## Checklist


MarvinSchenkel (approved):
Thanks @chrisuthe !

PR #3999, opened 2026-05-27
Skip non-music providers in library update callback dispatch

dmoo500:
## Summary
When updating a library item, `MediaControllerBase.update_item_in_library` iterates over all provider mappings and calls `on_item_updated` unconditionally after a cast.

Some mappings can point to plugin providers (for example smart playlist). Plugin providers do not implement `on_item_updated`, which can raise an `AttributeError` during Refresh Item / Update Metadata flows.

This change dispatches the callback only to actual `MusicProvider` instances.

## Changes
- Guard callback dispatch in `music_assistant/controllers/media/base.py` with `isinstance(provider, MusicProvider)`.
- Add regression test in `tests/test_library_sync.py` that verifies non-music providers are skipped.

## Validation
- `pytest tests/test_library_sync.py -k skips_non_music_providers` passed.
- `pre-commit run --all-files` passed except an existing unrelated repo-level mypy error in `music_assistant/providers/fastmcp_server/auth.py`.

## Context
- Checked existing PRs first; no open PR currently addresses this specific callback-dispatch bug.

### DIFF
--- a/music_assistant/controllers/media/base.py
+++ b/music_assistant/controllers/media/base.py
@@ -37,12 +37,12 @@ from music_assistant.helpers.compare import compare_media_item, create_safe_stri
 from music_assistant.helpers.database import UNSET
 from music_assistant.helpers.json import json_loads, serialize_to_json
 from music_assistant.helpers.util import guard_single_request, parse_optional_bool
+from music_assistant.models.music_provider import MusicProvider
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, Mapping
 
     from music_assistant import MusicAssistant
-    from music_assistant.models.music_provider import MusicProvider
     from music_assistant.models.plugin import PluginProvider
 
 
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/media/base.py
```diff
@@ -208,7 +208,8 @@ async def update_item_in_library(
         # notify music providers of the update so they can sync their own storage
         for prov_mapping in library_item.provider_mappings:
             if provider := self.mass.get_provider(prov_mapping.provider_instance):
-                provider = cast("MusicProvider", provider)
+                if not isinstance(provider, MusicProvider):
```
MarvinSchenkel: Let's use `provider.type != ProviderType.MUSIC` instead of an expensive `isinstance` check

MarvinSchenkel (changes requested):

music_assistant/controllers/media/base.py
dmoo500: done

MarvinSchenkel (approved):
Thanks @dmoo500

PR #4007, opened 2026-05-27
fastMCP Server: sync 0.3.20→0.3.33 (security, fixes, tests)

trudenboy:

This PR catches the upstream `fastmcp_server` provider up from `v0.3.20`
to `v0.3.33`, bundling twelve provider releases. The changes group into
four themes.

**Source**: [trudenboy/ma-provider-mcp](https://github.com/trudenboy/ma-provider-mcp) · tag [v0.3.33](https://github.com/trudenboy/ma-provider-mcp/releases/tag/v0.3.33)

### Security (v0.3.24, v0.3.26)

- **Wizard bootstrap moved from URL query string to URL fragment.**
  Tokens no longer ride in access logs, reverse-proxy logs (HA ingress
  / nginx), or cross-origin `Referer` headers.
- **Wizard HTML response carries `Referrer-Policy: no-referrer` and a
  strict `Content-Security-Policy`** (`default-src 'none'` plus
  narrowly-scoped `script-src` / `style-src` / `connect-src` /
  `img-src` directives) so a future inline-data edit accidentally
  introducing XSS cannot exfiltrate long-lived MA tokens.
- **Bearer-token audience is verified before Music Assistant is
  consulted.** Previously every request hit `authenticate_with_token`
  first, which refreshed MA's sliding-window expiry even when the
  audience was about to fail — an attacker holding a non-MCP MA token
  could keep it alive indefinitely by polling the MCP endpoint.
- **Connect-Wizard credential endpoints refuse plaintext HTTP from
  non-loopback hosts.** `/connect/login`, `/connect/exchange` and
  `/connect/token` carry secrets that were sniffable over plain HTTP
  from a LAN host; HTTPS and loopback are still accepted.
- **`play_announcement` validates the supplied URL scheme** and clamps
  `volume_level` to `[0, 100]`. The URL flowed to MA's player API
  unchecked, so a prompt-injected `file:///etc/passwd` or `data:` URL
  would have been fetched by the audio backend.

### Reliability fixes (v0.3.22, v0.3.23, v0.3.25, v0.3.29, v0.3.30)

- **Permission hot-swap could race the previous ASGI lifespan
  shutdown.** The unmount path now awaits shutdown to completion
  before returning, so a follow-up `apply_permission_change` no longer
  leaks background tasks or triggers `Task was destroyed but it is
  pending` warnings.
- **A failed mount no longer leaves the runtime half-mounted.** If
  `MCPServerRuntime.start` raises after a partial mount, the rollback
  tears the in-progress state down before re-raising so a retry
  starts from a clean slate.
- **Connect Wizard rejected the Home-Assistant ingress flow.** The
  plaintext guard introduced in v0.3.26 didn't recognise HA ingress
  (TLS terminates at HA's front door and is forwarded over a *local*
  socket, so `request.scheme == "http"` for a non-loopback host); the
  guard now trusts requests MA's `is_request_from_ingress` helper
  confirms, while still refusing direct LAN plaintext.
- **`playlists/add_track`, `add_tracks` and `remove_tracks` accept
  both the integer library id and the `library://playlist/<n>` URI**
  for `playlist_id`, normalising internally to the integer form MA
  requires. Previously passing the URI raised a Python `ValueError`
  inside MA.
- **`playlists/add_tracks` no longer falsely claims atomicity for
  small batches** — MA performs no atomicity at any batch size. The
  tool now always adds tracks one at a time with progress reporting
  and an explicit "not atomic" notice.
- **`get_track_by_uri` and `get_lyrics` reject non-track URIs** with a
  clean `ToolError` instead of silently coercing an album/playlist
  into a garbage `TrackBrief` (or returning `None` for non-track
  lyrics queries).
- **`list_players` hides offline devices by default and reports
  `state="unavailable"` when they are explicitly requested via
  `include_unavailable=True`.** MA never receives push updates from a
  device it has lost contact with, so the cached `state` stayed at
  `"idle"` and the brief looked identical to a working speaker that
  simply wasn't playing anything.

### Enhancements

- **Sphinx-style docstrings with `:param:` blocks across all tools** —
  every parameter previously surfaced to clients with an empty
  description in the JSON Schema. The updated docstrings give the LLM
  the constraints it needs to construct correct calls and pick the
  right tool out of similar siblings (search vs. list, favorites vs.
  library, `play_media` vs. `play_index`, volume_set vs.
  volume_up/down vs. volume_mute).
- **`play_media` and `transfer_queue` are now flagged as destructive
  operations** in their tool annotations (queue replacement / source
  drain), and the `play_announcement` timeout is raised from `15s` to
  `30s` to match audio-fetch latency on slow providers.
- **Search and recents tools clamp caller-supplied `limit`** to the
  same `[1, 200]` envelope the rest of the library tools enforce —
  previously `search_tracks` / `search_albums` / `search_artists` /
  `recently_added_tracks` / `recently_played` were unbounded.
- **Media-tool errors are more actionable.** `_resolve_uri` surfaces
  a distinct `ToolError` per failure class (`MediaNotFoundError` /
  `InvalidProviderURI` / `ProviderUnavailableError`) instead of
  flattening every failure to "Item not found"; the LLM caller can
  now distinguish "URI typo" from "provider offline".
- **`QueueBrief.item_count` is now `int | None`** — `None` when the
  upstream queue exposes no canonical total, rather than the
  previously silent `0` on a non-empty queue.
- **Origin-allowlist helpers split into a dedicated `origins.py`
  module**, with prior `http_bridge` symbols re-exported under their
  historical names for back-compat. Two new contract tests pin both
  the public module names and the back-compat aliases.
- **External-base-URL detection prefers a public
  `client.authenticated_user` attribute** when available, falling
  back to the underscore-prefixed internal form MA exposes today — a
  future MA rename to a public property is then picked up
  transparently.

### Player & queue surface — usability triage (v0.3.30, v0.3.32)

The `list_players` response previously made every device look
interchangeable to an LLM; v0.3.30 + v0.3.32 close that gap.

- **`list_players` and `get_player` expose `available`, `enabled`,
  `needs_setup`, `active_group`, `synced_to`** in every player brief,
  mirroring the same flags Music Assistant carries on its `Player`
  model. The previous response shape is a strict subset, so existing
  callers keep working.
- **`list_players` accepts `include_unavailable` and `include_disabled`
  parameters** (both default `False`, matching MA's own
  `return_disabled` default). Without the flags, offline and
  admin-disabled players are hidden upstream by MA's controller — the
  tool just forwards the flags rather than re-implementing the
  filter.
- **The `state` field is now a usability summary, not just a playback
  state.** `to_brief_player` evaluates a four-rung priority ladder —
  `unavailable` > `disabled` > `needs_setup` > `synced` > the raw
  `playback_state` — so a single field tells the LLM everything it
  needs to triage a device. The new values are additive: any caller
  doing equality checks against `idle` / `playing` / `paused` keeps
  working as before. A sync follower (whose audio is routed through
  the group's stream but whose own queue stays idle) now surfaces
  `state="synced"` instead of looking like a quiet speaker.
- **`QueueBrief` gains `available`**, mirroring `PlayerQueue.available`
  upstream and closing the parallel triage problem for queues.
- **`now_playing_summary` prompt updated** to reflect both new
  filters (`include_unavailable=True` for offline devices,
  `include_disabled=True` for admin-disabled devices) and to explain
  the `state="synced"` queue-routing semantic.

### Test coverage

- **Six tests that silently passed for the wrong reason are
  tightened** — most notably the log-handler wiring (the previous
  version no-op'd via a non-existent `Client` method), the
  `apply_permission_change` hot-swap assertion (the preset tag was
  also in defaults, so the test held whether the rebuild ran or
  not), and the `to_brief_player` powered-source check (now
  parametrised with a contradictory case so a precedence swap is
  caught).
- **Five new test files** add coverage for `provider/prompts.py`
  (was zero), `MCPServerProvider.update_config` (was zero), every
  malformed-JWT path in `_extract_jwt_audience` (now nine
  parametrised cases), all five library resource kinds plus the
  `queue_resource` (was uncovered), and an end-to-end hot-swap test
  that mounts a real FastMCP root and verifies the visible tool
  surface flips on `apply_permission_change`.
- **Test count climbs from 194 → 322** across the bundled releases.

- N/A

## Types of changes

## Checklist


MarvinSchenkel (approved):
Thanks @trudenboy

PR #4001, opened 2026-05-27
Fix Spotify Connect playback control reliability and error reporting

marcelveldt:
# What does this implement/fix?

Spotify Connect (exposed as an AudioSource) had several control and UX issues once a session was active or handed between devices: control silently broke after deselecting Music Assistant in the Spotify app, resuming from MA was unreliable, "MA is not the active device" surfaced as a misleading "no more tracks available" error, and pausing/stopping a live source cleared the queue so it could not be resumed.

## Changes

- Keep the matched Spotify music provider on `session_disconnected` so Web API control survives the Spotify app dropping the device.
- Prefer `PUT /me/player/play?device_id=…` for reliable resume, with transfer-with-play as fallback.
- Track whether MA is the active Spotify Connect device and fail fast in `get_stream_details` with a user-friendly message when it isn't.
- Set `can_initiate=False` on the AudioSource — cold-start from MA can't be made reliable via the Web API (it needs an existing Spotify playback context); reliable entry stays external (pick MA in the Spotify app).
- Surface provider `AudioError` messages verbatim through the queue and stream layers instead of flattening them to "no more tracks available".
- Don't clear the queue when a live source (radio / AudioSource) stops, so playback can be resumed.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pre-commit run --all-files` passes.
- [ ] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.
- [ ] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.

PR #4010, opened 2026-05-28
Spotify Connect: clearer transport errors and automatic stall recovery

marcelveldt:

Three small follow-ups to #4001 that make Spotify Connect playback control feel less fragile:

- **Friendlier errors for transport commands.** When MA is no longer the active Spotify device, next/previous/seek/volume now surface the same "select Music Assistant in the Spotify app" hint as play. Previously only play caught this — the others fell through to a generic 403.
- **Self-recovery from pipe stalls.** Occasionally librespot would start producing audio before MA attached a consumer; the FIFO would fill, librespot's write would block, and the device looked frozen until Spotify timed out (~14 s). A short watchdog now drains the FIFO once and lets librespot proceed on its own.
- **Codec-independent readiness sentinel.** The "librespot is up" check no longer keys on a PCM-specific log line, so a future backend or codec change doesn't silently break startup detection.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/streams/ogg_handler.py
+++ b/music_assistant/controllers/streams/ogg_handler.py
@@ -404,47 +404,67 @@ def _resync_ogg_buffer(buffer: bytearray) -> int:
 _MAX_BUFFER_SIZE = 65536
 
 
+async def chain_ogg_pages(
+    source: AsyncGenerator[bytes, None],
+    metadata_callback: Callable[[dict[str, str]], Any] | None = None,
+) -> AsyncGenerator[bytes, None]:
+    """
+    Yield continuous OGG data from a chained byte source, stitching chain boundaries.
+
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
Automated review (no CRITICAL/PROBLEM findings).

Verified:
- AudioFormat OGG/VORBIS + bit_rate=320 matches the new `--passthrough` librespot mode.
- Active-device guards in `on_source_control` / `on_volume_change` mirror the existing check in `get_stream_details`.
- Removed `--dither` and `--enable-volume-normalisation` are decode-only flags, correctly dropped under passthrough.

Thanks @marcelveldt :tada:

Note: the PR Labels workflow failed because the `bugfix` checkbox uses escaped backticks (\`bugfix\`) instead of plain backticks — the label regex won't match, so the bugfix label won't auto-apply. Worth fixing in the description or applying the label manually.

PR #4015, opened 2026-05-28
Adapt core methods to allow providing a username for item retrieval

fmunkes:
This moves a method currently in the HA integration to the server, and adds support for user-based filtering.

~Still to do: verification of library uris to ensure accessibility by a user if user id is given.~ done

## Types of changes

## Checklist


music_assistant/controllers/media/albums.py
```diff
@@ -129,6 +132,11 @@ async def library_items(
         :param album_types: Filter by album types.
         :param genre: Filter by genre id(s).
         """
+        user: User | None = None
+        if username_or_user_id:
+            # anything non-web cannot use the web context in _ensure_provider_filter
```
marcelveldt: why not?

music_assistant/controllers/media/albums.py
```diff
@@ -129,6 +132,11 @@ async def library_items(
         :param album_types: Filter by album types.
         :param genre: Filter by genre id(s).
         """
```
marcelveldt, line 137: Don't forget to also extend the docstrings

music_assistant/controllers/music.py
```diff
@@ -318,13 +313,18 @@ async def search(
...
         """Perform global search for media items on all providers.
 
         :param search_query: Search query.
         :param media_types: A list of media_types to include.
         :param limit: number of items to return in the search (per type).
+        :param username_or_user_id: get user from either id or name for provider filter
         """
+        user: User | None = None
+        if username_or_user_id:
+            user = await self.mass.webserver.auth.get_user_by_id_or_name(username_or_user_id)
```
marcelveldt: so there is one flaw in this logic, which is that a limited user can now request data from another one (that has provider filters set).

You always need to check if the user requesting the data with the filter applied has this rights.

music_assistant/controllers/music.py
```diff
@@ -3421,3 +3350,81 @@ def library_sync_back_enabled(self, provider: Provider, media_type: MediaType) -
...
+        artist: str | None = None,
+        album: str | None = None,
+        media_type: MediaType | None = None,
+        username_or_user_id: str | None = None,
+    ) -> MediaItemType | ItemMapping | None:
+        """Try to find a media item (such as a playlist) by name.
+
+        This is a copy of the HA integrations method with additional username specification.
+        The HA method shall become obsolete.
+        """
```
```suggestion
        """Try to find a media item (such as a playlist) by name."""
```

music_assistant/controllers/music/controller.py
```diff
@@ -3421,3 +3350,81 @@ def library_sync_back_enabled(self, provider: Provider, media_type: MediaType) -
...
+            search_results.playlists,
+            search_results.artists,
+            search_results.radio,
+            search_results.audiobooks,
+            search_results.podcasts,
+        ):
+            for _item in results:
+                # simply return the first item because search is already sorted by best match
+                return _item
+        return None
```
marcelveldt, line 2533: Future expansion: use plugins that have AI query capabilities to fulfill the request. That way you can even make typos or type in natural language.

music_assistant/controllers/media/albums.py
fmunkes: Did sound correct to me yesterday. I'll remove the comment.

music_assistant/controllers/music.py
fmunkes: That makes sense.
Authenticated user: provider_a, provider_b, provider_c
Requested user: provider_b, provider_c
-> give only b and c results
But shall the following situation raise an error, or only return the shared items?
Authenticated user: provider_a, provider_b
Requested user: provider_b, provider_c
-> give only b or error?

I'll go with the error for now.

fmunkes: Another thought - other methods like music/get_library_items or <media_type>/get always allow to retrieve an item in the lib, also, if the user does have a provider filter set. Do we want to adjust them to?

music_assistant/controllers/music/controller.py
fmunkes: I'll added that to the backlog and a comment in the code.

fmunkes:
I added an additional method which verifies, if a user may access an item - I could use that one for my comment above for get and get_library_items too. I will need it to replace this part in the HA integration code:
```python
if "://" in media_id_str:
    media_uris.append(media_id_str)
    continue
# try content id as library id
if media_type and media_id_str.isnumeric():
    with suppress(MediaNotFoundError):
        item = await self.mass.music.get_item(
            MediaType(media_type), media_id_str, "library"
        )
        if isinstance(item, MediaItemType | ItemMapping) and item.uri:
            media_uris.append(item.uri)
        continue
```

music_assistant/controllers/music.py
marcelveldt: Nope those methods should also have a guard in place, I am like 99% sure of that.
There is some common helper somewhere which checks the provider instance id

music_assistant/controllers/music.py
```diff
@@ -886,12 +905,23 @@ async def get_item_by_uri(
         )
 
     @api_command("music/recommendations")
-    async def recommendations(self) -> list[RecommendationFolder]:
-        """Get all recommendations."""
+    async def recommendations(
+        self, username_or_user_id: str | None = None
```
marcelveldt: I am starting to wonder why we are adding the user override now to all endpoints/methods.
Aren't we a bit over doing this ?

The goal is that, when executing a playback related action from HA, we temporary "impersonate" as a certain user for that call.

A much easier way to do this would be setting a contextvar from that outer call that needs this (play media, and the new play media by name) which tells the auth system to override the user context for that command only, and then all underlying methods will just transparently follow.

The only requirement here is that the API user doing that call must have admin rights.

wdyt ?

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,42 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
     # Let the handler decide if authentication is required
     # The handler will call require_authentication() if needed
     return cast("web.StreamResponse", await handler(request))
+
+
+class OptionalImpersonatedUser:
+    """Optional impersonated user context manager.
```
marcelveldt:
```suggestion
    """
    Optional impersonated user context manager.
```
smal multi-line docstring preference

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,42 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
...
+    def __init__(self, mass: MusicAssistant, username: str | None) -> None:
+        """Initialize UseImpersonatedUser."""
+        self.mass = mass
+        self.username = username
+        if (
+            username is not None
+            and (authenticated_user := get_current_user())
+            and authenticated_user.role != UserRole.ADMIN
+        ):
+            raise InsufficientPermissions("Can only impersonate another user as Admin.")
```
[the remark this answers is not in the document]
fmunkes: Addressed

music_assistant/controllers/music.py
```diff
@@ -3634,3 +3637,142 @@ async def _credit_artist_plays(
...
+                self.playlists.library_items,
+                self.radio.library_items,
+                self.tracks.library_items,
+                self.albums.library_items,
+                self.artists.library_items,
+                self.audiobooks.library_items,
+                self.podcasts.library_items,
+            )
+            if not media_type or media_type.value.lower() in x.__name__
+        ]
```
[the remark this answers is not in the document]
fmunkes: addressed

music_assistant/controllers/player_queues/controller.py
```diff
@@ -1350,12 +1348,11 @@ async def _handle_play_media(
...
-        if username and (user := await self.mass.webserver.auth.get_user_by_username(username)):
-            playback_user = user
-        else:
-            playback_user = get_current_user()
-        queue.userid = playback_user.user_id if playback_user else None
+        if playback_user := get_current_user():
+            queue.userid = playback_user.user_id
+            self.logger.debug(
+                "User %s requested playback.", playback_user.display_name or playback_user.username
+            )
```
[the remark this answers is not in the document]
fmunkes: I think this is as before, so something we want.

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,42 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
...
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> bool | None:
+        """Unset the impersonated user."""
+        set_impersonated_user(None)
+        return None
```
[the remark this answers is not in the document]
fmunkes: addressed

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,42 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
...
+
+
+class OptionalImpersonatedUser:
+    """Optional impersonated user context manager.
+
+    If username is None this class does nothing. Otherwise, only an Admin
+    may impersonate another user. This is used for calls from HA (e.g. play_media on
+    the player queues controller).
+    """
+
```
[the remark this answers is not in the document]
fmunkes: added tests, adapted for nested usage

music_assistant/controllers/music/controller.py
```diff
@@ -3634,3 +3637,142 @@ async def _credit_artist_plays(
...
+        self,
+        name: str,
+        artist: str | None = None,
+        album: str | None = None,
+        media_type: MediaType | None = None,
+        username: str | None = None,
+    ) -> MediaItemType | ItemMapping | None:
+        """Try to find a media item (such as a playlist) by name."""
+        async with OptionalImpersonatedUser(self.mass, username):
+            return await self._handle_item_by_name(name, artist, album, media_type)
```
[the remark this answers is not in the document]
fmunkes: If we want these tests, I'd ask for help in setting up a mass_minimal mock for pytest. Tried a couple of things, but don't know how. I.e. something like this raises an AttributeError for the webserver when the OptionalImpersonatedUser is called. I thought there might be a way without all the setup required to test the authentication.

```python
@patch("music_assistant.webserver.auth.get_user_by_username")
async def test_verify_item_uri(mock_get_user_by_username: Mock, music: MusicController) -> None:
    """Test verify_item_uri."""
    admin_user = User(user_id="admin", username="admin", role=UserRole.ADMIN)
    set_current_user(admin_user)
    music.mass.webserver = WebserverController(music.mass)

    mock_get_user_by_username.return_value = Mock(
        user_id="admin", username="admin", role=UserRole.ADMIN
    )

    await music.verify_item_uri("fooo")
```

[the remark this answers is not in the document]
marcelveldt: we can do that as a follow-up now that we've started refactoring the music controller

music_assistant/controllers/music/controller.py
```diff
@@ -2422,3 +2425,148 @@ async def _credit_artist_plays(
...
+
+    async def _handle_item_by_name(
```
marcelveldt: please move private methods to the bottom of the file

marcelveldt: should the method be _handle_get_item_by_name or just _get_item_by_name ?

music_assistant/controllers/music/controller.py
```diff
@@ -2422,3 +2425,148 @@ async def _credit_artist_plays(
...
+    @api_command("music/verify_item_uri")
+    async def verify_item_uri(self, uri: str, username: str | None = None) -> bool:
+        """Verify, if a uri specifies a valid item.
+
+        If username_or_user_id is specified, verifies additionally, if this user may access this item. This requires the requesting (i.e. authorized user) to be able to access this item as well.
+        """
+        async with OptionalImpersonatedUser(self.mass, username):
+            return await self._handle_verify_item_uri(uri)
+
+    async def _handle_verify_item_uri(self, uri: str) -> bool:
```
marcelveldt, line 2535: make sure that private methods are at the bottom, public methods at the top.

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,60 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
     # Let the handler decide if authentication is required
     # The handler will call require_authentication() if needed
     return cast("web.StreamResponse", await handler(request))
+
+
+class OptionalImpersonatedUser:
```
marcelveldt: should the name just be "ImpersonatedUser" ?
the fact that its optional is irrelevant for the model/class name

music_assistant/controllers/music/controller.py
fmunkes: I dropped the "handle"

music_assistant/controllers/webserver/helpers/auth_middleware.py
fmunkes: I renamed it.

music_assistant/controllers/music/controller.py
fmunkes: There are a couple of other private methods not belonging to this PR in this class not at the bottom - shall I do a follow-up maintenance PR, where I move these?

marcelveldt: yes that would be nice - we're slowly starting to refactor controllers here and there to make things more maintainable long-term and consistent

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,60 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
...
+                "Authentication is necessary to use OptionalImpersonatedUser."
+            )
+        if username is not None:
+            if (
+                authenticated_user.role != UserRole.ADMIN
+                and authenticated_user.username != self.username
+            ):
+                raise InsufficientPermissions("Can only impersonate another user as Admin.")
+        else:
+            self.username = authenticated_user.username
```
marcelveldt, line 348: **This breaks unauthenticated / protocol-triggered playback.**

`play_media` now wraps every call in `ImpersonatedUser`, and this raises `InsufficientPermissions` whenever there's no authenticated user in context — even when no `username` is passed. So hardware presets (squeezelite) and Spotify Connect / Plex / AirPlay handoffs (`play_media(queue_id, uri)` with no auth) will now fail. The suite doesn't catch it because `play_media` is mocked everywhere. Confirmed at runtime: `ImpersonatedUser(mass, None)` with `current_user=None` raises.

Fix: only require auth when actually impersonating *another* user; a call with no `username` and no auth should be a no-op (anonymous playback). Verified this keeps the existing context-manager test green.

(Also: this message and the `__init__` docstring still say `OptionalImpersonatedUser` / `UseImpersonatedUser` — class is `ImpersonatedUser`.)

```suggestion
        if authenticated_user is None:
            # No authenticated user in context (e.g. playback from a hardware
            # button or an external protocol). Impersonating another user is not
            # allowed, but a call without a username is a no-op so anonymous
            # playback keeps working.
            if username is not None:
                raise InsufficientPermissions(
                    "Authentication is necessary to impersonate another user."
                )
            return
        if username is not None:
            if (
                authenticated_user.role != UserRole.ADMIN
                and authenticated_user.username != self.username
            ):
                raise InsufficientPermissions("Can only impersonate another user as Admin.")
        else:
            self.username = authenticated_user.username
```

music_assistant/controllers/webserver/helpers/auth_middleware.py
```diff
@@ -278,3 +302,60 @@ async def auth_middleware(request: web.Request, handler: Any) -> web.StreamRespo
...
+
+    async def __aenter__(self) -> Self:
+        """Set the impersonated user if applicable."""
+        assert self.username is not None  # for type checking
+        if impersonated_user := await self.mass.webserver.auth.get_user_by_username(self.username):
+            set_impersonated_user(impersonated_user)
+            return self
+        raise InvalidDataError(f"A user with user id or name {self.username} is not available.")
```
marcelveldt: Paired with the `__init__` change above: when there's no user to impersonate, `self.username` stays `None`, so `__aenter__` should no-op instead of asserting. Also renamed the walrus var so it no longer shadows the module-level `impersonated_user` ContextVar.

```suggestion
        if self.username is None:
            # no-op: nothing to impersonate (no authenticated user, no username)
            return self
        if user := await self.mass.webserver.auth.get_user_by_username(self.username):
            set_impersonated_user(user)
            return self
        raise InvalidDataError(f"A user with user id or name {self.username} is not available.")
```

music_assistant/controllers/music/controller.py
```diff
@@ -2422,3 +2425,148 @@ async def _credit_artist_plays(
...
+            MediaType.AUDIOBOOK,
+            MediaType.PODCAST,
+        ]
+        if media_type not in allowed_media_types:
+            raise InvalidDataError(
+                "%s is not a supported media_type. Supported media_types are %s",
+                media_type,
+                allowed_media_types,
+            )
+        media_types = allowed_media_types if media_type is None else [media_type]
```
marcelveldt: Two issues:

1. The default is `media_type=None`, but `None not in allowed_media_types` is `True`, so the default call always raises and the `media_type is None` branch on the next line is dead code. `MediaType.UNKNOWN` hits the same trap.
2. `InvalidDataError` doesn't `%`-format like a logger — the extra args just land in `.args`, so the message never interpolates. Needs an f-string.

```suggestion
        if media_type in (None, MediaType.UNKNOWN):
            media_types = allowed_media_types
        elif media_type not in allowed_media_types:
            raise InvalidDataError(
                f"{media_type} is not a supported media_type. "
                f"Supported media_types are {allowed_media_types}"
            )
        else:
            media_types = [media_type]
```

music_assistant/controllers/player_queues/controller.py
```diff
@@ -1248,12 +1246,11 @@ async def _handle_play_media(
...
```
marcelveldt: Follow-up to the auth fix: once anonymous playback is a no-op again, `get_current_user()` here can be `None`, and this block then leaves a stale `queue.userid` from a previous play. The old code always reset it — suggest restoring that:

```suggestion
        playback_user = get_current_user()
        queue.userid = playback_user.user_id if playback_user else None
        if playback_user:
            self.logger.debug(
                "User %s requested playback.", playback_user.display_name or playback_user.username
            )
```

music_assistant/controllers/music/controller.py
```diff
@@ -2422,3 +2425,148 @@ async def _credit_artist_plays(
...
+        """Try to find a media item (such as a playlist) by name."""
+        async with ImpersonatedUser(self.mass, username):
+            return await self._get_item_by_name(name, artist, album, media_type)
+
+    @api_command("music/verify_item_uri")
+    async def verify_item_uri(self, uri: str, username: str | None = None) -> bool:
+        """Verify, if a uri specifies a valid item.
+
+        If username_or_user_id is specified, verifies additionally, if this user may access this item. This requires the requesting (i.e. authorized user) to be able to access this item as well.
+        """
```
marcelveldt: Doc nit: the param is `username`, not `username_or_user_id`, and the sentence is a bit of a run-on.

```suggestion
        """
        Verify whether a uri points to a valid, accessible item.

        :param uri: The uri to verify.
        :param username: Optional user to additionally verify access for. Requires
            the authenticated caller to also have access to the item.
        """
```

marcelveldt (commented):
Took a deeper pass on this — overall the impersonation context manager is a clean approach 👍. Two things I think we should sort before merging, plus a couple of small ones; left them inline as suggestions.

- **Auth requirement on `play_media`** (the big one): wrapping every call in `ImpersonatedUser` makes it raise when there's no authenticated user, which breaks hardware/protocol-triggered playback (presets, Spotify Connect, Plex, AirPlay, …). Inline fix makes the no-username case a no-op so anonymous playback keeps working — verified it keeps the existing CM test green.
- **`_get_item_by_name` default**: `media_type=None` (the default) always raises today. Fixed inline together with the `InvalidDataError` f-string.
- **Tests**: the new `music/item_by_name` and `music/verify_item_uri` have no coverage, and the auth regression slipped through because `play_media` is mocked in the suite. Worth a small test calling `play_media` with no auth context, plus happy-path tests for the two new commands.

Nice work otherwise!

music_assistant/controllers/music/controller.py
marcelveldt: maybe also set translation_key (and args) if we want a localized error to the user?

music_assistant/controllers/webserver/helpers/auth_middleware.py
marcelveldt: Done in d0d00eb8: `ImpersonatedUser` is now a no-op when there is no user to impersonate, so anonymous/protocol playback works again.

music_assistant/controllers/webserver/helpers/auth_middleware.py
marcelveldt: Done in d0d00eb8 (no-op path, and dropped the shadowing walrus var).

music_assistant/controllers/music/controller.py
marcelveldt: Done in d0d00eb8: `None`/`UNKNOWN` now resolve to all types and the message is an f-string.

music_assistant/controllers/player_queues/controller.py
marcelveldt: Done in d0d00eb8: `queue.userid` is reset again for anonymous playback.

music_assistant/controllers/music/controller.py
marcelveldt: Done in 41c66165.

marcelveldt (approved):
Thanks, @fmunkes !

PR #4017, opened 2026-05-29
Make universal player merge deterministic when link counts tie

sdhomecode:
## Summary

Make the equal-link-count branch of `_check_merge_universal_players` deterministic by breaking ties on `player_id` lexicographically. Previously the comparison used `>=`, so on ties the caller-side `universal_player` always won — but which side ends up as caller depends on iteration order of `self._players.values()`, which is not stable across server restarts.

In practice this manifests as `_2` / `_3` entity_id suffixes appearing/disappearing in the Home Assistant integration after every MA server restart. Most visible on multi-zone setups with multiple AirPlay endpoints per physical speaker (Juke Audio clusters, etc.). Linked issue has full repro details.

## Linked issue

music-assistant/support#5549

## Testing

- Manually verified the merge outcome on a 3-box Juke Audio cluster (19 zones, ~50 player records) is stable across multiple consecutive MA restarts after the patch. Before the patch, restarts produced intermittent reshuffles.
- No regressions observed on single-protocol players.

## Test coverage follow-up

This repo doesn't currently have a `tests/controllers/players/` directory — adding a unit test for `_check_merge_universal_players` would require spinning up that infrastructure. Happy to do that as a follow-up if maintainers want it; left out of this PR to keep the surface small.

## Breaking changes

None. The change only affects which of two equivalent UniversalPlayer instances "wins" a merge — both contain the same identifiers post-merge, just under different `player_id`s. After deployment, existing setups may see a one-time `player_id` change as the tiebreaker picks a different winner than before, but from that point on the choice is stable.

### DIFF
--- a/music_assistant/controllers/players/protocol_linking.py
+++ b/music_assistant/controllers/players/protocol_linking.py
@@ -641,13 +641,23 @@ class ProtocolLinkingMixin:
                 )
                 continue
 
-            # Determine which player absorbs the other (more protocols wins)
-            keep, remove = (
-                (universal_player, player)
-                if len(universal_player.linked_output_protocols)
-                >= len(player.linked_output_protocols)
-                else (player, universal_player)
-            )
+            # Determine which player absorbs the other (more protocols wins).
+            # On ties, fall back to a deterministic tiebreaker on player_id so
+            # the merge outcome is stable across server restarts. Without this,
+            # which UniversalPlayer "wins" depends on iteration order of
+            # self._players.values(), which can shift between runs and causes
+            # downstream player_id reshuffling (and broken entity bindings in
+            # consumers like the Home Assistant MA integration).
+            len_u = len(universal_player.linked_output_protocols)
+            len_p = len(player.linked_output_protocols)
+            if len_u > len_p:
+                keep, remove = universal_player, player
+            elif len_u < len_p:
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Nice fix! Could you please add a unit test for this case in tests/core/test_protocol_linking.py?

sdhomecode:
Added a regression test in `tests/core/test_protocol_linking.py::TestUniversalPlayerMerging::test_merge_deterministic_tiebreaker_on_equal_link_counts` (commit f487c958) that exercises the equal-link-count branch by running the merge with both possible callers and asserting the same survivor (the lex-smaller `player_id`) in both orientations. Couldn't run pytest locally (no MA dev env on this box), so heads up if CI flags anything stylistic.

MarvinSchenkel:
Looks like the PR breaks another test (test_merge_universal_players_on_identifier_copy), which makes sense since you change the order one-time. That would need a fix as well. Could you please make sure the tests pass?

OzGav:
@sdhomecode Please run pre-commit before pushing your changes. The formatter is failing.

sdhomecode:
Ready for review, thanks for your help!

MarvinSchenkel (approved):
Thanks @sdhomecode !

PR #4019, opened 2026-05-29
fast MCP server: debug & config namespaces, external-source playback, OpenClaw/Hermes presets (v0.7.1)

trudenboy:

## FastMCP Server provider → v0.7.1

Updates the existing `fastmcp_server` plugin provider from **v0.3.35** (current
`dev`) to **v0.7.1**, landing two off-by-default capability namespaces, an
onboarding convenience for the Connect Wizard, and richer now-playing state for
external "Connect"-style sources. A default installation sees **no new tool
surface** — every new tool is gated behind a permission `ConfigEntry` that
ships disabled, so operators opt in per capability.

**Source**: [trudenboy/ma-provider-mcp](https://github.com/trudenboy/ma-provider-mcp) · tag [v0.7.1](https://github.com/trudenboy/ma-provider-mcp/releases/tag/v0.7.1)

### `debug` namespace — introspection & troubleshooting (was v0.4.0)

Twelve read-mostly tools for diagnosing a running Music Assistant instance over
MCP, behind five off-by-default flags (`Debug: providers`, `Debug: inspect`,
`Debug: events`, `Debug: logs`, `Debug: reload`):

- **Topology / health** — `debug_list_providers`, `debug_health_summary`,
  `debug_list_package_versions`, `debug_list_webserver_routes`.
- **Inspection** — `debug_inspect_player`, `debug_inspect_queue`,
  `debug_inspect_provider`, `debug_inspect_provider_config`.
- **Events** — `debug_recent_events`, `debug_event_buffer_stats`, backed by a
  bounded in-memory ring buffer (capacity configurable, default off).
- **Logs** — `debug_tail_log`, reading only files under MA's own
  `storage_path` (no arbitrary filesystem access).
- **Lifecycle** — `debug_reload_provider` (the one mutating tool in this
  namespace, gated by its own flag).

Supporting internals live under `debug/`: a ring event buffer, a depth- and
cycle-protected object inspector (back-reference skip-list to avoid serialising
the whole object graph), and a storage-scoped log reader.

Robustness refinements (v0.7.1): the log-scan tools (`debug_tail_log` and the
error count inside `debug_health_summary`) now run their synchronous file I/O
off the event loop; `debug_health_summary` honours the `Debug: logs` gate
before reading any log file; reload serialisation is per-runtime rather than
process-global; and the object inspector's byte-budget accounting no longer
round-trips every value through `json.dumps`.

### `config` namespace — view & edit settings (v0.5.0)

Fourteen tools to read and write MA core, provider, and player configuration
over MCP, behind five off-by-default flags — four capability flags
(`Config: read settings`, `Config: edit provider settings`,
`Config: edit core settings`, `Config: edit player settings`) plus an
orthogonal `Config: allow writing secret values` gate:

- **Read** — `config_list_targets`, `config_get_core`, `config_get_provider`,
  `config_get_player`, `config_get_entries`, `config_get_dsp`. `SECURE_STRING`
  values are masked by MA's own serialiser before they leave the process.
- **Stage a single value** — `config_set_core_value`,
  `config_set_provider_value`, `config_set_player_value` (validate + diff,
  no persistence).
- **Persist** — `config_save_core`, `config_save_provider`,
  `config_save_player`, `config_save_dsp`. Writes delegate to MA's atomic save
  primitives (validate → encrypt via `to_raw` → persist → reload, with
  rollback) — the provider never hand-rolls encryption or persistence.
- **Provider actions** — `config_trigger_provider_action`.

Writes are validated against each `ConfigEntry` (type coercion, explicit range,
per-member options) before they reach MA, and an optional elicitation
confirmation gate (`require_confirmation`, default on) prompts before any
mutation. Secret-valued keys are rejected atomically unless the dedicated
secret flag is enabled — that flag is re-evaluated per request, so toggling it
takes effect without a provider reload.

### External / Connect-source now-playing in briefs (v0.7.0 / v0.7.1)

When a player streams from an external "Connect"-style source (Spotify Connect,
AirPlay, Yandex Ynison), MA streams the audio out-of-band, so the player's own
`playback_state` stays `idle` while audio is actually playing. The player and
queue query tools now treat the **active queue** as the source of truth for
playback state — matching what MA's own interface shows:

- `players_list_players` / `players_get_player` (and the `player://` resource)
  report `playing` / `paused` from the active queue instead of a misleading
  `idle`.
- A new `external_source` field carries the controlling provider's instance id,
  so a client can tell which provider is driving playback (and distinguish
  several instances of the same provider). It is `null` for normal,
  self-driven playback.
- The currently playing item shows the real track title rather than the source
  wrapper name, in both player and queue views.

This is an enhancement to existing query tools — **no new tools and no new
permission flags**; it is gated by the same player/queue query permissions
that already exist.

### Connect Wizard — OpenClaw & Hermes presets (v0.6.0)

The Connect Wizard's client catalogue gains two entries — **OpenClaw** and
**Hermes** (Nous Research) — alongside the existing clients (Claude Code /
Desktop, Cursor, Windsurf, VS Code, ChatGPT, Codex, Gemini, Cline, Zed). For
each, the wizard mints a per-client long-lived token and renders a
ready-to-paste config snippet pointing at the server's streamable-HTTP endpoint
with an `Authorization: Bearer` header — an OpenClaw `openclaw mcp set …`
command and a Hermes `~/.hermes/config.yaml` block respectively. This is a
pure onboarding convenience: **no new tools, no new permission flags, no
runtime/transport changes** — the snippets target the same endpoint and auth
the wizard already wires for every other client.

- N/A

## Types of changes

## Checklist


music_assistant/providers/fastmcp_server/tools/debug.py
```diff
@@ -0,0 +1,607 @@
...
+        :param lines: Number of lines to return (clamped to [1, 2000], default 200).
+        :param level: Optional level filter (e.g. ``"ERROR"``).
+        :param component_regex: Optional regex matched against the component name.
+        :param since_seconds: When set, only entries with parseable timestamps within
+            this many seconds of "now" are returned.
+        :param name: Log file basename within ``$HOME/.musicassistant/``. Only the
+            canonical log and its rotated siblings (``.log.1`` … ``.log.5``) are
+            allowed.
+        """
+        return tail.tail(
```
MarvinSchenkel: **Blocking I/O on the event loop.** `tail.tail(...)` performs synchronous file I/O (`path.open("rb")`, `fh.seek`, `fh.read`) up to `_MAX_SCAN_BYTES = 10 MB` in `debug/log_reader.py:181-204`, but it's called directly from this `async def` tool. On a Home Assistant Yellow / Pi with slow eMMC or SD, 10 MB of seek+read will stall MA's single event loop hundreds of ms.

Same concern at `health_summary` (`tools/debug.py:581`), which calls `SafeLogTail(mass).count_errors_last_5min()` synchronously — that internally calls `tail(lines=2000, ...)` and can hit the same byte cap.

Suggested fix:

```python
return await asyncio.to_thread(
    tail.tail,
    lines=lines,
    level=level,
    component_regex=component_regex,
    since_seconds=since_seconds,
    name=name,
)
```

and similarly wrap the `count_errors_last_5min()` call in `health_summary`.

music_assistant/providers/fastmcp_server/server.py
```diff
@@ -65,6 +68,8 @@ def __init__(
         # Mutable so apply_permission_change can hot-swap the allowed-tag set
         # without re-instantiating the TagFilterMiddleware closure.
         self._allowed_tags: set[str] = set()
+        self._event_buffer: Any = None  # provider.debug.event_buffer.EventBuffer | None
+        self._reload_lock: asyncio.Lock = asyncio.Lock()
```
MarvinSchenkel, line 72: `self._reload_lock` is allocated here but never referenced anywhere in the provider — the actual reload serialization happens via the module-level `_RELOAD_LOCK` in `tools/debug.py:64`.

Either drop this attribute, or plumb it through to `build_debug_server` / `_register_reload_tool` so the lock is per-runtime instead of shared across every `MCPServerRuntime` that imports the module (the latter is arguably the cleaner choice — module-level state unnecessarily serializes unrelated test instances).

MarvinSchenkel (commented):
Two problem-level findings inline. Suggestions from my console review (event_buffer docstring, health_summary reading logs when DEBUG_LOGS off, module-vs-instance lock scope, json.dumps in _State.charge) are not posted here.

music_assistant/providers/fastmcp_server/tools/debug.py
trudenboy: wrapped both scans in asyncio.to_thread

music_assistant/providers/fastmcp_server/server.py
trudenboy: plumbed reload_lock through per-runtime and dropped the module-level lock

MarvinSchenkel (approved):
Thx @trudenboy

PR #4025, opened 2026-05-29
Fix smart playlist track evaluation from Discover and background queue context

dmoo500:

Fixes two bugs in the smart playlist provider:

**1. Discover page shows no sample tracks for rule-based smart playlists**

Opening a smart playlist from Discover/Recommendations navigates using the provider UUID. The server's `get()` call resolves this to the library item and returns its DB integer `item_id`. `get_playlist_tracks` then receives that integer ID, but `_rules_store` is keyed by UUID — so the lookup returns `None` and the sample is empty.

Fix: `get_playlist_tracks` now falls back to `_resolve_to_provider_id` when the direct store lookup fails, so both UUID and DB integer IDs work correctly.

**2. Provider filter ignored during background queue refill**

`_fill_radio_tracks` runs as a background task without a request context, so `get_current_user()` returns `None` and `_ensure_provider_filter` cannot apply the queue owner's provider filter. Tracks from disallowed providers could appear in dynamic playlist refills.

Fix: Before fetching tracks, the queue owner's user is looked up via `queue.userid` and set as the current user context.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -78,7 +78,10 @@ from music_assistant.constants import (
 )
 from music_assistant.controllers.players.constants import PlayerLockPurpose
 from music_assistant.controllers.streams.audio_buffer import AudioBuffer
-from music_assistant.controllers.webserver.helpers.auth_middleware import get_current_user
+from music_assistant.controllers.webserver.helpers.auth_middleware import (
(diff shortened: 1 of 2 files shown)


MarvinSchenkel (approved):
Thanks @dmoo500

PR #4041, opened 2026-05-31
Hide background analysis concurrency from generic core-config editor

chrisuthe:
## What
Change `background_scan_concurrency` core-config entry (streams domain, `audio_analysis` category) as `hidden=True` so it isn't in stream settings.

## Why

Move it via front-end change to the audio analysis page (https://github.com/music-assistant/frontend/pull/1830)


MarvinSchenkel:
If we unconditionally hide it, why not remove the config option and use a constant ?

chrisuthe:
> If we unconditionally hide it, why not remove the config option and use a constant ?

The reason I kept the entry rather than removing it was that the value still needs to be saved somewhere, and the only save path goes through the registered config entries — drop the entry and the save just gets ignored, so the setting wouldn't stick. Since we're keeping it user-tunable (just relocating it to the Audio Analysis page), the entry has to stay. hidden=True keeps the save/validation working but takes it out of the generic config screen so it isn't shown in two places.

Happy to go a different route if you know a cleaner one.

MarvinSchenkel:
I think this is more of a hack than a feature (also taking into account the frontend PR). We have an established system for configuration. For example, we automatically render those setting entries for all Core controllers. 

Now this is new territory, we want something configurable for a NON Core controller (The AA controller) and still show custom sections in that page (the coverage details). @marcelveldt is on PTO now, but I would like to discuss this use case when he gets back. 

For now, I would just draft this PR and update the frontend with just a link to the setting in the streams controller until we have settled on an approach.

PR #4042, opened 2026-05-31
Implement the Sendspin `visualizer@v1` role and rework Hue Lights Sync

maximmaxim345:
Implements the `visualizer@v1` role on the Sendspin server and updates Music Assistant to aiosendspin 6.0.1. The older `visualizer@_draft_r1` role is left unchanged and still carries no beats. Clients connecting with the `visualizer@_draft_r1` role still remain fully functional. 

## Visualizer

The Sendspin player derives a per-track beat schedule from `smart_fades` analysis and streams it to visualizer clients over `visualizer@v1`. The Hue Entertainment plugin is reworked to use the v1 role, including beat-based colour cycling with selectable modes.
Older clients using `visualizer@_draft_r1` stay fully compatible after this PR.

## Player timing

Reports each player's lead-time and live-source hints to the push stream so it schedules the first chunk far enough ahead. This stops the AirPlay bridge from cutting off the start of a track.
Older clients not implementing this stay fully compatible after this PR.

## Repeat and shuffle

Repeat and shuffle now ride on controller state, following the aiosendspin 6.0 move off metadata. The server sets the controller state and keeps the legacy metadata copy in sync so older and newer clients both work.

## Known limitations

The Hue plugin's beat effects need the track's beat analysis. On the first play of a track whose analysis has not been computed yet, the lights use the peak and onset fallback for the first stretch (up to ~30s) until beats arrive.

During a smart-fades transition the beats of the crossfading tracks are used as is and can drift slightly while the two tracks overlap. Alignment is correct again once the transition completes.

The Hue bridge uses a small visualizer buffer for near-realtime delivery, so after a track change beats take a few seconds to start arriving (the first moments of the new schedule are kept near the playhead and not delivered yet). The lights use the peak and onset fallback until they do.

Beat data is loaded by a lightweight retry poller rather than an audio-analysis event subscription. This is intentional, to keep the changes contained to the Sendspin provider and avoid touching core Music Assistant code before the 2.9 release.

Visualizer pitch detection is disabled server-wide for now. It is the heaviest visualizer DSP and its result quality is still mixed, so it needs more testing before being enabled.

## Relevant Specification this PR implements
- https://github.com/Sendspin/spec/pull/86
- https://github.com/Sendspin/spec/pull/69
- https://github.com/Sendspin/spec/pull/81

## Testing

I tested this locally, on a Raspberry Pi 4, and a Home Assistant Green.
Lower powered clients may not be powerful enough to compute `beats` in time, but the Hue bridge falls back to `peaks` if thats the case.


music_assistant/providers/airplay/sendspin_bridge.py
```diff
@@ -201,6 +202,20 @@ async def stop(self) -> None:
 
         self.logger.debug("Sendspin bridge stopped for %s", self.airplay_player.display_name)
 
+    def _refresh_bridge_timing(self) -> None:
+        """Push the AirPlay startup latency to the bridge role.
```
MarvinSchenkel: Multi line docstrings should start on a new line

MarvinSchenkel (approved):
AWESOME stuff @maximmaxim345 !

PR #4046, opened 2026-06-01
Import album tracks when manually adding an album and Import Album Tracks setting enabled

OzGav:

The "Import album tracks" library option only took effect during a library sync via `_sync_library_albums`. When a user enabled the setting and added an album to the library from the UI, the album's tracks were never imported, since the manual add path (music/library/add_item) never invoked the album-track import logic.

This PR extracts the album-track import into a reusable `MusicProvider.import_album_tracks` method (plus a `library_sync_album_tracks_enabled` helper) and invokes it from `add_item_to_library` for any provider mapping that has the setting enabled, mirroring the sync behavior. This affects all streaming providers but was reported against Qobuz.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -1165,9 +1165,23 @@ class MusicController(CoreController):
         # add (or overwrite) to library
         ctrl = self.get_controller(full_item.media_type)
         library_item = await ctrl.add_item_to_library(full_item, overwrite_existing)
+        # optionally import all album tracks into the library, mirroring the behavior
+        # of the library sync (which only triggers on a (scheduled) full sync run)
+        if full_item.media_type == MediaType.ALBUM:
+            self._import_album_tracks_if_enabled(cast("Album", library_item))
         # perform full metadata scan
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/music.py
```diff
@@ -1165,10 +1165,24 @@ async def add_item_to_library(
...
+        # of the library sync (which only triggers on a (scheduled) full sync run)
+        if full_item.media_type == MediaType.ALBUM:
+            self._import_album_tracks_if_enabled(cast("Album", library_item))
         # perform full metadata scan
         await self.mass.metadata.update_metadata(library_item, overwrite_existing)
         return library_item
 
+    def _import_album_tracks_if_enabled(self, album: Album) -> None:
+        """Import all album tracks into the library for providers that have this enabled."""
+        for prov_mapping in album.provider_mappings:
```
MarvinSchenkel, line 1178: I think this will mess with multi instance / user configurations since it blindly triggers this on all linked providers? We only want to this to trigger an import on the provider instance on which we added an album to the library right?

MarvinSchenkel (approved):
Thx @OzGav !

PR #4050, opened 2026-06-01
Add webradio provider

OzGav:

Satisfies https://github.com/orgs/music-assistant/discussions/2410 and maybe https://github.com/orgs/music-assistant/discussions/2972

Meets the OHF objective of sustainability in that older devices can now receive a stream from MA. Removes the current advertised workaround in the docs of using the HA MPD integration to act as a middleman to create the radio stream.

Each configured station appears in Music Assistant as a virtual player with its own queue and is served on a stable HTTP URL using Shoutcast/ICY. 

A station produces audio through a single shared ffmpeg pipeline and fans the encoded chunks out to every connected listener; the pipeline starts on the first listener and stops once the last one disconnects, so an idle station consumes nothing. The producer is driven by `play_media` rather than `queue.current_item` so a PLAY_NOW or seek never plays the wrong track. Pause and seek are hidden in the UI because neither has a sensible meaning on a live broadcast.

ICY metadata (track titles, optional cover-art URL) is on by default in the "basic" profile and configurable per station. Codec defaults to MP3 for compatibility; FLAC is also available.

One small change in core: extracted the inline ICY metadata-frame construction in `serve_queue_flow_stream` into `helpers/audio.format_icy_metadata_frame` so the new provider and the existing flow handler share one implementation. 

## Types of changes

## Checklist


music_assistant/controllers/streams/controller.py
```diff
@@ -934,26 +935,24 @@ async def serve_queue_flow_stream(self, request: web.Request) -> web.StreamRespo
                 continue
 
             # if icy metadata is enabled, send the icy metadata after the chunk
+            # use the current (not buffered) item so the title stays aligned with audio
```
marcelveldt: This is a bad change to make as this will cause the metadata to be out of sync

music_assistant/providers/webradio/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+async def get_config_entries(
+    mass: MusicAssistant,
+    instance_id: str | None = None,
+    action: str | None = None,
+    values: dict[str, ConfigValueType] | None = None,
+) -> tuple[ConfigEntry, ...]:
+    """Return the provider-level config entries."""
+    del mass, instance_id, action, values
+    return (
+        ConfigEntry(
```
marcelveldt, line 52: We can make this easier maybe ?
why not dynamically register a player if you e.g. try to play http://massurl:8097/webradio/myplayer.mp3

this will auto create (or restore) a player+queue with that name an use mp3 as output format
if the player disconnects for more than x seconds, the player is unregistered
if there is queue items, stream empty

music_assistant/providers/webradio/provider.py
```diff
@@ -0,0 +1,942 @@
...
+        """
+        async for chunk in get_ffmpeg_stream(
+            audio_input=self.mass.streams.audio.get_queue_flow_stream(
+                queue=queue,
+                start_queue_item=start_item,
+                pcm_format=flow_pcm_format,
+            ),
+            input_format=flow_pcm_format,
+            output_format=output_format,
+            filter_params=self.mass.streams.audio.get_player_filter_params(
```
marcelveldt: this is weird to have filter params on the main stream

music_assistant/providers/webradio/provider.py
```diff
@@ -0,0 +1,942 @@
...
+            superseded by a skip).
+        """
+        async for chunk in get_ffmpeg_stream(
+            audio_input=self.mass.streams.audio.get_queue_flow_stream(
+                queue=queue,
+                start_queue_item=start_item,
+                pcm_format=flow_pcm_format,
+            ),
+            input_format=flow_pcm_format,
+            output_format=output_format,
```
marcelveldt, line 512: so be aware that not every output format will work with this. for example I'm pretty sure you cant join a flag stream mids-tream unless you wrap it into a ogg container

music_assistant/providers/webradio/provider.py
```diff
@@ -0,0 +1,942 @@
...
+    pending_media_event: asyncio.Event = field(default_factory=asyncio.Event)
+    # Populated by the producer once output config is known.
+    output_format: AudioFormat | None = None
+    icy_preference: str = "disabled"
+    icy_chunk_size: int = 0
+    current_metadata: bytes = b""
+
+
+class WebRadioProvider(PlayerProvider):
+    """Player provider that publishes each configured station as an HTTP audio stream."""
```
marcelveldt, line 109: I see some weird (AI created) method names here and there that can use a better name.
Also some of the code can be a bit more compact/tight, same for docstrings and comments.

The idea in general is good I think.
But what if we take it one step further and make this a default provider (just with a better name, because webradio is weird) or even make it a subcontroller of the streams controller. No config options, just dynamic player(queue) creation if you try to open the url as I described above

This would also open the opportunity to browse and play MA content on any player within HA due to the fact that the player(queue) is just dynamically generated

There is a lot of overlap however with the MultiClientStream used in UGP and squeezelite.
We should either migratie this into that helper or turn this little framework into the new standard.
For UGP, this is definitely good enough, one single stream with multiple consumers.

Food for thought....

music_assistant/controllers/streams/controller.py
OzGav: You're right that comment/structure shouldn't have changed. Reverted to the original shape so now the only delta is the call into the shared `format_icy_metadata_frame` helper, which produces byte-identical frames

music_assistant/providers/webradio/provider.py
OzGav: Good point — dropped. The broadcast pipeline now passes no filter params, so what listeners receive is the raw queue audio. If anyone wants per-station DSP later it can be a per-station config rather than reusing the per-player one

music_assistant/providers/webradio/provider.py
OzGav: Hmmm good point. I only tested FLAC with a fresh single listener, not mid-stream join or skip-restart, which is where it'd actually break. So now I restricted the codec dropdown to MP3 and AAC. You are right I think FLAC over OGG would also work but didn't seem worth the container complexity for the marginal benefit...?

music_assistant/providers/webradio/provider.py
OzGav: Firstly, I have tightened the code generally. 

Secondly, thanks for the bigger picture. Full disclosure, I'm not familiar with UGP, squeezelite, or MultiClientStream so I can't really judge which is best between folding into MultiClientStream or making this the new shared helper. I think that is a discussion for you and Marvin?

On the dynamic-player-by-URL model I can think of a few questions such as: what does the device get on first connect with an empty queue; whether persisted entries stick in player settings (marked unavailable like other offline players) or get aged out; whether the queue is reachable for pre-listener curation when no listener has been around recently; how the URL's codec extension and any stored codec preference reconcile; and whether anything guards against URL hits creating state unintentionally.

Given those open questions I'd lean toward keeping the explicit-config model for now as it matches how other player providers work and avoids the lifecycle and persistence questions. The downside though is that it doesn't unlock the "any HA media player can play MA content by URL" capability your bigger vision points at and that needs the dynamic approach.

I'm happy for the bigger restructure to be your call. If you think this PR is a useful starting point that can be refactored later, we could add it basically as-is and let you drive the bigger plan as a follow-up. If you'd rather it be done first,  you'd need to steer me onto the target.

OzGav: OK I have tried to get smarter about this and I think I understand more about the idea but it just brings up more questions!

So we have two approaches, current plugin model or core functionality model. I think they're optimising for different users, not the same problem expressed two different ways. I think plugin is simpler UX for a dumb internet radio in the kitchen as the user sets up the stream, gets shown the URL required, and enters that into the device. Dynamic-URL is essential for the "any HA media_player plays MA content via URL" capability you outlined and I don't think that's achievable any other way.

So the call really comes down to whether MA wants to commit to URL-addressable queues as core infrastructure. That feels bigger than this PR and like it'd benefit from your architectural call before I build. Two main questions:

Does the URL-addressable-queue idea belong as core MA infrastructure (sub-controller of streams), or as a provider?
If infrastructure, is now the right time to commit, or would you prefer the explicit provider lands now and we revisit later?
As I wrestled with these concepts I came up with this comparison table:

image

edit: I have also considered a hybrid model where the dynamic mechanism is the foundation, and the explicit provider is a thin convenience layer that sits on top of it.

marcelveldt: I think it should be both;

- It should be an independent player provider
- It has a config flow that allows you to setup one or more permanent players by specifying their name(s) --> this results in a URL outputted per player to use for streamer
- It also has a dynamic mode where you can create a temporary webplayer on the fly, this can be done by adding some api commands
- for the dynamic mode (e.g. called from HA for ANY player in the media browser) is that HA processes the media browser command for a player unknown to MA, calls this provider's api command to create a player on the fly --> enqueues the requested media --> temporary player ic cleaned up as soon as the actual stream is not read for x amount of time

Our configentry model should be able to handle this (creating a list of permanent players to create), if not we should look into expanding it OR we just create custom frontend (not preferred but can be considered).

In any case I hope we can come up with a better name as webradio is a bit weird - maybe something like "Web Player" ? Although that clashes a bit with our builtin web player running sendspin (but we can consider making that switcheable between sendspin and unsynced playback of this provider)

marcelveldt: BTW: Let me know if you want my help with any of this

OzGav: Ah, I had not thought of using an API call to do this. That clears up some of my concerns.

If dynamic creation is triggered by an API call rather than by an HTTP request hitting an open URL, then:

- Nothing is created accidentally, only when something explicitly asks for it.
- HA will have already added what it wants to play before the device connects, so the queue is never empty at first connect. That removes the silence question entirely.
- The temporary players don't need persisted settings since they're explicitly throwaway.
- The permanent players keep working the way they do today with the same UI, same setup, same URL copy-paste.

A few things to check:

- I think the extra concepts you are talking about would be additive to this PR so is it an option to merge what's on the branch now, with temporary players and the API command as a follow-up PR? Smaller pieces, easier to review?
- On the name, "Web Player" works for me in principle, I also thought of "Web Broadcast"?
- I'm already using a multi-value stations `ConfigEntry` (one name per line, each becomes a player). Is that the setup you had in mind for the permanent list?

As for extra help I'm happy with anything you have the capacity to provide!

OzGav: @marcelveldt any further guidance?

PR #4053, opened 2026-06-01
Add Settings to allow Control of default similar_track action

chrisuthe:

Adds three controls:
- Select which similarity engine to use for generic similar_tracks calls (default to 18dim index, allow choosing CLAP if it's built) 
- Select which preset to use if using 18dim index
- Set diversity control if using the 18dim index
This is stand alone from the same settings for the "Discover" page. 

Adds routing debug logs
Address a O(N) CLAP seed Lookup performance gap (results: O(1))

_none_

## Types of changes

## Checklist


music_assistant/providers/sonic_similarity/__init__.py
```diff
@@ -204,11 +206,58 @@ async def get_config_entries(
...
             depends_on=CONF_ENABLE_CLAP_INDEX,
             depends_on_value=True,
         ),
+        # Engine selector for Similar Tracks; shown only when CLAP is enabled.
+        ConfigEntry(
+            key=CONF_SIMILAR_TRACKS_ENGINE,
+            type=ConfigEntryType.STRING,
+            default_value=SIMILAR_ENGINE_18DIM,
+            label="Similar Tracks engine",
+            description="Which index powers library-wide Similar Tracks. The 18-dim engine "
```
MarvinSchenkel: While I don't disagree, let's put this in words average users also understand, for example 'what it sounds like vs. what you describe it as' (remember you mentioned this in one of our calls)

music_assistant/providers/sonic_similarity/__init__.py
```diff
@@ -204,11 +206,58 @@ async def get_config_entries(
...
+                ConfigValueOption("Balanced", "balanced"),
+                ConfigValueOption("Vibe (mood + timbre)", "vibe"),
+                ConfigValueOption("Party (rhythm-heavy)", "party"),
+                ConfigValueOption("Genre + Era (stay close)", "genre_era"),
+                ConfigValueOption("Discover (novelty-leaning)", "discover"),
+            ],
+            depends_on=CONF_SIMILAR_TRACKS_ENGINE,
+            depends_on_value=SIMILAR_ENGINE_18DIM,
+        ),
+        ConfigEntry(
```
MarvinSchenkel, line 223: If you add `range(0, 1.0)` here, you should get a fancy slider

music_assistant/providers/sonic_similarity/__init__.py
MarvinSchenkel, line 1: Just had a play with this. I think we should re-arrange / re-word some of the categories to make everything a bit more clear to the end user.. Or link to the documentation in certain sections. For example, the 'similar track preset' is not very clear without actually reading the documentation. 


Suggestion for ordering
Generic
* Enable CLAP
* Enable free-text-search (they kinda go together righ?)

Similarity search (split from generic)
* Engine
* Preset
* Diveristy

Discover
* Show xxx
* Preset
* Diversity

Status
....
Think it makes sense to put this at the bottom, maybe even make it advanced. I think average users won't have to rebuild the index?

MarvinSchenkel:

music_assistant/providers/sonic_similarity/__init__.py
chrisuthe: Agree, going to use the following descriptors everywhere I can: 

18-dim → "Matches on measured sound traits — tempo, energy, loudness, key"
CLAP → "Matches on overall character — how a listener would describe the sound"

music_assistant/providers/sonic_similarity/__init__.py
chrisuthe: Good callout.

music_assistant/providers/sonic_similarity/__init__.py
chrisuthe: image
image

MarvinSchenkel (approved):
Thanks @chrisuthe

PR #4058, opened 2026-06-02
fix(tidal): serve DASH manifests via HTTP route instead of data: URI

libre-7:
## Problem

Tidal returns dynamic DASH manifests (MPD) with a sliding segment window (~35s). ffmpeg needs to re-fetch the manifest during playback to obtain the next segment window. The current code encodes the manifest as a `data:` URI — ffmpeg reads it once, plays the initial segments, then when it tries to re-fetch the MPD for the next window, the `data:` URI has nothing left. The stream dies silently at ~35s.

Closes #5077

## Fix

Decode the base64 manifest and register it as a real HTTP endpoint on the internal stream server using the existing `register_dynamic_route` API. ffmpeg can re-fetch the MPD on demand just like any other HTTP stream.

## Design decisions

- **Content-hashed path** (`/tidal-dash/{md5(manifest)}`): identical manifests reuse the same route, avoiding redundant registrations. The `RuntimeError` on duplicate registration is caught silently.
- **Fixed TTL cleanup** (10 minutes): the route self-destructs after 10 minutes via `call_later`. This is generous enough for even long tracks. The previous attempt (PR #3369) used a timer that fired too early for long/slow streams. A 600s TTL eliminates that race while preventing route leaks.
- **No new dependencies**: uses `base64`, `hashlib` (stdlib), and `aiohttp.web\Response` (already a dependency).

## Testing

- Verified syntax on both modified files
- Unit test updated to assert HTTP route URL instead of `data:` URI
- Test verifies `register_dynamic_route` is called and `call_later` schedules cleanup
- The test manifest base64 decodes to `manifestData` and produces a deterministic hash

### DIFF
--- a/music_assistant/providers/tidal/streaming.py
+++ b/music_assistant/providers/tidal/streaming.py
@@ -2,9 +2,12 @@
 
 from __future__ import annotations
 
+import base64
+import hashlib
 from sqlite3 import OperationalError
 from typing import TYPE_CHECKING
 
+from aiohttp import web
 from music_assistant_models.enums import ContentType, ExternalID, StreamType
 from music_assistant_models.errors import MediaNotFoundError
 from music_assistant_models.media_items import AudioFormat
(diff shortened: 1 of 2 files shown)


libre-7:
Holding for review — withdrawn pending user approval

PR #4062, opened 2026-06-02
Fix Tidal DASH playback stuttering by serving manifests as HTTP routes

libre-7:
## Types of changes

- [x] Bug fix `bugfix`

## Summary

Tidal returns DASH MPD manifests as `data:application/dash+xml;base64,...` URIs. ffmpeg reads the manifest once, then attempts to re-fetch the MPD during playback to verify segment availability — but a `data:` URI cannot be read twice, so the stream fails after the initial segment window (~35 seconds).

## Fix

**`music_assistant/providers/tidal/streaming.py`**

- Decode the base64 manifest, content-hash it, and register it as a temporary HTTP route on the stream server via `register_dynamic_route`
- A self-extending idle buffer cleans up the route after `track.duration + 300s` of inactivity — the handler reschedules the deadline on every manifest fetch
- `Cache-Control: no-cache` prevents proxy caching of the live manifest

**`tests/providers/tidal/test_streaming.py`**

- 5 new tests covering the manifest handler response body/content-type/headers, duplicate route registration (cleanup still scheduled on reuse), `_remove_dash_route` delegation, and error swallowing
- Existing DASH test updated to assert HTTP route URL instead of `data:` URI

## Testing

Verified on a Chromecast group and single speaker:
- Tidal tracks play fully through (previously cut at ~35s)
- Track transitions advance GUI metadata correctly
- Seek forward/backward to any position works
- Crossfade and gapless playback unaffected
- No 404s or route cleanup races observed

Closes #5077


libre-7:
Lint fix pushed: removed `os.process_cpu_count` compat patch from test file. The `# type: ignore[attr-defined]` was flagged as unused on Python 3.14+ where `os.process_cpu_count` exists natively. The patch was only needed for local 3.12 environments — CI runs 3.14 so its unnecessary there.

Test run (3770 passed, 6 skipped, 0 failed) was clean — the lint failure was the only blocker.

MarvinSchenkel:
@jozefKruszynski could you have a look at this one please?

music_assistant/controllers/streams/audio.py
```diff
@@ -1613,6 +1613,19 @@ async def get_queue_item_stream(
...
+                # duration + a short buffer, stop yielding so the flow stream
+                # loop can advance to the next queue item. Bounded sources
+                # (FLAC, MP3, local files) finish naturally well before this
+                # point, making the break a no-op for them.
+                if (
+                    streamdetails.duration
+                    and (bytes_received / pcm_format.pcm_sample_size + seek_position)
+                    >= streamdetails.duration + 5
+                ):
+                    break
```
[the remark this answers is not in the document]
MarvinSchenkel: I agree here, changing this changes CORE streaming behaviour, so it feels very tricky to me to accept this as a workaround for something like Tidal. Preferably, we should come up with a solution that is scoped to Tidal to avoid breaking other parts of MA.

jozefKruszynski:
Will do, I commented over on the issue thread that I'll take a look as soon as I'm back from holidays.

libre-7:
Addressed Copilot's `usedforsecurity=False` feedback in `ebac3cd`. The MD5 call is content-deduplication only, not cryptographic — matches the existing pattern used in `helpers/images.py`, `apple_music/recommendations.py`, and `airplay_receiver/__init__.py`.

libre-7:
**Removed the `audio.py` duration cap** — it was an unnecessary workaround.

After dumping the raw Tidal DASH MPD and confirming it is `type="static"` with a finite `SegmentTimeline` (example from track `198584630`: 51 segments of 176128 samples + 1 final segment of 91438 samples, bounded by `mediaPresentationDuration="PT3M25.758S"`), the stream terminates naturally without any duration cap. Verified this by testing on a single Chromecast speaker — track transitions happen cleanly, ffmpeg exits after the last segment, and the queue advances normally.

The cap touched core playback logic (`audio.py`) without justification. 3 files → 2 files now.

The Chromecast *group* metadata hang is a separate pre-existing universal group player lifecycle issue (flow streams restart without clean `Finished Queue Flow` signals), not related to this change.

libre-7:
This PR is still an active issue — Tidal DASH playback on Chromecast group speakers is broken without this fix. It will be reviewed by @jozefKruszynski and all feedback has been addressed. Please keep this open.

jozefKruszynski:
> This PR is still an active issue — Tidal DASH playback on Chromecast group speakers is broken without this fix. It will be reviewed by @jozefKruszynski and all feedback has been addressed. Please keep this open.

Hi, sorry for the long delay in getting to this.
If the issues have been addressed, can you please mark them as resolved. I'll try to have a look at things today.

jozefKruszynski:
Can you please clean up the body of the PR to reflect the actual PR contents, it still talks about the reverted change to audio.py and describes tests that no longer exist.

The PR title also needs to change please. Check the project standards for PR titles. Essentially the PR title should be a human user focused descriptive title, as opposed to using the conventional commit standard.

I will add some inline review comments also.

music_assistant/providers/tidal/streaming.py
```diff
@@ -57,7 +69,53 @@ async def get_stream_details(self, item_id: str) -> StreamDetails:
         # 4. Parse stream URL
         manifest_type = stream_data.get("manifestMimeType", "")
         if "dash+xml" in manifest_type and "manifest" in stream_data:
-            url = f"data:application/dash+xml;base64,{stream_data['manifest']}"
+            # Tidal returns a dynamic DASH manifest (MPD) that ffmpeg needs
```
jozefKruszynski: This statement is not true as you mentioned in another review comment. The manifest is static. Please update the comment to reflect this.

music_assistant/providers/tidal/streaming.py
```diff
@@ -57,7 +69,53 @@ async def get_stream_details(self, item_id: str) -> StreamDetails:
...
+            # to re-fetch during playback to get the next segment window.
+            # A data: URI can't be re-fetched, so we serve the decoded
+            # manifest from a real HTTP endpoint on the stream server.
+            manifest_bytes = base64.b64decode(stream_data["manifest"])
+            manifest_hash = hashlib.md5(manifest_bytes, usedforsecurity=False).hexdigest()
+            route_path = f"/tidal-dash/{manifest_hash}"
+            cleanup_id = f"tidal-dash-cleanup-{manifest_hash}"
+
+            # Use track duration + buffer so the route lives through seeks.
+            # ffmpeg's periodic manifest fetches extend the deadline further
```
jozefKruszynski: Wording here also reflects dynamic manifest fetching which you have established is not the case.

music_assistant/providers/tidal/streaming.py
```diff
@@ -17,6 +21,14 @@
 
     from .provider import TidalProvider
 
+# Seconds of inactivity after which a DASH manifest route is cleaned up.
+# The cleanup timer resets each time ffmpeg fetches the manifest, so this
+# is only an idle timeout applied AFTER active playback (or seeks) stop.
+# Set to track duration + 300s so that seeks and queue transitions always
+# land on a live route, even when the old ffmpeg process has been dead for
+# minutes before the new one starts fetching.
+_MINIMAL_DASH_ROUTE_TTL: int = 300
```
jozefKruszynski: This is actually never used as a minimum, and the naming makes it slightly confusing.
You state in the comment preceding it that it is in fact an idle buffer, please name it as such.

jozefKruszynski (dismissed):

libre-7:
All review comments addressed:

1. `_MINIMAL_DASH_ROUTE_TTL` → `_DASH_ROUTE_IDLE_BUFFER` — renamed to reflect it is an idle buffer (`873d4a6`)
2. **Line 72 comment** — corrected "dynamic" wording to accurately describe the static manifest + re-fetch behavior
3. **Line 82 comment** — changed "periodic manifest fetches" → "manifest re-fetches"
4. **PR body** — removed stale `audio.py` section, corrected test description
5. **PR title** — updated to human-focused format per project standards

jozefKruszynski:
Minor niggle that the comments on line 74 of streaming.py still refer to "updated" segments, which is not really the case as the manifest is static.

Other than that the changes look good.

@MarvinSchenkel I would give this a thumbs up after the necessary rebase.

libre-7:
Fixed the "updated segment windows" wording too — changed to "read the segment timeline" in `a420834`. Also rebased onto latest `dev`. Ready for another look.

marcelveldt (approved):
LGTM
Thanks, @libre-7 !

jozefKruszynski:
Nice work @libre-7 
Thank you.

PR #4068, opened 2026-06-03
Mark players unavailable on controller disconnection for HEOS

MarvinSchenkel:

When a HEOS-connected AVR is fully power-cycled, pyheos internally marks the device as unavailable, but Music Assistant was not listening for the disconnection event. This left the player showing as available in MA while the underlying transport connection was gone, causing every playback command to fail with `CommandError: Not connected to device` until the MA addon was restarted.

The fix subscribes to pyheos's `add_on_disconnected` callback in `_setup_controllers`. When the callback fires, `set_device_info()` is called on each registered HEOS player, which reads the (now-False) `available` attribute from the pyheos device object and propagates it via `update_state()`. Recovery on reconnection was already handled correctly: pyheos fires `EVENT_PLAYERS_CHANGED` with a `PlayerUpdateResult` after reloading players, which triggers the existing `_handle_controller_event` → `discover_players()` path.

- related issue https://github.com/music-assistant/support/issues/5565

## Types of changes

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01RCxBC9UUetxpuHWMZv8vF1)_


MarvinSchenkel:
@Tommatheussen my issues agent came up with this fix, could you sanity check this please?

Tommatheussen:
LGTM, I can test on phyiscal devices later today if you want

MarvinSchenkel:
Would be great since I do not have any :)

Tommatheussen:
@MarvinSchenkel technically, the changes are correct, the problem is a bit deeper though. pyheos doesn't invoke the _on_disconnected all that often, seems to happen mainly when we programmatically disconnect (so a known disconnection), or when the device gets rebooted. Basically when the connection really gets cleanly closed.
Seems turning off the AVR doesn't properly close the connection, resulting in commands timing out instead. This timeout doesn't trigger the callback.
I'll check with the pyheos creator if this is intented or not, to me it makes sense that commands timing out (at least connection heartbeat commands) would indicate a broken connection.

Tommatheussen:
@MarvinSchenkel I think it's ok to continue with this one, as it will already solve some of the connection issues with HEOS devices. When the connection is closed cleanly (like rebooting the devices) this will already make sure they are marked unavailable accordingly.

I am working on an additional fix (in `pyheos) to make sure timeouts on the heartbeat, when the connection isn't cleanly closed, are also triggering the disconnection logic. Once that's in place, it will just follow the same disconnection triggers to then MA is prepared for that.

PR #4070, opened 2026-06-03
Fix AcoustID scan coverage stalling

OzGav:

The audio-analysis background scan only recorded successful analyses, so AcoustID never persisted a row for tracks it skips (already have an MBID/ISRC) or can't match. Those tracks stayed "pending" forever and were fully re-streamed and re-fingerprinted on every nightly run with no progress.

- Record already-identified tracks (existing MBID/ISRC) as an analysis result instead of silently declining, so they count as analyzed and drop out of the nightly scan for good.
- Persist a retryable no-match marker for tracks that fingerprint cleanly but find no match, so they are no longer re-streamed every night but are retried after a 60-day cooldown (the AcoustID/MusicBrainz database grows over time).

If the controller later exposed row age (or a generic retry mechanism), then we could drop the `retry_after` field and use that instead

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/acoustid_lookup/provider.py
+++ b/music_assistant/providers/acoustid_lookup/provider.py
@@ -21,6 +21,7 @@ from music_assistant_models.errors import (
 
 from music_assistant.controllers.cache import use_cache
 from music_assistant.helpers.compare import create_safe_string
+from music_assistant.helpers.datetime import utc_timestamp
 from music_assistant.helpers.tags import write_identifier_tags
 from music_assistant.helpers.throttle_retry import ThrottlerManager, throttle_with_retries
 from music_assistant.helpers.util import parse_title_and_version
@@ -31,7 +32,7 @@ from music_assistant.providers.musicbrainz import MusicbrainzProvider
 if TYPE_CHECKING:
     from music_assistant_models.config_entries import ProviderConfig
     from music_assistant_models.enums import ProviderFeature
-    from music_assistant_models.media_items import AudioFormat
+    from music_assistant_models.media_items import AudioFormat, Track
     from music_assistant_models.provider import ProviderManifest
     from music_assistant_models.streamdetails import StreamDetails
 
@@ -44,6 +45,12 @@ CONF_WRITE_TAGS_BACK = "write_tags_back"
 CONF_ANALYSE_STREAMING = "analyse_streaming"
 
 DEFAULT_MIN_SCORE = 0.85
+# Days before a track that fingerprinted but found no match is retried. The AcoustID
+# database grows over time, so an unidentifiable track may match on a later attempt.
+NO_MATCH_RETRY_DAYS = 60
+# No-match results are stored at this version so they always read as stale and are
+# re-offered to the provider, which then gates the actual retry on NO_MATCH_RETRY_DAYS.
+NO_MATCH_ANALYSIS_VERSION = -1
 MAX_FINGERPRINT_SECONDS = 120
 MAX_CANDIDATES = 5
 # Per-recording cap on stored release-groups; sized to fit popular tracks
@@ -133,13 +140,23 @@ class AcoustidLookupProvider(AudioAnalysisProvider):
         if track is None:
             self.logger.debug("Skipping %s — track is not in the library", session_id)
             return False
-        if track.mbid:
+        if track.mbid or track.get_external_id(ExternalID.ISRC):
+            # Already identified, so fingerprinting would be wasted work. Record the
+            # existing identifiers as a result so coverage counts the track and the scan
+            # does not revisit it.
+            await self._record_existing_identifiers(streamdetails, track)
             self.logger.debug(
-                "Skipping %s — track already has a MusicBrainz Recording Id", session_id
+                "Skipping %s — track already identified (MBID/ISRC present); "
+                "recorded existing identifiers",
+                session_id,
             )
             return False
-        if track.get_external_id(ExternalID.ISRC):
-            self.logger.debug("Skipping %s — track already has an ISRC", session_id)
+
+        if await self._within_no_match_cooldown(streamdetails):
+            self.logger.debug(
+                "Skipping %s — earlier AcoustID lookup found no match; still within retry cooldown",
+                session_id,
+            )
             return False
 
         fingerprinter = self._create_fingerprinter(audio_format.sample_rate, audio_format.channels)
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -65,6 +65,22 @@
...
+    Rows without a ``retry_after`` value cover their track permanently; rows carrying
+    one cover it only until that epoch (seconds) falls due. Callers must bind a ``:now``
+    parameter to the current epoch.
+
+    :param analysis_data_col: Qualified analysis_data column reference, e.g. ``aa.analysis_data``.
+    """
+    return (
+        f"(json_extract({analysis_data_col}, '$.extra_data.retry_after') IS NULL "
+        f"OR json_extract({analysis_data_col}, '$.extra_data.retry_after') > :now)"
+    )
```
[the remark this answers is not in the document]
MarvinSchenkel: Not an issue

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -65,6 +65,22 @@
...
+    Build the SQL predicate for whether a stored analysis row still covers its track.
+
+    Rows without a ``retry_after`` value cover their track permanently; rows carrying
+    one cover it only until that epoch (seconds) falls due. Callers must bind a ``:now``
+    parameter to the current epoch.
+
+    :param analysis_data_col: Qualified analysis_data column reference, e.g. ``aa.analysis_data``.
+    """
+    return (
+        f"(json_extract({analysis_data_col}, '$.extra_data.retry_after') IS NULL "
```
MarvinSchenkel: This is hacky. We should not use an `extra_data` (ie dump whatever you want) field for central functionality. Since only Acoustid would need a retry-after mechanism, I would like to solve this inside the provider (see my other comment)

music_assistant/providers/acoustid_lookup/provider.py
```diff
@@ -329,6 +339,39 @@ async def _finalize(self, session_id: str) -> AudioAnalysisData | None:
...
 
+    def _no_match_marker(self) -> AudioAnalysisData:
+        """
+        Build a no-match marker result for a track that fingerprinted but found no match.
+
+        The marker carries a ``retry_after`` epoch so the background scan reattempts the
+        track once the cooldown elapses rather than re-fingerprinting it on every run.
+        """
+        retry_after = int(utc_timestamp()) + NO_MATCH_RETRY_DAYS * 86400
+        return AudioAnalysisData(extra_data={"acoustid_no_match": True, "retry_after": retry_after})
```
MarvinSchenkel: I would save this AnalysisData with `-1` for version number, so it is offered to the AA provider the next time the analysis runs. Then your provider can just reject the analysis when the `retry_after` period has not passed yet. This way, we can keep the entire retry logic isolated in the Acoustid provider.

MarvinSchenkel (changes requested):

music_assistant/controllers/streams/audio_analysis.py
OzGav: OKey dokey I've removed this

music_assistant/providers/acoustid_lookup/provider.py
OzGav: Done as suggested. A no-match result is now stored at `analysis_version` = -1, so the track keeps getting re-offered, and the provider declines the session in `start_analysis` while it's still inside the `retry_after` window. Thus it's not re-fingerprinted until the cooldown passes. All the retry logic is isolated in the AcoustID provider now.

One supporting change was needed to make the -1 actually get re-offered: the background-scan candidate query was selecting purely on row existence, so any stored row (including the -1 one) excluded the track. I made it version-aware so a track is offered for a domain when it has no row, or its row's `analysis_version` is below that domain's current version. It keys off the `analysis_version` column only, and it's no problem for the other providers today since none of them store rows below their current version.

MarvinSchenkel:
I don't get why we still have changes in the AA Controller when we are trying to isolate this logic into the AcoustiD provider? Can you explain why we need them?

OzGav:
I think this is actually a bug, separate from AcoustID. We already track staleness as the coverage endpoint reports `stale_version`, and `start_analysis` re-runs analysis when a stored row is below the provider’s current version. But the nightly scan never acts on it, its candidate query picks tracks purely on whether a row exists, NOT on version. 

So once any row exists, the scan skips that track forever, and stale rows only ever get refreshed when a track happens to be played. After a provider bumps its `analysis_version`, the background scan won’t re-analyse anything — a track you never play keeps its outdated analysis indefinitely.

So the controller change makes the scan’s candidate query version-aware: a track is offered when it has no row, OR its row is below the provider’s current version. That fixes the stale-rescan gap for every provider, and the AcoustID -1 retry just builds on the corrected behaviour.

MarvinSchenkel:
I was talking to @chrisuthe about this and we might need to drag this centrally after all. @chrisuthe , maybe you can describe your use case here and we can see if we can come to solution that works for both? If that's too risky, we could also just merge this PR as-is and work on a retry / failed-queue mechanism for 2.10.

chrisuthe:
Sorry Y'all did not see this ping. 

There is indeed a bug in the scan (sorry @OzGav) - it is only picking up tracks that don't have an entry, it needs to compare the version and then include any tracks as candidates if the version in the AA table is lower than the version the plugin is running... 

(I would suggest that we merge that fix separately though... https://github.com/music-assistant/server/pull/4123)

The bugfix doesn't solve the short term problem of "How do I not start every scan with the same 100 tracks that I know won't resolve and then run out of time but that I want to check up on in 30 days"

Short term the proposal form @OzGav utilizes version field for a non-original purpose (-1 means come back to me) and uses the extra_data field arguably again for a non-original purpose of storing the next time this should be checked... I personally would suggest that should not be the long term solution as it leaves every other AA provider to decide to either adopt the same non-standard uses or come up with their own (plus, since this is a problem across providers adding logging/controls for this sort of scenario seems appropriate).  I wouldn't want to see 'placeholder' rows in the AA table, I'd rather keep track of non-analyzed files separately.

Could be a perfectly fine temporary fix for 2.9.0 I guess, if we're okay with a possible migration and non-standard uses for those fields.

Long Term.... I'd like to have a way to have failures _not_ get included as candidates to be analyzed if they failed previously, and allow those to be re-scanned later if the error/provider says it should (and of course selectively/customizable by the AA providers).

I've got a branch (backend/frontend) that I think accomplishes this I was holding for 2.10, I doubt we want to try to squeeze it into 2.9.0 (I've made it include 4123).

image

https://github.com/chrisuthe/server/tree/feat/aa-failure-tracking
https://github.com/chrisuthe/frontend/tree/feat/aa-failure-tracking

MarvinSchenkel (approved):

PR #4072, opened 2026-06-03
Fix Bluesound ungroup crashing on non-existent pyblu client attributes

OzGav:

While looking at typing fixes for the Bluesound provider I came across a bug in the code.

`BluesoundPlayer.ungroup()` reads `self.client.leader` and `self.client.provider.player_map(...)`. But `self.client` is the pyblu Player, which has no `leader `and no `provider `attribute so the first line throws `AttributeError`.

The fix gets the group leader from `self.sync_status.leader` and looks up its player id via the provider's player_map, mirroring how the existing `synced_to` property already does it.

This hasn't been encountered before because the normal `ungroup `command routes through `set_members` (which works) and never calls this method. The only path that calls `ungroup()` directly is universal group teardown, so it only triggers when a Bluesound player that's part of a native sync group is a member of a universal group.

I can't test this as I don't have a Bluesound device but it just aligns `ungroup()` with the working pattern already used in `synced_to` and `update_attributes`, so it's low risk.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/bluesound/player.py
+++ b/music_assistant/providers/bluesound/player.py
@@ -253,9 +253,11 @@ class BluesoundPlayer(Player):
 
     async def ungroup(self) -> None:
         """Handle UNGROUP command for BluOS player."""
-        leader = self.client.leader
-        leader_player_id = self.client.provider.player_map((leader.ip, leader.port))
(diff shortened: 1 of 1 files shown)


MarvinSchenkel (approved):
Thx Gav!

PR #4074, opened 2026-06-03
Final Typing fixes for Bluesound provider

OzGav:

Last set of typing fixes

Points to note

  - `cast("str", …)` on a few pyblu fields. pyblu types these as `str | None`. The casts keep the existing behaviour exactly (these values were already used as-is); for the `dict.get(...)` keys a `None `is harmless. Whether those fields can actually be None mid-playback is a pre-existing question, not something introduced here but happy to follow up separately if you'd prefer real guards.
  - `int(source_id)` in `select_source`. pyblu's `load_preset` is typed `preset_id: int`.  `source_id` is the numeric string form of `preset.id` so this produces an identical HTTP request
  - duration cast keeps the float as-is (an inline comment notes that `int()` would change precision).
  - Two `# type: ignore` remain, both commented inline: `close()` (pyblu upstream lacks a `-> None` return annotation) and the sub-second `poll_interval = 0.5` assignment (base attr is typed int).

Upstream PR for pyblu is here https://github.com/LouisChrist/pyblu/pull/75

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/bluesound/player.py
+++ b/music_assistant/providers/bluesound/player.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import time
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 from music_assistant_models.config_entries import ConfigEntry, ConfigValueType
 from music_assistant_models.enums import IdentifierType, PlaybackState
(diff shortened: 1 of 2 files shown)


OzGav:
@Cyanogenbot FYI

MarvinSchenkel (approved):

PR #4081, opened 2026-06-04
Type throttle_with_retries via Protocol instead of Provider bound

OzGav:

Inline with the recent typing changes to the cache decorator (`@use_cache`, which uses a structural `_Cacheable` Protocol), this applies the same treatment to `throttle_with_retries`.

The decorator only ever touches` self.logger` and `self.throttler`, but it bounded its type var to the nominal Provider class. Helper API-client classes that hold a throttler without subclassing Provider (apple_music, tidal, yousee) therefore couldn't satisfy the bound and needed `# type: ignore[type-var]` on every decorated method. The decorator itself also needed a `# type: ignore[attr-defined]` on `self.throttler`, because Provider doesn't actually declare a throttler attribute.

This replaces the Provider bound with a structural `_Throttleable` Protocol (logger + throttler), so any duck-typed class with those two attributes satisfies it directly

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/throttle_retry.py
+++ b/music_assistant/helpers/throttle_retry.py
@@ -12,7 +12,7 @@ from contextlib import asynccontextmanager
 from contextvars import ContextVar
 from email.utils import parsedate_to_datetime
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Concatenate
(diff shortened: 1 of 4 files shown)


MarvinSchenkel (approved):
LGTM

PR #4067, opened 2026-06-03
Treat Retry-After as a floor for rate limits, not an exact target

rnewman:
# What does this implement/fix?

When extending the Bandcamp integration I noticed bad behavior: lots of requests, which triggered a Bandcamp rate limit response, after which we make 10 retries 3 seconds apart, followed by long-lasting failure.

The existing throttle/retry helper treats the server `backoff_time` as a target, not a floor, which is neither the intent of the `Retry-After` header in a 429, nor useful behavior in these real scenarios.

Rather than hack something in for Bandcamp, this PR attempts to shift the overall codebase a little towards the right state without breaking too many things.

I chose to introduce a `RateLimited` error in `models`, raise it for 429 HTTP responses distinctly from `ResourceTemporarilyUnavailable`, and adjust the throttle/retry helper to apply exponential backoff in this case.

Existing users of `ResourceTemporarilyUnavailable` — transient errors like 503s, and odd users like Spotify's 1-second token refresh — keep the existing behavior.

Additionally, this PR makes three complementary changes:

1. Apply additive jitter to these backoff times. This is generally good practice as a citizen of the internet, helping to avoid thundering herds in the case of servers not already configured to add jitter.
2. Cap long `Retry-After` values to one hour; again, a concession to practicality.
3. Drop Bandcamp's retries from 10 to 5.

Requires `music-assistant/models` with `RateLimited` (error_code 25).

**Related issue (if applicable):**

- [Models change](https://github.com/music-assistant/models/pull/242) — unit tests will fail until this lands.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [x] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pre-commit run --all-files` passes.
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked.
[...]

PR #4089, opened 2026-06-04
Apple Music: stream library tracks and harden transient-error handling

teancom:

I have an Apple Music library of about 96k tracks, very near their upper-limit of 100k. I've had issues in the past where the 'tracks sync' job times out / 'crashes' / etc., but if I left it alone it would eventually muddle through. However, after https://github.com/music-assistant/server/pull/4006 was merged it started failing completely, which I tracked down to MA being OOMkilled once I hit around 70k tracks through the sync, because the added enrichment + holding all of it in RAM until the end meant my HA's server 8GB was not enough. This PR incorporates the recent changes from #4067, fixes up the retry logic to smooth out Apple Music's rough edges (the "further" you get into your library pagination-wise, the slower the API responses get for instance), and (most importantly for large libraries like mine) writes as it goes so that we don't store the entire library's worth of tracks in RAM in one go (my server now stays flat at 5GB used for the entire three and a half hour run).

For testing, I did 10 full runs with various settings for the number of tracks to return per page, and 30 had the best ratio of "wall clock time" vs. "Retries". I *did* see all of the various errors we handle during my testing, from 429s to timeouts, partial / empty responses, etc., and they are now cleanly handled via retry (with backoff!) or treated as fatal (mostly when we run out of retries, which I did multiple times until lowering the track limit to 30).

The library sync materialized the whole ~100k-track listing in memory with heavy includes (~87KB/track), OOM-killing the server around 70k tracks on 8GB HA/MA hosts. Stream the listing in bounded windows instead, enriching and yielding each window so the full library is never held at once.

Also retry transient Apple API failures with exponential backoff (504/500/429 and dropped-connection/timeout transport errors), and re-fetch mid-pagination truncations rather than dropping items. Page tracks at 30 vs the 50 default to reduce deep-offset 504s.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/apple_music/api_client.py
+++ b/music_assistant/providers/apple_music/api_client.py
@@ -2,13 +2,14 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, cast
+import functools
+from collections.abc import AsyncGenerator, Awaitable, Callable
+from typing import TYPE_CHECKING, Any, Concatenate, cast
 
-from aiohttp import ClientTimeout
+from aiohttp import ClientConnectionError, ClientPayloadError, ClientTimeout
 from music_assistant_models.enums import MediaType
 from music_assistant_models.errors import (
     MediaNotFoundError,
-    MusicAssistantError,
     RateLimited,
     ResourceTemporarilyUnavailable,
 )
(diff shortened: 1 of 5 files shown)


OzGav:
Just checking as a bunch of typing fixes were just merged. Did you work from that base?

teancom:
@OzGav I rebased a couple of times, and did full runs to smoke-test after each. So it's been tested against every commit except 758937819 which appears to just be YouTube music stuff.

MarvinSchenkel:
Will have a look, for now removed the backport label since we have a 2.9 release next Wednesday

MarvinSchenkel (approved):
Thanks @teancom 🎉

PR #4114, opened 2026-06-06
Final typing fixes for the Music controller

OzGav:

Stage 3 of 3 for the typing fixes.

Points to note

1. Casting the per-type controller to its generic base — `cast("MediaControllerBase[MediaItemType]", ctrl)`. `get_controller()` returns one of eight controllers (artists, albums, tracks, …), and each only accepts its own kind of item. When the code picks a controller by media_type and then calls add_item_to_library / update_item_in_library / match_providers, the type checker can't see that the controller and the item match, so it rejects the call. The cast states what the code already guarantees: the controller was chosen from the item's own media_type, so they always line up at runtime. It's a no-op at runtime and the dispatch itself is unchanged.

2. `browse()` now declares `ItemMapping` in its return type. This widens the public (`@api_command`) return to `Sequence[MediaItemType | ItemMapping | BrowseFolder]`. It isn't a workaround it makes the signature honest. Provider `browse()` can already return `ItemMapping` values (via the recommendations path), and the one internal caller already handles them. So the type now matches what the method actually returns today

3. One small change outside the controller, in `helpers/util.py`. Taking `music.py` out of the mypy exclude list meant its calls into `TaskManager` were type-checked for the first time, which surfaced that it passes the manager a coroutine that returns a value, while the signature only allowed None-returning ones. `TaskManager` runs these fire-and-forget and discards the result, so widening the accepted coroutine type (`Coroutine[Any, Any, None] → …, Any]`) is correct and that's the only reason `util.py` appears in this PR.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -44,6 +44,7 @@ from music_assistant_models.media_items import (
     Playlist,
     Podcast,
     ProviderMapping,
+    Radio,
     RecommendationFolder,
     SearchResults,
     Track,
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
Thanks @OzGav !

PR #4116, opened 2026-06-07
Fix Universal Group Player producing no audio on some members

OzGav:

When playing to a Universal Group, members that can't use chunked streaming (e.g. Squeezelite, some older Chromecast firmware) got no sound. Universal groups always streamed using chunked encoding to every member, regardless of each member's own configured streaming style so any member that couldn't handle chunked failed silently, and the whole group then stopped after a few seconds.

Each member already fetches the audio over its own connection, so the streaming style can safely differ per member. We now pick it from each member's own player settings — the same one it uses for normal (non-group) playback — instead of forcing chunked on everyone. We also guard against the specific crash that left HTTP/1.0 members silent.

- related issue https://github.com/music-assistant/support/issues/5214

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/universal_group/player.py
+++ b/music_assistant/providers/universal_group/player.py
@@ -583,12 +583,15 @@ class UniversalGroupPlayer(Player):
         # use each member's own config to support mixed members (e.g. a chunked
         # Chromecast alongside an HTTP/1.0 Squeezelite). Fall back to the group config
         # when the request has no player_id.
+        http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
         if child_player_id:
-            http_profile = await self.mass.config.get_player_config_value(
-                child_player_id, CONF_HTTP_PROFILE, default="default", return_type=str
-            )
-        else:
-            http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
+            try:
+                http_profile = await self.mass.config.get_player_config_value(
+                    child_player_id, CONF_HTTP_PROFILE, default=http_profile, return_type=str
+                )
+            except KeyError:
+                # player_id may be stale/invalid; fall back to the group profile
+                pass
         if http_profile == "chunked" and request.version < HttpVersion11:
             # chunked encoding is not allowed on HTTP/1.0; fall back to
             # connection-close streaming to avoid raising in resp.prepare()
--- a/music_assistant/providers/universal_group/player.py
+++ b/music_assistant/providers/universal_group/player.py
@@ -7,7 +7,7 @@ from copy import deepcopy
 from time import time
 from typing import TYPE_CHECKING, cast
 
-from aiohttp import web
+from aiohttp import HttpVersion11, web
 from music_assistant_models.config_entries import ConfigEntry, ConfigValueOption, ConfigValueType
 from music_assistant_models.constants import PLAYER_CONTROL_FAKE, PLAYER_CONTROL_NONE
 from music_assistant_models.enums import (
@@ -580,6 +580,17 @@ class UniversalGroupPlayer(Player):
         }
         resp = web.StreamResponse(status=200, reason="OK", headers=headers)
         http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
+        if http_profile == "chunked" and request.version < HttpVersion11:
+            # chunked transfer-encoding is forbidden for HTTP/1.0 clients (e.g. some
+            # Squeezelite and older Chromecast firmware). Unlike the per-player flow
+            # stream, every UGP member shares the group's single http_profile, so an
+            # HTTP/1.0 member would otherwise crash on resp.prepare() and get no audio.
+            # Fall back to connection-close streaming so these members still play.
+            self.logger.debug(
+                "Disabling chunked encoding for UGP stream to HTTP/1.0 client %s",
+                child_player_id or request.remote,
+            )
+            http_profile = "no_content_length"
         if http_profile == "forced_content_length":
             # some clients (notably older Chromecast firmware) refuse to play unless
             # they see a Content-Length header up front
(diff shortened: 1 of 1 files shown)


music_assistant/providers/universal_group/player.py
```diff
@@ -579,7 +579,24 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.StreamResponse:
...
+        else:
+            http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
+        if http_profile == "chunked" and request.version < HttpVersion11:
+            # chunked encoding is not allowed on HTTP/1.0; fall back to
+            # connection-close streaming to avoid raising in resp.prepare()
+            self.logger.debug(
+                "Disabling chunked encoding for UGP stream to HTTP/1.0 client %s",
+                child_player_id or request.remote,
+            )
+            http_profile = "no_content_length"
```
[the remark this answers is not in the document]
OzGav: The `universal_group` provider has zero existing test coverage, so there's nothing to extend. I plan to pass on this suggestion.

music_assistant/providers/universal_group/player.py
```diff
@@ -579,7 +579,24 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.StreamResponse:
...
+        # http_profile is per-connection, not part of the shared encoded stream, so
+        # use each member's own config to support mixed members (e.g. a chunked
+        # Chromecast alongside an HTTP/1.0 Squeezelite). Fall back to the group config
+        # when the request has no player_id.
+        if child_player_id:
+            http_profile = await self.mass.config.get_player_config_value(
+                child_player_id, CONF_HTTP_PROFILE, default="default", return_type=str
+            )
+        else:
+            http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
```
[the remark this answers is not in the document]
OzGav: Fair. Accepting suggestion.

music_assistant/providers/universal_group/player.py
```diff
@@ -579,7 +580,25 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.StreamResponse:
             "Content-Type": get_mime_type(output_format_str),
         }
         resp = web.StreamResponse(status=200, reason="OK", headers=headers)
+        # http_profile is per-connection, not part of the shared encoded stream, so
+        # use each member's own config to support mixed members (e.g. a chunked
+        # Chromecast alongside an HTTP/1.0 Squeezelite). Fall back to the group config
+        # when the request has no player_id.
```
marcelveldt:
```suggestion
```
No need for the too verbose comments

music_assistant/providers/universal_group/player.py
```diff
@@ -579,7 +580,25 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.StreamResponse:
...
         http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
+        if child_player_id:
+            # player_id may be stale/invalid; fall back to the group profile
+            with contextlib.suppress(KeyError):
+                http_profile = await self.mass.config.get_player_config_value(
+                    child_player_id, CONF_HTTP_PROFILE, default=http_profile, return_type=str
+                )
+        if http_profile == "chunked" and request.version < HttpVersion11:
```
marcelveldt, line 591: Instead if silently ignoring a misconfigured client, we should consider raising an exception with a meaningful message, so the user can adjust the player config.

This is also the case for the other stream paths btw; it would make sense to add this same defense here as well

music_assistant/providers/universal_group/player.py
```diff
@@ -579,7 +580,25 @@ async def _serve_ugp_stream(self, request: web.Request) -> web.StreamResponse:
             "Content-Type": get_mime_type(output_format_str),
         }
         resp = web.StreamResponse(status=200, reason="OK", headers=headers)
+        # http_profile is per-connection, not part of the shared encoded stream, so
+        # use each member's own config to support mixed members (e.g. a chunked
+        # Chromecast alongside an HTTP/1.0 Squeezelite). Fall back to the group config
+        # when the request has no player_id.
         http_profile = cast("str", self.config.get_value(CONF_HTTP_PROFILE, "chunked"))
+        if child_player_id:
```
marcelveldt, line 585:
```suggestion
        # prefer the child (protocol) player configuration
        if child_player_id:
```

music_assistant/providers/universal_group/player.py
OzGav: Thanks Marcel, that makes sense and I'm happy to go with raising instead of silently falling back.

One thing I wanted to check with you so I get it right. I think the only case that actually fails here is when the `player_id` from the query string doesn't resolve to a known player at all (for example a member that's been removed mid-stream, or a stale URL). I say that because the `http_profile` value itself never raises. `get_player_config_value` returns whatever value is set, so the only thing that throws is the lookup for an unknown player. So I don't think it's really a misconfiguration the user can fix in the player settings? Given that, would a clear `HTTPNotFound` ("Unknown player: ...") be the right thing to raise? I think that would also be consistent with the check already at the top of `_serve_ugp_stream` for the group player itself.

On the other stream paths, I think they're already covered. When I looked, the flow, queue-flow and announcement handlers in the stream controller all validate the player up front and raise `HTTPNotFound` before reading the config, so the UGP path looks like the odd one out because it never validated the child `player_id`. Having said that, you know this code far better than I do, so if there's a case I'm missing, point me at it and I'll add the same guard.

Happy to push this once you confirm the approach.

marcelveldt: yeah unknown player can safely result to the group leader default

OzGav: Perfect, thanks. That's actually what the current code already does.` http_profile` starts as the group default and only gets overridden if the child player's config resolves, so an unknown player keeps the group leader default. So I think we're good as-is?

MarvinSchenkel (approved):
LGTM, thanks @OzGav .

PR #4126, opened 2026-06-08
Spread metadata maintenance schedule across the day

MarvinSchenkel:

The daily metadata maintenance tasks were hard-scheduled at 04:00 local time. The
"scan missing artist metadata" task queries the shared MusicBrainz mirror
(`musicbrainz-mirror.music-assistant.io`), so every instance firing at the same
wall-clock minute caused a request spike on the mirror.

This spreads the schedule across the full day using a per-instance random time.
Since the schedule is persisted and restored on boot, existing installs are
migrated rather than left stuck on their old 04:00 time.

- Pick a random daily time across the full 24h instead of a fixed 04:00
- Register the maintenance tasks under new (`_v2`) ids so existing installs drop
  the persisted 04:00 schedule and pick up the randomized one
- Clean up the orphaned legacy task state in `ConfigController._migrate`
- Add/update tests

- N/A

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -1427,8 +1427,39 @@ class ConfigController:
                 )
                 changed = True
 
+        # Drop the persisted schedule for the metadata maintenance tasks that were hardcoded
+        # to run at 04:00 local. They are now registered under new ("_v2") task ids with a
+        # randomized full-day schedule (to avoid spiking the shared musicbrainz mirror), so the
+        # old persisted state is orphaned and can be removed.
+        # TODO: remove after 2.10 release
+        if self._migrate_metadata_maintenance_schedule():
+            changed = True
+
         if changed:
             await self._async_save()
+
+    def _migrate_metadata_maintenance_schedule(self) -> bool:
+        """Remove the orphaned persisted state for the pre-randomization metadata task ids."""
+        core_config = self._data.get(CONF_CORE)
+        if not isinstance(core_config, dict):
+            return False
+        tasks_config = core_config.get("tasks")
+        if not isinstance(tasks_config, dict):
+            return False
+        task_states = tasks_config.get("scheduled_task_states")
+        if not isinstance(task_states, dict):
+            return False
+        legacy_task_ids = (
+            "metadata_missing_artist_metadata_scan",
+            "metadata_playlist_metadata_scan",
+            "metadata_thumb_cache_cleanup",
+        )
+        removed = [task_id for task_id in legacy_task_ids if task_id in task_states]
+        for task_id in removed:
+            del task_states[task_id]
+        if removed:
+            LOGGER.info("Removed orphaned metadata maintenance schedule state for %s", removed)
+        return bool(removed)
 
     def _migrate_fully_kiosk_multi_instance(self) -> bool:
         """Collapse legacy multi-instance Fully Kiosk configs into a single provider instance."""
--- a/music_assistant/controllers/config.py
+++ b/music_assistant/controllers/config.py
@@ -1431,7 +1431,7 @@ class ConfigController:
         # to run at 04:00 local. They are now registered under new ("_v2") task ids with a
         # randomized full-day schedule (to avoid spiking the shared musicbrainz mirror), so the
         # old persisted state is orphaned and can be removed.
-        # TODO: remove after 2.10 release
+        # TODO: remove after 2.9 release
         if self._migrate_metadata_maintenance_schedule():
             changed = True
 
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/config.py
```diff
@@ -1427,9 +1427,40 @@ async def _migrate(self) -> None:
...
+        if not isinstance(tasks_config, dict):
+            return False
+        task_states = tasks_config.get("scheduled_task_states")
+        if not isinstance(task_states, dict):
+            return False
+        legacy_task_ids = (
+            "metadata_missing_artist_metadata_scan",
+            "metadata_playlist_metadata_scan",
+            "metadata_thumb_cache_cleanup",
+        )
```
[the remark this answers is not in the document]
MarvinSchenkel: Thanks, but I'll keep these hard-coded here. This migration is a snapshot of historical state: it must always target the *old* (pre-`_v2`) ids, which no longer have a live constant at all. Importing the current `metadata.py` constants would actually point at the new `_v2` ids and break the cleanup, and if those constants ever change again the migration should still reference the original strings. This also matches the existing migrations in this file (fully_kiosk, default_enqueue_option_radio), which hard-code their legacy keys for the same reason.

tests/core/test_background_tasks.py
```diff
@@ -65,7 +70,7 @@ async def _wait_for_task_status(
 
 
 @pytest.fixture
-async def tasks_controller(mass_minimal: MusicAssistant) -> AsyncGenerator[TasksController, None]:
+async def tasks_controller(mass_minimal: MusicAssistant) -> AsyncGenerator[TasksController]:
```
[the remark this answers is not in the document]
MarvinSchenkel: This is pre-existing code that isn't touched by this PR, and it follows the existing fixture pattern already used throughout this file. `pre-commit` (including mypy with strict generics) passes on this branch — on the project's pinned Python 3.14 typeshed, `AsyncGenerator`'s send/return type params have PEP 696 defaults, so a single type argument is valid. Leaving as-is to avoid unrelated churn.

music_assistant/controllers/config.py
```diff
@@ -1427,9 +1427,40 @@ async def _migrate(self) -> None:
                 )
                 changed = True
 
+        # Drop the persisted schedule for the metadata maintenance tasks that were hardcoded
+        # to run at 04:00 local. They are now registered under new ("_v2") task ids with a
+        # randomized full-day schedule (to avoid spiking the shared MusicBrainz mirror), so the
+        # old persisted state is orphaned and can be removed.
+        # TODO: remove after 2.9 release
+        if self._migrate_metadata_maintenance_schedule():
```
marcelveldt, line 1435: not for now but in the next release cycle we should refactor the config controller and isolate all migrations into its own module/file.

music_assistant/controllers/metadata.py
```diff
@@ -219,9 +218,10 @@ def _is_safe_imageproxy_request_path(path: str) -> bool:
 CONF_ENABLE_ONLINE_METADATA = "enable_online_metadata"
 CONF_PREFER_LOCAL_GENRES = "prefer_local_genres"
 CONF_ENABLE_RADIO_METADATA_LOOKUP = "enable_radio_metadata_lookup"
-MISSING_ARTIST_METADATA_SCAN_TASK_ID = "metadata_missing_artist_metadata_scan"
-PLAYLIST_METADATA_SCAN_TASK_ID = "metadata_playlist_metadata_scan"
-THUMB_CACHE_CLEANUP_TASK_ID = "metadata_thumb_cache_cleanup"
+# "_v2" ids drop the legacy 04:00 schedule on existing installs (see ConfigController._migrate).
```
marcelveldt:
```suggestion
```
irrelevant

tests/core/test_background_tasks.py
```diff
@@ -449,13 +456,22 @@ async def test_core_maintenance_tasks_register_nightly_schedules(
...
     assert artist_scan_task.translation_key == "background_task.scan_missing_artist_metadata"
-    assert artist_scan_task.schedule == maintenance_schedule
     assert artist_scan_task.metadata == {"task_domain": "metadata_missing_artist_metadata_scan"}
 
     assert playlist_scan_task.translation_key == "background_task.refresh_playlist_metadata"
-    assert playlist_scan_task.schedule == maintenance_schedule
     assert playlist_scan_task.metadata == {"task_domain": "metadata_playlist_metadata_scan"}
 
+    # Metadata maintenance tasks pick a random time spread across the full day (instead of a
+    # fixed 04:00) to avoid spiking the shared MusicBrainz mirror, but share one time per instance.
```
marcelveldt:
```suggestion
    # Metadata maintenance tasks pick a random time spread across the full day
    # to avoid spiking the shared MusicBrainz mirror, but share one time per instance.
```

marcelveldt (approved):

PR #4127, opened 2026-06-08
Don't enqueue next track onto a stopped queue

MarvinSchenkel:

Fixes playback resuming by itself a while after the user stops or pauses, observed on WiiM (LinkPlay) players.

A next-track preload — notably Smart Fades crossfade/beat analysis, which can take tens of seconds — can finish well after the queue was stopped or paused. The enqueue-next guard only checked `queue.active` and `session_id`, neither of which a stop/pause changes, so it still called `enqueue_next_media`. On WiiM/LinkPlay players, issuing `SetNextAVTransportURI` on a stopped device makes the device start playing, so playback resumed on its own seconds-to-minutes after the user stopped/paused. (Per the UPnP AVTransport spec, staging a next URI should not start playback — this is firmware behaviour, but MA should not be enqueuing onto a non-playing queue regardless.)

Fix (defense in depth):
- Skip enqueue-next when the queue is no longer playing (`queue.state != PLAYING`).
- Cancel the in-flight preload/enqueue-next tasks in `stop()` so a slow preload finishing after stop can't enqueue.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues.py
+++ b/music_assistant/controllers/player_queues.py
@@ -683,6 +683,10 @@ class PlayerQueuesController(CoreController):
         self._check_player_permission(queue_id)
         # cancel any pending play_index calls for this queue to prevent conflicts
         self.mass.cancel_timer(f"queue_play_index_{queue_id}")
+        # cancel in-flight preload/enqueue-next so it can't enqueue after stop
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
Solution LGTM but note a few copilot comments

PR #4129, opened 2026-06-08
Fix volume jump when crossfade intro and body normalize differently

MarvinSchenkel:
## Problem

In **per-item crossfade mode**, the incoming track is streamed **twice**:

1. As the **crossfade fade-in** at *prep time* (during the previous track), the result of which is baked into `CrossfadeData` and replayed at the start of the track.
2. As the track's **own body** when it actually plays.

Each call computes volume normalization independently from `streamdetails.loudness` *as known at that moment*. For a track whose EBU R128 loudness has **not been measured yet**, the intro is generated with `DYNAMIC` (`loudnorm`) under the default `fallback_dynamic`. If the measurement then lands *before the body plays* (the analyzer finalizes mid-transition), the body re-evaluates to `MEASUREMENT_ONLY` with a static gain. The replayed intro (loudnorm) and the body (static gain) no longer match → an **audible volume jump at the intro→body seam**, typically ~20–45s into the track.

This was confirmed from production logs: a cold YouTube Music track (`We Are All Astronauts` – Doves) had its loudness measured fresh *after* the crossfade intro was committed but *before* the body played (−11.4 LUFS, LRA 14.4). The intro played on loudnorm (which boosted the quiet intro), the body applied a static −5.6 dB → the jump went **down**. The analyzer only runs when no stored measurement exists, so the fresh measurement proves the value was absent when the intro was baked.

## Fix

- Record the normalization **mode the intro was baked with** on `CrossfadeData.normalization_mode`.
- Add `normalization_override` to `get_queue_item_stream`, which skips the just-in-time loudness hydration + mode re-evaluation and forces the given mode.
- On replay, when the intro was `DYNAMIC`, **pin the body to `DYNAMIC`** too, collapsing `DYNAMIC → MEASUREMENT_ONLY` into `DYNAMIC → DYNAMIC`.

Static intro modes (`MEASUREMENT_ONLY` / `FIXED_GAIN`) already match the body, so they're left to re-evaluate normally and can still adopt a fresher measurement. Flow mode is unaffected (it normalizes the track once). `CrossfadeData` created before this change defaults `normalization_mode=None` → no override → previous behavior (backward-compatible).

## Tests

`tests/controllers/streams/test_normalization_override.py`:
- DYNAMIC override forces `loudnorm` and ignores an available measurement (hydration/re-eval skipped).
- No override re-evaluates to `MEASUREMENT_ONLY` with the expected `volume=-5.6dB`.
- `CrossfadeData.normalization_mode` defaults to `None` and round-trips a pinned value.

## Notes

- Pre-commit / pytest were **not** run locally (this machine lacks the pinned Python 3.14 + `uv` toolchain). CI / a local `pre-commit run --all-files` + `pytest tests/controllers/streams/test_normalization_override.py` should validate.
- Residual: on very high-LRA tracks, `DYNAMIC → DYNAMIC` still has a small transient seam (two independent single-pass `loudnorm` instances), but far smaller than a multi-dB static step. Fully removing it would require not splitting the track into two streams (out of scope).

### DIFF
--- a/music_assistant/controllers/streams/audio.py
+++ b/music_assistant/controllers/streams/audio.py
@@ -142,6 +142,11 @@ class CrossfadeData:
     queue_item_id: str
     # Offset for the fade_in track's elapsed time calculation, to account for crossfade duration and trim
     elapsed_time_offset: float = 0.0
+    # Volume normalization mode the intro PCM ('data') was baked with at crossfade-build time.
+    # Used to pin the next track's body to the same mode, so the replayed intro and the body
+    # cannot disagree (e.g. a DYNAMIC intro followed by a MEASUREMENT_ONLY body once the
+    # loudness measurement lands mid-transition, which causes an audible volume jump).
+    normalization_mode: VolumeNormalizationMode | None = None
 
 
 def _snap_supported_rate_up(target: int, supported_sample_rates: list[int]) -> int:
@@ -1451,6 +1456,7 @@ class StreamsAudio:
         seek_position: int = 0,
         playback_speed: float = 1.0,
         raise_on_error: bool = True,
+        normalization_override: VolumeNormalizationMode | None = None,
     ) -> AsyncGenerator[bytes]:
         """
         Get the (PCM) audio stream for a single queue item.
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/streams/audio.py
```diff
@@ -1740,6 +1751,11 @@ async def get_queue_item_stream_with_smartfade(
         crossfade_buffer_size = (crossfade_buffer_size // frame_size) * frame_size
         fade_out_data: bytes | None = None
 
+        # pin the body to DYNAMIC when the intro was baked DYNAMIC, else a late measurement flips it and causes a volume jump
```
```suggestion
        # pin the body to DYNAMIC when the intro was baked DYNAMIC,
        # else a late measurement flips it and causes a volume jump
```

tests/controllers/streams/test_normalization_override.py
marcelveldt, line 1: what an (AI generated) overcomplicated test - can we simplify it ?

marcelveldt (approved):

PR #4130, opened 2026-06-08
Send Sendspin album artwork for radio and Spotify Connect streams

maximmaxim345:
Album artwork was only sent inside the `queue_item` branch, so radio and Spotify Connect streams never updated it. Instead it is now sent based on `current_media.image_url`, so radio shows the now-playing cover instead of the static station logo.

### DIFF
--- a/music_assistant/providers/sendspin/player.py
+++ b/music_assistant/providers/sendspin/player.py
@@ -829,23 +829,15 @@ class SendspinPlayer(SendspinBasePlayer):
                     f.cancel()
         # self.group_members will be updated by the group event callback
 
-    async def _send_album_artwork(
-        self, current_media: PlayerMedia, queue_item: QueueItem | None
-    ) -> str | None:
+    async def _send_album_artwork(self, current_media: PlayerMedia) -> str | None:
         """
         Send album artwork to the sendspin group.
 
         Args:
             current_media: The current player media.
-            queue_item: The current queue item, or None for raw streams
-                (radio, Spotify Connect) with no resolved MA queue item.
         """
-        # PlayerMedia.image_url is synthesized from stream_metadata.image_url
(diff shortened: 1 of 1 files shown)


music_assistant/providers/sendspin/player.py
```diff
@@ -829,29 +829,36 @@ async def set_members(
...
-                    if (artwork_role := self._artwork_role) is not None:
-                        await artwork_role.set_album_artwork(image)
-            # Clear artwork if none available
+            image_data: bytes | None = None
+            if artwork_url is not None:
+                # Fetch from the resolved URL so the bytes match artwork_url, even when
+                # radio now-playing art differs from the queue item's own image.
+                fetched = await self.mass.metadata.get_thumbnail(artwork_url, provider="builtin")
+                if isinstance(fetched, bytes):
+                    image_data = fetched
```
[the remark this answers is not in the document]
maximmaxim345: Pre-existing and not specific to this PR: the same unguarded fetch exists across image/playlist/feed fetch paths.
But also don't think it would even be exploitable in either case since the response is never returned to the caller.

music_assistant/providers/sendspin/player.py
```diff
@@ -829,29 +829,36 @@ async def set_members(
                     f.cancel()
         # self.group_members will be updated by the group event callback
 
-    async def _send_album_artwork(self, current_item: QueueItem) -> str | None:
+    async def _send_album_artwork(
+        self, current_media: PlayerMedia, queue_item: QueueItem | None
```
marcelveldt: you can simplify all this logic by simply using the player.state.current_media (or player.__final_current_media) as that handles all this already for all sources

maximmaxim345: Done.

MarvinSchenkel (approved):
LGTM!

PR #4133, opened 2026-06-08
Adjust Chromecast playback defaults (HTTP Profile 3 + flow mode)

MarvinSchenkel:

Adjusts two Chromecast player defaults for more reliable playback. Both remain per-player settings users can override, and existing players keep their stored values.

- **HTTP Profile 3 (forced content length)** instead of Profile 2 (no content length). Some Cast devices play unreliably without a content length (e.g. streams stopping partway). This matches what providers like Bluesound already use.
- **Flow mode enabled by default.** Cast devices handle a continuous queue flow stream more reliably than enqueueing individual tracks (same rationale as DLNA).

- N/A

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/chromecast/constants.py
+++ b/music_assistant/providers/chromecast/constants.py
@@ -6,7 +6,7 @@ from music_assistant_models.config_entries import ConfigEntry
 from music_assistant_models.enums import ConfigEntryType
 
 from music_assistant.constants import (
-    CONF_ENTRY_HTTP_PROFILE,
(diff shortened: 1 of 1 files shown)


marcelveldt:
I would also default flow mode for cast players because the downside of profile 3 is that we need to guess the content length, resulting in gaps between tracks

marcelveldt (approved):
It is kind of lame that this is needed but if people report this works for them, then lets go.
Hopefully Google fixes this mess later again.

PR #4148, opened 2026-06-09
Lastfm improvements

OzGav:

The "Discover Similar Tracks" row seeds from your most-played tracks, but niche tracks silently fail Last.fm's `track.getSimilar`, so the row quietly collapses onto whichever one or two seeds Last.fm actually recognises and ignores the rest of your listening. So we now fetch more seeds (top 5 → 10) so enough survive to build a representative row. The same seeds are now also filtered to tracks that have actually been played, so a brand-new library full of unplayed tracks doesn't produce a row from arbitrary zero-play seeds.

The genre rows were seeded by date, so they only changed once a day no matter how often recommendations rebuilt. They're now seeded by date + hour, so the sampled items rotate each hour. In-library items are still filtered out before the expensive resolution step, so the extra rotation only ever resolves genuinely external items, which then cache so no wasted lookups and no bigger burst of API calls per rebuild.

With that rotation in place we can also widen the genre chart pool (40 → 60). It's still a single fetch and the same number resolved per rebuild, so the extra items are resolved gradually across the day's rotations and cached, surfacing more of the genre over time for no extra API cost at row build time.

Finally, the genre rows only ever used your single top genre. They now pull your top 3 and cycle through them day by day, so the genre on offer varies instead of being fixed to one tag.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/lastfm_recommendations/api_client.py
+++ b/music_assistant/providers/lastfm_recommendations/api_client.py
@@ -21,6 +21,8 @@ from music_assistant.providers.lastfm_recommendations.constants import CONF_API_
 _DEFAULT_API_KEY: str = app_var(12)
 
 if TYPE_CHECKING:
+    import logging
+
     from aiohttp import ClientSession
 
     from music_assistant.providers.lastfm_recommendations import LastFMRecommendationsProvider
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
LGTM

PR #4156, opened 2026-06-10
Fix cache cleanup missing most records and skip needless startup vacuum

MarvinSchenkel:

Users with very large libraries hit out-of-memory errors during cache database compaction, and most expired cache entries were never cleaned up — so the cache kept growing.

Two issues:

- **Cleanup only scanned the first 500 records.** `auto_cleanup` fetched rows via `get_rows`, which defaults to `limit=500`, so on large caches the vast majority of expired entries were never deleted and accumulated indefinitely. It now deletes expired entries in a single set-based `DELETE` statement, which also avoids materializing the (potentially huge) cached data blobs in memory. Entries marked `allow_expired_cache` are still preserved for stale-while-revalidate.
- **Startup VACUUM ran unconditionally.** A full vacuum rebuilds the whole database, which is wasteful (and memory-hungry) when little space is actually free. The startup vacuum now runs only when at least 20% of the database file can be reclaimed (a cheap freelist check), so it no longer does a full rebuild for negligible gain.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/cache/constants.py
+++ b/music_assistant/controllers/cache/constants.py
@@ -13,6 +13,8 @@ CONF_CLEAR_CACHE = "clear_cache"
 DEFAULT_CACHE_EXPIRATION = 86400 * 30  # 30 days
 DB_SCHEMA_VERSION = 8
 MAX_CACHE_DB_SIZE_MB = 2048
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):
LGTM!

PR #4166, opened 2026-06-10
Automatically check if CPU is supported for Audio Analysis

chrisuthe:

Gate incompatible hardware for AA providers.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -145,7 +145,9 @@ def verify_cpu_supports_ml_inference() -> None:
     if torch.backends.cpu.get_cpu_capability() not in ("AVX2", "AVX512"):
         raise SetupFailedError(
             "On-device audio analysis requires a CPU with AVX2 support "
-            "(Intel Haswell / AMD Zen or newer). This CPU does not support AVX2."
+            "(Intel Haswell / AMD Zen or newer). This CPU does not support AVX2. "
+            "If you are running in a virtual machine (e.g. Proxmox), changing the "
+            "CPU type to 'host' may expose AVX2 to the guest."
         )
 
 
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -142,7 +142,9 @@ def verify_cpu_supports_ml_inference() -> None:
         return
     import torch  # noqa: PLC0415
 
-    if torch.backends.cpu.get_cpu_capability() not in ("AVX2", "AVX512"):
+    # denylist of known-insufficient capabilities so a future torch version
+    # reporting a new, higher capability string fails open instead of closed
+    if torch.backends.cpu.get_cpu_capability() in ("DEFAULT", "NO AVX"):
         raise SetupFailedError(
             "On-device audio analysis requires a CPU with AVX2 support "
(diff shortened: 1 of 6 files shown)


music_assistant/helpers/util.py
```diff
@@ -129,6 +130,27 @@ def get_total_system_memory() -> float:
...
 
 
+def verify_cpu_supports_ml_inference() -> None:
+    """
+    Verify the CPU can run on-device ML (torch) inference.
+
+    :raises SetupFailedError: If this is an x86 CPU without AVX2 support, which
+        torch's FBGEMM quantized backend requires.
+    """
+    if platform.machine().lower() not in ("x86_64", "amd64", "i386", "i686", "x86"):
```
MarvinSchenkel, line 140: Claude had a nice suggestion:

"The allowlist not in ("AVX2", "AVX512") fails closed if a future torch version reports a new, higher capability string (torch has changed these before). Inverting to a denylist of the known-insufficient values is more future-proof:"

Can we test if this also works as expected?

`if torch.backends.cpu.get_cpu_capability() in ("DEFAULT", "NO AVX"):`

marcelveldt: @chrisuthe what about this suggestion ? We can also do it later as we now want to get this fix out

chrisuthe: Good call — done in 9689b21. Verified against torch's documented capability strings (DEFAULT, NO AVX, AVX2, AVX512, plus the non-x86 ones we already filter out earlier), and added a test case asserting an unknown future string like AVX10 now passes instead of raising.

music_assistant/helpers/util.py
```diff
@@ -129,6 +130,29 @@ def get_total_system_memory() -> float:
...
+        # non-x86 (ARM) machines run quantized inference via QNNPACK instead of FBGEMM
+        return
+    import torch  # noqa: PLC0415
+
+    # denylist of known-insufficient capabilities so a future torch version
+    # reporting a new, higher capability string fails open instead of closed
```
chrisuthe:
```suggestion
```

MarvinSchenkel (approved):

NL-DUX:
I was a bit surprised of this change this release, as I formerly WAS able to execute Smart Fades Analysis and all of a sudden MA tells me my CPU is too old? What happened to its previous capabilities? I think the choice of using a CPU instruction set may have the unintended side effect of needlessly shutting users out.

In my case it's an INTEL Atom C3538 (Synology machine, 20GB RAM)

Smart Fades Analysis was slow on my machine, sure, but I reckon that's a performance choice for me to make? The slowness didn't bother me when music was playing, and it bothered me even less when it was working whole days in the background when the music was switched off. Seeing as it's a one time deal where music is analyzed (mostly in the background for me) and its analysis result data is reused in the future, I'm having trouble to see the necessity of having a minimum amount of CPU set, per se. I was a thousand successful songs in when this change was implemented.

I am interested in its core functionality (who doesn't want great fades in music?) but it's also supposed to be a great addition when linking to Philips Hue lighting, and my home and garden are riddled with Hue lights. 

Could the hard limit perhaps soften up a little, or can the old analysis mode return as a (unsupported) legacy option?

PR #4169, opened 2026-06-11
Store track artist in playlog

OzGav:

I'm still not 100% happy with how the Last.fm recommendations are seeded, and while looking into seeding them from the playlog I found the playlog doesn't store the track artist.

For library items that doesn't matter (we can look the artist up), but for streamed tracks that were never added to the library the artist is lost even though it's right there on the media item when the play is logged. That makes a big chunk of listening (anything played straight from a streaming provider) unusable for recommendation seeding.

This PR records the primary artist alongside each playlog entry:
- new `artist` column on the `playlog` table and migration
- populated in `mark_item_played` from the played media item

No backfill of the DB is needed as the playlog is pruned at 90 days, so existing rows age out and the column fully populates within that window. New plays benefit immediately.

While here, this also adds a `recently_played_tracks(limit, played_after_timestamp, userid)` accessor on the music controller. `recently_played` returns `ItemMapping`, which has no artist field, so there was no way to read the new column back; this accessor returns recent fully-played tracks from the playlog *including* the stored artist, for callers that need play history with artist attribution.

This is groundwork only and the recommendation-seeding change that uses this will be a separate PR.

## Types of changes

## Checklist


MarvinSchenkel:
cc @marcelveldt 

I have been toying with the idea of enriching the playlog with more identifiable information so that we can do more things with play history. I think simply adding an artist is not the solution here, as we will probably need even more if we ever want to show detailed play history statistics in the future. Think 'breakdown of plays per genre' etc

music_assistant/controllers/music/controller.py
```diff
@@ -752,6 +761,45 @@ async def recently_played(
             )
         return result
 
+    async def recently_played_tracks(
```
marcelveldt, line 739: I don't like this hacky bandaid tbh

music_assistant/controllers/music/controller.py
```diff
@@ -1380,11 +1428,16 @@ async def mark_item_played(
...
+        # store the primary artist so streaming plays not in the library remain seedable
+        item_artists = getattr(media_item, "artists", None)
+        artist = item_artists[0].name if item_artists else None
+
         params = {
             "item_id": media_item.item_id,
             "provider": media_item.provider,
             "media_type": media_item.media_type.value,
             "name": media_item.name,
+            "artist": artist,
```
marcelveldt, line 1418: what problem is this going to solve ?

OzGav: The PR description explains it in detail. Basically I can t use the play log to provide seeds for last.fm without an artist

marcelveldt: and we cant just do a cheap library lookup ? so that lastfm lookups only act on library items ?

OzGav: This is for people who are listening to streaming tracks which are not in their library. If we limited to library only then we may not get enough seeds or they wont be representative of the users actual play history.

MarvinSchenkel: @marcelveldt see my comment above

marcelveldt: Going back to this, why don't you simply request the track then from the provider ?
We just need to make sure that we prefer a cached path over an API call. Expanding the playlog is not the way to go in any case.

marcelveldt:
> cc @marcelveldt
> 
> I have been toying with the idea of enriching the playlog with more identifiable information so that we can do more things with play history. I think simply adding an artist is not the solution here, as we will probably need even more if we ever want to show detailed play history statistics in the future. Think 'breakdown of plays per genre' etc

I think that is a bit irrelevant to this PR.
If you want playcounts per genre, we should extend the logic to also attribute the genre in the playlog, just like we do for artists and albums when a track is played.

OzGav:
Yeah OK I was looking to avoid API calls. However, I checked how the caching works for this and the recent-play window used in the new PR is 30 days and that's the same as the `get_track` cache TTL, plus every build re-warms it, so the seeds basically stay cached no matter how often it refreshes. So as it turns out no real API cost to worry about.

So the [new PR](https://github.com/music-assistant/server/pull/4457) ditches the need for the playlog change and just resolves each recent play back to its track through `get_provider_item`, so it uses the cached `get_track` and only falls back to the API if it has to. Library plays come straight from the DB anyway.

PR #4184, opened 2026-06-12
Fix persistent MusicBrainz rate limiting during Last.fm recommendation rebuilds

OzGav:

Users are still hitting rate limiting on the MusicBrainz mirror during Last.fm recommendation rebuilds, even after #4146 and #4155. Once the first 429  arrived, it became self-sustaining: every rate-limited request slept for 0s and then retried *without* going through the throttler, so they all burst at once while other callers kept firing during the backoff window.

So this PR ensures that:

1. **Retries are now throttled.** `throttle_with_retries` re-acquires a throttler slot for every attempt instead of only the first one. This was the actual bug.
2. **A 429 now pauses the whole provider.** The throttler gets a shared cooldown for the backoff window, so other callers stop firing into a window where every request just extends the block. Only set on 429 so a plain 502/503 still only backs off the affected request. Bypassed (time-critical playback) calls are unaffected.
3. **MusicBrainz pacing smoothed.** 10 req/10s allowed all 10 requests in one instant burst. Now it is 1 request per 1.25s which provides a bit of headroom.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/throttle_retry.py
+++ b/music_assistant/helpers/throttle_retry.py
@@ -118,14 +118,29 @@ class ThrottlerManager:
         self.retry_attempts = retry_attempts
         self.initial_backoff = initial_backoff
         self.throttler = Throttler(rate_limit, period)
+        self._cooldown_until: float = 0.0
+
+    def set_cooldown(self, seconds: float) -> None:
+        """
+        Pause all (non-bypassed) acquires, e.g. after a rate-limit response.
+
+        :param seconds: Duration of the pause; never shortens an active cooldown.
+        """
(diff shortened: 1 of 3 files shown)


marcelveldt:
This is over engineered completely.
We have the throttler for this which can handle throttle with retries just fine.

My main issue is more that the lastfm provider is making so many musicbrainz calls, we should stop that as soon as possible. If we cant find an easy fix, lets remove lastfm from the default providers because it doing a call to MB for every track is just unacceptable.

OzGav:
Superseded by #4190 which removes the Last.fm provider's MusicBrainz ISRC lookup instead of throttling them

PR #4186, opened 2026-06-12
Fix Last.fm Discover rows showing owned tracks under a different version name

OzGav:

Last.fm track names often carry a scrobbled version suffix (e.g. 'Ventura Highway - 2006 Remaster'), which the library pre-filter could never match against the owned track: the version suffix defeated the name comparison, and the space-separated search query format meant the library search returned no candidates at all. Strip the version suffix and search as 'Artist - Title' so owned tracks are recognised. A name match is overruled when both sides carry recording MBIDs that differ, as those are genuinely different tracks.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/lastfm_recommendations/recommendations.py
+++ b/music_assistant/providers/lastfm_recommendations/recommendations.py
@@ -19,6 +19,7 @@ from music_assistant_models.media_items import (
 
 from music_assistant.constants import CONF_USERNAME
 from music_assistant.helpers.compare import compare_strings
+from music_assistant.helpers.util import parse_title_and_version
 from music_assistant.providers.lastfm_recommendations.constants import (
     CACHE_CATEGORY_RESOLVED_ITEMS,
     CACHE_EXPIRATION_SECONDS,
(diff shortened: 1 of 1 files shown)


music_assistant/providers/lastfm_recommendations/recommendations.py
```diff
@@ -188,21 +189,27 @@ async def _is_in_library(self, item_data: dict[str, Any], media_type: MediaType)
                 artist_info if isinstance(artist_info, str) else artist_info.get("name", "")
             )
             if name and artist_name:
-                search_query = f"{artist_name} {name}"
+                # Last.fm track names carry scrobbled version suffixes ("- 2006 Remaster");
+                # strip those so variants of an owned track still match
+                clean_name, _ = parse_title_and_version(name, strip_for_search=True)
```
marcelveldt, line 194: why are we throwing away the version info ?
shouldn't we be matching for same versions as well ?

OzGav: I dont think so. Spotify, for example, just slaps a (Remaster) label on everything. My intent is people are shown tracks they dont currently have and simple remasters are, for all intents and purposes, the same track. This principal is what Musicbrainz subscribes to

OzGav: Even if there is a false match this just means a different version of the track someone has in their library is not shown and this still meets the intent of showing the users music that they don't have in their library (in any version)

marcelveldt: hmmm OK
I'm actually in the camp where I think a remaster is a different version but OK.

marcelveldt (approved):

PR #4196, opened 2026-06-13
Spotify Connect: don't block volume changes when no music provider is installed

MarvinSchenkel:
> 🤖 This issue was auto-triaged and this PR was opened in draft for a maintainer to review.

When Spotify Connect is used **without** a matching Spotify music provider installed, changing the volume from the Spotify app no longer worked on 2.9.0 (it worked on 2.8.0).

A volume change on the player fires the `on_volume_change` hook on the Spotify Connect AudioSource so the Spotify app's volume slider can be kept in sync. That sync goes through the Spotify Web API, which is only available when a matching Spotify music provider is configured. Without one, `_on_volume` raised `UnsupportedFeaturedException`, and because the hook is awaited from `PlayerController.cmd_volume_set`, the exception propagated and aborted the volume change.

There is nothing to sync back to when the Web API is unavailable, so `on_volume_change` now returns early in that case instead of raising.

- https://github.com/music-assistant/support/issues/5627

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/spotify_connect/__init__.py
+++ b/music_assistant/providers/spotify_connect/__init__.py
@@ -431,6 +431,11 @@ class SpotifyConnectProvider(PluginProvider):
         """Sync the Spotify app's volume slider with the player's new volume."""
         if source_id != AUDIO_SOURCE_ID:
             return
+        # Syncing the slider needs the Web API, which is only available with a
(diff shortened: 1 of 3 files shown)


A-KumarSharma:
I’ve love to work on this as contributor

marcelveldt:
this one is no longer needed with the go-librespot connect provider

PR #4199, opened 2026-06-13
Add correct playlog information when retrieving audiobooks in audiobooks controller

fmunkes:

Within the frontend's audiobooks view the in-progress icons are correctly referring to the logged in user as the library_items method is used. However, when using any other method to receive an audiobook (e.g. music/audiobook/get) the playlog information might be incorrect. (This is e.g. present in the abs' provider's browse implementation).

I made the base_query a property, so we can dynamically add the current user for playlog retrieval.

## Types of changes

## Checklist


music_assistant/controllers/media/audiobooks.py
```diff
@@ -60,10 +63,24 @@ def __init__(self, mass: MusicAssistant) -> None:
             FROM audiobooks
             LEFT JOIN playlog ON playlog.item_id = audiobooks.item_id AND playlog.media_type = 'audiobook'
```
marcelveldt, line 74: so maybe I have missed this but I'm wondering why we're joining the playlog here in the first place.
I'm not sure if inflating the base query this way is a good practice. I think we should only join/use the playlog on places where its actually needed. So we also have to avoid workarounds like proposed in this PR

fmunkes: The joining in the base query has been there before me ;-)
I added the user_id to the extra_join_parts of the library_items some time ago. If I understand you correctly, you'd prefer to have all the joining in the respective methods - this works for library_items, but some of the `get_library_items...` methods, e.g. `    get_library_items_by_prov_id` are decorated with `final` in the base controller class. Would it then be fine to lift this restriction, so we can overwrite this method in the audiobooks controller?

marcelveldt: This could as well be me doing it this way and now I'm questioning myself haha.
I was just thinking why we're joining playlog here for audiobooks controller but not on any other controller/mediaitem. I'll dig through the code to see if I can discover my reasoning :-)

marcelveldt:
I tweaked the PR a tiny bit to make it more universal.

marcelveldt (approved):
Thanks, @fmunkes !

marcelveldt:
If you want this also backported to stable, please assign the "backport-to-stable" label

PR #4210, opened 2026-06-14
Fix translations in Audiobookshelf and iTunes podcasts

fmunkes:

Not all translation strings made it from the frontend to the backend after https://github.com/music-assistant/server/pull/4200. Others did not match in key. Fixed here for ABS and iTunes podcasts.

What do we want to do with player options and sound modes? Their translation keys are critical in terms of HA, and still in the frontend.

## Types of changes

## Checklist


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -1034,10 +1035,11 @@ async def recommendations(self) -> list[RecommendationFolder]:
         # from _browse_lib_audiobooks, i.e. Authors, Narrators etc.
         # Podcast libs do not have filter folders, so always the root folders.
         browse_items: list[MediaItemType | BrowseFolder] = []
-        translation_key = "libraries"
+        _prefix = f"provider.{DOMAIN}.recommendations"
+        translation_key = f"{_prefix}.libraries"
         if len(self.libraries.audiobooks) <= 1:
             if len(self.libraries.podcasts) == 0:
-                translation_key = "library"
+                translation_key = f"{_prefix}.library"
```
marcelveldt: you dont need to set the prefix yourself - this is auto derived as long as you specify the key in your strings.json

fmunkes: Hmm, if I remove the prefix e.g. for the browse folder (so just `browse.<my key>`) it doesn't work anymore.

marcelveldt: the translation_key should only be the key, no prefix at all

fmunkes: I adjusted it in the way I think you mean, but for me this does not work.
I did clear the cache to be certain. Also, the generated strings.json does show the keys as it should.
2026-06-14-170945-screenshot-region

marcelveldt: OK, I'll have a look - might be a bug or some oversight

marcelveldt:
Pushed a rework on top: the translation keying changed in #4216 - media item names now resolve under `media.<type>.*`, so i merged dev in and re-nested your new abs strings under `media.folder.*` / `media.recommendations.*` (at the provider root they didn't actually resolve). also reverted the two fully-qualified keys back to bare - `in_progress_items` / `episodes_recently_added` live in the shared catalog now - and dropped the trending_podcasts + library/libraries bits since those are handled centrally too. should be green shortly.

marcelveldt (approved):
Thanks, @fmunkes !

PR #4211, opened 2026-06-14
Adapt QQ Music provider to qqmusic-api 0.6

xiasi0:
## Summary

This updates the QQ Music provider for `qqmusic-api-python` 0.6.x and the current Music Assistant provider APIs.

Changes include:

- Bump `qqmusic-api-python` to 0.6.6 and align the `cryptography` requirement with the new dependency constraints.
- Replace the old module-level QQ Music API calls with the 0.6 `QQClient` client/module interfaces.
- Adapt search, artist, album, playlist, recommendation, stream URL, playlist edit, and QR login handling to the 0.6 response models and method signatures.
- Preserve the existing QR login user flow while storing the full 0.6 credential payload, including `encryptUin`, so authenticated config options remain available after a successful scan.
- Decode encrypted 0.6 lyric payloads before exposing lyrics to Music Assistant.
- Set `metadata.description_language` for QQ Music descriptions so MA can handle localized metadata explicitly.
- Keep the existing QQ Music provider tests lean while updating them for the 0.6 API shape.

## Music Assistant compatibility

This also adapts the provider to newer MA behavior by:

- Adding `description_language` alongside metadata descriptions.
- Keeping authenticated-only config entries hidden until QR login succeeds.
- Avoiding additional discovery sections; the provider continues to expose its existing recommendation folders only.


music_assistant/providers/qqmusic/__init__.py
```diff
@@ -747,6 +778,33 @@ async def _resolve_stream_url(
...
+    def _extract_stream_url(self, url_response: Any, item_id: str) -> str:
+        """Extract absolute stream URL from qqmusic-api 0.6 or legacy URL payload."""
+        if isinstance(url_response, dict) and isinstance(url_response.get(item_id), str):
+            return str(url_response[item_id])
+        response = self._to_dict(url_response)
+        url_items = response.get("midurlinfo") or response.get("data") or []
+        if not isinstance(url_items, list):
+            return ""
+        cdn_base = str(
+            getattr(
```
OzGav: Accessing `self._qq_song._SONG_URL_FALLBACK_DOMAIN` via `getattr` ties us to a private library constant so if it's renamed or removed in a future update, this silently falls back to the hardcoded URL with no warning. Can we get the base domain from a public API, or at least log when the fallback is used so a future break is visible?

music_assistant/providers/qqmusic/__init__.py
```diff
@@ -279,112 +257,88 @@ async def _start_qr_auth(
...
-            "Waiting for scan confirmation timed out. "
-            "You can click Start QR login again or use Check QR status."
-        )
+        deadline = time.monotonic() + 120
+        while time.monotonic() < deadline:
+            try:
+                await _check_qr_auth(values, client=client, qr=qr)
+                return
+            except InvalidDataError as err:
+                if "expired" in str(err).lower() or "rejected" in str(err).lower():
```
OzGav: Branching on "expired"/"rejected" substrings in the error message is fragile as a wording change anywhere upstream silently alters the retry behaviour here. Could we distinguish these states by exception type (or a status field on the result) rather than by message text?

music_assistant/providers/qqmusic/__init__.py
```diff
@@ -678,28 +658,57 @@ async def _get_recommend_payload_cached(
...
+        """Normalize and decrypt QQ Music lyric responses."""
+        if callable(decrypt := getattr(response, "decrypt", None)):
+            response = decrypt()
+        lyric_obj = dict(self._to_dict(response))
+        if str(lyric_obj.get("crypt") or "0") != "1":
+            return lyric_obj
+        for key in ("lyric", "trans", "roma"):
+            value = str(lyric_obj.get(key) or "").strip()
+            if value and _HEX_LYRIC_PATTERN.fullmatch(value):
+                with suppress(Exception):
```
OzGav: This will swallow any failure and leave the raw encrypted hex in the lyrics field with no signal. Suggest narrowing to the specific exception `qrc_decrypt` raises and logging at debug so a decrypt regression is diagnosable.

music_assistant/providers/qqmusic/__init__.py
```diff
@@ -210,20 +189,20 @@ async def setup(
...
-        values.get(CONF_MUSICID) and values.get(CONF_MUSICKEY)
-    )
-
-
-def _has_qr_pending(values: dict[str, ConfigValueType]) -> bool:
-    return bool(values.get(CONF_QR_IDENTIFIER))
+    credential_json = str(values.get(CONF_CREDENTIAL_JSON) or "").strip()
+    if not credential_json:
+        return False
+    with suppress(Exception):
```
OzGav: Can this narrow to ValidationError/ValueError? A broad suppress here could mask unexpected failures as "not verified" and send the user back through QR login for the wrong reason.

OzGav:
A conflict has popped up as well

xiasi0:
Updated the branch with the latest `dev` and addressed the QQ Music review comments:

- QR polling now checks `QRCodeLoginEvents` directly instead of matching error message text.
- Credential verification now only suppresses validation/value errors.
- QRC lyric decrypt failures now catch the expected exceptions and log at debug level.
- Stream URL fallback domain usage now logs when the library fallback constant is unavailable.

I also kept the existing QR login user flow unchanged and verified the QQ Music provider checks locally.

OzGav (approved):
LGTM!

OzGav:
If any changes to the docs are required then please PR those targetting the beta branch.

PR #4231, opened 2026-06-15
fix for #5652 - Fix YouTube Music radio / similar tracks failing with KeyError 'endpoint'

hippich:

**Bug:** "Start Radio" (and "similar tracks" / dynamic radio) on YouTube Music tracks would fail with `Error handling message: player_queues/play_media: 'endpoint'`.

**Root cause:**
`YoutubeMusicProvider.get_similar_tracks()` calls `get_song_radio_tracks()`, which uses `ytmusicapi.get_watch_playlist(..., radio=True)`. The library's response parser (`get_tab_browse_id`) can raise a raw `KeyError('endpoint')` (or similar) when the watch/radio response contains a tab without the expected `["tabRenderer"]["endpoint"]` structure. This exception was not caught anywhere in the provider or the radio flow (`get_dynamic_radio_tracks` → `tracks.similar_tracks` → `play_media(radio_mode=True)`), which only catches `MusicAssistantError` in the main media handling path. The unhandled error aborted the entire play command.

**Fix:**
Wrap the `get_song_radio_tracks()` call in `get_similar_tracks()` with a try/except. On failure, log a warning and return `[]`. This lets radio mode gracefully fall back to the seed/base track(s) instead of crashing, matching the existing defensive style used elsewhere in the provider for YTM response fragility.

- related issue: https://github.com/music-assistant/support/issues/5652

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/ytmusic/__init__.py
+++ b/music_assistant/providers/ytmusic/__init__.py
@@ -625,9 +625,13 @@ class YoutubeMusicProvider(MusicProvider):
     @use_cache(3600 * 24, allow_expired_cache=True)  # Cache for 1 day
     async def get_similar_tracks(self, prov_track_id: str, limit: int = 25) -> list[Track]:
         """Retrieve a dynamic list of tracks based on the provided item."""
-        result = await get_song_radio_tracks(
-            headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
-        )
+        try:
+            result = await get_song_radio_tracks(
+                headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
+            )
+        except Exception as err:
+            self.logger.warning("Failed to get similar/radio tracks from YT Music: %s", err)
+            return []
         if "tracks" in result:
             tracks = []
             for track in result["tracks"]:
--- a/music_assistant/providers/ytmusic/__init__.py
+++ b/music_assistant/providers/ytmusic/__init__.py
@@ -50,7 +50,7 @@ from music_assistant_models.media_items import (
 )
 from music_assistant_models.streamdetails import StreamDetails
 from ytmusicapi.constants import SUPPORTED_LANGUAGES
-from ytmusicapi.exceptions import YTMusicServerError
+from ytmusicapi.exceptions import YTMusicError, YTMusicServerError
 from ytmusicapi.helpers import get_authorization, sapisid_from_cookie
 
 from music_assistant.constants import (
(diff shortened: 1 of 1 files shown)


music_assistant/providers/ytmusic/__init__.py
```diff
@@ -625,9 +625,13 @@ async def remove_playlist_tracks(
...
-        result = await get_song_radio_tracks(
-            headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
-        )
+        try:
+            result = await get_song_radio_tracks(
+                headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
+            )
+        except Exception as err:
+            self.logger.warning("Failed to get similar/radio tracks from YT Music: %s", err)
+            return []
```
marcelveldt, line 635: So if parsing one song fails we fail everything ?

music_assistant/providers/ytmusic/__init__.py
```diff
@@ -625,9 +625,13 @@ async def remove_playlist_tracks(
...
     async def get_similar_tracks(self, prov_track_id: str, limit: int = 25) -> list[Track]:
         """Retrieve a dynamic list of tracks based on the provided item."""
-        result = await get_song_radio_tracks(
-            headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
-        )
+        try:
+            result = await get_song_radio_tracks(
+                headers=self._headers, prov_item_id=prov_track_id, limit=limit, user=self._yt_user
+            )
+        except Exception as err:
```
marcelveldt: Never catch a bare exception but try narrow it down, for example KeyError in this case

OzGav:
@marcelveldt Im not sure this is the correct fix. It might just need a library bump but @MarvinSchenkel will need to comment when he is back.

hippich:
with last commit:

1) narrowed down exceptions types.
2) adjusted code parsing track results, so if we can pull partial results from the response - try to do that instead of failing the whole thing.
3) copilot's comment about returning [] in the cached method - moved fallback [] return to public non-cached function and original logic + patches pushehed into private function.

no idea about upgrade of the library as a proper fix, but intuitively i feel like even with upgrade of the underlying library we still need to be defensive about handling its results.

marcelveldt:
The library should have been bumped a few days ago.
Can you have a look at the outstanding comments of copilot and give a reply ?
when you are ready for a (final?) review, press the "ready for review" button. Thanks!

MarvinSchenkel:
Do we still need this? According to [this](https://github.com/sigma67/ytmusicapi/pull/936) PR, this exact error has been fixed upstream in the library.

hippich:
I just upgraded MA and looks like it is working without my patch. So I would say this is no longer needed. Closing.

PR #4243, opened 2026-06-15
Fix image download from CDNs that reject our User-Agent

OzGav:

A user reported that artwork for the Swedish radio station P3 (added via Radio Browser) never downloads. The image URL loads fine in a browser.

The cause is the User-Agent. Music Assistant fetches images server-side and identifies itself with a `Music Assistant/… aiohttp/… Python/…` User-Agent. The station's artwork is served by Akamai, whose bot protection returns 403 for any User-Agent it doesn't recognise as a known-good client and our identifying string falls into that blocked bucket.

So now when an image fetch comes back 401/403, retry it once with a User-Agent that these CDNs allow. We keep our normal, self-identifying User-Agent for the first attempt (and for every host that already works), so the only behaviour change is for hosts that were rejecting us outright and those now succeed on the retry. Verified end-to-end against the reporting station: first attempt 403, retry returns the real image.

- related issue https://github.com/music-assistant/support/issues/5675

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/images.py
+++ b/music_assistant/helpers/images.py
@@ -17,9 +17,10 @@ from pathlib import Path
 from typing import TYPE_CHECKING, cast
 
 import aiofiles
-from aiohttp.client_exceptions import ClientError
+from aiohttp.client_exceptions import ClientError, ClientResponseError
 from PIL import Image, UnidentifiedImageError
 
+from music_assistant.constants import APPLICATION_NAME
 from music_assistant.helpers.security import is_safe_path
 from music_assistant.helpers.tags import get_embedded_image
 from music_assistant.models.metadata_provider import MetadataProvider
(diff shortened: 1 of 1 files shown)


music_assistant/helpers/images.py
```diff
@@ -281,6 +281,30 @@ async def get_image_data(
...
+    try:
+        async with mass.http_session_no_ssl.get(url, raise_for_status=True) as resp:
+            return await resp.read()
+    except ClientResponseError as err:
+        # Some CDNs (e.g. Akamai) 403 our default User-Agent; retry once with
+        # a UA they allowlist before giving up.
+        if err.status not in (401, 403):
+            raise
+        fallback_ua = (
+            f"{APPLICATION_NAME}/{mass.version} (Wget/1.24.5; +https://music-assistant.io)"
```
marcelveldt: why not use this as default agent then ?

OzGav: I didnt want to risk breaking some other call site.

marcelveldt: OK, I adjusted it a tiny bit to not do the ugly retry but change the user agent only for the images path

marcelveldt (approved):

PR #4246, opened 2026-06-16
docs: add translated README links

dowithless:
This PR adds links to translated versions of the README to help non-English-speaking users access translated documentation.

Supported languages:

- German
- Spanish
- French
- Japanese
- Korean
- Portuguese
- Russian
- Chinese

Translations are automatically synchronized and require no maintenance effort from project maintainers.

This change only affects README documentation and does not modify project code, dependencies, CI workflows, or repository settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,16 @@
 Music Assistant
 ==================================
 
+<!-- Keep these links, translations synced daily. -->
+[Deutsch](https://zdoc.app/de/music-assistant/server) |
+[Español](https://zdoc.app/es/music-assistant/server) |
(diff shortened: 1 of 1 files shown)


marcelveldt:
In general, the readme on the server project itself is only read by developers and not users so I'm not really understanding why we need this. An English README is just fine for this purpose. And in fact, our readme is very limited, because we have a dedicated documentation website (music-assistant.io).

If I were to add support for translations anywhere, that would be the docs site, not the readme of the underlying project readme.

Sorry for the trouble of creating the PR but imo there is no need to merge this.

PR #4260, opened 2026-06-16
Record explicit album/artist/track plays as user-initiated

chrisuthe:

Restores the Discover **Recently played** row, which stopped updating after upgrading to 2.9.x.

**Why it broke:** the row read the playlog with `user_initiated_only=True` (`user_initiated = 1`), but the most common in-app plays never set that flag — albums were credited `user_initiated=False`, artists were never marked on enqueue, and individual tracks were hard-coded `False` in the playback-report path. Only playlist/genre/podcast enqueues set it, so for anyone who mostly plays albums nothing new matched the filter. The 2.9.x migration added the column with `DEFAULT 1`, backfilling existing rows, which is why the row stayed **frozen on pre-upgrade items** rather than going empty. (The Albums "last played" sort kept working because it reads the library table's `last_played`, a separate path.)

**The fix —** the Discover **Recently played** row now reads the playlog with `user_initiated_only=False`, so it surfaces everything that actually played (newest-first), behaving like a history of what you heard rather than only what you explicitly clicked. This is the behavior discussed in review ([thread](https://github.com/music-assistant/server/pull/4260#issuecomment)).

**Also: corrected `user_initiated` tagging.** Independent of the row's filter, the flag was being recorded inconsistently. This PR makes it record a play as user-initiated whenever the user explicitly chose that exact item:

- **Albums** — `_mark_album_played` now passes `user_initiated=True` (already gated on the album being explicitly enqueued).
- **Artists** — the ARTIST enqueue branch now marks the artist, mirroring the existing playlist/genre/podcast branches.
- **Single tracks** — the playback-report path sets the flag from a new `_is_user_initiated_play()` helper (`media_item in queue.enqueued_media_items`), so a directly-played track counts but auto-advanced / radio tracks do not.

**Artist credit decoupling (the non-obvious part):** an album row is written in one place, so flipping its flag is safe; an artist row is rewritten on *every* track via `_credit_artist_plays`. That method is now always a side-effect write (`user_initiated=False`, parameter removed) using a non-downgrading upsert — `user_initiated = playlog.user_initiated OR excluded.user_initiated` — so an explicit artist play stays user-initiated across the repeated credits its tracks generate, and a track/album play never flags its artists. Order-independent, which matters since these run as `create_task`s.

**a playlog row is `user_initiated=True` iff that exact item was explicitly enqueued.** 

The default Recently played row no longer filters on it (it shows everything that has played in reverse chronological order, as expected for 'a row of stuff I have heard').

- related issue music-assistant/support#5671

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -2049,7 +2049,7 @@ class MusicController(CoreController):
                 icon="mdi-motion-play",
                 items=cast(
                     "UniqueList[MediaItemType | ItemMapping | BrowseFolder]",
-                    await self.recently_played(limit=10, user_initiated_only=True),
+                    await self.recently_played(limit=10, user_initiated_only=False),
                 ),
             ),
             RecommendationFolder(
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -3599,12 +3599,7 @@ class MusicController(CoreController):
         queue_id: str | None,
         skip_ids: set[str],
     ) -> None:
-        """Credit each (library-resolvable) artist with a play, skipping skip_ids.
-
-        An artist credited this way is a side-effect of a track or album play, so its
-        playlog row is recorded as not user-initiated. An existing user-initiated flag
-        (set when the artist itself was explicitly played) is preserved, never downgraded.
-        """
+        """Credit each (library-resolvable) artist with a play, skipping skip_ids."""
         # ON CONFLICT keeps an explicit user-initiated artist play sticky across the
         # repeated side-effect credits its tracks generate.
         upsert_query = (
@@ -3648,4 +3643,3 @@ class MusicController(CoreController):
             for user_id in user_ids:
                 playlog_entry["userid"] = user_id
                 await self.database.execute(upsert_query, playlog_entry)
-        await self.database.commit()
--- a/music_assistant/controllers/music.py
+++ b/music_assistant/controllers/music.py
@@ -1473,7 +1473,6 @@ class MusicController(CoreController):
                 timestamp=timestamp,
                 user_ids=user_ids,
                 queue_id=queue_id,
-                user_initiated=user_initiated,
                 skip_ids=set(skip_artist_ids or ()),
             )
         await self.database.commit()
@@ -3598,10 +3597,28 @@ class MusicController(CoreController):
         timestamp: float,
         user_ids: list[str],
         queue_id: str | None,
-        user_initiated: bool,
         skip_ids: set[str],
     ) -> None:
-        """Credit each (library-resolvable) artist with a play, skipping skip_ids."""
+        """Credit each (library-resolvable) artist with a play, skipping skip_ids.
+
+        An artist credited this way is a side-effect of a track or album play, so its
+        playlog row is recorded as not user-initiated. An existing user-initiated flag
+        (set when the artist itself was explicitly played) is preserved, never downgraded.
+        """
+        # ON CONFLICT keeps an explicit user-initiated artist play sticky across the
+        # repeated side-effect credits its tracks generate.
+        upsert_query = (
+            f"INSERT INTO {DB_TABLE_PLAYLOG} "
+            "(item_id, provider, media_type, name, image, fully_played, "
+            "seconds_played, timestamp, queue_id, user_initiated, userid) "
+            "VALUES (:item_id, :provider, :media_type, :name, :image, :fully_played, "
+            ":seconds_played, :timestamp, :queue_id, :user_initiated, :userid) "
+            "ON CONFLICT(item_id, provider, media_type, userid) DO UPDATE SET "
+            "name = excluded.name, image = excluded.image, "
+            "fully_played = excluded.fully_played, seconds_played = excluded.seconds_played, "
+            "timestamp = excluded.timestamp, queue_id = excluded.queue_id, "
+            f"user_initiated = {DB_TABLE_PLAYLOG}.user_initiated OR excluded.user_initiated"
+        )
         for artist in artists:
             db_artist = await self.artists.get_library_item_by_prov_id(
                 artist.item_id, artist.provider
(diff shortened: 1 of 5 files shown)


music_assistant/controllers/player_queues.py
```diff
@@ -3370,6 +3375,16 @@ def _enqueued_album_for_track(
                 return None
         return enqueued
 
+    def _is_user_initiated_play(self, queue: PlayerQueue, media_item: MediaItemType) -> bool:
+        """Return whether a played item was explicitly chosen by the user.
```
```suggestion
        """
        Return whether a played item was explicitly chosen by the user.
```

music_assistant/controllers/music.py
```diff
@@ -3598,10 +3597,28 @@ async def _credit_artist_plays(
         timestamp: float,
         user_ids: list[str],
         queue_id: str | None,
-        user_initiated: bool,
         skip_ids: set[str],
     ) -> None:
-        """Credit each (library-resolvable) artist with a play, skipping skip_ids."""
+        """Credit each (library-resolvable) artist with a play, skipping skip_ids.
```
```suggestion
        """
        Credit each (library-resolvable) artist with a play, skipping skip_ids.
```

tests/core/test_artist_play_credit.py
```diff
@@ -120,6 +130,46 @@ async def test_album_play_credits_album_artists_with_dedup(mass: MusicAssistant)
     assert await _play_count(mass, DB_TABLE_ARTISTS, skipped.item_id) == 0
 
 
+async def test_track_credit_does_not_flag_artist_user_initiated(mass: MusicAssistant) -> None:
+    """A track play credits its artist as a side-effect, never as user-initiated.
```
marcelveldt:
```suggestion
    """
    A track play credits its artist as a side-effect, never as user-initiated.
```

tests/core/test_artist_play_credit.py
```diff
@@ -120,6 +130,46 @@ async def test_album_play_credits_album_artists_with_dedup(mass: MusicAssistant)
...
+
+    await mass.music.mark_item_played(
+        track, fully_played=True, user_initiated=True, userid=user.user_id
+    )
+
+    assert await _artist_user_initiated(mass, artist.item_id, user.user_id) == 0
+
+
+async def test_explicit_artist_play_survives_track_credit(mass: MusicAssistant) -> None:
+    """An explicit artist play stays user-initiated after later side-effect credits.
```
marcelveldt:
```suggestion
    """
    An explicit artist play stays user-initiated after later side-effect credits.
```

marcelveldt:
We should maybe consider adding 2 different homescreen rows:

1) Last enqueued items (current behavior)
2) Last played items (everything that was actually played)

Because if you have some playlist playing and you think "hey what was that song I just heard" you wont be able to find it on the recently played with the current behavior.

So if we change the current "recently played" to lift the whole "user initiated" we can get that behavior and then add another optional widget row to get only the specifically enqueued items. Does that make any sense or not ?

music_assistant/controllers/player_queues.py
chrisuthe: Fixed in ae83f8b8d

music_assistant/controllers/music.py
chrisuthe: fixed within ae83f8b8d

tests/core/test_artist_play_credit.py
chrisuthe: fixed by ae83f8b8d

tests/core/test_artist_play_credit.py
chrisuthe: fixed by  ae83f8b8d

chrisuthe:
> We should maybe consider adding 2 different homescreen rows:
> 
> 1. Last enqueued items (current behavior)
> 2. Last played items (everything that was actually played)
> 
> Because if you have some playlist playing and you think "hey what was that song I just heard" you wont be able to find it on the recently played with the current behavior.
> 
> So if we change the current "recently played" to lift the whole "user initiated" we can get that behavior and then add another optional widget row to get only the specifically enqueued items. Does that make any sense or not ?

Yea this is how I would want my "Recently played" row to function, like a history of what I've heard, not what I've clicked on. Do you think it's worth a whole different row or is there a place a setting makes sense? I think it makes sense to have this land, fixing previous behavior, then add the setting or the additional row in a subsequent PR, do you agree?

music_assistant/controllers/music.py
```diff
@@ -3626,8 +3643,9 @@ async def _credit_artist_plays(
...
                 "timestamp": timestamp,
                 "queue_id": queue_id,
-                "user_initiated": user_initiated,
+                "user_initiated": False,
             }
             for user_id in user_ids:
                 playlog_entry["userid"] = user_id
-                await self.database.insert(DB_TABLE_PLAYLOG, playlog_entry, allow_replace=True)
+                await self.database.execute(upsert_query, playlog_entry)
+        await self.database.commit()
```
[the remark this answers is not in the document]
chrisuthe: The AI Gets credit for a good catch here- this is indeed redundant with line 1478.

marcelveldt:
Yeah a setting in the music controller (because that is where this recommendation is coming from, right?) makes sense OR a separate widget row. In any case it would make sense to fix the filter now so the row behaves more as we described above, showing what actually played.

Long-term refactor/cleanup should be that we introduce a dedicated library recommendations provider that hosts these builtin recommendation rows because I'm starting to feel that we shouldn't have done this directly in the controller itself (and make the controller just the orchestrator, not providing anything itself) but that is a follow-up for later when we get bored.

chrisuthe:
> Yeah a setting in the music controller (because that is where this recommendation is coming from, right?) makes sense OR a separate widget row. In any case it would make sense to fix the filter now so the row behaves more as we described above, showing what actually played.
> 
> Long-term refactor/cleanup should be that we introduce a dedicated library recommendations provider that hosts these builtin recommendation rows because I'm starting to feel that we shouldn't have done this directly in the controller itself (and make the controller just the orchestrator, not providing anything itself) but that is a follow-up for later when we get bored.

Flipped the default Recently played row to user_initiated_only=False so it now shows everything that actually played, as discussed in 9b7fa19. Kept the corrected user_initiated tagging in place — that's the groundwork for the follow-up. I'll do the optional "recently enqueued" row / music-controller setting as a separate PR.

marcelveldt (approved):
Thanks, @chrisuthe !

PR #4197, opened 2026-06-13
support cgroups on linux for memory detection

nerzhul:
# What does this implement/fix?

Currently we look for sysconf, which is not exact available memory when using containers. We must read cgroup values under Linux.

This will permit to better tune music-assistant when memory constraints are sets on containers

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [x] Enhancement to an existing feature — `enhancement`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [x] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [ x `pre-commit run --all-files` passes.
- [ ] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.
- [ ] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.
- [ ] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate.

PR #4269, opened 2026-06-16
Respect container cgroup memory limits when detecting system memory

marcelveldt:

`get_total_system_memory()` only read the host's physical RAM, so inside a memory-constrained container (e.g. `docker run --memory=4g` on a 32GB host) it reported the full 32GB. Music Assistant then offered buffer presets and gated heavy audio-analysis providers as if all that RAM were available, risking OOM-kills.

On Linux it now returns `min(physical RAM, cgroup memory limit)` so the container's real allocation is respected. This is a reworked, tested take on the idea from #4197.

- Read cgroup **v2** (`memory.max`) first, with a **v1** (`memory/memory.limit_in_bytes`) fallback.
- Use the process's own cgroup (from `/proc/self/cgroup`) with a mount-root fallback for the common namespaced-container case.
- Treat `max` / the v1 near-`INT64_MAX` sentinel as "unlimited", and fail open to host RAM on any read error or non-Linux platform.
- Compared to #4197, this drops the fragile `/proc/self/mountinfo` parsing in favour of the conventional `/sys/fs/cgroup` mount points, splits the logic into small testable helpers, and adds fixture-based unit tests (v2, v1, leaf-vs-root preference, sentinels, and the `min()` selection).

Credit to @nerzhul (#4197) for raising the cgroup gap.

> Note: this does not change behaviour for Home Assistant OS add-ons, which generally run without a per-container memory limit and so continue to see the shared host RAM.

- Supersedes #4197

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -119,15 +119,137 @@ async def warn_if_missing_x86_64_v2(logger: logging.Logger) -> None:
 
 
 def get_total_system_memory() -> float:
-    """Get total system memory in GB."""
-    try:
-        # Works on Linux and macOS
-        total_memory_bytes = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
-        return total_memory_bytes / (1024**3)  # Convert to GB
-    except AttributeError, ValueError:
-        # Fallback if sysconf is not available (e.g., Windows)
-        # Return a conservative default to disable buffering by default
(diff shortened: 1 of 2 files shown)


nerzhul:
Hello,Thanks for taking time to polish my PR and make it mergeable.Can you fix your git commit to put me as Co-Author, please, as îm the original author.That will permit proper rights and show me in the git logThanks

PR #4272, opened 2026-06-17
Validate return_url before appending JWT token

s0yd4RK:
Closes the token leakage described in GHSA-j369-4c4w-7qmq.

The /auth/login endpoint was appending the JWT token to a user-supplied
return_url without any validation. This mirrors the existing
is_allowed_redirect_url() check from _handle_auth_authorize to the
login endpoint.


marcelveldt:
Thanks for the PR/fix!
Small lint error pops up in CI, can you have a look?

music_assistant/controllers/webserver/controller.py
```diff
@@ -743,9 +743,13 @@ async def _handle_auth_login(self, request: web.Request) -> web.Response:
...
                 "user": auth_result.user.to_dict(),
             }
-
             # If return_url provided, append code parameter and return as redirect_to
             if return_url:
+                # SECURITY FIX (GHSA-j369-4c4w-7qmq): Validate return_url before
+                # appending the JWT token, mirroring the check in _handle_auth_authorize.
+                is_valid, _ = is_allowed_redirect_url(return_url, request, self.base_url)
+                if not is_valid:
+                    return web.Response(status=400, text="Invalid return_url")
```
marcelveldt: Thanks for jumping on this 🙏

One thing though — I don't think this fully closes the leak yet. `is_allowed_redirect_url()` is a three-way classifier: it returns `(True, "external")` for any valid-but-unknown external URL (anything that isn't same-origin / localhost / private IP / base_url / a registered pattern). Since we only check `is_valid` here, a `return_url` like `https://evil.com` still passes and we still append the token to it — which is exactly the vector the advisory describes.

`_handle_auth_authorize` gets away with the `is_valid`-only check because it never appends the token (that happens later in `_handle_auth_callback`, which gates `external` behind a consent screen). This endpoint appends the token immediately and has no consent step, so it should only auto-redirect to `trusted` destinations:

```suggestion
                # SECURITY FIX (GHSA-j369-4c4w-7qmq): only forward the token to trusted
                # destinations. is_allowed_redirect_url returns (True, "external") for any
                # unknown external URL, so checking is_valid alone still leaks the JWT.
                _, category = is_allowed_redirect_url(return_url, request, self.base_url)
                if category != "trusted":
                    return web.Response(status=400, text="Invalid return_url")
```

All the legit callers (web UI = same-origin, app = `musicassistant://`, HA = `homeassistant.local` / `my.home-assistant.io`, LAN = private IP) already resolve to `trusted`, so this shouldn't break any real flow.

Could you also add a quick regression test asserting an external/blocked `return_url` is rejected and the token isn't appended? Would've caught this, and it protects the backport.

marcelveldt (approved):
Thanks, @s0yd4RK !

PR #4303, opened 2026-06-17
Fix live metadata not refreshing for grouped AirPlay players

marcelveldt:

Live streams with their own "now playing" metadata (e.g. BBC Sounds) got stuck on the first track when played to a Sync Group whose leader is a Universal Player outputting via AirPlay (shairport-sync). The title updated once at startup and then never refreshed.

The queue resolves its current item each poll from the player's raw `current_media`, and a sync group mirrors its leader's raw value. A Universal Player returned `None` while outputting through a protocol player, so the group could never resolve the current item and `_update_queue_from_player` bailed out before ever reaching the live-metadata refresh. This fixes it at the player level, so the existing per-second update path drives the refresh for grouped AirPlay exactly like every other player — no extra background polling.

Changes:
- `UniversalPlayer.current_media` now surfaces the active output protocol player's raw `current_media` while outputting via a protocol (instead of returning `None`), so the existing update path resolves the queue item and refreshes metadata on its normal cadence.
- Reads the protocol player's *raw* `current_media` (not `.state`) to avoid routing back through the group's `__final_current_media`.
- Added a regression test for `UniversalPlayer.current_media` with an active output protocol.

- related issue https://github.com/music-assistant/support/issues/5668

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/universal_player/player.py
+++ b/music_assistant/providers/universal_player/player.py
@@ -111,6 +111,12 @@ class UniversalPlayer(Player):
         """Return the current media being played by the player."""
         if ext_player := self._get_external_source_protocol_player():
             return ext_player.current_media
+        if protocol_player := self._get_active_output_protocol_player():
+            # while playing through an output protocol player, surface its raw current_media
+            # so consumers of this player's raw value (e.g. a sync group mirroring its leader)
+            # can resolve the active queue item. Reading the protocol player's .state here would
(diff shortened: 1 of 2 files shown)


marcelveldt:
@OzGav I didnt add the backport label since I think we should first test this for a week or so in nightly/beta if it doe snot have side effects. If it all works as expected, you can later assign the label to auto backport it

PR #4313, opened 2026-06-18
fastMCP Server: Connect Wizard fixes for reverse-proxy deployments

Sawtaytoes:

Two small fixes to the MCP Server provider's **Connect Wizard**, both of which I ran into running Music Assistant behind a reverse proxy (Nginx Proxy Manager) that terminates TLS:

**1. The wizard couldn't mint tokens behind a reverse proxy.**
`_scheme_guard` only lets a credential request through if it's HTTPS, loopback, or HA ingress. With a proxy terminating TLS, MA's own socket sees plain HTTP from a public host, so `/connect/login`, `/connect/exchange`, and `/connect/token` are rejected with *"Plaintext credential traffic from non-loopback hosts is not allowed"* — even though the proxy already sends `X-Forwarded-Proto: https`. aiohttp doesn't use that header to set `request.scheme`, so there was no way to finish the wizard short of an SSH tunnel to `localhost`.

I added an opt-in **`trust_forwarded_proto`** setting (boolean, advanced, **default off**). When enabled, the guard treats the request as HTTPS if a forwarded `X-Forwarded-Proto: https` (or `X-Forwarded-Scheme: https`, which NPM uses) is present, reading the first hop of a comma-separated list. It's off by default on purpose: that header is forgeable by anything that can reach MA directly, so it should only be trusted when a reverse proxy is known to set it and strip client-supplied copies. This mirrors the existing HA-ingress exception and the `X-Forwarded-Proto` handling already in `origins.py` / `websocket_client.py`.

**2. Empty client list after username/password sign-in.**
`init()` only renders the client list and permissions when you're already authenticated at page load (a bootstrap fragment or a cached session). The interactive login handler revealed the wizard panel but never called the render, so a fresh sign-in showed an empty *"Pick your AI client"* list until you reloaded the page. I pulled the reveal-and-populate logic into a shared `showWizard()` and call it from both paths.

- none

## Types of changes

<sub>(The proxy fix is technically a new opt-in config option, but both changes exist to restore a wizard that's broken behind a reverse proxy, so I've labelled the PR `bugfix`. Happy to relabel or split into two PRs if you'd prefer.)</sub>

## Checklist

- [x] `pre-commit run --all-files` passes. <sub>(ruff, ruff-format, codespell, check_config_entries, and build_translations_source all pass on the changed files; the regenerated `translations/en.json` is committed. The only mypy output is pre-existing and in unrelated providers.)</sub>
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable. <sub>(5 new scheme-guard cases + updated entry-count assertion in `test_connect_wizard.py` / `test_config_entries.py`.)</sub>
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. <sub>(N/A — no model changes.)</sub>
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. <sub>(N/A — the wizard HTML lives in this provider's `connect/page.py`, not the frontend repo.)</sub>
- [ ] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate. <sub>(The new option is self-documented via its in-app description in `strings.json`; happy to add a docs PR for `plugins/fastmcp-server` if wanted.)</sub>


OzGav:
https://github.com/music-assistant/.github/blob/51c584e009c4600aecc4845961f4374eb5b962f2/AI_POLICY.md

Sawtaytoes:
I copied it into my local Docker container and got it working:

image

OzGav:
I don't believe it is a bug. It is not supported.

Sawtaytoes:
I reviewed the AI policy doc.

I worked with AI for a few hours before getting this PR made. I did a lot of testing beforehand and even tested it locally.

I can explain everything including the actual problem **except** what I'm going to list here:
1. I don't know enough about HTTP and HTTPS to understand `X-Forwarded-Proto` nor `X-Forwarded-Scheme`. I can learn about them, but I only know them as related to proxying. From my testing, it looks like the MCP server was looking at `http://localhost:8095` and not my proxied URL which is why it kept erroring even though I had `https` in the URL and the correct credentials:
    image
2. I'm proficient in JavaScript, but I don't Python. For instance, this has code-smell to me as a JS dev, but I have no clue if it's any good in Python:
    image
    I feel like there are better ways to write this. `return True` seems dumb when you could literally just return the result directly. But I don't know enough Python to say either way. I can ask AI to give me more options, but without a background in Python, it'd be hard for me to know.
    
I would be responsible for this code, and I'm fine with that. I'm a professional software engineer, and I did review the AI-generated code. It looked like it hit all the right spots based on the issue I ran into with proxy headers.

Sawtaytoes:
PR description updated to match PR template.

OzGav:
Please rebase to fix the test failures

Sawtaytoes:
@OzGav Rebase complete!

OzGav:
@trudenboy

marcelveldt:
@trudenboy since you are the maintainer of the MCP provider, can you give this your green light ?

Jakeshadow:
The `X-Forwarded-Proto` issue is something I ran into as well — MCP Connect wizards that rely on `request.scheme` break behind every TLS-terminating reverse proxy, and aiohttp doesn't pick up the forwarded header on its own.

`trust_forwarded_proto` (default off) is the right default stance. For anyone else landing on this PR with the same problem: if you're using nginx as the proxy, make sure your `proxy_set_header X-Forwarded-Proto $scheme;` is in the location block for the MCP endpoint, and that the proxy strips any client-supplied `X-Forwarded-*` headers. I wrote up the full nginx + MCP reverse proxy config (with TLS termination, SSE stream timeouts, and this exact header issue) at https://deploymcp.dev if you want a copy-paste starting point.

trudenboy:
> since you are the maintainer of the MCP provider, can you give this your green light ?

@marcelveldt 
Looks good from the MCP provider side - green light to merge. 👍
I'll port this into the provider repo so a forward-sync doesn't revert it.

OzGav (approved):

PR #4328, opened 2026-06-18
Only run affected tests on PRs (pytest-testmon)

marcelveldt:
Draft / proof-of-concept. Follow-up to #4325 (now merged).

Even after splitting out ffmpeg, every PR still runs the full ~4200-test suite. testmon records which tests cover which code and, on a PR, re-runs only the tests affected by the diff.

**How it works**
- Baseline (`.testmondata`) is rebuilt and cached on every push to dev/stable (full run).
- PRs restore that baseline read-only and run `pytest --testmon`, which selects only the affected tests. A one-line controller change runs a handful of tests instead of all of them.
- Cache key includes the dependency hash, so any dep change invalidates the baseline and forces a full run.
- `test-ffmpeg` is left alone (only a couple of files, not worth it).

**Caveats / why this is a draft**
- testmon can occasionally under-select (miss a test that should have run). The safety net is that **every push to dev/stable still runs the full suite**, so anything missed on a PR is caught right after merge.
- This does **not** reduce dependency install time (torch etc.) — that's the bigger chunk of the job and a separate problem (torch marker split, or running CI inside our `Dockerfile.base`).
- Dropped `--cov` on this job (testmon and coverage both hook the tracer, and nothing consumed the coverage xml anyway).

Opening as draft to validate the cache/select behaviour over a few real PRs before committing to it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -107,8 +107,24 @@ jobs:
         # --index-strategy: allow PyPI packages when also using the PyTorch extra index
         # https://docs.astral.sh/uv/pip/compatibility/#packages-that-exist-on-multiple-indexes
         run: uv pip install --system --index-strategy unsafe-best-match . .[test] -r requirements_all.txt
+      # testmon only re-runs tests whose covered code changed since the cached baseline.
+      # The baseline is rebuilt on every push to dev/stable; PRs restore it read-only.
+      # The dep hash in the key invalidates the baseline whenever dependencies change.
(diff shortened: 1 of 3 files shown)


marcelveldt:
Going a different route. testmon fights coverage/codecov (shared tracer) and can under-select. Switching to HA-style path-based selection — a PR touching only one provider runs just that provider's tests, and coverage still works (plain pytest --cov on the subset, uploaded to Codecov with flags). Follow-up PR incoming.

PR #4344, opened 2026-06-19
YouTube Music: ~1000% faster playback start by getting audio from the player API instead of yt-dlp

rathlinus:

This changes how the YouTube Music provider resolves the audio stream.

Instead of going through yt-dlp (which needs deno to solve the signature cipher
and a separate PO token server running), it now calls YouTube's player API
directly with the ANDROID_VR client. That client returns plain stream URLs that
don't need any descrambling, so yt-dlp, deno and the PO token server can all be
dropped.

The main reason is speed. In a local test, starting a song went from about 3.7
seconds down to about 0.3 seconds, because we no longer scrape the page and run
the cipher VM every time you press play.
| Phase | yt-dlp + deno cipher | after |
|---|---|---|
| Resolve | 3581 ms median | 220 ms |
| TTFB | 106 ms | 76 ms |
| click to play | 3664 ms (~3.7 s) | 291 ms (~0.3 s) |

I also changed the ytmusicapi helpers to reuse one YTMusic object per thread
instead of creating a new one on every call. ytmusicapi makes an extra request
the first time each new object is used, so the old code paid that cost on every
request. Reusing it roughly halved search and browse times locally.

A few things to be aware of:

- There is a new required setting, Visitor Data, that you copy once from a logged
  in session. It can't be derived from the cookie, and without it the player
  request gets blocked.
- The PO Token Server URL setting, the deno requirement and the "Premium
  required" check at setup are removed.
- Audio quality is a bit lower. The new path tops out around 150 kbps opus / 130
  kbps AAC. The old path could reach 256 kbps on Premium accounts, which this one
  can't _yet_.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`

## Checklist


marcelveldt:
> the "Premium required" check at setup are removed.

Which turns this into a illegal way to consume Youtube Music and is per our policy not allowed.

Also, you can argue that by not fetching the actual YTM streams but plain YouTube, this is no longer a YTM provider but more a way to listen to the audio of youtube videos, with alike quality.

Maybe you should at least offer this as an option. Prefer speed over playback-start over quality.
Also our queue logic fetches streamdetails in advance for the next track so only the first/fresh start of a song on YTM will be this slow, subsequent fetches will then be faster. Also you should probably put back a check for valid subscription to avoid Google taking this down within hours after its merged.

Anyways, let's wait until @MarvinSchenkel is back from holidays as he's the maintainer of the YTM provider.

rathlinus:
@marcelveldt  > Which turns this into a illegal way to consume Youtube Music and is per our policy not allowed.

I have it so didnt think about that, yeah we can add that back in. 

I will be working on getting the High quality streams going. Also not sure if that is interesting or possible to expose: I also have the lyrics for and the search suggestions ready if there is interest in implementing this.

rathlinus:
I have done it. I can now get the full 258 kbps in under 200ms.

PR #4375, opened 2026-06-21
Fix Party URL when webserver URL has trailing /

OzGav:

Stripping any trailing slash from base_url before appending /?join=... prevents the // when the configured webserver URL already ends in /.

- related issue https://github.com/music-assistant/support/issues/5710

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/party/__init__.py
+++ b/music_assistant/providers/party/__init__.py
@@ -514,7 +514,7 @@ class PartyPlugin(PluginProvider):
             )
 
         base_url = self.mass.webserver.base_url
-        assert base_url, "webserver base_url is not configured"
+        assert base_url  # for type-checker only
         return f"{base_url}/?join={code}"
 
     async def get_party_player(self) -> str | None:
--- a/music_assistant/providers/party/__init__.py
+++ b/music_assistant/providers/party/__init__.py
@@ -514,9 +514,8 @@ class PartyPlugin(PluginProvider):
                 f"https://app.music-assistant.io/?remote_id={remote_access.remote_id}&join={code}"
             )
 
-        base_url_value = self.mass.webserver.config.get_value("base_url")
-        base_url = str(base_url_value) if base_url_value else f"http://localhost:{DEFAULT_PORT}"
-        return f"{base_url.rstrip('/')}/?join={code}"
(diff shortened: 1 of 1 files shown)


marcelveldt:
Shouldnt we just strip it at the webserver itself then ?

OzGav:
That was option B.... if that is better I will switch to that.

OzGav:
Actually it already is stripped at the webserver,  the `base_url` property does `.removesuffix("/")`. The party plugin's actual bug is that it bypasses that property and reads the raw config value. Fixing that!

music_assistant/providers/party/__init__.py
```diff
@@ -514,8 +514,7 @@ async def get_party_url(self) -> str | None:
                 f"https://app.music-assistant.io/?remote_id={remote_access.remote_id}&join={code}"
             )
 
-        base_url_value = self.mass.webserver.config.get_value("base_url")
-        base_url = str(base_url_value) if base_url_value else f"http://localhost:{DEFAULT_PORT}"
+        base_url = self.mass.webserver.base_url or f"http://localhost:{DEFAULT_PORT}"
```
marcelveldt: base url is supposed to always have a value
maybe just do an assert to satisfy the type checker ?

OzGav: It passed mypy but I can throw one in?

OzGav: Oh hang on I think you mean vice the OR

OzGav: ok done!

music_assistant/providers/party/__init__.py
```diff
@@ -514,8 +513,8 @@ async def get_party_url(self) -> str | None:
                 f"https://app.music-assistant.io/?remote_id={remote_access.remote_id}&join={code}"
             )
 
-        base_url_value = self.mass.webserver.config.get_value("base_url")
-        base_url = str(base_url_value) if base_url_value else f"http://localhost:{DEFAULT_PORT}"
+        base_url = self.mass.webserver.base_url
+        assert base_url, "webserver base_url is not configured"
```
marcelveldt:
```suggestion
        assert base_url  # for type-checker only
```

marcelveldt (approved):
Thanks, @OzGav !

PR #4389, opened 2026-06-22
fix: restore execute bit on bundled provider streaming binaries

steamEngineer:

Provider streaming binaries (e.g. the airplay/RAOP CLI) are shipped inside the published wheel **without** the execute bit. In the slim runtime image they therefore fail to launch. This restores the execute bit in two complementary places:

- **`Dockerfile`** — after the wheel install (and again in the entrypoint as a safety net), `chmod +x` every file under `.../music_assistant/providers/*/bin/`, excluding docs/scripts/config (`*.md`, `*.sh`, `*.conf`, `.gitkeep`).
- **`airplay/helpers.py`** — `get_cli_binary()` now defensively restores the execute bit on the resolved CLI path before first use, so non-Docker installs are covered too.

- N/A

## Types of changes

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,11 @@ RUN uv pip install \
     --no-cache \
     "music-assistant@dist/music_assistant-${MASS_VERSION}-py3-none-any.whl"
 
+# Provider streaming binaries are bundled in the wheel without the execute bit.
+RUN find "$VIRTUAL_ENV/lib" -path '*/music_assistant/providers/*/bin/*' -type f \
+    ! -name '*.md' ! -name '*.sh' ! -name '*.conf' ! -name '.gitkeep' \
+    -exec chmod +x {} +
+
 # Pre-compile Python bytecode for faster startup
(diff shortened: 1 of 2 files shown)


marcelveldt:
This is not true: The binaries are shipped with the executable bit already set.

Otherwise we would have a massive amount of complaints.
How are you running MA ?

Also this shouldn't be fixed from within the code but within the docker container.
Did you test the docker container ?

steamEngineer:
Oh my gosh, I feel stupid. You're so right.

I'm running ghcr.io/music-assistant/server:beta and a bind-mount of ./server/music_assistant over site-packages (scripted rsync + container restart from WSL). That replaced the wheel-installed binaries with host copies, and those host files were missing +x in my environment...

I should have validated using the official docker image before opening this up. (Also missed the 777)

Closing this PR. I'll fix up my dev workflow. 

Sorry for the noise. _And Thank you for for everything you've done for ma!_

PR #4406, opened 2026-06-23
Recover flow stream restart on Cast groups (players that don't report idle)

OzGav:

Casting a flow-mode queue to a Google/Nest speaker **group**, playback stalls at any flow restart, for example,when a sample-rate change in `smart` mode occurs. The first track plays, then the queue sits in "playing" forever and only a manual skip recovers it. Single Cast devices are unaffected.

The flow stream is sent as `STREAM_TYPE_LIVE`. At a restart, MA ends the HTTP response; a Cast **group** treats the EOF as a
live-stream underrun and reports `BUFFERING` (which pychromecast surfaces as `player_is_playing`, so MA maps it to `PlaybackState.PLAYING`) instead of going `IDLE`. `_resume_on_idle` only restarted on `IDLE`, which never arrived leading to a permanent stall.

`_resume_on_idle` now also treats the flow as finished once the player has played to the end of the streamed audio:
- **fast path** on the player's *reported* position reaching the streamed total (the underrun report), guarded by a fresh-status check so a stale position from the pre-restart stream can't skip a short track;
- **extrapolation backstop** for players that stop sending updates entirely.

Well-behaved players keep using the existing `IDLE` path. Also clarifies the `flow_mode_sample_rate` description: the
smart/bit-perfect restart briefly interrupts players that can't switch rate gaplessly, and the fixed-rate options avoid restarts.

Also clarifies the flow_mode_sample_rate description: the smart/bit-perfect restart briefly interrupts players that can't switch rate gaplessly (grouped speakers), and the fixed rate options avoid restarts.

- related issue https://github.com/music-assistant/support/issues/5715

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues/controller.py
+++ b/music_assistant/controllers/player_queues/controller.py
@@ -139,6 +139,8 @@ class PlayerQueuesController(CoreController):
         self._transitioning_players: set[str] = set()
         self._play_action_refcount: dict[str, int] = {}
         self._last_counted_play: dict[str, str] = {}
+        # queue_id -> session_id whose flow stream was fully generated
+        self._flow_buffer_completed: dict[str, str] = {}
         self.manifest.name = "Player Queues controller"
         self.manifest.description = (
             "Music Assistant's core controller which manages the queues for all players."
@@ -907,6 +909,7 @@ class PlayerQueuesController(CoreController):
             # At this point index is guaranteed to be int
             queue.index_in_buffer = index
             queue.flow_mode_stream_log = []
+            self._flow_buffer_completed.pop(queue_id, None)
             target_player = self.mass.players.get_player(queue_id)
             if target_player is None:
                 raise PlayerUnavailableError(f"Player {queue_id} is not available")
@@ -1327,6 +1330,9 @@ class PlayerQueuesController(CoreController):
 
         # capture session_id so we can bail out if playback restarts
         original_session_id = queue.session_id
+        # record so player providers can detect flow EOF without an idle report
+        if original_session_id is not None:
+            self._flow_buffer_completed[queue_id] = original_session_id
 
         async def _resume_on_idle() -> None:
             # wait for the player to finish playing the buffered audio and go idle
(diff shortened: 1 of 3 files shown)


marcelveldt:
Is this the right place to fix it ? In the end the issue is in the cast provider so we might also investigate an alternative fix in the cast provider itself where we account for the end of stream + flow mode + receiving buffering.
Also, very weird that cast simply ignore EOF and is expecting more data to come.

OzGav:
That's what it seemed like. OK might not be until tomorrow but I'll have a look at the cast provider.

OzGav:
Looking back at my notes and the logs when I reproduced the problem. At a real track end the group reported BUFFERING with currentTime ≈ the streamed length, and never IDLE/FINISHED so from that I assumed EOF was ignored. And then the group only ever went IDLE after MA sent the new LOAD, with idleReason: 'INTERRUPTED'. 

Whilst the issue is only with Cast devices I assumed that it could happen with others hence the fix in the player queues controller. 

On doing the fix in the cast provider instead I think we would basically spot that we're in flow mode and the player's sat buffering at the end, and mark it as stopped so the normal next-track logic takes over. The one catch is telling a real end from a normal mid-song network buffer. To do that safely the provider needs to know how much we'd actually streamed, and that lives on the queue/streams side at the moment. So we'd either pass that down to the provider, or fall back to something like "it's been buffering too long" which is a bit less safe.

marcelveldt:
Or simplify it with basic check:

if seconds_elapsed > 30 and prev state == playing and new_state == buffering --> treat state as idle
the queue logic will only kick in if the full stream was consumed and were at the end of the track so all you need to fix is the state here within the cast provider, not pollute the queue controller with a massive amount of logic to cover an edge case of a single player provider.

OzGav:
I tried this and moved the detection into the cast provider where I mapped flow-mode BUFFERING-after-PLAYING to IDLE once `current_time` is past a threshold and with the changes to the queue-controller reverted.

It works fine for normal-length tracks, but it stalls on short ones. First repro track was a 14s intro (2Cellos – Thunderstruck  (Intro)) and at the rate change the group reported:

 `playerState: 'BUFFERING', currentTime: 14.258367`

 …and then just sat there (still BUFFERING through the next GET_STATUS poll, never recovered). With a 30s gate it never fires.

The problem is that `current_time` is the streamed length of the flow session at the restart, not the track length so it resets on each restart. So the problem is "short leading track in a session, then a higher-rate track forces the restart." We could accept a stall on those and lower the threshold, but it doesn't fully close it and 15s wouldn't my test case (14.26s); I'd have to drop to ~10s, which starts overlapping possible mid-song network rebuffering and risks false end-of-stream detection.

The reason the controller version doesn't need a magic number is that it compares the player's reported position against what we actually streamed (total_streamed), so it resumes correctly at ~14s.

So I see three ways to go:
1. Provider-side, fixed threshold. Simplest and keeps it out of the controller, but accepts a stall on short leading tracks 
(manual skip to recover), and no single threshold is safe (too high stalls short tracks, too low false-triggers on network  buffering).
2. Provider-side, fed the real signal. Pass total_streamed (or just a "flow stream finished sending" flag) down to the cast 
provider so its buffering→idle decision is robust. Keeps the state correction in the provider like you suggested, at the cost of the provider depending on queue/streams state.
3. Controller-side (current PR) . No magic number and no stall, because the "has it finished?" check already sits next to the total_streamed / queue_buffer_completed state it needs.

Thoughts?

marcelveldt (approved):
Thanks, @OzGav !

PR #3748, opened 2026-04-20
Plex: Add audiobook/podcast support with position sync

zenibako:
# What does this implement/fix?

Adds **library type-aware audiobook & podcast support** to the Plex provider. A single Plex provider can be configured as **music**, **audiobooks**, or **podcasts**, with auto-detection, type-appropriate sync options, multi-part streaming, and bidirectional resume-position sync.

Rebased on / integrated with the recent Plex refactor (#4179). The generic config-controller change this branch originally carried has been split out into #4414 (now a small `_build_sync_entries` refactor, no behaviour change); Plex instead exposes its dynamic, library-type-aware features by overriding the provider's `supported_features` property.

**Related issue:** N/A

## Types of changes

- [ ] Bugfix — `bugfix`
- [ ] New feature — `new-feature`
- [x] Enhancement to an existing feature — `enhancement`
- [ ] New provider — `new-provider`
- [ ] Breaking change — `breaking-change`
- [x] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## What's Changed

- **`library_type` setting** (`music` / `audiobooks` / `podcasts`) — one provider per mode; `supported_features` and the shown sync options follow the chosen type.
- **Auto-detection** — `get_section_info` inspects each section's `enableTrackOffsets` ("Store track progress") to pre-select a sensible library/type; results are cached and validated on read. Sections already used by another instance are filtered out.
- **Audiobooks** — Plex `Album`+`Track`s → MA `Audiobook`+`MediaItemChapter`s with cumulative timings and multi-part `StreamDetails` (skips tracks without media).
- **Podcasts** — Plex `Album`+`Track`s → MA `Podcast`+`PodcastEpisode`s (incl. `PODCAST_EPISODE` streaming).
- **Resume sync (bidirectional)** — `get_resume_position()` sums per-track `viewOffset`s; `on_played()` writes back via `track.updateTimeline()` / `album.markPlayed/Unplayed`.
- **Type-switch cleanup** — `update_config` removes stale provider mappings and purges old sync config values; reads raw config to avoid cross-instance recursion.
[...]

PR #4414, opened 2026-06-23
Extract library-sync config-entry building into a helper

zenibako:

Pure readability refactor of `ConfigController.get_provider_config_entries`: the library-sync `ConfigEntry` building is extracted from the body of that method into a dedicated **`_build_sync_entries`** helper. **No behaviour change.**

### Background

This branch was originally split out of #3748 to add a dynamic `get_supported_features(values)` hook in the config controller. Per review feedback (thanks @marcelveldt), that approach is dropped — providers that need dynamic features should override the `supported_features` **property** in the provider, which the controller already reads for loaded providers. So no controller-side hook is needed, and this PR keeps only the harmless extraction.

## Types of changes

- [ ] Bugfix — `bugfix`
- [ ] New feature — `new-feature`
- [ ] New provider — `new-provider`
- [ ] Breaking change — `breaking-change`
- [x] Refactor (no behaviour change) — `refactor`

## What's Changed

- **`controllers/config.py`** — moved the music library-sync `ConfigEntry` construction into `_build_sync_entries(manifest, supported_features, provider)` (placed in the private-method section). The feature resolution (loaded provider's `supported_features` property, else the static module `SUPPORTED_FEATURES`) stays inline, unchanged. Adds `import builtins` for the `builtins.set[...]` annotation (the class defines a `set` method, so a bare `set` in a signature would resolve to it).

## Testing

- New `tests/controllers/config/test_build_sync_entries.py` (4 cases): non-music provider → no entries; library features map to their sync entries; an `*_EDIT` feature surfaces the sync-back option; no features → no entries.
- `ruff` / `ruff-format` / `mypy` / `check_method_order` pass; `pytest tests/controllers/config/` passes.

## Notes

The dynamic supported-features behaviour for Plex lives in #3748 via a `supported_features` property override.


marcelveldt:
supported_features is a property in the Provider model. So you should override that property in the provider. There should be no need for the get_supported_features method

zenibako:
> supported_features is a property in the Provider model. So you should override that property in the provider. There should be no need for the get_supported_features method

OK @marcelveldt, I refactored and now it's effectively just breaking out that logic into a helper with tests.

If this is too small, I can close this PR out. I will update https://github.com/music-assistant/server/pull/3748 with these changes too.

marcelveldt:
Nice, the config controller was next on the list to get a refactor (like some of the other controllers are already undergoing). This is a nice step in that direction.

marcelveldt (approved):
Thanks, @zenibako !

PR #4415, opened 2026-06-23
Fix Home Assistant control of universal players running an external source

maximmaxim345:

Only `play`, `pause`, `stop`, `next`, `previous`, and `seek` were proxied to the
active external source protocol player. `volume_set`, `volume_mute`, and `power`
fell through to the base `Player` and raised `NotImplementedError` whenever Home
Assistant tried to control a Chromecast running Spotify Connect (or any other
external source).

The universal player now only forwards pause/seek/next/previous. Volume and mute
are left to the base `Player`, which already routes them to the protocol player,
so we no longer advertise them here. Power and `play_media` stay excluded too.

- related issue music-assistant/support#5443

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/universal_player/player.py
+++ b/music_assistant/providers/universal_player/player.py
@@ -27,8 +27,7 @@ if TYPE_CHECKING:
     from .provider import UniversalPlayerProvider
 
 
-# Volume, mute and power are intentionally excluded: the base Player resolves
-# those to the protocol player itself, so advertising them here is not needed.
+# Volume and mute are excluded: the base Player resolves those to the protocol player.
 FORWARDED_FEATURES = {
     PlayerFeature.PAUSE,
     PlayerFeature.SEEK,
--- a/music_assistant/providers/universal_player/player.py
+++ b/music_assistant/providers/universal_player/player.py
@@ -27,6 +27,15 @@ if TYPE_CHECKING:
     from .provider import UniversalPlayerProvider
 
 
+# Volume, mute and power are intentionally excluded: the base Player resolves
+# those to the protocol player itself, so advertising them here is not needed.
+FORWARDED_FEATURES = {
+    PlayerFeature.PAUSE,
+    PlayerFeature.SEEK,
+    PlayerFeature.NEXT_PREVIOUS,
+}
+
+
 class UniversalPlayer(Player):
     """
     Universal Player implementation.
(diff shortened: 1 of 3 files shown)


music_assistant/providers/universal_player/player.py
```diff
@@ -75,7 +76,7 @@ def available(self) -> bool:
     def supported_features(self) -> set[PlayerFeature]:
         """Return the supported features of the player."""
         if ext_player := self._get_external_source_protocol_player():
-            return ext_player.supported_features
+            return ext_player.supported_features - {PlayerFeature.PLAY_MEDIA}
```
marcelveldt: we shouldn't do this - only pick a curated list of features that may come from a protocol player.
For instance power is useless and we should refrain from using that from a protocol player

music_assistant/providers/universal_player/player.py
```diff
@@ -127,6 +128,27 @@ def source_list(self) -> list[PlayerSource]:
...
+            return ext_player.volume_muted
+        return self._attr_volume_muted
+
+    @property
+    def powered(self) -> bool | None:
+        """Return if the player is powered on."""
+        if ext_player := self._get_external_source_protocol_player():
+            return ext_player.powered
+        return self._attr_powered
+
```
marcelveldt:
```suggestion
```

A protocol player does not need power. If this is intended for cast players and people really want power control, they can enable a fake power control in the settings. A cast player does not have power control so its pointless to add it - previously the power control was abused to set/clear the active app which is now done at the right place

music_assistant/providers/universal_player/player.py
```diff
@@ -162,6 +184,34 @@ async def seek(self, position: int) -> None:
             await ext_player.seek(position)
             self.mass.players.trigger_player_update(ext_player.player_id, debounce_delay=2)
 
+    async def volume_set(self, volume_level: int) -> None:
+        """Handle VOLUME_SET command on the player."""
+        ext_player = self._get_external_source_protocol_player()
```
marcelveldt: I find the name of this helper method confusing... Is it an external source or is it a protocol player ?

music_assistant/providers/universal_player/player.py
```diff
@@ -162,6 +184,34 @@ async def seek(self, position: int) -> None:
...
+        if ext_player is None:
+            raise UnsupportedFeaturedException(
+                f"{self.display_name} has no active source to control volume on"
```
marcelveldt: that is weird.... you don't control the volume of a source but you control volume of a player(protocol)

music_assistant/providers/universal_player/player.py
maximmaxim345: It now only forwards pause/seek/next/previous, also dropped volume since `Player` already handles that.

music_assistant/providers/universal_player/player.py
maximmaxim345: Renamed it to `_get_protocol_player_with_external_source`.

marcelveldt (approved):
Thanks, @maximmaxim345 !

PR #4432, opened 2026-06-24
Add native player sleep timers

teancom:
This is an MVP for [#5526 ](https://github.com/orgs/music-assistant/discussions/5526), explicitly with no user-facing changes, just adding the API. Assuming this approach works we can add UI support afterwards (honestly, hoping someone else does that part - my UI / design skills are... lacking.) I've tested this locally - it's very simple, but it works. I explicitly didn't persist the timer to the database figuring that if the server restarts, all players would have stopped playback already.

ETA: https://github.com/music-assistant/models/pull/276

Add controller-level sleep timer APIs that store an in-memory expiry per player, expose the expiry through player extra attributes, stop playback via the existing player command path when the timer expires, and clean up timers when players unregister. Cover set, clear, expiry, and unregister behavior with player controller tests.

- #5526

## Types of changes

## Checklist


marcelveldt:
Can we either add a attribute to the player's extra_attributes or a dedicated attribute on the Player model ?
Next to this I think that we need an extra event specified so we will have something like "player_sleep_timer_updated"

formatBCE:
I don't think we need sleep timer updated event - it will be just end time field, right? It's not changing every second, so it would be enough to send player updated event with new value, when new duration set or it expired.

teancom:
@marcelveldt https://github.com/music-assistant/models/pull/276 will need to be merged / bumped to, I assume the tests are *just* about to fail as I type this. 😄

marcelveldt:
models PR is building and should be auto bumped soon

teancom:
@marcelveldt Re-based to get the models bump, checks are passing.

@formatBCE I didn't see your comment until I'd already done the models stuff - do you think we actively *shouldn't* have an event? Or did you just not want to gate this PR on it?

marcelveldt:
> I don't think we need sleep timer updated event - it will be just end time field, right? It's not changing every second, so it would be enough to send player updated event with new value, when new duration set or it expired.

Technically we don't NEED it (because you will simply also get the player updated event but this would be nice for components that are only interested in the sleep time change (e.g. a home assistant entity) so it doesn't hurt to have it.

music_assistant/controllers/players/controller.py
```diff
@@ -486,6 +489,55 @@ def get_player_control(
...
+            msg = "Sleep timer duration must be greater than zero seconds"
+            raise InvalidDataError(msg)
+        player = self._get_player_with_redirect(player_id)
+        try:
+            # guard against absurd durations that overflow the float timestamp math
+            expires_at = time.time() + seconds
+        except OverflowError:
+            msg = "Sleep timer duration is too large to schedule"
+            raise InvalidDataError(msg) from None
+        player._attr_sleep_timer_expires_at = expires_at
```
marcelveldt: This is actually bad. We should add a new final method to the player model to set it, which data is internally stored in a __sleep_timer_expires_at so it doesnt get modified out of our control. See some examples in the last part of the player model which is marked as internal only (e.g. the stop called logic)

formatBCE:
> do you think we actively *shouldn't* have an event? Or did you just not want to gate this PR on it?

Former. But Marcel makes a point.

formatBCE:
I'm also thinking if these things should persist between server restarts. It doesn't matter much, right? Because all players will be stopped/disconnected anyway.

teancom:
> I'm also thinking if these things should persist between server restarts. It doesn't matter much, right? Because all players will be stopped/disconnected anyway.

That was my thought, that we didn't need persistence. I *guess* if someone set a sleep timer, restarted their server, and restarted their player, all manually, then they may expect the timer to still be extant? But that's enough of an edge case / oddity I wasn't going to worry about it (especially for an explicit 'mvp').

formatBCE:
> That was my thought, that we didn't need persistence. I *guess* if someone set a sleep timer, restarted their server, and restarted their player, all manually, then they may expect the timer to still be extant? But that's enough of an edge case / oddity I wasn't going to worry about it (especially for an explicit 'mvp').

If I'm restarting playback, I will restart timer for sure. :)

music_assistant/controllers/players/controller.py
teancom: Fixed (well, updated, you tell me if this is what you were looking for).

OzGav:
Related feature request
https://github.com/orgs/music-assistant/discussions/5526

marcelveldt (approved):
Nice work @teancom !

PR #4436, opened 2026-06-25
fix(snapcast): fix ~65s stop delay caused by spurious inactivity timer

vintvinst:
Two related bugs caused ~65s delay when stopping Snapcast playback:

1. set_in_use(False) was scheduling a new call_later(60.0) timer even after request_stop_stream() had already been called explicitly. request_stop_stream() resets _stop_timer_started_at to None, which set_in_use() interpreted as 'no timer scheduled yet' and started a new 60s countdown. Fix: skip scheduling if _stop_requested is already True.

2. The inactivity timer was hardcoded to 60s, causing unnecessary delays even in non-buggy code paths. Fix: reduce to 3s, sufficient to survive skip/seek transitions.

Tested with external Snapcast server v0.31.0 + MA 2.9.3:
- Before: ~65s from cmd_pause to audio stop
- After:  ~5s (limited by FFmpeg teardown + idle_threshold 500)

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/snapcast/ma_stream.py
+++ b/music_assistant/providers/snapcast/ma_stream.py
@@ -254,9 +254,9 @@ class SnapcastMAStream:
             self._stop_timer_started_at = None
             if self._stop_timer:
                 self._stop_timer.cancel()
-        elif self._stop_timer_started_at is None:
+        elif self._stop_timer_started_at is None and not self._stop_requested:
             self._stop_timer_started_at = self._mass.loop.time()
-            self._stop_timer = self._mass.loop.call_later(60.0, self.request_stop_stream)
+            self._stop_timer = self._mass.loop.call_later(3.0, self.request_stop_stream)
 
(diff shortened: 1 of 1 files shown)


chrisuthe:
Going to leave a longer than normal comment so @OzGav  can put eyes and over-rule me. 

Two things here: You're properly catching a re-arm issue, which we should absolutely fix, so, great work!

The 60->3 change is adjusting a fallback that a healthy system shouldn't hit. No harm from my point of view in changing it but you're just cleaning up orphaned streams faster (if I understand snapcast well enough).  There was a PR that I saw go by that I think forced a default stream, but ideally I think we'd check to see if a default stream exists and just handle the stop()/regroup dynamically if needed OR just make sure even on external snapcast servers that MA is managing there is always a default group..

No harm after some research in adjusting this, but there is more to be done to ensure folks don't have to do other configuration when an external snapcast server is in use, I think.

chrisuthe (approved):
Looks good with the note that we could do a more complete fix down the road.

PR #4442, opened 2026-06-25
Make live audio analysis a passive observer so it can never stall playback

chrisuthe:

When the analysis queue fills, the producer used to wait for room — which could drag it below real-time and starve players to silence. So we drop chunks instead of ever blocking. A passive observer should lose data, never stall playback.

When a track ends, the producer has already burst-filled the queue, so the end-of-stream marker landed on a full queue and got dropped — which could leave the worker waiting forever and leak the session. So we evict one chunk to guarantee the marker always lands, and the worker always terminates on its own.

With the marker delivered reliably, the overall session timeout we'd added is redundant — and on normal tracks it would have thrown away good analysis and only logged it 5 minutes later — so we removed it. Genuinely stuck providers are already evicted by the existing per-chunk timeout.

[- related issue <link to issue>](https://github.com/music-assistant/support/issues/5718)

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/streams/audio_analysis.py
+++ b/music_assistant/controllers/streams/audio_analysis.py
@@ -294,9 +294,7 @@ class AudioAnalysisController:
             if is_last_chunk:
                 finalized = True
                 # The terminator must always land or the worker blocks forever on get().
-                # _on_chunk is the sole producer and runs single-threaded with no await
-                # between the full() check and put_nowait, so evicting one chunk to make
-                # room is race-free.
+                # Sole producer, no await before put_nowait, so evicting to make room is race-free.
                 if queue.full():
                     with contextlib.suppress(asyncio.QueueEmpty):
                         queue.get_nowait()
--- a/music_assistant/controllers/streams/audio_analysis.py
+++ b/music_assistant/controllers/streams/audio_analysis.py
@@ -287,21 +287,20 @@ class AudioAnalysisController:
 
         finalized = False
 
-        # Reuse the background path's per-track budget to bound the live session too.
-        session_timeout_seconds = max(
-            BACKGROUND_PER_TRACK_TIMEOUT_SECONDS,
-            int((streamdetails.duration or 0) * BACKGROUND_PER_TRACK_TIMEOUT_DURATION_MULTIPLIER),
-        )
-
         async def _on_chunk(position_seconds: int, pcm_data: bytes, is_last_chunk: bool) -> None:  # noqa: ARG001
             nonlocal finalized
             if finalized or session_key not in self._active_sessions:
                 return
             if is_last_chunk:
                 finalized = True
-                # Don't block _set_eof; _finalize_session still cleans up if this drops.
-                with contextlib.suppress(asyncio.QueueFull):
-                    queue.put_nowait(None)
+                # The terminator must always land or the worker blocks forever on get().
+                # _on_chunk is the sole producer and runs single-threaded with no await
+                # between the full() check and put_nowait, so evicting one chunk to make
+                # room is race-free.
+                if queue.full():
+                    with contextlib.suppress(asyncio.QueueEmpty):
+                        queue.get_nowait()
+                queue.put_nowait(None)
                 self.mass.create_task(_finalize_session())
                 return
             # Awaited inline by the audio producer — drop rather than ever block playback.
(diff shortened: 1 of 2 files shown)


marcelveldt:
huh? didnt we already have this

chrisuthe:
> huh? didnt we already have this

Yes but, I may have under-estimated the upside to the 1 second vs immediate eviction, my assumption was if we're behind... it's not going to finish in a second as it has 30s of work built up on front of it... 

Second part of this is probably a bit too defensive, really only providing value if the worker dies completely... 

Fine if you think this is low value for code bloat and want to kill it @marcelveldt

marcelveldt (commented):
Coming back to this with fresh eyes — I think it's more complicated than it needs to be. The non-blocking puts are the real fix and they're correct. But the session budget is really only papering over a leak we create ourselves by dropping the EOF sentinel. Notes inline.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -287,28 +287,42 @@ async def start_analysis(
 
         finalized = False
 
+        # Reuse the background path's per-track budget to bound the live session too.
+        session_timeout_seconds = max(
```
marcelveldt: With the sentinel delivered reliably (see the comment further down), this whole budget can go away. The 300s value is borrowed from the background path where it means a real per-track decode budget; here it's just an arbitrary "eventually give up" that only exists to unwedge the worker we leak below.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -287,28 +287,42 @@ async def start_analysis(
...
         async def _on_chunk(position_seconds: int, pcm_data: bytes, is_last_chunk: bool) -> None:  # noqa: ARG001
             nonlocal finalized
             if finalized or session_key not in self._active_sessions:
                 return
             if is_last_chunk:
                 finalized = True
-                await queue.put(None)
+                # Don't block _set_eof; _finalize_session still cleans up if this drops.
+                with contextlib.suppress(asyncio.QueueFull):
+                    queue.put_nowait(None)
```
marcelveldt: This is where it goes wrong. The producer dumps the whole decoded buffer into the queue in a burst (buffer is 60-1200s, worker drains at ~5x), so for basically any track longer than the queue depth the queue is full at EOF and this `None` gets dropped. The worker then drains its backlog and blocks forever on `queue.get()` — the session is never popped and the cancel callbacks don't fire (EOF was received) — so the worker leaks. That's the *only* reason `_finalize_session` needs a timeout at all.

Reliably deliver the terminator instead (single-threaded, no await in between, so race-free):

```python
if is_last_chunk:
    finalized = True
    if queue.full():
        with contextlib.suppress(asyncio.QueueEmpty):
            queue.get_nowait()  # evict one chunk so the sentinel always lands
    queue.put_nowait(None)
    self.mass.create_task(_finalize_session())
    return
```

Then the worker always terminates on its own and `_finalize_session` can go back to a plain `await worker`.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -287,28 +287,42 @@ async def start_analysis(
...
                 return
-            try:
-                await asyncio.wait_for(
-                    queue.put(pcm_data), timeout=REAL_TIME_PACE_INTERVAL_SECONDS_CEILING
-                )
-            except TimeoutError, asyncio.QueueFull:
-                return
+            # Awaited inline by the audio producer — drop rather than ever block playback.
+            with contextlib.suppress(asyncio.QueueFull):
+                queue.put_nowait(pcm_data)
```
marcelveldt, line 306: This is the actual fix and it's spot on: `_on_chunk` is awaited inline on the producer path (`AudioBuffer._put`), so a non-blocking drop-on-full here is all that's needed to guarantee live analysis can never slow the producer or stall playback.

music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -287,28 +287,42 @@ async def start_analysis(
...
         async def _finalize_session() -> None:
-            """Await the worker, then dispatch finalize to each provider."""
+            """Await the worker within the session budget, then finalize each provider."""
             worker = self._workers.pop(session_key, None)
             if worker is not None:
-                with contextlib.suppress(asyncio.CancelledError):
-                    await worker
+                try:
+                    with contextlib.suppress(asyncio.CancelledError):
+                        await asyncio.wait_for(worker, timeout=session_timeout_seconds)
```
marcelveldt: On timeout we *cancel* the providers — so we throw the analysis away instead of finalizing it. Combined with the dropped sentinel above, normal-length tracks (especially streaming ones, which don't get the nightly background scan) would routinely produce nothing and only log it 5 minutes later. None of this branch is needed once the sentinel is delivered reliably.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Got it, I put a timer on something because I allowed it to reliably fill the buffer to the point where the queue will be full at EOF and not ever call the actual callbacks that we intended to wrap this up.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Yea, drop a single chunk, make room for the sentinel, get a proper close out of it vs dropping everything for most songs.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: Understood, we now don't need the 300s situation, we will always make room for the sentinel so it'll always close, and not hang.

music_assistant/controllers/streams/audio_analysis.py
chrisuthe: got it!

chrisuthe:
Okay @marcelveldt  I think I captured where you wanted this to go now, forcing the marker in at the end of the file no matter what. Can you take one more look?

marcelveldt (approved):
Looks good, thanks @chrisuthe !

PR #4450, opened 2026-06-26
BBC Sounds: refactor for auntie-sounds 2.0

### LINKED ISSUE #7: Errors in metadata updates when between songs
I'm using the bbc sounds extension in Music Assistant and it's working very well. 

I am getting the following errors in the logs and it seems that the stream details don't update when there isn't a song currently playing - ie, during the links between songs. The following error is produced constantly during links but not at all during music.

It isn't actually interfering with the audio, but it notable that the metadata doesn't clear at the end of the songs, it stays showing as the previous song until the next one starts.

```
2026-03-09 15:39:08.484 ERROR (MainThread) [music_assistant] Error doing task: Task exception was never retrieved
  File "/home/lloyd/github/music-assistant-server/music_assistant/providers/bbc_sounds/__init__.py", line 360, in _update_on_demand_stream_metadata
    stream_details = await self._get_playable_stream_details(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        item_id=stream_details.item_id, media_type=stream_details.media_type
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/lloyd/github/music-assistant-server/music_assistant/providers/bbc_sounds/__init__.py", line 271, in _get_playable_stream_details
    episode_info = await self.client.streaming.get_by_pid(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        item_id, include_stream=True, stream_format=self.stream_format
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/lloyd/github/music-assistant-server/.venv/lib/python3.13/site-packages/sounds/auth.py", line 32, in _impl
    method_output = await method(self, *method_args, **method_kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lloyd/github/music-assistant-server/.venv/lib/python3.13/site-packages/sounds/streaming.py", line 196, in get_by_pid
    json_resp = await self._get_json(
                ^^^^^^^^^^^^^^^^^^^^^
        url_template=SignedInURLs.PID_PLAYABLE, url_args={"pid": pid}
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/lloyd/github/music-assistant-server/.venv/lib/python3.13/site-packages/sounds/base.py", line 129, in _get_json
    raise NotFoundError(message)
sounds.exceptions.NotFoundError: Not Found
```

kieranhogg:

This updates the BBC Sounds provider against the auntie-sounds library v2.0 API.

In the process:

- Improves its use of caching
- Fixes kieranhogg/auntie-sounds/issues/7
- Adds some extra menu items for both UK and non-UK listeners

Apologies it's a larger than ideal PR, I would have preferred split this. As it's self-contained, it started became easier to keep together so I could test the integration the moving parts.

## Types of changes

## Checklist


music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -185,34 +171,20 @@ async def handle_async_init(self) -> None:
...
-
-            try:
-                await self.client.auth.authenticate(
-                    username=str(self.config.get_value(CONF_USERNAME)),
-                    password=str(self.config.get_value(CONF_PASSWORD)),
-                )
-            except exceptions.LoginFailedError as e:
-                raise LoginFailed(e)
+        # Two simple internal caches to reduce API calls
+        self._stream_details_cache: dict[str, tuple[StreamDetails, float]] = {}
```
OzGav: What's the thinking behind the hand-rolled dictionary cache here instead of `use_cache`/`mass.cache`? If it's because` StreamDetails` can't be serialised, can we say so in a comment? As it stands it never removes old entries, so it keeps growing, and `copy.copy` only copies the top level, so `data` and` stream_metadata` are still shared with the cached version.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -407,17 +466,16 @@ async def _station_programme_display(self, station: LiveStation) -> StreamMetada
             return StreamMetadata(title=title, artist=None, image_url=station.image_url)
         return None
 
-    @use_cache(3600 * 6)  # Cache for 6 hours
```
OzGav, line 410: Same question in a couple of other places...why remove the caching?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -255,35 +220,122 @@ async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
...
+
+    async def _catch_up_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Get stream details for catch-up content."""
+        cached, expiry = self._stream_details_cache.get(item_id, (None, 0.0))
+
+        if cached is not None and time.time() < expiry:
+            self.logger.debug(f"Cache hit for {item_id}, returning")
+            return copy.copy(cached)
+
+        programme = await self._get_programme(item_id)
```
OzGav: This calls` _get_programme` (which calls `get_by_pid`), then below you call `get_by_pid` again for the same id. Can the first call just fetch the stream too, so it's one request instead of two?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -255,35 +220,122 @@ async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
...
+
+        self.logger.debug(f"{station} {ma_radio} {type(ma_radio)}")
+        raise MusicAssistantError("No valid radio stream found")
+
+    @use_cache(expiration=_Constants.DEFAULT_EXPIRATION)
+    async def _get_programme(self, pid: str) -> dict[str, Any] | None:
+        """Fetch and normalise a programme from the API."""
+        episode = await self.client.streaming.get_by_pid(
+            pid,
+            # include_stream=False,
```
OzGav: Delete?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -255,35 +220,122 @@ async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
...
-        stream_details = await self.adaptor.new_streamable_object(episode_info)
-        if not stream_details:
+
+        if not episode:
+            return None
+
+        # Normalize object to dict to avoid cache serialization issues
+        return {
+            "id": episode.id,
+            "titles": getattr(episode, "titles", None),
```
OzGav: Do you need getattr here or can the info be read directly?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -144,38 +131,37 @@ async def get_config_entries(
...
+                password=str(password),
+            )
+        else:
+            self.client = SoundsClient(
+                session=self.mass.http_session,
+                logger=self.logger,
+                timezone=LOCAL_TIMEZONE,
+            )
+            # This seems odd, but e.g. clearing the username and password will result in a stuck session
+            await self.client.logout()
```
OzGav: Should this be worked around here or in `auntie-sounds`?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -144,38 +131,37 @@ async def get_config_entries(
...
 
     async def handle_async_init(self) -> None:
         """Handle async initialization of the provider."""
-        self.client = SoundsClient(
-            session=self.mass.http_session,
-            logger=self.logger,
-            timezone=LOCAL_TIMEZONE,
-        )
-
+        # If we have an account, authenticate. Testing shows all features work without auth
```
OzGav, line 142: Previously you handled LoginFailed. Is that not needed anymore?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -255,35 +220,122 @@ async def get_podcast_episode(self, prov_episode_id: str) -> PodcastEpisode:
...
+    async def get_radio(self, prov_radio_id: str) -> Radio:
+        """Get full radio details by id."""
+        self.logger.debug(f"Getting radio for {prov_radio_id}")
+        station = await self.client.stations.get_station(prov_radio_id, include_stream=True)
+        if station:
+            ma_radio = await self.adaptor.new_object(station, force_type=Radio)
+            if ma_radio and isinstance(ma_radio, Radio):
+                return ma_radio
+        else:
+            raise MusicAssistantError(f"No station found: {prov_radio_id}")
```
OzGav: You are using `MusicAssistantError` in a number of places but we have better errors. For example, here you should use `MediaNotFoundError`

music_assistant/providers/bbc_sounds/metadata.py
```diff
@@ -0,0 +1,44 @@
...
+        if image_url and _Constants.BLANK_IMAGE_NAME in image_url:
+            image_url = None
+    elif isinstance(now_playing, dict):
+        titles = now_playing.get("titles") if now_playing else None
+        title = titles.get("secondary", "") if titles else "Unknown title"
+        artist = titles.get("primary", "") if titles else "Unknown artist"
+        image_url = now_playing.get("image_url")
+        if image_url and _Constants.BLANK_IMAGE_NAME in image_url:
+            image_url = None
+    return StreamMetadata(title=title, artist=artist, image_url=image_url)
```
OzGav: title/artist/image_url only get set inside the two branches. If it's ever neither, this throws an error

music_assistant/providers/bbc_sounds/metadata.py
```diff
@@ -0,0 +1,44 @@
...
+def _find_segment(segments: list[Segment], elapsed_time: int) -> Segment | None:
+    segment = None
+    if segments:
+        segment = next(
+            (
+                s
+                for s in segments
+                if isinstance(s, Segment)
+                and s.offset
+                and int(cast("int", s.offset.get("start")))
```
OzGav: Surprised you needed the cast here, I though the int would have been ok....

OzGav (commented):
A couple of things.

kieranhogg:
> A couple of things.

Thanks, I'll take a look, this is why reviews are super helpful!

I've not looked properly yet but some of the cache stuff was likely written to work around cache issues before the recent cache improvements, so they might be redundant now.

music_assistant/providers/bbc_sounds/adaptor.py
```diff
@@ -224,17 +231,16 @@ def _get_attr(self, obj: Any, attr_path: str, default: Any = None) -> Any:
...
 
-    type ConvertableTypes = Station | LiveStation | StationSearchResult
-    convertable_types = (Station, LiveStation, StationSearchResult)
+    ConvertableTypes = Station | LiveStation | StationSearchResult
 
-    def can_convert(self, source_obj: ConvertableTypes) -> bool:
+    def can_convert(self, source_obj: Any) -> bool:
         """Check if this converter can convert to a Station object."""
-        return isinstance(source_obj, self.convertable_types)
+        return isinstance(source_obj, self.ConvertableTypes)
```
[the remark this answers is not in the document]
kieranhogg: This isn't correct as of 3.10+:

image

music_assistant/providers/bbc_sounds/adaptor.py
```diff
@@ -622,29 +623,27 @@ async def _convert_radio_clip(self, clip: RadioClip) -> Track | MAPodcastEpisode
...
+    def can_convert(self, source_obj: Any) -> bool:
         """Check if this converter can convert to a Browsable object."""
         can_convert = False
         if self.context.force_type:
-            can_convert = issubclass(self.context.force_type, self.output_types)
+            can_convert = issubclass(self.context.force_type, self.OutputTypes)
         else:
-            can_convert = isinstance(source_obj, self.convertable_types)
+            can_convert = isinstance(source_obj, self.ConvertableTypes)
         return can_convert
```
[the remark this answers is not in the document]
kieranhogg: As above.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -566,121 +539,91 @@ async def _render_browse_item(
...
+                if sub_menu is None:
+                    break
+            else:
+                if isinstance(sub_menu, MenuItem) and sub_menu.sub_items is not None:
                     for item in sub_menu.sub_items:
-                        new_item = await self._render_browse_item(item)
-                        if new_item:
+                        if new_item := await self._render_browse_item(
+                            item, path_parts=[f"{self.domain}:/", *path_parts]
+                        ):
```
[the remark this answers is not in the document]
kieranhogg: While not the clearest, each part in path_parts is joined with "/" later, hence omitting the extra one here.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -701,19 +645,25 @@ async def browse(self, path: str) -> Sequence[MediaItemType | ItemMapping | Brow
             *[part for part in path_parts if len(part) > 0],
         ]
 
+        # A large part of the menu content is pre-loaded into self.menu
+        # These are the exceptions, so get the extra content
         if sub_path == "":
             return await self._get_menu()
+        # Categories and collections aren't in the API menus
         if sub_path == "categories" and sub_sub_path:
             return await self._get_category(sub_sub_path)
```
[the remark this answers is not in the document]
kieranhogg: As above.

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: It's to catch the very small edge case (mostly just me testing) of:
- User is logged in
- User removes username and password in MA
- Auth session/cookie then hangs around for a bit then will eventually error as there is no user/pass to renew session

It feels like a MA situation rather than a library one, but I will have a think.

music_assistant/providers/bbc_sounds/adaptor.py
```diff
@@ -387,17 +388,14 @@ def _format_podcast_episode_title(self, episode: PodcastEpisode) -> str:
...
 
-    def can_convert(self, source_obj: ConvertableTypes) -> bool:
+    def can_convert(self, source_obj: Any) -> bool:
         """Check if this converter can convert to a Podcast object."""
-        # Can't use type alias here https://github.com/python/mypy/issues/11673
         if self.context.force_type:
-            return issubclass(self.context.force_type, self.output_types)
-        return isinstance(source_obj, self.convertable_types)
+            return issubclass(self.context.force_type, self.OutputTypes)
+        return isinstance(source_obj, self.ConvertableTypes)
```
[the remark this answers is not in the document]
kieranhogg: As above.

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: I _think_ this is because to use MA's cache, I need to know what the object type is returned by get_by_pid() in advance. The only way to do that is to get the object, then convert it. Which then defeats the point of using the cache.

music_assistant/providers/bbc_sounds/metadata.py
kieranhogg: Yeah, I think the proper option is to ensure .offset.get("start") doesn't return None which will need breaking out of the generator I think.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -566,121 +509,91 @@ async def _render_browse_item(
...
+        )
+        items = []
+        if schedule and schedule.sub_items:
+            for folder in schedule.sub_items:
+                new_folder = await self._render_browse_item(folder, path_parts=path_parts)
+                if new_folder:
+                    items.append(new_folder)
+        return items
+
+    async def _get_category(
```
OzGav, line 570: You restored the cache decorator on `station_list` but not here?

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -566,121 +509,91 @@ async def _render_browse_item(
...
-                    )
-                    if station.metadata.images
-                    else None
-                ),
-            )
-            for station in await self._station_list(include_local=show_local)
-        ]
+        return []
+
+    async def _get_collection(
```
OzGav, line 584: You restored the cache decorator on `station_list` but not here?

OzGav:
Marking as draft so we can keep track of what needs attention. Please set to ready for review when the comments are resolved.

OzGav:
I see you added a 30 day cache to get_radio and  _catch_up_stream_details. Is that not too long, particularly for the latter?

kieranhogg:
> I see you added a 30 day cache to get_radio and  _catch_up_stream_details. Is that not too long, particularly for the latter?

I will happily make either a little shorter. I was fairly convinced 30 days was fine for both as the details won't change (radio stream URLs have changed every year or so IIRC), but as we do grab the stream URL with a JWT token I think I've changed my mind and I might make them both even shorter now you've asked the question.

OzGav:
You are the expert here so if they are all good then no problems!

kieranhogg:
> You are the expert here so if they are all good then no problems!

Well the other consideration, in which you definitely know more than me on, is _catch_up also sets the callback function. Is there any issues there with caching? Can't quite think it through...

OzGav:
Got Claude to trace it out and turns out it is not just the callback, StreamDetails has path (the stream URL), data, and the callback all set to serialize="omit", and `use_cache` stores as JSON. So a cache hit returns it with the URL and callback stripped. Best to not cache `_catch_up_stream_details` at all. You could cache the `get_by_pid` fetch if you want, but build the StreamDetails fresh each call. `get_radio` is fine

music_assistant/providers/bbc_sounds/metadata.py
```diff
@@ -0,0 +1,47 @@
...
+            image_url = None
+        metadata = StreamMetadata(title=title, artist=artist, image_url=image_url)
+    elif isinstance(now_playing, dict):
+        titles = now_playing.get("titles") if now_playing else None
+        title = titles.get("secondary", "") if titles else "Unknown title"
+        artist = titles.get("primary", "") if titles else "Unknown artist"
+        image_url = now_playing.get("image_url")
+        if image_url and _Constants.BLANK_IMAGE_NAME in image_url:
+            image_url = None
+        metadata = StreamMetadata(title=title, artist=artist, image_url=image_url)
```
OzGav: `currently_playing_song` returns `Segment | None` and `_find_segment` returns a `Segment`, so this is only ever called with a `Segment`. This part is therefore no longer required and you can narrow to

`def _segment_to_metadata(now_playing: Segment) -> StreamMetadata | None:`

music_assistant/providers/bbc_sounds/adaptor.py
OzGav: I know this is existing but it would be good to narrow this down

OzGav: I know this is existing but it would be good to narrow this down

kieranhogg: Very fair, I think I'll probably catch anything higher up so I can throw a relevant MA error instead.

OzGav:
```suggestion
class ConversionError(MusicAssistantError):
```
Not a blocker. Sorry I didn't notice before. The import will be needed at the top as well.

OzGav (approved):

OzGav:
Thanks Kieran. Looks good! One thing if you are keen.

music_assistant/providers/bbc_sounds/adaptor.py
kieranhogg: Added, thanks. Once this lands, the plan was to take a look at the adaptor again as I haven't touched it since the first version, and it could sure use a bit of love!

PR #4457, opened 2026-06-27
Seed Last.fm personalized rows from recent plays

OzGav:

Rank the Discover Similar Artists/Tracks seeds from the recent-play log within a rolling window instead of lifetime play_count, so seeds track current listening: stale heavily-played favourites and old one-off album playthroughs drop out once they age past the window.

The playlog stores no artist, so each recent play is resolved back to its track via a cache-preferred provider lookup (library plays resolve from the local database) to recover the artist for seeding. This avoids expanding the playlog schema while still covering streaming plays not in the library.

## Types of changes

## Checklist


music_assistant/providers/lastfm_recommendations/recommendations.py
```diff
@@ -730,58 +739,96 @@ async def _get_geo_based_recommendations(self) -> AsyncIterator[RecommendationFo
...
+    async def _resolve_seed_track(self, play: ItemMapping) -> _SeedTrack | None:
+        """
+        Resolve a recently played track to an artist/track seed, or None if unavailable.
 
-    async def _get_similar_artists_from_seeds(self, seed_artists: list[Artist]) -> list[Artist]:
+        :param play: A recent play entry from the playlog.
+        """
+        # A removed provider or a dropped streaming id leaves the play unresolvable; skip it.
+        with suppress(MediaNotFoundError, ProviderUnavailableError):
+            track = await self.mass.music.tracks.get_provider_item(play.item_id, play.provider)
```
marcelveldt: this means an api call for each track to the provider if the item is not in the cache.
consider using self.mass.music.tracks.get_library_item_by_provid which is a db call only (but will return None if the item is not in the lib). If you don't want this to be limited to the library only we'll have to revisit this heavily to somehow ensure we only fetch cached items of provider item's

marcelveldt (changes requested):

music_assistant/providers/lastfm_recommendations/recommendations.py
OzGav: Yeah. On the previous PR you suggested querying the API and I incorrectly assumed `get_track` was called during the playback path but looking at it again it isn't so those using streaming providers will end up with more API calls for sure. I note though that when this provider does its first query then all of that info is cached and will be subsequently used for all further calls. The providers currently cache track info for 30 days so that could be extended.

The problem with querying database only is that streaming provider users could be playing heaps of tracks and none of those will get used as they aren't in the library. I realised this during the shift to the use of the playlog vice the play counts and that is what kicked this whole thing off. So I think limiting this to in-library only will be a poor user experience for streaming users.

I think the original plan to add the artist name to the playlog is still valid and the simplest option?

The other thing I thought of is to cache the artist name when `mark_item_played` fires for the sole use of this provider. I would gate it on this provider being loaded so we wouldn't store the info for nothing.

marcelveldt: I now understand the problem you're facing better.

- when playing content from a streaming provider (e.g. an album, playlist) we already have the (minimal) track data available so unless the data is too minimal, we don't call the get_track endpoint to save API calls
- at time of playback the track data is available to us but we throw it away and only store the id's in the playlog table

- if you later want to query that playlog table again (which can be reconstructed into simple ItemMapping objects), you will know what item played but none of the metadata is there, including the artist(s) or album of a track

- Doing a get_item or get_track call to resolve the item for the lastfm provider will result in a bunch of api calls so if possible to avoid we should do so. Unless we think this is very limited and like you said will only be once as the data is cached after that. But if the lastfm plugin is quering different tracks all the time that won't be good.

Idea 1:
Write the track object to cache if it's already complete. A bit fragile as you do not exactly know the cache category/key per provider so you may want to write it into some special bucket.

Idea 2:
Write the artist and album to the playlog table (maybe as json objects?)

Idea 3:
Just accept the calls if we can guarantee that it will be a limited number.

OzGav: Thanks. I don't think idea 3 is the solution when we have ideas 1 and 2. Personally I prefer idea 2 as I feel it is the neatest and that data may prove useful in the future as it will be available to all.

Idea 1 can work and I was thinking of something like writing the artist (and album) into a dedicated cache bucket at mark_item_played time, only when we already have the data in hand and only when this provider is loaded so we don't store for nothing:

```
# in mark_item_played, gated on the lastfm_recommendations provider being loaded
await self.mass.cache.set(
    key=f"{media_item.provider}.{media_item.item_id}",
    data={"artist": ..., "album": ...},  # data we already have, no extra fetch
    category=CACHE_CATEGORY_RECENT_PLAY_ARTISTS,  # our own bucket
    expiration=3600 * 24 * 30,
)
```

The seed resolution would then read from our own bucket and falls back to `get_library_item_by_provid` (DB only) skipping if neither hits. Using our own category also sidesteps the fragility you mentioned about not knowing each provider's cache key/category as we're not touching the providers' `get_track` buckets, just our own.

Having said that, I still lean towards idea 2 for the durability/reuse reasons above. Happy to go whichever way you think fits best.

marcelveldt: OK, let me think about this then how to do it in a clean way because just appending columns to the playlog table breaks its compatibility with an itemmapping and that we wanted it to be as lightweight as possible

marcelveldt: artists are now present on the playlog (array of itemmappings)

OzGav: OK redid this to use the playlog artists column you added, thanks!

I have changed:

- The seed path no longer calls `get_provider_item` (the per-track API/cache lookup you flagged). It now reads the artist mappings straight off the playlog rows.
- Added a small helper on the music controller, `recently_played_tracks()`, which returns recent fully-played tracks with the artists recorded at play time.
- Plays logged before the artists column existed have no artist recorded, so they're skipped. Those rows pick up an artist again the next time the track is played, so the seed set fills back in on its own with normal listening thus no migration/backfill is needed.

OzGav:
@codspeedbot fix this regression

OzGav:
@codspeedbot fix this regression

music_assistant/controllers/music/controller.py
```diff
@@ -106,13 +100,21 @@
...
     from music_assistant.models.provider import Provider
     from music_assistant.providers.builtin import BuiltinProvider
 
 
+class RecentPlayedTrack(NamedTuple):
+    """A recently played track from the playlog, with the artists recorded at play time."""
+
+    item_id: str
+    provider: str
+    name: str
```
marcelveldt:
```suggestion
    track: ItemMapping
```

OzGav: Fixed with the conflict resolution as well.

OzGav:
@marcelveldt I think this is ready again.

marcelveldt (approved):
Not a very big fan of the named tuple but also no inspiration how to handle it differently so all good for now. Thanks, @OzGav !

PR #4461, opened 2026-06-27
Fix min/max volume scaling lost on protocol/external volume redirect

Hopperpop:
> **_NOTE:_**  Bugfix with the help of Opus4.8

When a player's volume command was redirected to a linked protocol player (e.g. a non-Google Chromecast soundbar wrapped in a UniversalPlayer) or to an external PlayerControl, the already-scaled device volume was discarded and the raw logical volume was forwarded instead. The protocol redirect then re-scaled against the protocol player's own (default 0-100) limits, so the user-facing player's min/max volume config was ignored on write — even though it was correctly honored on read, producing an asymmetry where reading the volume worked but setting it did not.

Forward the already-scaled device_volume in both the external-control and protocol-redirect branches so the min/max limits configured on the user-facing player are honored. The protocol player has no limits of its own, so its scaling is an identity pass-through.

- Not reported yet

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -3757,15 +3757,20 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
                     f"Player control {control_name} is not available"
                 )
             assert player_control.volume_set is not None
-            await player_control.volume_set(volume_level)
+            # forward the already-scaled device volume; the external control sets the
+            # raw device volume and does not apply min/max scaling of its own
+            await player_control.volume_set(device_volume)
             return
         if protocol_player := self.get_player(player.state.volume_control):
-            # redirect to protocol player volume control
+            # redirect to protocol player volume control.
+            # forward the already-scaled device volume so the min/max limits configured
+            # on this (user-facing) player are honored; the protocol player has no
+            # limits of its own, so its scaling is an identity pass-through.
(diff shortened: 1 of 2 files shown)


marcelveldt:
nice catch @Hopperpop, this looks correct to me 👍

I double checked the assumption and it holds: protocol players don't get the min/max volume config entries (they get an empty set of default entries), so `scale_volume_to_device` on the protocol player is indeed an identity pass-through and there's no risk of double scaling. Same reasoning for the external control branch.

would be nice to add a small regression test for this in a follow-up (a non-default min/max being honored when the volume command is redirected to a protocol player), but I don't want to block the fix on it.

thanks!

marcelveldt:
I went ahead and pushed the regression test directly (thanks for enabling maintainer edits 🙏). It covers both redirect paths — protocol player and external control — and asserts that a non-default `max_volume` is honored on write. I verified it fails against the old behaviour (forwarded the raw `volume_level`) and passes with your fix.

Good to merge from my side once CI is green.

marcelveldt (approved):
Thanks, @Hopperpop !

Hopperpop:
No problem. Glad I could help and glad it will be fixed. My speaker is too powerful and I would like to scale it down, but using it with the current release only makes my household angry.

PR #4470, opened 2026-06-28
Fix discover page not loading due to MusicBrainz recommendation scan

marcelveldt:

The discover page stopped loading on recent nightlies. The birthday/in-memoriam recommendations added in #3833 turned MusicBrainz into a recommendations provider, but its scan runs on the discover request itself: for every library artist with a MusicBrainz ID it fetches full artist details across a multi-day window, for both birth and death dates. MusicBrainz lookups are rate-limited to ~1/sec, so on a larger library the scan takes minutes — and because the recommendations controller awaits all providers together without a timeout, the whole discover page is blocked until it finishes. That's why the only sign in the logs is a stream of MusicBrainz rate-limit messages.

This moves the scan off the request path and adds a safety net so one slow provider can no longer block discover:

- MusicBrainz recommendations are now precomputed in a background task and served from cache, so `recommendations()` returns immediately. The cache is warmed on startup and refreshed once a day.
- The library scan is now single-pass — each artist is fetched once and bucketed by birthday and death day — instead of re-fetching per day and per date field.
- The result is cached until the next day so the relative labels (today / tomorrow / in N days) stay correct; the underlying MusicBrainz lookups remain cached and are reused on later refreshes, so the rate-limited work only happens once.
- Added a per-provider timeout in the recommendations controller so a single slow provider can never block the entire discover page again.
- Updated the MusicBrainz recommendation tests for the new background/caching behaviour.

- Regression introduced by #3833

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music/constants.py
+++ b/music_assistant/controllers/music/constants.py
@@ -18,6 +18,10 @@ DYNAMIC_RADIO_DYNAMIC_TARGET: Final[int] = 50
 
 CACHE_CATEGORY_SEARCH_RESULTS: Final[int] = 10
 
+# max time to wait for a single provider's recommendations before skipping it,
+# so one slow provider can never block the whole discover page
+RECOMMENDATIONS_PROVIDER_TIMEOUT: Final[int] = 30
+
 DATABASE_CLEANUP_TASK_ID: Final[str] = "music_database_cleanup"
 PROVIDER_MAPPING_CORRECTION_TASK_ID: Final[str] = "music_provider_mapping_correction"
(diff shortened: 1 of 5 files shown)


OzGav:
Also wondering if these rows should be able to be disabled at the provider level to save API calls for those that don’t intend to have these rows displayed?

marcelveldt:
> Also wondering if these rows should be able to be disabled at the provider level to save API calls for those that don’t intend to have these rows displayed?

Yes, I'm going to rework the recommendations engine in a follow-up.

PR #4487, opened 2026-06-29
Load Discover recommendation rows on demand and improve Recently Played

chrisuthe:

This improves the recommendations system and makes the Discover page load only the rows it actually shows, via a two-method model (as discussed in review):

- **`music/recommendations`** returns every available row (builtin + providers) as `RecommendationFolder`s **without items** — a fast call that serves as both the Discover skeleton and the row-toggle catalog.
- **`music/recommendations/items`** (`provider`, `item_id`) returns one row's items — this is where backend fetches live, so hidden rows cost nothing.

> **Note:** an earlier revision achieved partial laziness with a `wanted` row filter on the single bulk call. Per review feedback it was replaced by this full rows/items split, with the fetch-cost concern solved server-side (see §4): rows are static or served from a persistent cache, and per-row item fetches reuse the per-row/bulk caching each provider already had.

## 1. Refactor — recommendations subcontroller

Default rows live in a `mass.music.recommendations` subcontroller backed by a registry of small source units (`RecommendationSource`), each exposing a row descriptor and an items builder — the same two-method shape the providers implement. Existing rows, their stable `item_id`s and order are unchanged. Per-row failures are isolated: a throwing source or provider logs and yields an empty result instead of breaking the response. The rows call uses a short per-provider timeout (rows are contractually fast); the items call keeps the standard provider timeout.

## 2. "Recently Played" fixes ([support#5671](https://github.com/music-assistant/support/issues/5671))

- Rolls up to the container the user actually played (album / playlist / artist) instead of every constituent track, via `user_initiated_only=True` — playing three albums shows three albums.
- Podcasts and audiobooks surface as their **container** (the show, the book), not the episode/chapter; a played episode credits its parent podcast so the show appears even from episode-only listening.
- Two new granular rows: **Recent Artists** and **Recent Tracks** (with translations).

## 3. `enabled_by_default` per row

Each row carries an **`enabled_by_default`** hint (default `True`). Advanced/heavy or noisier rows ship off by default; the client hides them until the user opts in. Per-user visibility and ordering stay owned by the frontend's `discover.rows` preference (persists and syncs across clients), so the server stays stateless.

## 4. Every provider implements the two methods

`get_recommendations()` (rows, no items, fast) and `get_recommendation_items(item_id)` are now the provider contract; the old bulk `recommendations()` method is removed. Two provider shapes:

- **Static rows** (kion_music, yandex_music, deezer, musicme, neteasecloudmusic, nicovideo, opensubsonic, qqmusic, nugs, zvuk_music, bandcamp, tunein, itunes_podcasts, smart_playlist, sonic_similarity, lastfm, musicbrainz): the rows call is a hardcoded descriptor list with **zero backend I/O** (local config/auth gates only); each row's items call does that row's dedicated fetch, keeping the caching the bulk method had (folder-level `@use_cache` with `base_class=RecommendationFolder`).
- **Atomic payload** (ytmusic, apple_music, tidal, plex, soundcloud, bbc_sounds, audiobookshelf, yousee): the backend serves rows+items in one blob, so a shared `RecommendationPayloadMixin` gives them a persistent stale-while-revalidate cache with single-flight deduplication (concurrent cold calls trigger exactly one backend fetch, shielded from caller timeouts); rows derive from the cached payload with full section fidelity, items extract from the same payload. Total backend I/O is unchanged from before — one bulk fetch per TTL — while only the requested rows' items cross the wire.

Also included:

- Provider browse (`<provider>://recommendations`) reworked onto the same two methods; the user-visible browse tree is unchanged.
- The fastmcp `metadata_recommendations` tool mirrors the split: it returns row briefs (**breaking:** `item_uris` dropped — it could only be empty or force N fetches) and a new `metadata_recommendation_items` tool fetches one row's items.
- Per-provider tests assert the zero-I/O rows property, per-row fetch isolation, warm-cache-hit serving, and single-flight behavior.
- Upstream `dev` is merged in, including porting Plex's new extended recommendations / "Mixes For You" (#3736) into the payload shape.

## Dependencies (companion PRs)

- [music-assistant/models PR 302](https://github.com/music-assistant/models/pull/302) (**merged**) — adds `enabled_by_default` to the shared `RecommendationFolder`; released in `music-assistant-models==1.1.164`, which this PR pins.
- [music-assistant/frontend PR 2141](https://github.com/music-assistant/frontend/pull/2141) — fetches the row catalog, renders row shells immediately, and requests items per enabled row.

## Types of changes

## Checklist

- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked. — [models#302](https://github.com/music-assistant/models/pull/302)
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. — [frontend#2141](https://github.com/music-assistant/frontend/pull/2141)


dmoo500:
Should we perhaps consider having all RecommendationFolder providers update their systems with this revision to offer a central configuration page for enabling/disabling Recommendation Settings (for the Discover Page)? Personally, I find it somewhat annoying to have to navigate to each provider's settings to turn it on or off.

My idea is to always include a default enable/disable setting in the RecommendationFolder and then display a settings page via the UI with all providers, where everyone can make their own adjustments.

chrisuthe:
> Should we perhaps consider having all RecommendationFolder providers update their systems with this revision to offer a central configuration page for enabling/disabling Recommendation Settings (for the Discover Page)? Personally, I find it somewhat annoying to have to navigate to each provider's settings to turn it on or off.
> 
> My idea is to always include a default enable/disable setting in the RecommendationFolder and then display a settings page via the UI with all providers, where everyone can make their own adjustments.

Strongly agree — a single Discover settings surface beats per-provider toggles. This is exactly what @marcelveldt  poked me about suggesting I didn't take this far enough: every row (library + provider) declares an enabled_by_default, the client fetches a lightweight "list of rows" and toggles them in one central page, and the choices persist server-side so they sync across clients. It also lazy-loads each row's items, so disabled rows cost nothing.

marcelveldt:
Exactly 👍

music_assistant/controllers/music/recommendations/sources/base.py
```diff
@@ -0,0 +1,93 @@
...
+    UniqueList,
+)
+
+if TYPE_CHECKING:
+    from music_assistant.mass import MusicAssistant
+
+RecommendationItems = Sequence[MediaItemType | ItemMapping | BrowseFolder]
+
+
+class RecommendationSource(ABC):
```
marcelveldt: I don't think there is need for any of this if we just have the recommendations endpoint/command returning the available recommendations (without items) and a dedicated recommendations/items endpoint to get the items of a recommendation row (by key / item_id)

chrisuthe: Hm, well, I went around a bit, should have asked for some direction, I had it set up like that but most of the providers, they were going to be making the same call twice (or caching results) to do give available recommendations, and then give the items.  Didn't seem logical when it was all bundled up into a single call already, but maybe I just didn't disect the call enough per provider to realize we could really break it down and get part from each provider.

I'll dig around, see if I can get a read on what you envisioned here. I'll stick this back in draft and work through it.

marcelveldt: I think then we need to optimize the providers - they should be able to provide a list of recommendations they support (which may be static or a heavily cached dynamic list) because this is key in speeding up this whole recommendation list fetching

chrisuthe:
# Follow-up: true rows/items separation per provider

The two-method recommendations API is fully shipped; for eight payload-shaped providers,
`get_recommendations()` still derives rows from the (cached) bulk fetch. True separation would
make listing rows genuinely independent of fetching items. Ranked by tractability:

| Provider | Current bulk fetch | Separation path | Feasibility |
|---|---|---|---|
| **yousee** | One GraphQL query aliasing every section | Query section ids/names without item fields for rows; query a single aliased section for items | **High — easiest win** |
| **apple_music** | `me/recommendations?include[personal-recommendation]=contents` | Drop the `include` param for rows (returns recommendation objects with titles, no contents); fetch one row via `me/recommendations/{id}?include=contents` | **High** (verify the per-id endpoint honors `include`) |
| **deezer** (partial) | One GQL `get_recommendations` feeds 4 of its 8 rows | Per-row GraphQL queries instead of the combined one; the other 4 rows are already separate | **High** |
| **plex** | One `/hubs/sections/{key}` fetch returns all hubs with items | Each hub carries its own `key`/`hubKey` endpoint — rows from a `count=0`-style hub listing, items via the hub's own path | **Medium-high** (plexapi supports per-hub reload; `count=0` behavior needs testing) |
| **tidal** | 5 pages fetched whole, modules merged | Parse/cache page *structure* (module titles) separately from items; many modules expose a `dataApiPath` ("view all") usable as a per-row items endpoint | **Medium** — parser rework; not all modules have the per-module path |
| **bbc_sounds** | One experience-menu call | Menu items carry their own URIs — rows from the menu skeleton, items by fetching the individual container | **Medium** (depends on the client library exposing per-container fetches) |
| **audiobookshelf** | One `personalized` view per library returns all shelves | No per-shelf endpoint; would mean reimplementing known shelves via dedicated queries (`recently-added` → items sorted by addedAt, etc.), losing server-curated shelf composition | **Low-medium — probably not worth it** |
| **ytmusic** | One `get_home` blob; sections have no stable ids and locale-dependent titles | No row-list API exists; only real path is replacing home sections with dedicated endpoints (charts, moods; `mixed_for_you` already is) — which changes the product | **Low — blocked by the backend** |

All other providers (17 static-row + the single-row ones) already have full separation — no work needed.

Notes:
- The yousee/deezer (GraphQL) and apple (`include` param) calls are grounded in the current fetch
  code; the plex `count=0`, tidal `dataApiPath`, and bbc per-container claims are educated reads
  of those APIs — verify against the client libraries before picking them up.
- Even unconverted, these providers are efficient under the new architecture: the payload mixin
  caches the bulk fetch (persistent SWR, single-flight), so backend I/O is one fetch per TTL
  regardless of how many rows/items calls arrive, and only requested rows cross the wire.
  This list is optimization/polish, not a gap.

music_assistant/controllers/music/recommendations/sources/base.py
chrisuthe: Done in the latest revision — music/recommendations now returns the available rows without items, and music/recommendations/items fetches a single row's items by provider + item_id. Every provider implements the pair natively (static row lists where possible, a cached-payload mixin where the backend returns one blob). This file is what remains of the base machinery: each builtin row is now just a descriptor() + get_items() pair, mirroring the provider contract — the old bulk build() path is gone.

music_assistant/controllers/music/recommendations/controller.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> UniqueList[MediaItemType | ItemMapping | BrowseFolder]:
+        """
+        Get the items for a single recommendation row.
+
+        :param provider: The provider instance id owning the row, or "library" for builtin rows.
+        :param item_id: The item_id of the row, as returned by the recommendations listing.
+        """
+        try:
+            if provider == "library":
+                return UniqueList(await library_items(self.mass, item_id))
```
marcelveldt: why isnt "library_items" already returning a UniqueList here ?

music_assistant/controllers/music/recommendations/library.py
```diff
@@ -0,0 +1,130 @@
...
+                    MediaType.TRACK,
+                    MediaType.PLAYLIST,
+                    MediaType.ARTIST,
+                ],
+                user_initiated_only=True,
+                always_include_media_types=[MediaType.PODCAST, MediaType.AUDIOBOOK],
+            )
+        case "recently_added_tracks":
+            return await mass.music.tracks.library_items(
+                limit=10, order_by="timestamp_added_desc", summary=False
```
marcelveldt: can we turn summary to True on all these because otherwise you get a a shitton of data on each object

music_assistant/controllers/music/recommendations/library.py
```diff
@@ -0,0 +1,130 @@
...
+    from music_assistant.mass import MusicAssistant
+
+
+def library_rows() -> list[RecommendationFolder]:
+    """Return the built-in library recommendation rows, without items."""
+    return [
+        _folder("in_progress", "In progress", "in_progress_items", "mdi-motion-play"),
+        _folder("recently_played", "Recently played", "recently_played", "mdi-motion-play"),
+        _folder(
+            "recently_added_tracks",
```
marcelveldt: idea to turn all these into constants ?

music_assistant/controllers/music/recommendations/library.py
```diff
@@ -0,0 +1,130 @@
...
+                user_initiated_only=True,
+                always_include_media_types=[MediaType.PODCAST, MediaType.AUDIOBOOK],
+            )
+        case "recently_added_tracks":
+            return await mass.music.tracks.library_items(
+                limit=10, order_by="timestamp_added_desc", summary=False
+            )
+        case "recently_added_albums":
+            return await mass.music.albums.library_items(
+                limit=10, order_by="timestamp_added_desc", summary=False
```
marcelveldt: not relevant to this PR but a future reminder is that we should make all these sort options a static enum

music_assistant/controllers/music/constants.py
```diff
@@ -34,10 +34,14 @@
 # how long to cache combined search results (fast path for repeated searches)
 SEARCH_CACHE_EXPIRATION_COMBINED: Final[int] = 600
 
-# max time to wait for a single provider's recommendations before skipping it,
-# so one slow provider can never block the whole discover page
+# max time to wait for a single recommendation row's item fetch before skipping
+# the row, so one slow provider row can never block an items request
 RECOMMENDATIONS_PROVIDER_TIMEOUT: Final[int] = 30
```
marcelveldt: do we still need this ? because we are no longer fetching all items in the one single call.
get_recommendations now returns early and items is a separate call per row
or do you mean for misbehaving providers that are slow in returning their recommendations ?

chrisuthe: It's the 2nd one of those, a slow provider can timeout and now we'll just stop trying to load that row after 30 seconds, I updated the name to be more descriptive.

music_assistant/controllers/music/controller.py
```diff
@@ -1465,6 +1458,13 @@ async def mark_item_played(
                 queue_id=queue_id,
                 skip_ids=set(skip_artist_ids or ()),
             )
+        if isinstance(media_item, PodcastEpisode) and media_item.podcast:
```
marcelveldt, line 1461: unrelated change or rebase needed ?

music_assistant/controllers/music/recommendations/controller.py
chrisuthe: No good reason, removed re-wrap, good to go.

music_assistant/controllers/music/recommendations/library.py
chrisuthe: Done, good catch!

music_assistant/controllers/music/recommendations/library.py
chrisuthe: Done! Now the ids can't drift between the functions, so bonus.

music_assistant/controllers/music/recommendations/library.py
chrisuthe: 10/4

music_assistant/controllers/music/controller.py
chrisuthe: Intentional, part of this PR's Recently Played fixes that started this path... A played episode credits its parent podcast (via _credit_podcast_play, added below) so the show surfaces in Recently Played even for episode-only listening.

music_assistant/controllers/music/recommendations/library.py
```diff
@@ -0,0 +1,130 @@
...
+            "Recently added tracks",
+            "recently_added_tracks",
+            "music-note-plus",
+        ),
+        _folder(
+            "recently_added_albums",
+            "Recently added albums",
+            "recently_added_albums",
+            "music-note-plus",
+        ),
```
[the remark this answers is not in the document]
chrisuthe: Agreed and changed in https://github.com/music-assistant/server/pull/4487/commits/bfd49f03fde9ba250d963d7a3a162d49a3bf028f

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -893,40 +899,72 @@ async def _played_recommendations(self) -> RecommendationFolder:
...
+            )
+        if self._show_faves:
+            recos.append(
+                RecommendationFolder(
+                    item_id="subsonic_starred_albums",
+                    provider=self.domain,
+                    name="Starred Items",
+                    translation_key="starred_items",
+                )
+            )
```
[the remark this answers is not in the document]
chrisuthe: agreed and changed in https://github.com/music-assistant/server/pull/4487/commits/bfd49f03fde9ba250d963d7a3a162d49a3bf028f

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -893,40 +899,72 @@ async def _played_recommendations(self) -> RecommendationFolder:
...
+        if self._show_played:
+            recos.append(
+                RecommendationFolder(
+                    item_id="subsonic_most_played",
+                    provider=self.domain,
+                    name="Most Played Albums",
+                    translation_key="most_played_albums",
+                )
+            )
```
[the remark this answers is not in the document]
chrisuthe: Yes, done in https://github.com/music-assistant/server/pull/4487/commits/bfd49f03fde9ba250d963d7a3a162d49a3bf028f

music_assistant/controllers/music/recommendations/controller.py
```diff
@@ -0,0 +1,135 @@
...
+        """Initialize the controller and register its api commands."""
+        self.mass = mass
+        self.logger = logging.getLogger(f"{MASS_LOGGER_NAME}.music.recommendations")
+        self.mass.register_api_command(
+            "music/recommendations",
+            self.get_recommendations,
+            required_scope=Scope.LIBRARY_READ,
+        )
+        self.mass.register_api_command(
+            "music/recommendations/items",
```
marcelveldt, line 49: please also bump API_SCHEMA_VERSION - we add a new api command here and the existing music/recommendations changed behavior (rows without items), so clients can gate on the schema version

music_assistant/controllers/music/controller.py
```diff
@@ -2772,6 +2632,45 @@ async def _credit_artist_plays(
...
+            ":seconds_played, :timestamp, :queue_id, :user_initiated, :userid) "
+            "ON CONFLICT(item_id, provider, media_type, userid) DO UPDATE SET "
+            "name = excluded.name, image = excluded.image, "
+            "fully_played = excluded.fully_played, seconds_played = excluded.seconds_played, "
+            "timestamp = excluded.timestamp, queue_id = excluded.queue_id, "
+            f"user_initiated = {DB_TABLE_PLAYLOG}.user_initiated OR excluded.user_initiated"
+        )
+        playlog_entry: dict[str, Any] = {
+            "item_id": podcast.item_id,
+            "provider": podcast.provider,
```
marcelveldt: this stores the provider-scoped ids of the episode's parent, while an explicit play of a library show stores a library-scoped row - so the same show can end up twice in recently played. can we resolve to the library item first like _credit_artist_plays does, and only fall back to the provider ids when the podcast is not in the library ?

music_assistant/controllers/music/recommendations/library.py
```diff
@@ -0,0 +1,160 @@
...
+) -> UniqueList[MediaItemType | ItemMapping | BrowseFolder]:
+    """Return the items for one built-in library row (empty for an unknown item_id)."""
+    items: Sequence[MediaItemType | ItemMapping | BrowseFolder]
+    match item_id:
+        case LibraryRowID.IN_PROGRESS:
+            items = await mass.music.in_progress_items(limit=10)
+        case LibraryRowID.RECENTLY_PLAYED:
+            items = await mass.music.recently_played(
+                limit=10,
+                media_types=[
```
marcelveldt, line 104: was it intentional to leave genre out here ? playing a genre writes a user-initiated playlog entry so with the old MediaType.ALL behavior genres showed up in recently played. fine by me if intentional, otherwise add MediaType.GENRE

music_assistant/models/recommendation_payload.py
```diff
@@ -0,0 +1,142 @@
...
+                return folder.items
+        return UniqueList()
+
+    async def _recommendation_payload(self) -> list[RecommendationFolder]:
+        """Return the full recommendations payload, cached and deduplicated."""
+        task = self._recommendation_payload_task
+        if task is None or task.done():
+            # single-flight: share one in-flight fetch between concurrent callers,
+            # as the cache decorator does not dedupe concurrent cold misses
+            task = asyncio.create_task(self._cached_recommendation_payload())
```
marcelveldt: small race here: use_cache stores its result in a background task, so a call landing right after this task completes (but before the store lands) misses the cache and does a second full backend fetch - the typical cold start rows -> items sequence hits exactly that. keep the last payload in memory until the store lands (or store it explicitly) - that also avoids re-deserializing the full payload from the cache db on every warm call

tests/models/test_recommendation_payload.py
```diff
@@ -0,0 +1,380 @@
...
+            return None, False, False
+        data = self._cache_store[key]
+        base_class = kwargs.get("base_class")
+        if base_class is not None and data is not None:
+            # mirrors production get_with_freshness: a list of dicts reconstructs as a
+            # list of base_class instances, a single dict as one base_class instance
+            if isinstance(data, list):
+                return [base_class.from_dict(item) for item in data], True, True
+            return base_class.from_dict(data), True, True
+        return data, True, True
```
marcelveldt: the fake always returns is_fresh=True so the stale-while-revalidate path is never exercised at mixin level. can you add an expiry flag and a test that stale data is served immediately while exactly one background refresh is scheduled ? bonus: a test where the sole waiter is cancelled (the rows timeout case) and the fetch still warms the cache

music_assistant/models/recommendation_payload.py
```diff
@@ -0,0 +1,142 @@
...
+    concurrent fetches.
+    """
+
+    _recommendation_payload_task: asyncio.Task[list[RecommendationFolder]] | None = None
+    _recommendation_refresh_task: asyncio.Task[list[RecommendationFolder]] | None = None
+
+    async def _recommendation_rows_from_payload(self) -> list[RecommendationFolder]:
+        """Return all payload folders as rows, without items."""
+        return [
+            RecommendationFolder(
```
marcelveldt: this copy drops type, image, is_playable and media_type from the payload folders - inert today, but yousee already passes media_type args that now do nothing and a future provider setting type/image would silently lose them here. just copy those fields too

music_assistant/models/recommendation_payload.py
```diff
@@ -0,0 +1,142 @@
...
+    _MixinBase = _RecommendationPayloadHost
+else:
+    _MixinBase = object
+
+# Cache settings of the payload entry, shared between the @use_cache decorator on
+# _cached_recommendation_payload and the explicit store in _refresh_recommendation_payload.
+# The key mirrors use_cache's key construction: the wrapped function's __name__, with no
+# further arguments appended as that method takes none.
+_PAYLOAD_CACHE_KEY = "_cached_recommendation_payload"
+_PAYLOAD_CACHE_EXPIRATION = 3600
```
marcelveldt: this fixed 1h TTL replaces the longer intervals some of these providers had (yousee 24h, soundcloud/plex/bbc 3h) so those backends now get hit a lot more often - make this an overridable class attribute ?

music_assistant/models/recommendation_payload.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Mixin serving recommendation rows and items from a single (cached) bulk payload.
+
+    The provider implements _fetch_recommendation_payload() with its existing bulk
+    backend fetch+parse; the mixin derives both the fast rows call and the per-row
+    items call from that payload, with persistent SWR caching and deduplication of
+    concurrent fetches.
+    """
+
+    _recommendation_payload_task: asyncio.Task[list[RecommendationFolder]] | None = None
```
marcelveldt, line 81: these task slots are created with raw asyncio.create_task: not cancelled on provider unload, and on python 3.14 a shielded fetch that fails after its waiter timed out ends up as a raw loop-handler error instead of the provider logger. track them via mass.create_task or cancel them from unload, and attach a done-callback for the logging

music_assistant/providers/lastfm_recommendations/__init__.py
```diff
@@ -190,9 +195,34 @@ async def _refresh_recommendations(self) -> None:
...
+
+    async def get_recommendation_items(
+        self, item_id: str
+    ) -> UniqueList[MediaItemType | ItemMapping | BrowseFolder]:
+        """
+        Get the items for a single recommendation row.
+
+        :param item_id: The item_id of the row, as returned by get_recommendations.
+        """
+        for folder in self._recommendation_folders:
```
marcelveldt, line 226: _refresh_recommendations clears and repopulates this list in place, so during the (slow, rate limited) rebuild this returns empty for rows that were just listed. build into a local list and swap it at the end ?

music_assistant/models/recommendation_payload.py
chrisuthe: Made it an overridable class attribute (recommendation_payload_ttl); yousee keeps 24h, soundcloud/plex/bbc_sounds 3h.

music_assistant/models/recommendation_payload.py
chrisuthe: background work now goes through mass.create_task with done-callbacks routing exceptions to the provider logger and the task handles stay exposed for unload cancellation

music_assistant/models/recommendation_payload.py
chrisuthe: Switched to dataclasses.replace(folder, items=UniqueList()) so every field survives and added a test constructing a folder with image/type/is_playable/media_type set.

music_assistant/models/recommendation_payload.py
chrisuthe: Good catch.... mixin now keeps the payload in memory (memory is set before the persistent store is scheduled), so the cold rows→items sequence does exactly one fetch, and warm calls skip the cache-db deserialization entirely + added test

tests/models/test_recommendation_payload.py
chrisuthe: Test added for the stale data with a single background refresh, plus the bonus test to ensure a single cancelled waiter still warms the cache.

music_assistant/providers/lastfm_recommendations/__init__.py
chrisuthe: Agreed, now builds into a local list, swaps only on success (a failed rebuild also keeps the old rows now). Test to make sure mid-rebuild results in previous generation being served.

music_assistant/controllers/music/recommendations/library.py
chrisuthe: Not intentional! Good catch, GENRE restored.

music_assistant/controllers/music/controller.py
chrisuthe: Yes we can, that is a better way to handle it and this now resolves the parent to its library item first like _credit_artist_plays, provider ids only as fallback.  A couple tests to cover both the no-duplicate and fallback cases included.

music_assistant/controllers/music/recommendations/controller.py
chrisuthe: Bumped to 39

marcelveldt (approved):
Amazing work, @chrisuthe !

chrisuthe:
@dmoo500  - I believe this was blocking some of you work, should be good to go now!

PR #4488, opened 2026-06-29
Fix startup crash from provider config entry missing 'domain'

chrisuthe:
When a provider failed to load, its last_error was written back to the provider config key via a guard-free set(). If the config had been removed in the meantime (e.g. removing an unsupported provider while a load/retry was still in flight), set() auto-created the parent dict, leaving a stub entry with no 'domain' key. get_provider_configs then crashed on prov_conf["domain"], which fails startup via create_builtin_provider_config (KeyError: 'domain'), unrecoverable even in safe mode.

Harden both sides:
- get_provider_configs / get_provider_config skip (or clearly reject) any entry without a 'domain', so a malformed entry can never wedge startup.
- Add a guarded update_provider_last_error() that only writes when the config entry still exists, and use it from the load-failure paths so a removed provider is never resurrected as a domain-less stub.

[Issue 5728](https://github.com/music-assistant/support/issues/5728)

## Types of changes

## Checklist


music_assistant/controllers/config/providers.py
```diff
@@ -85,12 +86,18 @@ async def get_provider_configs(
         prov_entries = {x.domain for x in self.mass.get_provider_manifests()}
         configs: list[ProviderConfig] = []
         for prov_conf in raw_values.values():
-            if provider_type is not None and prov_conf["type"] != provider_type:
+            # guard against malformed/partial entries (e.g. a removed provider that had an
+            # error written back to its key after removal, leaving a stub without a domain).
+            # Skip them rather than crash on a missing key. See #5728.
```
marcelveldt: instead of working around it here, shouldnt we just clean malformed entries in the migration logic ?

chrisuthe: @marcelveldt  good point, I was pre-occupied with "It worked before, just guard against it" vs the obvious choice to handle it completely. This update should be more of what you're looking for I think.

marcelveldt:
Looks good now @chrisuthe but do note that this is not automatically backportable to stable (since dev deviated too much from stable now with all the refactoring) - if it needs to be backported, it needs to be done manually against the stable branch.

marcelveldt (approved):
Nice work @chrisuthe !

PR #4494, opened 2026-06-29
Handle Spotify's refresh-token changes

OzGav:

Sources:
https://developer.spotify.com/blog/2026-06-18-refresh-token-expiration
https://developer.spotify.com/documentation/web-api/tutorials/refreshing-tokens

Spotify has changed how refresh tokens work: they no longer always return a new refresh token when you refresh an access token, and they've introduced a 6-month expiry on them. The first change was crashing setup with a custom client ID (KeyError: 'refresh_token'), leaving the provider stuck in "requires attention". We now keep using the existing token when Spotify doesn't send a new one (as their docs instruct).

For the second change, an expired token now returns `invalid_grant`, which we treat as unrecoverable, clearing the stored credentials and prompting re-authentication rather than silently failing to refresh.

- related issue reported here https://github.com/orgs/music-assistant/discussions/5719

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/spotify/helpers.py
+++ b/music_assistant/providers/spotify/helpers.py
@@ -80,7 +80,11 @@ async def get_spotify_token(
                 # enforces a 6-month lifetime); retrying won't recover it, so fail now and
                 # let the caller clear the stored token and prompt re-authentication.
                 if "invalid_grant" in err or "revoked" in err:
-                    raise LoginFailed(f"Refresh token no longer valid for {session_name}: {err}")
+                    raise LoginFailed(
+                        f"Refresh token no longer valid for {session_name}: {err}",
+                        translation_key="refresh_token_invalid",
+                        translation_owner="provider.spotify",
+                    )
                 # the token failed to refresh, we allow one retry
                 await asyncio.sleep(2)
                 continue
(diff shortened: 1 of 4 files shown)


music_assistant/providers/spotify/helpers.py
```diff
@@ -76,14 +76,20 @@ async def get_spotify_token(
         ) as response:
             if response.status != 200:
                 err = await response.text()
-                if "revoked" in err:
-                    raise LoginFailed(f"Token revoked for {session_name}: {err}")
+                # invalid_grant means the refresh token is revoked or expired (Spotify
+                # enforces a 6-month lifetime); retrying won't recover it, so fail now and
+                # let the caller clear the stored token and prompt re-authentication.
+                if "invalid_grant" in err or "revoked" in err:
+                    raise LoginFailed(f"Refresh token no longer valid for {session_name}: {err}")
```
marcelveldt: if you want, you can also add a translation key to the exception and add it to the provider's strings.json so we can return a human friendly and translatable error to the user

OzGav: OK I think I have set it up correctly.

marcelveldt (approved):
Thanks, @OzGav !

marcelveldt:
@OzGav note that is not automatically backportable to stable. If you want that, you need to do so manually and open a PR against stable branch

PR #4510, opened 2026-06-30
Enable Don't Stop the Music mode for metadata/plugin similar-track providers

OzGav:

The autoplay config gated the 'similar' mode on music providers only, so a similar-tracks source that is a metadata or plugin provider (e.g. sonic_similarity) could not enable it, even though the runtime fetch path already consults those providers. Gate on get_providers_supporting_feature instead, matching the fetch path.

- related issue https://github.com/music-assistant/support/issues/5752

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues/controller.py
+++ b/music_assistant/controllers/player_queues/controller.py
@@ -314,9 +314,10 @@ class PlayerQueuesController(CoreController):
         :param playlist_options: Library playlists to offer for the 'playlist' autoplay mode.
             Only populated when serving the entries to the UI; the parse path can omit it.
         """
-        similar_tracks_available = any(
-            ProviderFeature.SIMILAR_TRACKS in provider.supported_features
-            for provider in self.mass.music.providers
+        # Match the runtime fetch path: similar tracks can also come from metadata/plugin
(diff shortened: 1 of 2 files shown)


marcelveldt:
this fix needs to be made against stable branch only - per definition not against dev as dev is completely deviated (and this will be fixed in another way).

So if you want to fix this in stable, change the PR title to refer to the old feature name and only fix it in stable.
dev will be fixed during the current rebuild of this feature anyways.

music_assistant/controllers/player_queues/controller.py
```diff
@@ -153,42 +314,147 @@ def get_queue_config_entries(
         :param playlist_options: Library playlists to offer for the 'playlist' autoplay mode.
             Only populated when serving the entries to the UI; the parse path can omit it.
         """
-        return queue_config_entries(self.mass, playlist_options)
+        # Match the runtime fetch path: similar tracks can also come from metadata/plugin
```
MarvinSchenkel, line 317: I don't think we need a comment here at all, it's pretty obvious :).

OzGav:
OK closing in favour of https://github.com/music-assistant/server/pull/4512 which targets the stable branch.

PR #4514, opened 2026-07-01
Fix first queued item being skipped when playing onto an idle queue

marcelveldt:

`PLAY` onto an idle or empty queue started playback at the **second** enqueued item instead of the first. On an idle queue `current_index` is `None`, so `insert_at_index` resolved to `1` and `play_index()` jumped past index `0` — skipping the first track whenever several items were enqueued at once (e.g. playing an album or a multi-track selection onto a stopped player).

`PLAY` now inserts at and starts from index `0` when the queue is idle/empty. Active-queue behaviour is unchanged (insert right after the current/buffered index and jump to it), and the `NEXT` / `ADD` / `REPLACE_NEXT` branches are untouched. The bug predates the player-queues refactor (#4509).

## Types of changes

### DIFF
--- a/music_assistant/controllers/player_queues/queue_loader.py
+++ b/music_assistant/controllers/player_queues/queue_loader.py
@@ -121,13 +121,16 @@ class QueueLoaderMixin(_PlayerQueuesBase):
             return
         # handle play: replace current loaded/playing index with new item(s)
         if option == QueueOption.PLAY:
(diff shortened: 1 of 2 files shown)


marcelveldt:
backport needs to be manual towards stable as dev has a fully refactored queue controller (not compatible at all anymore with stable)

PR #4531, opened 2026-07-02
Remove obsolete AirPlay ignore volume reports setting

MarvinSchenkel:

The AirPlay `ignore_volume` ("Ignore volume reports sent by the device itself") setting is now obsolete and actively counterproductive.

#3980 fixed the root cause of the volume jumps on protocol-linked native receivers (e.g. HEOS/Denon): `sync_volume_level` no longer copies the parent's receiver-scale volume onto the AirPlay child, and instead keeps the child volume learned from DACP feedback. That fix *relies on* DACP volume feedback being processed. The `ignore_volume` setting suppresses exactly that feedback, so enabling it makes MA send a stale volume on stream (re)start — the very snap users were trying to avoid.

For a standalone AirPlay device (no protocol parent competing for volume control) the setting served no purpose either. So it has no valid use left and is removed. The hardcoded Apple-device guard (Apple reports on its own volume scale) is kept.

- related issue https://github.com/music-assistant/support/issues/5603

## Types of changes

## Changes

- Remove the `ignore_volume` config entry and all reads of it (`player.py`, `provider.py`, `constants.py`).
- Keep the internal Apple-device volume-feedback guard.
- Drop the setting from `strings.json` / `en.json` and the provider README.
- Remove the now-defunct test and config mock.


MarvinSchenkel:
Closed because apparently this is needed for some devices that send messed up volume reports..

PR #4561, opened 2026-07-03
Fix snapcast announcement volume being restored before playback finished

MarvinSchenkel:

Snapcast announcements restore the player volume (and switch the group back to the previous stream) as soon as the *server* has received all announcement audio. Snapclients however play audio `buffer_ms` (default 1000ms) behind the server, so the volume was reverted ~1 second before the announcement actually finished on the speakers, and the tail of the announcement got clipped by the stream switch-back.

- Wait for the client buffer to play out after the announcement stream finished, before restoring volume and switching the group back to the previous stream.
- Apply the same buffer wait before raising the volume on external servers too (previously built-in only).
- The existing "Snapserver buffer size" setting is now also shown when using an external server (the buffer cannot be read via the Snapcast API, so the user must mirror the server's `stream.buffer` setting there). It keeps configuring the buffer of the built-in server as before.
- Add regression tests for the announcement call ordering.

- related issue https://github.com/music-assistant/support/issues/4377

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/snapcast/__init__.py
+++ b/music_assistant/providers/snapcast/__init__.py
@@ -78,18 +78,6 @@ async def get_config_entries(
 
     return (
         ConfigEntry(
-            key=CONF_SERVER_BUFFER_SIZE,
-            type=ConfigEntryType.INTEGER,
-            range=(200, 6000),
-            default_value=1000,
-            required=False,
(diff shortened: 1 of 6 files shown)


MarvinSchenkel:
Waiting for user confirmation to see if the issue is fixed

PR #4568, opened 2026-07-03
Pace background audio analysis to stop it saturating the CPU

oldrobotdev:

The background loudness scan runs its ffmpeg processes without throttling, pegging every core on small hosts. On the reporter's Pi CM5 that is ~95% CPU and 85C for the whole nightly run with ~18k tracks pending (music-assistant/support#5781).

The loudness provider does its analysis inside an ffmpeg subprocess, so it does not pass through the offload governor (`_run_offloaded`) that bounds the Python analysis providers. Its only backpressure is the dispatch loop in `_run_background_streaming_inner`, which used to sleep a floor between chunks. That floor was removed in #4449 when the realtime path was reworked. This restores it (`BACKGROUND_PACE_INTERVAL_SECONDS_FLOOR = 0.250`), capping each scanned track at ~4x realtime. The sleep backpressures both the source-decode and ebur128 ffmpeg processes through their pipes, so total CPU work is unchanged, just spread over wall time.

Measured on 8 FLACs (3 min each, 2 tracks concurrent, one core per ffmpeg): mean ffmpeg CPU while active drops from ~190% to 2.5% of a core, peak aggregate from 577% to 7.5%, and the ebur128 result is byte-identical. A large first-time library will not finish in one 4h run and resumes across nights, which the scan already handles with its run budget and deferred counter.

**Related issue:**

- related issue https://github.com/music-assistant/support/issues/5781

## Types of changes

## Checklist

- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable. (`tests/controllers/streams/test_audio_analysis.py`, incl. a new pacing test; ruff, ruff-format and mypy are clean on the changed files.)
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. (n/a)
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. (n/a)
- [ ] I have raised a PR against the documentation repository. (n/a — no user-facing docs change)

### DIFF
--- a/music_assistant/controllers/streams/audio_analysis.py
+++ b/music_assistant/controllers/streams/audio_analysis.py
@@ -53,11 +53,8 @@ BACKGROUND_SCAN_RUN_BUDGET_SECONDS = 4 * 3600
 # player streams, so a chunk may wait behind other work before it computes.
 CHUNK_HANG_GUARD_SECONDS = 120.0
 # Floor on wall-seconds between consecutive background chunk dispatches (one chunk = one
-# audio-second), capping each scanned track at ~4x realtime. The offload governor (niced
-# pool, half-the-cores semaphore, solo lock) only bounds Python offloads; providers that do
-# their work in ffmpeg subprocesses (loudness ebur128) are otherwise unthrottled, and an
-# unpaced scan runs source-decode plus analysis ffmpeg flat-out per track, pegging small
-# hosts for the whole run. Pacing the dispatch backpressures both processes via their pipes.
+# audio-second), capping each scanned track at ~4x realtime so a background analyse doesn't
+# consume all resources. Nice and slow is preferred for nightly background scans.
 BACKGROUND_PACE_INTERVAL_SECONDS_FLOOR = 0.250
 # OS nice value for analysis worker threads (Linux): keeps analysis below playback so the
 # scheduler favors the event loop and ffmpeg under contention.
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/streams/audio_analysis.py
```diff
@@ -52,6 +52,13 @@
 # treated as stuck and evicted. Generous because analysis runs one offload at a time while a
 # player streams, so a chunk may wait behind other work before it computes.
 CHUNK_HANG_GUARD_SECONDS = 120.0
+# Floor on wall-seconds between consecutive background chunk dispatches (one chunk = one
+# audio-second), capping each scanned track at ~4x realtime. The offload governor (niced
+# pool, half-the-cores semaphore, solo lock) only bounds Python offloads; providers that do
+# their work in ffmpeg subprocesses (loudness ebur128) are otherwise unthrottled, and an
+# unpaced scan runs source-decode plus analysis ffmpeg flat-out per track, pegging small
+# hosts for the whole run. Pacing the dispatch backpressures both processes via their pipes.
```
marcelveldt:
```suggestion
# Floor on wall-seconds between consecutive background chunk dispatches (one chunk = one
# audio-second), capping each scanned track at ~4x realtime so a background analyse doesn't
# consume all resources. Nice and slow is preferred for nightly background scans.
```

oldrobotdev: Thanks, applied.

marcelveldt (approved):
Thanks, @oldrobotdev !

PR #4576, opened 2026-07-04
AudioMuse-AI provider

NeptuneHub:

This PR enable Music Assistant to have an AudioMuse-AI plugin in order to use it for sonic analysis with:
- Jellyfin
- Navidrome

This first PR only introduce as functionality:
- Radio => it will use AudioMuse-AI similar track
- Search Text => it will use DCLAP text search

It also fix:
- Give priority to result from AudioMuse-AI plugin in Radio functionality
- Jellyfin 12.0 auth token
- disables shuflling when the plugin is activated because ored matter

No issue but related discussion here: https://github.com/orgs/music-assistant/discussions/5720

I tested it with Jellyfin 12.0 RC2 and Navidrome 0.62.0 by compiling in this way on my local machine:
```
# one-time: install ffmpeg + uv if you don't have them
sudo apt install -y ffmpeg
curl -LsSf https://astral.sh/uv/install.sh | sh

# from the repo root:
git checkout audiomuse-ai     # you're already on this branch
scripts/setup.sh              # creates .venv, installs deps + pre-commit
source .venv/bin/activate     # activate the venv
python -m music_assistant --log-level debug
Then open http://localhost:8095.

# Each subsequent run only needs the last two lines:

source .venv/bin/activate
python -m music_assistant --log-level debug
```

## Types of changes

## Checklist


OzGav:
Tests are failing. You need to run pre-commit before pushing.

There also appears to be unrelated changes to the music controller, Jellyfin provider and Claude.md. 

Also I have concerns about the changes to the tracks controller. Plugin/metadata providers now take priority over the track's own music provider for all users, not just those running AudioMuse. That needs discussion before it goes in.

There is also a separation of concerns violation. `radio_playlist` shouldn't know anything about `audiomuse_ai`. If ordered similarity results need different handling, expose that through the framework (a provider feature/property) or a `radio_playlist` config option, so any provider can benefit and the coupling is removed.

So I think some explanation and discussion is required for these structural questions.

NeptuneHub:
The main changes are:
- **Adding of the New AudioMuse-AI plugin provider**
- **Remove shuffle**: I noticed that by default MA shuffle the result, but this don't make sense shuffle everything by default. Plugin return return a specific order that must be mantained. For this reason for AudioMuse-Ai i removed the shuffle;
- **Give priority to the plugin result**: This is also important, otherwise the plugin result for RadioMix are not used.

If there is something that could be done in a better and different way, please give me more example of something that did a similar implementation or explain be better and I'll fix.

For some reason run  `uv run pytest` on my local machine is not fesable because it takes hours. If github will raise other issue I'll check for sure.

Also if someone would like to help in the implementation, I'll be very happy: my final goals is to give an alternative to the final user and each help to reach this golas is really well accepted.

OzGav:
Thanks for the explanation. The goals make sense. The issue is how they're implemented.

Ordering in `radio_playlist`: the problem is your change checks whether the `audiomuse_ai` provider exists, not whether it produced the tracks. Providers must never reference each other by name. Two acceptable approaches: (1) add something generic like `ProviderFeature.ORDERED_SIMILARITY` (or a property on `PluginProvider`) that `radio_playlist` checks on the provider that actually returned the results, or (2) make it a `radio_playlist` config option ("preserve provider order"). Either way `audiomuse_ai` never appears in `radio_playlist` code.

Priority in the tracks controller: your change makes plugin/metadata providers beat the track's own music provider for every user, including people who don't run AudioMuse. That's a behaviour change we can't just slip in. Look at the existing provider priority attribute, maybe respecting that instead of hardcoding the tier order would let users/setup decide, and is worth discussing with Marcel before you rework it.

These are large decisions that you will need guidance from Marcel or Marvin I would say.

As for tests:, you don't need the full suite locally. Run just the relevant files, e.g. `uv run pytest tests/controllers/music/ tests/providers/audiomuse_ai/` . That takes seconds. And `pre-commit run --all-files` catches the lint/format failures before you push.

Also please split the music controller sync fix and the Jellyfin auth fix into their own PRs as they're unrelated and the Jellyfin one reaches into aiojellyfin private attributes, which will break when that library changes.

OzGav:
Marking as draft until these issues are resolved.

NeptuneHub:
In the last commit I did the implementation suggested AND I removed the extra fix to place in a future PR as you suggested. Please let me know if now could work.

music_assistant/controllers/music/media/tracks.py
```diff
@@ -333,89 +333,127 @@ async def similar_tracks(
         """
         Get a list of similar tracks for the given track.
 
+        Results accumulate (de-duplicated) up to ``limit`` from every provider
+        offering similarity, consulted in order of each provider's ``priority``
```
MarvinSchenkel, line 337: I don't see why this core functionality needs to be so rigorously adjusted for adding a new provider? Could you elaborate on this or split this in a separate PR to keep the scope of this one 'AudioMuse provider' only?

music_assistant/models/plugin.py
```diff
@@ -32,6 +32,22 @@ class PluginProvider(Provider):
     Plugin Provider implementations should inherit from this base model.
     """
 
+    @property
```
MarvinSchenkel: Same question here, why do we need to change the plugin model?

music_assistant/controllers/music/media/tracks.py
NeptuneHub: Because it was poorly implemented in a no-sense way.
It give priority to the Music Provider and only if Music Provider don't return nothing it give priority to plugin provider.
This is a no sense because Music Provider notiusly more or less return something, but this something is low quality. So you install a plugin provider to increase the quality of the Music provoider, but in this way it is never called.

This is **MANDAOTRY** for having AudioMuse-AI plugin provider working, I'll not split the PR anymore.

music_assistant/models/plugin.py
NeptuneHub: Removed as you requested.

music_assistant/controllers/music/media/tracks.py
MarvinSchenkel: Ah I get you, the plugin would never be called for similar tracks in case a music provider would already return a sufficient amount of similar tracks.

I'd rather you raise this change in a separate PR because I want to centralize the discussion. This is a **very** important part of MA and I would like to properly solve this. That way we can focus on the Audiomuse provider in this PR.

music_assistant/providers/radio_playlist/__init__.py
```diff
@@ -173,26 +177,38 @@ async def get_dynamic_tracks(
...
+                    seen_dynamic.add(track)
+                    dynamic_tracks.append(track)
                 if len(dynamic_tracks) >= DYNAMIC_RADIO_DYNAMIC_TARGET:
                     break
 
+        # A provider that returns similar tracks best-match-first (declared via
+        # its `ordered_similarity` property, e.g. a sonic-similarity plugin)
+        # makes `dynamic_tracks` meaningfully ordered: preserve that order (its
+        # picks on top, top-up below) instead of the default shuffled mix.
+        if any(
```
MarvinSchenkel, line 193: Also here, let's make this part of the "Allow plugins to also deliver similar tracks"-PR. That cleans this PR from any non-plugin related code and we can address them in parallel.

NeptuneHub:
I respect the work you are doing. I maintain a project, so I understand that every project has its own rules and preferences.

That said, I don't want to keep spending time splitting and reworking this PR. I have already put a fair amount of effort into adapting it, and I don't plan to continue changing it.

This PR:

- Adds support for AudioMuse-AI, the first opensource sonic analysis, that support multiple of your Music Server (Jellyfin, Navidrome, recently also Plex);
- Fixes a small bug in the plugin priority.

If you would prefer to implement these changes in a different way, that is completely fine. In that case, please feel free to close this PR and implement the solution in the way that best fits the project.

morgancurrie:
The people want to know -- what's happening here? We're dying to get audiomuse support in music assistant!!

MarvinSchenkel:
@NeptuneHub Totally understand where you are coming from. We are just discussing how we can best tackle the challenge of implementing 'priority' or 'preference' for picking **how** things like 'similar tracks' should be resolved MA-wide.

Once we have that figured out, we'll loop that into this PR and get it merged.

NeptuneHub:
@MarvinSchenkel this PR has _"Allow edits and access to secrets by maintainers"_ enabled. 
As soon as you decide how to proceed feel free to to commit any needed change against the PR branch to proceed with the merge.

Thanks for your times.

Delivator:
Really hope this will make it's way in an upcoming release. AudiomuseAI is such a key part for me when listening to my local music library!

PR #4581, opened 2026-07-04
Add Google Drive filesystem provider

OzGav:

Adds a new Google Drive music provider. This uses the same library as HA.

The provider-agnostic plumbing lives in a new shared `filesystem_cloud` package so an upcoming OneDrive provider (separate PR) can reuse it without changes.

Users must create a Google OAuth client (ID + secret). The process is described in the HA docs and our docs will reference that with minor clarifications (such as using our callback). Multiple instances are supported and the instance is named after the chosen folder. The OAuth scope is read only.

Under the hood it follows the same recipe as the WebDAV provider, subclass `LocalFileSystemProvider` and only replace the parts that touch storage. Items keep their human-readable relative path as their identity, so everything path-derived that the local provider already does (folder artwork, .lrc lyrics, cue sheets, audiobook chapters) works unchanged on Drive.

Two Drive-specific problems needed solving:

- Drive doesn't have paths, it has opaque file IDs. Paths are resolved through a per-folder listing cache which the sync walk fills as a side effect so the sibling-file probes the inherited logic makes (artwork, lyrics, playlists) normally cost no extra API calls.
- Drive access tokens expire after ~1 hour, which would kill long audiobook streams. Audio is therefore served through a dynamic route on the MA webserver that proxies the Drive download and attaches a freshly refreshed token on every request. Range headers are forwarded so seeking works, and tag parsing / embedded cover art extraction (ffmpeg) read over the same URL.

Sync concurrency is capped at 4 (same as WebDAV).

`CloudFileSystemProvider` holds all the generic logic above.  A concrete cloud provider implements just three hooks against its API, list a folder's children, download a file, open a streaming download plus its own auth/setup. The hooks translate client-library errors into MA errors (`ProviderUnavailableError` for API/transport failures, `LoginFailed` for auth) so the base can handle failures uniformly.

- related backlog item https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=23192235

That item recently was updated to say look at pyfilesystem.org but that appears to now be parked. I looked into the underlying Google Drive provider that it used to point to and it appears to me that using the same library as HA (which I did here) is the better solution.

## Types of changes

## Checklist


music_assistant/providers/filesystem_cloud/base.py
```diff
@@ -0,0 +1,356 @@
...
+        Proxy a cloud download through MA, adding a fresh auth header.
+
+        Because this runs per request, the token is always valid - so even a
+        multi-hour audiobook can't outlive it.
+        """
+        path = request.query.get("path")
+        if not path:
+            raise web.HTTPBadRequest(text="Missing path")
+        try:
+            file_id = await self._resolve_id(path)
```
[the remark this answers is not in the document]
OzGav: Fair.

music_assistant/providers/filesystem_google_drive/provider.py
```diff
@@ -0,0 +1,167 @@
...
+        # name here, but Drive only accepts the opaque ID from the folder URL
+        if self.root_folder_id != "root":
+            try:
+                meta = await self._get_file_meta(self.root_folder_id, "id, name, mimeType")
+            except GoogleDriveApiError as err:
+                msg = (
+                    f"Drive folder ID '{self.root_folder_id}' not found. Use the ID from the "
+                    "folder URL (drive.google.com/drive/folders/<ID>), not the folder name."
+                )
+                raise SetupFailedError(msg) from err
```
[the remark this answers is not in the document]
OzGav: This can never run. The library call inside _get_file_meta already catches every aiohttp network error itself and re-raises it as its own exception type (GoogleDriveApiError), which the existing except already handles

music_assistant/providers/filesystem_google_drive/auth.py
```diff
@@ -0,0 +1,119 @@
...
+            "client_secret": self._client_secret,
+            "refresh_token": self._refresh_token,
+            "grant_type": "refresh_token",
+        }
+        async with self.mass.http_session.post(OAUTH_TOKEN_URL, data=data) as resp:
+            if resp.status in (400, 401):
+                # invalid_grant and friends: the refresh token was revoked or expired
+                raise LoginFailed(f"Google token refresh failed: {await resp.text()}")
+            resp.raise_for_status()
+            payload = await resp.json()
```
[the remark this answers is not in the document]
OzGav: This actually would get caught by the library but it probably is neater to catch it here

music_assistant/providers/filesystem_google_drive/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+  "domain": "filesystem_google_drive",
+  "name": "Google Drive",
+  "description": "Stream music, audiobooks and podcasts stored in Google Drive",
+  "documentation": "https://music-assistant.io/music-providers/google-drive/",
+  "type": "music",
+  "requirements": ["python-google-drive-api==0.1.0"],
+  "codeowners": ["@ozgav"],
+  "multi_instance": true,
+  "stage": "experimental"
```
OzGav:
```suggestion
  "stage": "beta"
```

music_assistant/providers/filesystem_cloud/base.py
```diff
@@ -0,0 +1,356 @@
...
+            return entry[1]
+        raise MediaNotFoundError(f"Cloud path not found: {file_path}")
+
+    async def exists(self, file_path: str) -> bool:
+        """Check if a cloud file/folder exists."""
+        if not file_path:
+            return False
+        try:
+            return await self._lookup(self._normalize_path(file_path)) is not None
+        except ProviderUnavailableError, MediaNotFoundError:
```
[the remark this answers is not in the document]
OzGav: No this is valid Python 3.14 syntax

music_assistant/providers/filesystem_cloud/base.py
```diff
@@ -0,0 +1,377 @@
...
+
+    # ------------------------------------------------------------------
+    # streaming
+    # ------------------------------------------------------------------
+
+    def _stream_url(self, path: str) -> str:
+        """Build the MA-hosted URL that proxies this cloud file."""
+        base = f"{self.mass.streams.base_url}/{self.instance_id}_stream"
+        return f"{base}?path={quote(path)}"
+
```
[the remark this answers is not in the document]
OzGav: OK this is confirmed real. I reproduced it by syncing audiobooks from an actual Google Drive account: every file failed with Cloud path not found: http:... because the chapter scan passed the URL-derived parent_path into `_scandir`.

The root cause is older than this PR though. `FileSystemItem.parent_path` derives from `absolute_path`, but `_scandir` (and `exists/_read_file`) take relative paths. Most call sites already use `relative_parent_path`, and the handful of audiobook/podcast ones that didn't only worked on local providers because `os.path.join` with an absolute second argument happens to be a no-op. The WebDAV provider already had to add URL-to-relative conversion workarounds for the same reason.

Fixed by making relative paths the single currency: the remaining call sites now use `relative_parent_path`, `parent_name` is derived from the relative path, and the now-unused `parent_path` property is removed (it had no other consumers). Behaviour is unchanged for local/SMB/NFS/WebDAV, and audiobook sync from Google Drive now works and was verified before/after with a multi-file audiobook. Added a regression test covering parent-folder scans on cloud items.

music_assistant/providers/filesystem_cloud/base.py
```diff
@@ -0,0 +1,377 @@
...
+    async def _handle_stream_request(self, request: web.Request) -> web.StreamResponse:
```
MarvinSchenkel, line 288: Let's also add a `HTTPNotFound` to this method in case we are **not** requesting a file with an audio extension. Otherwise people on the same network can just download files via the streamserver.

OzGav: Good idea. Done.

music_assistant/providers/filesystem_cloud/base.py
```diff
@@ -0,0 +1,390 @@
...
+
+# (id, name, is_dir, checksum, size) as returned by _api_list_children
+RawItem = tuple[str, str, bool, str, int | None]
+
+# extensions the stream route will serve; playlists/cue/images are read
+# server-side and never fetched over HTTP, so audio is all it needs to proxy
+AUDIO_STREAM_EXTENSIONS = TRACK_EXTENSIONS | AUDIOBOOK_EXTENSIONS | PODCAST_EPISODE_EXTENSIONS
+
+
+class CloudFileSystemProvider(LocalFileSystemProvider):
```
MarvinSchenkel, line 53: [CRITICAL] Playlist track resolution is broken on cloud providers: the inherited `_parse_playlist_line` (filesystem_local `__init__.py`) builds its first candidate via `Path(self.get_absolute_path(playlist_path)) / line).resolve()`, which joins onto `base_path` (= the opaque folder ID, a relative string) and resolves against the server process CWD. Cloud `resolve()` then raises `MediaNotFoundError`, but the candidate loop only does `contextlib.suppress(FileNotFoundError)`, so the exception escapes to the outer `except MusicAssistantError` and no further candidate is tried - including root-relative lines that would have resolved. Every m3u/pls entry fails.

Reproduced against this PR's own `_StubCloudProvider` (tree contains `Artist/track.mp3`):

```python
>>> await provider._parse_playlist_line("track.mp3", "Artist")
result: None
warning: Could not parse track.mp3 to track: Cloud path not found: Users
```

Note the mangled `Users/...` path - the CWD leaked in and resolution stopped there.

WebDAV hit this exact issue and fixed it by overriding `_parse_playlist_line` (webdav/provider.py), joining relative to the playlist folder with `posixpath.normpath` and suppressing `(FileNotFoundError, MediaNotFoundError)`. `CloudFileSystemProvider` needs the same override (or better: make the local implementation relative-path based, like this PR already did for `parent_path`, and let webdav drop its copy). Please also add a regression test - the suite covers artwork/lyrics/audiobook siblings but not playlist entries.

music_assistant/providers/filesystem_google_drive/auth.py
```diff
@@ -0,0 +1,122 @@
...
+            "grant_type": "refresh_token",
+        }
+        try:
+            async with self.mass.http_session.post(OAUTH_TOKEN_URL, data=data) as resp:
+                if resp.status in (400, 401):
+                    # invalid_grant and friends: the refresh token was revoked or expired
+                    raise LoginFailed(f"Google token refresh failed: {await resp.text()}")
+                resp.raise_for_status()
+                payload = await resp.json()
+        except ClientError as err:
```
MarvinSchenkel, line 118: [PROBLEM] `_refresh` maps every failure to `LoginFailed`: `resp.raise_for_status()` raises `ClientResponseError` (a `ClientError` subclass) for 5xx, and network blips also land in the `except ClientError` branch. Drive tokens expire hourly, so a long sync will refresh mid-run; one transient Google 503 then surfaces as "authentication failed" and aborts the sync as an auth error (the sync walker only skips-and-continues on `ProviderUnavailableError`). Keep `LoginFailed` for the 400/401 branch and raise `ProviderUnavailableError` for other statuses/transport errors.

music_assistant/providers/filesystem_cloud/base.py
OzGav: Good catch, confirmed. The inherited code built a path with the server's working directory baked in, which can never exist on a cloud provider. The resulting error also wasn't suppressed, so the other candidate paths were never tried.

I went with your "or better" option, since this PR already made the parent folder logic relative. The shared `_parse_playlist_line` in `filesystem_local` now works with relative paths: it first tries the line relative to the playlist folder (using `posixpath.normpath`), then the raw line, and suppresses both `FileNotFoundError` and `MediaNotFoundError`. The WebDAV override is deleted and cloud providers just inherit the fixed version. Absolute playlist lines still work through the raw line fallback because `resolve()` handles joining to the base path.

Added regression tests for playlist entries against the stub cloud provider (relative, root-relative, and unresolvable lines). Also tested manually on Drive with a playlist mixing line styles, and on the local provider with a playlist containing an absolute path to check nothing regressed there.

music_assistant/providers/filesystem_google_drive/auth.py
OzGav: Agreed. A temporary Google outage during a token refresh shouldn't be reported as an auth failure and kill the sync. `_refresh` now raises `LoginFailed` only for 400/401 responses (token revoked or expired) and ProviderUnavailableError for anything else, including network errors. The sync walker already handles that by logging and continuing.

MarvinSchenkel (approved):
Thanks Gav, nicely on time for the beta :)

PR #4582, opened 2026-07-04
Emby Music Provider: add last played date

hatharry:


OzGav:
@hatharry speedy resolution as always. Can you please make sure you don't bypass our PR template in the future.

OzGav (approved):

PR #4589, opened 2026-07-05
Add OpenSubsonic Internet Radio to the MusicAssistant Subsonic Provider

Drizzt321:

Implements the OpenSubsonic Radio type into the MusicAssistant Radio. Small and focused, does have an enable/disable Radio, just like for Podcast.

https://opensubsonic.netlify.app/categories/internet-radio/

No related issue

## Types of changes

## Checklist

- [x] `pre-commit run --all-files` passes. - see note below
https://github.com/music-assistant/music-assistant.io/pull/761

### NOTE: 
The `pre-commit run --all-files` does fail on `plex/__init__.py:1742`, but this is a failure that exists on `dev` branch


Drizzt321:
Added documentation update PR https://github.com/music-assistant/music-assistant.io/pull/761

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -671,6 +666,21 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
                 item.id,
                 item.content_type,
             )
+        elif media_type == MediaType.RADIO:
+            # An internet radio station is an unbounded live stream: emit a
+            # direct HTTP stream to its streamUrl, no seeking, no proxying.
+            # ffmpeg autodetects the container, so content type is unknown.
+            station = await self._find_radio_station(item_id)
```
OzGav: Why are you doing this instead of just grabbing from the library?

music_assistant/providers/opensubsonic/parsers.py
```diff
@@ -285,6 +292,43 @@ def parse_artist(
     return artist
 
 
+def parse_radio(instance_id: str, sonic_radio: SonicInternetRadioStation) -> Radio:
+    """
+    Parse an InternetRadioStation into a Music Assistant Radio.
+
+    The station's streamUrl is carried in the single ProviderMapping's `details`
```
OzGav: General comment throughout. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

OzGav (changes requested):
The PR title doesn't make sense. @khers can you also review this one

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: Ah, yes, missed this. Was concentrating on hitting/focusing on the subsonic-server setup, this has been corrected.

music_assistant/providers/opensubsonic/parsers.py
Drizzt321: Ah, got it. In my own Javadocs sometimes I tend to leak that for the docs, since callers for my work sometimes need to know specifics. Fixed up all the docstrings that are from me, although I noticed a few pre-existing ones that might fall under this, but I left them alone.

music_assistant/providers/opensubsonic/__init__.py
```diff
@@ -97,6 +99,12 @@ async def get_config_entries(
             required=True,
             default_value=True,
         ),
+        ConfigEntry(
+            key=CONF_ENABLE_RADIOS,
+            type=ConfigEntryType.BOOLEAN,
+            required=True,
+            default_value=True,
+        ),
```
OzGav: This should be removed. Core already has a toggle to control the sync of radio items.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -524,6 +503,22 @@ async def get_library_podcasts(self) -> AsyncGenerator[Podcast]:
             for channel in channels:
                 yield parse_podcast(self.instance_id, channel)
 
+    async def get_library_radios(self) -> AsyncGenerator[Radio]:
+        """Yield the internet radio stations in the user's library."""
+        # Keep this enable_radios gate: when off we yield nothing, which is what
+        # drives MA's sync to purge already-synced stations (orphan removal).
+        # Looks redundant with library_sync_radios, but isn't.
```
OzGav: You will need to explain this. This option is causing the problem you describe. Under normal circumstances, at sync, if the stations are still on the server then nothing will get removed from the MA database.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -875,62 +945,59 @@ async def _played_recommendations(self) -> RecommendationFolder:
...
+        library_item = await self.mass.music.radio.get_library_item_by_prov_id(
+            prov_radio_id, self.instance_id
+        )
+        if library_item is not None:
+            for mapping in library_item.provider_mappings:
+                if mapping.provider_instance == self.instance_id and mapping.details:
+                    return prov_radio_id, mapping.details
+        # Fallback: station not in the library (e.g. streamed without a prior sync).
+        station = await self._find_radio_station(prov_radio_id)
+        return station.id, station.stream_url
```
OzGav: How could this happen when all stations are getting sync'ed to the MA library?

tests/providers/opensubsonic/test_parsers.py
```diff
@@ -67,6 +73,28 @@ async def test_parse_artists(example: pathlib.Path, snapshot: SnapshotAssertion)
...
+
+    radio = parse_radio("xx-instance-id-xx", station)
+    parsed = radio.to_dict()
+    parsed["external_ids"].sort()
+    assert snapshot == parsed
+
+    # Load-bearing invariant: the stream URL must round-trip into the single
+    # provider mapping's `details` field — that is what get_stream_details reads
+    # to build the StreamType.HTTP path. A snapshot alone wouldn't flag a silent
+    # move of this field, so assert it explicitly.
```
OzGav: Overly wordy.

tests/providers/opensubsonic/test_provider.py
```diff
@@ -0,0 +1,294 @@
...
+def test_radio_config_key_wired() -> None:
+    """
+    Pin the CONF_ENABLE_RADIOS key string contract.
+
+    This key is the contract between the config UI (the __init__.py ConfigEntry)
+    and what handle_async_init reads. The toggle tests force-set _enable_radios
+    and so can't catch a wrong key string; this pins it directly, mirroring
+    CONF_ENABLE_PODCASTS == "enable_podcasts".
+    """
+    assert CONF_ENABLE_RADIOS == "enable_radios"
```
OzGav: How is this value adding?

OzGav (changes requested):

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -74,9 +76,16 @@
     from libopensonic.media import Playlist as SonicPlaylist
     from libopensonic.media import PodcastEpisode as SonicEpisode
 
+    # InternetRadioStation is not re-exported from libopensonic.media top-level
+    # in py-opensonic 10.0.0 (unlike its siblings) - import from the submodule.
+    from libopensonic.media.media_types import (
+        InternetRadioStation as SonicInternetRadioStation,
```
khers, line 81: THis is the first consumer which is likely why it was never re-exported. This should be fixed in that library.

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -172,38 +183,6 @@ def is_streaming_provider(self) -> bool:
         """
         return False
 
-    async def _get_podcast_episode(self, eid: str) -> SonicEpisode:
```
khers, line 175: This is a really unfortunate diff. It looks like you are moving code around in the file, why?

music_assistant/providers/opensubsonic/sonic_provider.py
khers: seconded, this comment does not make sense with the code as written

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -875,62 +945,59 @@ async def _played_recommendations(self) -> RecommendationFolder:
...
-        except DataNotFoundError:
-            self.logger.debug("Lyrics not found for '%s'", track.id)
-            return None
-        if not lyrics:
-            return None
-        return parse_structured_lyrics(lyrics[0])
+    async def _resolve_radio_stream(self, prov_radio_id: str) -> tuple[str, str]:
+        """Resolve a station's id and stream URL, raising MediaNotFoundError if unknown."""
+        # Prefer the URL stored in the library item's mapping details (set at sync
+        # time) over re-fetching the whole station list; fall back if not synced.
```
khers: There is no way that we could get here if the item exists on the subsonic sever but not in MA. If you have a scenario where you saw that behavior it is likely a bug and we should fix it.

tests/providers/opensubsonic/test_parsers.py
khers: I appreciate that your LLM added comments, but they should be written in a way that lets non-native speakers easily understand what is happening. I agree that the extra tests need a comment because they look odd. But this comment does not convince me that we need them instead of simply removing the extra tests.

tests/providers/opensubsonic/test_provider.py
```diff
@@ -0,0 +1,294 @@
...
+    """Return an OpenSonicProvider with radio enabled (the default case)."""
+    return _make_provider(enable_radios=True)
+
+
+# --- feature flag -----------------------------------------------------------
+
+
+def test_radio_feature_declared() -> None:
+    """LIBRARY_RADIOS must be advertised so MA surfaces the Radios browse node."""
+    assert ProviderFeature.LIBRARY_RADIOS in SUPPORTED_FEATURES
```
khers: I don't understand why this test is necessary. It simply confirms that a developer added item still exists in a set. Testing set membership is not a valuable thing to do on its face. If the set is runtime constructed and needs to have certain items, then that may be worth testing, but this test simply confirms that line 48 in the provider is still there.

tests/providers/opensubsonic/test_provider.py
```diff
@@ -0,0 +1,294 @@
...
+    assert sd.allow_seek is False
+    assert sd.can_seek is False
+    # the fallback fetched + filtered the list (no singular getter):
+    provider.conn.get_internet_radio_stations.assert_awaited_once()
+
+
+async def test_get_stream_details_radio_library_hit_without_details_falls_back(
+    provider: OpenSonicProvider,
+) -> None:
+    """A synced item whose mapping has no stored URL falls through to the station list."""
```
khers: This test is unnecessary, we cannot have a radio station item with a URL (the py-opensonic library will enforce its presence by refusing to construct a response without one. We should not silently handle this and instead surface the error that caused it.

khers (changes requested):
I don't mind adding support for open subsonic internet radio (never added it because I don't use it), but this PR needs work.

1. Make sure that the corner cases you are handling are necessary and not bugs elsewhere (URL-less stations for instance)
2. Don't move code around without a really strong reason. And if you have a strong reason, do it in a separate commit (and possibly PR) because it made the diff here unreadable.
3. Comments are great but make sure they are worded in a concise manner that is understandable by a non-native speaker.

Drizzt321:
@OzGav @khers changed to draft, I'll be doing testing things via the MA Dev app, and GH automatically loads those into the PR when I push to my GH, I'll respond to all comments and here when I've gone through everything.

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: This is to match the current AGENTS.md `Private methods should be at the bottom of the file, public at the top`, same cleanup was done in PRs [4365](https://github.com/music-assistant/server/pull/4365) and [4418](https://github.com/music-assistant/server/pull/4418). This file was grandfathered in `scripts/lint_baselines/private_methods_not_last.txt`, and since the PR touches it anyway I brought it into compliance and removed the baseline entry; no logic changed, tests pass. If you'd rather keep this PR's diff minimal, happy to revert the reordering + restore the baseline entry and do the compliance pass as a separate PR, your call.

music_assistant/providers/opensubsonic/__init__.py
Drizzt321: Removed, I was just mirroring the existing Enable Podcast toggle.

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: Removed along with the Radio flag, not relevant anymore.

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: Removed, a station not in the library no raises MediaNotFoundError at stream time.

How I hit it was that I have another service which pushes via MA API using a token in order to push in a radio URI that does exist on the subsonic-server, but which hasn't been synced via the Subsonic Radio Sync task yet. I've changed my client to add an item to the library with the URI, and then add that to the queue.

tests/providers/opensubsonic/test_parsers.py
Drizzt321: Removed along with the extra asserts, yes, wasn't needed.

tests/providers/opensubsonic/test_provider.py
Drizzt321: Removed along with the Radio Enable toggle removal.

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: Yes, removed, now raises a MediaNotFoundError at stream time instead of re-fetching.

tests/providers/opensubsonic/test_provider.py
Drizzt321: Removed

tests/providers/opensubsonic/test_provider.py
Drizzt321: Test is gone, and the silent fall-through. Now raises a MediaNotFoundError for missing row or missing stored URL instead of quietly doing a re-fetch of the station list.

music_assistant/providers/opensubsonic/sonic_provider.py
Drizzt321: You're right, but the upstream py-opensonic doesn't re-export InternetRadioStation in [`__init__.py`](https://github.com/khers/py-opensonic/blob/v10.1.0/src/libopensonic/media/__init__.py), only the `media.media_types` sub-module. Since this is the first time it's been used, probably slipped through the cracks. Want me to go PR upstream to fix that, and then we can fix this here?

Drizzt321:
@khers @OzGav thanks for the review, I've replied inline to all your comments. One point still open I'd like to highlight, the method reordering in `music_assistant/providers/opensubsonic/sonic_provider.py`, this was to comply with the most recent AGENTS.md and linting rules.

OzGav (commented):
Looks way better from my perspective. Eric will respond is due course with his thoughts. I expect you should pursue PR-ing the upstream library. Also tests are failing.

music_assistant/providers/opensubsonic/sonic_provider.py
khers: Yes

khers: I think you will find the maintainer fairly easy to work with ;)

music_assistant/providers/opensubsonic/sonic_provider.py
```diff
@@ -172,38 +179,6 @@ def is_streaming_provider(self) -> bool:
         """
         return False
 
-    async def _get_podcast_episode(self, eid: str) -> SonicEpisode:
```
khers, line 175: Why did you move these functions? Please keep changes to the code you need.

khers (commented):
It is not clear why the code reorganization in the sonic_provider.py file was necessary, please but it back and only add what you need.

MarvinSchenkel:

OzGav:
Sorry this has now been superseded by https://github.com/music-assistant/server/pull/5150

PR #4592, opened 2026-07-05
Add username parameter to the search, search_library and library_items methods of music controller

fmunkes:

Add username to global search and library_items method of media controllers. Will allow me to add username support to the HA integration for music_assistant.get_library & music_assistant.search.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music/controller.py
+++ b/music_assistant/controllers/music/controller.py
@@ -286,6 +286,7 @@ class MusicController(MusicDatabaseSetupMixin, CoreController):
         media_types: list[MediaType] = MediaType.ALL,
         limit: int = 25,
         library_only: bool = False,
+        username: str | None = None,
     ) -> SearchResults:
         """
         Perform global search for media items on all providers.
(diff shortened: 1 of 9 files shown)


marcelveldt:
While looking at this PR I suddenly realized something;

- Impersonating of a user depends on access rights. Currently only an (api) user with admin rights can actually impersonate as somebody else. This is good as it is like this for security reasons.

- The current HA user/token we auto generate for the directly connected HA instance (MA running as App/add-on on HA) is NOT an admin user so this will fail by default for all add-on users.

- Instead of adding the username argument to all methods manually, we could consider adding a "allow_impersonate" argument to the api_command handler which then auto injects and accepts the username/user field on methods annotated with this (including setting the contextvar etc.)

I will think about this for a tiny bit (especially the HA token usecase).

fmunkes:
Ok, sure, take your time!

marcelveldt:
Closing this one - superseded by #4613 which centralizes impersonation in the api command dispatch. music/search and all music/*/library_items now accept an optional "user" argument (user_id or username, "username" also works as alias), handled centrally so no per-method wrappers are needed anymore. So for the HA integration you can simply pass user/username to these commands. Note that the HA system user now has a service role which is allowed to impersonate, so this works out of the box for add-on users. Thanks for the groundwork here - it directly shaped the final design!

PR #4594, opened 2026-07-05
Implement SQL-based explicit filter for smart playlists

dmoo500:

Implements SQL-based filtering for the explicit content filter in smart playlists, replacing the inefficient POST-filter that operated on small samples.

**⚠️ This PR depends on #4597** - must be merged after #4597.

**Problem:** Smart playlists with "explicit only" filter showed only 1-3 tracks instead of expected 83 because the POST-filter was applied to a small 75-track random sample (6.3% explicit rate = ~5 explicit tracks, further reduced by recency).

**Solution:** 
- Uses the new `explicit` parameter from `TracksController.library_items()` (added in PR #4597)
- Smart playlist provider passes `explicit=rules.explicit` through all code paths
- Ensures accurate filtering before random sampling
- POST-filter removed in library mode, kept in seed mode

**Filter logic:**
- **Allow explicit** (`None`): No filter, shows all tracks
- **Explicit only** (`True`): SQL filter for explicit=1
- **No explicit** (`False`): SQL filter for explicit IS NULL OR = 0

- Depends on: PR #4597

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/smart_playlist/__init__.py
+++ b/music_assistant/providers/smart_playlist/__init__.py
@@ -727,8 +727,7 @@ class SmartPlaylistProvider(PluginProvider):
                     and t.metadata.popularity >= rules.min_popularity
                 ]
 
-            # Apply explicit filter in library mode
-            tracks = _filter_by_explicit(tracks, rules.explicit)
+            # Explicit filter is now handled at SQL level via _get_library_tracks()
 
             if rules.year_from is not None or rules.year_to is not None:
                 tracks = [
@@ -1101,6 +1100,7 @@ class SmartPlaylistProvider(PluginProvider):
             return await self._get_library_tracks(
                 favorite=None,
                 genre_ids=None,
+                explicit=rules.explicit,
                 limit=min(rules.limit * 3, 2000),
                 user_provider_filter=user_provider_filter,
             )
@@ -1110,6 +1110,7 @@ class SmartPlaylistProvider(PluginProvider):
         base_tracks = await self._get_library_tracks(
             favorite=favorite,
             genre_ids=genre_ids,
+            explicit=rules.explicit,
             limit=min(rules.limit * 5, 2000),
             user_provider_filter=user_provider_filter,
         )
@@ -1152,7 +1153,10 @@ class SmartPlaylistProvider(PluginProvider):
 
         if rules.favorites_only:
             for track in await self._get_library_tracks(
-                favorite=True, limit=fetch_limit, user_provider_filter=user_provider_filter
+                favorite=True,
+                explicit=rules.explicit,
+                limit=fetch_limit,
+                user_provider_filter=user_provider_filter,
             ):
                 if track.uri:
                     track_sets[track.uri] = track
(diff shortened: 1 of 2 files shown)


music_assistant/providers/smart_playlist/__init__.py
```diff
@@ -1210,16 +1219,32 @@ async def _get_library_tracks(
...
         user_provider_filter: list[str] | None = None,
     ) -> list[Track]:
         """Fetch library tracks with optional filters."""
+        extra_query_parts = []
+
+        # Build explicit filter SQL (no core changes needed!)
+        if explicit is not None:
+            if explicit:
+                # Explicit only: only tracks with explicit=true
+                extra_query_parts.append("json_extract(tracks.metadata, '$.explicit') = 1")
```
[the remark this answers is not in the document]
dmoo500: Tests added

music_assistant/controllers/music/media/tracks.py
```diff
@@ -197,9 +197,15 @@ async def library_items(
...
-        extra_query_parts: list[str] = []
-        extra_join_parts: list[str] = []
+        # Extract any extra query parts from kwargs (for extensibility)
+        extra_query_params: dict[str, Any] = kwargs.get("extra_query_params", {})
+        extra_query_parts: list[str] = (
+            kwargs.get("extra_query_parts", []).copy() if kwargs.get("extra_query_parts") else []
+        )
+        extra_join_parts: list[str] = (
+            kwargs.get("extra_join_parts", []).copy() if kwargs.get("extra_join_parts") else []
+        )
```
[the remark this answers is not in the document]
dmoo500: Intentionally not implemented. The kwargs dict is not mutated by this code - we only extract the explicit value via .get(). The extra_query_parts list is created fresh in this function, not extracted from kwargs, so there's no mutation of caller data. Defensive copying would add unnecessary overhead without providing any safety benefit here.

dmoo500:
To make sure my latest changes are correct – could someone request a copilot review?

music_assistant/controllers/music/media/tracks.py
```diff
@@ -196,10 +196,26 @@ async def library_items(
         :param order_by: Order by field (e.g. 'sort_name', 'timestamp_added').
         :param provider: Filter by provider instance ID (single string or list).
         :param genre: Filter by genre id(s).
+        :param played_only: Filter to only played tracks.
+        :param kwargs: Additional filters (e.g. explicit: bool | None).
```
marcelveldt: let's not do this as this is [a] too unspecific and [b] may open the door for sql injection some day
this is the public method, also available on the api.

so maybe just instead either;

a) simply add an explicit filter to the tracks.library_items (true/false/none)
b) from smart playlists, use get_library items by query manually

dmoo500: The reason I chose this approach is that I didn't want to have to adjust all the providers around this parameter again, but yes - it makes more sense, especially if it's exposed via API.

dmoo500: Ah... because the explicit default is None, I don't need to adjust it at all providers - Sorry, I totally overlooked that.

dmoo500: PR #4597 added for the core changes

marcelveldt (approved):
Thanks, @dmoo500 !

PR #4620, opened 2026-07-06
Fix Home Assistant integration losing authentication after restart or over time

MarvinSchenkel:

The Home Assistant integration repeatedly loses authentication to the Music Assistant server when running as the HAOS add-on, failing setup with `Authentication failed, addon discovery not completed yet` (or a reauth prompt). Two variants are reported: it breaks **on every restart** of the MA container (a reload fixes it) and it breaks **over time** (after ~90 days).

Root cause is in `get_homeassistant_system_user_token()`: it deleted and re-minted the token on every announce (so each MA restart invalidated the token the integration still held), and the token silently died at the 90-day absolute lifetime cap with no rotation path (the integration cannot reauth while running as an add-on).

This keeps the short-lived, expiring token but makes the announce idempotent and rotates the token proactively:

- The announced token is persisted (in the auth DB, next to the `jwt_secret` that can mint any token, so no added attack surface) and reused as-is on subsequent announces. Re-announcing the same payload is a no-op: the Supervisor dedupes it and HA only reloads the entry when the token actually changed.
- A replacement token is only minted when the current one is missing, expired or revoked, or within 7 days of its 90-day absolute cap. The superseded token stays valid until it expires on its own, so HA keeps working until it reloads with the new one.
- The add-on now re-announces daily, so rotation also happens on long uptimes and a missed announce (e.g. HA-side discovery flow aborted during startup) self-heals.
- Expired token rows for the HA system user are cleaned up on rotation.

- Fixes https://github.com/home-assistant/core/issues/171938
- Fixes https://github.com/home-assistant/core/issues/158174

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/webserver/auth.py
+++ b/music_assistant/controllers/webserver/auth.py
@@ -63,8 +63,7 @@ TOKEN_LONG_LIVED_EXPIRATION = 365  # Long-lived tokens (1 year, no auto-renewal)
 # Max days a sliding short-lived session may live from creation before re-auth.
 TOKEN_ABSOLUTE_MAX_EXPIRATION = 90
 TOKEN_GUEST_EXPIRATION = 1  # Guest sessions: short fixed lifetime, no renewal
-# Days before the absolute cap at which the HA integration token is rotated, so the
-# replacement is announced (and picked up by HA) before the current one stops working.
+# Days before the absolute cap at which the HA integration token is rotated
 HA_TOKEN_ROTATION_MARGIN = 7
 
 HA_TOKEN_SETTING_KEY = "ha_integration_token"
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/webserver/auth.py
```diff
@@ -410,35 +410,39 @@ async def get_homeassistant_system_user(self) -> User:
...
-        The token auto-renews on use (expires after 30 days of inactivity).
+        The addon calls this on every startup to (re)announce the token to the HA
+        integration, which cannot reauth while running as an addon. Re-minting on each
+        announce invalidated the token the integration still held (breaking setup on every
+        restart, issue #158174) and a short-lived token silently expired over time
+        (issue #171938). A fresh long-lived token is therefore only minted when the
+        integration does not already have a valid one.
 
-        :return: Authentication token for the Home Assistant system user.
+        :return: A newly minted token to announce, or None when a valid token already
```
marcelveldt: I don't think this is valid to assume ? You have no idea if the HA integration actually consumed the token.
What if a race happens with MA + HA booting on the same host ?

marcelveldt (approved):

PR #4627, opened 2026-07-06
Preserve player config when replacing registered players

dpersek:

Load the persisted full player config when `register_or_update()` replaces an existing player object.

This reuses the normal registered-player setup path so replacement state is published after config load and protocol links are refreshed consistently. Same-instance updates stay on the lightweight path, while true replacements are serialized to avoid overlapping setup hooks.

This PR also adds regression coverage for replacement config preservation, overlapping replacements, same-instance updates, and a Squeezelite/universal-player replacement shape found while investigating support#5745.

This should be treated as a related in-memory replacement lifecycle fix only. It does not change the code path that writes ordinary player settings to `settings.json`; those writes go through `save_player_config()` / `ConfigController.set()` and are not modified here. It should not be treated as proven closure for the on-disk settings rewrite reported in support#5745.

Validation:

- `PYTHONDONTWRITEBYTECODE=1 .venv/bin/pytest -p no:cacheprovider --no-cov tests/controllers/players/test_player_controller.py::TestRegisterOrUpdate -q` -> 4 passed.
- `PYTHONDONTWRITEBYTECODE=1 .venv/bin/pytest -p no:cacheprovider --no-cov tests/controllers/players/test_player_controller.py tests/controllers/players/test_player_protocol_state.py tests/controllers/players/test_protocol_linking.py -q` -> 209 passed.
- `.venv/bin/pre-commit run --all-files` -> passed.

Scope / non-goals:

- No provider discovery, Squeezelite protocol, playback, queue, frontend, migration behavior, or `settings.json` write path changed.
- No hardware/container recreation smoke test was run; proof is controller-level regression coverage plus adjacent player/protocol tests.
- No claim that this closes the on-disk settings rewrite mechanism reported in support#5745.
- No real music-service account, media library, playback device, Home Assistant state, LAN discovery, or user credentials were used.

- related issue https://github.com/music-assistant/support/issues/5745

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -1593,24 +1593,7 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
             # We use the 'initialized' attribute to indicate that the player
             # is still in the process of being registered so we can filter it out where needed.
             self._players[player_id] = player
-            # update state to ensure player.state reflects the final attributes
-            # (e.g. player type) set after super().__init__() in the player subclass,
-            # before we fetch config (which relies on state.type for entry resolution)
-            player.update_state(signal_event=False)
-            # ensure we fetch and set the latest/full config for the player
-            player_config = await self.mass.config.get_player_config(player_id)
-            player.set_config(player_config)
-            # update state again now that config is loaded
-            player.update_state(signal_event=False)
-            self._save_underlying_player_id(player)
-            # call hook after the player is registered and config is set
-            await player.on_config_updated()
-
-            # Handle protocol linking
-            self._evaluate_protocol_links(player)
-
-            # now we're ready to signal the player is added and available
-            player.set_initialized()
+            await self._setup_registered_player(player, signal_state_event=False)
             self.logger.info(
                 "Player (type %s) registered: %s/%s",
                 player.state.type.value,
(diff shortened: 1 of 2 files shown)


OzGav:
Whilst I can see that this solves A problem, what evidence do you have that is solves the linked issue?

dpersek:
Fair question. I added a follow-up regression test, `test_squeezelite_universal_player_replacement_preserves_reported_settings`, that is shaped around the linked support report: a Squeezelite-backed `universal_player` replacement starts with the provider/default name, while the persisted full `PlayerConfig` has a custom display name, `hide_in_ui=True`, announcement min/max overrides, and `play_media_overrides_group=False`.

That test proves this branch reloads the full persisted `PlayerConfig` before the replacement publishes state or refreshes protocol links. On the previous replacement path, the replacement object could publish state from its default/base config; on this branch the custom name and those persisted settings remain applied.

So the evidence is now issue-shaped controller regression coverage plus the adjacent player/protocol test slice. I still have not run a full Docker container recreation with real Squeezelite hardware, so I’m not claiming that level of proof.

OzGav:
The test confirms the in-memory fix, which I wasn't disputing. My question was about the mechanism: the issue shows settings.json being rewritten on disk. Which code path fixed here writes those values to storage? If there isn't one, this PR fixes a related bug but we shouldn't close #5745 with it

dpersek:
You're right. I re-checked the storage paths and this PR does not add or change the code path that writes those reported settings back to `settings.json`.

The path fixed here is the in-memory replacement path: `register_or_update()` now reloads the full persisted `PlayerConfig` before the replacement publishes state or refreshes protocol links. The storage writes for player config go through `save_player_config()` / `ConfigController.set()` -> `settings.json`, and this diff does not change a writer for the custom name, `hide_in_ui`, or announcement settings shown in support#5745.

So I agree this should be treated as a related controller bug, not as proven closure for support#5745. I updated the PR body to remove the `Fixes` wording and narrow the claim.

OzGav:
This and your other PR are bypassing our PR template. Please don't do that. Also note https://github.com/music-assistant/.github/blob/main/AI_POLICY.md

dpersek:
Understood, updated both PRs to be in accordance with your PR template.

marcelveldt:
Closing this one in favor of #4634, which addresses the actual cause of the settings loss (see my comment there): the wipe happens in the universal player restore path due to a corrupted persisted `player_type`, not in the in-memory replace path this PR touches - so this change can't fix #5745. The replacement hardening itself isn't wrong (re-hydrating the full config and setting the initialized flag on a replaced player object are real gaps), so feel free to resubmit that separately with an accurate description if you think it's still worth it.

PR #4632, opened 2026-07-06
Speed up library sync when nothing has changed

marcelveldt:
## Problem

Every library sync re-checked each item by loading the complete media item from the database — metadata, artists, album, artwork and all — even when nothing changed since the last sync. For a 10k-track library that meant ~30k queries and ~110k object constructions of pure waste, recurring every few hours via the scheduled syncs (and considerably slower on a Raspberry Pi than on the machine measured below).

## Changes

- Library sync now uses a new lightweight lookup (`get_library_item_sync_details`) that fetches only the few fields needed to detect changes — favorite flag, date added and the raw provider mappings — instead of constructing full media items. Tracks and audiobooks fetch their few extra decision fields (album link, author/narrator type, resume state).
- The full load/update path only runs when an item is actually new or changed. Behavior is otherwise identical: favorites propagation, `date_added` updates, album backfill, unavailable-track skip and the deletion pass are all preserved.
- Genre sync resolves incoming names against a short-lived cached genre lookup and skips all writes when the item is already mapped to the same genres.
- The provider-mapping lookup now filters on `media_type`, so it uses the unique index instead of scanning the table.
- New regression test: syncing twice with the test provider asserts the second sync performs zero library writes and constructs zero full `Track`/`Album`/`Artist`/`Audiobook` objects.

## Measured (10k tracks / 500 albums / 50 artists, test provider, M-series Mac)

| | dev | this PR |
|---|---|---|
| no-change re-sync CPU | 11.2 s | **3.9 s** |
| SQL execute time within re-sync | 5.9 s | 0.2 s |
| full media item constructions during the check | ~110k | 0 |

Initial sync also drops ~14% (36.9 s → 31.7 s) from the cheaper existence checks.


tests/integration/test_no_change_resync.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+        async with wait_for_sync_completion(mass):
+            await mass.music.start_sync()
+        # sync tasks are done; drain until fully idle while spies are still active
+        elapsed = 0.0
+        while mass.music.active_sync_tasks and elapsed < 60.0:
+            await asyncio.sleep(0.25)
+            elapsed += 0.25
+
```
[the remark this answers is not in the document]
marcelveldt: The genre scan is raw SQL only (no MediaItem hydration, no controller writes) so nothing could actually slip past the spies here. Tightened the test anyway: the spies now stay active until the scan has fully completed.

music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -554,3 +557,53 @@ async def _authors_narrators(self, column: str) -> UniqueList[str]:
...
+                JOIN artists ON artists.item_id = audiobook_artists.artist_id
+                WHERE audiobook_artists.audiobook_id = audiobooks.item_id
+                AND artists.artist_type = '{ArtistType.NARRATOR.value}'
+            ) AS has_narrator_artists
+            , json_type(audiobooks.authors, '$[0]') AS first_author_type
+            , json_type(audiobooks.narrators, '$[0]') AS first_narrator_type
+            , playlog.fully_played AS fully_played
+            , playlog.seconds_played * 1000 AS resume_position_ms
+        """
+        extra_joins = (
```
MarvinSchenkel, line 597: [PROBLEM] This playlog join can fan out: without a session user (the normal case for a scheduled sync) rows for *all* users match, and `get_library_item_sync_details` runs with `limit=1` and no `ORDER BY`, so SQLite returns an arbitrary resume-state row. The `base_query` above guards against exactly this with a correlated subquery (`ORDER BY p2.timestamp DESC LIMIT 1` — "so the join can never fan out the result set"). In a multi-user install with divergent resume positions this either triggers a spurious `update_item_in_library` on every sync (defeating the optimization) or skips a genuine resume-state change, depending on which row wins. Suggest mirroring the base_query join:

```sql
LEFT JOIN playlog ON playlog.id = (
    SELECT p2.id FROM playlog p2
    WHERE p2.item_id = audiobooks.item_id AND p2.media_type = 'audiobook'
    [AND p2.userid = :playlog_userid]
    ORDER BY p2.timestamp DESC LIMIT 1)
```

music_assistant/controllers/music/media/genres.py
```diff
@@ -1048,6 +1048,25 @@ async def sync_media_item_genres(
...
+            if (normalized := self._normalize_genre_name(name))
+        }
+        stored_rows = await self.mass.music.database.get_rows_from_query(
+            f"SELECT DISTINCT alias FROM {gm} "
+            "WHERE media_type = :media_type AND media_id = :media_id",
+            {"media_type": media_type.value, "media_id": media_id_int},
+            limit=0,
+        )
+        stored_aliases = {row["alias"] for row in stored_rows}
+        if None not in stored_aliases and stored_aliases == incoming_aliases:
```
MarvinSchenkel: [PROBLEM] This short-circuit assumes alias→genre resolution is static, but it changes over time: a user can create a new genre whose name/alias matches an already-stored alias, and the daily scan grows aliases via `_ensure_aliases`. Previously the next sync re-resolved every alias and `to_add` created the new mapping; now the sync returns early because the alias *strings* are unchanged, so the mapping is never created via sync. The compensating `_bulk_scan_unmapped_genres` only resolves names from items' `metadata.genres` JSON — names delivered via `get_item_genre_names` aren't guaranteed to be there, so those items can be permanently skipped. Consider invalidating the short-circuit when the genre/alias set changed since the mappings were written (e.g. compare against `max(timestamp)` of the genres table captured per sync run).

MarvinSchenkel (commented):
Reviewed — CI is green and the earlier Copilot comment is addressed, but found two issues in the change-detection logic that should be looked at before merging.

music_assistant/controllers/music/media/audiobooks.py
marcelveldt: Good catch! This predates #4629 landing on dev (was mirroring the old base_query join). Merged dev in and applied the same correlated subquery here, so the lookup now always picks the single most recent playlog row.

music_assistant/controllers/music/media/genres.py
marcelveldt: Fair point. Reworked it: the fast path now performs the actual name->genre resolution (primary name first, then aliases, excluded skipped) from a short-lived cached snapshot (5s TTL) instead of comparing stored alias strings. Unknown names still take the full path, and a newly created genre that claims a stored alias now remaps on the next sync - added a regression test for exactly that scenario.

MarvinSchenkel (approved):

PR #4634, opened 2026-07-07
Preserve universal player settings during startup restore

### LINKED ISSUE #5745: Squeezelite player settings/configuration rewritten on startup

2.9.4

Docker Container

### The problem

For quite a few versions now, player configuration in squeezelite players (which are the only players I use) is being overwritten on startup, when the container is recreated. 

I can verify that the settings are correct in the UI and in the settings.json, and when MASS restarts, I can watch in realtime in the UI as players that should be hidden in the UI are added back. Comparing the settings.json post boot shows anywhere from a couple to a dozen players have had their settings restored to default, depending on the boot.

I've attached settings.json of the players before and after a restart, plus a file showing the differences.

[settings.txt](https://github.com/user-attachments/files/29419132/settings.txt)
[settings-after.json](https://github.com/user-attachments/files/29419112/settings-after.json)
[settings-before.json](https://github.com/user-attachments/files/29419111/settings-before.json)

### How to reproduce

Edit default player settings in the settings page - name, hide in ui, min/max announcement volume, and play override are the ones I use.

Restart music assistant docker container, with --force-rebuild (I thought I had noticed the problem on regular docker restarts, but when capturing logs for this at least on this version it is only occurring on container recreates).

Compare custom player settings, many (not all) are changed back to default settings. It can vary from a couple players to a dozen or so.

### Music Providers

N/A

### Player Providers

Squeezelite

### Full log output

[docker-logs-startup.redacted.txt](https://github.com/user-attachments/files/29419248/docker-logs-startup.redacted.txt)

### Additional information

Docker compose file:
services:
  music-assistant-server:
    image: ghcr.io/music-assistant/server:latest # <<< Desired release version here (or use beta to get the latest beta version)
    container_name: music-assistant-server
    restart: unless-stopped
    # Network mode must be set to host for MA to work correctly
    network_mode: host
    volumes:
      - /home/dakota/musicassistant/data:/data/
      - /mnt/dildakicon:/media/
    # privileged caps (and security-opt) needed to mount smb folders within the container
    cap_add:
      - SYS_ADMIN
      - DAC_READ_SEARCH
    security_opt:
      - apparmor:unconfined
    environment:
      # Provide logging level as environment variable.
      # default=info, possible=(critical, error, warning, info, debug)
      - LOG_LEVEL=debug
    ulimits:
      nofile:
        soft: 65536
        hard: 65536

N/A

Home Assistant Container

Linux

- [x] Yes

dpersek:

Preserve restored Universal Player configs when a linked protocol child was temporarily rewritten as a regular player during startup.

The issue's startup log shows restored Universal Players being deleted because their linked Squeezelite child rows briefly had `player_type: player`, followed by the same Universal Player IDs being recreated with default names/settings. The restore path now treats a persisted `protocol_parent_id` pointing at the Universal Player as the stronger signal that the child still belongs to that wrapper, so the existing Universal Player config is kept instead of deleted and recreated.

This also keeps `config/players` summary reads from mutating raw player config data that can later be saved to `settings.json`.

Fixes music-assistant/support#5745.

Validation:

- Added a regression that reproduces the destructive restore decision for `upe45f0170ef67` / `e4:5f:01:70:ef:67` and verifies the Universal Player config is not removed.
- Added a persistence regression proving a `config/players` summary read followed by a save leaves the raw `settings.json` player row unchanged.
- Replayed the issue attachment shape locally by taking the reported `settings-before.json`, changing the four affected Squeezelite child rows to the transient `player_type: player` state shown in the startup log, then running restore + save; the four affected Universal Player rows remained unchanged.
- `.venv/bin/pytest --no-cov -q tests/controllers/players/test_protocol_linking.py::TestUniversalPlayerRestoreOrphanCleanup tests/controllers/config/test_persistent_storage_save.py` -> passed.
- `.venv/bin/pytest --no-cov -q tests/controllers/players/test_protocol_linking.py` -> passed.
- `.venv/bin/pre-commit run --all-files` -> passed.

Scope:

- Does not change protocol child registration or type detection.
- Does not change stale Universal Player cleanup when a child has no parent link or has moved to another parent.

- related issue https://github.com/music-assistant/support/issues/5745

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/config/players.py
+++ b/music_assistant/controllers/config/players.py
@@ -501,9 +501,14 @@ class PlayerConfigMixin:
             # update default name if needed
             if name and name != existing_conf.get("default_name"):
                 self.set(f"{CONF_PLAYERS}/{player_id}/default_name", name)
-            # update player_type if needed
-            if existing_conf.get("player_type") != player_type:
-                self.set(f"{CONF_PLAYERS}/{player_id}/player_type", player_type.value)
+            # NOTE: player_type is deliberately NOT updated here. This is called from
+            # Player.__init__ where the type can still be a transient class default
+            # (the real type is often only determined during provider setup).
+            # Overwriting the persisted type here corrupts the stored config if
+            # registration is interrupted (e.g. shutdown while clients reconnect),
+            # which in turn caused universal player configs to be deleted on the
+            # next startup. Genuine player type changes are persisted by
+            # update_state once the player is fully registered.
             return
         # config does not yet exist, create a default one
         conf_key = f"{CONF_PLAYERS}/{player_id}"
--- a/music_assistant/controllers/config/players.py
+++ b/music_assistant/controllers/config/players.py
@@ -501,14 +501,9 @@ class PlayerConfigMixin:
             # update default name if needed
             if name and name != existing_conf.get("default_name"):
                 self.set(f"{CONF_PLAYERS}/{player_id}/default_name", name)
-            # NOTE: player_type is deliberately NOT updated here. This is called from
-            # Player.__init__ where the type can still be a transient class default
-            # (the real type is often only determined during provider setup).
-            # Overwriting the persisted type here corrupts the stored config if
-            # registration is interrupted (e.g. shutdown while clients reconnect),
-            # which in turn caused universal player configs to be deleted on the
-            # next startup. Genuine player type changes are persisted by
-            # update_state once the player is fully registered.
+            # deliberately do NOT update player_type here: this is called from
+            # Player.__init__ where the type can still be a transient class default.
+            # Genuine type changes are persisted by update_state after registration.
             return
         # config does not yet exist, create a default one
         conf_key = f"{CONF_PLAYERS}/{player_id}"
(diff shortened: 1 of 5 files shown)


marcelveldt:
Thanks - this one targets the actual failure path. I traced the startup log from the issue report myself and the wipe is indeed the universal player restore deleting configs because the squeezelite child rows briefly had `player_type: player` persisted.

The root cause is that `Player.__init__` calls `create_default_player_config` with the constructor-time player type, which for squeezelite is still the class default (`player`) - the real type (`protocol`) is only set later during setup. So every client (re)connect flipped the persisted row and healed it moments later via `update_state`. If registration got interrupted in between (shutdown while clients reconnect, which is exactly what happens on a container recreate), the corrupted row was flushed to disk and the next boot deleted the universal player config.

I pushed a commit to this branch that fixes that root cause: `create_default_player_config` no longer rewrites `player_type` on existing configs - genuine type changes are already persisted by `update_state` after registration. Your restore guard remains useful as defense-in-depth and to heal configs that are already corrupted in the wild, and the deepcopy fix for the summary reads is a nice catch as well.

marcelveldt (approved):
Thanks, @SyntaxSawdust !

PR #4636, opened 2026-07-07
Raise transient provider errors instead of caching them as negatives

marcelveldt:
## Problem

Follow-up to #4616. There, the metadata/lyrics/art providers whose `None` return conflated "genuinely not found" with "transient fetch error" were opted out of negative-result caching (`cache_none=False`) as the safe stopgap — so their negatives (no lyrics, no cover art, no bio) are still re-fetched from the network on every access.

The correct fix is to stop conflating the two at the source: a transient error should propagate (and never be cached), while a genuine "not found" returns `None` (and is cached).

## Changes

- **lrclib / coverartarchive / itunes_artwork / wikipedia**: these now raise/propagate on transient failures (network, 5xx, unparsable response) and return `None` only for a genuine miss (404/204, no result, no cover art). Their `cache_none=False` opt-out is dropped, so genuine negatives are cached again.
- **Metadata enrichment**: the artist/album/track loops now wrap each provider call in `try/except` (log + continue), mirroring the existing playlist path, so a provider raising on a transient error degrades gracefully instead of aborting enrichment of the item. The lyrics path was already guarded.
- theaudiodb / fanart.tv / acoustid keep `cache_none=False` — their `None` only ever signals an error (a genuine miss is a data object that is already cached), so there is nothing to gain from the refactor there.

## Tests

- Enrichment resilience: a provider that raises must not abort artist/album/track enrichment for the remaining providers.
- lrclib: a 404 returns `None` (cached negative); a network error propagates instead of being cached.

### DIFF
--- a/music_assistant/providers/coverartarchive/__init__.py
+++ b/music_assistant/providers/coverartarchive/__init__.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-import aiohttp.client_exceptions
 from music_assistant_models.enums import ExternalID, ImageType, ProviderFeature
 from music_assistant_models.media_items import Album, MediaItemImage, MediaItemMetadata, UniqueList
 
@@ -88,8 +87,7 @@ class CoverArtArchiveMetadataProvider(MetadataProvider):
             )
         )
 
-    # None can signal a network failure as well as "no cover art", so don't cache it
-    @use_cache(86400 * 30, cache_none=False)
+    @use_cache(86400 * 30)
     async def _get_cover_art_url(self, release_group_id: str) -> str | None:
         """
         Retrieve cover art URL for a release group.
@@ -99,14 +97,12 @@ class CoverArtArchiveMetadataProvider(MetadataProvider):
         # Try 1200px first, fall back to 500px
         for size in ("front-1200", "front-500"):
             url = f"{CAA_BASE_URL}/release-group/{release_group_id}/{size}"
-            try:
-                async with self.mass.http_session.head(url, allow_redirects=True) as response:
-                    if response.status == 200:
-                        return str(response.url)
-            except (
-                aiohttp.client_exceptions.ClientConnectorError,
-                aiohttp.client_exceptions.ServerDisconnectedError,
-                TimeoutError,
-            ):
-                self.logger.debug("Failed to retrieve cover art from %s", url)
+            async with self.mass.http_session.head(url, allow_redirects=True) as response:
+                if response.status == 200:
+                    return str(response.url)
+                if response.status != 404:
+                    # a non-404 status (5xx, 429, ...) is a transient error — propagate it
+                    # instead of caching it as "no cover art"
+                    response.raise_for_status()
+        # a 404 for both sizes means this release group genuinely has no cover art
         return None
--- a/music_assistant/providers/coverartarchive/__init__.py
+++ b/music_assistant/providers/coverartarchive/__init__.py
@@ -4,7 +4,9 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+import aiohttp
 from music_assistant_models.enums import ExternalID, ImageType, ProviderFeature
+from music_assistant_models.errors import ResourceTemporarilyUnavailable
 from music_assistant_models.media_items import Album, MediaItemImage, MediaItemMetadata, UniqueList
 
 from music_assistant.controllers.cache import use_cache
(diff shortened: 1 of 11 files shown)


music_assistant/providers/lrclib/__init__.py
```diff
@@ -152,24 +150,17 @@ async def get_track_metadata(self, track: Track) -> MediaItemMetadata | None:
...
-            self.logger.debug("Error parsing response from LRCLib API: %s", err)
-            return None
+        async with self.mass.http_session.get(
+            f"{self.api_url}/get", params=params, headers=headers
+        ) as response:
+            # 204/404 mean there are genuinely no lyrics for this track; any other failure
+            # (5xx, network, malformed json) propagates so it is not cached as "no lyrics"
+            if response.status in (204, 404):
+                return None
+            response.raise_for_status()
```
MarvinSchenkel: [PROBLEM] `get_track_lyrics` (metadata/controller.py:328) loops over lyrics providers with no error handling, so a transient LRCLIB failure (5xx/network, or `RetriesExhausted` — `@throttle_with_retries` only retries `ResourceTemporarilyUnavailable`, not `ClientResponseError`) now aborts the loop, skipping remaining lyrics providers such as genius_lyrics, and propagates out of the API command instead of the previous graceful "no lyrics". Suggest wrapping the per-provider call in `get_track_lyrics` in the same try/except-and-continue pattern this PR added in enrichment.py.

music_assistant/providers/coverartarchive/__init__.py
```diff
@@ -99,14 +97,12 @@ async def _get_cover_art_url(self, release_group_id: str) -> str | None:
...
-                TimeoutError,
-            ):
-                self.logger.debug("Failed to retrieve cover art from %s", url)
+            async with self.mass.http_session.head(url, allow_redirects=True) as response:
+                if response.status == 200:
+                    return str(response.url)
+                if response.status != 404:
+                    # a non-404 status (5xx, 429, ...) is a transient error — propagate it
+                    # instead of caching it as "no cover art"
+                    response.raise_for_status()
```
MarvinSchenkel: [PROBLEM] The radio artwork path can't handle the newly raised exceptions: the provider loops in metadata/radio.py catch only `(ProviderUnavailableError, ResourceTemporarilyUnavailable, InvalidDataError)` and `update_radio_stream_artwork` catches only `MusicAssistantError`, so a raw `ClientResponseError` from coverartarchive/itunes_artwork/wikipedia now skips all remaining providers in the loop and escapes into the `mass.call_later` background task (streams/audio.py) and bbc_sounds as an unhandled task exception. Either broaden the exception handling in radio.py (log-and-continue, mirroring enrichment.py) or translate transient failures into `ResourceTemporarilyUnavailable`, which existing callers already catch and `throttle_with_retries` retries.

MarvinSchenkel (commented):
Reviewed. Note: both existing Copilot comments are incorrect against the current diff — `_get_cover_art_url` raises via `raise_for_status()` for any non-200/404 (only a double-404 caches `None`, covered by the 503 test), and wikipedia's `_get_json` has no `raise_on_error` parameter at all; transient failures propagate unconditionally. However, two call sites of these providers can't handle the newly raised exception types — see inline comments.

music_assistant/providers/coverartarchive/__init__.py
```diff
@@ -88,25 +87,18 @@ async def get_album_metadata(self, album: Album) -> MediaItemMetadata | None:
...
-                        return str(response.url)
-            except (
-                aiohttp.client_exceptions.ClientConnectorError,
-                aiohttp.client_exceptions.ServerDisconnectedError,
-                TimeoutError,
-            ):
-                self.logger.debug("Failed to retrieve cover art from %s", url)
+            async with self.mass.http_session.head(url, allow_redirects=True) as response:
+                if response.status == 200:
+                    return str(response.url)
```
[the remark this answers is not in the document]
marcelveldt: Good catch — fixed. A non-404 now raises ResourceTemporarilyUnavailable instead of being cached as no-cover-art.

music_assistant/providers/wikipedia/__init__.py
```diff
@@ -176,9 +176,17 @@ async def _fetch_bio(self, lang: str, title: str) -> str | None:
         return None
 
     async def _get_json(
-        self, url: str, params: dict[str, str] | None = None
+        self, url: str, params: dict[str, str] | None = None, *, raise_on_error: bool = False
     ) -> dict[str, Any] | None:
```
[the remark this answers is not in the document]
marcelveldt: Fixed — _get_json no longer swallows; a transient failure now raises ResourceTemporarilyUnavailable (only a real 404 returns None), so _wikidata_sitelinks won't cache {} on error either.

music_assistant/providers/lrclib/__init__.py
marcelveldt: Good catch, you're right — that loop wasn't guarded. Wrapped the per-provider call in get_track_lyrics with try/except-and-continue like the enrichment loops, so a failing provider is skipped instead of aborting the lookup.

music_assistant/providers/coverartarchive/__init__.py
marcelveldt: Went with your ResourceTemporarilyUnavailable suggestion — the providers now raise that on transient errors, so radio.py's existing catches (and throttle_with_retries) handle them with no changes needed there. Thanks!

MarvinSchenkel (approved):

PR #4642, opened 2026-07-07
Reduce image proxy-id overhead when serializing media items

marcelveldt:
Serializing media items injects a `proxy_id` for every image occurrence in every outbound message: a single 500-track listing performs ~1,500 `compute_image_id` calls, each doing a fresh SHA-256 + lock + LRU shuffle. On top of that, every LRU miss unconditionally rewrote the id mapping to the cache db — so after each restart, browsing replayed one sqlite write per unique image, and libraries larger than the 10k LRU cap kept repeating those writes forever.

Changes:
- Add a forward `(provider, path) → id` memo so repeat serializations are a single lock-free dict read (~7.5x faster warm path and no more lock contention between the event loop and executor threads).
- Skip re-persisting mappings that are already stored: an in-process marker plus a cheap expiration probe (new `CacheController.get_expiration` helper) turn the post-restart write storm into reads. A stored row is only rewritten once it has burned through half its 1-year TTL, so long-lived ids stay resolvable across restarts.

### DIFF
--- a/music_assistant/controllers/cache/controller.py
+++ b/music_assistant/controllers/cache/controller.py
@@ -205,6 +205,28 @@ class CacheController(CoreController):
                 return data, is_fresh, True
         return None, False, False
 
+    async def get_expiration(
+        self,
+        key: str,
+        provider: str = "default",
+        category: int = 0,
+    ) -> int | None:
+        """
+        Return the expiration timestamp (epoch seconds) of a cache entry, if any.
+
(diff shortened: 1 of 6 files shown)


MarvinSchenkel (approved):
Reviewed: forward-memo fast path, eviction, persist-marker dedup on failure, half-TTL re-persist logic, and the expires-only probe in get_expiration all check out. The earlier Copilot comment was already addressed by the follow-up commit (probe only the expires column + key validation). CI green — approving and merging.

PR #4646, opened 2026-07-07
Advertise both IP families via mDNS and respect a specific bind IP when publishing

OzGav:

The `_mass._tcp` zeroconf advertisement only ever contained the single auto-detected (usually IPv4) publish address, even when the host also has routable IPv6 addresses or when the webserver was explicitly bound to one specific (IPv6) address. IPv6-capable clients were therefore always pointed at the IPv4 address, and binding the server to a specific address advertised an address the server does not even listen on.

- discovery: advertise the primary address of both IP formats when bound to all interfaces, and only the bound address when bound to one specific address
- webserver: when bound to one specific address, use that address as the publish address (and for the auto base_url) instead of the auto-detected primary IP

I don't think this will break any existing behaviour. Testing was limited to confirming no regression in an IPv4 network as I dont have exclusively IPv6

- related issue https://github.com/music-assistant/support/issues/4977

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/discovery/controller.py
+++ b/music_assistant/controllers/discovery/controller.py
@@ -27,9 +27,8 @@ from music_assistant.constants import (
     CONF_ZEROCONF_INTERFACES,
     INGRESS_SERVER_PORT,
     VERBOSE_LOG_LEVEL,
-    WILDCARD_BIND_IPS,
 )
-from music_assistant.helpers.util import get_ip_addresses, get_ip_pton, get_zeroconf_args
+from music_assistant.helpers.util import get_ip_pton, get_zeroconf_args
 from music_assistant.models.core_controller import CoreController
 
 if TYPE_CHECKING:
@@ -263,7 +262,7 @@ class DiscoveryController(CoreController):
             zeroconf_type,
             name=f"{server_id}.{zeroconf_type}",
             addresses=[
-                await get_ip_pton(address) for address in await self._get_publish_addresses()
+                await get_ip_pton(address) for address in self.mass.webserver.publish_addresses
             ],
             port=self.mass.webserver.publish_port,
             properties=self.mass.get_server_info().to_dict(),
@@ -279,21 +278,6 @@ class DiscoveryController(CoreController):
             self.logger.error(
                 "Music Assistant instance with identical name present in the local network!"
             )
-
-    async def _get_publish_addresses(self) -> list[str]:
-        """Return the IP addresses to include in the Music Assistant mDNS advertisement."""
-        webserver = self.mass.webserver
-        addresses = [webserver.publish_ip]
-        if webserver.bind_ip and webserver.bind_ip not in WILDCARD_BIND_IPS:
-            return addresses
-        # bound to all interfaces: also advertise the primary address of the other
-        # IP family (if any) so both IPv4-only and IPv6-only clients can connect
-        publish_is_ipv6 = ":" in webserver.publish_ip
-        for ip in await get_ip_addresses(include_ipv6=True):
-            if (":" in ip) != publish_is_ipv6:
-                addresses.append(ip)
-                break
-        return addresses
 
     def _on_mdns_service_state_change(
         self,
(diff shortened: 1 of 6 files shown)


music_assistant/controllers/discovery/controller.py
```diff
@@ -261,7 +262,9 @@ async def _register_mass_service(self) -> None:
         info = AsyncServiceInfo(
             zeroconf_type,
             name=f"{server_id}.{zeroconf_type}",
-            addresses=[await get_ip_pton(self.mass.webserver.publish_ip)],
+            addresses=[
+                await get_ip_pton(address) for address in await self._get_publish_addresses()
+            ],
```
marcelveldt, line 266: this is weird. It should read from the webserver - not invent the wheel itself

music_assistant/controllers/streams/controller.py
```diff
@@ -344,12 +345,15 @@ async def setup(self, config: CoreConfig) -> None:
...
+        self._bind_ip = bind_ip = str(config.get_value(CONF_BIND_IP))
         publish_ip = str(config.get_value(CONF_PUBLISH_IP) or CONF_VALUE_AUTO)
         if publish_ip == CONF_VALUE_AUTO:
-            # resolve the "auto" default (or an unset value) to this server's primary IP
-            publish_ip = (await get_ip_addresses(include_ipv6=True))[0]
+            if bind_ip not in WILDCARD_BIND_IPS:
+                publish_ip = bind_ip
+            else:
+                # resolve the "auto" default (or an unset value) to this server's primary IP
+                publish_ip = (await get_ip_addresses(include_ipv6=True))[0]
```
marcelveldt: Aren't we conflating the bind ip and the publish IP here ?
If we want to change this, it should be the DEFAULT bind + publish IP having this logic, not secretly overwrite it

music_assistant/controllers/discovery/controller.py
```diff
@@ -277,6 +280,21 @@ async def _register_mass_service(self) -> None:
...
+            return addresses
+        # bound to all interfaces: also advertise the primary address of the other
+        # IP family (if any) so both IPv4-only and IPv6-only clients can connect
+        publish_is_ipv6 = ":" in webserver.publish_ip
+        for ip in await get_ip_addresses(include_ipv6=True):
+            if (":" in ip) != publish_is_ipv6:
+                addresses.append(ip)
+                break
+        return addresses
+
```
marcelveldt: We need to trust the webserver to have a property/method to provide us the actual/configured bind IP + publish address

music_assistant/controllers/discovery/controller.py
OzGav: Yes. I’ve moved this into the webserver so it now resolves its publish addresses at setup and exposes them as publish_addresses, and discovery just reads that. The address selection logic and its tests live with the webserver now.

music_assistant/controllers/discovery/controller.py
OzGav: Same resolution as your first point.

music_assistant/controllers/streams/controller.py
OzGav: You’re right that it reads that way, though to clarify it only changed what “auto” resolved to, an explicitly configured publish IP was never touched. But I’ve dropped the streams change from this PR anyway to keep it scoped to the mDNS advertisement fix. Happy to raise the auto-publish-vs-bind question separately if you think it’s worth discussing.

tests/controllers/webserver/test_publish_addresses.py
```diff
@@ -0,0 +1,33 @@
...
+        "fd00::10",
+    ]
+    assert _get_publish_addresses(None, "192.168.1.10", ALL_ADDRESSES) == [
+        "192.168.1.10",
+        "fd00::10",
+    ]
+    assert _get_publish_addresses("::", "fd00::10", ALL_ADDRESSES) == [
+        "fd00::10",
+        "192.168.1.10",
+    ]
```
marcelveldt, line 22: do we also want a IPv6 only test here that tests a ipv6 IP (or wildcard) and a IPv6 base url ?

OzGav: Yeah thats a good idea. Ill have something up shortly.

marcelveldt (approved):
LGTM, thanks @OzGav !

PR #4650, opened 2026-07-07
Fix stale active output protocol on sync group leader after group stop

MarvinSchenkel:

When a sync group is stopped, the leader's `active_output_protocol` was only cleared when the leader no longer reported PLAYING. Real devices (e.g. Wiim via AirPlay) keep reporting PLAYING for a few seconds after the stop command, so the clear was skipped and the stale protocol persisted. The next *solo* playback on that player then wrongly used the group's protocol (AirPlay) instead of the preferred native protocol.

- Replace the racy conditional clear in `_dissolve_syncgroup` with an unconditional delayed clear (same 5s `call_later` pattern the player controller uses); the timer is cancelled automatically when a new session sets the protocol.
- Add a unit test verifying the dissolve schedules the delayed protocol clear even while the leader still reports PLAYING.

- reported on Discord (Wiim Amp + Sonos sync group: solo playback stuck on AirPlay after group stop)

## Types of changes

## Checklist


marcelveldt (approved):

music_assistant/providers/sync_group/player.py
```diff
@@ -729,12 +729,15 @@ async def _dissolve_syncgroup(self) -> None:
...
-        if sync_leader and sync_leader.state.playback_state != PlaybackState.PLAYING:
-            sync_leader.set_active_output_protocol(None)
+        # Delayed-clear the leader's active protocol: the leader may still report
+        # PLAYING briefly after stop; a new session cancels the timer.
+        if sync_leader:
+            self.mass.call_later(
+                5,
+                sync_leader.set_active_output_protocol,
+                None,
+                task_id=f"clear_active_protocol_{sync_leader.player_id}",
```
marcelveldt: what if playback is started again in this timeframe, will the delayed clear be cancelled ? otherwise it gets overwritten.

also, this is weird for a user that you have to wait 5 seconds before issuing a playback command.
instead we should wait for the player to be in the state we expect and then clear the protocol

marcelveldt (changes requested):
see my inline comment

music_assistant/providers/sync_group/player.py
MarvinSchenkel: Both points addressed:

1. **Cancellation on restart**: yes — `Player.set_active_output_protocol` cancels the `clear_active_protocol_{player_id}` timer as its first action on every call, so any new session (which sets a protocol) automatically cancels the pending clear.

2. **Fixed 5s delay**: replaced with an event-driven clear (`PlayerController.schedule_active_output_protocol_clear`) that polls every 0.5s and clears as soon as the player no longer reports PLAYING, with a 10s deadline as fallback. The solo-stop path in `_handle_cmd_stop` now uses the same helper.

marcelveldt: why not use the existing helper that can wait for a certain state instead of polling ?

marcelveldt: done — switched to the existing `_wait_for_playback_state` helper instead of polling. it clears once the leader reports IDLE (with a 10s timeout as fallback), and the clear is deduped per player so a new session cancels the pending one via `cancel_task`.

marcelveldt (approved):

PR #4654, opened 2026-07-08
Fix several bugs in the Jellyfin provider

OzGav:

Had Claude do a health check on the Jellyfin provider since it is unmaintained.

This is a batch of small fixes:

- Searching for anything with a hyphen in it (like "AC-DC") crashed the album search, which took the whole Jellyfin search down with it.
- Starting playback of a track that was deleted from the server, or that has no duration metadata, crashed instead of failing gracefully.
- Logging in with a passwordless Jellyfin account never worked, because the literal text "None" was sent as the password. An empty password is now sent correctly.
- With two Jellyfin servers configured, albums, artists and playlists could resolve to the wrong server (tracks were already correct). All item types are now tied to the specific server instance they came from.
- The checks that pick which Jellyfin libraries to sync used accidental substring matching instead of exact comparison. Fixed.
- Cleaned up a pile of unused constants and corrected two misleading comments.

No database changes, no config changes, existing installs keep working as-is. All provider tests pass (snapshots updated for the instance-id change).

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/jellyfin/__init__.py
+++ b/music_assistant/providers/jellyfin/__init__.py
@@ -24,7 +24,7 @@ from music_assistant_models.media_items import (
 )
 from music_assistant_models.streamdetails import StreamDetails
 
-from music_assistant.constants import UNKNOWN_ARTIST_ID_MBID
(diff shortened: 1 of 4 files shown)


MarvinSchenkel (approved):
LGTM thanks @OzGav

PR #4685, opened 2026-07-09
Fix flow mode not switching sample rate between tracks on wrapped players

MarvinSchenkel:

Flow mode streams were stuck at the first track's sample rate on wrapped players (e.g. players behind a Universal Group): the mid-flow restart check re-resolved the player from the queue id, which returns the wrapper player. That wrapper has no audio/protocol config entries, so the check fell back to 44.1kHz/16bit defaults and the sample-rate switch never triggered.

- Pass the protocol player (the player actually requesting the flow stream, and the one the flow PCM format was anchored on) into `get_queue_flow_stream()`
- New `_flow_restart_context()` helper resolves the flow mode config and supported sample rates from that player, falling back to the queue's player when not given
- Add unit tests for the helper (protocol player preferred, queue-player fallback, no player at all)

- related issue https://github.com/music-assistant/support/issues/5748

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/streams/audio.py
+++ b/music_assistant/controllers/streams/audio.py
@@ -2059,6 +2059,7 @@ class StreamsAudio:
         start_queue_item: QueueItem,
         pcm_format: AudioFormat,
         session_id: str | None = None,
+        flow_player: Player | None = None,
     ) -> AsyncGenerator[bytes]:
         """
         Get a flow stream of all tracks in the queue as raw PCM audio.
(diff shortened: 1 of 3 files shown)


MarvinSchenkel:
Waiting for user confirmation in the related issue

marcelveldt:
I'm pretty sure this one was already fixed last week or the week before where the active protocol would be resolved to get the active setting

marcelveldt:
Rebased on latest dev - the conflicts were just overlap with the queue session plumbing from #4793, the fix itself is unchanged. I also traced this through the current code and I have to walk back my earlier comment: this was not already fixed. The flow restart check still resolved the player from the queue id, and since the audio settings live on the protocol player's own config, the wrapper indeed falls back to 44.1kHz defaults.

marcelveldt (approved):

PR #4686, opened 2026-07-09
Recompute dynamic playlist metadata when its tracks are refreshed

MarvinSchenkel:

The genres (and other metadata) of dynamic builtin playlists like "Random Album (from Library)" never updated: dynamic playlists are excluded from the periodic metadata batch job and nothing recomputed their metadata when their content regenerated, so the genre chips stayed stale forever.

- Schedule `update_metadata(force_refresh=True)` for a dynamic playlist after a user-initiated refresh of its tracks, so metadata is recomputed from the freshly generated content
- Browse requests, playback refills and refreshes of regular playlists are unaffected
- Add unit tests covering all four cases

- related issue https://github.com/music-assistant/support/issues/3899

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music/media/playlists.py
+++ b/music_assistant/controllers/music/media/playlists.py
@@ -145,10 +145,12 @@ class PlaylistController(MediaControllerBase[Playlist]):
             page += 1
 
         if refresh_dynamic_metadata and library_item is not None:
-            # runs after the loop so it reads the freshly cached track sample
-            self.mass.create_task(
-                self.mass.metadata.update_metadata(library_item, force_refresh=True)
-            )
+            # dynamic playlists are excluded from the periodic metadata batch job, so
+            # recompute here from the freshly cached track sample. clearing last_refresh
+            # requests the recompute the way the add/remove track handlers do, without
+            # relying on force_refresh (whose purpose is to bypass caches).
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/music/media/playlists.py
```diff
@@ -156,6 +144,12 @@ async def tracks(
                 yield track
             page += 1
 
+        if refresh_dynamic_metadata and library_item is not None:
+            # runs after the loop so it reads the freshly cached track sample
+            self.mass.create_task(
+                self.mass.metadata.update_metadata(library_item, force_refresh=True)
```
marcelveldt: we should carefully watch for side effects here as I'm pretty sure "force_refresh" will bypass the cache in the entire path

MarvinSchenkel: You mean it will also re-fetch the playlist tracks for example?

marcelveldt: Pushed a small change for this: instead of force_refresh we just clear last_refresh to trigger the recompute (same pattern the add/remove track handlers already use). So no cache-bypass in play at all - the metadata refresh just reads the already-cached sample, no re-fetch. Resolving.

PR #4734, opened 2026-07-11
MSX Bridge: Party Mode QR on TVs, direct streamserver delivery, playback and CSRF hardening

trudenboy:

Brings the MSX Bridge provider up to its **v1.4.4** release. Areas: Party Mode on the TV (menu page, kiosk overlay, and QR stamped into cover art), a direct-streamserver delivery mode (now the default, hardened for Docker/NAT), a Sendspin bridge so TVs join sample-synchronized groups, kiosk improvements, and a round of playback-robustness / security hardening.

## Party Mode support on the TV

When the **Party plugin** has guest access enabled, the bridge puts the party join QR code on the biggest screen in the room:

- **Web kiosk overlay** — QR + party name + configured caption, shown in both kiosk variants (display-only and display + TV audio). The overlay polls a small status endpoint and refetches the QR image only when the join code rotates (opaque `qr_version`), so there is no periodic image flicker on TV browsers.
- **Native MSX page** — the TV menu gains a **Party** entry opening an MSX content page with the same QR rendered as PNG (older Tizen/webOS engines cannot decode SVG).
- **QR stamped into playback cover art** — the native MSX player cannot render overlays, so during a party the playback background is routed through a compositor endpoint that stamps the join QR into the cover's corner; covers revert automatically when the party ends. Guardrails: only MA-hosted image sources are accepted (no open redirect / SSRF proxy), the cover fetch does not follow redirects, remote streaming-service covers are routed through the MA image proxy first, and any failure degrades to a redirect to the original image.
- **Endpoints** — `GET /api/party` (status; the raw guest join URL is deliberately never exposed, clients only get a relative QR image URL so reverse-proxy setups work), `GET /api/party/qr.svg` / `qr.png` (rendered with `segno`, `Cache-Control: no-store`), and `GET /api/party/qr-cover.png` (compositor).
- **Resilience** — a missing, disabled, or misbehaving Party plugin degrades to "no party" (5 s timeout + short TTL cache) and can never break the MSX menu or kiosk; a TV parked on the QR page keeps its player registered (idle-timeout activity refresh).
- **Event-loop hygiene** (addresses review feedback in this PR): all segno/PIL image work runs in a worker thread via `asyncio.to_thread`, matching `helpers/images.py`; concurrent cache misses for the same cover coalesce into one shared fetch+composite task (shielded from requester cancellation), and the QR render is memoized per join-code rotation.

## Direct streamserver delivery ("MA Streamserver" mode)

A third stream-delivery option next to the per-TV proxy and shared-buffer modes: the TV is redirected to the URL resolved by `streams.resolve_stream_url()` and fetches audio straight from the MA streamserver — the same pattern Chromecast/Sonos/WiiM/Squeezelite use. One less ffmpeg process per TV; the per-player output-codec config and DSP are applied by the core. Guardrails:

- Flow-mode URLs (e.g. crossfade on a non-gapless player) are rejected and served through the local per-track proxy instead, preserving per-track progress on the TV.
- Any resolution failure falls back to the proxy pipeline automatically.
- **Docker/NAT support**: the redirect URL's host is rewritten to the host the TV already uses to reach the provider — the streamserver's advertised IP (e.g. a container IP with `publish_ip` AUTO) is unreachable from the LAN, which previously left the TV in a silent retry loop.
- This mode is now the **default for new installations** (existing configs keep their stored value); the per-TV proxy and shared-buffer modes remain selectable.

## Sendspin bridge: TVs join sample-synchronized groups

Each TV can be registered as a Sendspin client (default-on, can be disabled per provider config), so it joins sample-synchronized playback groups together with any Music Assistant players — not just other MSX TVs. When a synchronized stream starts, the TV automatically opens the web kiosk in Sendspin mode; if the TV's browser cannot run the Sendspin client, playback transfers back to the regular HTTP player automatically and the failure is logged.

## Sendspin web-kiosk client bundled with the provider

The kiosk's synchronized-audio mode previously loaded `sendspin-js` **1.0.0 from a CDN at runtime** — dead on LAN-only TVs and two protocol majors behind the server. The current `@sendspin/sendspin-js` **3.2.0** is now vendored into the provider's static tree (import specifiers rewritten for direct browser loading, with a CI guard that resolves every static/dynamic specifier against the vendored files), and the kiosk is adapted to the 3.x API. Browsers without WebCodecs no longer advertise opus, so the SDK never reaches its non-vendored opus fallback.

## Kiosk improvements

- The center equalizer is a real audio spectrum analyzer driven by the live playback signal (HTTP kiosk mode); the Sendspin kiosk keeps its decorative animation, and browsers without Web Audio fall back to it automatically.
- The kiosk URL accepts four display toggles (`controls`, `party`, `viz`, `lyrics` — on by default, `=0` disables) for dedicated displays; disabled features are skipped entirely, not just hidden. The provider status page gained a Kiosk URL Builder composing these links interactively.
- The party QR overlay is centered at the top of the screen.

## Playback robustness and hardening

- A pending track change could serve the previous track's stream (or time out spuriously): the audio handler now explicitly arms the wait for the newly requested track before enqueueing.
- Concurrent group-stream requests could spawn a duplicate orphaned encoder; shared-stream creation is now serialized.
- WS-position staleness and player idle tracking use the monotonic clock, so an NTP step (common on RTC-less hosts) can no longer corrupt reported progress or mass-unregister active TVs. `elapsed_time_last_updated` stays wall-clock per the core contract.
- Playback-control endpoints accept only GET+POST and reject browser cross-site requests (`Sec-Fetch-Site`), closing a LAN CSRF vector; TVs, same-origin players and automations are unaffected.
- Provider unload no longer raises while closing active TV WebSocket connections (the close handshake mutated the client set mid-iteration).

New dependency: `segno==1.6.6` (pure-Python QR generation, no transitive deps) in the provider manifest and `requirements_all.txt`.

Covered by the provider test suite (203 passed / 1 skipped), including Party tests (status/error paths, no join-URL leakage, QR content types and caching, compositor SSRF/redirect guards, off-event-loop and single-flight regression tests), redirect-mode unit+integration tests with fallback and host-rewrite coverage, Sendspin bridge tests, race reproducers for the stale-media and duplicate-encoder fixes, clock-jump simulations, CSRF tests, and vendored-SDK loadability guards.

- Requested in discussion music-assistant#5631 (Android TV Party)
- Provider-repo feature request: trudenboy/ma-provider-msx-bridge#65

## Types of changes

- [x] Dependencies bump — `dependencies`

## Checklist


music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2051,10 +2211,182 @@ async def _handle_queue(self, request: web.Request) -> web.Response:
...
+                    allow_redirects=False,
+                ) as resp:
+                    if resp.status != 200:
+                        raise ValueError(f"cover fetch returned HTTP {resp.status}")
+                    cover_bytes = await resp.read()
+                import segno  # noqa: PLC0415  # only needed when the Party plugin is used
+
+                qr_buf = io.BytesIO()
+                segno.make(party.join_url, error="m").save(qr_buf, kind="png", scale=8)
+                cached = _stamp_qr_on_cover(cover_bytes, qr_buf.getvalue())
```
MarvinSchenkel: `_stamp_qr_on_cover()` does a synchronous PIL decode + resize + PNG re-encode of a 512px cover, running directly on the event loop on every cache miss. MA's own image code offloads all PIL work to a thread (see the `asyncio.to_thread` calls throughout `music_assistant/helpers/images.py`). Since this same server also streams audio on that loop, a burst of TV requests right after a QR rotation (new cache key) will stall playback for every connected player while each composite runs.

Suggest offloading:
```python
cached = await asyncio.to_thread(_stamp_qr_on_cover, cover_bytes, qr_buf.getvalue())
```
The `segno.make(...).save(...)` calls are lighter but could ride along in the same thread hop.

trudenboy: Fixed. Both the PIL composite and the segno QR generation now run off the event loop via asyncio.to_thread — the handlers delegate to sync helpers (_render_qr / _render_qr_cover), so the QR render rides along in the same thread hop as the composite, matching the helpers/images.py pattern.

MarvinSchenkel:
@trudenboy . Looks good to go, could you resolve the conflict so I can merge it?

trudenboy:
> @trudenboy . Looks good to go, could you resolve the conflict so I can merge it?

@MarvinSchenkel 
Conflict resolved, pr ready to merge

MarvinSchenkel (approved):
Thanks @trudenboy

PR #4746, opened 2026-07-11
Fix album version parsing and album_versions for filesystem_local provider

allmazz:

### Fixes album version parsing in the filesystem_local provider

Fixes a bug where the parsed album version was erased.

### Fixes album_versions endpoint when both albums are from filesystem_local provider

When searching other versions in the local database, the search query must not contain artists because they are not stored in the `search_name` column. Before this fix it always returned empty results.

## Types of changes

## Checklist


OzGav:
Thanks for the PR! I verified both fixes and they check out.

A couple of small points:

Could you add a regression test for the version erasure bug? It was a control flow issue rather than a data issue, and nothing in the current test suite would catch it if that NFO block gets reshuffled later. That's the one thing I'd like to see before merging.

Minor nit, non-blocking: in `_parse_album`, `info.get("title", info.get("name"))` will return `None` when the NFO has an empty `<title>` element, because the present key suppresses the fallback to `<name>`. info.get("title") or info.get("name") would handle that case.

Not for this PR, but `tracks.versions()` has the exact same bug you fixed here for albums: it always searches with `f"{track.artist_str} - {track.name}"`, so track versions from local providers also come back empty. Happy to see that as a follow-up PR if you're up for it.

Nice find on both bugs!

allmazz:
> Thanks for the PR! I verified both fixes and they check out.
> 
> A couple of small points:
> 
> Could you add a regression test for the version erasure bug? It was a control flow issue rather than a data issue, and nothing in the current test suite would catch it if that NFO block gets reshuffled later. That's the one thing I'd like to see before merging.
> 
> Minor nit, non-blocking: in `_parse_album`, `info.get("title", info.get("name"))` will return `None` when the NFO has an empty `<title>` element, because the present key suppresses the fallback to `<name>`. info.get("title") or info.get("name") would handle that case.
> 
> Not for this PR, but `tracks.versions()` has the exact same bug you fixed here for albums: it always searches with `f"{track.artist_str} - {track.name}"`, so track versions from local providers also come back empty. Happy to see that as a follow-up PR if you're up for it.
> 
> Nice find on both bugs!

Thank you for highlighting `name` fallback issue. I've fixed it and added a test, but there are no tests fully covering _parse* methods anyway.

_**Maybe** I will fix it sometime..._

OzGav (approved):
Thanks @allmazz

PR #4791, opened 2026-07-13
Add OneDrive filesystem provider

OzGav:

Adds a new OneDrive music provider. This uses the same library as HA's OneDrive integration (onedrive-personal-sdk).

This builds on the shared `filesystem_cloud` package introduced with the Google Drive provider (#4581)

Users must create a Microsoft OAuth client (application ID + secret) in the Azure portal. Details are in the docs PR. The provider uses the consumer (personal account) endpoints and the OAuth scope is read only (Files.Read.All). The configured folder is a path (e.g. Documents/Music) which is resolved to its Graph item ID once at init; multiple instances are supported and the instance is named after the chosen folder.

Listing and metadata go through the SDK, but downloads go straight to Microsoft Graph: the SDK's download helper forwards neither Range headers (needed for seeking) nor the response headers. Graph 302-redirects downloads to a pre-signed URL, which also means the auth header is correctly dropped on the cross-host hop.

Two OneDrive-specific quirks needed handling:

- Microsoft rotates the refresh token on every redemption and the old one only stays valid briefly. Token refreshes are therefore serialized behind a lock (parallel refreshes during a sync could otherwise invalidate the token) and the newest refresh token is persisted encrypted to config so it survives restarts.
- Well-known Microsoft auth errors are mapped to actionable messages: an expired client secret (AADSTS7000222), copying the Secret ID instead of the secret's Value (AADSTS7000215), and revoked/expired authorization (invalid_grant) each tell the user exactly what to fix instead of dumping raw JSON.

File checksums use Graph's quickXorHash (a stable content hash) with a size fallback for files that lack one.

- related issue https://github.com/orgs/music-assistant/projects/2/views/1?pane=issue&itemId=23192235

## Types of changes

## Checklist


music_assistant/providers/filesystem_onedrive/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    :param values: The (intermediate) raw values for config entries sent with the action.
+    """
+    if values is None:
+        values = {}
+
+    if action == CONF_ACTION_AUTH and values.get("session_id"):
+        client_id = str(values.get(CONF_CLIENT_ID) or "")
+        client_secret = str(values.get(CONF_CLIENT_SECRET) or "")
+        # the frontend masks a previously stored secret; fetch the real value
+        if client_secret == SECURE_STRING_SUBSTITUTE and instance_id:
```
MarvinSchenkel, line 73: [CRITICAL] Reauthorization passes the encrypted stored client secret to Microsoft because `get_raw_provider_config_value` returns the persisted encrypted value; decrypt it with `mass.config.decrypt_string(...)` before calling `authorize()`.

music_assistant/providers/filesystem_onedrive/auth.py
```diff
@@ -0,0 +1,170 @@
...
+            # network trouble or a Microsoft outage, not an auth problem
+            raise ProviderUnavailableError(f"Microsoft token refresh failed: {err}") from err
+        self._access_token = str(payload["access_token"])
+        self._expires_at = time.time() + float(payload["expires_in"])
+        # Microsoft rotates the refresh token on every use and the old one only
+        # stays valid for a limited time, so persist the newest to config to
+        # survive restarts
+        if (new_refresh := payload.get("refresh_token")) and new_refresh != self._refresh_token:
+            self._refresh_token = new_refresh
+            self.mass.config.set_raw_provider_config_value(
```
MarvinSchenkel, line 150: [PROBLEM] Rotated refresh tokens use the delayed save path, so a crash within the five-second save window can lose the only valid token; pass `immediate=True` when persisting it.

music_assistant/providers/filesystem_onedrive/provider.py
```diff
@@ -0,0 +1,164 @@
...
+        # the SDK just needs a coroutine that returns a fresh access token
+        self.client = OneDriveClient(self.auth.async_get_access_token, mass.http_session)
+        self._root_folder_name: str | None = None
+
+    @property
+    def instance_name_postfix(self) -> str | None:
+        """Return a (default) instance name postfix for this provider instance."""
+        return self._root_folder_name
+
+    async def handle_async_init(self) -> None:
```
MarvinSchenkel, line 80: [PROBLEM] The new provider has no tests for token rotation, folder resolution, listing conversion, or Range forwarding; add focused OneDrive provider and auth tests based on the Google Drive tests.

music_assistant/providers/filesystem_onedrive/auth.py
OzGav: Pretty serious edge case but easy to fix so done!

music_assistant/providers/filesystem_onedrive/__init__.py
OzGav: Yeah I didn't try a re-auth so didn't find this one. Fixed. Also this code was copied from the Google Drive provider, so the same bug exists there on dev. I will fix that in a separate small PR.

music_assistant/providers/filesystem_onedrive/provider.py
OzGav: Fair. These have been added.

MarvinSchenkel (approved):
Another nice addition, thanks @OzGav !

PR #4802, opened 2026-07-14
Offer fake mute control for players with protocol-provided volume

OzGav:

Fixes the report that squeezelite players no longer have a fake mute option. Under player settings the mute control only offered "None", even though the documentation says fake mute is available for squeezelite devices.

The cause appears to be the fake mute option is only shown when a player natively supports volume. Since the protocol player rework, squeezelite devices are wrapped by the universal player, and the wrapper does not have native volume itself. Its volume comes from the linked squeezelite protocol player. The settings builder only looked at the native capability, so the option quietly disappeared for these players. It still worked for players like sync groups that report volume natively, which matches what the reporter observed.

Fake mute itself works fine for these players. It simply sets the volume to zero and restores the previous level on unmute, using whatever volume control is configured, including protocol volume. Only the option list was wrong.

The fix offers the fake mute option when the player has any usable volume path, native or via a linked protocol player. As a side effect this also makes mute on a sync group of squeezelite players work again, because group mute forwards to the members, which previously had no mute control to forward to.

- related issue https://github.com/music-assistant/support/issues/5832

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/config/players.py
+++ b/music_assistant/controllers/config/players.py
@@ -735,7 +735,9 @@ class PlayerConfigMixin:
             ConfigValueOption(PLAYER_CONTROL_NONE),
         ]
         mute_options.append(ConfigValueOption(PLAYER_CONTROL_NONE))
-        if player.supports_feature(PlayerFeature.VOLUME_SET):
+        # fake mute drives the volume control, so offer it when the player has any
+        # usable volume path (native or via a linked protocol player)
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Looks good to me, can we add a regression test before we merge this?

OzGav:
Yes we can! Tests added.

MarvinSchenkel (approved):

PR #4803, opened 2026-07-14
Keep metadata scan tasks running when a library row has corrupt metadata JSON

OzGav:

A user reported that their nightly "Scan missing artist metadata" task always failed with "Task failed: malformed JSON". That error message comes from sqlite itself. The scan tasks use `json_extract` in their database queries, and sqlite raises a fatal error the moment json_extract reads a metadata value that is not valid JSON. So a single corrupt row in the artists or playlists table  made the whole query fail before any item was processed. The result was that automatic metadata enrichment silently stopped working for the entire library, which the user noticed as artist images never being downloaded.

This change does two things for both affected scan tasks (artist metadata scan and playlist metadata refresh):

1. The scan queries now filter out rows with invalid metadata using sqlite's `json_valid` function, so one corrupt row no longer takes down the whole scan. Rows with empty (NULL) metadata are still scanned as before.
2. Each corrupt row is reported in the task log so the user knows about the problem and how to fix it, for example: "'Some Artist' has corrupt metadata and was skipped. To repair, remove 'Some Artist' from the library; it will be re-added with fresh metadata on the next library sync (artists id 1234)." The task then finishes as partial success instead of failing outright.

Claude verified against sqlite 3.50 with a deliberately corrupted row: the old query reproduces the exact reported error, the guarded query skips the bad row and returns everything else, and the detection query identifies the corrupt row.

Following Marvin's suggestion, the corrupt rows found by the last scan pass are also kept in a private list on the metadata controller and returned through `get_diagnostics()`. They show up as a `core.metadata` section in the diagnostics report, so we can capture affected rows directly from a report attached to an issue. The section is omitted entirely while no corrupt rows are known, and repaired rows drop out again on the next scan pass. Note the list starts empty after a restart until the scheduled scan has run.

- related issue https://github.com/orgs/music-assistant/discussions/5846#discussioncomment-17636407

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/metadata/controller.py
+++ b/music_assistant/controllers/metadata/controller.py
@@ -394,13 +394,17 @@ class MetaDataController(
     async def _scan_missing_artist_metadata(self) -> None:
         """Scan for artists with missing metadata."""
         update_current_task_progress_text("Searching for artists with missing metadata")
+        await self._report_corrupt_metadata_rows(DB_TABLE_ARTISTS)
         missing_images = (
             f"(json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') ISNULL "
             f"OR json_extract({DB_TABLE_ARTISTS}.metadata,'$.images') = '[]')"
         )
         missing_description = f"json_extract({DB_TABLE_ARTISTS}.metadata,'$.description') ISNULL"
         never_refreshed = f"json_extract({DB_TABLE_ARTISTS}.metadata,'$.last_refresh') ISNULL"
-        query = f"({missing_images} OR {missing_description}) AND {never_refreshed}"
+        query = (
+            f"{_valid_metadata_guard(DB_TABLE_ARTISTS)} AND "
+            f"({missing_images} OR {missing_description}) AND {never_refreshed}"
+        )
         artists = await self.mass.music.artists.get_library_items_by_query(
             limit=METADATA_SCAN_BATCH_SIZE,
             order_by="random",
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
As discussed on Discord:

I think we should figure out **why** we seem to have corrupted metadata json in our database and fix the root cause + provide a once of migration correction rather than fighting the symptom.

marcelveldt:
> I think we should figure out **why** we seem to have corrupted metadata json in our database and fix the root cause + provide a once of migration correction rather than fighting the symptom.

I was about to say the same - fix at the root, cleanup existing malformed entries - and not workaround it

OzGav:
I asked the reporter

> Do you have any idea as to how your database could have been corrupted. Server shutdown ungracefully?

and they said

> Yes that could have happened. We do suffer occasional power cuts, but I'm not aware of anything like that while MA might have been doing anything likely to actively update the db, but possible I guess.

MarvinSchenkel:
> I asked the reporter
> 
> > Do you have any idea as to how your database could have been corrupted. Server shutdown ungracefully?
> 
> and they said
> 
> > Yes that could have happened. We do suffer occasional power cuts, but I'm not aware of anything like that while MA might have been doing anything likely to actively update the db, but possible I guess.

You mentioned he was a database developer? Can we ask him to pull the actual corrupted metadata from the local database so we can see the actual corruption? That will help a lot.

MarvinSchenkel:
@marcelveldt Could we simply add a diagnostics for these corrupted rows so we can easily capture them?

music_assistant/controllers/metadata/controller.py
```diff
@@ -394,13 +403,17 @@ def _get_metadata_lookup_task_id(uri: str) -> str:
     async def _scan_missing_artist_metadata(self) -> None:
         """Scan for artists with missing metadata."""
         update_current_task_progress_text("Searching for artists with missing metadata")
+        await self._report_corrupt_metadata_rows(DB_TABLE_ARTISTS)
```
MarvinSchenkel: Instead of **pro**actively scanning, can't we just use the `except` to only do **re**active reporting?

OzGav: What are you thinking to react to?

MarvinSchenkel (approved):
LGTM, thanks @OzGav

PR #4804, opened 2026-07-14
Bulk-resolve Sonic Similarity candidates to stop event-loop stalls

chrisuthe:

Sonic Similarity reranking/filtering resolved every ANN candidate with a full `tracks.get()` inside an unbounded `asyncio.gather` — ~200 DB round-trips + model builds per `/similar` call (~800 per Discover refresh), on the event-loop thread.

Scoring only reads genres, artist names and album year, so candidates are now resolved with **one bulk library query per provider**; filter and rerank share that single resolution, and the rare analyzed-but-unlibraried candidate uses a bounded provider fallback. Scoring math is unchanged, so rankings are identical.

The one core touch is a `provider_item_ids` batch parameter added to `get_library_items_by_prov_id`.

Measured on a 9,973-track library — `sonic_similarity/similar` p95 under a concurrency ramp:

| concurrent calls | before | after | speedup |
|---|---|---|---|
| 1 | 1.1 s | 46 ms | 24× |
| 8 | 8.8 s | 442 ms | 20× |
| 32 | 31.6 s | 1.5 s | 21× |

credit @MarvinSchenkel  for the find!

## Types of changes

## Checklist


music_assistant/providers/sonic_similarity/provider.py
```diff
@@ -1121,38 +1130,73 @@ def _track_genres(track: Track) -> set[str]:
...
+        # provider under a small semaphore so their network latency, unlike the
+        # old unbounded fan-out, can't starve the loop.
+        misses = [key for key in seen if key not in resolved]
+        if misses:
+            semaphore = asyncio.Semaphore(PROVIDER_FALLBACK_CONCURRENCY)
+
+            async def _fallback(item_id: str, provider: str) -> None:
+                async with semaphore:
+                    try:
+                        track = await self.mass.music.tracks.get_provider_item(item_id, provider)
```
marcelveldt: Be aware that this is an API call for the provider so always be careful when executing it.

- we should always prefer an existing library item whenever possible (get_library_item_by_prov_id
- we should request items from providers sparsely and never in giant bulk/bursts 
- providers are heavily rate limited - the more calls we do for nothing/little gain, the more they rate limit us

chrisuthe: Good callout, let me make sure we're limiting things for non-library items.

chrisuthe: So, after some back and forth, this could become a problem if someone has audio analysis on  and listens to a bunch of spotify or other provider songs that are not library items. in that situation they'd all be indexed as expected and could be within the 200 candidates, which would then mean we could be making large burst calls to get artist, year, genre etc so we can further filter/weight. 

To avoid this, I can remove the metadata bonus from the ranking for any tracks not in library, so we basically wait until they are selected (meaning, they will be 1 of only 12-25 visible, used results by the time we hit external services).

This raises a bigger question in my head @marcelveldt , do we need a shadow library entry for these items, to avoid calling the provider for data that isn't changing (genre, year, etc)?

marcelveldt: One thing that popped in my head a couple of times (and its actually a model we had in the code years ago) is to simply put everything we see come across the wire in the library db - just none of their provider mappings have "in_library" set to true so it basically means it is just there for reference. I wont show up in any listing, its just there as a placeholder with whatever data we had at the moment we saw this item

When the item later gets added to the library, the metadata fulfill request kicks in and will complement the data.
The harsh truth however is that many streaming providers only provide very limited metadata in their listings results (so playlist tracks, album tracks etc.) which makes sense from a data hygiene and performance aspect, we actually do the same with our list results.

The other downside of this is that you start filling the library db with tracks you may never ever see again, meaning unbounded growth with items that are not the main interest. It does however help a ton with resolving cases like this. If we ever saw this item in any result, we have its data "cached", ready to use without needing to do an api call to get the full data.

An alternative is that we smartly use the cache db for this (which has expiry builtin) to quickly store these items for later use but I have not yet seen a clear model how we can ever get that implemented in a clean way without hacks. Also, going back to what I said above about the "list" items being just compact variants of these media items, they are probably lacking the metadata we care about anyways so we can wonder if this is all worth it.

So for now (until we ever have some global cache for all MA users) the stance still is that we should be very careful in making too many calls to streaming providers (or metadata providers). A quick win for now could just be that you try a fetch in the library first - if that does not deliver you any results at all, you do a small batch of resolving provider items.

For example whenever you try to do a "get_track" on a spotify track, you are within their rate limiter. While you are doing these calls, the user will not be able to do a search, fetch the contents of an album, playlist etc.  and it can even impact audio playback if the rate limiter really kicks in.


Hope you get my point

chrisuthe: Point taken, interesting that we've already had a shadow library in the past. I do hate setting up for a situation where we get the same known-static data more than once from a rate-limited provider but it'd need a second call for most tracks anyway, or even just end up not having the information we need the way it sounds. Maybe some day we solve this with that public repository of data that we can let our users use unrestricted/fast enough to not have to try store everything locally if they don't want to.

I wonder if a "Store information on everything I play, even if I don't add it to my library" would be a way to enable us to use the ID lookup, meta-data providers, etc to flesh out the DB enough to make this all work, but then again, as noted, that is an essentially uncapped bloat-fest, depending on the user.

I'm going to _not_ hit providers of non-library tracks to get the items we need for post-search weighting (year, genre) - this will mean we won't ever have the possibility of hitting 200+ calls if someone only listens to (and processes) all their music from a non-local source and puts none of it in library.  I will make a call if any tracks of that qualification (not in library, but previously listened to/analyzed) make the final post-weighted list of 12-15 tracks. That should give us a safety net without completely ruling those tracks out of being used.

music_assistant/providers/sonic_similarity/provider.py
```diff
@@ -1121,38 +1129,49 @@ def _track_genres(track: Track) -> set[str]:
...
+        Bulk-resolve candidate library tracks, keyed by (item_id, provider).
+
+        Scoring (filters + rerank) only reads genres, artist names and album
+        year — never display — so this batches one library query per provider
+        instead of a full tracks.get() per candidate (the old N+1 that stalled
+        the event loop). It stays library-only and never calls a provider:
+        candidates not in the library (e.g. streamed-but-not-added tracks the
+        analysis provider still indexed) are simply left out of the map, so
+        scoring can never burst a rate-limited provider API. Callers read a
+        missing key as no rerank bonus, and drop it when filtering.
```
OzGav: The general rule is to not have explanations of inner workings in the docstring. Totally fine as inline comments if it is needed.

marcelveldt (approved):
Thanks, @chrisuthe !

PR #4372, opened 2026-06-20
Adapt methods for audiobook collections support

fmunkes:
# What does this implement/fix?

Adds additional methods for audiobook collections. I tried to move more logic into sql queries to make it faster too. The frontend does not use the collections method, only the other two. 

Frontend PR: https://github.com/music-assistant/frontend/pull/1941
Models PR: https://github.com/music-assistant/models/pull/267

**Related issue (if applicable):**

- related issue <link to issue>

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [x] Enhancement to an existing feature — `enhancement`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [ ] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [x] The code change is tested and works locally.
- [x] `pre-commit run --all-files` passes.
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.
- [ ] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.
- [ ] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate.

PR #4806, opened 2026-07-14
Allow collapsing of collections in base media controller

fmunkes:

supersedes https://github.com/music-assistant/server/pull/4372

Not sure with the collection item ordering yet, had a chat with chatgpt about that.
Unfortunately the type hints are not inherited, so the overloads became a bit excessive. Now with this implementation we can just use our regular library_items call to get media items, and have them collapse in the UI.

This PR has the change for controllers using the base controller's "library_items" and the audiobooks controller. But the query is universal, so if what I added is fine, I'm happy to do the adjustments for albums, artists etc as well, so we may allow collections there as well.

Requires: https://github.com/music-assistant/models/pull/305 (merged) and new https://github.com/music-assistant/models/pull/309
Frontend PR: https://github.com/music-assistant/frontend/pull/2155

## Types of changes

## Checklist


fmunkes:
Regarding https://github.com/music-assistant/server/pull/4806/commits/564cfa730daec3ce0f3b4c9eae44dc477806d38f - I do not understand, why I need to assert a list here. The test is fine, the type hints don't indicate anything else than a list of something.

music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -158,11 +208,11 @@ async def library_items(
         provider: str | list[str] | None = None,
         genre: int | list[int] | None = None,
         played_only: bool = False,
-        without_collections: bool | None = None,
         *,
         summary: bool = True,
+        collapse_collections: bool = False,
         **kwargs: Any,
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/controllers/music/media/base.py
```diff
@@ -528,6 +592,37 @@ async def search(
...
+                item_id=db_rows[0]["name"],
+                name=db_rows[0]["name"],
+                provider="library",
+                provider_mappings=set(),
+                items=UniqueList(
+                    [
+                        self.item_cls.from_dict(self._parse_db_row(json_loads(x)))
+                        for x in json_loads(db_rows[0]["media_data"])
+                    ]
+                ),
```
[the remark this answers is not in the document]
fmunkes: that should never happen I think

music_assistant/controllers/music/media/base.py
```diff
@@ -1182,9 +1341,44 @@ async def get_library_items_by_query(  # noqa: PLR0913
...
+            items: list[ItemCls | MediaCollection[ItemCls]] = []
+
+            def _parse_method(x: str) -> ItemCls:
+                if summary:
+                    return cast("ItemCls", self._parse_summary_row(json_loads(x)))
+                return cast(
+                    "ItemCls",
+                    self.item_cls.from_dict(self._parse_db_row(json_loads(x))),
+                )
+
```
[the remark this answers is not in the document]
fmunkes: These are json.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1974,101 @@ def _parse_summary_artist_mappings(
...
+            END COLLATE NOCASE
+        )
+        GROUP BY collection_title
+
+        UNION ALL
+
+        SELECT 'single', name, search_name, search_sort_name, {json_object} FROM joined_table
+            WHERE {collections_column} IS NULL
+                OR {collections_column} = "[]"
+        """
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/controllers/music/media/base.py
```diff
@@ -528,6 +592,37 @@ async def search(
...
 
+    async def get_collection(self, name: str) -> MediaCollection[ItemCls]:
+        """Get a single collection."""
+        query_params: dict[str, Any] = {}
+        sql_query, base_query_params = self._build_final_query([], [], None, summary=False)
+        for key, value in base_query_params.items():
+            query_params.setdefault(key, value)
+        sql_query = await self._adapt_query_for_collections(
+            sql_query, query_params, summary=False, order_by=None, collection_name=name
+        )
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1974,101 @@ def _parse_summary_artist_mappings(
...
+        UNION ALL
+
+        SELECT 'single', name, search_name, search_sort_name, {json_object} FROM joined_table
+            WHERE {collections_column} IS NULL
+                OR {collections_column} = "[]"
+        """
+
+        if collection_name:
+            adapted_sql_query += f") WHERE name = '{collection_name}'"
+            return adapted_sql_query
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/controllers/music/media/base.py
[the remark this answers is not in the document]
marcelveldt: verified this with an actual sqlite test on 3.40.1 (which is what our docker image ships) and 3.53: the json subtype does not survive the subquery boundary, so json_group_array receives plain TEXT and produces an array of json *strings* here. The double decode is correct, copilot is wrong.

music_assistant/controllers/music/media/base.py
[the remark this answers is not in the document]
marcelveldt: same as the other thread: verified these are json strings at runtime, so the double decode is correct here.

music_assistant/controllers/music/controller.py
```diff
@@ -887,6 +889,7 @@ async def get_item_by_uri(
     ) -> MediaItemType | BrowseFolder:
         """Fetch MediaItem by uri."""
         media_type, provider_instance_id_or_domain, item_id = await parse_uri(uri)
+        self.logger.error("%s %s %s", media_type, provider_instance_id_or_domain, item_id)
```
marcelveldt: leftover debug logging, please remove (it's also spamming the test logs at error level)

music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -484,6 +486,32 @@ async def _resolve_media_items(
...
+                media_item.item_id
+            )
+            if collection_item_media_type != MediaType.AUDIOBOOK:
+                self.logger.error("Collections are only available for audiobooks.")
+                return []
+            if TYPE_CHECKING:
+                assert isinstance(media_item, MediaCollection)
+            book: Audiobook | None = None
+            for item in media_item.items:
+                _book = Audiobook.from_dict(item)
```
marcelveldt: This will crash at runtime: the items of a collection are already `Audiobook` objects at this point - both when resolved server side (`get_collection` builds real objects) and when passed in over the api (the serialization strategy in the models PR deserializes them into `Audiobook` objects). mashumaro's `from_dict` raises a `ValueError` on anything that is not a dict. Just use `item` directly here - the `items[0]` fallback below is then fine as well.

music_assistant/helpers/collections.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def get_collection_item_id(collection_name: str, item_media_type: MediaType) -> str:
+    """Get item id for a collection."""
+    return COLLECTION_ITEM_ID_SEPARATOR.join((item_media_type.value.lower(), collection_name))
+
+
+def get_collection_name_from_item_id(collection_item_id: str) -> str:
+    """Get collection's name from item id."""
+    return collection_item_id.rsplit(COLLECTION_ITEM_ID_SEPARATOR, maxsplit=1)[-1]
```
marcelveldt: rsplit splits off the *last* `___`, so a collection name that itself contains `___` gets mangled (`audiobook___Foo___Bar` -> `Bar`) and the lookup will fail. Split from the left instead: `collection_item_id.split(COLLECTION_ITEM_ID_SEPARATOR, maxsplit=1)[1]`

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,104 @@ def _parse_summary_artist_mappings(
...
+
+        sql_query = f"""
+        SELECT * FROM (
+
+            WITH
+                joined_table as ({sql_query}),
+                collection_extract as (
+                    SELECT
+                        name as media_name,
+                        json_extract(iter_coll.value, "$.title") as collection_title,
```
marcelveldt: please use single quotes for the json path literals ($.title, $.sequence and the $.collections variant above): double quoted strings in sqlite are the deprecated DQS misfeature. It happens to work because default builds still allow it, but builds compiled with SQLITE_DQS=0 will throw 'no such column' here - and the rest of the codebase consistently uses single quotes for json paths.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,104 @@ def _parse_summary_artist_mappings(
...
+                )
+            SELECT
+                'collection' as type,
+                collection_title as name,
+                replace(lower(collection_title),' ','') as search_name,
+                replace(lower(collection_title),' ','') as search_sort_name,
+                json_group_array(media_data) as media_data
+            FROM (
+                SELECT * FROM collection_extract
+                ORDER BY collection_title,
```
marcelveldt, line 2088: fyi: relying on the subquery ORDER BY to control the aggregation order of json_group_array is undocumented sqlite behavior. I verified it works on 3.40.1 (what the docker image ships) and 3.53, so fine for now, but please add a small inline comment so nobody 'optimizes' this subquery away later. sqlite 3.44+ supports ORDER BY inside the aggregate itself (which is guaranteed), we can switch once the image moves past 3.40.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,104 @@ def _parse_summary_artist_mappings(
...
+
+        if collection_name:
+            sql_query += " WHERE type = 'collection' AND name = :collection_name"
+            return sql_query
+
+        if search:
+            sql_query += "WHERE search_name LIKE :search"
+            return sql_query
+
+        supported_order_keys = ("name", "sort_name")
```
marcelveldt: Everything except name/sort_name is silently dropped here - e.g. `timestamp_added_desc` or even `name_desc` leaves the union completely unordered, which also makes limit/offset pagination non-deterministic. At minimum support the `_desc` variants and fall back to a stable default order; timestamp based sorting could be supported by selecting an aggregated value (e.g. `MAX(timestamp_added)`) in both branches of the union.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,104 @@ def _parse_summary_artist_mappings(
...
+        self,
+        sql_query: str,
+        query_params: dict[str, Any],
+        summary: bool,
+        order_by: str | None,
+        collection_name: str | None = None,
+        search: str | None = None,
+    ) -> str:
+        # get column names of base query
+        db_rows = await self.mass.music.database.get_rows_from_query(
```
marcelveldt: nit: this runs the full base query (limit 1) on every listing call just to discover column names. The column set is static per controller + summary flag, so this can be computed once and cached.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,104 @@ def _parse_summary_artist_mappings(
...
+                        json_extract(iter_coll.value, "$.sequence") as collection_sequence,
+                        {json_object} as media_data
+                    FROM (
+                        SELECT * FROM joined_table
+                    ), json_each({collections_column}) as iter_coll
+                )
+            SELECT
+                'collection' as type,
+                collection_title as name,
+                replace(lower(collection_title),' ','') as search_name,
```
marcelveldt, line 2075: note: `search_name` for singles is fully normalized via create_safe_string (unaccented, alphanumeric only) but for collections it's only lower + space-strip, so collection titles with accents or punctuation won't be found by search (search term 'kanguru' will never match 'känguru'). Fine as a followup.

marcelveldt (changes requested):
Had a good look at this one, including the two remaining copilot threads. Those are wrong: I verified with an actual sqlite test on 3.40.1 (which is what our docker image ships) and 3.53 that the json subtype does not survive the subquery boundary, so `json_group_array` produces an array of json *strings* and the double decode is exactly right. Resolved both threads.

That said, I did find some real issues - see inline. The media_resolver one is the important one (playing a collection will crash). CI red is expected until the models PR is merged + released + pinned here; most of the mypy errors will disappear with that bump. Also please link the frontend companion PR - this removes the old `music/audiobooks/collections` endpoint, so server and frontend need to land together.

Nice direction btw, having collections collapse straight in the regular library_items flow is exactly where this should live.

music_assistant/controllers/music/controller.py
fmunkes: removed.

music_assistant/helpers/collections.py
fmunkes: Fixed. Good catch!

music_assistant/controllers/music/media/base.py
fmunkes: Fixed. I always learn something new...

music_assistant/controllers/music/media/base.py
fmunkes: I added a comment. I didn't even realize that I am using something undocumented here...

music_assistant/controllers/music/media/base.py
fmunkes: I added support for all common sort keys using your hint of MAX(...) or SUM(...) when applicable. For a stable default order, I used name.
I additionally added the duration sort key at the example of the audiobooks controller, which we can then extend if we were to add collection support to other media types.

music_assistant/controllers/music/media/base.py
fmunkes: I added caching, I use the category to distinguish between summary / non-summary.

music_assistant/controllers/music/media/base.py
marcelveldt: We had to update our base image anyways and that includes the newer sqlite build so we are safe anyways (I hope haha)

fmunkes:
Alright. Thanks a ton for the extensive feedback! I'll linked the frontend PR as well.

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: fixed as well.

music_assistant/controllers/music/media/base.py
fmunkes: Noted. I try to get to this one in a weekend follow-up

music_assistant/controllers/player_queues/media_resolver.py
marcelveldt: yep, this is sorted now - using the item directly. thanks!

music_assistant/controllers/music/media/base.py
marcelveldt: almost there - the `json_object = None` on the line right after the cache.get means it always falls through and recomputes, so the cache never actually kicks in. drop that line and it's good.

music_assistant/controllers/music/media/base.py
marcelveldt: fyi these are actually still double-quoted ("$.title" etc). works fine on our sqlite build so not blocking, just flagging in case you meant to switch them to single quotes.

marcelveldt:
nice work, the feedback from the last round is addressed - playback, the id split and the sort keys all look good now.

two things left before this can go green:
- lint is still red because the `1.1.168` models pin doesn't include `MediaType.COLLECTION` yet - that needs models#309 merged + released first, then bump the pin here.
- one small caching bug, see inline (the `json_object = None` line defeats the cache you added).

the double-quoted json paths are still there too but that's non-blocking on our sqlite build.

music_assistant/controllers/music/media/base.py
fmunkes: oops. Debugging leftover. Gone now.

music_assistant/controllers/music/media/base.py
```diff
@@ -1779,3 +1982,154 @@ def _parse_summary_artist_mappings(
...
+                    "json_object(" + ",".join([f"'{x}',{x}" for x in db_rows[0].keys()]) + ")"  # noqa: SIM118
+                )
+            else:
+                json_object = "json_object()"
+            await self.mass.cache.set(
+                key=cache_key_json_object, category=int(summary), data=json_object
+            )
+
+        collections_column = (
+            "collections" if summary else 'json_extract("metadata", "$.collections")'
```
fmunkes: I converted the json_extract occurences to single quotes with the exception of this occasion. If I change this to
```"collections" if summary else "json_extract('metadata', '$.collections')"``` I'm left with a "malformed json" sqlite error for non-summary queries

fmunkes: Fixed with a fresh head. Just had to remove the quotes around metadata. I think all feedback comments are addressed now.

marcelveldt:
models release is building and once it's bumped, I'll trigger a rebase here

marcelveldt (approved):
Nice work @fmunkes !

PR #4813, opened 2026-07-15
Fix Sonos S1 enqueue failing with UPnP error 701

OzGav:

Fixes the linked regression where a Sonos S1 player plays the first track but rejects every enqueue of the next track with UPnP Error 701.

I believe the cause is a leftover from the stream URL refactor. `PlayerMedia.uri` used to hold the resolved HTTP stream URL but nowadays it holds the MA media uri such as `library://track/123`, and players resolve the stream URL themselves. The `create_didl_metadata` helper gained a url parameter for this reason and the dlna and wiim providers were updated to pass it, but the two call sites in the Sonos S1 provider appear to have been missed. The speaker therefore received a `NextURI` pointing at the stream server while the metadata res element referenced the unresolvable MA media uri, and S1 firmware rejects that mismatch on `SetNextAVTransportURI`.

The fix passes the resolved stream URL to `create_didl_metadata` in both call sites, matching the dlna and wiim providers. A regression test covers both and fails against the previous code.

I could not verify this on real S1 hardware, but the difference between the provider fixes made the omission obvious, and it lines up exactly with the reported regression window.

- related issue https://github.com/music-assistant/support/issues/5852

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos_s1/player.py
+++ b/music_assistant/providers/sonos_s1/player.py
@@ -207,7 +207,7 @@ class SonosPlayer(Player):
             # Sonos really does not like FLAC streams without duration
             stream_url = stream_url.replace(".flac", ".mp3")
 
-        didl_metadata = create_didl_metadata(media)
+        didl_metadata = create_didl_metadata(media, stream_url)
         is_announcement = media.media_type == MediaType.ANNOUNCEMENT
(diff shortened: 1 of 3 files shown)


MarvinSchenkel (approved):
Nice find!

PR #4814, opened 2026-07-15
Allow adding players to a dynamic sync group when all members are offline

OzGav:

Fixes the linked report where a dynamic sync group with a standing member that is currently unavailable refuses to accept any other players.

In dynamic mode the group derives its groupable candidates from the members that are currently available. When every listed member was offline this produced an empty set, and the fallback that offers any compatible player only triggered for a group with no members at all. The controller then rejected every join attempt against that empty set, so the group was unusable until a preset member came back online. This appears to be an unintended side effect of the protocol linking rework, since before that rework the code fell through to offering all compatible players whenever no member was available.

The fix runs the member aggregation first and falls back to offering any compatible player whenever it produces nothing, which can only happen when no member is available. Regression tests cover both the offline fallback and the normal case where an available member supplies the candidates, and the fallback test fails against the previous code.

- related issue https://github.com/music-assistant/support/issues/5810

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sync_group/player.py
+++ b/music_assistant/providers/sync_group/player.py
@@ -215,19 +215,20 @@ class SyncGroupPlayer(Player):
         member_ids = self._attr_group_members if self._attr_group_members else []
         # current members bypass the allow-list filter (filter constrains joiners only)
         current_members = set(member_ids)
-        if member_ids:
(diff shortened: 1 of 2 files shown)


MarvinSchenkel (approved):
LGTM, thanks @OzGav

PR #4821, opened 2026-07-15
Remove outdated note that AirPlay 2 can't group

Kyzcreig:
## What

Removes the outdated note in the `airplay_protocol` setting description that says AirPlay 2 doesn't support audio synchronization / grouping.

## Why

The description currently ends with:

> NOTE: AirPlay 2 currently does not support audio synchronization. Grouping/syncing with other players is only available when using AirPlay 1 (RAOP).

That isn't accurate anymore — AirPlay 2 players can group. It's also contradicted by the code itself: the `can_group_with` docstring in `providers/airplay/player.py` says *"RAOP and AP2 players can group with other RAOP and/or AP2 players."* This came up on music-assistant/support#5831 where a maintainer noted the description needs updating, so I figured I'd send the fix.

## Change

Drops just the stale trailing `NOTE:` sentence from the `airplay_protocol` description in `providers/airplay/strings.json`, and regenerated `translations/en.json` via `uv run -m scripts.build_translations`. No behavior change — text only. The rest of the description (RAOP vs AP2 explanation + "automatic selection usually works") is untouched.


OzGav:
Whilst we really appreciate all contributions, I just gave you our AI policy in the discussion and you have used it here again, bypassing the PR template. Please replace the post with the template and you can change the note to "Grouping/syncing with other players is supported on both protocols. If grouped playback is out of sync, adjust the delay correction (under advanced settings)" instead of deleting it.

MarvinSchenkel (approved):
Thanks @Kyzcreig

PR #4839, opened 2026-07-17
Fix fake mute never reporting muted state

OzGav:

With fake mute control enabled, `cmd_volume_mute` set the fake mute flag before calling `_handle_cmd_volume_set`, which unconditionally clears that flag (a volume change normally implies an unmute). As a result the player kept reporting `volume_muted=false` after muting, the UI kept sending mute commands instead of unmute, and each repeated mute overwrote the stored previous volume with 0 so the player could never be unmuted back to its original volume.

This PR sets the flag after the internal volume command instead, and only store the previous volume when not already fake muted.

- related issue https://github.com/music-assistant/support/issues/5870

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -949,7 +949,7 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
                 player.state.name,
             )
             if muted:
-                if not player.extra_data.get(ATTR_FAKE_MUTE):
+                if player.extra_data.get(ATTR_FAKE_MUTE) is False:
                     # on a repeated mute command the volume is already 0
                     player.extra_data[ATTR_PREVIOUS_VOLUME] = player.state.volume_level
                 await self._handle_cmd_volume_set(player_id, 0)
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -949,7 +949,8 @@ class PlayerController(ProtocolLinkingMixin, CoreController):
                 player.state.name,
             )
             if muted:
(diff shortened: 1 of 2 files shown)


music_assistant/controllers/players/controller.py
```diff
@@ -949,9 +949,12 @@ async def cmd_volume_mute(self, player_id: str, muted: bool) -> None:
                 player.state.name,
             )
             if muted:
-                player.extra_data[ATTR_PREVIOUS_VOLUME] = player.state.volume_level
-                player.extra_data[ATTR_FAKE_MUTE] = True
+                if not player.extra_data.get(ATTR_FAKE_MUTE):
```
marcelveldt:
```suggestion
                if player.extra_data.get(ATTR_FAKE_MUTE) is False:
```

music_assistant/controllers/players/controller.py
```diff
@@ -949,9 +949,12 @@ async def cmd_volume_mute(self, player_id: str, muted: bool) -> None:
...
+                    # on a repeated mute command the volume is already 0
+                    player.extra_data[ATTR_PREVIOUS_VOLUME] = player.state.volume_level
```
marcelveldt, line 955: why overwrite it if it's already set ?

OzGav: Actually after I accepted the suggestion I realised the flag is None when unmuted (it gets popped, not set to False), so is False never matched. Also the diff didn’t read well so I’ve redone it to make the intent more obvious.

marcelveldt (approved):
Thanks, @OzGav !

PR #4846, opened 2026-07-18
Add sendspin encryption support

arturpragacz:

Add sendspin encryption support.

Fully backwards-compatible: unencrypted (legacy) clients still work.
There is a global sendspin provider toggle to disable that compatibility.
Limitation: the webplayer is currently not updated yet, so it will not work with compatibility disabled.

## Types of changes

## Checklist


music_assistant/providers/sendspin/provider.py
```diff
@@ -379,6 +549,207 @@ def is_virtual_player(self, player_id: str) -> bool:
...
+        """Abort an in-flight or parked PIN pairing session, restoring normal service."""
+        session = self._pin_sessions.pop(client_id, None)
+        if session is None:
+            return
+        self._cancel_pin_idle_timeout(client_id)
+        await self._end_pairing_quietly(client_id)
+        if session.task is not None:
+            with suppress(Exception):
+                await session.task
+        await self._refresh_player(client_id)
```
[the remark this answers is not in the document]
arturpragacz: Pin session task is only cancelled during provider unload.

marcelveldt:
I reviewed most of the changes and they look solid. It's a big change in any case. 
Since we have the compatibility mode enabled by default, I think we should just merge this now so we can test in nightly (and later beta) and then fix whatever we still may run into and if the UX of the pairing flow makes sense.

Just flagged a few minor review comments in-line.

music_assistant/providers/sendspin/__init__.py
```diff
@@ -40,4 +48,18 @@ async def get_config_entries(
...
+            type=ConfigEntryType.BOOLEAN,
+            default_value=True,
+        ),
+        ConfigEntry(
+            key=CONF_MIN_PIN_LENGTH,
+            type=ConfigEntryType.INTEGER,
+            range=(4, 12),
+            default_value=DEFAULT_MIN_PIN_LENGTH,
+            advanced=True,
+        ),
```
marcelveldt, line 63: why do we want this to be configurable ?
isn't it true that the client reports what pin type it supports/requires ?

music_assistant/providers/sendspin/player.py
```diff
@@ -137,9 +175,82 @@ def format_to_display_string(fmt: SupportedAudioFormat) -> str:
...
+    PairAbortReason.CONCURRENT_ATTEMPT: "pairing_error_concurrent",
+    PairAbortReason.LOCKED_OUT: "pairing_error_locked_out",
+    PairAbortReason.METHOD_NOT_SUPPORTED: "pairing_error_method_unsupported",
+    PairAbortReason.PIN_LENGTH_UNACCEPTABLE: "pairing_error_pin_length",
+    PairAbortReason.PIN_MISMATCH: "pairing_error_pin_mismatch",
+    PairAbortReason.USER_CANCELLED: "pairing_error_cancelled",
+}
+
+
+def _error_alert(err: Exception) -> _AlertText:
```
marcelveldt: should we move all this into a helpers.py ?

arturpragacz: Done.

music_assistant/providers/sendspin/__init__.py
arturpragacz: It is important to understand that our PIN pairing is not simply a password that grants access to the client.

It is a mutual authentication mechanism, where both the server and the client prove to each other that they are who they say they are, based on the shared knowledge of the same secret.

This is why the specification allows both sides to define the level of security they are comfortable with, represented by the PIN length. A longer PIN provides greater security, but also requires slightly more effort to type.

That said, I will lower the default here to the minimum value of 4. This means that for the majority of users who will realistically never change this setting, there will be no impact, and only the client's setting will effectively be the one that determines the required PIN length.

arturpragacz:
>Since we have the compatibility mode enabled by default, I think we should just merge this now so we can test in nightly (and later beta) and then fix whatever we still may run into and if the UX of the pairing flow makes sense.

Yes, I agree. There are some inherent UI quirks to this approach, but it is definitely robust enough for testing.

music_assistant/providers/sendspin/__init__.py
marcelveldt: As long as it's an advanced setting I am fine with that - like you said, the majority of users will not care about this setting and then like/want to have the minimum to be set to 4 characters.

marcelveldt (approved):
Amazing work, @arturpragacz !

PR #4850, opened 2026-07-19
Don't allow ffmpeg to try range requests when using POST data

khers:
ffmpeg chooses not to include Rage headers when doing a POST reconnect. Because of this, we need to ensure that it knows to pull the whole stream and decode+discard to seek.

Fixes: #5210

- related issue https://github.com/music-assistant/support/issues/5210

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/ffmpeg.py
+++ b/music_assistant/helpers/ffmpeg.py
@@ -546,6 +546,13 @@ def get_ffmpeg_args(
                 "-reconnect_on_http_error",
                 "5xx,429",
             ]
+            if "-post_data" in extra_input_args:
+                # ffmpeg does not include Range headers on POST reconnects, so byte-range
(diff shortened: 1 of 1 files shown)


marcelveldt (approved):
The change looks good to me. I'm assuming you tested if this actually works and solves the issue @khers ?

marcelveldt:
If verified fixed on nightly/beta, we should consider a backport to stable as well.

khers:
> The change looks good to me. I'm assuming you tested if this actually works and solves the issue @khers ?

I can't recreate the problem on the setup I have but I have tested that this still works on my working install. This is the best conclusion I can come to for why we are seeing the behavior in beta that we are.

khers:
> If verified fixed on nightly/beta, we should consider a backport to stable as well.

The issue it addresses is not in stable so I don't think it needs a backport.

PR #4462, opened 2026-06-27
Hue Lights Sync: music-structure visualizer — Club/Pulse modes, auto strobe, palettes & live preview

silasmariusz:
# What does this implement/fix?

A music-structure-aware visualizer upgrade for the **Hue Lights Sync** (`hue_entertainment`) provider. The lights now follow the *structure* of the music — verse, build, drop, sustain, breakdown — instead of only reacting to instantaneous loudness, and gain two new base effects, a distributed/auto strobe overlay, a 100-palette system, per-light control, and a live in-browser preview. Everything is configurable from the provider settings (English labels + examples) and applies live while music plays. Implemented against the public Hue Entertainment API; no new dependencies.

https://github.com/user-attachments/assets/0c36356d-f54d-413c-b46f-ee48dcc83959

**Related issue (if applicable):**

- n/a

## Highlights

**Music-structure detector** (`structure.py`)
- A small, cheap (band sums + EMAs) section state machine — `NORMAL / RISE / DROP / SUSTAIN / BREAK` — derived from the *shape* of the energy.
- Track-relative: smoothed loudness vs a slowly-decaying ceiling + spectral brightness, so it classifies real continuous club music (a sustained-loud drop reads as `SUSTAIN`; a filter-down where the kick keeps going but the highs vanish reads as `BREAK`) instead of sitting on `NORMAL`. Degrades gracefully with no onsets and free-runs tempo when beats are sparse.

**Two new base modes** (`analyzer.py`)
- **Club** — structure-aware: soft groove pulse in the verse, lights recruited one by one through a build, a hard white slam on the drop, a breathing blackout on breakdowns.
- **Pulse** — distributed decay-to-black sequential beat fire: one light fires and eases to black each beat, the hit chasing around the room, the whole room blooming on the downbeat. The per-light fire envelope is anchored to render time (not beat audio time), so it stays visible under any latency-compensation offset.

**Distributed + auto strobe overlay** (`strobe_overlay.py`)
- An overlay on the selected lights only: a hard, low-duty club strobe with a coverage cap, non-repeating random subsets, and a "return to preset" hand-back.
- **Auto strobe**: the structure detector drives engagement and flash rate — silent in verses/breaks, accelerating through builds, a hard 1/16 on the drop and an accent on the landing — all tempo-locked to the detected BPM. Tuned for house / trance / techno.

**Palettes, rotation & eased transitions**
[...]

PR #4853, opened 2026-07-19
Hue Lights Sync: optional colour output mode (RGB / Accurate / Vivid)

silasmariusz:
Adds a **Colour output** dropdown to the Hue Lights Sync provider: how rendered colours are encoded on the Entertainment wire.

- **RGB** (default, today's behaviour) — the widest colour range and the most dynamic fades. The HueStream RGB channel is 8-bit per colour with brightness baked into the magnitude, so very dim colours can lose hue accuracy (visible green/blue casts on some bulbs at low brightness).
- **Accurate (CIE xy)** — chromaticity and full 16-bit brightness are sent separately: colours stay consistent across mixed Hue models and dim cleanly.
- **Vivid (CIE xy)** — like Accurate, but saturated colours are stretched to the bulb's gamut edge, so cross-model consistency does not cost punch.

Notes for review:
- **Depends on music-assistant/hue-entertainment#2** (adds `ColorMode` and the xy/gamut math to the library). The provider soft-imports it: with the released lib 0.1.2 the dropdown is hidden and the wire format stays RGB exactly as today, so this is safe to merge before the library update — the feature auto-activates once the `hue-entertainment` pin is bumped.
- **Stacked on #4462** — the diff shows both changesets until #4462 merges; the colour-output change itself is the last commit only.
- Tested live on a 7-area setup (all three modes) — the xy path fixes the low-brightness green cast that RGB shows on some bulbs.

## Types of changes

## Checklist


marcelveldt:
So the question I have is if Accurate is the best representation, why don't we just pick that by default ?
I mean, what would be the reason for a user to change it ?

silasmariusz:
Hi @marcelveldt ...


> why don't we just pick that by default ?

Agree & good question....


"Accurate" means accurate to the source colour, not the most colourful the lamp can do. Now colours come from sRGB, and the sRGB triangle is smaller than the gamut of a Hue lamp. sRGB red sits at 0.640, 0.330 while gamut C red is at 0.692, 0.308.

So in Accurate mode the lamp never reaches its deepest colours. Not less often, never. That outer part of the lamp range is simply never used.

Boost takes the saturated colours and pushes them out to the edge of the lamp own gamut. Same hue, same brightness, only as deep as the lamp can go. On a music light show you see it right away on reds and blues.


[WHY NOT RGB?]
RGB is the old path. **The bridge does the conversion itself and you get the widest range**, but there is no separate brightness channel, so the same three numbers carry colour and level together. [pls note] **On fast music changes that flickers**. I kept it only so existing setups do not change behaviour.

So instead of leaving three modes, what about this:

1. always use the Hue colour path (xy)
2. one checkbox, "Boost colours", on by default
3. hide RGB, or drop it if you prefer 
(i would leave it, as you decided to push it first into beta, maybe lets keep it for a short of time)

Then there is nothing to understand. Accurate colours, or accurate hues at full strength.

Waiting for feedback

marcelveldt:
yes that sounds perfect!

silasmariusz:
Pushed e65306d - a fix for the 1/4..4/4 bar counter used by the section logic and the preview.

What was wrong: beat_in_bar and bar_phase divided (now - last downbeat) by the smoothed inter-beat interval. That average is never exact, and the error adds up across a bar, so the counter sometimes skipped a count or repeated one even when the beat schedule itself was correct. I reproduced it by replaying a perfect 128 BPM schedule and sampling the counter every 20 ms: the 1,2,3,4 cycle broke several times per track.

The fix: note_beat records each pushed beat together with its position in the bar, and the two getters look up the latest recorded beat at or before the render time. The old formula only runs before any beat has arrived, or once the schedule has run dry. The same replay after the fix shows zero broken transitions over 3 minutes.

Covered by unit tests on my side (the standalone box shares these engine files byte for byte): a synthetic perfect schedule and live tracker output both drive StructureDetector and assert the strict 1,2,3,4 cycle.

silasmariusz:
Pushed a9c3a83 - a new "When the beat disappears" setting plus a silence idle.

Why: on real DJ sets and melodic house the beat schedule legitimately dries up for long stretches (breakdowns, ambient intros). Until now each mode had its own fallback, and in full silence the lights either froze on the last frame or kept free-running a pulse into an empty room.

What it does: while music keeps playing without a detected beat, the lights either keep pulsing at the last known tempo, dimmed (ghost pulse), do a slow tempo-free swell (breathe), or react only to onsets as before - one setting, applied the same way in every base mode. Streams that never had a beat schedule keep their original look, so servers without beat analysis are unaffected. After a few seconds of full silence the room drops to a faint breathing drift and wakes as soon as audio returns.

Also in this commit: the bridge manager's update_settings had fewer parameters than the bridge's and raised TypeError on live updates of palette, strobe or pulse settings; it now forwards every keyword through.

Covered by unit tests on my side (the standalone box shares these files byte for byte): ghost pulses at the remembered tempo and is dimmed, breathe swells without flicker, beat-less streams keep the plain onset look, silence dims to idle and wakes on audio.

marcelveldt:
This PR is about adding one single extra option for the color mode, with one single extra checkbox for the "boost colors".
Yes it contains the whole gigantic bunch of changes from your other PR and various unrelated commits.
This is impossible for us to review, let alone approve.

Like discussed ion the previous PR:
Keep the PR's smaller and focused with clear reasoning why a change is needed so we can help you get your improvements merged. Thanks!

MarvinSchenkel:
Putting this PR in draft until we have merged the stacked PR first.

silasmariusz:
Yes you’re right. Please let me handle all this once I get back from vacation.

PR #4857, opened 2026-07-19
Add DSP gain and balance filters

OzGav:

Adds two new sound adjustments to the audio pipeline:

- Gain - adjusts the overall volume up or down by a set amount.
- Balance - shifts the sound left or right between the speakers. To do this it quietens the opposite side rather than boosting the near side, so the audio never gets louder than the original and there's no risk of distortion.

If either control is left at its centre/zero setting, it's skipped entirely and has no effect on the sound.

Added unit tests in `tests/helpers/test_dsp.py` covering the volume change, the left/right balance direction, the quieting amount, and that a centred/zero setting is left untouched.

Frontend PR: https://github.com/music-assistant/frontend/pull/2154

- related issue 

## Types of changes

## Checklist

- [X] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.

### DIFF
--- a/music_assistant/helpers/dsp.py
+++ b/music_assistant/helpers/dsp.py
@@ -156,12 +156,15 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
     if isinstance(dsp_filter, GainFilter) and dsp_filter.gain != 0:
         filter_params.append(f"volume={dsp_filter.gain}dB")
     if isinstance(dsp_filter, BalanceFilter) and dsp_filter.balance != 0:
-        # attenuate only the channel opposite the slider direction, so there is
-        # no positive gain and thus no clipping risk
-        attenuation = (100 - abs(dsp_filter.balance)) / 100
-        if dsp_filter.balance > 0:
-            filter_params.append(f"pan=stereo|FL={attenuation}*FL|FR=FR")
-        else:
-            filter_params.append(f"pan=stereo|FL=FL|FR={attenuation}*FR")
+        # balance is a stereo operation; on a non-stereo source the FL/FR pan
+        # expression would output silence, so only apply it to stereo streams
+        if input_format.channels == 2:
+            # attenuate only the channel opposite the slider direction, so there is
+            # no positive gain and thus no clipping risk
+            attenuation = (100 - abs(dsp_filter.balance)) / 100
+            if dsp_filter.balance > 0:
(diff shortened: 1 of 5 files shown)


music_assistant/helpers/dsp.py
```diff
@@ -151,5 +153,15 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
             filter_params.append(
                 f"equalizer=frequency=9000:width=18000:width_type=h:gain={dsp_filter.treble_level}"
             )
+    if isinstance(dsp_filter, GainFilter) and dsp_filter.gain != 0:
+        filter_params.append(f"volume={dsp_filter.gain}dB")
+    if isinstance(dsp_filter, BalanceFilter) and dsp_filter.balance != 0:
```
marcelveldt, line 158: [CRITICAL] Balance assumes stereo here. On a mono source this pan expression produces silence, so can we skip it unless `input_format.channels == 2` (or explicitly upmix first) and add mono coverage?

music_assistant/helpers/dsp.py
```diff
@@ -151,5 +153,15 @@ def filter_to_ffmpeg_params(dsp_filter: DSPFilter, input_format: AudioFormat) ->
             filter_params.append(
                 f"equalizer=frequency=9000:width=18000:width_type=h:gain={dsp_filter.treble_level}"
             )
+    if isinstance(dsp_filter, GainFilter) and dsp_filter.gain != 0:
```
marcelveldt, line 156: [PROBLEM] Can we also keep neutral Gain/Balance filters out of the effective chain? At `0` they emit no FFmpeg params, but `AudioDSPDetails` still reports them and marks the path non-bit-perfect.

marcelveldt:
Be aware that the audio chain is currently being reworked so these changes are now overlapping and should be adjusted for the new audio chain.

music_assistant/helpers/dsp.py
OzGav: Fixed. Balance now only applies when the source is stereo (channels == 2) and passes mono through untouched, with a test added.

music_assistant/helpers/dsp.py
OzGav: Done. Neutral Gain/Balance filters are now excluded from the effective chain, so they aren't reported in AudioDSPDetails and no longer mark the path non-bit-perfect.

marcelveldt (approved):
Thanks, @OzGav !

PR #4876, opened 2026-07-20
Send progress metadata on Sendspin playback-state transitions

chrisuthe:

Resuming a paused Sendspin player did not send a `server/state` with `metadata.progress`, so every connected Sendspin client kept showing the stale (paused) playback position for the rest of the track after any pause/resume. Per the Sendspin spec (README:1445), the server must send `progress` on every playback-state change (play, pause, resume, seek, speed change).

Root cause: resuming a paused player reuses the same `current_media` object (no new track). The metadata/progress push is only triggered by a debounced callback keyed on `current_media` identity changes (title/uri/queue_item_id/etc.), so a same-track pause/resume never fires it.

A related issue: `PlayerMedia.corrected_elapsed_time` extrapolates unconditionally from `elapsed_time_last_updated`, with no playing/paused check. Since the reused media object's anchor was stamped at pause time, a subsequent progress push would over-count the entire paused duration as playback time.

Fix, in `SendspinPlayer.group_event_cb` (fires whenever the group's playback state actually transitions):
- On `PAUSED -> PLAYING`, refresh `current_media.elapsed_time_last_updated` to now before `update_state()` runs, so `corrected_elapsed_time` extrapolates from the resume moment instead of through the paused span.
- On any `PLAYING`/`PAUSED` transition (leader only), explicitly schedule `send_current_media_metadata()`, independent of whether track identity changed.

- N/A

## Types of changes

## Checklist


tests/providers/sendspin/test_group_state_progress.py
```diff
@@ -0,0 +1,117 @@
+"""
+Tests for progress metadata being pushed on playback-state transitions.
+
+Regression coverage for: resuming a paused Sendspin group did not send a
+server/state with metadata.progress, because current_media's identity is
+unchanged across pause/resume, so the debounced media-updated callback never
+fired (README:1445 requires progress on every play/pause/resume/seek).
```
teancom: ~~Should probably mention this is the Sendspin spec's README, not the `server` repo. And I try not to use line numbers in comments since they immediately rot (this one is already off by a few lines compared to `head` of spec's main).~~ Actually I would just drop the whole parenthetical.

music_assistant/providers/sendspin/player.py
```diff
@@ -1016,8 +1016,28 @@ def _on_group_stopped(self) -> None:
...
+        if (
+            isinstance(event, GroupStateChangedEvent)
+            and event.state == PlaybackStateType.PLAYING
+            and self._attr_playback_state == PlaybackState.PAUSED
+            and self._attr_current_media is not None
+        ):
+            # current_media's identity is unchanged on resume, so update_state()
+            # below won't debounce a metadata push; refresh the anchor now so
+            # corrected_elapsed_time doesn't extrapolate through the paused span.
+            self._attr_current_media.elapsed_time_last_updated = time.time()
```
teancom: If I'm tracing through everything right (not guaranteed, this gets hairy), this is never *actually* used/read. It mutates `_attr_current_media` (the provider-attr object from `play_media`, elapsed fields None), while `send_current_media_metadata` reads `self.state.current_media` (player.py:1489), the derived object rebuilt from queue anchors, which never consults `_attr_current_media`'s elapsed fields in the queue-active branch. Sendspin is a flow-mode MA-queue player, so real playback always takes that branch. And the actual field that's sent, the derived media anchor, still has the possibility of the double-count bug (depending on event-loop scheduling).

teancom (changes requested):
Good catch!

music_assistant/providers/sendspin/player.py
chrisuthe: Good catch, and you're right — traced it through and _attr_current_media really is never read here. send_current_media_metadata() reads self.state.current_media, which the queue controller rebuilds fresh from its own cached elapsed-time fields, never touching _attr_current_media at all.

Turned out there were two anchors that needed fixing, not one: the player's own _attr_elapsed_time_last_updated (only advanced while audio is actively streaming, so it freezes across a pause), and the queue controller's cached copy of it (only refreshed via a debounced callback that the metadata push otherwise races ahead of). Updated fix refreshes both — the player's anchor before calling super(), and self.state.current_media directly right after, so the immediate push is correct without waiting on the debounce. Rewrote the tests to check both of those instead of _attr_current_media. Pushed in 5cdb016b.

tests/providers/sendspin/test_group_state_progress.py
chrisuthe: Good point, done — dropped the parenthetical entirely rather than trying to word around the line-number issue. Pushed in 5cdb016b.

music_assistant/providers/sendspin/player.py
```diff
@@ -1016,23 +1016,34 @@ def _on_group_stopped(self) -> None:
...
+        )
+        if is_resume:
+            # _attr_elapsed_time_last_updated is only advanced by playback.py's commit
+            # loop, which stops while paused - so it's still anchored to the moment
+            # playback paused. Fast-forward it now, before update_state() below flips
+            # playback_state to PLAYING, so corrected_elapsed_time doesn't extrapolate
+            # across the paused span.
+            self._attr_elapsed_time_last_updated = time.time()
         super().group_event_cb(group, event)
+        if is_resume and self.state.current_media is not None:
```
teancom, line 1037: This isn't gated to exclude followers in a group, which means followers can update the leader's PlayerMedia object directly. And the player-level `_attr_elapsed_time_last_updated` refresh below should also be leader-only. We can fix it here and there, or when setting `is_resume`.

teancom (changes requested):

music_assistant/providers/sendspin/player.py
chrisuthe: Jeeze I need to pay attention, good catch! Check it out now, should be good in both scenarios and updated the test_synced_follower test.

teancom (approved):
LGTM! Thank you, @chrisuthe !

PR #4878, opened 2026-07-20
Fix flow-stream EOF recovery for universal-player-wrapped Cast devices

distante:

A non-Google (third-party) Chromecast is exposed as a `PlayerType.PROTOCOL` player and wrapped by a **Universal Player**, which is the entity that owns the queue. The end-of-queue flow-stream recovery added in #4406 resolves the queue-owning player from the active cast group or the player's own id. For a wrapped Cast that lands on the Cast's own id instead of the Universal Player that actually owns the queue, so the flow-stream-finished check always returns `False`. Since pychromecast reports the end-of-stream underrun as `BUFFERING` (which it treats as "playing"), the `BUFFERING → idle` recovery never fires and the player stays stuck reporting **"playing"** forever at the end of the queue — the elapsed-time counter never stops.

The fix also resolves the queue through the protocol parent, so a wrapped Cast finds the Universal Player's queue. No behaviour change for native/standalone Casts (which have no protocol parent) or Cast groups (the group still takes precedence). Verified on-device: at EOF the recovery check was `False` for the Cast's own id but `True` for its protocol parent.

Regression tests added under `tests/providers/chromecast/` cover the wrapped-Cast (recovery fires via the protocol parent), standalone-Cast, and Cast-group cases.

- related issue https://github.com/music-assistant/support/issues/5732

## Types of changes

## Checklist


marcelveldt:
As far as I'm aware the fix from https://github.com/music-assistant/server/pull/4406 was only needed for cast groups and those are never marked as protocol only.

Are you now telling that regular cast devices face the same issue ?

OzGav:
Not quite, regular Google devices aren't affected. The report behind this is a cheap third-party cast speaker (a Terris CCM 283), so being non-Google it gets wrapped by a universal player. The #4406 fix already runs on every cast device, not just groups, but for a wrapped device it was checking the cast's own queue instead of the universal player's, so it always came back "not finished" and the player sat on "playing" forever at the end of the queue. This PR points the check at the correct queue and doesn't change when the recovery is allowed to kick in.

I considered what this PR might break and the answer is basically nothing. The new lookup only applies to wrapped devices, so Google casts, groups and stereo pairs behave exactly as they do today. And a false trigger is as unlikely as it was in #4406, by the time we mark the stream as finished, the device already has all the audio, so if it's still stuck buffering at that point it really is the end of the stream. So far this is only confirmed on this one device, but the fix does nothing unless a device actually gets stuck in that state. The contributor has verified it on the device itself, and the PR includes tests covering the wrapped, native and group cases. On that basis I think this is sound.

OzGav (approved):

distante:
> As far as I'm aware the fix from https://github.com/music-assistant/server/pull/4406 was only needed for cast groups and those are never marked as protocol only.
> 
> Are you now telling that regular cast devices face the same issue ?

As @OzGav  said, the regular cast devices work as expected. The issue is with non-Google Chromecast devices. I was able to reproduce the issue everytime with the internal Ocean waves files.

PR #4887, opened 2026-07-21
Cast dashboards to display devices

MarvinSchenkel:

Adds the ability to show a Music Assistant dashboard (party mode, now-playing) on a display device or client — a Chromecast/Nest Hub today, an Android/Apple TV app or plain webview tomorrow — while audio keeps playing.

Dashboards are deliberately **not** part of the provider model: a dashboard endpoint is not a player or a music object. Instead, any API client can register itself as a dashboard endpoint, and show/hide are intent events on the event bus.

- n/a (feature)

## Types of changes

## How it works
- Clients call `dashboard/register` (`dashboard_id`, `name`, `supported_types`, `icon`). Registrations are owned by the websocket connection and cleaned up automatically on disconnect.
- `dashboard/show` / `dashboard/hide` (scope `users.invite`) emit intent-only `DASHBOARD_SHOW` / `DASHBOARD_HIDE` events (dashboard type + player_id, no URL). Native clients render locally; URL-based viewers resolve their URL via `dashboard/get_url`, which mints a one-time viewer code (dedicated `dashboard_viewer` guest user) into a fully-built frontend URL.
- `dashboard/dashboards?dashboard=<type>` lists endpoints filtered by supported type (the party page only offers party-capable endpoints); `DASHBOARDS_UPDATED` / `DASHBOARD_SESSIONS_UPDATED` keep clients live.
- In-server implementers use `DashboardController.register_dashboard_handler(device, on_show, on_hide)` — the controller resolves the URL and awaits the callback directly. The chromecast provider is the first implementer (`providers/chromecast/dashboard.py`): video-capable cast devices register at discovery and the existing MA cast app shows the dashboard; audio + dashboard coexist in one app session.

## Hardening (from review)
- Registrations are connection-owned: colliding or foreign `register`/`unregister` calls are rejected, in-server handlers can't be hijacked via the public API.
- `dashboard/get_url` requires `users.invite`, or ownership of an endpoint with a matching active session — a bare viewer token can't mint new codes.
- Show intent is validated up front on both paths (e.g. now-playing requires a `player_id`).
- Fix in the API arg parser: `set[...]` annotations were silently dropped to `None`.

## Dependencies (must land/deploy for this to work)
- **models** — companion PR music-assistant/models#320 (`DashboardType`, `DashboardDevice`, `DashboardSession`, dashboard events). Must be released and the pin bumped here before CI is green.
- **frontend** — companion PR music-assistant/frontend#2176 (cast button + viewer session + kiosk views).
- **cast receiver** — companion PR music-assistant/cast-receiver#1. ⚠️ **This must be deployed to `cast.music-assistant.io`** — the receiver is what renders the dashboard on the device, so the feature is inert in production until that deploy ships.

## Checklist

- [x] The code change is tested and works locally (tests under `tests/controllers/dashboard/` and `tests/providers/chromecast/`; verified end-to-end on real devices and with a scripted external API client).
- [x] `pytest` passes for the touched areas, and tests have been added under `tests/` (full suite runs in CI).
- [ ] I have read and complied with the project's AI Policy for any AI-assisted contributions.


music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        provider = self.mass.get_provider(provider_instance, provider_type=PlayerProvider)
+        if provider is None:
+            msg = f"Provider not found: {provider_instance}"
+            raise ProviderUnavailableError(msg)
+        provider.check_feature(ProviderFeature.SHOW_DASHBOARD)
+
+        remote_access = self.mass.webserver.remote_access
+        if not remote_access.is_enabled or not remote_access.remote_id:
+            msg = "Remote access must be enabled to cast dashboards"
```
marcelveldt: why is this ? can't we just send the frontend url to the device or at least support a reverse proxied scenario ?

music_assistant/models/player_provider.py
```diff
@@ -68,6 +70,40 @@ async def discover_players(self) -> None:
...
+    async def show_dashboard(
+        self, device_id: str, path: str, remote_id: str, dashboard_code: str
+    ) -> None:
+        """
+        Show a Music Assistant dashboard on a display device.
+
+        Only called for providers that declare the SHOW_DASHBOARD feature.
+
+        :param device_id: Provider-scoped device ID, as returned by `get_dashboard_devices`.
+        :param path: Frontend route to show on the display (e.g. "/party").
```
marcelveldt: See my comment above - this call should receive a URL
so not path, remote_id and dashboard_code

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,147 @@
...
+        return list(self._sessions.values())
+
+    @api_command("dashboard/show", required_scope=Scope.USERS_INVITE)
+    async def show_dashboard(self, provider_instance: str, device_id: str, path: str) -> None:
+        """
+        Show a Music Assistant dashboard on a display device.
+
+        :param provider_instance: Instance ID of the provider owning the display device.
+        :param device_id: Provider-scoped device ID, as returned by `dashboard/devices`.
+        :param path: Frontend route to show on the display (e.g. "/party").
```
marcelveldt: instead of a path I would limit this to a enum with a specific dashboard
then the controller will convert this to a full URL to the special dashboard with a access token

the underlying call at provider level should return the dashboard URL to open

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,208 @@
...
+        :param provider_instance: Instance ID of the provider owning the display device.
+        :param device_id: Provider-scoped device ID, as returned by `dashboard/devices`.
+        :param dashboard: Dashboard to show on the display.
+        :param player_id: Player to show, required when dashboard is NOW_PLAYING.
+        """
+        provider = self.mass.get_provider(provider_instance, provider_type=PlayerProvider)
+        if provider is None:
+            msg = f"Provider not found: {provider_instance}"
+            raise ProviderUnavailableError(msg)
+        provider.check_feature(ProviderFeature.SHOW_DASHBOARD)
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 117fd94 — `show_dashboard` now validates `provider.type == ProviderType.PLAYER` (cheap enum check) and raises `ProviderUnavailableError` for a non-player instance id, instead of relying on the runtime-unenforced `get_provider` type hint.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,208 @@
...
+        Hide a Music Assistant dashboard from a display device.
+
+        :param provider_instance: Instance ID of the provider owning the display device.
+        :param device_id: Provider-scoped device ID, as returned by `dashboard/devices`.
+        """
+        provider = self.mass.get_provider(provider_instance, provider_type=PlayerProvider)
+        if provider is None:
+            msg = f"Provider not found: {provider_instance}"
+            raise ProviderUnavailableError(msg)
+        provider.check_feature(ProviderFeature.SHOW_DASHBOARD)
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 117fd94 — same `provider.type == ProviderType.PLAYER` guard applied to `dashboard/hide`.

music_assistant/providers/chromecast/provider.py
```diff
@@ -273,3 +345,46 @@ async def _recheck_multichannel_child(self, player_id: str, disc_info: CastInfo)
...
+            msg = f"Unknown Cast device: {device_id}"
+            raise PlayerUnavailableError(msg) from err
+
+        def _connect() -> pychromecast.Chromecast:
+            """Create the Chromecast connection and wait for it to come up (blocking)."""
+            chromecast = pychromecast.get_chromecast_from_cast_info(
+                disc_info, self.mass.discovery.aiozc.zeroconf
+            )
+            chromecast.wait()
+            return chromecast
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 117fd94 — the on-demand connect now uses `chromecast.wait(timeout=DASHBOARD_CONNECT_TIMEOUT)` and raises `PlayerUnavailableError` if it doesn't come up, so an unreachable device can no longer stall `dashboard/show`.

music_assistant/controllers/dashboard/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+MusicAssistant DashboardController.
+
+Provides the core-level API for casting Music Assistant dashboards (e.g. Party
+mode, Quiz mode) to display devices, delegating the actual casting transport
+to whichever player provider owns the display device.
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 117fd94 — dropped the stale Quiz mode example; the docstring now lists only the supported Party / Now Playing dashboards.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,210 @@
...
+
+        url = await self._resolve_dashboard_url(dashboard, player_id)
+        await provider.show_dashboard(device_id=device_id, url=url)
+
+        device_name = device_id
+        for device in await provider.get_dashboard_devices():
+            if device.device_id == device_id:
+                device_name = device.name
+                break
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 72f4141 — the device name (and the dashboard url) are now resolved *before* `provider.show_dashboard`, so the only post-cast steps are the in-memory session write and the `DASHBOARD_SESSIONS_UPDATED` signal, which can't fail. A dashboard showing now always implies a tracked session.

OzGav:
Satisfies https://github.com/orgs/music-assistant/discussions/5181

music_assistant/providers/chromecast/provider.py
```diff
@@ -127,6 +145,63 @@ async def get_diagnostics(self) -> dict[str, SerializableType]:
...
+        try:
+            await self.mass.loop.run_in_executor(None, send_show_dashboard, chromecast, url)
+        except TimeoutError as err:
+            msg = f"Timed out launching app on {chromecast.name}"
+            raise PlayerUnavailableError(
+                msg,
+                translation_key="app_launch_timeout",
+                translation_owner=self.translation_owner,
+                translation_args=[chromecast.name],
+            ) from err
```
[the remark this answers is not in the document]
MarvinSchenkel: Connections don't actually accumulate here: `_dashboard_connections` is keyed by `device_id`, so a repeated failed `show_dashboard` to the same device reuses the cached entry (or replaces it if the socket dropped) — it's bounded to one per device, never stacking. It's also cleaned up by `hide_dashboard` and on `unload`.

The only real gap is that an on-demand connect that then times out on app launch leaves one live socket cached for a device showing nothing until the next hide/unload — bounded and rare. Worth noting the naive "disconnect on failure" is unsafe: `_get_or_create_chromecast` can return an *active player's* `cc`, so a blanket teardown would kill a live audio player. Leaving as-is for now given it's bounded and self-heals; will revisit with a `created`-flag guard if it proves noisy.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,214 @@
...
+            msg = f"Player provider not found: {provider_instance}"
+            raise ProviderUnavailableError(msg)
+        provider.check_feature(ProviderFeature.SHOW_DASHBOARD)
+
+        # resolve everything that can fail before the cast, so a failure never
+        # leaves a dashboard showing without a tracked session
+        device_name = await self._resolve_device_name(provider, device_id)
+        url = await self._resolve_dashboard_url(dashboard, player_id)
+        await provider.show_dashboard(device_id=device_id, url=url)
+        self._sessions[(provider_instance, device_id)] = DashboardSession(
```
marcelveldt: So DashBoardDevice should declare what type(s) of DashboardType it supports.
This way we can offer both a hosted experience with URL and custom experiences where the dashboard is built into a (native) app running on a device.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,214 @@
...
+
+    from music_assistant.mass import MusicAssistant
+
+DASHBOARD_VIEWER_USERNAME = "dashboard_viewer"
+DASHBOARD_VIEWER_DISPLAY_NAME = "Dashboard Viewer"
+DASHBOARD_CODE_EXPIRY_HOURS = 1
+APP_MA_HOST = "https://app.music-assistant.io"
+
+
+class DashboardController(CoreController):
```
marcelveldt, line 51: consider adding a register_dashboard command here so any api client can register itself as a dashboard endpoint - like for example a android/apple tv client.

you also need a "get_url_for_dashboard" helper in this controller for url-based dashboard viewers so they can retrieve the dashboard URL

music_assistant/models/player_provider.py
```diff
@@ -68,6 +70,36 @@ async def discover_players(self) -> None:
...
+        """
+        Hide a Music Assistant dashboard from a display device.
+
+        Only called for providers that declare the SHOW_DASHBOARD feature.
+        Must be a graceful no-op when no dashboard is currently showing.
+
+        :param device_id: Provider-scoped device ID, as returned by `get_dashboard_devices`.
+        """
+        raise NotImplementedError
+
```
marcelveldt: this does not belong in the provider model.
if an existing provider (e.g. cast) wants to implement this as a (bonus) feature, it just calls the "register_dashboard" on the dashboards controller

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,214 @@
...
+        self._signal_sessions_updated()
+
+    async def _resolve_device_name(self, provider: PlayerProvider, device_id: str) -> str:
+        """Return the display name for a device, falling back to its id."""
+        for device in await provider.get_dashboard_devices():
+            if device.device_id == device_id:
+                return device.name
+        return device_id
+
+    async def _resolve_dashboard_url(self, dashboard: DashboardType, player_id: str | None) -> str:
```
marcelveldt: so this will become a public function so any provider can use it to resolve a dashboard type into an url

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,214 @@
...
+    ) -> None:
+        """
+        Show a Music Assistant dashboard on a display device.
+
+        :param provider_instance: Instance ID of the provider owning the display device.
+        :param device_id: Provider-scoped device ID, as returned by `dashboard/devices`.
+        :param dashboard: Dashboard to show on the display.
+        :param player_id: Player to show, required when dashboard is NOW_PLAYING.
+        """
+        provider = self.mass.get_provider(provider_instance, provider_type=PlayerProvider)
```
marcelveldt: idea: instead of calling in on providers, we simply emit an event on the eventbus that a dashboard needs to be shown. This way it will also work from clients

music_assistant/providers/chromecast/dashboard.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def unregister(self, uuid: UUID) -> None:
+        """
+        Unregister a Cast device that is no longer discovered.
+
+        :param uuid: Cast device uuid, as reported by discovery.
+        """
+        if unregister_callback := self._unregister_callbacks.pop(str(uuid), None):
+            unregister_callback()
+
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 3ebbb26 — `unregister` now also pops and disconnects the cached on-demand connection (non-blocking `disconnect(0)`; the socket thread exits on its own). Safe as noted: `_dashboard_connections` never holds a live MA player connection.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,404 @@
...
+        :param player_id: Player to show, required when dashboard is NOW_PLAYING.
+        :raises InvalidCommand: If dashboard is NOW_PLAYING without a player_id, or unsupported.
+        """
+        if dashboard == DashboardType.PARTY:
+            return "/party"
+        if dashboard == DashboardType.NOW_PLAYING:
+            if not player_id:
+                msg = "player_id is required to show the now_playing dashboard"
+                raise InvalidCommand(msg)
+            return f"/now-playing?player={player_id}"
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 3ebbb26 — the now_playing route now builds its query with `urlencode`, so reserved characters in a player_id can no longer produce an ambiguous route.

music_assistant/providers/chromecast/dashboard.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            await self.mass.loop.run_in_executor(None, send_show_dashboard, chromecast, url)
+        except TimeoutError as err:
+            msg = f"Timed out launching app on {chromecast.name}"
+            raise PlayerUnavailableError(
+                msg,
+                translation_key="app_launch_timeout",
+                translation_owner=self.provider.translation_owner,
+                translation_args=[chromecast.name],
+            ) from err
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 3ebbb26 — the wrapper now preserves the helper's original message (device name + actual reason) and uses a new generic `show_dashboard_failed` translation key instead of mislabeling everything as a launch timeout. `app_launch_timeout` stays, it's still used by the play_media path.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,404 @@
...
+from music_assistant.helpers.api import api_command
+from music_assistant.helpers.guest_access import get_or_create_guest_user
+from music_assistant.models.core_controller import CoreController
+
+if TYPE_CHECKING:
+    from music_assistant.mass import MusicAssistant
+
+DASHBOARD_VIEWER_USERNAME = "dashboard_viewer"
+DASHBOARD_VIEWER_DISPLAY_NAME = "Dashboard Viewer"
+DASHBOARD_CODE_EXPIRY_HOURS = 1
```
marcelveldt, line 34: what happens if it expires ? is a dashboard viewer client able to ask for a refresh or a keep alive while it is consuming the dashboard ?

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,404 @@
...
+        icon: str | None = None,
+    ) -> None:
+        """
+        Register the calling client as a dashboard endpoint.
+
+        :param dashboard_id: Unique id chosen by the registering client.
+        :param name: Display name for the dashboard endpoint.
+        :param supported_types: Dashboard types this endpoint can show, defaults to all
+            when omitted; an explicitly empty set is rejected.
+        :param icon: Optional material design (mdi-*) or MA/lucide icon name.
```
marcelveldt: let's not do this - we are trying to move away from those static icon sets in backend definitions.
maybe a provider hint or something is better

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,404 @@
...
+            name=registration.device.name,
+            dashboard=dashboard,
+            player_id=player_id,
+        )
+
+        if registration.on_show is not None:
+            # resolve everything that can fail before invoking the callback, so a failure
+            # never leaves a dashboard showing without a tracked session
+            url = await self._resolve_dashboard_url(dashboard, player_id)
+            await registration.on_show(dashboard, url, player_id)
```
marcelveldt: this assumes that the on_show callback always needs the URL
I would flip that and let the consumer resolve the URL (that is also what a remote client would have to do) 
This way you leave flexibility for a client that wants to render a custom page based on the values

music_assistant/controllers/dashboard/controller.py
MarvinSchenkel: Another option is that only server side dashboard provider push an icon and we use a default (eg `tv-minimal`) for dashboards that register themselves via the api?

marcelveldt: I would just use a provider(hint) for this

music_assistant/controllers/dashboard/controller.py
MarvinSchenkel: This expiry is for the exchange token, the guest token is valid for 24. Should be long enough for casting.

music_assistant/providers/chromecast/dashboard.py
```diff
@@ -0,0 +1,195 @@
...
+            return
+
+        hidden = await self.mass.loop.run_in_executor(None, send_hide_dashboard, chromecast)
+        if not hidden:
+            self.logger.debug("No dashboard was showing on %s", chromecast.name)
+
+        if device_id in self._dashboard_connections:
+            del self._dashboard_connections[device_id]
+            await self.mass.loop.run_in_executor(None, chromecast.disconnect, 10)
+
```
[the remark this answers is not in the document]
MarvinSchenkel: Fixed in 3b4663f — `_on_hide` now disconnects only the popped on-demand connection object (`_dashboard_connections` only ever holds connections we opened on-demand, never a registered player's `cc`), so hiding can no longer tear down an active player's connection. Added a regression test covering a live player alongside a stale on-demand cache entry.

music_assistant/controllers/dashboard/controller.py
```diff
@@ -0,0 +1,401 @@
...
+
+        :param dashboard: Dashboard to load.
+        :param player_id: Player to show, required when dashboard is NOW_PLAYING.
+        :raises InsufficientPermissions: If the caller has neither the required scope
+            nor a matching active session of its own.
+        """
+        if not self._can_resolve_url_for_caller(dashboard, player_id):
+            msg = "Insufficient permissions to resolve a dashboard url"
+            raise InsufficientPermissions(msg)
+        return await self.resolve_dashboard_url(dashboard, player_id)
```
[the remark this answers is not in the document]
MarvinSchenkel: Intentional — authorization is checked before request validation on purpose, so we don't leak whether a request is valid to callers who aren't allowed to resolve a url in the first place. An authorized caller (USERS_INVITE, or an endpoint owner with a matching active session) always passes the permission check and then gets the proper `InvalidCommand` from `resolve_dashboard_url`. Leaving the order as-is.

marcelveldt (approved):
Nice!

PR #4898, opened 2026-07-22
fix(player_queues): reset elapsed_time with the item switch in play_index

teancom:

Tracked a bug in the mobile app where we were displaying the wrong time for a few seconds at the beginning of a track, to this.

Unlike next/previous, play_index left elapsed_time untouched when it set the new current_item, so queue updates paired the new item with the old item's elapsed time until the player reported position. Reset it to the start/seek position together with the item switch.

## Types of changes

## Checklist

### DIFF
--- a/tests/controllers/player_queues/test_play_index_elapsed.py
+++ b/tests/controllers/player_queues/test_play_index_elapsed.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import time
 from unittest.mock import AsyncMock, MagicMock, Mock
 
+from music_assistant_models.errors import MediaNotFoundError
 from music_assistant_models.player_queue import PlayerQueue
 from music_assistant_models.queue_item import QueueItem
 
@@ -47,6 +48,7 @@ def _controller_with_stale_queue() -> tuple[
     ctrl.player_media_from_queue_item = AsyncMock()  # type: ignore[method-assign]
     ctrl.mass = MagicMock()
     ctrl.mass.players.play_media = AsyncMock()
+    ctrl.logger = MagicMock()  # the retry-fallback path logs a warning
     return ctrl, queue, signals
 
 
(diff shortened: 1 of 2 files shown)


tests/controllers/player_queues/test_play_index_elapsed.py
```diff
@@ -0,0 +1,75 @@
...
+    await ctrl.play_index(QUEUE_ID, 1)
+
+    assert queue.current_item is not None
+    assert queue.current_item.queue_item_id == "new"
+    assert queue.elapsed_time == 0
+    assert queue.elapsed_time_last_updated >= before
+    # no signaled state may pair the new item with the previous item's elapsed
+    assert all(elapsed == 0 for item_id, elapsed in signals if item_id == "new"), signals
+
+
```
chrisuthe, line 69: One gap I noticed: both new tests have _load_item succeed on the first try, so attempt is always 0 in queue.elapsed_time = seek_position if attempt == 0 else 0. That means the retry branch — where the originally requested item fails to load and we fall back to the next one — never actually gets exercised.

That's worth a test on its own to me, because it's the case this line was written for: if you ask to resume item A at seek_position=45 and A turns out to be unplayable, we fall back to item B — and B has nothing to do with A's resume point, so elapsed_time should land on 0, not 45. Right now nothing would catch it if that ever regressed (e.g. someone "simplifies" the ternary to just queue.elapsed_time = seek_position).

```suggestion
async def test_play_index_retry_fallback_discards_failed_items_seek_position() -> None:
    """Falling back to another item after a load failure zeroes elapsed_time, not seek_position."""
    ctrl, queue, signals = _controller_with_stale_queue()
    fallback_item = QueueItem(
        queue_id=QUEUE_ID, queue_item_id="fallback", name="fallback", duration=200
    )
    ctrl._queue_data[QUEUE_ID].items.append(fallback_item)
    ctrl._load_item = AsyncMock(side_effect=[MediaNotFoundError("unplayable"), None])
    ctrl._get_next_index = Mock(return_value=2)

    await ctrl.play_index(QUEUE_ID, 1, seek_position=45)

    assert queue.current_item is not None
    assert queue.current_item.queue_item_id == "fallback"
    assert queue.elapsed_time == 0
    assert all(elapsed == 0 for item_id, elapsed in signals if item_id == "fallback"), signals
```
You'll need to import MediaNotFoundError for that, though as well.

chrisuthe (changes requested):
LGTM - One test I think could help catch in the future but not strictly speaking required.

tests/controllers/player_queues/test_play_index_elapsed.py
teancom: Good catch, thank you!

chrisuthe (approved):
Looking great!

PR #4902, opened 2026-07-22
Fix AirPlay 2 pairing credentials not persisted to live player config

Randalix:

Fixes #5895 — AirPlay 2 pairing to an LG webOS TV reported success, but the very next playback attempt still failed and `cliairplay` fell back to asking for interactive pairing again.

Root cause: `AirPlayPlayer._finish_pairing()` computes the HAP credentials and logs `"Finished AirPlay pairing for ..."`, but only stores them into the transient `values` dict returned by the (read-only, `CONFIG_PLAYERS_READ`-scoped) `config/players/get_entries` call. Nothing on that call path writes the value back into the player's live `PlayerConfig`, so `stream.py`'s `self.player.config.get_value(CONF_AIRPLAY_CREDENTIALS)` at play time finds nothing, no `--auth` is passed to `cliairplay`, and it asks for pairing again — even though pairing had just succeeded.

The sibling method `_reset_pairing()` a few lines below already persists correctly via `self.config.update(...)` for the clear-credentials case; `_finish_pairing()` was simply missing the equivalent call for the success case.

Fix: add the missing `self.config.update({cred_key: credentials})` call, mirroring `_reset_pairing`.

**Verified against real hardware** (LG OLED48C17LB, webOS 03.53.45, AirPlay 2 via the `hass_players` provider): before the fix, pairing followed by an immediate `play_media` reproduced the bug exactly (pairing succeeds, then `cliap2` asks for pairing again, second PIN prompt can't be answered — `"No active pairing session"`). With the fix applied, the same pair-then-play sequence streams successfully with no second pairing prompt.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/airplay/player.py
+++ b/music_assistant/providers/airplay/player.py
@@ -1011,9 +1011,15 @@ class AirPlayPlayer(Player):
         credentials = await self._active_pairing.finish_pairing(pin=str(pin))
         self._active_pairing = None
 
-        # Store credentials with the protocol-specific key
+        # Store credentials with the protocol-specific key. `values` only feeds the
+        # ConfigEntry response of this (read-only) config/players/get_entries call, so
+        # without also persisting to the live config here, stream.py's subsequent
+        # config.get_value(cred_key) lookup at play time finds nothing and cliairplay
+        # falls back to asking for interactive pairing again, even though pairing just
+        # succeeded (see _reset_pairing below, which already does this for its case).
         cred_key = self._get_credentials_key(protocol)
         values[cred_key] = credentials
+        self.config.update({cred_key: credentials})
 
         self.logger.info(f"Finished {protocol_name} pairing for {self.display_name}")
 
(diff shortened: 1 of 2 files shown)


Randalix:
Re-verified against the real LG webOS TV after switching from a raw `config.update()` to `save_player_config()`: paired, restarted the server, and confirmed the credentials survive the restart (encrypted in the persisted config) and playback starts immediately afterward with no second pairing prompt. Previously (with just the in-memory `config.update()` fix) the credentials were lost on every restart, which would have meant re-pairing after every nightly server restart in practice.

marcelveldt:
Did you actually test this on the latest dev branch because a lot has happened compared to the stable branch.
Because I can not reproduce your issue at all with the credentials not persisting. Besides that, current dev has a slightly changed AirPlay implementation in place.

marcelveldt:
Ah, now I think I know what is going on. Your LG TV requires pairing but it is wrapped as a universal player since AirPlay is just a protocol to this device. This is actually the first time I have seen a 3rd party device requiring pin pairing.

OzGav:
Also tests are failing.

marcelveldt:
Pushed a small fixup to get CI green: fixed the mypy errors in the new tests and compacted some comments/docstrings. The fix itself checks out - nice find and thanks for the thorough analysis!

marcelveldt (approved):
Thanks, @Randalix !

PR #4908, opened 2026-07-22
Fix ffmpeg stderr log flood on corrupted/malformed audio streams

chrisuthe:

An ffmpeg subprocess that logged 50 "Invalid data found when processing input" lines (a corrupted/malformed source file) started promoting every subsequent stderr line to `ERROR` for the rest of its life — one bad AAC file produced thousands of `ERROR` lines during background audio analysis.

Report the condition once instead of per-line, and abort the stream via a detached task (an inline `await self.close()` would make the stderr-reader task await itself, raising `RuntimeError`). Also:
- Reap the process before checking `returncode` in `get_ffmpeg_stream`/`get_ffmpeg_overlay_stream`, so a stream aborted mid-decode raises `AudioError` instead of silently truncating.
- Guard `AsyncProcess.close()`'s one unguarded `send_signal(SIGINT)` against `ProcessLookupError`, matching the SIGKILL delivery already guarded in the same method — making `close()` exception-safe for every caller that awaits it in a `finally`.

- related issue https://github.com/music-assistant/support/issues/5874

## Types of changes

## Checklist


music_assistant/helpers/ffmpeg.py
```diff
@@ -242,6 +255,16 @@ async def _log_reader_task(self) -> None:
                     self.logger.debug("Detected input duration: %s seconds", duration)
             del line
 
+    async def _abort_on_decode_errors(self) -> None:
+        """Close the process after excessive decode errors, swallowing any close() error."""
+        try:
+            await self.close()
+        except Exception as err:
```
OzGav: Can this be narrowed?

chrisuthe: Yes, But, this is the only process in asyncprocess.close that isn't wrapped with suppress. I think we should guard it there, which removes the erquirement to have any except clause to narrow... update inbound. Good catch!

marcelveldt (approved):
LGTM, Thanks @chrisuthe !

PR #4901, opened 2026-07-22
Remove the fixed output limiter

OzGav:
# What does this implement/fix?

Removes the fixed -2 dB limiter that was appended to every player's output filter chain, along with its per-player "Enable limiting to prevent clipping" setting.

**Related issue (if applicable):**

- related issue 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) — `bugfix`
- [ ] New feature (non-breaking change which adds functionality) — `new-feature`
- [ ] Enhancement to an existing feature — `enhancement`
- [ ] New music/player/metadata/plugin provider — `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) — `breaking-change`
- [ ] Refactor (no behaviour change) — `refactor`
- [ ] Documentation only — `documentation`
- [X] Maintenance / chore — `maintenance`
- [ ] CI / workflow change — `ci`
- [ ] Dependencies bump — `dependencies`

## Checklist

- [ ] The code change is tested and works locally.
- [X] `pre-commit run --all-files` passes.
- [X] `pytest` passes, and tests have been added/updated under `tests/` where applicable.
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.
- [X] I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.
- [ ] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate.

PR #4919, opened 2026-07-23
Add multi channel for local audio

vingerha:

Extends local audio to allow to play multi-channel. 

Key changes:
- helpers/ffmpeg.py : channel layouts are now named per channel count instead of collapsing everything to stereo; the source's channel count is also parsed back from FFmpeg's stream output.
- sendspin/playback : session PCM format no longer hardcodes stereo. It defaults to stereo unless the advertised count is concrete and within range >2 and <=8.
- local_audio/pa_simple.py : detect the device's channel order and reorder where needed on ALSA only, deliberately not on pulse otherwise the correction is applied twice
- local_audio/sendspin_bridge.py : BRIDGE_CHANNELS replaced by the device's real channel count 
- config: stereo remains 'default', one must actively select the multi-channel from the player>advanced>sendspin output channels (if capable)

Testing:
Has been executed on a Denon 3800h 7.1 setup (channel-test files for stereo, 3.0, 4.0-quad, 5.0, 5.1, 7.1) and WiiM (stereo) to verify if ffmpeg was still OK.

Open/not supported: DSP / EQ, this remains stereo only

## Types of changes

## Checklist


OzGav:
Make sure you run pre-commit before pushing. There are 190 errors that could have been auto corrected. 

The (AI) comments in the code need to be removed or completely slimmed down. Also make sure docstrings don't explain inner workings of the code

OzGav:
And did you look at this before you submitted it because there are 100s of blank lines in the code?! Marking as draft

vingerha:
> Make sure you run pre-commit before pushing. There are 190 errors that could have been auto corrected.
> 
> The (AI) comments in the code need to be removed or completely slimmed down. Also make sure docstrings don't explain inner workings of the code

Hi, yes, I am working on it.... I mix windows with linux and did not have a VSC locally so these 'errors' were a surprise, understandable but not expected....... and I am not the GH expert

vingerha:
@OzGav all fine now I hope, installed a local dev env to run the pre-commit and full pytest
Now the question is...is the code anywhere ok :)

EDIT: the branch is outdated every few minutes (logically), how often does this require to update?

music_assistant/providers/sendspin/playback.py
```diff
@@ -1245,19 +1249,20 @@ def _read_pipeline_config(self, player_id: str) -> _PipelineConfig:
...
+        if output_plan is not None:
+            if not requires_transform:
+                output_plan.output_details.dsp.output_limiter = False
+            if self._queue_id is not None and self._queue_session_id is not None:
+                self.player.mass.streams.audio_processing.update_output(
+                    output_plan.output_details.player_ids[0],
+                    output_plan,
+                    queue_id=self._queue_id,
+                    session_id=self._queue_session_id,
+                )
```
OzGav: This looks like a botched merge as it is reverting PR 4901

music_assistant/providers/sendspin/playback.py
```diff
@@ -1273,21 +1278,43 @@ def _select_session_pcm_formats(self) -> tuple[AudioFormat, SendspinAudioFormat]
...
+        # so a 5.1/7.1 client gets a session pipeline carrying its native
+        # layout. This is deliberately conservative: it defaults to stereo and
+        # only raises the count when the leader advertised a concrete, in-range
+        # (>2, <=_MAX_SESSION_CHANNELS) value. Critically, _get_member_output_
+        # format() falls back to returning self._pcm_format when the leader has
+        # no resolved sendspin role yet (e.g. at stream start); reading a
+        # channel count from that self-referential fallback is unsafe, so
+        # anything outside the strict multichannel range leaves the session at
+        # stereo — matching stock behavior — rather than propagating a stray
+        # count into the whole pipeline.
```
OzGav: We can't have essays as code comments

music_assistant/providers/sendspin/playback.py
```diff
@@ -1273,21 +1278,43 @@ def _select_session_pcm_formats(self) -> tuple[AudioFormat, SendspinAudioFormat]
         lossy — higher rates yield no perceivable quality gain there. Member
         clients with a different preferred rate up/down-sample in their own DSP
         step on the receiving side.
+
+        Channel count follows the leader's advertised capability (from its Sendspin
+        ClientHello) instead of being forced to stereo, so multichannel-capable
+        clients (e.g. a 5.1/7.1 ALSA or PipeWire sink) get a session pipeline that
+        can actually carry their native layout. Falls back to stereo when the
+        client hasn't advertised a channel count, and is clamped to a sane ceiling
+        so a malformed/hostile ClientHello can't blow up the pipeline.
```
OzGav: General comment throughout:

Use Sphinx-style docstrings with :param: syntax. For simple functions, a single-line docstring is fine. Multi line docstrings should have the first line on its own line to maintain indentation with the following. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/sendspin/playback.py
```diff
@@ -1273,21 +1278,43 @@ def _select_session_pcm_formats(self) -> tuple[AudioFormat, SendspinAudioFormat]
...
+        # format() falls back to returning self._pcm_format when the leader has
+        # no resolved sendspin role yet (e.g. at stream start); reading a
+        # channel count from that self-referential fallback is unsafe, so
+        # anything outside the strict multichannel range leaves the session at
+        # stereo — matching stock behavior — rather than propagating a stray
+        # count into the whole pipeline.
+        session_channels = 2
+        advertised = int(leader_output.channels) if leader_output.channels else 2
+        if 2 < advertised <= _MAX_SESSION_CHANNELS:
+            session_channels = advertised
```
OzGav: In a sync group, every member receives that same stream. There's no step anywhere that converts the audio to what each individual member can actually handle. Either convert per member, or restrict multichannel to ungrouped players until the grouped case is demonstrated.

music_assistant/helpers/ffmpeg.py
```diff
@@ -30,6 +30,24 @@
...
+    3: "2.1",
+    4: "quad",
+    5: "5.0",
+    6: "5.1",
+    7: "6.1",
+    8: "7.1",
+}
+
+
+def _channel_layout(channels: int) -> str:
```
OzGav, line 46: This silently falls back to "stereo" for unmapped counts, which would emit contradictory args like `-ac 9 -channel_layout stereo`. Fail loudly or omit `-channel_layout` instead.

music_assistant/helpers/ffmpeg.py
```diff
@@ -30,6 +30,24 @@
...
+_CHANNEL_LAYOUT: Final[dict[int, str]] = {
+    1: "mono",
+    2: "stereo",
+    3: "2.1",
+    4: "quad",
+    5: "5.0",
+    6: "5.1",
+    7: "6.1",
+    8: "7.1",
+}
```
OzGav, line 43: This table must stay in sync with `_SOURCE_CHANNEL_ORDER` in `pa_simple.py` as one writes the channel order, the other interprets it. Nothing enforces that today, and a mismatch plays channels on the wrong speakers with no error. Derive both from one definition or add a test asserting agreement, and cover the same channel counts in both (3/4/5/7 are currently only in this one).

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -3,12 +3,17 @@
...
+import logging
 import os
 import threading
 from typing import Any, ClassVar, Final, Self
 
+import numpy as np
+
 from .constants import volume_pct_to_amplitude
 
+LOGGER = logging.getLogger("local_audio.pa_simple")
```
OzGav: [PROBLEM] Don't invent a standalone logger here "local_audio.pa_simple" sits outside the music_assistant logger hierarchy, so the log level a user sets for this provider in MA has no effect on it, and all the new debug logging in this file won't follow the provider's settings. The rest of this provider does it properly: `sendspin_bridge.py` uses `provider.logger.getChild(...)`. Either pass the provider's logger into these functions, or name the logger under the provider's hierarchy so it inherits the configured level.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -42,6 +47,65 @@ class _PASampleSpec(ctypes.Structure):
...
+    client's *defaulted* channel map with PA_ERR_INVALID, while accepting
+    the byte-identical spec with an explicit map; (2) declaring the data's
+    true slot order (MA/FFmpeg order, from _SOURCE_CHANNEL_ORDER) makes the
+    server itself reorder by position name to the sink's map — hardware
+    channel order is then correct by construction on this backend, with no
+    client-side byte remapping.
+
+    The map is built via libpulse's own pa_channel_map_parse from position
+    names rather than hardcoded enum integers, so the numeric positions are
+    always whatever the linked libpulse defines.
```
OzGav: Delete

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +787,380 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
         )
+
     return sinks
 
 
+# The channel order MA/FFmpeg actually writes PCM in for each channel count
+# it produces (matches helpers/ffmpeg.py's _CHANNEL_LAYOUT scope — only
+# channel counts with an unambiguous standard layout are covered; anything
+# else has no single correct "default" order to remap from).
+_SOURCE_CHANNEL_ORDER: Final[dict[int, list[str]]] = {
```
OzGav, line 760: What about 3/4/5/7 ?

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +787,380 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+# Driver-reported positions with no counterpart in _SOURCE_CHANNEL_ORDER's
+# vocabulary, aliased to the source position whose content they should carry.
+# See the comment in build_channel_remap_index for the reasoning.
+_PHYSICAL_POSITION_ALIASES: Final[dict[str, str]] = {
+    "rear-left-of-center": "side-left",
+    "rear-right-of-center": "side-right",
+}
+
+
+def build_channel_remap_index(
```
OzGav, line 797: No tests cover this

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +787,380 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+        "Remap computed for %d-channel device: source %s -> physical %s (index %s)",
+        channels,
+        source_order,
+        physical_channel_map,
+        index,
+    )
+    return index
+
+
+def remap_pcm_channels(
```
OzGav, line 872: No tests cover this

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +787,380 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+    """
+    frame_bytes = channels * bytes_per_sample
+    # Truncate to a whole number of frames — a partial trailing frame (can
+    # happen at a stream boundary) can't be meaningfully reordered and is
+    # dropped rather than corrupted; callers write in frame-aligned chunks
+    # in normal operation, so this is a no-op in practice.
+    usable_len = (len(data) // frame_bytes) * frame_bytes
+    arr = np.frombuffer(data[:usable_len], dtype=np.uint8).reshape(-1, channels, bytes_per_sample)
+    remapped = arr[:, index, :]
+    return remapped.tobytes() + data[usable_len:]
```
OzGav, line 893: Comment says partially trailing frame is dropped?

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -129,15 +157,43 @@ def __init__(
...
-        # doesn't change. This is independent of BRIDGE_CHANNELS, which is
-        # the audio *stream's* channel count (always 2).
+        # remap sink). Drives both the advertised Sendspin format and the
+        # actual PCM stream opened against the sink, and is used for
+        # pa_cvolume_set in PAVolumeController calls — a mismatch against the
+        # sink's real channel_map can leave the displayed reference_volume
+        # updated while soft_volume (real gain) doesn't change.
         self.pa_channels: int = device_info.get("max_output_channels", BRIDGE_CHANNELS)
+        # The device's real physical channel order (from pa_simple's
+        # get_card_port_physical_positions(), EDID-detected where
```
OzGav: This function doesnt exist

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -661,6 +763,9 @@ async def _audio_writer_pulse(self) -> None:
                     data = self._apply_format_conversion(data)
                 else:
                     data = self._apply_software_volume(data)
+                data = self._remap_channels(
```
OzGav: This _remap_channels call in _audio_writer_pulse is dead code: `_channel_remap_index` is explicitly None when `backend == "pulse"` (line 183). If it ever did activate, `self.bit_depth // 8` is wrong after the 24-bit repack at line 763 (data is 3 bytes/sample by then). Remove the call.

OzGav (changes requested):
This approach seems viable and worth approving eventually, but as submitted it isn't ready to merge. To get it there I believe it needs:

- An opt-in config toggle because as it stands, every multichannel device starts receiving all content (including plain stereo) as 8-channel PCM on upgrade, with no way to turn it off.
- A defined answer for sync groups, even if that answer is "multichannel disables when grouped". Right now every group member receives the leader's channel count and nothing converts it.
- The device and session channel counts clamped consistently (shared rule, one place). Devices advertising 1 or >8 channels currently get their stream opened at a different count than the session delivers, and 3/4/5/7-channel devices have no channel-order handling.
- The `playback.py` merge redone as lines 1252–1265 resurrect code that #4901 deliberately removed.
- The files cleaned up

Happy to see this land once those are addressed — multichannel output to an AVR is a genuinely valuable capability.

vingerha:
Thanks for the quick elaborate feedback :)   
 I can work on most but for the opt-out.... From what I found is that the stereo/left/right/mono are defined for all (!) players via the constants. If I add multi-channel to that, then all players will show this option but most will not be able to play mc... any design help welcome :)

music_assistant/providers/sendspin/playback.py
vingerha: addressed, the dev updates faster than lightning

music_assistant/providers/sendspin/playback.py
vingerha: clear, I did a lot of testing before finding out my Denon was a channel order mishap

music_assistant/providers/sendspin/playback.py
vingerha: I will change the comment

music_assistant/providers/sendspin/playback.py
vingerha: I will restrict it, converting per member can be a next PR

music_assistant/helpers/ffmpeg.py
vingerha: clear

music_assistant/helpers/ffmpeg.py
vingerha: addressed, I hope it is correct, you must get your checks from somewhere too

music_assistant/providers/local_audio/pa_simple.py
vingerha: Open...t.b.d.

music_assistant/providers/local_audio/pa_simple.py
vingerha: will be removed

music_assistant/providers/local_audio/pa_simple.py
vingerha: comment altered

music_assistant/providers/local_audio/pa_simple.py
vingerha: adding a test, and fully admitting this is claude input ony, I understan what I read but not 100% sure if this is how it should be

music_assistant/providers/local_audio/pa_simple.py
vingerha: adding a test, and fully admitting this is claude input ony, I understan what I read but not 100% sure if this is how it should be

music_assistant/providers/local_audio/pa_simple.py
vingerha: altered,

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: indeed, part of issues with Denon, see above......changed comment

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: removed

vingerha:
@OzGav if you run some AI/check/etc. that reveals issues to be covered, kindly let me know if I can run these too, reducing risks on repeated comments :)

music_assistant/providers/local_audio/pa_simple.py
vingerha: used injection and simplified logging a bit

vingerha:
> * An opt-in config toggle because as it stands, every multichannel device starts receiving all content (including plain stereo) as 8-channel PCM on upgrade, with no way to turn it off.

Addressed, see separate commit, Where I do see the use, tbh it is quite a bit of code for this excpetional channel mode.

> * A defined answer for sync groups, even if that answer is "multichannel disables when grouped". Right now every group member receives the leader's channel count and nothing converts it.

Addressed

> * The device and session channel counts clamped consistently (shared rule, one place). Devices advertising 1 or >8 channels currently get their stream opened at a different count than the session delivers, and 3/4/5/7-channel devices have no channel-order handling.

Addressed

> * The `playback.py` merge redone as lines 1252–1265 resurrect code that #4901 deliberately removed.
> * The files cleaned up

both addressed as well as other comment

Repeating thanks for the earlier speedy comment/feedback :)

OzGav:
I or someone else will look at this again soon but in answer to your question about AI review assistance we have this which provides some code style requirements and explanations https://github.com/music-assistant/server/blob/dev/AGENTS.md 

Then I would ensure your code matches the patterns seen elsewhere in the code. Dont let the AI guess. Make sure the code remains simple and maintainable and there is nothing that could be considered a hack.

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -125,19 +161,69 @@ def __init__(
...
+            get_device_uuid(self.device_name, device_info.get("hostapi", 0))
+        )
+        self.pa_channels: int = (
+            self.device_channels
+            if self.mass.config.get_raw_player_config_value(
+                _config_client_id, CONF_OUTPUT_CHANNELS, "stereo"
+            )
+            == OUTPUT_CHANNELS_MULTICHANNEL
+            else BRIDGE_CHANNELS
+        )
```
OzGav: Changing the toggle has no effect until the provider is reloaded as the bridge reads the setting once at construction and its advertised formats are fixed from then on. Users would expect this dropdown to apply on next play, so this will be reported as broken. Rebuild the bridge on config change (`player.on_config_updated()` fires on every save), or at minimum make the UI say a reload is required.

vingerha:
> I or someone else will look at this again soon but in answer to your question about AI review assistance we have this which provides some code style requirements and explanations https://github.com/music-assistant/server/blob/dev/AGENTS.md
> 
> Then I would ensure your code matches the patterns seen elsewhere in the code. Dont let the AI guess. Make sure the code remains simple and maintainable and there is nothing that could be considered a hack.

I had a look at the code and frm what I can see it is in line, the only thing is that I donot know how to write test stuff and I am not 100% sure if what Claude provided is good enough, it looks good and covers what I expect but ... just saying that testing is rather new to me :)

music_assistant/providers/sendspin/playback.py
```diff
@@ -446,6 +448,19 @@ async def add_member(self, player_id: str) -> None:
         async with self._state_lock:
             if player_id in self._members:
                 return
+            if self._pcm_format.channels > 2:
+                # The session PCM format is fixed at stream start and every
+                # member receives that same stream, with no per-member
+                # conversion. Admitting a member to a multichannel session
+                # would push a layout it may be unable to render, so the join
+                # is refused while multichannel is active.
+                self.player.logger.warning(
```
OzGav: This refusal is silent and it loops. `sync_members` reconciles the session to the target member set, so a refused player is still missing on the next pass, gets re-added, and gets refused again, repeating the warning forever. From the user's side they add the player to the group, the UI shows it as a member, it plays nothing, and the only clue is a server log. Either show the refusal where the user can see it by failing the group command itself, or do not refuse at all and instead fall back to a stereo session when a member joins. That would match what already happens when the group exists before playback starts.

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -125,19 +161,69 @@ def __init__(
...
+        # updated while soft_volume (real gain) doesn't change.
+        # What the device can do, clamped by the same rule the session format
+        # uses so the two can never disagree.
+        self.device_channels: int = resolve_channel_count(
+            device_info.get("max_output_channels", BRIDGE_CHANNELS)
+        )
+        # What this bridge actually runs at. Multichannel is opt-in and off by
+        # default, so nothing changes on upgrade. The toggle lives on the protocol
+        # (Sendspin) player, which is where the device's advertised capability is
+
```
OzGav:
```suggestion
```

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +778,396 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+    the target channel count — the map the driver will *actually* use for
+    playback, which for HD Audio HDMI is negotiated from the sink's ELD at
+    prepare time. This is the API speaker-test derives its printed
+    "N - Position Name" labels from, so the result matches what listening
+    tests verify.
+
+    snd_pcm_query_chmaps() (the *available* maps list) is only a fallback:
+    on real hardware its first entry has been observed to be the generic
+    CEA allocation order rather than the active map — using it reproduced
+    an incorrect remap identical to EDID-derived data. Available ≠ active.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/local_audio/pa_simple.py
OzGav: This is still here.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -436,6 +496,13 @@ def __init__(
             rate=rate,
             channels=channels,
         )
+        # Explicit channel map declaring the data's slot order (MA/FFmpeg
+        # order). Required: the pulse server has been observed to reject
+        # multichannel streams that rely on the client's defaulted map
+        # (PA_ERR_INVALID at 8ch) while accepting the same spec with an
+        # explicit map — and declaring the true order makes the server
+        # route channels by position name to the sink's own map.
```
OzGav: Please trim this down

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -547,6 +614,21 @@ def enumerate_alsa_devices() -> list[dict[str, Any]]:
 
         description = _re.sub(r"\s*\(hw:\d+,\d+\)$", "", name).strip()
 
+        # Extract "hw:C,D" so the device can be matched against pactl's
+        # card-port EDID data by card index, and so it can be opened
+        # directly for a chmap query (query_alsa_chmap) even though this
+        # backend never opens a PA sink object — PipeWire still enumerates
+        # the card's ports as long as it's running alongside, regardless of
+        # who actually writes audio.
```
OzGav: I dont think this is correct information?

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +778,396 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+    remapped = arr[:, index, :]
+    return remapped.tobytes() + data[usable_len:]
+
+
+# ALSA channel position enum values (alsa/pcm.h SND_CHMAP_*), verified
+# directly against libasound.so.2's snd_pcm_chmap_name()/
+# snd_pcm_chmap_long_name() output rather than assumed from documentation —
+# confirmed e.g. 14="RLC"/"Rear Left Center", matching speaker-test's own
+# printed labels exactly. Only positions a real consumer device can
+# plausibly report are included.
```
OzGav: Dont explain what you did in the code, that can go in the PR description if at all. So slim this down.

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +778,396 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+    """
+    Build a channel-reorder index mapping standard PCM order to device order.
+
+    Remaps MA's standard PCM channel order onto the device's real physical
+    channel order.
+
+    Backend-agnostic: works identically whether the caller then applies it
+    to bytes headed for a PulseAudio/PipeWire sink or a raw ALSA hw: device
+    via PortAudio — this only computes *which source channel goes in which
+    output slot*, independent of how those bytes get written.
```
OzGav: No how it works comments

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +778,396 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+
+    :param channels: Number of channels in the PCM stream being written.
+    :param physical_channel_map: The device's real channel order (from
+
```
OzGav:
```suggestion
```

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -660,11 +778,396 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
...
+        logger.debug("No remap for %d-channel device: physical order unknown", channels)
+        return None
+    # Translate driver-reported positions that have no counterpart in the
+    # FFmpeg source vocabulary. HD Audio HDMI devices report the pair beyond 5.1
+    # as rear-left/right-of-center (CEA back-pair naming); FFmpeg 7.1 has no
+    # such positions. Pair them with the side channels so that the chmap's
+    # rear-left/right (matching names) carry the source's rear content and
+    # RLC/RRC carry the side content. For 5.1-in-7.1 sources the side pair
+    # is silent, so this pairing choice is inaudible there; it only matters
+    # for true 7.1 sources, where CEA semantics make it a defensible default.
```
OzGav: Trim this down. What does the future dev need to know?

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -76,6 +87,31 @@
...
+    Return the first frame in `data` that carries signal.
+
+    Returns the first frame where any sample exceeds the silence threshold,
+    or the very first frame if the whole scanned range is silent.
+
+    Diagnostic-only: chunks commonly start with a silent lead-in (encoder
+    priming, or genuine silence in the source), which makes logging frame 0
+    useless for verifying channel content/order. Scans up to the first 2000
+    frames (a few hundred ms) rather than the whole chunk, since this is
+    only for a one-time log line, not the audio path itself.
```
OzGav, line 102: There should not be as many lines in the docstring as there are in the function...

music_assistant/providers/local_audio/sendspin_bridge.py
```diff
@@ -125,19 +161,69 @@ def __init__(
...
+        # uses so the two can never disagree.
+        self.device_channels: int = resolve_channel_count(
+            device_info.get("max_output_channels", BRIDGE_CHANNELS)
+        )
+        # What this bridge actually runs at. Multichannel is opt-in and off by
+        # default, so nothing changes on upgrade. The toggle lives on the protocol
+        # (Sendspin) player, which is where the device's advertised capability is
+
+        # visible. The client id is derived here rather than reused from start():
+        # this read runs during construction, before start() assigns _device_uuid.
```
OzGav: Slim this down to two lines. "Multichannel is opt-in and off by default, so nothing changes on upgrade. Toggle lives on the sendspin protocol player."

OzGav (changes requested):
Just some more tidy up., Also the tests were reviewed and are solid, thanks for the honesty when you weren't sure.

music_assistant/providers/sendspin/playback.py
vingerha: OK, I just removed the whole bit and then it plays stereo fine. What I noticed is that when adding members to a group, those new members are not recognised without a restart, this is also in the 'stock' nightly so I will not fix that ... at least not in this PR :)

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: Got it working, I will update the PR later today when I have addressed the other comments.

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: so pre-commit does not always remove unnecessary empty lines :)

music_assistant/providers/local_audio/pa_simple.py
vingerha: will remove it, the comment/help can be found in the _build_pa_channel_map

music_assistant/providers/local_audio/pa_simple.py
vingerha: simplified/improved

music_assistant/providers/local_audio/pa_simple.py
vingerha: simplified/improved

music_assistant/providers/local_audio/pa_simple.py
vingerha: simplified/improved

music_assistant/providers/local_audio/pa_simple.py
vingerha: simplified/improved

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: simplified/improved

music_assistant/providers/local_audio/sendspin_bridge.py
vingerha: simplified/improved

music_assistant/providers/local_audio/pa_simple.py
vingerha: simplified/improved

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -547,6 +604,17 @@ def enumerate_alsa_devices() -> list[dict[str, Any]]:
 
         description = _re.sub(r"\s*\(hw:\d+,\d+\)$", "", name).strip()
 
+        # Extract "hw:C,D" for a direct chmap query (query_alsa_chmap) and
+        # keep the card index for later matching.
+        alsa_card_index: int | None = None
```
OzGav: This isn't used anywhere save for one log line?

music_assistant/providers/local_audio/pa_simple.py
```diff
@@ -618,6 +712,16 @@ def enumerate_pa_sinks() -> list[dict[str, Any]]:
         master_device: str | None = properties.get("device.master_device")
         is_remap = master_device is not None or driver == "module-remap-sink.c"
         alsa_card_name: str | None = properties.get("alsa.card_name")
+        active_port: str | None = sink.get("active_port")
```
OzGav: This isn't used anywhere?

OzGav (commented):
Nice. Just two minor points from me. This will get another review from Marvin or Marcel

music_assistant/providers/local_audio/pa_simple.py
vingerha: indeed, for log only. I felt it important to know this bit when things do not work as expected

EDIT, I think I can resolve it via the hw string ... 

EDIT2: yes..alsa_hw_string provides similar, updating pa_simple now

music_assistant/providers/local_audio/pa_simple.py
vingerha: yep a remnant from when I struggled with the Denon channels, removing it

iVolt1:
@vingerha Good to see you moving this PR forward. I just saw this PR yesterday and have one concern regarding multiple surround cards.

I submitted #5132 two days ago to fix some issues with the local_audio provider currently on dev, mainly duplicate sound card handling. This probably isn't much of an issue for HDMI, but it does affect analog surround cards — in my case an HDA surround instance and two identical Creative X-Fi cards, as the screenshots below show.

On your branch, the HDA surround instance isn't picked up at all, and only one of the two X-Fi surround instances is recognized. When I played a 7.1 test file to the Creative X-Fi Analog Surround 7.1 instance on your branch, all channels came out the front pair.

The grouping or not grouping configuration I am still trying to wrap my head around so I hope that is well documented for users.

Your branch players:

image


My branch players: 


image


@OzGav Assuming you want both PRs to move forward, then one of the PRs will need to be rebased. 

Full disclosure, I previously developed a working multichannel surround player provider that was simple, separate from the local_audio provider, not for grouping, and dependent on the local_audio player provider to supply the correct remap sinks for analog cards.  It worked fine on my analog surround cards with multichannel surround flac sources and although I never tested on the hdmi output, I think it would work or could be made to work.  I thought the separate player provider was an easier way to for a user to setup surround by only selecting the card to be used and for most users surround would be a dedicated listening experience anyway and not for whole house audio.  Due to time constraints I never created a PR for the multichannel player provider as I was planning to get to it after #5132 moved to dev.  Integrating surround into the local audio player provider is probably the better alternative, but it may take a little more effort for a user to configure and know what to expect with surround output. Thanks.

vingerha:
> On your branch, the HDA surround instance isn't picked up at all,

? except for the double sound cards.. .I see 4 HD_Audio on 'my' branch too

> When I played a 7.1 test file to the Creative X-Fi Analog Surround 7.1 instance on your branch, all channels came out the front pair.

Did you set the player to use multi-channel? Default it is stereo

> Grouping

If grouped: only stereo comes out, I was only able to test this with my laptop and that worked fine enough

> #5132

No clue how the maintainers look at that but I would guess that they first want that one done so I will likely rebase afterwards

> Documentation

I wait for signals from Marcel or Marvin. As soon as they see a reasonably short-ish path to acceptance then I will update the documentation.
Keeping this alive and in sync/rebased is an effort and I want to be careful spending more time with this after fixing OzGav'sreview. Tbh, without the proper signals I am not sure if this ends up on the backburner.

vingerha:
@iVolt1 
Re on misisng cards, I believe I misread it before, the fact that you see multiple cards is due to the PR5132, not ? Then that explains it I guess, and then also repeating that I likely need to rebase posterior

And 'your'branch seems to miss the USB one...but that could also be that it was 'down' when you activated your container

iVolt1:
> @iVolt1 Re on missng cards, I believe I misread it before, the fact that you see multiple cards is due to the PR5132, not ? Then that explains it I guess, and then also repeating that I likely need to rebase posterior

That is the reason.
> 
> And 'your'branch seems to miss the USB one...but that could also be that it was 'down' when you activated your container

That is true.  I had removed it at that point.  That is a USB 7.1 device that I will change the ha audio profile to 7.1 from the default stereo to also test with later. 


This may have been a mistake on my part due to what was connected to the sound cards, so please disregard for now:
```
    When I played a 7.1 test file to the Creative X-Fi Analog Surround 7.1 instance on your branch, all channels came out the front pair.

Did you set the player to use multi-channel? Default it is stereo
```

I think what I would want to occur is a distinct local audio sendspin player with "surround" in the name to differentiate it from the stereo pairs and multichannel stereo players sort of like like there is for Creative X-Fi Analog Surround 7.1 in your list but with the card number in the name e.g Creative_X-Fi_card0_Analog_Surround_7.1.  The surround 'cards' could possibly default to the max channel layout since that is really all they need to be used for as the stereo pairs and multichannel stereo instances cover everything else.

vingerha:
> I think what I would want to occur is a distinct local audio sendspin player with "surround" in the name to differentiate it from the stereo pairs and multichannel stereo players sort of like like there is for Creative X-Fi Analog Surround 7.1 in your list but with the card number in the name e.g Creative_X-Fi_card0_Analog_Surround_7.1. The surround 'cards' could possibly default to the max channel layout since that is really all they need to be used for as the stereo pairs and multichannel stereo instances cover everything else.

To have default stereo was a request bu OzGav ... so ... 
EDIT: this had a good reason but we can review this when/after this goes live, it should be as minimal as posisble and also that has proven to be not that small :)

OzGav:
@iVolt1 thanks for testing on real hardware with multiple cards, that is a setup neither vingerha nor I can test. Glad the front pair result turned out to be expected behaviour. Your duplicate card findings confirm the two PRs affect each other. And I agree with your own conclusion that building this into local_audio is better than a separate provider, one version is easier to maintain than two. Testing the dual card setup again in the future would be great.

@vingerha thanks for the sustained effort here, every review point has been addressed properly.

On the stereo default, keeping it was deliberate and I would argue for it staying. If multichannel were the default, everyone with a surround capable device would have all their music, including normal stereo music, sent as 8 channel audio the day they upgrade, without touching anything. That works against the receiver doing its own surround processing, uses four times the bandwidth for no benefit, and silently changes behaviour for people who never asked for it. Keeping it off by default also means the new code only runs for users who chose it. @iVolt1 your idea of separate surround players that default to full surround is worth a look, but only after both PRs are in, so let's make that a follow-up issue rather than grow this one.

My suggested way ahead, with the final call sitting with Marvin and Marcel:

[#5132](https://github.com/music-assistant/server/issues/5132) goes in first, it is a bugfix and this PR builds on correct card detection.
[#4919](https://github.com/music-assistant/server/issues/4919) then rebases on top. That needs care, since player identities come from device names and #5132 changes how duplicate cards are named.

@vingerha since you asked for a clear signal: from my side the review points are done, so after the rebase this is ready for Marvin and Marcel to look at. I will flag it with them. The work has been good and it shows.

vingerha:
> @vingerha since you asked for a clear signal: from my side the review points are done, so after the rebase this is ready for Marvin and Marcel to look at. I will flag it with them. The work has been good and it shows.

As I also expected on the prio for the other PI and super, thanks... there is still hope :) After rebasing on https://github.com/music-assistant/server/issues/5132 I hope that iVolt can test this as indeed... not a chance that I will be able to install or simulate such a piece of hardware :)

iVolt1:
I tested @vingerha's branch last night and was able to get some good results.  

Playing a 7.1 test file to the ICUSBAUDIO7DICUSBAUDIO7D_multichannel_stereo player player with multichannel configuration enabled played to all 8 channels with the correct channel identifying words.  I played a 7.1 flac track and it played to the correct channels.  

I played the same track in 5.1 format and it played on all channels except surround as expected.  I was able to join in a stereo player and seemed to play with all 7.1 or 5.1 channels mixed correctly while ICUSBAUDIO7D_multichannel_stereo player continued playing to the correct channels. 

Setting ICUSBAUDIO7D_multichannel_stereo back to stereo makes the 71. or 5.1 play what sounds like mixed to all channels.  @vingerha I assume your ffmpeg work does that.

This is all very positive.  The only thing I propose would be in addition to the *_stereo_pair remap players and the *_multichannel_stereo remap players for analog multichannel cards, there be another remap player suffixed *_analog_surround or *_surround that would default to multichannel in the player sendspin configuration while the *_multichannel_stereo remap players continue to default to two channel.  This would save users a lot of clicks changing configurations as well as seeing what the options for using the players are.  If this sounds like a good approach, I can work with my AI to get this into my current PR.

vingerha:
Thanks for testing, you are only the second one that was able to confirm it works :)

> Setting ICUSBAUDIO7D_multichannel_stereo back to stereo makes the 71. or 5.1 play what sounds like mixed to all channels. @vingerha I assume your ffmpeg work does that.

You mean it does play stereo but mixes the back/side/etc. into the front left/right? What I did to test is using flac files that 'tell' you the speaker it outputs to and setting the player on stereo the back/side did not get passed onto the front. I tested active switching too so am not sure what effect you are hearing. Do you have sample file(s) ?

> The only thing I propose would be in addition to the *_stereo_pair remap players and the *_multichannel_stereo remap players for analog multichannel cards, there be another remap player suffixed *_analog_surround or *_surround that would default to multichannel in the player sendspin configuration while the *_multichannel_stereo remap players continue to default to two channel. 

As mentioned by @OzGav, default is supposed to be stereo so he would need to review this. 
But, even if this is OK, I would prefer this for a next PR, adding things that dynamically/conditionally set 2<>more channels does not make easier as the playerdevice may change of channel count and/or order too, my Denon did/does weird things and I can not verify more devices so I prefer to first see if the bae version is able to work with multiple/different hdmi output devices before adding on top.

iVolt1:
> As mentioned by @OzGav, default is supposed to be stereo so he would need to review this. But, even if this is OK, I would prefer this for a next PR, adding things that dynamically/conditionally set 2<>more channels does not make easier as the playerdevice may change of channel count and/or order too, my Denon did/does weird things and I can not verify more devices so I prefer to first see if the bae version is able to work with multiple/different hdmi output devices before adding on top.

The proposed  *_analog_surround or *_surround players I am referring to would be for analog multichannel cards only but maybe that would be appropriate for hdmi also.  I am hoping to get an hdmi cable hooked up from my HA/MA box to my Onkyo SR507 receiver today to test the hdmi side of things.

vingerha:
> The proposed *_analog_surround or *_surround players I am referring to would be for analog multichannel cards

At the moment, it is all stereo for all devices by default. If a device is detected as being capable of multi-channel then th eoption to change the channel mode shows multi-channel and between brackets the number of channels. As I have only one mutltichannel device being my Denon via hdmi, can you please confirm this works for the analog ones ?

chrisuthe:
First off — nice work, this clearly took some real digging. The chmap querying in particular is the right instinct: asking the driver instead of guessing, and refusing to act when the answer doesn't line up.

A couple of easy asks before the bigger thought:

**Can we pull the `volume_control: none` fix out into its own PR?** It looks correct, it's just unrelated to multichannel, and it's the kind of thing that probably wants a `bugfix` + `backport-to-stable` label of its own. Same for the README line that goes with it. Just simplifies what's left here.

---

## The bigger thought: I'd really like to fix the spec first

Right now we're blind-matching channels in a few different places, and the tables don't even agree with each other. 5-channel audio is `4.1` according to aiosendspin's `resolve_av_format()` and `5.0` according to `_CHANNEL_LAYOUT` in this PR — different repos, so nothing catches it.

And because we only ever pass a *count*, we lose things that a count can't express. 3-channel L/R/C, which was used a good bit in mid 50's recording, is always assumed to be `2.1`, in both tables. ffmpeg says `3.0 = FL+FR+FC` and `2.1 = FL+FR+LFE`, so that centre channel goes straight to my subwoofer. There's no way to tell it apart from the count alone, and `_LAYOUT_TO_CHANNELS` maps both names back to `3` anyway, so we can't recover it on the parse side either.

So my strong preference would be to amend the Sendspin spec to carry the channel identity and order, rather than building more inference on top of a bare integer:

```python
class SupportedAudioFormat:      # the hello: "my speakers are wired this way"
    channels: int
    channel_map: list[str] | None = None      # ["FL","FR","FC","LFE","RLC","RRC","BL","BR"]

class AudioFormat:               # the stream: "these bytes are in this order"
    channels: int
    channel_map: list[str] | None = None      # ["FL","FR","FC","LFE","BL","BR","SL","SR"]
```

Both ends, deliberately — once the device's wiring *and* the stream's order are both declared, the mismatch is computable and we can decide explicitly who fixes it. That's basically what `pa_channel_map` already does, and it's why the PulseAudio path in this PR is clean while the ALSA path needs ~350 lines of ctypes and a numpy shuffle on every chunk. Give every client a channel map and they all get the pulse path's ergonomics.

We'd need to agree a vocabulary — PulseAudio names, ALSA `SND_CHMAP_*` and ffmpeg tokens don't map 1:1 (this PR already had to alias HD Audio's `rear-left-of-center`/`rear-right-of-center` onto the side pair because ffmpeg has no name for them).

@kahrendt — I'd want you to weigh in here for the sendspin discussion, esp with your multi-channel experience.

That suggestion is deliberately channel-based only. Object formats like Atmos are a different problem: a bed plus objects with per-frame coordinates, not a fixed map, and out of scope here I suspect.

---

## If we do go ahead with this as-is

Not blocking on the spec, but these need closing, because the failure mode in each case is silent rather than loud:

- **DSP isn't actually blocked, it just eats channels.** The per-channel EQ preamp at `dsp.py:103` (dev @ `69a282fc`) emits `pan=stereo|FL=...|FR=...` unconditionally — balance, width and crossfeed all check `channels == 2` but this one doesn't. On a 7.1 stream that leaves FL/FR and silences the other six. Exit code 0, byte count unchanged, still labelled 7.1.
- **Smart fades break for a solo multichannel player.** `fades.py:103/116/139` still build `"mono" if channels == 1 else "stereo"` next to `-ac {channels}`, and `-channel_layout` wins over `-ac` — the buffer gets read as stereo at 4x the frame count, so channels scramble *and* fade timing lands at a quarter of where it should. Worked fine before because `select_flow_pcm_format` hardcoded stereo.

- **Joining mid-playback.** The `is_grouped` check covers grouped-at-start, but `add_member`/`sync_members` run live and never re-select the format, so a stereo player joining an active session gets 8-channel PCM. Nothing downstream converts it — `get_player_output_plan` never compares input channels to output channels.

- **The opt-in isn't really an opt-in.** `CONF_OUTPUT_CHANNELS == "multichannel"` only takes effect in the local_audio bridge. The native path *does* read the key (`sendspin/playback.py:1235-1239`) but whitelists `{stereo,left,right,mono}` at `:1241`, so `"multichannel"` is silently coerced to `"stereo"` — its width comes instead from `preferred_format`, which aiosendspin sets from `compatible[0]`, the first format the client advertises. Advertise 8 channels first and you get a multichannel session with no config at all, straight into the three items above.

## Two smaller things

**The ffmpeg channel-layout fix and the downmix matrix have to move together.** I initially thought `_channel_layout()` was a clean standalone fix, but it isn't — right now two bugs cancel out. On `dev`, `-ac 6 -channel_layout stereo` doesn't mislabel, it *actually downmixes* (I get 1130496 bytes in, 376832 out), and that accidental early downmix uses libswresample's matrix, which handles BL/BR correctly. Fix the labelling on its own and the downmix defers to ours at `ffmpeg.py:726-730`, which references SL/SR but never BL/BR — so 5.1 surrounds go silent. On a 5.1 source with signal only in the surrounds I get 0.0624/0.0623 on dev today vs 0.0000/0.0000 with just the layout fix. Something like this fixes the matrix and holds up on 7.1 too:

```
pan=stereo|FL=FL+0.707*FC+0.707*LFE+0.707*BL+0.707*SL
          |FR=FR+0.707*FC+0.707*LFE+0.707*BR+0.707*SR
```

(Worth a sanity check on headroom as that's five sources folded into each output now.)

**I don't think you need the stderr layout parsing at all.** `_LAYOUT_TO_CHANNELS`, the regex and `FFMpegStreamInfo.channels` look redundant since ffprobe already populates `streamdetails.audio_format.channels` at `tags.py:666`, before the streaming ffmpeg ever runs. And the parsed value gets written in `_apply_input_stream_info`, which fires on the first stderr line, well after `select_pcm_format` has already picked the format — so it can't influence anything. Worth noting `ffmpeg_proc.input_format` is the *same object* as `streamdetails.audio_format` (see the comment at `audio.py:596`), so it's quietly mutating stream state mid-playback too. Dropping it takes ~110 lines out and removes the count→layout round-trip problem entirely.

---

Genuinely think this is doable and worth landing — I'd just rather we spend the effort on the spec so there's less to guess at, and so we're not maintaining three tables that have to agree.

(Note: AI has been used to do STT and edit/format my response since I suck at typing)

vingerha:
@chrisuthe 

Thanks Chris and hereby the update

> **Can we pull the `volume_control: none` fix out into its own PR?** 

Reverted, see commit

> **DSP isn't actually blocked, it just eats channels.** 
> **The opt-in isn't really an opt-in.**
> **The ffmpeg channel-layout fix and the downmix matrix have to move together.**
> **I don't think you need the stderr layout parsing at all.**
> **Smart fades break for a solo multichannel player.**

Done, see various commits, I hope this in line with the expectations :)

> * **Joining mid-playback.** The `is_grouped` check covers grouped-at-start, but `add_member`/`sync_members` run live and never re-select the format, so a stereo player joining an active session gets 8-channel PCM. Nothing downstream converts it — `get_player_output_plan` never compares input channels to output channels.

This I found more complex than the rest — two options:
- Preferred: leave as-is, same treatment as for above volume_control: none. An add/remove in the sync group player already require a restart in the current code, so this doesn't regress anything.
- Alternatively if you believe this is important to add to an already larger PR: please provide pointers :)

> **The bigger thought: I'd really like to fix the spec first**

Awaiting your + @kahrendt's input on this before touching it further.

OzGav:
Marking as draft as Chris reviews the way ahead for the local in and out providers.

chrisuthe:
Hey @vingerha  and others - had some discussion internally with the sendspin and MA team at ohf about this and I want to urge us to go in another direction.

Right now I think we should pull the local audio out plugin from MA completely, and rely on either a CLI player installed locally or a docker container or HAOS add-on to expose local audio as a sendspin player. This means the hardware handling aspects are kept  out of Music Assistant so it stays as agnostic as it can be about what you're running it on. 

We've got the multi-sendspin container and HAOS add on that I put a good bit of effort into (Shoutout @scyto as well) that is in use at a good number of places, as well as an upcoming sendspin-CPP cli that mirrors the usefulness of the squeezelite CLI. The plan is to make a SUPER basic HAOS plugin out of that as well, so folks that just want local audio out can use that without any extra work, but the code stays out of the server.

Taking thoughts and opinions for and against of course, but I think this is a safer way forward..

vingerha:
@chrisuthe 
As per discord: pity of the effort, also aside other effort my multi-channel work.... but the cli appaorch seems OK, although not yet working for multi-channel (you already know that) and I hope this has a reasonable outlook.

chrisuthe:
Work will continue as a different thread with MA core and Sendspin pos v1

PR #4947, opened 2026-07-23
Add DSP convolution filter

OzGav:

Adds the new `ConvolutionFilter` DSP type, which applies an impulse response to the audio, room correction (REW, Acourate, Audiolense), speaker correction, or headphone simulation.

Impulse responses are uploaded and stored server side rather than embedded in the filter, so the DSP config only carries a small id and the same file can be shared across players and presets. Files live in a `dsp_irs` subdirectory of the storage path with a metadata record in the config store, managed by three new commands: `config/dsp_irs/list`, `/upload` and `/remove`. Uploads are transcoded to wav on the way in, which validates they are really audio. Removing an impulse response that is still in use blanks it from any config or preset referencing it instead of leaving a dangling id.

Frontend PR https://github.com/music-assistant/frontend/pull/2190
Models PR: https://github.com/music-assistant/models/pull/336

## Types of changes

## Checklist

- [X] For changes to shared models, the companion PR in `music-assistant/models` is linked.
- [X] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked.


OzGav:
Setting as draft as Steven identified we need a new event so need a models bump to permit that.

music_assistant/controllers/streams/audio.py
```diff
@@ -1181,8 +1188,18 @@ def get_player_output_plan(
...
             for dsp_filter in enabled_filters:
-                params = filter_to_ffmpeg_params(dsp_filter, input_format)
+                # ffmpeg fails to open the graph if the impulse response file is gone,
+                # which costs the player all audio, so drop the filter instead
+                is_unknown_ir = (
+                    isinstance(dsp_filter, ConvolutionFilter)
+                    and dsp_filter.ir_id not in known_ir_ids
+                )
+                if is_unknown_ir:
+                    continue
```
OzGav:
```suggestion
                if isinstance(dsp_filter, ConvolutionFilter) and dsp_filter.ir_id:
                    # ffmpeg fails to open the graph if the impulse response file is gone,
                    # which costs the player all audio, so drop the filter instead
                    if dsp_filter.ir_id not in known_ir_ids:
                        self.logger.warning(
                            "Skipping the convolution filter of player %s: "
                            "impulse response %s is not stored",
                            player_id,
                            dsp_filter.ir_id,
                        )
                        continue
```
Would a log line be worth it here? When we drop the filter the player just sounds different and nothing in the log says why, so I think support would only ever see "my convolution stopped working"

music_assistant/controllers/config/dsp.py
```diff
@@ -123,6 +150,139 @@ async def remove_dsp_preset(self, preset_id: str) -> None:
...
+                ir_path,
+                timeout=IR_TRANSCODE_TIMEOUT,
+            )
+            if returncode != 0:
+                raise InvalidDataError(
+                    f"Uploaded file is not valid audio: {output.decode().strip()}"
+                )
+            # a lossy codec smears the impulse across neighbouring samples, which
+            # corrupts the very measurement the response is supposed to apply
+            source_format = ContentType.try_parse((await async_parse_tags(upload_path)).format)
```
OzGav: I had Claude play with this one. If I do `ffmpeg -c:a libmp3lame out.wav` the file comes back as format wav with codec_name mp3, so I think it gets accepted even though it's an mp3 inside? Meanwhile flac in ogg and Wave64 get rejected for being the wrong container even though the audio is lossless. Is that right?

I know you called out the ogg and mp4 cases as deliberate and that;s understood, but the wav one is the case the check is meant to catch, isn't it? Would reading the stream codec be better here? I had a look and it wouldn't need much, AudioTags is in our repo and `parse()` already grabs the audio stream, so a small codec property would do it with no models change. `try_parse` seems happy with codec names too, pcm_s16le, flac, alac and wavpack all come back lossless and mp3, aac, vorbis and opus don't. Happy either way, just wondering if it's worth doing here or if you'd rather keep it separate.

music_assistant/controllers/config/dsp.py
```diff
@@ -123,6 +150,139 @@ async def remove_dsp_preset(self, preset_id: str) -> None:
...
+                str(MAX_IR_SECONDS + 1),
+                "-c:a",
+                "pcm_f32le",
+                ir_path,
+                timeout=IR_TRANSCODE_TIMEOUT,
+            )
+            if returncode != 0:
+                raise InvalidDataError(
+                    f"Uploaded file is not valid audio: {output.decode().strip()}"
+                )
```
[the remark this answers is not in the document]
OzGav: Seems low probability but is free to add.

music_assistant/controllers/config/dsp.py
```diff
@@ -165,3 +326,89 @@ def _clear_dsp_preset_assignments(self, preset_id: str) -> None:
...
+
+        :param config: The DSP configuration to check.
+        """
+        stored: dict[str, dict[str, Any]] = self.get(CONF_PLAYER_DSP_IRS, {})
+        for dsp_filter in config.filters:
+            # an empty id is the "none selected yet" value, which stays allowed
+            if not isinstance(dsp_filter, ConvolutionFilter) or not dsp_filter.ir_id:
+                continue
+            if dsp_filter.ir_id not in stored:
+                raise InvalidDataError(f"Unknown impulse response: {dsp_filter.ir_id!r}")
```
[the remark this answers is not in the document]
OzGav: I don't think this can happen. The IR library is new in this PR so there are no legacy records

music_assistant/helpers/dsp.py
```diff
@@ -19,43 +21,62 @@
...
 
     Args:
         dsp_filter: DSP filter configuration
         input_format: Audio format containing sample rate
+        ir_dir: Directory holding convolution impulse response files, used to
+            resolve a ConvolutionFilter's ir_id to a file path
 
     Returns:
         List of FFmpeg filter parameter strings
     """
```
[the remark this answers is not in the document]
OzGav: Yes correct.

music_assistant/controllers/config/dsp.py
maximmaxim345: I always forget that wav supports mp3...

My fix was more for catching accidentally wrong file outputs since lossy codecs mess with the impulse response enough to be noticeable.

From what I saw IRs usually come in wav, but sometimes also as flac when generated for Room Correction. Haven't seen any other container for them.

I can revert 95da109 again if you like. I think nobody has lossy IRs either way unless they manually generated/encoded them. In that case if somebody uses lossy IRs it's their problem.

OzGav: I'm new to this but when I looked for some to test with I only found WAV or FLAC. I think we can make it work with a property but equally people using this should hopefully know what they are doing and not use a compressed IR file?

maximmaxim345: Reverted 95da109.
Just tested Fruity Convolver in FL Studio, that even accepts mp3s so lets keep it up to the user.

maximmaxim345 (approved):
Thanks @OzGav !
Even when doing Room Correction, some things just can't be done with a simpler PEQ.
This makes me want to try out measuring my Speakers again with fancy DRC software.

PR #5016, opened 2026-07-25
Don't resume an unrelated track from the global playlog

### LINKED ISSUE #5913: Announcement to an idle sync group starts unrelated playback via playlog "any recent item" fallback
### Before you begin

- [x] I have searched the [open and closed issues](https://github.com/music-assistant/support/issues?q=is%3Aissue) and this hasn't been reported yet.
- [x] I have read the [troubleshooting guide](https://www.music-assistant.io/faq/troubleshooting/) and it did not solve my problem.
- [x] I am running a [supported installation](https://www.music-assistant.io/installation/) (Home Assistant add-on or Docker container).

(Closest prior report I found is #4791 — "Announcements resume issues" — which was closed/fixed and was Squeezelite-specific. This one is a different code path: it involves sync-group players and the playlog fallback described below.)

### What happened?

An announcement sent to a **sync group** that was **idle** (nothing playing, nothing paused) causes Music Assistant to start playing music once the announcement finishes. The track it picks is frequently unrelated to that room — it appears to be whatever was most recently played anywhere in the system.

Setup: a Home Assistant automation announces "Front door opened." via `chime_tts` + Piper to an HA `media_player` group whose members are two MA sync-group players (`Kitchen Speakers Group`, `Master Bedroom Speakers Group`).

Nothing was playing on either group at the time. After the announcement, music started on its own.

The MA log for one occurrence (Nabu Casa host + `authSig` JWT redacted by me):

```
17:39:20 INFO    music_assistant.players :: Playback announcement to player Master Bedroom Speakers Group (with pre-announce: False): https://<REDACTED>.ui.nabu.casa/media/local/sounds/temp/chime_tts/<hash>.mp3?authSig=<REDACTED>
17:39:20 WARNING music_assistant.players :: Detected announcement to playergroup Master Bedroom Speakers Group while group member Master Closet Speaker is playing other content, this may lead to unexpected behavior.
17:39:31 INFO    music_assistant.Sync Group Player :: Idle-grace expired for syncgroup Master Bedroom Speakers Group, dissolving
17:39:39 WARNING music_assistant.player_queues :: Ignore queue command: An announcement is in progress
17:39:39 INFO    music_assistant.player_queues :: Resumed queue Master Bedroom Speakers Group from playlog (any recent item)
```

Note the `Idle-grace expired ... dissolving` at 17:39:31 — the group was already idle, which is why the grace timer was counting down *during* the announcement. Despite nothing having been playing, the queue is resumed at 17:39:39 via the playlog's `any recent item` match.

**Expected:** an announcement to an idle player/group should restore the previous state (i.e. stay idle) and not start playback.

**Actual:** playback starts with an arbitrary recently-played track.

### How to reproduce

1. Create a sync group with a few members.
2. Play something on that sync group, then stop it and let it go idle (long enough that the idle-grace dissolve is pending/fires).
3. Send an announcement to that sync group (e.g. `music_assistant.play_announcement`, or a TTS announcement from a Home Assistant automation targeting the group).
4. When the announcement finishes, the group starts playing a track that was not playing before — often one last played on a *different* player.

### Music Assistant version

2.9.9

### How do you run Music Assistant?

Docker container

### Diagnostics report or log file

I'm filing this via the GitHub API/CLI, which can't drag-and-drop a file attachment, so the pertinent log excerpt is inlined above rather than attached. Happy to upload the full diagnostics JSON or server log on request.

Supporting details from the running instance:

- `music_assistant`: 2.9.9, `music_assistant_models`: 1.1.129.post1
- Providers loaded: 40 (0 disabled, 0 in error), queues total: 48
- Affected players are `sync_group` provider players; their members are `universal_player` provider players
- Announcement source: Home Assistant `chime_tts` + `tts.piper`, targeting an HA media_player group of two MA sync groups

### Anything else?

I dug through the 2.9.9 source and there appear to be two independent contributing factors. Sharing in case it's useful — flagging that the runtime call chain is my reading of the code, not something I instrumented, so please treat the *analysis* as a hypothesis; the log evidence above is what I actually observed.

**1. The sync-group reform path resumes without checking whether it was playing.**

The default announcement handler `PlayerController._play_announcement` correctly gates the restore on the prior state:

```python
# music_assistant/controllers/players/controller.py (2.9.9)
elif prev_state == PlaybackState.PLAYING:
    # player was playing something before the announcement - try to resume that here
    await self._handle_cmd_resume(player.player_id, prev_source, prev_media)
```

But announcing to a sync group ungroups/re-forms the group, and the sync group's own `play()` issues a resume with no equivalent guard:

```python
# music_assistant/providers/sync_group/player.py (2.9.9, and unchanged on dev)
async def play(self) -> None:
    """Send PLAY (unpause) command to given player."""
    await self._form_syncgroup()
    async with self._await_leader_playback():
        await self.mass.players.cmd_resume(
            self.player_id, self._attr_active_source, self._attr_current_media
        )
```

The `Ignore queue command: An announcement is in progress` line immediately preceding the resume looks like this firing while `ATTR_ANNOUNCEMENT_IN_PROGRESS` is still set, then succeeding once it clears.

**2. `resume()` on an emptied queue falls back to "any recent item" globally.**

Because the group had gone idle and dissolved, the queue had no current item, so `resume()` falls through to the playlog:

```python
# music_assistant/controllers/player_queues.py (2.9.9)
# Queue is empty, try to resume from playlog
if await self._try_resume_from_playlog(queue):
    return
```

and `_try_resume_from_playlog` ends with an unscoped catch-all:

```python
filter_attempts.append((queue.userid, queue.queue_id, "userid + queue_id match"))
filter_attempts.append((None, queue.queue_id, "queue_id match"))
filter_attempts.append((queue.userid, None, "userid match"))
filter_attempts.append((None, None, "any recent item"))   # no queue, no user
```

That last attempt asks `music.recently_played()` for the most recent user-initiated item **anywhere in the system**, which is why the resulting track is often from an unrelated room. This is the `(any recent item)` match type in the log line.

Both are still present on `dev` as of 2.10.0.dev2026072406 — `sync_group/player.py::play()` is unchanged, and the fallback moved to `music_assistant/controllers/player_queues/queue_loader.py` (`_try_resume_from_playlog`, ~line 441) with the `(None, None, "any recent item")` attempt intact.

A restore-after-announcement that finds an empty queue arguably shouldn't reach the global playlog fallback at all, since "resume what was playing" and "start something plausible" are different intents.

geofffranks:
Resuming an empty queue (e.g. after an announcement on an idle player/sync
group) fell through to an unscoped "any recently played item" fallback, which
started the most recent play from anywhere in the system.

Removed that global fallback so a resume only considers this queue's (and its
user's) own history. Adds two unit tests; queue/user-scoped lookups unchanged.

Closes music-assistant/support#5913.

## Types of changes

### DIFF
--- a/music_assistant/controllers/player_queues/queue_loader.py
+++ b/music_assistant/controllers/player_queues/queue_loader.py
@@ -442,11 +442,16 @@ class QueueLoaderMixin(_PlayerQueuesBase):
         """
         Try to resume playback from playlog when queue is empty.
 
-        Attempts to find user-initiated recently played items in the following order:
+        Attempts to find user-initiated recently played items for *this* queue
+        or its user, in the following order:
         1. By userid AND queue_id
         2. By queue_id only
         3. By userid only (if available)
-        4. Any recently played item
+
+        Only this queue's/user's own history is considered — a resume must never
+        start a track that was played elsewhere. The former global "any recent
+        item" fallback was removed because it resumed an unrelated track (e.g.
+        after an announcement on an idle player or sync group). See support #5913.
 
         :param queue: The queue to resume playback on.
         :return: True if playback was started, False otherwise.
@@ -459,7 +464,6 @@ class QueueLoaderMixin(_PlayerQueuesBase):
         filter_attempts.append((None, queue.queue_id, "queue_id match"))
         if queue_data.userid:
             filter_attempts.append((queue_data.userid, None, "userid match"))
-        filter_attempts.append((None, None, "any recent item"))
 
         for userid, queue_id, match_type in filter_attempts:
             items = await self.mass.music.recently_played(
--- a/tests/controllers/player_queues/test_resume_from_playlog.py
+++ b/tests/controllers/player_queues/test_resume_from_playlog.py
@@ -1,0 +1,65 @@
+"""
+Tests for resuming an empty queue from the playlog.
+
+Covers the scope of the playlog fallback used when a queue is resumed while
+empty. A resume must only continue *this queue's/user's* own recently-played
+history. It must never pull a globally-recent item that was played elsewhere,
+which is how a restore after an announcement on an idle (sync group) player
+ended up starting an unrelated track. See support #5913.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, Mock
+
+from music_assistant.controllers.player_queues import PlayerQueuesController
+
+
+def _make_controller() -> tuple[PlayerQueuesController, AsyncMock]:
+    """Bare controller instance with only the dependencies the resume path touches."""
+    controller = PlayerQueuesController.__new__(PlayerQueuesController)
+    controller.logger = Mock()
+    handle_play_media = AsyncMock()
+    controller._handle_play_media = handle_play_media  # type: ignore[method-assign]
+    return controller, handle_play_media
+
+
+async def test_resume_from_playlog_ignores_globally_recent_item() -> None:
+    """A resume must not start a track that only exists in the unscoped/global playlog."""
+    controller, handle_play_media = _make_controller()
+    foreign = Mock(uri="spotify://foreign", name="Foreign")
+
+    async def recently_played(**kwargs: object) -> list[Mock]:
+        # the track exists only in the global (unscoped) playlog,
+        # not in this queue's or user's own history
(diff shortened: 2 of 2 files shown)


marcelveldt:
This is most certainly not the right fix for the issue that was reported.

PR #5028, opened 2026-07-25
Report per-provider search status in search results

jozefKruszynski:

Reports how each provider contributed to a global search, so clients can tell
"this provider found nothing" apart from "this provider timed out / errored"
instead of guessing from an empty result.

`SearchResults` now carries `provider_search_statuses` (per provider,
`complete` / `timeout` / `failed`). `_search_provider` returns its status
alongside its results; `search()` collects them (library and shortcut-skipped
providers are `complete`).
Caching is unchanged — combined results are still
only cached when every provider completed, so cached payloads only ever carry
`complete` statuses.

Companion PRs:
https://github.com/music-assistant/models/pull/331
https://github.com/music-assistant/frontend/pull/2207

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music/controller.py
+++ b/music_assistant/controllers/music/controller.py
@@ -19,6 +19,7 @@ from music_assistant_models.enums import (
     EventType,
     MediaType,
     ProviderFeature,
+    ProviderSearchStatus,
     ProviderType,
     TaskStatus,
 )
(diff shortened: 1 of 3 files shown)


marcelveldt:
I think this is a bit over-engineered. A much more simple take on this is by simply throwing an exception here (based on MusicAssistantError so its even translatable).

- Throw a timeout error if the provider timed out
- Throw another specific error if fetching the items failed

This will already naturally travel to the frontend today, no additional plumbing needed at all.

Note that we have adjusted the frontend nowadays so it will do separate queries for the global search instead of one big unified one. Frontend does one call to get the library results (very fast) and then subsequent calls per (selected) providers. So this whole flow already works well if you combine it with propagating errors if they arrise. We should just not swallow the errors but convert them to a known error where needed and raise.

OzGav:
Any progress on this one Jozef?

PR #5039, opened 2026-07-26
Reconnect SMB mount when stale during library scan enumeration

kiwipaulrob:

Adds automatic SMB mount recovery to `SMBFileSystemProvider` by overriding `_enumerate_files_for_sync` in the subclass.

## Problem

When the SMB server behind a CIFS mount temporarily disconnects during a library scan (transient network blip, NAS reboot, etc.), `os.scandir` raises an `OSError` at the mount root. The parent class's `recursive_iter` helper catches this and populates `root_scan_errors`, which causes `sync_library` to **abort the entire scan** and mark the provider as **unavailable** until a manual restart.

For users with large libraries (200k+ files), even a brief SMB disconnect can abort days of scan progress and require manual intervention to recover.

## Solution

The override in `SMBFileSystemProvider`:

1. Checks `ismount()` before starting the walk and remounts if the CIFS mount is stale
2. If the parent's walk returns `root_scan_errors`, unmounts and remounts the share with exponential backoff (1s, 2s, 4s), then retries
3. Only passes errors through to the parent's abort logic after 3 failed attempts

This prevents transient SMB disconnects from aborting the scan, while still failing definitively if the server is genuinely unreachable.

- Related support issue: https://github.com/music-assistant/support/issues/5917
- Previous (withdrawn) mount-options PR: #5023

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) - `bugfix`

## Checklist


OzGav:
Please dot submit PRs for review which are failing CI. That is why we have the checklist. Marked as draft, p,ease mark as ready for review when it is.

kiwipaulrob:
Sorry about that.

kiwipaulrob:
You're absolutely right — I should have marked this as a draft before submitting for review. I've converted it to draft now.

I'll work through the CI steps (pre-commit, pytest) and mark it ready for review once everything passes. Sorry for jumping the gun — I'll make sure future PRs follow the proper workflow.

Thank you for the guidance.

kiwipaulrob:
CI is now all green — lint and test both pass. The issues were:

1. `FileSystemItem` type needed importing in the SMB provider (F821)
2. Method ordering: private methods need to be grouped together alphabetically
3. Formatting: `ruff format` applied to the new code

The one remaining lint finding (`CPY001` copyright notice) is pre-existing in the original file and is CI-filtered, so it doesn't block.

I've marked the PR as ready for review. Thanks for your patience with the CI workflow — I'll make sure future PRs are drafted first until CI passes.

OzGav:
Thanks for the follow-up attempt, but as I see it there is a serious flaw in the retry loop that would cause library data loss. Stepping through what happens:

1. When a sync runs, Music Assistant first makes a list of every file it already knows about from previous scans. It then walks the share to build a second list of every file it can see right now. At the end it compares the two lists and deletes anything from the library that was in the old list but not seen in the new scan. That is how removed files get cleaned up. There are two safety checks before the delete step: the sync stops if the scan reported errors at the top level of the share, and it stops if the scan found nothing at all.

2. Now add your retry loop and imagine the NAS drops out partway through a scan of a big library. The first attempt gets through 130k of 290k files, then the share stops responding and the scan records the error.

3. Your code sees the error, clears the error list, unmounts, waits, and tries again. The second and third attempts do the same. Importantly, the list of files found so far is never reset, so it still holds the 130k files from the first partial attempt.

4. After the third failed try the loop just ends and the sync carries on. At this point the error list is empty, because your retry code cleared it each time, and the "files seen" list holds 130k files.

5. The safety checks now pass when they should not. No errors recorded, so no stop. Files were found, so no stop. The sync moves on to the compare-and-delete step and removes the 160k files it never got to, along with their playlist entries, favourites and play counts.

So if the retries run out, a short NAS outage ends up deleting a large chunk of the library. The current code in dev stops the sync in that same situation and nothing is lost. The comment in your code says the errors are handed back so the sync will stop safely, but the code clears them first, so that never happens. I am surprised you didn't see this in your testing.

There are more problems beyond that one. Because the file lists are kept between attempts, a retry that succeeds will process the same files twice. If the remount itself fails, which is the most likely thing when the server is down, the code throws an error on the first try and the backoff never happens. When the code gives up it leaves the share unmounted, so playback and browsing are broken until the provider is restarted. And the retry only kicks in when the top folder of the share fails to read, while the errors in the original issue were thousands of failures in subfolders, which this code never notices. There are also no tests covering any of this new behaviour.

On the remounting you are doing, it is not needed at all. When an SMB connection drops, the OS does not unmount the share. The mount stays in place and the Linux kernel automatically reconnects as soon as the server is reachable again, using the same credentials. So in the realistic failure case, a NAS that reboots or a network blip, the mount heals itself and there is nothing for Music Assistant to do. Unmounting and remounting while the server is down cannot succeed anyway, and risks leaving the share in a worse state than if it had been left alone.

So the approach here is not correct. The fix should be done higher up in the shared filesystem provider code where it will also benefit all the file system providers. What should happen is the scan should give up quickly once it hits a run of consecutive read errors, instead of grinding through every remaining file at a couple of minutes per failure, which is what turned a ten second outage into days in your case. A scan that had any errors should never run the delete step, so a failed scan can never remove anything from the library. And after a failed sync the provider should retry sooner than the normal sync interval and mark itself available again once a scan succeeds, rather than staying unavailable until someone restarts it. None of that requires touching the mount, and it would protect a local USB drive that disappears mid-scan just the same as an SMB share. I would suggest closing this PR, as the retry-and-remount idea is not salvageable, but a PR along the lines of the above would be welcome. Alternatively, I can work on it.

kiwipaulrob:
Thank you for the thorough review — this is excellent feedback and you're right on every point. The data loss scenario from clearing `root_scan_errors` without resetting the file list is a genuine bug, and the approach is not salvageable.

I've closed PR #5039.

For the approach you described (consecutive-error detection in the shared filesystem provider, with a scan that had errors never running deletions, and automatic retry), that makes much more sense — it would fix this for every filesystem provider, not just SMB.

If you're happy to work on it, please go ahead — you clearly have a much better understanding of where the fix belongs in the codebase. I'm happy to test the fix on our Frigate/NVR setup (which has a large library and sees the issue regularly) if that would help.

PR #5062, opened 2026-07-26
Follow-up fixes and enhancements to author and narrator support

fmunkes:

Follow up fixes to artist support of type ArtistType.AUTHOR / ArtistType.NARRATOR:
- ProviderFeature.LIBRARY_ARTIST doesn't indicate, what kind of artists are supported. Add an additional API endpoint giving all library-supported artist types
- ProviderFeature.AUTHOR_AUDIOBOOKS / ProviderFeature.NARRATOR_AUDIOBOOKS are for streaming providers, where the provider might have additional audiobooks of an author/ narrator not part of the library.
    - Remove from audiobookshelf, including methods. Doesn't make any sense, audiobookshelf is not a streaming provider, and we sync all artists to the library
    - Do not use these ProviderFeatures to check, if an audiobook.author/ narrator should be a string or Artist
- Introduce a new music model property supported_artist_types, which is instead used for verification. By default this is ArtistType.SINGER, but not used anywhere.

Nothing of this is breaking, the only provider using this until now was abs, and it is fix in this PR as well.

The frontend companion PR depends on these changes.
https://github.com/music-assistant/frontend/pull/2000

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/music/media/artists.py
+++ b/music_assistant/controllers/music/media/artists.py
@@ -95,8 +95,8 @@ class ArtistsController(MediaControllerBase[Artist]):
             required_scope=Scope.LIBRARY_READ,
         )
         self.mass.register_api_command(
-            f"music/{api_base}/supported_library_artist_types",
-            self.get_supported_library_artist_types,
+            f"music/{api_base}/library_artist_types",
+            self.get_library_artist_types,
             required_scope=Scope.LIBRARY_READ,
         )
 
@@ -841,7 +841,7 @@ class ArtistsController(MediaControllerBase[Artist]):
                 result.append(candidate)
         return result[:limit]
 
-    async def get_supported_library_artist_types(self) -> list[ArtistType]:
+    async def get_library_artist_types(self) -> list[ArtistType]:
         """Get all supported in-library artist types."""
         artist_types: list[ArtistType] = []
         query = f"SELECT DISTINCT artist_type FROM {DB_TABLE_ARTISTS}"
--- a/music_assistant/controllers/music/media/artists.py
+++ b/music_assistant/controllers/music/media/artists.py
@@ -95,7 +95,7 @@ class ArtistsController(MediaControllerBase[Artist]):
             required_scope=Scope.LIBRARY_READ,
         )
         self.mass.register_api_command(
-            f"music/{api_base}/supported_artist_types",
+            f"music/{api_base}/supported_library_artist_types",
             self.get_supported_library_artist_types,
             required_scope=Scope.LIBRARY_READ,
         )
--- a/music_assistant/controllers/music/media/artists.py
+++ b/music_assistant/controllers/music/media/artists.py
@@ -94,6 +94,11 @@ class ArtistsController(MediaControllerBase[Artist]):
             self.similar_artists,
             required_scope=Scope.LIBRARY_READ,
         )
+        self.mass.register_api_command(
+            f"music/{api_base}/supported_artist_types",
+            self.get_supported_artist_types,
+            required_scope=Scope.LIBRARY_READ,
+        )
 
     @property
     def summary_query(self) -> tuple[str, dict[str, Any]]:
@@ -835,6 +840,15 @@ class ArtistsController(MediaControllerBase[Artist]):
                     continue
                 result.append(candidate)
         return result[:limit]
+
+    async def get_supported_artist_types(self) -> list[ArtistType]:
+        """Return all supported artist types."""
+        artist_types: list[ArtistType] = []
+        query = f"SELECT DISTINCT artist_type FROM {DB_TABLE_ARTISTS}"
+        rows = await self.mass.music.database.get_rows_from_query(query)
+        for row in rows:
+            artist_types.append(ArtistType(row["artist_type"]))
+        return artist_types
 
     async def remove_item_from_library(self, item_id: str | int, recursive: bool = True) -> None:
         """Delete record from the database."""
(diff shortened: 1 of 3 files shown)


fmunkes:
I'll mark this one as draft, I might find some additional smaller adjustments to make the author/ narrator available in the UI.

fmunkes:
> @fmunkes The title or description of this pull request needs a fix:
> 
>     * Checklist item not ticked: I have read and complied with the project's [AI Policy](https://github.com/music-assistant/.github/blob/main/AI_POLICY.md) for any AI-assisted contributions.
> 
> 
> Release notes are generated from the title, and the template carries what reviewers need, so please edit them before this is reviewed.

I didn't use AI here...

music_assistant/models/music_provider.py
```diff
@@ -81,6 +81,16 @@ def is_streaming_provider(self) -> bool:
         """
         return True
 
+    @property
+    def supported_artist_types(self) -> set[ArtistType]:
+        """
+        Return all supported artist types by this provider.
+
+        Note, that this property only is used, if an artist-related ProviderFeature like
+        ProviderFeature.LIBRARY_ARTISTS is supported by the provider.
```
OzGav: Is this docstring right? As far as I can see the only place supported_artist_types is used is `_validate_audiobook_author_narrator_types`, which runs on audiobook sync for every provider regardless of `LIBRARY_ARTISTS` e.g. teddycloud doesn't support `LIBRARY_ARTISTS` but still relies on the `{SINGER}` default here to keep its string authors valid. Should the note instead say something like "determines whether audiobook authors/narrators must be Artist objects or plain strings"?

music_assistant/models/music_provider.py
```diff
@@ -1053,32 +1063,34 @@ async def import_album_tracks(self, prov_album_id: str, album_name: str | None =
...
-        if ProviderFeature.AUTHOR_AUDIOBOOKS in self.supported_features and not all(
-            isinstance(author, Artist) for author in prov_item.authors
+        if ArtistType.AUTHOR in self.supported_artist_types and not all(
+            (isinstance(author, Artist) and author.artist_type == ArtistType.AUTHOR)
+            for author in prov_item.authors
         ):
             raise MusicAssistantError(
-                f"Provider {self.name} supports ProviderFeature.AUTHOR_AUDIOBOOKS, but"
+                f"Provider {self.name} supports ArtistType.AUTHOR, but"
                 f" item {prov_item.name} does not exclusively provide Artist instances."
```
OzGav: Should the error messages be updated too? The checks now also require the artist type to be author or narrator, but the message still only complains about the item not exclusively providing Artist objects. So an Artist with the wrong type would fail with a message that doesn't explain the real problem. Maybe something like "does not exclusively provide Artist objects of type author"? Same for the docstring, which still mentions supported features rather than artist types.

music_assistant/models/music_provider.py
```diff
@@ -1053,32 +1063,34 @@ async def import_album_tracks(self, prov_album_id: str, album_name: str | None =
 
     def _validate_audiobook_author_narrator_types(self, prov_item: Audiobook) -> None:
         """Validate that authors/narrators types match the provider's supported features."""
-        if ProviderFeature.AUTHOR_AUDIOBOOKS in self.supported_features and not all(
-            isinstance(author, Artist) for author in prov_item.authors
+        if ArtistType.AUTHOR in self.supported_artist_types and not all(
+            (isinstance(author, Artist) and author.artist_type == ArtistType.AUTHOR)
+            for author in prov_item.authors
         ):
             raise MusicAssistantError(
```
OzGav: I know its legacy code but maybe InvalidDataError? (A couple more below)

music_assistant/models/music_provider.py
OzGav: I think because of your change this comment is no longer accurate. The mechanism that flips authors/narrators between strings and Artists is now `supported_artist_types`, not a ProviderFeature

music_assistant/controllers/music/media/artists.py
```diff
@@ -822,6 +827,15 @@ async def get_library_artist_similar_artists(
                 result.append(candidate)
         return result[:limit]
 
+    async def get_supported_library_artist_types(self) -> list[ArtistType]:
```
OzGav: I might be reading this wrong, but is "supported" the right word here? I think this returns the artist types that are actually present in the library right now, rather than what's supported. So I think a fresh install, or an ABS setup that hasn't synced yet, would return an empty list even though authors and narrators are fully supported. If presence in the library is the intent (which seems right for the frontend), maybe rename to `get_library_artist_types` and `music/artists/library_artist_types` while the frontend PR is still open? If it's meant to be about capability, I think it would need to come from the providers' `supported_artist_types` instead. What do you think?

OzGav (commented):
A couple of questions from me!

music_assistant/models/music_provider.py
fmunkes: You are right. I'll clarify the doc string that it is (as of now) only used for author/ narrator verification.

music_assistant/models/music_provider.py
fmunkes: I updated all of them.

music_assistant/models/music_provider.py
fmunkes: Done.

music_assistant/models/music_provider.py
fmunkes: Another one. Thanks for being so thorough with reading the code and comments. Updated as well.

music_assistant/controllers/music/media/artists.py
fmunkes: Your wording is better. I'll change the method's to get_library_artist_types. Frontend PR is updated as well.

OzGav (approved):

PR #5076, opened 2026-07-27
Handle remote storage going away during a library sync

OzGav:

Fixes a problem reported where a short NAS outage during a scan of a large SMB library left the scan running for days and the provider switched off until it was set up again. 

The fix sits in the shared filesystem provider code, so the local, SMB, NFS, WebDAV and cloud providers all get it.

This PR:

- stops a scan when the storage stops responding. Only an error on the top level folder used to stop a sync. Errors on sub folders were logged and skipped, and on a share that is gone each of those had to time out first, which is what turned a ten second outage into days of scanning. The scan now counts folders that fail one after the other, resets the count on every folder it reads fine, and gives up once too many fail in a row. Folders skipped because their name has characters the share does not support are not counted.

-  stops library items being deleted after an incomplete scan. A scan that could not read some folders still counted as a success, so everything below those folders looked like it had been removed and was deleted from the library. The same applied to an item the scan could see but could not look at closely enough to tell whether it was a folder or a file, which can hide a whole branch of the library. Deletions and the cleanup that follows them are now skipped whenever either happens. Files that were scanned fine are still processed, and the sync is marked as partly failed, so the reason shows in the task list along with the count and a few example paths of what could not be read.

 - stops a file that cannot be read from being deleted. Reading a file's details can fail for all sorts of reasons, from a corrupt download to the share dropping out part way through, and the file then fell out of the scan result and was deleted from the library, taking its album and artist with it if nothing else was left in them. Such a file is now kept as still present, since the scan just walked past it and knows it is there. Nothing is deleted, the rest of the deletion pass still runs, and one broken track cannot stop a provider from tidying up. A cue sheet that fails to parse keeps the track ids from the previous scan, otherwise the album would lose every track it describes.

-  stops an unreachable SMB server being reported as a login failure. Every problem during setup was reported that way, so a NAS that is simply switched off looked like a wrong username or password, which is easy to hit when the server boots faster than the NAS after a power cut. A login failure is now only reported when the server refused access.

-  fixes the error you get when a provider is reloaded while the server is down. The mount point is a local folder, but once a share has been mounted on it and the server goes away, even looking at that folder fails. Setup takes that as the folder not being there and then falls over trying to create it, so the user is told "File exists" about a folder in /tmp when the real problem is that their NAS is off. Setup no longer asks first, so the reload gets as far as the mount and fails with the real reason. The NFS provider had the same problem.

One thing left out on purpose: after a sync is aborted the provider is marked unavailable and recovers at the next scheduled sync, up to twelve hours later. Bringing that down with a shorter retry belongs with the sync scheduling rather than in the provider, so it is better as a separate change if we want it.

- related issue https://github.com/music-assistant/support/issues/5917

## Types of changes

## Checklist


music_assistant/providers/filesystem_local/helpers.py
```diff
@@ -28,6 +31,53 @@
...
+        :param err: The error raised while reading the directory.
+        :param is_root: True if the directory is the provider's root path.
+        """
+        if is_root:
+            self.fatal = err
+            return
+        self.failed_dirs += 1
+        self.consecutive_failures += 1
+        if self.consecutive_failures >= MAX_CONSECUTIVE_SCAN_ERRORS:
+            self.fatal = err
```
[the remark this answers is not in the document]
OzGav: The idea is correct but the fix is in the wrong place as this also affect cloud and WebDAV providers here so the fix was put in `_record_dir_failure`.

music_assistant/providers/filesystem_local/helpers.py
```diff
@@ -269,19 +326,18 @@ def recursive_iter(
...
             try:
                 item = next(scan_iter)
             except StopIteration:
+                scan_errors.record_dir_read()
                 break
             except OSError as err:
                 log.warning("Error while scanning directory %s: %s", path, err)
-                if is_root and scan_errors is not None:
-                    scan_errors.append(err)
+                _record_dir_failure(scan_errors, err, path=path, base_path=base_path, log=log)
```
MarvinSchenkel, line 406: Entry-level stat failures can still produce a "clean" scan that runs deletions — the same failure mode this PR fixes at directory level, one level lower. Two spots in `recursive_iter`:

- a non-EINVAL `OSError` from `item.is_dir()`/`is_file()` logs debug and `continue`s — if the entry was a directory, its whole subtree is missed with `failed_dirs == 0`
- a non-EINVAL `OSError` from `FileSystemItem.from_dir_entry` skips the file silently

Scenario: the storage blips while the walker is mid-iteration of one directory. `next()` keeps serving buffered entries but every stat fails, so all entries are skipped, `StopIteration` then counts the directory as read, the scan finishes "clean" and every skipped file lands in `deleted_files`.

Suggest recording these on `scan_errors` too (`failed_dirs += 1`, or a separate failed-entries counter that also blocks deletions).

OzGav: Good catch, both spots record now. `ScanErrors` gets a `failed_entries` counter next to `failed_dirs,` and deletions are skipped when either is non zero. `ENOENT` is not counted, since a file that vanishes mid scan really is gone. Entry failures do not feed the abort counter either, a failed directory listing still handles that.

The same hole exists one level up. `_process_item_async` catches everything and returns False, and the caller only adds to `cur_filenames` on True, so a blip while reading tags also drops the file into `deleted_files`. That one is more likely since it reads the whole file. It is handled differently though: a failed entry in the walk hides part of the tree we never saw, so blocking the pass is the only safe move, but a file that fails to process was found on the storage, so it is simply kept in the scan result. Nothing is treated as removed and the rest of the deletion pass still runs, so one broken track cannot stall a provider. A CUE sheet that fails to parse keeps the track ids from the previous scan, otherwise the album would lose every track. This covers cloud and WebDAV too.

As mitigation, when a scan does have to skip something, the sync is marked as failed and its closing message names what was missed, including up to five example paths, so a skipped deletion pass shows up in the task list instead of being silent. Files that fail to process are also listed individually by path.

MarvinSchenkel (approved):
Thanks @OzGav

PR #2356, opened 2025-08-26
Added Local Audio Source Provider

Torrax:
Note this plugin runs a raw PCM streams and will require optimizations in the MA FFMPEG to bring the buffering time closer to 1s to start the audio.

Currently without the optimizations, there will be roughly a 10-second start time to play/pause the audio. With optimizations, I was able to get this down to ~1 second (Player side buffer).

PR #5079, opened 2026-07-27
Adds a new Local Audio Source provider plugin

meiser79:

Adds a new Local Audio Source provider plugin

This PR is based on PR https://github.com/music-assistant/server/pull/2356, it updates the plugin to the current `AudioSource`-based plugin API (following `vban_receiver` as the reference implementation, per review feedback), switches the capture backend from ALSA to PulseAudio/PipeWire, and adds an optional signal-triggered auto-start mode for sources that are always connected (e.g. USB sound cards).

## Types of changes

## Checklist


meiser79:
But I guess, the new setup flow described at https://developers.music-assistant.io/setup-flows/ is missing.

OzGav:
This is great as its a top requested provider!

A quick first look shows a couple of things
- You already know about the new setup flow
- Large parts duplicate `local_audio/pa_simple.py` so you need to extract that out into shared helpers
- Narrow the suppress(Exception): blocks

Looking forward to seeing this merged!

edit: Please also follow this guidance "Use Sphinx-style docstrings with :param: syntax. For simple functions, a single-line docstring is fine. Multi line docstrings should have the first line on its own line to maintain indentation with the following. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally."

music_assistant/providers/local_audio_source/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "plugin",
+  "domain": "local_audio_source",
+  "stage": "alpha",
+  "name": "Local Audio Source",
+  "description": "Pipe realtime audio from a PulseAudio/PipeWire capture source (e.g. Bluetooth receiver or line-in) to any Music Assistant player.",
+  "codeowners": ["@music-assistant"],
```
OzGav:
```suggestion
  "codeowners": ["@meiser79"],
```

music_assistant/providers/local_audio_source/manifest.json
```diff
@@ -0,0 +1,14 @@
...
+  "credits": [
+    "Torrax",
+    "meiser79"
+  ],
+  "documentation": "",
```
OzGav:
```suggestion
  "documentation": "https://music-assistant.io/plugins/local-audio-in/",
```

music_assistant/providers/local_audio_source/README.md
```diff
@@ -0,0 +1,73 @@
+# Local Audio Source — the AUX-in cable for Music Assistant
```
OzGav: This whole file is not required here but should be used as a basis for a PR against the BETA branch of the docs. See the existing the pages for format and typical inclusions.

music_assistant/providers/local_audio_source/provider.py
```diff
@@ -0,0 +1,722 @@
...
+        # actually started — see the pending-claim timeout in _sensor_loop.
+        self._auto_trigger_pending_since: float = 0.0
+        self._sensor_task: asyncio.Task[None] | None = None
+        # Dedicated executor for all blocking libpulse-simple calls (pa_simple_new/
+        # read/close), kept off MA's shared default executor so a slow/stuck
+        # PulseAudio/PipeWire call from this provider can't starve unrelated work
+        # elsewhere in the server. max_workers=4 (not 1): a stuck read must not be
+        # able to block a queued close on the same single worker thread — see the
+        # blocking-call caveat on PASimpleRecordStream itself.
+        self._pa_executor = ThreadPoolExecutor(
```
OzGav: 6:26 code:comment is not a good ratio!

music_assistant/providers/local_audio_source/provider.py
```diff
@@ -0,0 +1,722 @@
...
+        # the sensor doesn't later try to stop a queue that isn't ours
+        # anymore. When _trigger_playback() itself is what led here, it
+        # already set _auto_triggered_queue to this exact queue_id before
+        # play_media() ran, so this check is a no-op in that case.
+        if self._auto_triggered_queue and self._auto_triggered_queue != queue_id:
+            self._auto_triggered_queue = None
+        # Record this request's session id so a later on_source_unselected can
+        # tell whether it is the live teardown or a stale callback from a
+        # superseded same-queue request.
+        self._active_session_id = stream_session_id
```
OzGav: Remove the comments.

music_assistant/helpers/pulseaudio.py
```diff
@@ -0,0 +1,123 @@
...
+PA_SAMPLE_S24LE: Final = 9
+
+PCM_FORMAT_TO_BIT_DEPTH: Final[dict[str, int]] = {
+    "u8": 8,
+    "s16le": 16,
+    "s16be": 16,
+    "s24le": 24,
+    "s24be": 24,
+    "s24-32le": 24,
+    "s24-32be": 24,
```
OzGav: The original `pa_simple.py` table mapped these to 32, and the commit message for the change says this "fixes" a reported value of 32 to the "correct" 24 significant bits.

Downstream in `sendspin_bridge.py`, `bit_depth` is used to compute `bytes_per_sample` and to decide whether to repack samples to a tightly-packed 3-byte layout before writing to PulseAudio. For `s24-32`, the sample actually occupies 4 bytes when sent (24 significant bits padded into a 32-bit container) so this repack path would now trigger for a format that isn't tightly packed, and local_audio_source's own format resolution appears to have the same problem. Can you clarify the intent behind the change, and whether the downstream byte-packing logic was accounted for?

music_assistant/providers/local_audio_source/provider.py
```diff
@@ -0,0 +1,719 @@
...
+        content_type = (
+            ContentType.PCM_S32LE
+            if bit_depth >= HIGH_BIT_DEPTH_THRESHOLD
+            else ContentType.PCM_S16LE
+        )
+        return AudioFormat(
+            content_type=content_type,
+            codec_type=content_type,
+            sample_rate=match["sample_rate"] or DEFAULT_SAMPLE_RATE,
+            bit_depth=bit_depth,
```
OzGav: This should be 32, not the source's original value, whenever `content_type` above is promoted to `PCM_S32LE` otherwise the two fields disagree, and the pause/silence keepalive (which sizes padding from `bit_depth` alone) inserts misaligned silence into the byte stream.

music_assistant/providers/local_audio_source/provider.py
```diff
@@ -0,0 +1,719 @@
...
+            self._sensor_task.cancel()
+            with suppress(asyncio.CancelledError):
+                await self._sensor_task
+        async with self._capture_lock:
+            await self._stop_capture_stream()
+        self._pa_executor.shutdown(wait=False)
+
+    async def get_audio_sources(self) -> list[AudioSource]:
+        """Return the single AudioSource this plugin exposes."""
+        return [self._build_audio_source(await self._resolve_audio_format())]
```
OzGav:
```suggestion
        return [self._audio_source]
```
I don't think you need to rebuild the `audio_source` information all the time. That's costing a lot in terms of spawning and killing `pactl`. This gets called on every Live Inputs browse (once per instance) and again on every playback resolution. You already have the info you need and it doesn't change. `get_stream_details()` already does the live resolve at the point it actually matters.

OzGav (dismissed):
Looks really good to me. Just a couple of items to review and you need to fix the failing CI. Thanks!

music_assistant/providers/local_audio_source/setup_flow.py
```diff
@@ -0,0 +1,101 @@
...
+                    required=True,
+                ),
+                ConfigEntry(
+                    key=CONF_THUMBNAIL_IMAGE,
+                    type=ConfigEntryType.STRING,
+                    default_value="",
+                    required=False,
+                    depends_on=CONF_ICON_PRESET,
+                    depends_on_value=ICON_PRESET_CUSTOM,
+                ),
```
OzGav, line 92: I see you've moved these here. The other entries in this flow pass `value=prefill.get(...)` but these two don't. On a reconfigure they will revert to default and then get saved back over the stored values. So someone coming here just to change the input device would lose their icon choice and thumbnail URL. Was this the plan?

Hedda:
FYI, MA developers have now added a "Sendspin Source" plugin/addon to Music Assistant server 2.10 BETA releases, see:

* https://beta.music-assistant.io/plugins/sendspin-source/

with this codebase for reference:

  * https://github.com/music-assistant/server/tree/dev/music_assistant/providers/sendspin_source

  * https://github.com/music-assistant/server/pull/5658

And the new Music Assistant server [2.10.0 RC 7](https://github.com/music-assistant/server/releases/tag/2.10.0rc7) release retire old local audio provider in favor of the Sendspin source plugin:

* https://github.com/music-assistant/server/pull/5965

meiser79:
@Hedda Could you help me understanding why you are posting this info here? I don't see any relation to this local audio source plugin, but maybe I'm wrong.

Hedda:
If use Sendspin audio provider for local-output then would it not also make sense to use the new Sendspin source for local-input? 🤔 

> Could you help me understanding why you are posting this info here? 

With this you could create a solution that uses a local Sendspin client with source role enabled on the device for local input.

Even if you do not want that then at least have another alternative reference to vban_receiver as reference implementation:

* https://github.com/music-assistant/server/tree/dev/music_assistant/providers/sendspin_source

and

* https://github.com/music-assistant/server/tree/dev/music_assistant/providers/vban_receiver

> Could you help me understanding why you are posting this info here? 

@meiser79 See and compare to this but compare and consider similar solution for audio input -> https://github.com/music-assistant/server/pull/5965

* https://github.com/music-assistant/server/pull/5965

Ping @chrisuthe and @marcelveldt

meiser79:
I'm working on a "local-audio-in-addon" using https://github.com/music-assistant/local-audio-addon as a reference.

MarvinSchenkel:
@meiser79 Awesome stuff! That's exactly the place where we position this architecturally (ie outside of MA). However, we do envision local audio in to be part of the existing addon. So before you start, please align with @chrisuthe to avoid any disappointments.

Are you on Discord by any chance? Would be great to align there.

meiser79:
Obsolete

Hedda:
> @meiser79 Awesome stuff! That's exactly the place where we position this architecturally (ie outside of MA). However, we do envision local audio in to be part of the existing addon. So before you start, please align with @chrisuthe to avoid any disappointments.

Please see this related feature request asking to add local audio input support to the new Music Assistant local audio addon:

* https://github.com/music-assistant/local-audio-addon/issues/23

PR #5088, opened 2026-07-27
Fix Audiobookshelf configuration loading

marcelveldt:

Audiobookshelf could fail to load or save its configuration because its library state was unavailable while determining provider capabilities.

- Initialize library state when the provider is created.
- Add coverage for capability detection before connection setup.

- N/A

## Types of changes

## Checklist

### DIFF
--- a/tests/providers/audiobookshelf/test_init.py
+++ b/tests/providers/audiobookshelf/test_init.py
@@ -1,0 +1,10 @@
+"""Tests for Audiobookshelf provider initialization."""
+
+from music_assistant_models.enums import ProviderFeature
+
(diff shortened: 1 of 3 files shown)


tests/providers/audiobookshelf/test_init.py
```diff
@@ -0,0 +1,10 @@
+"""Tests for Audiobookshelf provider initialization."""
```
chrisuthe, line 1: Should make the regression test exercise the empty-libraries pre-init state directly, rather than relying on the fixture's pre-populated library to reach the assertion.

PR #5118, opened 2026-07-29
Measure true peak during loudness analysis

MarvinSchenkel:

`AudioAnalysisData.true_peak` has always been stored as NULL. Two independent problems, either one enough on its own:

- The `ebur128` filter ran without `peak=true`. ffmpeg measures no peak unless asked, so the summary contains no `True peak:` section at all.
- `_TRUE_PEAK_RE` expected `dBTP`, but ffmpeg prints the peak in `dBFS` — the value would have been dropped even when present.

Neither surfaced, because `_finalize` guards with `if true_peak is not None`, so a permanently missing peak looks exactly like "this file had no peak data", and no consumer reads the field. On a real library, all 428 stored `loudness_analysis` rows had an integrated loudness and no peak.

This matters for peak-limited normalization: without a peak, static gain can push a quiet-but-hot item past full scale, and there is no measurement to clamp against.

- Request `peak=true` on the `ebur128` filter.
- Match the unit ffmpeg actually prints (`dBFS`).
- Bump `analysis_version` to 2 so v1 rows are re-measured instead of keeping their NULL peak forever.
- Add regression tests: one parses a verbatim ffmpeg summary, one asserts the filter requests peak measurement.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/loudness_analysis/provider.py
+++ b/music_assistant/providers/loudness_analysis/provider.py
@@ -33,7 +33,7 @@ CONF_WRITE_REPLAYGAIN_TAGS = "write_replaygain_tags"
 
 _INTEGRATED_RE = re.compile(r"Integrated loudness:.*?I:\s*(-?\d+(?:\.\d+)?)\s*LUFS", re.DOTALL)
 _LRA_RE = re.compile(r"Loudness range:.*?LRA:\s*(-?\d+(?:\.\d+)?)\s*LU", re.DOTALL)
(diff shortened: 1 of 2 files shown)


music_assistant/providers/loudness_analysis/provider.py
```diff
@@ -133,7 +133,8 @@ async def _start_analysis(
             input_format=audio_format,
             output_format=audio_format,
             audio_output="NULL",
-            filter_params=["ebur128=framelog=verbose"],
+            # ebur128 measures no peak unless asked
```
marcelveldt:
```suggestion
```

marcelveldt (approved):

PR #5120, opened 2026-07-29
Remove stale Audiobookshelf playback sessions to avoid repeated sync failures

foobarth:

## Problem
When ABS closes an idle playback session, sync_open_session() raises AbsSessionSyncError. The on_played() callback (fired every 30s during playback) catches it and falls back to direct progress updates, but keeps the stale session in self.sessions — so every subsequent report cycle retries the dead session and fails again, generating repeated log spam.
Problem with the naive fix
Removing the stale session on the first failure permanently demotes the user to the non-session fallback (update_my_media_progress). That fallback only updates the play position — it does not feed into Audiobookshelf's listening statistics (time listened, session history). For users who care about those stats, a transient sync failure (e.g. ABS briefly unreachable, or a session that times out but could be recreated) would silently disable stats for the rest of the playback.

## Fix
Track consecutive sync failures on the SessionHelper via a new failed_sync_count field, reset on success:
- Success → failed_sync_count = 0
- AbsSessionSyncError → increment the counter
- 5 consecutive failures → remove the session from self.sessions and log a warning with the attempt count
This way transient errors don't permanently disable ABS listening stats — only genuinely dead sessions (5 consecutive failures over ~2.5 minutes of 30s reporting) are dropped. A fresh session is automatically created when a new stream starts via _get_playback_session().
## Testing
Added test_session_sync.py with three cases:
- Session survives 4 consecutive failures (failed_sync_count == 4, still in sessions)
- Session removed on the 5th consecutive failure
- A successful sync resets the failure counter

All existing Audiobookshelf provider tests pass.

_None_

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -1240,12 +1240,19 @@ for more details.
                     ),
                 )
                 session_helper.last_sync_time = now
+                session_helper.failed_sync_count = 0
                 self.logger.debug("Synced playback session, position %s s.", position)
                 return True
             except AbsSessionSyncError:
-                self.logger.debug(
-                    "Was unable to sync session. Falling back to non-session approach."
-                )
+                session_helper.failed_sync_count += 1
+                if session_helper.failed_sync_count >= 5:
+                    self.logger.warning(
+                        "Unable to sync session %s after %s attempts - "
+                        "falling back to non-session approach.",
+                        session_helper.abs_session_id,
+                        session_helper.failed_sync_count,
+                    )
+                    self.sessions.pop(prov_item_id, None)
             return False
 
         if media_type == MediaType.PODCAST_EPISODE:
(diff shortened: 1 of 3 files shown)


OzGav:
Please adjust the title to remove the "fix:" portion. We just want the title to be a user facing clear description. The label(s) take care of the rest. (Same comment for your other PR as well)

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -1246,6 +1246,7 @@ async def _update_by_session(session_helper: SessionHelper, duration: int) -> bo
                 self.logger.debug(
                     "Was unable to sync session. Falling back to non-session approach."
                 )
+                self.sessions.pop(prov_item_id, None)
```
fmunkes: I would suggest to only remove the session after some consecutive failures. For a couple users the abs statistics are important, and the non-session sync does not count into them. Maybe give the SessionHelper an additional failed_sync_count, and e.g. after 5 times, we remove the session. In that case, I would also log a warning instead of a debug.
What do you think?

foobarth: Good catch there, fixed accordingly.

fmunkes (approved):
Thanks @foobarth!

PR #5131, opened 2026-07-29
Snapshot Sendspin playback progress at natural end of stream

chrisuthe:

`SendspinGroup.stop()` freezes the group's playback progress before it stops transport, but can only resolve the live position while the push stream is still up. On a natural end of stream `SendspinPlaybackSession` stops the push stream first, so that freeze re-emits whatever anchor was last pushed rather than the real end position. When another track follows, the next metadata push masks it.

Snapshotting at the point transport actually stops fixes it without reordering anything. `group.stop()` already calls `freeze_progress()` on every path it runs, so this adds no new metadata emission — it makes the existing one carry a real position. `snapshot_progress` is gated on the same condition as the `group.stop()` call below it, and the `is_stopped` guard means a stream that was already torn down can never be frozen.

The decision stays in the provider rather than the library because only the caller knows whether a teardown is final: `SendspinGroup.stop_stream()` exists for stopping transport mid-transition and deliberately preserves `PLAYING`, which is exactly the case where freezing would be wrong. A `stop_stream(snapshot_progress=...)` upstream in aiosendspin would be cleaner and would drop the reach into `player._metadata_role`, but it needs a library release against the pinned `aiosendspin[server]==7.0.0`, so it is deferred.

This is the same defect as #5129 at a second site. The two touch different files and are independent in either merge order — `SendspinPlayer.stop()` cancels the playback session before calling `group.stop()`, so this snapshot is inert on the user-STOP path. The condition is deliberately expressed at both sites rather than extracted, since each has its own answer to whether the teardown is a transition.

- https://github.com/music-assistant/support/issues/5813

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sendspin/playback.py
+++ b/music_assistant/providers/sendspin/playback.py
@@ -941,7 +941,11 @@ class SendspinPlaybackSession:
                     # and let the new playback handle the transition.
                     producer_stopped_cleanly = False
             with suppress(Exception):
-                self._stop_push_stream()
+                # Same condition as the group.stop() below, so we snapshot on exactly the
+                # paths where a group STOP - and therefore a freeze - is already emitted.
+                self._stop_push_stream(
+                    snapshot_progress=producer_stopped_cleanly and not self._cancel_requested
+                )
             await self._clear_join_catchup()
             await self._clear_member_pipelines()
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
LGTM

PR #5132, opened 2026-07-29
Add duplicate sound card support and ensure cards are not muted by alsa on startup

iVolt1:

Follow-up to PR #3724. Five improvements:

PA stream pre-warming (configurable) — PA streams opened at provider startup rather than at first play, eliminating cold-start stream-open latency (~50–200 ms) that causes sync offset in groups. Sinks remain RUNNING while the provider is active (intentional — warm streams ready for playback); they return to IDLE when the provider is disabled or MA stopped.

Because a held-open stream prevents idle suspend of the device (relevant to USB audio on virtualized hosts — see [needs verification: exact repo/ issue link for the hassio-audio-plugin module-suspend-on-idle report] for the failure mode this can aggravate), pre-warming is exposed as a config option (prewarm_streams, default on). Users on USB/VM setups who don't need tight sync-group alignment can disable it.

ALSA suspend/resume at topology creation — After creating remap sinks, runs pactl suspend-sink / resume on each ALSA master. Resets driver state.  Safe on all ALSA-card masters. 

Master sink volume pinned at 100% — module-device-restore can restore arbitrary volume on the master sink, stacking hidden attenuation on every zone sink. Pinned explicitly after topology creation.

Duplicate card support — Two identical cards previously caused remap sink name collision, silently skipping the second card's topology. Fix: detects duplicate alsa.card_name and appends the ALSA card index (_card{N}) to sink names and display names. Single-card setups unaffected.

ALSA playback-switch unmute at topology creation — On systems with multiple identical cards, the second-enumerated card can come up with some playback-enable mixer switches muted by default. The provider now forces all Playback Switch mixer elements on for each ALSA-card master at topology creation, via libasound directly (ctypes — no alsa-utils dependency). Writes are limited to boolean playback switches; volumes, capture and routing controls are never touched.

Note: Claude.ai wrote the PR description and I edited as needed.

## Types of changes

## Checklist


OzGav:
Please fix the template for the PR description and adjust the title to be a user facing explanation (plain english single sentence) of the changes here

chrisuthe:
Thanks for the follow-up + great work. Both bugs here are real, and I've no objection in principle to this
provider reaching into ALSA — it already suspends/resumes sinks and pins master volume. Two
asks, then some nitpicks.

### 1. The unmute writes much more than it diagnosed

`pa_simple.py:806-813` turns on *every* mixer element that has a playback switch, on every
discovery pass. Meanwhile your own comment at `sendspin_bridge.py:1049-1056` names the actual
culprits: the surround/center/side channel enables. On the Creative cards this targets, that
wider set also includes the analog input monitors (`Line`, `Mic`, `Aux`), so this can mix live
input into the output on machines that never had the bug.

**I'd like to see it approached as detect-then-correct:** read the playback switch on the
channel-enable elements the remap topology actually feeds, and turn on only the ones you find
off. That gives you one behaviour instead of four problems — it's a no-op on healthy cards, it
stops overriding deliberate mixer settings on every reload, and the code then documents its own
diagnosis. Worth checking first whether the mute is on the PA sink rather than the hardware
mixer; if so, use the sink mute you already have.

Could you send `amixer -c <N> scontrols` for an affected card, so we can size this on real
data rather than argument?

### 2. `player_id` depends on which cards are currently plugged in

`player_id` is `uuid5(ns, f"{sink_name}:{hostapi}")` (`sendspin_bridge.py:93-95`), and
`sendspin_bridge.py:1013-1018` appends `_card{N}` only while a duplicate is present. So adding
or removing one of two identical cards re-IDs the *other* card's players. Orphaned players lose their volume, DSP settings and sync-group
membership with no warning.

**I'd like to see it approached as a stable suffix, applied unconditionally, with a fallback
lookup:** derive it from `device.bus_path` instead of the card index (the master sink name
already carries it), apply it to every card rather than only to duplicates, and when resolving
a player fall back to the old un-suffixed name if the new key finds nothing. Unconditional plus
stable means the name never moves again; the fallback means no existing user is orphaned. Much
cheaper to do now, while the provider is `stage: alpha`, than after it reaches beta.

Could you also send `pactl list sinks | grep -E 'Name:|device\.bus_path'` to confirm bus_path
is populated on those cards? This is how @scyto approached it on our multi-sendspin-player container and it seems to have proven out pretty well. 

### Nits — none of these block

- **Tests.** `normalize_card_name()` (`remap_topology.py:54-70`) is pure and needs no
  PulseAudio. A test pinning the naming is also the
  tripwire for #2.
- **Errors are invisible.** The status string (`pa_simple.py:757-760`) is only ever
  debug-logged, and `set_playback_switch_all`'s return is dropped. `attach_failed` at debug is
  correct; `open_failed` / `register_failed` / `load_failed` are real anomalies.
- **ctypes convention.** `pa_simple.py:767-791` re-declares nine prototypes on every call;
  this file caches them once in `_load_lib()` / `_get_lib()` (`:69-104`).
- **`prewarm_streams` shape.** The stated rationale is sync-group timing, but the toggle is
  global — so disabling it for a USB-passthrough reason silently degrades sync groups. Could it
  be derived from group membership instead of a setting, so it's on exactly when a
  sync group needs it?

iVolt1:
@chrisuthe Thanks for the review.  I was actually finalizing some work on #2 when I saw you review.  It uses a 4 character short_hardware_tag() from the bus identity string e.g. "alsa_output.pci-0000_04_00.0.analog-surround-71" == cbf7.   "(USB)" is also added as suffix for USB devices.

image

#1 and the nits were also addressed.  I am testing these changes at the iVolt1/server@local_audio_hdmi_naming branch on my repo before I move them to this branch on this PR.

I will move the tests in tomorrow.

chrisuthe:
Hey @iVolt1- had some discussion internally with the sendspin and MA team at ohf about this and I want to urge us to go in another direction.

Right now I think we should pull the local audio out plugin from MA completely, and rely on either a CLI player installed locally or a docker container/HAOS add-on to expose local audio as a sendspin player. This means the hardware handling aspects are kept  out of Music Assistant so it stays as agnostic as it can be about what you're running it on. 

We've got the multi-sendspin container and HAOS add on that I put a good bit of effort into (Shoutout @scyto as well) that is in use at a good number of places, as well as an upcoming sendspin-CPP cli that mirrors the usefulness of the squeezelite CLI. The plan is to make a SUPER basic HAOS plugin out of that as well, so folks that just want local audio out can use that without any extra work, but the code stays out of the server.

Taking thoughts and opinions for and against of course, but I think this is a safer way forward..

iVolt1:
My main concern is that local audio should be easy for non-technical users since multi-room audio sounds so cool. I think the current implementation is easy where by default all possible players including, remap sinks, multichannel_stereo, and surround sinks are presented with descriptive names with the only limitations being sample rates and bit depth as provided by the back end which is separate issue which sounds like what you are trying to address.  I have seen many local audio how-tos that have so many steps that it is easy to get overwhelmed or make a mistake and just give up.

I did take a quick look at your work a few weeks ago and was not sure how it fit in, but it did look like it required a lot of configuration so I moved on, but it sounds like you are making it more plug and play.  I do wonder how latency will be with your approach.

With that said, I trust your team is much more immersed in the quickly evolving MA/HA sendspin architecture and I appreciate your assessing and moving forward with my PRs thus far.

My motivation is to once again have the same kind of local audio setup I was able to have where for spotify-connect where I took Alex Babel's Homepod Connect addon and used it with my own addon to enumerate remap sinks and attach a shairport-sync to them and control playback and volume from a HA media card.  It all synchronized very well with the actual airplay players as well. The media card was easy to share with family members and guests.  I abandoned that to get the same functionality in favor of MA and a year and a half later I am still trying replicate it and my family is growing impatient with me.

Thanks for asking for my input on this shift and I hope to see a stable local audio provider in soon!

OzGav:
The concern though it won't be easy to cover all combinations of hardware out there and we are opening up a huge support can of worms. If we can defer the hardware support to the operating system and sit sendspin on top of that then that might be the actual easiest path.

iVolt1:
Regarding supporting hardware, for HAOS it is plugin audio that provides the sound hardware drivers, albeit a currently limited list compared to most linux distributions. A while back I put in a request to have the snd_ctxfi driver added to the plugin and it was implemented a couple of releases later.  I considered requesting that all pci-e (pci might be might be a little old) cards with known good linux drivers be added.  I know HAOS tries to be a slim OS but a more complete list of sound car drivers should not add much bloat.

Hardware support should not be much of an issue for MA as any issues need to be resolved by the underlying OS.  If the MA team could work with the HAOS team to get sound hardware support expanded, that would be great.

scyto:
there are few moving pieces here

1. what the core HAOS team plan to do in terms of audio support - give the ask so many have had for HAOS to support sound cards and BT audio it seems like thats the right place to do the basic support of surfacing devices.
2. i think it is right to say by default it should be simple for default devices @iVolt1 i would push and say you never really looked at the multiroom audio - it has a wizard, for USB and PCIE devices its click click click and done - this could absolutely simplified down into -just surface all devices as-is and then advanced settings can be used to massage that - the biggest thing is when multichannel USB or pcie devices are used to provide multiple stero pairs - note surfacing say the 7.1 channels by default on a single PCIE device or USB device can can also confusion unless the device is very clear to MA and the OS which jack speakers are connected to by default.... i don't think this is a big issue
3. then there is the issue of those that want BT audio devices and the complexities of the OS wasn't designed for that

in terms of pros/cons

having hardware support in MA seems like it is good for both HAOS and running standalone in docker

HAOS subsystemms are something a sound container will fight with - both @chrisuthe and I have fought that battle over 9 months - and its never done, this is esp true for BT audio

I am initial inclination is that it would be great if HAOS could do the basics for audio subsystems for PCIE there are only really a few kernel drivers that would be needed if not included

i think an effort to get the multiroom audio container to be an 'MA audio container provider' has several advantages
one allows for all the work as default simplicity to be worked out outisde of the MA interation cycle and prove whats needed for simplicity @chrisuthe  am i missing something? I havent looked to see if you did anything on this front in moving the wizard to smart defaults, i know it was built when we didn't know what the smart defautls should be....

in parallel maybe there is something to do here with proposing that HAOS do more, for example it will need pipewire instead of PA and BLUEZ to support hi defintion BT codecs (I am blcoked on adding that suppotr to https://github.com/scyto/ha-bluetooth-audio-manager) as an example

as that progress then the decision could be taken to just merge whatever we feel is needed for a complete audio solution 

to argue against myself, the reason to not do this approach is if the standalone MA container (not in HAOS) is the primary scenario and a P1 deliverable - because then it becomes two approaches.... but i think thats unavoidable for most things when doing HAOS Supervisor and Standalone Versions of a container....

iVolt1:
Assuming local audio player provider continues to run on MA in more or less the current state and the multichannel audio piece can be completed I would have nothing more to add except for fifo output support which seems very doable.  On the HAOS side the audio_plugin needs two things configured and persisted, sound card sample rate/bit depth and sound card profile selection similar to how pavucontrol works.  Assuming a front end could be built for those in MA, my experience with trying to do things like just restart the hassio_audio container with a HA automation will fail due to lack of access. A linux host has pavucontrol or maybe even the linux distribution's sound/volume control should the user be aware of that.  I have not found a linux tool to make it easy to change sample rate/bit depth yet on PA or pipewire. 

@marcelveldt addressed some device recovery issue with #5325 which I don't really understand how the need was discovered what was done to actually address it.  As the local audio player provider currently works, reloading the plugin after a change in sound cards or profiles is generally required but a hot-plug functionality should be possible.

As the list of players in the screenshot above shows, non-multichannel and stereo pairs and multichannel_stereo players for surround cards are surfaced automatically just by the fact the provider is running.  I would want the multichannel_surround instances surfaced as well with their default 6 or 8 channel configuration. It should not be hard for new users to figure out the stereo pairs can be used as individual players for multiroom audio (basically a AmpliPi non-amplified player for a fraction of the cost) and multichannel stereo to play as one player for 3 or 4 receiver outputs and that multichannel_surround players can play multichannel sources to the correct channels.  Since PA is managing the streams it is possible for more than one stream to play to any of the player instances for overlays etc.

Hopefully future discussions with ohf can focus on the audio plugin's need for easier configuration persistence and sound card profile selection.

iVolt1:
I am guessing it may be bad form to keep adding improvements and features to an in process PR, but I have completed two important features in the branch attached to this PR and welcome any feedback and reviews provided.

1) The default player naming has been fine tuned to ensure uniqueness and a good sort order.  Any settings related to current sound cards will not be lost as that is all handled by a unique id in the background: 
image



2)  Pulse audio sound card profiles can now be set in the Local Audio Out player provider configuration screen Advanced options.  This eliminates a big pain point when trying set the correct card profile for surround, spdif, etc which required a lot of work at the command line on HAOS.  The profiles also include what card inputs are used so it will take some coordination when using those inputs in other apps or the Local Audio In provider. Most other Linux distributions have pavucontrol to handle setting profiles and more:
image



The remaining pain point to address is that the HAOS audio plugin does not retain customized daemon.conf settings (maximum sample rates and maximum bit depth primarily) over updates of the plugin or reboots of HAOS.  There is currently no obvious way to address that but retained daemon.conf settings could probably be done by the audio plugin devs as the audio plugin as of this week does allow a /data/custom.pa to add commands to system.pa. I currently don't have any uses for that functionality, especially if sound card profiles can be set in MA. Retaining higher maximum sample rates and maximum bit depth settings instead of the default 44100/16 will be a big win to a lot of users.

Hopefully the Local Audio Out provider will ready soon so multichannel output can be added.

iVolt1:
@OzGav  @marcelveldt @MarvinSchenkel   Can I please get a review or some feedback on this?  I am willing to convert this to a draft and create a new PR if you prefer.  

My local_audio_2 branch for this PR is feature complete and glitch free as far as I can tell after many days of testing.

The only things I know of on the roadmap after this PR are portaudio fixes as suggested in issue [#6029]  (https://github.com/music-assistant/support/issues/6029) and @vingerha 's #4919.

Hopefully the HA pulse audio plugin sample rate/bit depth persistence will be resolved soon by PR  [Add ability to add custom PulseAudio configurations](https://github.com/home-assistant/plugin-audio/pull/230#top).

Thanks.

vingerha:
Not sure if you read the comment in my PR, citing **Right now I think we should pull the local audio out plugin from MA completely**

iVolt1:
> Not sure if you read the comment in my PR, citing **Right now I think we should pull the local audio out plugin from MA completely**

@vingerha I think that was one person's opinion and no decision has been made.  It would be a big mistake to remove the local audio player provider.

vingerha:
> I think that was one person's opinion

This is not just a single person, this is a key developer for sendspin and I received similar feedback on Discord  .... But I agree it is vague....still I stopped spending time on it as it is quite an effort to rebase, effort I prefer to spend elsewhere. Would be nice if this is more transparent soon.

chrisuthe:
Addressed outside Music Assistant Core

PR #5151, opened 2026-07-30
Add Overcast podcast provider

OzGav:

Adds Overcast as a music provider, as requested in https://github.com/orgs/music-assistant/discussions/5027. Overcast is the 3rd ranked podcast app for iOS behind Apple Podcasts and Spotify.

Sign in with the Overcast email and password, and a user's subscriptions are added to the library along with their progress state. Finished episodes show as played, part-finished ones pick up where theye were left off.

Everything flows one way. Overcast gives other apps no way to write back, so nothing done here changes the Overcast account.

Overcast only permits downloading the subscription list about ten times a day, so it is saved and reused for four hours and refreshed in the background.

Two small pieces of shared code moved into the shared podcast code, since every podcast provider needs them. gPodder, iTunes and podcastfeed will be migrated in a follow up PR

## Types of changes

## Checklist


music_assistant/providers/overcast/provider.py
```diff
@@ -0,0 +1,457 @@
...
+                instance_id=self.instance_id,
+                domain=self.domain,
+            )
+        if newest_applied is not None and newest_applied != self._last_applied_ts:
+            self._last_applied_ts = newest_applied
+            await self.mass.cache.set(
+                key=CACHE_KEY_LAST_APPLIED,
+                provider=self.instance_id,
+                category=CACHE_CATEGORY_OPML,
+                data=newest_applied.isoformat(),
```
teancom: If I'm reading this right - and I could not be! - this sets 'newest_applied' globally, so if one feed has a transient error, its updates will be silently skipped the next time.

1. Feed A imports an update from `2026-07-30`; watermark becomes `2026-07-30`.
2. Feed B fails to download.
3. Feed B later recovers with an unimported update from `2026-07-29`.
4. The update is skipped because `2026-07-29 <= 2026-07-30`.

OzGav: You're reading it right, good catch. I'll fix this by keeping a separate watermark for each feed instead of one shared value, so a feed that fails a sync keeps its old watermark and its updates apply normally the next time it loads.

Looking into this also turned up a related issue where a synced state could overwrite a newer position you'd reached in MA itself, so I've added a guard for that here too. The underlying cause is in the core playlog write rather than this provider, so I'll do that fix separately.

music_assistant/providers/overcast/provider.py
```diff
@@ -0,0 +1,472 @@
...
+        remaining = self._rate_limited_until - time.monotonic()
+        if remaining <= 0:
+            self._rate_limited_until = None
+            return 0
+        return math.ceil(remaining)
+
+    async def _get_opml_subscriptions(self) -> dict[str, OvercastSubscription]:
+        opml_text = await self._fetch_opml_text()
+        if self._opml_cache is None or self._opml_cache[0] != opml_text:
+            self._opml_cache = (opml_text, parse_extended_opml(opml_text))
```
MarvinSchenkel: Let's put the helper call in an `await asyncio.to_thread` because it could be blocking.

MarvinSchenkel (approved):
Nice work Gav, thanks 🎉 !

fmunkes:
@OzGav Can you please ping me, when the PR for gpodder/ iTunes is ready - we have to make certain, that a library sync always updates the cache, no matter if it exists or not. Thanks!

PR #5156, opened 2026-07-30
A finished queue keeps its tracks so you can play it again

marcelveldt:

Reaching the end of a queue emptied it. That threw away any trace of what you had been listening
to, and left a finished queue looking exactly like one you never used, with nothing to tell you
whether pressing play would do anything at all.

A queue that plays through now stays where it is and is marked as finished instead:

- The tracks stay in the queue, with the position on the one that finished, so you can still see
  what played. Pressing play starts the queue over from the first track.
- Anything with a natural end works this way, music or audiobook or podcast alike, so what happens
  at the end of a queue no longer depends on what you were listening to.
- Radio and other live sources are left alone. They have no end, so stopping one still leaves it in
  the queue ready to start again.
- Playing something new replaces a finished queue. Only an explicit "add to queue" keeps the old
  tracks, and then the queue continues with what you just added rather than replaying its last item.
- While a queue sits at its end, seek and next/previous are reported as unavailable, so clients grey
  them out and leave play as the one thing you can do.

With Autoplay on a queue normally never reaches this point, since Autoplay has already added what
comes next. It only ends when there is genuinely nothing left to add.

- Refs https://github.com/music-assistant/support/issues/5913

## Types of changes

## Checklist

Models: https://github.com/music-assistant/models/pull/339 (merged, released as 1.1.176)
Frontend: https://github.com/music-assistant/frontend/pull/2234


marcelveldt:
Both low-confidence notes were right, fixed in bd223e9:

- `play_index` cleared `ended` before validating the player and the item id, so a start that failed there left the queue not-finished while still parked on its last item — the next play press would have replayed that item instead of restarting. The flag is now cleared only once an item actually loaded.
- `_continue_ended_queue` left `next_item` at the `None` that `mark_ended` set, so adding a batch onto a finished queue reported nothing after the first added item until playback recomputed it. It refreshes it now.

Regression test for each.

PR #5161, opened 2026-07-30
Dlna manual ips

olifluous:

**In short:** if your DLNA speakers live on a different network than Music
Assistant (a separate VLAN or subnet — common in homes that isolate IoT
devices), MA currently can't find them, because the "hello, I'm here"
broadcasts speakers send don't cross network boundaries. With this change
you can simply type the speaker's IP address into the DLNA provider's
settings, and MA will contact it directly — no multicast reflection or
network scanning needed. (Your firewall must still allow the speaker to
reach MA's stream server for audio, as with any cross-network player.)

### Technical summary

Adds a "manual discovery IP addresses" config option (the shared `CONF_ENTRY_MANUAL_DISCOVERY_IPS` entry, as already used by Sonos, Chromecast, Bose SoundTouch and others) to the DLNA player provider, for
setups where the renderers sit on another VLAN/subnet and multicast SSDP
cannot reach them.

For each configured address, the central discovery controller sends a
**directed unicast SSDP M-SEARCH** (single UDP packet to `ip:1900`) during
its existing discovery cycle — immediately when the provider loads, then on
the regular 5-minute cycle. The reply's LOCATION header provides the device
description URL, and from there the response flows through the exact same
callback path as a multicast discovery. No multicast, no broadcast, no
network scanning; retry-forever comes free with the existing cycle.

### Why not provider-local, like Roku/Samsung WAM handle the same option?

`roku_media_assistant` and `samsung_wam` already accept manual IPs and
resolve them provider-locally — they can, because their devices sit on fixed, vendor-known HTTP ports (8060 / 55001), so an IP is enough to construct the device URL directly. Generic DLNA renderers have no fixed port or path for their description XML (every vendor differs), so the only universal way to resolve a bare IP is the SSDP M-SEARCH round-trip itself. Since SSDP for UPnP providers is centralized in the `DiscoveryController` (which already supports directed targets for the network-scan fallback), the unicast search lives there, behind a small optional provider hook
(`upnp_manual_discovery_addresses`, default empty).

Alternative considered: reading the config key generically in the controller for every `upnp_discovery` provider — rejected because Roku/WAM already own that key with different semantics; the opt-in hook keeps them untouched.

### Blast radius

- Only `dlna`, `roku_media_assistant` and `samsung_wam` use central UPnP
  discovery; only DLNA implements the new hook. Roku/WAM behavior is
  unchanged.
- Config-gated no-op: with no manual IPs configured (every existing user),
  the discovery cycle is bit-for-bit unchanged — the pre-existing discovery
  tests pass with assertions unmodified.
- With IPs configured: one unicast M-SEARCH per unique valid address per
  cycle, sent concurrently (so N addresses don't serialize N × the SSDP MX
  timeout inside the discovery lock).

### Error handling

- Addresses are validated as IPv4 (after whitespace strip) before searching; invalid entries
  (hostnames, IPv6, typos) log a warning naming the provider and are skipped — they never
  abort the cycle. `_run_upnp_search`'s error handler is additionally widened to `except
  (OSError, ValueError)` as defense in depth.
- IPv6 manual addresses are explicitly out of scope (warned and skipped).
- Unreachable-but-valid IPs simply don't answer and are retried next cycle.

Note for reviewers: `except AddressValueError, AttributeError:` (no parentheses) is PEP 758 syntax, valid on the project's Python 3.14 floor and already used elsewhere in the codebase.

### Testing

- Unit tests: unicast target wiring, multicast/unicast dedupe (device answering both is
  dispatched once), invalid-address skip, config-gated no-op, DLNA config hook. `pre-commit 
  run --all-files` clean.
- Hardware validation on a Naim Mu-so Qb 2 across a firewalled/routed VLAN boundary (multicast 
  impossible), MA dev server on the other side:
  - Discovery via manual IP + unicast M-SEARCH: player registered.
  - Playback: flow stream served and audible; pause/volume commands work.
  - Power-cycle recovery: player marked unavailable by the poll, then
    rediscovered automatically on the next 5-minute cycle — at a *new*
    description port (the Qb's Rygel stack uses an ephemeral port per
    boot, 39288 → 54970 in our test). This is also why the config takes an
    IP and resolves via M-SEARCH: any stored description URL would break
    on every reboot.
  - Negative check: an invalid entry (`lounge.speaker.local`) logs
    `Ignoring invalid manual discovery address for DLNA: ... (only IPv4
    addresses are supported)` and is skipped; a whitespace-padded valid IP
    is stripped and works; the discovery cycle continues undisturbed.

### Disclosure

Developed with AI assistance (Claude) per the OHF AI policy; every line
reviewed and owned by a human.

 - related discussion: addresses the DLNA part of: https://github.com/orgs/music-assistant/discussions/5189 

## Types of changes

## Checklist

- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked. -> none
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. -> none
- [x] I have raised a PR against the documentation repository targeting the main or beta branch as appropriate. (https://github.com/music-assistant/music-assistant.io/pull/823)


OzGav:
The PR title need to be a user facing sentence as to what this does

music_assistant/controllers/discovery/controller.py
```diff
@@ -415,7 +449,7 @@ async def on_response(discovery_info: CaseInsensitiveDict) -> None:
                 await async_upnp_search(on_response, search_target=search_target)
             else:
                 await async_upnp_search(on_response, search_target=search_target, target=target)
-        except OSError as err:
+        except (OSError, ValueError) as err:
```
marcelveldt, line 452:
```suggestion
        except OSError, ValueError as err:
```

On python 3.14+ this is valid syntax as well

marcelveldt:
I'm not sure if I'm a fan of adding the manual IP configuration in the discovery controller.
Discovery is about automatic discovery, not for manual connections to player IP's.
For all we know, there are (or will be) player providers that can only rely on dynamic data coming from SSDP announcements and then this manual IP configuration will give some false expectations.

Why not just add this as a (advanced) configuration option on the DLNA provider ? Like we do for Sonos and Cast.

MarvinSchenkel:
Agreeing with Marcel here as well. Have a look at manual ip config for the other providers and exclusively add this to the DLNA provider, instead of in the central discovery controller. Putting this PR to draft so we can keep track of things, feel free to mark it as 'ready for review' again when you want us to have another look 🙏

olifluous:
ok, give me some time, I will review how the others implemented this. Give me some time and i'll mark it "ready for review" when i've managed to do it in the same way.

MarvinSchenkel:
@olifluous How is it going? Do you need any help to get this over the finish line?

PR #5165, opened 2026-07-30
Library item counts now respect the user's provider filter

chrisuthe:

`library_count()` counted the whole table and ignored the current user's `provider_filter`, so a user with a filter set got a total that did not match the list they could actually see. The list path applies the filter (`_ensure_provider_filter` → `get_library_items_by_query`); the count path never did. What leaks is an aggregate row count per media type, not item data — the symptom is a count that disagrees with the list.

The provider filter is expressed in `_apply_filters` as a correlated `EXISTS` subquery over `provider_mappings`. That term-building moves into a `@final` private `_provider_filter_clause()` helper on `MediaControllerBase` — the second instance of the existing `_search_filter_clause()` idiom in that file — and the listing path plus all three `library_count()` implementations now call it, so the two cannot drift apart again:

- `base.py` generic count (tracks, playlists, radios, podcasts, audiobooks)
- `artists.py` count (`artist_type` / `favorite_only` / `album_artists_only`)
- `albums.py` count (`favorite_only` / `album_types`)

The extraction is behaviour-preserving: the generated SQL string and the bound params are byte-identical to the previous inline strings for every `provider_filter` × `in_library_only` combination, and `tests/test_library_sync.py` now pins the exact clause for each branch so that stays true. Each count only adds the term when `_ensure_provider_filter(None)` actually returns a filter, so a user without a filter keeps exactly today's SQL — including the bare `get_count(table)` fast path in the generic case.

### Genres

`GenreController` gains a `library_count()` override that keeps the pre-change behaviour. Genres are library-only items without `provider_mappings`, which is why `GenreController.library_items()` already overrides the base to skip both the provider filter and `in_library_only`; without the matching count override, a filtered user's genre count would have dropped to 0 while the genre list stayed full.

## Types of changes

## Checklist


marcelveldt (approved):
LGTM, thanks @chrisuthe !

PR #5167, opened 2026-07-30
Mount the NFS export as configured and scan the Subfolder inside it

chrisuthe:

The NFS provider's optional **Subfolder** setting only worked when the subfolder was itself a separate NFS export. `mount()` joined Subfolder onto the export path and mounted the combined path, but NFSv3's `rpc.mountd` only hands out a filehandle for a path that is itself exported — so `Export=/mnt/vault` + `Subfolder=Music` failed with access denied unless `/mnt/vault/Music` happened to be exported too. `strings.json` documents the field as "use if your music is stored in a sublevel of the export", which is exactly the case that did not work.

The pattern was copied from `filesystem_smb`, where CIFS genuinely does support a subdirectory in the UNC path — **SMB is correct and is left untouched here.** NFSv4 users were unaffected, since the v4 pseudo-filesystem makes subdirectories of an export mountable.

The root cause was one attribute doing two jobs: `filesystem_nfs/__init__.py` hands the `/tmp/{instance_id}` mountpoint to the base class as its `base_path`, so the mountpoint and the scan root were the same value and the subfolder had nowhere to go but into the mount source.

### What changed

- `NFSFileSystemProvider` now keeps the two apart. `mount_path` is the mountpoint and owns the mount lifecycle (the `mount` target, the `umount` target, the `ismount` diagnostic, `makedirs`); `base_path` is `mount_path/<subfolder>` and stays the scan/serve root the base class works from. The mount source is the export exactly as configured.
- The subfolder is validated with `is_safe_path(subfolder, mount_path)` before any mount, and checked for existence after mounting. Without that second check the failure was silent — `check_write_access` swallows every exception into a debug log and the sync then took the clean-but-empty branch, so the user got an empty library and no error. Both errors are translated and name only the subfolder, never the internal `/tmp` path.
- A one-off settings migration folds a stored `subfolder` into `export_path`, reproducing a byte-identical mount source, so installs whose subfolder happened to be a separate export of its own keep working across the upgrade (`instance_name_postfix` comes out identical too).

Since `get_absolute_path`/`is_safe_path` are evaluated against `base_path`, this also *tightens* path confinement: item paths are now confined to the subfolder rather than the whole export.

### Notes for reviewers

- **The migration is guarded by a one-shot `nfs_subfolder_migrated` marker, and has to be.** These migrations re-run on every boot, and unlike its siblings this fold cannot recognise its own output — a subfolder stored after this fix is byte-identical to a legacy one. Without the marker the next restart would fold a freshly configured subfolder into the export path and silently restore the very bug being fixed. The marker follows the existing `encryption_key_migrated` precedent and is claimed even when no instance needed folding.
- It is the first migration here that needs `decrypt_string`, because `export_path`/`subfolder` live encrypted in `setup_data`. That is why it runs in the post-encryption phase next to `migrate_provider_setup_data` rather than in `migrate()`, as a new sibling taking both callbacks so the existing hook's signature is undisturbed. Decrypt failures are caught per instance and skipped (as `_rotate_encrypted_values` does) so one unreadable value cannot abort config-controller setup.
- **Known and accepted:** the migration faithfully preserves currently-*broken* configs as broken — it cannot distinguish them from working ones without probing the NFS server. Affected users (including the reporter) reconfigure once with the now-working field, which is what they have to do today anyway. Worth a line in the release notes.
- **Why not try the joined path first and fall back to the export when it is refused?** That needs no migration and would self-heal affected installs on restart, which is a real advantage. It was still rejected: it picks the layout from a network outcome rather than from what NFS guarantees, so a *transient* first-attempt failure would silently mount the parent export instead — wider than the admin scoped, and invisible because scanning still works — and `base_path` would permanently mean the mountpoint on one path and a subdirectory on the other. One deterministic layout plus a one-off fold was preferred.
- Downgrading stays clean: older code joins nothing onto the folded `export_path` and mounts exactly the source it built before.
- The `controllers/config/**` files will trip the advisory provider-scope warning. The migration is the compatibility half of this same fix rather than an unrelated change, which is why it rides along.
- Deliberately **not** labelled `backport-to-stable`: a settings migration in a patch release is a different risk profile than the code fix on its own.

- related issue https://github.com/music-assistant/support/issues/5949

## Types of changes

## Checklist


chrisuthe:
Tested against Truenas 25.10.4 Config:
image

Working

OzGav (commented):
From a solution standpoint I think this is good. The migration until 2.13 might be a little long from what I have seen before but Marvin or Marcel will advise. The only thing is the conflict that has popped up now that #5192 moved the unmount code into a helper, that will need careful sorting first. I noticed that the helper needs to be given `mount_path` rather than `base_path`, because `base_path` can now point at the subfolder inside the mount, and the helper quietly does nothing if the path it gets isn't a mountpoint.

chrisuthe:
> From a solution standpoint I think this is good. The migration until 2.13 might be a little long from what I have seen before but Marvin or Marcel will advise. The only thing is the conflict that has popped up now that #5192 moved the unmount code into a helper, that will need careful sorting first. I noticed that the helper needs to be given `mount_path` rather than `base_path`, because `base_path` can now point at the subfolder inside the mount, and the helper quietly does nothing if the path it gets isn't a mountpoint.

Thanks for looking this over — and for catching the mount_path vs base_path trap in the helper. That was exactly right, and it's the part of this that would most easily have slipped through.

I've merged dev and reconciled with #5192. The provider now uses the shared helpers/mount.unmount() and its own unmount method is gone. Worth flagging how close that came to going wrong: the automatic merge resolved without complaint but kept base_path at all three call sites — so it would have imported fine, read fine, and then quietly done nothing, precisely the silent no-op you described. All three now pass mount_path.

One extra wrinkle the helper brought with it: it raises SetupFailedError when it can't free a mountpoint, where the old method only warned. At the post-mount cleanup — the one that runs just before the "subfolder not found" error — that would have swapped the actionable message for an unmount failure, so that call is now best-effort and suppressed. The pre-mount unmount still propagates, since a mountpoint that can't be freed before mounting is a genuine setup failure.

On the 2.13 marker: happy to shorten it to whatever Marvin or Marcel prefer. For reference I matched the newest marker already in that file — the PROVIDER_SETUP_FLOW_KEYS block is also 2.13. It's a two-line change, and the only residue after removal is an unused nfs_subfolder_migrated flag left in settings.json, which is harmless.

Also fixed the translation_owner omission Copilot flagged; without it both new errors fell back to the raw English string rather than resolving under provider.filesystem_nfs.

music_assistant/controllers/config/controller.py
```diff
@@ -81,7 +82,15 @@ async def setup(self) -> None:
         # one-off: move pre-setup-flow provider values into (encrypted) setup_data.
         # runs here, after encryption is initialized, so string values are encrypted
         # at rest (the migrate() pass in _load() runs before encryption is available).
-        if migrate_provider_setup_data(self._data, self.encrypt_string):
+        setup_data_migrated = migrate_provider_setup_data(self._data, self.encrypt_string)
+        # one-off: fold a stored NFS subfolder into its export path. Runs here for the same
+        # reason (both keys are encrypted setup_data), and after the move above so a legacy
+        # install's values have landed in setup_data by now.
+        # TODO: remove after 2.13 release
```
MarvinSchenkel: Why not 2.10?

music_assistant/controllers/config/migrations.py
```diff
@@ -386,6 +389,77 @@ def migrate_provider_setup_data(data: dict[str, Any], encrypt: Callable[[str], s
     return changed
 
 
+# TODO: remove after 2.13 release
```
MarvinSchenkel: Why not 2.10?

music_assistant/controllers/config/controller.py
chrisuthe: because I like to let AI make things up.. //fixed.

music_assistant/controllers/config/migrations.py
chrisuthe: because I like to let AI make things up...  //fixed.

MarvinSchenkel (approved):
Nice, thanks @chrisuthe 🎉

PR #5169, opened 2026-07-31
Pick a new go-librespot API port when the configured one is unavailable

OzGav:

The problem in the issue report may already be fixed on dev. The reporter is on version 2.9.10, but current dev has been changed to hand out each port only once.

While looking into the report, one remaining gap was found, each instance keeps the port it was given at startup and never picks another one. If that port is no longer free when the background process needs to start again (for example, because something else on the system took it in the meantime), the process fails to start, retries on the same port five times, and the plugin then shuts itself down with an error. That matches what the log in the issue shows.

This PR fixes that. Before starting the background process, the provider now checks whether its port is still free. If it isn't, the provider picks a new one and carries on. A conflict over a port now fixes itself on the next start attempt instead of taking the provider down. A test for this behaviour is included.

- related issue https://github.com/music-assistant/support/issues/5952

## Types of changes

## Checklist


marcelveldt (approved):
LGTM

PR #5180, opened 2026-07-31
Sync uploaded albums, artists and tracks from YouTube Music

whitty:

This adds further to the changes from #3156 . [As discussed](https://github.com/music-assistant/server/pull/3156#issuecomment-4301799341) the resolved albums could miss a reverse link in the database.

These changes simply enumerate the upload versions of the album track and artist APIs, which seems to resolve the links in the database.

As a result 
* I can browse uploaded albums and songs
* The song<-->album links are solid ([issue from #3156](https://github.com/music-assistant/server/pull/3156#issuecomment-4301799341))
* "Favourited" upload music now turns up in the favourited auto playlist

The only non-trivial change surrounds that bit of patching in `_get_album()` - I've added code to explicitly avoid the exception path and reproduce the side-effects of that path assuming it had succeeded.  I have no indication the default handling there is needed, from my account data, but I thought it prudent to include them, since someone had already felt the need to patch them in the other path.

## Types of changes

## Checklist

      Note `pre-commit` fails to run under `pyenv` which i use for isolation and python-version management.  The scripts break out of my venv and then complains the system python is wrong. 

### DIFF
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -54,14 +54,23 @@ async def get_album(
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album() -> dict[str, Any]:
+        private = False
         if prov_album_id.startswith("FEmusic_library_privately_owned_release"):
+            private = True
             ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
             album = ytm.get_library_upload_album(browseId=prov_album_id)
+
+            for album_track in album.get("tracks", []):
+                # note: side-effect of the album/playlist videoId xreference below is
+                # to default these values if not present, so I'm reproducing that here
+                # given we don't need to (and can't) do the query that wraps it
+                album_track["isAvailable"] = album_track.get("isAvailable", True)
+                album_track["likeStatus"] = album_track.get("likeStatus", "INDIFFERENT")
+
         else:
             ytm = ytmusicapi.YTMusic(language=language)
             album = ytm.get_album(browseId=prov_album_id)
-
-        if "audioPlaylistId" in album:
+        if "audioPlaylistId" in album and not private:
             # Track id's from album tracks do not match with actual album tracks. E.g. a track
             # points to the videoId of the original version, while we want the album version
             try:
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -59,14 +59,6 @@ async def get_album(
             private = True
             ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
             album = ytm.get_library_upload_album(browseId=prov_album_id)
-
-            for album_track in album.get("tracks", []):
-                # note: side-effect of the album/playlist videoId xreference below is
-                # to default these values if not present, so I'm reproducing that here
-                # given we don't need to (and can't) do the query that wraps it
-                album_track["isAvailable"] = album_track.get("isAvailable", True)
-                album_track["likeStatus"] = album_track.get("likeStatus", "INDIFFERENT")
-
         else:
             ytm = ytmusicapi.YTMusic(language=language)
             album = ytm.get_album(browseId=prov_album_id)
@@ -86,8 +78,6 @@ async def get_album(
             for album_track in album.get("tracks", []):
                 if playlist_track := playlist_tracks_by_title.get(album_track.get("title")):
                     album_track["videoId"] = playlist_track["videoId"]
-                    album_track["isAvailable"] = playlist_track.get("isAvailable", True)
-                    album_track["likeStatus"] = playlist_track.get("likeStatus", "INDIFFERENT")
         return album
 
     return await _run_ytmusic(_get_album)
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -172,17 +172,26 @@ async def get_library_artists(
 
     def _get_library_artists() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        artists = (
+
+        artists = []
+        # Avoid duplicate ids from overlapping sources
+        seen_ids = set()
+
+        for artist in (
             ytm.get_library_subscriptions(limit=9999)
             + ytm.get_library_artists(limit=9999)
             + ytm.get_library_upload_artists(limit=9999)
-        )
-        # Sync properties with uniformal artist object
-        for artist in artists:
-            artist["id"] = artist["browseId"]
+        ):
+            # Sync properties with uniformal artist object
+            artist["id"] = artist["browseId"].removeprefix("MPLA")
             artist["name"] = artist["artist"]
             del artist["browseId"]
             del artist["artist"]
+
+            if artist["id"] not in seen_ids:
+                artists.append(artist)
+                seen_ids.add(artist["id"])
+
         return artists
 
     return await _run_ytmusic(_get_library_artists)
--- a/music_assistant/providers/ytmusic/helpers.py
+++ b/music_assistant/providers/ytmusic/helpers.py
@@ -182,7 +182,11 @@ async def get_library_artists(
 
     def _get_library_artists() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        artists = ytm.get_library_subscriptions(limit=9999)
+        artists = (
+            ytm.get_library_subscriptions(limit=9999)
+            + ytm.get_library_artists(limit=9999)
+            + ytm.get_library_upload_artists(limit=9999)
+        )
         # Sync properties with uniformal artist object
         for artist in artists:
             artist["id"] = artist["browseId"]
@@ -201,7 +205,7 @@ async def get_library_albums(
 
     def _get_library_albums() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        return ytm.get_library_albums(limit=9999)
+        return ytm.get_library_albums(limit=9999) + ytm.get_library_upload_albums(limit=9999)
 
     return await _run_ytmusic(_get_library_albums)
 
@@ -231,7 +235,7 @@ async def get_library_tracks(
 
     def _get_library_tracks() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        return ytm.get_library_songs(limit=9999)
+        return ytm.get_library_songs(limit=9999) + ytm.get_library_upload_songs(limit=9999)
 
     return await _run_ytmusic(_get_library_tracks)
 


OzGav:
You can only choose one type of change, and to me it sounds like a bugfix actually. Also please improve the PR title. It should be a user facing single sentence that describes what this does.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -50,14 +50,23 @@ async def get_album(
...
+
         else:
             ytm = ytmusicapi.YTMusic(language=language)
             album = ytm.get_album(browseId=prov_album_id)
-
-        if "audioPlaylistId" in album:
+        if "audioPlaylistId" in album and not private:
             # Track id's from album tracks do not match with actual album tracks. E.g. a track
             # points to the videoId of the original version, while we want the album version
             try:
```
whitty, line 68: To further clarify with the change in #3156 the code was taking the `except` path here and returning.  Now it does not.

The mapping to correct the track list does not seem to be needed for uploaded albums.  The additional side-effects of setting `isAvailable` and `likeStatus` to defaults does not seem to be related to the mapping, so I reproduced it above.  It might be reasonable to move that to a common code-path.

All of that said - none of my queries are missing `isAvailable` or `likeStatus` so I don't know if the change in this function is needed at all.  I include it simply because I can't prove everyone's queries will return the same as mine.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -50,14 +50,23 @@ async def get_album(
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album() -> dict[str, Any]:
+        private = False
         if prov_album_id.startswith("FEmusic_library_privately_owned_release"):
+            private = True
             ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
             album = ytm.get_library_upload_album(browseId=prov_album_id)
+
+            for album_track in album.get("tracks", []):
```
MarvinSchenkel: This is already handled in `_parse_track` so this code seems redundant?

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -169,7 +178,11 @@ async def get_library_artists(
 
     def _get_library_artists() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        artists = ytm.get_library_subscriptions(limit=9999)
+        artists = (
```
MarvinSchenkel: This causes duplicates because `get_library_artists()` returns artist ids with a `MPLA` prefix while the others dont, so we need to:
- Use `artist["id"] = artist["browseId"].removeprefix("MPLA")` to make sure IDs are unique
- Also implement dedupe logic because the same artist can come from multiple function calls now

MarvinSchenkel:

music_assistant/providers/ytmusic/helpers.py
whitty: OK - should the code below which does the same thing also be removed?

lines 85-86

OzGav:
The title still needs work. People in the future are not going to find this if they search for YouTube for example.

music_assistant/providers/ytmusic/helpers.py
whitty: Implemented

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -54,14 +54,15 @@ async def get_album(
     """Async wrapper around the ytmusicapi get_album function."""
 
     def _get_album() -> dict[str, Any]:
+        private = False
```
OzGav, line 57: The private flag is only needed because the audioPlaylistId fix up on lines 65 to 79 sits after the if/else instead of inside it. That block only ever runs for non upload albums, so move it into the else branch under line 64. The flag on lines 57 and 59 and the and not private on line 65 all go away, and the two paths read for themselves.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -77,8 +78,6 @@ def _get_album() -> dict[str, Any]:
             for album_track in album.get("tracks", []):
                 if playlist_track := playlist_tracks_by_title.get(album_track.get("title")):
                     album_track["videoId"] = playlist_track["videoId"]
-                    album_track["isAvailable"] = playlist_track.get("isAvailable", True)
-                    album_track["likeStatus"] = playlist_track.get("likeStatus", "INDIFFERENT")
```
OzGav, line 81: From what I understand removing the `isAvailable` and `likeStatus` lines is not removing duplicate default handling. Both keys are already present on the album track and on the playlist track, and they belong to a video, not to the album entry. 

You keep swapping in the playlist track's `videoId`, so after this change the id comes from one video while the availability and like flags come from another. A greyed out track will look playable and a liked track can lose its favourite mark. 

I think you need to put those two lines back. If you have account data showing they are actually wrong, that is worth its own PR with the evidence, since it has nothing to do with uploads.

music_assistant/providers/ytmusic/helpers.py
OzGav, line 171: The docstring still says this wraps the ytmusicapi get_library_artists function, which is no longer what it does. Update it to say it returns the library artists from subscriptions and uploads, with duplicates removed.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -173,13 +172,26 @@ async def get_library_artists(
...
-        for artist in artists:
-            artist["id"] = artist["browseId"]
+
+        artists = []
+        # Avoid duplicate ids from overlapping sources
+        seen_ids = set()
+
+        for artist in (
+            ytm.get_library_subscriptions(limit=9999)
+            + ytm.get_library_artists(limit=9999)
```
OzGav, line 182: `ytm.get_library_artists` returns every artist that has a song in the user's library, not the ones they follow and nothing to do with uploads. Adding it here is a separate behaviour change from the fix your title describes, it will drop a large number of new artists into people's libraries on their next sync, and it costs another paged request on every sync. Please take it out of this PR. If you did mean to add it, it needs its own PR and a line in the description explaining the intent.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -173,13 +172,26 @@ async def get_library_artists(
...
+            + ytm.get_library_upload_artists(limit=9999)
```
OzGav, line 183: Uploaded artists now reach the library, but `get_artist` at the top of this file still hands the id straight to `ytm.get_artist`. An uploaded artist id is not a channel id, so that call falls through to the fallback and returns the name Unknown, and `get_artist_albums` and `get_artist_toptracks` in the provider both come back empty for the same reason. `ytmusicapi` has `get_library_upload_artist` for this case, in the same way `get_album` already has an upload branch. Did you open one of these artists in the UI, and what did you get?

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -173,13 +172,26 @@ async def get_library_artists(
...
+        ):
+            # Sync properties with uniformal artist object
+            artist["id"] = artist["browseId"].removeprefix("MPLA")
```
OzGav, line 186: Two things. First, add a short comment saying what the MPLA prefix is and why it is stripped, because that strip is the only reason the same artist coming from two endpoints collapses into one entry. Second, and more important, if an artist that already synced with an MPLA id now syncs with the stripped id, that is a change to stored user data and the old rows are orphaned. Please check which endpoint returns the prefix on your account and let us know if existing ids change, since that needs a migration rather than a silent switch

music_assistant/providers/ytmusic/helpers.py
OzGav, line 203: Same docstring point as above. It now returns library albums plus uploaded albums, so say that.

OzGav, line 233: Same docstring point as above.

music_assistant/providers/ytmusic/helpers.py
```diff
@@ -222,7 +234,7 @@ async def get_library_tracks(
 
     def _get_library_tracks() -> list[dict[str, Any]]:
         ytm = ytmusicapi.YTMusic(auth=headers, language=language, user=user)
-        return ytm.get_library_songs(limit=9999)
+        return ytm.get_library_songs(limit=9999) + ytm.get_library_upload_songs(limit=9999)
```
OzGav, line 237: Did you test what happens if an upload has no linked artist as I think that might cause a failure here?

OzGav (commented):
The PR title is describing the bug. It should be what this PR changes. for example "Sync uploaded albums, artists and tracks from YouTube Music"

OzGav:
Just marking as draft so we know which needs our attention. Please set to ready to review when you would like another look

music_assistant/providers/ytmusic/helpers.py
whitty: Sorry - I only implemented this because I was asked [here](https://github.com/music-assistant/server/pull/5180#discussion_r3703110378).  I do not know the reasoning, nor did I even know that detail at all.

On this entire PR - I just threw the hammer at everything and with the original set of changes (original PR) suddenly I had my uploaded albums and artists resolving.  I didn't do it through understanding the YTM API, I simply tried it out, saw that the metadata looked about the same, and that it worked.

Since then I've been chasing your requests with this PR.

All I do is apply the original patches via docker monkeypatching on my live HA and it gives me what I need (the rest of my library seems to work).  I don't know enough about the expected DB requirements, or what the ytmusic API does

music_assistant/providers/ytmusic/helpers.py
whitty: I'm not sure exactly what you're getting hat here.  When I navigate to an artist that is purely uploaded (a band of a personal friend disbanded at the start of the century who does not have an upload presence) I see a page with an icon (circle with two letters of the name) generated from the name, a list of the single album, and a list of the tracks of that artist (in alphabetical order).

If I navigate to the Album from there I see a list of tracks (in album track order this time) and a back-reference to the Artist

If I navigate to an individual track I see "Appears on" referencing the album (and "Similar tracks" courtesy of _something_).  From my naive understanding these look completely cross-referenced and complete.  My expectation is that youtube has produced appropriate IDs that are self-consistent with its own APIs (as long as the appropriate `get_album()` version is called based on the previous PR).

OzGav: Thanks for the detail. The page looks complete because the albums and tracks come from the Music Assistant library. They were linked during sync. That part of your change works and I am not disputing it.

The gap is in the provider itself. When Music Assistant asks the provider for an uploaded artist it goes through `get_artist` at the top of this file. That function sends the id straight to `ytm.get_artist`. That call only understands normal channel ids. For an uploaded artist it fails and the code falls back to an artist called Unknown with no tracks. Today that shows up as an empty top tracks call and two wasted requests every time the page is opened. It also means any future code that needs the provider to describe an uploaded artist gets nothing.

You solved exactly this for albums in #3156  with the `FEmusic_library_privately_owned_release` check in `get_album`. Please do the same here. Add a branch in `get_artist` for uploaded artist ids and use `get_library_upload_artist` from `ytmusicapi`. The ids for uploaded artists look like `FEmusic_library_privately_owned_artist_detail`... in the `ytmusicapi `docs. Please confirm that against your own account before you pick the prefix to check. That call returns the uploaded tracks for the artist and not a full artist page. So the branch needs to build the artist dict itself with the name and the track list in the shape the rest of the provider expects.

Without that the feature is only half done. Uploaded artists would be in the library but the provider could not answer questions about them. I would rather see it complete in this PR than merge it with a known hole.

music_assistant/providers/ytmusic/helpers.py
OzGav: No need to apologise. You did what was asked and you were right to. Marvin's comment answers the question I raised. The MPLA prefix only comes from `get_library_artists()`. The subscriptions list never had it. So nothing that is already in anyone's library changes id and no migration is needed. That part of my comment is resolved.

The only thing left on this line is a short comment so the next person knows why the strip is there. This would do.

```
# get_library_artists returns channel ids with an MPLA prefix that the other
# calls do not. Strip it so the same artist from two lists gets the same id.
```

One thing to be aware of. This strip and the dedupe below both exist because of the `get_library_artists()` call. I have asked in another comment whether that call belongs in this PR. @MarvinSchenkel should comment on that before you change anything here. If it goes then the strip and the dedupe go with it and this line becomes simpler again.

music_assistant/providers/ytmusic/helpers.py
whitty: This was in response to [this comment here](https://github.com/music-assistant/server/pull/5180#discussion_r3749645518)

I originally duplicated this code onto the "privately owned" branch (again because I still don't understand why its there), and was told it was not necessary, and I asked if this code should be similarly removed.

OzGav: Thanks for the link. That explains it and I can see why it looked the same. It is not quite the same though.

The lines you first added to the upload branch read a value from the track and wrote the same value back to the same track. The only effect was to fill in a default when the key was missing. Marvin was right that _parse_track already does that so those lines did nothing.

Lines 80 and 81 look similar but they read from playlist_track and write to album_track. Those are two different records. The line above them copies the videoId from the playlist entry onto the album entry. These two lines copy the availability and like status that go with that videoId. The True and INDIFFERENT are only the fallback if the playlist entry has no value. So they are not default handling. They keep the three values for a track in step with each other.

You asked whether those lines should go too and did not get an answer. Sorry about that. The answer is no. Please put lines 80 and 81 back as they were. Nothing else on this block needs to change.

music_assistant/providers/ytmusic/helpers.py
whitty: This was previously incorrect and I think changing it in that way might be the wrong direction.  This function (combined with the async inner) _implements_ the `get_library_artists()` contract against the ytmusic backend.  The fact it called `get_library_subscriptions()` should be by-the-by (assuming that `get_library_subscriptions()` was ever the right implementation of the abstract `provider::get_library_artists()`

Alternatively changing it to say "it calls X and Y" (and later Z) feels like the sort of "doesn't need to be a comment" comment

PR #5188, opened 2026-07-31
Apple Music: preserve a stored custom developer token across setup flow runs

LosCV29:

After noticing Apple Music is rate limiting playback, I signed up for a developer account. I applied the correct developer token in my MA account, but everytime I save it, it gets overwritten by the community shared token which brings back the rate limiting.

- related issue https://github.com/music-assistant/support/issues/5958

## Why

With the recent rate limiting of the shared bundled developer token (music-assistant/support#5955), a personal developer token is the effective escape hatch (music-assistant/support#5931) — verified: with a personal token active, all `Rate Limiter` backoffs, `Skipping unplayable item`, `produced no audio data` and HLS m3u8 fetch timeouts disappear, and flow-mode track transitions are clean again. But today the token cannot *stay* configured: any flow re-run silently reverts it (music-assistant/support#5958 — same defect in the 2.9.x releases, where `get_config_entries` unconditionally overwrites the stored value).

## Notes

- One extra validation request (`/v1/test`) is made only when a stored custom token differs from the bundled one.
- A natural follow-up (not in this PR, per #5931) would be offering the app-token field as an optional/advanced step even while the bundled token validates, so users can enter a personal token through the UI in the first place.

I've tested it locally and right now I have it applied. But a MA update will revert my fix.

## Types of changes

## Checklist


LosCV29:
Thank you for all of the work you guys all do. Its much appreciated!

LosCV29:
@MarvinSchenkel I had Claude help me submit this PR. I closed it on accident and just now reopened it. Can you please review? I would really like to get around this apple rate limiting for my music. Again... thank you for all you do! Please let me know if you need anything from me.

MarvinSchenkel:
> @MarvinSchenkel I had Claude help me submit this PR. I closed it on accident and just now reopened it. Can you please review? I would really like to get around this apple rate limiting for my music. Again... thank you for all you do! Please let me know if you need anything from me.

Are you sure this is still an issue on the `dev` branch? We introduced a setup flow there that should allow this. 

I saw that you fixed your 429 issue by disabling IPv6. I am currently wayyy deep in the 429 rabbithole, running a bunch of simulations to figure out what is causing this. Will keep you posted.

LosCV29:
> > @MarvinSchenkel I had Claude help me submit this PR. I closed it on accident and just now reopened it. Can you please review? I would really like to get around this apple rate limiting for my music. Again... thank you for all you do! Please let me know if you need anything from me.
> 
> Are you sure this is still an issue on the `dev` branch? We introduced a setup flow there that should allow this.
> 
> I saw that you fixed your 429 issue by disabling IPv6. I am currently wayyy deep in the 429 rabbithole, running a bunch of simulations to figure out what is causing this. Will keep you posted.

Thanks for getting back to me!! I really appreciate it. Pretty sure it's still there on dev, but it shows up differently now. The runtime part looks right — provider.py does prefer a stored custom token. The problem is the setup flow itself: it always goes with the bundled MUSIC_APP_TOKEN as long as that one validates, and only asks for a manual token if the bundle is expired. So if I've got my own developer token stored (which migrates over from the 2.9.x music_app_token field), the flow just ignores it. so whenever the flow re-runs to refresh the music user token, the MusicKit login happens under the bundled dev token — but user tokens are tied to whichever dev token requested them. So I end up with a user token minted under your token and a provider making calls with mine, and everything 403s. Not a clobber like on 2.9 anymore, but same end result. The PR just makes the flow use a stored custom token (if it still validates) for the whole run, so the pair stays matched, and falls back to the bundled one otherwise. Since the music user token expires every so often, I'm forced to re-run the flow regularly — so I'd probably hit this every time.

Also small correction on the 429s... since you're in the middle of that. I didn't fix mine by disabling IPv6, those turned out to be two different problems. The 429s went away completely once I switched to a personal developer token, so it really seems keyed to the shared token/fingerprint. The IPv6 thing was a separate issue with the audio CDN: over v6, Akamai would just never answer ffmpeg's second range request (the moov atom fetch) — debug logs showed "First chunk received after 116-226s" with zero errors because the reconnect option retries silently. Pinning aod-ssl/audio-ssl.itunes.apple.com to IPv4 in /etc/hosts fixed the stalls on the spot. Happy to dump my notes on either one if it helps your simulations.

I hope I properly explained this. Apologies for the back and forth. Again... many thanks for getting back to me.

MarvinSchenkel:
So I ran some simulations yesterday, and it seems we are getting throttled by Apple on **account level** (eg Team ID). Even if I generate a new dev token from the same private key, I am still getting throttled. Only using a developer token from completely different team / account resolves the issue. So that matches your findings as well.

The 'good' news is that we can quickly recover from a 429 (within 0.5s), so I attempted to fix that in [this](https://github.com/music-assistant/server/pull/5333) PR by tweaking the throttler. 

I'll have a play with this PR later to see what the best way forward is in terms of supplying a custom token 👍

LosCV29:
> So I ran some simulations yesterday, and it seems we are getting throttled by Apple on **account level** (eg Team ID). Even if I generate a new dev token from the same private key, I am still getting throttled. Only using a developer token from completely different team / account resolves the issue. So that matches your findings as well.
> 
> The 'good' news is that we can quickly recover from a 429 (within 0.5s), so I attempted to fix that in [this](https://github.com/music-assistant/server/pull/5333) PR by tweaking the throttler.
> 
> I'll have a play with this PR later to see what the best way forward is in terms of supplying a custom token 👍

Updated the branch to resolve the conflict with #5333 — adapted it to the new state check, so a throttled validation should no longer drop the stored token (same idea as your change).  the checks came back all green.

let me know if you need anythign from me, brother. Im just a vibe coder trying to contribute any way I can.... you guys are the REAL MVPs. Thanks for all your work and attention to this.

MarvinSchenkel:
Just did some digging. This issue is already fixed in the `dev` branch, so I will close this PR 🙏

LosCV29:
> Just did some digging. This issue is already fixed in the `dev` branch, so I will close this PR 🙏

You're the MAN! Thanks!!

PR #5207, opened 2026-08-01
Fix squeezelite players stuck unavailable after restart 

gdesmott:
I used `claude` to help me debugging and fixing this issue. I validated the patch and it fixes a real life bug I'm experiencing.

After a Music Assistant restart (add-on restart, upgrade, or host reboot), Squeezelite players reconnect at the socket level but are **never registered / marked available** — the dependent `universal_player`s stay `available: false` indefinitely. Recovery requires restarting the squeezelite client or MA again. Reproduces reliably on every restart.

## Root cause

An initialization-order race in the Squeezelite provider:

- The SLIMProto socket server is started in **`handle_async_init`** (`await self.slimproto.start()`), so it begins accepting connections immediately.
- The event callback is subscribed only later, in **`loaded_in_mass`** (`self.slimproto.subscribe(self._handle_slimproto_event)`).
- `aioslimproto` **does not buffer or replay events** — `signal_event()` iterates `self._subscribers`, which is empty until `subscribe()` runs. A `PLAYER_CONNECTED` fired in that window is dropped.
- There is **no reconciliation** of already-connected clients on load (base `on_provider_loaded` is an empty stub; the provider never enumerates `self.slimproto.players`). Registration happens *only* reactively in the `PLAYER_CONNECTED` handler.

On restart the reconnecting clients complete their `HELO` during that window, so their `PLAYER_CONNECTED` is lost. Because idle Squeezelite has no TCP keepalive and doesn't reconnect on its own, the player never recovers.

Verified on a live 2.9.10 instance — server log during the failing restart:

```
… Loaded player provider Universal Player          # wrappers registered, children not connected yet
… Starting SLIMProto server on port 3483
… Socket client connected: ('192.168.1.141', …)    # 4 pCPs reconnect
… Player connected: 2c:cf:67:b2:0c:79 (+3)          # HELO parsed  ← BEFORE the line below
… Loaded player provider Squeezelite                # provider loaded → subscribe() happens after this
  (no "Player available" for the squeezelite rooms — stuck)
```

Client-side confirmation while stuck: the pCP’s socket to `:3483` is `ESTABLISHED` (`/proc/net/tcp`) and `squeezelite` is running, yet MA reports the player unavailable.

## The fix

Reconcile already-connected clients in `loaded_in_mass`, right after subscribing — mirroring the `PLAYER_CONNECTED` branch of the event handler. Idempotent (guards on `get_player(...) is None`; `setup()` uses `register_or_update`).

### Why here (and not "subscribe before start")

Subscribing before `start()` would also close the window, but it lets the connect handler run during `handle_async_init`, before the provider is fully loaded/available — riskier. Reconciling in `loaded_in_mass` keeps registration where it already happens (provider ready) and is purely additive.

## Related issues

Fixes the deterministic restart case behind the "Squeezelite / slimproto player disappears" reports — e.g. music-assistant/support#5266, discussions `#5196` / `#1296`. Likely also explains intermittent occurrences (any `PLAYER_CONNECTED` that lands while the provider is momentarily unsubscribed is lost with no recovery path).

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/squeezelite/provider.py
+++ b/music_assistant/providers/squeezelite/provider.py
@@ -79,27 +79,29 @@ class SqueezelitePlayerProvider(PlayerProvider):
         # Validate ALL required ports before starting ANY services
         await self._validate_all_ports(control_port, telnet_port, json_port)
 
-        # Only proceed with server creation after all ports are validated
-        try:
-            self.slimproto = SlimServer(
-                cli_port=telnet_port or None,
-                cli_port_json=json_port or None,
-                ip_address=self.mass.streams.publish_ip,
-                name="Music Assistant",
-                control_port=control_port,
-            )
-            # start slimproto socket server
-            await self.slimproto.start()
-        except Exception as err:
-            # Ensure cleanup on any initialization failure
-            await self._cleanup_server()
-            raise SetupFailedError(f"Failed to start SlimProto server: {err}") from err
+        # create the server here (also validates config and sets up the CLI) but defer
+        # start() to loaded_in_mass, so we subscribe to events before accepting clients
+        self.slimproto = SlimServer(
+            cli_port=telnet_port or None,
+            cli_port_json=json_port or None,
+            ip_address=self.mass.streams.publish_ip,
+            name="Music Assistant",
+            control_port=control_port,
+        )
 
     async def loaded_in_mass(self) -> None:
         """Call after the provider has been loaded."""
         await super().loaded_in_mass()
         assert self.slimproto is not None  # for type checker
+        # subscribe before starting the socket server: aioslimproto does not buffer
+        # events, so a client connecting before we subscribe would be missed entirely
         self.slimproto.subscribe(self._handle_slimproto_event)
+        try:
+            await self.slimproto.start()
+        except Exception as err:
+            # ports were validated during setup, so a failure here is unlikely
+            self.unload_with_error(err)
+            return
         self.mass.streams.register_dynamic_route(
             "/slimproto/multi", self._serve_multi_client_stream
         )


gdesmott:
> Checklist item not ticked: pre-commit run --all-files passes.

getting this error but I did not touch this file so I guess it's unrelated to my change.

```
🐶 Ruff Linter...........................................................Failed
- hook id: ruff-check
- exit code: 1

S310 Audit URL open for permitted schemes. Allowing use of `file:` or custom schemes is often unexpected.
   --> scripts/check_package_safety.py:147:14
    |
146 |     try:
147 |         with urllib.request.urlopen(url, timeout=10) as response:
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
148 |             return json.loads(response.read())
149 |     except urllib.error.HTTPError as err:
    |

Found 1 error.
```

marcelveldt:
Did you actually test this on the current DEV branch instead of 2.9 stable ?

music_assistant/providers/squeezelite/provider.py
```diff
@@ -100,6 +100,17 @@ async def loaded_in_mass(self) -> None:
         await super().loaded_in_mass()
         assert self.slimproto is not None  # for type checker
         self.slimproto.subscribe(self._handle_slimproto_event)
+        # Reconcile players that connected during provider startup. The socket server is
```
marcelveldt: A simpler solution could maybe be to just start the slimproto socket server later.
We already check all the required ports as part of setup and we can also simply umload with an error if the start failed for some reason

gdesmott:
> Did you actually test this on the current DEV branch instead of 2.9 stable ?

I did not. I hot patched `provider.py` on top of `2.9.10` to test the fix.

music_assistant/providers/squeezelite/provider.py
gdesmott: Something like this then?

I just tested and that seems to fix my problem as well.

MarvinSchenkel:
@gdesmott Can you please confirm this is tested against dev instead of 2.9.10?

gdesmott:
Is there an easy way to install `dev`? I don't have a testing env so have to deploy this on my actual pi running HA and MA.

MarvinSchenkel:
> Is there an easy way to install `dev`? I don't have a testing env so have to deploy this on my actual pi running HA and MA.

Yep, have a look at our [dev docs](https://github.com/music-assistant/server/blob/dev/DEVELOPMENT.md). It's basically cloning the repo and executing our setup script. Then you can run MA locally on your development machine.

gdesmott:
Just tested my branch directly and it worked fine. :+1:

MarvinSchenkel (approved):
Thanks @gdesmott 🎉

gdesmott:
When will this change be released?

Do we need to wait for the `2.10` branch?

PR #5217, opened 2026-08-01
Report AirPlay speakers that stay silent while shown as playing

marcelveldt:

A speaker that never answers the server's clock cannot work out when to start, so it plays nothing. The session still establishes and audio keeps flowing, so Music Assistant showed it as playing indefinitely with nothing in the log to explain the silence.

The binary now reports that state, and it is surfaced as a warning naming the speaker, so the cause is visible instead of being invisible.

Needs the matching airplay-cli change: music-assistant/airplay-cli#41

- https://github.com/music-assistant/support/issues/5312

## Types of changes

## Checklist


marcelveldt:
Took the suppressed suggestion on `_parse_clock_ready`, both halves:

- The warning is gated on `mode != ntp` — an NTP-timed session has no clock of ours to answer, so it can't be a stall.
- A stalled line is treated as carrying no projection explicitly, rather than relying on the binary happening to send `ready_at_unix_ms=0`.

Behaviour is unchanged against today's binary (a stall is PTP-only and always numbered 0), but it no longer depends on that staying true. Test added for the NTP case.

PR #5228, opened 2026-08-02
Fix Flow Mode progress after abandoned stream probes

alectogeek:

Some players open the same Flow Mode URL multiple times while probing the stream. A probe can leave an unfinished, non-tail play-log entry before the connection producing audible playback appends the next item.

Flow progress mapping currently treats every unfinished entry as the active seven-day sentinel. That pins the queue title/index to the abandoned entry and lets elapsed time exceed the track duration.

This change treats only the unfinished tail as active. For an unfinished non-tail entry, it recovers the bytes-derived streamed duration from the shared queue item before mapping cumulative player time. Regression tests cover both the abandoned non-tail entry and the normal active-tail behavior.

- No public issue filed; reproduced against Kodi Flow Mode with duplicate HTTP consumers.

## Types of changes

## Checklist

## Local validation

- `pytest -q tests/controllers/player_queues --no-cov`: `219 passed`
- Ruff: passed
- mypy for the changed playback tracker and test: passed
- `pre-commit run --all-files`: passed
- `git diff --check`: passed

### DIFF
--- a/music_assistant/controllers/player_queues/playback_tracker.py
+++ b/music_assistant/controllers/player_queues/playback_tracker.py
@@ -323,13 +323,31 @@ class PlaybackTrackerMixin(_PlayerQueuesBase):
         played_time = 0.0
         queue_index: int | None = queue.current_index or 0
         track_time = 0.0
-        for play_log_entry in queue_data.flow_mode_stream_log:
+        flow_log = queue_data.flow_mode_stream_log
+        for log_index, play_log_entry in enumerate(flow_log):
             # seconds_streamed is bytes-derived stream-time, so the boundary check
             # doesn't need a speed factor. Only the still-streaming tail entry has
             # seconds_streamed=None; we'll break inside it before the sentinel matters.
             if play_log_entry.seconds_streamed is not None:
                 # NOTE: 'seconds_streamed' can be 0 if there was a stream error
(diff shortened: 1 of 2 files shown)


marcelveldt:
I know that some (DLNA) players indeed probe a stream once to get the contenttype or duration but so far I have never seen any side effects from that. Because the queue flow stream starts from the supplied start item it should not desync.

What issue does this exactly fix that you observed ?

alectogeek:
Yep, playback itself does not desync. 

After I fixed continuous playback for My Wave (Yandex Music), I observed the issue while using Kodi in Flow Mode. Playback continued normally and the tracks and titles changed, but the progress slider no longer reflected the position within the current track and effectively remained at 100%.   
                                                                                                                                                                            
I could reproduce it by starting My Wave from Browse and waiting for transitions between tracks. After applying this fix, under the same conditions, the slider resets at track transitions and shows the current position correctly again.                                                                                                           
                                                                                                                                                                            
So the issue I observed was specifically the incorrect progress display, not an audible playback failure or the queue starting from the wrong item.

MarvinSchenkel (approved):
Pushed 1 little lookup optimization. Other than that, this looks good to me. Thanks @alectogeek 🙏

PR #5230, opened 2026-08-02
Add party duration config option

lordhi:

Adds party_duration configuration to the party plugin, defaulting to the current 8 hours, but allowing for longer or shorter durations
This is to support events lasting longer than 8 hours, or being able to create the token in advance so that the QR code can be physically printed or otherwise prepared

Claude was used to generate the unit test.
(Which was then modified to be parametrized rather than having separate tests for different periods)

## Types of changes

## Checklist


lordhi:
Screenshots below of the config, with Advanced enabled to show the value.

Please note the Enable Guest Access button was not working in my local environment (before or after this change), and so local testing was done using curl
(This issue seemed to originate from the front end, as two separate requests were coming through, the second setting almost all party configs to None)

```
curl -s http://localhost:8095/api \
  -H "Authorization: Bearer eyJhbG..." \
  -H "Content-Type: application/json" \
  -d '{"message_id":"1","command":"config/providers/save","args":{"provider_domain":"party","instance_id":"party","values":{"enable_guest_access":true}}}'
```

image

image

I'm aware of [another pull request](https://github.com/music-assistant/server/pull/4912) that touches the same files from someone trying to fix a related issue, happy to either wait until that gets merged in or to assist with any conflicts that arise

music_assistant/providers/party/strings.json
```diff
@@ -15,6 +15,10 @@
         "__auto__": "Auto (last active player)"
       }
     },
+    "party_duration": {
```
MarvinSchenkel, line 18: Let's make clear here that the unit is 'hours'. Currently it just says '1'.

music_assistant/providers/party/__init__.py
```diff
@@ -186,6 +188,12 @@ async def get_config_entries(self) -> tuple[ConfigEntry, ...]:
                 default_value="",
                 required=False,
             ),
+            ConfigEntry(
+                key=CONF_PARTY_DURATION,
+                type=ConfigEntryType.INTEGER,
+                default_value=guest_access.DEFAULT_JOIN_CODE_EXPIRY_HOURS,
+                advanced=True,
```
MarvinSchenkel, line 195: Let's also and `range=(1, 168)` here to restrict the possibilities. You will also get a nice slider with this.

music_assistant/providers/party/__init__.py
```diff
@@ -129,6 +130,7 @@ class PartyConfig(DataClassDictMixin):
     # Custom party settings
     party_name: str | None
     qr_text: str | None
+    party_duration: int
```
MarvinSchenkel: Are you planning to expose this config to the frontend at a later stage? If not, we won't need to add it to the PartyConfig dataclass.

MarvinSchenkel (commented):
I left a few comments and then I think this is a good-enough enhancement to merge. 

In the future, we could maybe look into something like 'fixed QR codes with configurable expiration blocks' to make printed QR codes even more 'fixed', but I gotta think about the security aspect a bit more. 

Putting this in draft so we can keep track of what things need our attention. Please mark this PR as 'ready for review' again when you want me to have another look.

lordhi:
> I left a few comments and then I think this is a good-enough enhancement to merge.

Thanks! I'll fix these when I've got a chance.

> In the future, we could maybe look into something like 'fixed QR codes with configurable expiration blocks' to make printed QR codes even more 'fixed', but I gotta think about the security aspect a bit more.

My initial thought was being able to configure what the token would be, but that seemed like a can of worms security wise which is why I went with this as a low surface fix for the issue I was having
Fixed QR codes with different expiries would be fairly nice, although an ideal solution there would probably need an equivalent of the long-lived access token pane.

Might be worth a wider relook at how auth is handled in general for these, so that instead of a user being created and every time the QR code is scanned it logs into that user, the QR code being scanned generates a new user with details according to however the QR code is configured and maybe a mac address. Then one could block/ban bad actors, etc, although still imperfectly it'd be more than can currently happen.
(Although that's also all complete overkill for home use)

music_assistant/providers/party/__init__.py
lordhi: Is there a different frontend exposure outside of what I've already done?
It does currently display and can be edited on the front end, unless that's a different frontend from the settings menu?

lordhi:
- UI elements updated to clarify time is in hours
- Range has been added to change over to a slider, 1-168 hours as suggested (168 hours = 1 week)
- Party duration removed from guest frontend code as suggested

&nbsp;

- Translation build has been rerun
- Pytest passes locally
- UI has been retested locally with these changes (setting to 7 hours, saving, refreshing, checking it's changed, setting to 1, saving, refreshing)
- I have not retested the expiry logic after this round of changes as they shouldn't have any impact (and I'm still having the front end issue of not being able to enable guest access)

image
image

music_assistant/providers/party/__init__.py
lordhi: Read the manual and figured it out

MarvinSchenkel (approved):
Thanks @lordhi 🎉 !

PR #5273, opened 2026-08-03
Reuse the Home Assistant device list instead of re-fetching it constantly

marcelveldt:

Every time an ESPHome speaker connects or reconnects, Music Assistant asked Home Assistant for the complete list of devices — one full list per speaker, no batching. Speakers reconnect often (wifi blips, reboots, power cuts), so this repeated all day long, and a house full of speakers coming back at once meant one full list each.

Home Assistant has no compact device listing the way it does for entities (see #5270), so the list is now simply reused for a minute instead of being re-requested every time.

## Changes

- The device list is cached for 60 seconds instead of being fetched on every lookup.
- A burst of speakers connecting at the same time now shares a single fetch rather than one each.
- The player picker, startup and late-joining players all reuse the same list.

- follow-up to #5270

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/hass/__init__.py
+++ b/music_assistant/providers/hass/__init__.py
@@ -196,6 +196,7 @@ class HomeAssistantProvider(PluginProvider):
     _player_controls: dict[str, PlayerControl] | None = None
     _unsubscribe_controls: Callable[[], None] | None = None
     _unsubscribe_entity_registry: Callable[[], None] | None = None
+    _unsubscribe_device_registry: Callable[[], None] | None = None
     _engine_refresh_task: asyncio.Task[None] | None = None
     _ai_engines: list[AIEngine]
     _tts_engines: list[TTSEngine]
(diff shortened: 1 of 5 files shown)


marcelveldt:
Copilot is right here, and my test was hiding it - the fake cache stored synchronously, so the lock looked like exact single-flight when it is not.

Measured it properly with the store modelled as async (worker thread + db write, like the real one): a burst of 2, 3, 5, 10 or 20 concurrent lookups all end up at 2 fetches. So the lock does keep it flat instead of one-per-caller, it just does not reach one, because the caller right behind the first still finds the cache cold.

Fixed the comment and the tests to say that instead. Two fetches per burst instead of twenty is the win I was after, so keeping this as is.

Proper fix is single-flight inside use_cache itself (it has none today - only the stale-while-revalidate refresh is deduped). That is useful for all ~390 use_cache sites, not just this one, so it gets its own PR.

PR #5275, opened 2026-08-03
Allow multiple sessions in Audiobookshelf - fixes autoplay behavior

fmunkes:

The new autoplay behavior preloads another audiobook or podcast episode into the queue, if these media items are used. In the case of Audiobookshelf this means the creation of a new listening session. However, the provider currently only supports a single session, thus the newly enqueued item invalidates the current stream and session.

This PR changes:
- sessions can be created per media item by using device ids containing the media item's id
- a background task runs and closes unused open session. The "not-used anymore" is determined by the expiration set on the stream details, and the last occured sync with some safety offset
- get_resume_position does not create sessions anymore, and uses API calls outside of sessions instead

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -223,7 +223,7 @@ class Audiobookshelf(RecommendationPayloadMixin, MusicProvider):
         self.sessions: dict[str, SessionHelper] = {}  # key is the mass_item_id
         self.create_session_lock = asyncio.Lock()
         # see _get_playback_session for explanation
-        self.session_cycle = cycle(range(5))
+        self.session_cycle = cycle(range(10))
         base_url = str(self.get_setup_value(CONF_URL))
         username = str(self.get_setup_value(CONF_USERNAME))
         password = str(self.get_setup_value(CONF_PASSWORD))
@@ -1081,30 +1081,22 @@ for more details.
         self, item_id: str, media_type: MediaType
     ) -> tuple[bool, int, datetime | None]:
         """Return finished:bool, position_ms: int."""
-        # this method is called _before_ get_stream_details, so the playback session
-        # is created here.
-        session = await self._get_playback_session(mass_item_id=item_id)
-
+        # do not create a session here, is this method is called outside of stream acquisition
         item_ids = item_id.split(" ")
         abs_item_id = item_ids[0]
         episode_id = item_ids[1] if len(item_ids) == 2 else None
         progress = await self._client.get_my_media_progress(
             item_id=abs_item_id, episode_id=episode_id
         )
-        # only the progress object has a timestamp of the progress (not the session)
-        # last_update is in ms epoch
-        # If there is an open session, that session might have the old progress time,
-        # whereas the explicit progress call above gives the most recent time.
+        if progress is None:
+            # fallback to internal position
+            raise NotImplementedError
+        # The progress' last_update is in ms epoch
         timestamp = from_utc_timestamp(progress.last_update / 1000) if progress else None
-        current_time = (
-            progress.current_time
-            if progress is not None and progress.current_time is not None
-            else session.current_time
-        )
-        finished = current_time > session.duration - PLAYBACK_REPORT_INTERVAL_SECONDS
+        current_time = progress.current_time if progress.current_time is not None else 0.0
         self.logger.debug("Resume position %s: obtained.", current_time)
         return (
-            finished,
+            progress.is_finished,
             int(current_time * 1000),
             timestamp,
         )
--- a/music_assistant/providers/audiobookshelf/__init__.py
+++ b/music_assistant/providers/audiobookshelf/__init__.py
@@ -992,7 +992,7 @@ for more details.
 
             # Abs allows a single session per device id. MA may load another item into
             # the queue if autoplay is enabled. This then deletes the previous open session.
-            # We cycle through self.session_cycle to prevent this
+            # We cycle through self.session_cycle to prevent this.
             session_cycle_count = next(self.session_cycle)
             obsolete_session_id: str | None = None
             for id_, session_helper in self.sessions.items():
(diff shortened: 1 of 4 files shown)


music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -984,9 +990,24 @@ async def _get_playback_session(self, mass_item_id: str) -> AbsPlaybackSessionEx
...
+                if session_helper.cycle_count == session_cycle_count:
+                    obsolete_session_id = id_
+                    break
+            if obsolete_session_id:
+                # Remove old reference. Creating a new session with the same device id in
+                # abs will automatically close the session there, no separate call needed.
+                self.sessions.pop(obsolete_session_id)
+
+            client_name = f"Music Assistant {self.instance_id} {session_cycle_count}"
+            device_id = f"{self.instance_id}_{session_cycle_count}"
```
chrisuthe: Am I reading this right to say the cycle advances once per session *creation*, and a session is created for every candidate MA merely inspects — not just the ones it plays — so a single autoplay resolution can burn more than five slots and evict the session that is currently playing?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -354,6 +357,9 @@ async def unload(self, is_removed: bool = False) -> None:
         # run the unload chain first: RecommendationPayloadMixin cancels and awaits its
         # payload tasks, so no fetch is still running against the clients logging out below
         await super().unload(is_removed)
+        for session_helper in self.sessions.values():
+            with suppress(AbsError, AbsSessionNotFoundError):
+                await self._client.close_open_session(session_id=session_helper.abs_session_id)
```
chrisuthe: If I'm reading it correctly `suppress(AbsError, AbsSessionNotFoundError)` does not cover connection failures, so unloading against an unreachable ABS server stalls for up to five sequential connect timeouts and then raises out of `unload()`, skipping `logout()` and every `_on_unload_callbacks` entry.

`return_exceptions=True` guarantees the rest of `unload()` still runs, and going concurrent caps the wait at one connect timeout instead of five. `asyncio` is already imported, so nothing new comes into scope.

```suggestion
        # close the tracked sessions concurrently, so an unreachable server can neither
        # stall nor abort the unload below
        await asyncio.gather(
            *(
                self._client.close_open_session(session_id=x.abs_session_id)
                for x in self.sessions.values()
            ),
            return_exceptions=True,
        )
        self.sessions.clear()
```

Building the generator up front also removes a `RuntimeError: dictionary changed size during iteration` window — the original awaits with `self.sessions` live, and `on_played` can pop from it (`:1282`) without holding `create_session_lock`.

chrisuthe (commented):
Two things, one unintended consequence of object inspection and one unload scenario I'd change around if I'm understanding it correclty.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: Yes. The session is created in get_stream_details/ get_resume_position. If I play a normal playlist of e.g. audiobooks, MA only loads the next item, so this is fine.
However now that I tried, you are right, even a simple add to queue advances the counter, even if that item is played way later. So what I'm proposing here is not good enough...

While I could create a session per item, this will flood abs with open stale sessions, as I can only close them when I know for sure that the next MA call uses these methods again. And that is for example not true for pause. I don't have an idea yet.

Edited: it is only get_resume_position and this makes sense now. I have an idea. Session creation then only in get_stream_details.

fmunkes:
I resolved the above and changed the approach:
The autoplay, and collection handling both call `get_resume_position`. In the old implementation this created a session right there, and in `get_stream_details` this session was then used. You were completely right that this used the 5 slots too quickly. I updated the PR description to reflect what the PR does now. In short indefinite amount of sessions, which are regularly closed if they are not used anymore.

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -2166,3 +2188,40 @@ def _browse_recommendation_items(
...
             browse_items = list(self._browse_root())
         return UniqueList(browse_items)
+
+    async def _cleanup_open_sessions_loop(self) -> None:
+        """Close unused open sessions."""
+        while True:
+            await asyncio.sleep(STREAMDETAILS_EXPIRATION_S)
+            current_time = time.time()
+
+            sessions_to_close = [
```
OzGav: I might be misreading this, but the cleanup job make its list of “sessions to close” before it takes the lock? So if someone presses play at just the wrong moment, couldn’t the session get revived and handed to a brand-new stream, and then the cleanup job closes it anyway because it was already on the list?

music_assistant/providers/audiobookshelf/__init__.py
```diff
@@ -2166,3 +2188,40 @@ def _browse_recommendation_items(
...
+                (session_key, session)
+                for session_key, session in self.sessions.items()
+                if current_time - session.last_sync_time > (STREAMDETAILS_EXPIRATION_S * 2)
```
OzGav: What happens if I pause an audiobook and come back to it half an hour later? While it’s paused nothing is syncing, so after 10 minutes doesn’t the cleanup job think the session is abandoned and close it?

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: you're right, I moved it into the lock.

music_assistant/providers/audiobookshelf/__init__.py
fmunkes: If you pause, your are not using the StreamDetails. And they have a flag, that they expire after STREAMDETAILS_EXPIRATION_S, 5min. If you pause for less than 5min, MA will continue to use them. After 5 min MA will run get_stream_details again. Between 5 - 10 min (due to the safety offset), this will return an open session, but reset the last_synced (see get_playback_session, and this is while a lock is acquired), so it won't be closed anymore. After 10 min, the session was closed, so a fully new one is created.

OzGav (approved):
Thanks Fabian!

PR #5289, opened 2026-08-04
Make radio M3U export/import survive its own round trip

chrisuthe:

M3U export/import already existed for playlists and radios (#3387). The playlist path works; radio did not survive its own round trip — exporting the radio library and importing that same file back produced unplayable stations named after their URIs.

Two causes:

- **Import dropped the name.** Export writes it into `#EXTMA`, but import read only `item.title`, which `generate_m3u` never emits for a station because it gates `#EXTINF` on a known duration.
- **The MA URI was stored as a stream URL.** Import appended `builtin://radio/...` verbatim as a builtin `StoredItem` item_id, where builtin expects a plain stream URL. ffprobe failed, `get_library_radios` swallowed it, and every station — whatever provider owned it — became a dead `available=False` builtin row.

**Export** pages the library through `iter_library_items()` rather than a hardcoded `limit=10000`, carries the favourite flag, and writes unknown duration as `-1` so a station gets an `#EXTINF` line with its name.

**Import** runs as a background task, sequentially, and reports entries it cannot resolve through the task's failure channel instead of skipping them silently. It no longer casts to `BuiltinProvider`: entries resolve through `parse_uri` and are written with `mass.music.add_item_to_library`, so provider-owned stations restore against their own provider and builtin ones keep their raw stream URL as the stored item_id. Artwork falls out of builtin's `library_add`, and nothing is ffprobed on import.

Two shared-helper bugs the radio path depends on are fixed at the root; both also affect playlist import. An entry with no `#EXTPROV` produced an **empty** `provider_mappings` set — and an item with no mappings never reaches `library_add`, leaving a permanently unplayable library entry — plus an item_id truncated to the last path segment. Separately, the media type silently defaulted to `track`.

`BuiltinProvider.export_radios` and `import_radios` are unused as a result and removed.

Related: https://github.com/orgs/music-assistant/discussions/4617 — the radio half of the export/import request.

## Types of changes

## Checklist


music_assistant/controllers/music/media/radio.py
```diff
@@ -244,3 +287,58 @@ def _parse_summary_row(self, db_row: Mapping[str, Any]) -> RadioSummary:
...
+        total = len(parsed_items)
+        for index, item in enumerate(parsed_items):
+            update_current_task_progress_from_index(
+                index, total, f"Importing station {index + 1}/{total}"
+            )
+            try:
+                await self._import_radio_item(item)
+            except MusicAssistantError as err:
+                label = (item.metadata or {}).get("name") or item.title or item.path
+                report_current_task_failure(f"{label}: {err}")
```
MarvinSchenkel, line 313: Should we add another `except Exception` here to catch network blips etc? Otherwise stations that come after the error will be aborted

chrisuthe: we 100% should

chrisuthe: Resolved in 574d1dc

MarvinSchenkel (approved):

PR #5294, opened 2026-08-04
Announce only the stable address of each IPv6 network

marcelveldt:

On networks using IPv6 privacy extensions the host adds a fresh temporary address every rotation, and the old ones stick around for a while before expiring. We announce every address we can find, so on a normal machine that list was 11 entries: the LAN address, a VPN address, two real IPv6 addresses, and seven rotating temporaries — six of which the system had already retired. The list is only built once at startup, so those entries also go stale as the day goes on.

Speakers and clients pick an address out of that list to reach us, so most of what we were offering them was unusable.

We now announce a single address per network per interface, which is the stable one.

- helpers: when collecting our own addresses, keep only the first address of each IPv6 network

On the test machine this takes the announced list from 11 addresses down to 4, keeping both real IPv6 addresses and dropping only the temporaries.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/helpers/util.py
+++ b/music_assistant/helpers/util.py
@@ -993,6 +993,7 @@ def _enumerate_ip_addresses(include_ipv6: bool) -> tuple[str, ...]:
             _sock6.close()
     # get all IP addresses of all network interfaces
     adapters = ifaddr.get_adapters()
+    seen_ipv6_prefixes: set[tuple[str, bytes]] = set()
     for adapter in adapters:
         for ip in adapter.ips:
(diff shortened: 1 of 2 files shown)


marcelveldt:
Closing this - the approach doesn't hold up. Keeping the first address of each /64 works on macOS, where the stable address is listed first, but on Linux the kernel prepends global addresses, so the first one is the newest temporary. That would leave us announcing an address the kernel drops after 7 days of uptime, with no stable one left - worse than the noise it was meant to fix. Handling it in #5293 instead, without touching the shared address helper.

PR #5299, opened 2026-08-04
Fix remote access failing to reach the Music Assistant API

marcelveldt:

Remote access reached Music Assistant's own API through the *advertised* address: either the base URL you configured, or the published IP address. Neither is guaranteed to work from the machine itself — a published address does not always exist on the machine (containers, or a router doing NAT), and a configured external URL sends the call out through DNS and your reverse proxy just to come back in. When that failed, remote access stopped working while everything on your own network kept playing fine.

It now uses the address the webserver actually listens on, the same way the Sendspin connection was fixed in #5267.

- Remote access reaches the API and its HTTP proxy on the webserver's own bind address, or loopback when it listens on all interfaces.
- With SSL enabled, the certificate is not verified for this one connection: it never leaves the machine, and a certificate is issued for your hostname rather than for a local address.
- The address rule shared with the Sendspin connection now lives in one place.
- Added tests for the address selection and for the caller.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/webserver/controller.py
+++ b/music_assistant/controllers/webserver/controller.py
@@ -114,6 +114,20 @@ def _get_publish_addresses(
     return addresses
 
 
+def _get_internal_connect_ip(bind_ip: str | None, publish_ip: str) -> str:
+    """
+    Return the IP address to reach a server running on this host.
+
+    :param bind_ip: The server's configured bind IP (None or a wildcard means all interfaces).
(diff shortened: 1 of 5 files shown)


marcelveldt:
Checked the suppressed comment on the new ws test: MagicMock does support async iteration (it just yields nothing), and the test logs no exception with DEBUG capture on. Leaving it as is.

PR #5303, opened 2026-08-04
Align Spotify app volume with MA player volume on connection

OzGav:

Fixes the volume jump on the first volume press in the Spotify app when using the Spotify Connect provider.

When a Spotify session started, the Spotify app showed the device volume as 100 even though the linked player was at a lower level. The first volume press in the app then sent a value close to 100 and the player volume jumped up.

The cause is in how we configure go-librespot. We set `external_volume` so the player owns the actual volume, but in that mode go-librespot ignores the `initial_volume` value we wrote to its config and simply reports 100 to Spotify. Nothing ever told Spotify the player's real volume, so the two sides disagreed until the first button press.

This PR sends the player's current volume to go-librespot when a session is claimed, and again if the session reconnects. The Spotify app now starts at the player's real volume and the buttons adjust from there. The `initial_volume` config value is removed since it was never applied.

- related issue https://github.com/music-assistant/support/issues/5965

## Types of changes

## Checklist


alanmcruickshank:
Snap - I've got a similar branch doing the same thing: https://github.com/alanmcruickshank/music-assistant-server/tree/review-issue-5965-proposal

I've confirmed my changes fix the issue in my dev setup. I'm testing this branch now...

alanmcruickshank:
@OzGav - yes confirmed this fixes the issue. Thanks for the swift resolution! 👍

OzGav:
Thanks for confirming!

MarvinSchenkel (approved):

PR #5304, opened 2026-08-04
Ensure login providers endpoint doesn't crash when HA provider has no config URL

OzGav:

Fixes a bug from the setup flow refactor that made login impossible on standalone installs with a Home Assistant provider configured. The `auth/providers` endpoint crashed with an HTTP 500 before the login form could load, blocking all login methods including username/password.

The HA provider now keeps its URL in setup data, but the auth controller still reads it from provider config. On standalone installs that returns `None`, tripping a bare `assert`. The fix reads the URL with `get_setup_value("url")`, which the hass provider itself already uses and which falls back to config entries, so add-on installs are unaffected. If no URL is available the HA OAuth option is skipped instead of raising, so built-in login always stays reachable.

I have not reproduced this live on a standalone install, but the traceback in the issue points at this exact assert, and I've added regression tests for both the missing URL case and the normal case.

- related issue https://github.com/music-assistant/support/issues/5987

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/webserver/auth.py
+++ b/music_assistant/controllers/webserver/auth.py
@@ -10,7 +10,7 @@ import secrets
 from collections.abc import Mapping
 from datetime import datetime, timedelta
 from sqlite3 import IntegrityError, OperationalError
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 import jwt as pyjwt
 from music_assistant_models.auth import (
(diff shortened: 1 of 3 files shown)


music_assistant/controllers/webserver/auth.py
```diff
@@ -1793,9 +1793,15 @@ async def _setup_login_providers(self) -> None:
...
-            assert isinstance(ha_url, str)
+            # the connection URL is collected by the HA provider's setup flow
+            # (with fallback to a config entry value for add-on installs)
+            ha_url = ha_provider.get_setup_value("url")
+            if not isinstance(ha_url, str) or not ha_url:
+                self.logger.warning(
+                    "Home Assistant provider has no URL configured, "
+                    "Home Assistant OAuth login is not available"
+                )
+                return
```
marcelveldt: We should create a public method or property in the HA provider to read the url instead of having external logic needing to fish for it themselves

OzGav: Yes better idea. The Home Assistant provider now has a public url property that returns the configured URL, or None if there isn’t one.

MarvinSchenkel (approved):
LGTM, thanks @OzGav

PR #5307, opened 2026-08-04
Remove duplicated config stub from the player controller tests

marcelveldt:

The player controller tests stubbed `get_raw_player_config_value` with the same hand-rolled closure in nine different places — every copy returned the same min/max volume defaults and the same `"auto"` fallback, and only differed in which extra config key it answered. This replaces those copies with one shared helper, so a change to the stub no longer has to be made nine times.

No test changed what it asserts.

- n/a

## Types of changes

## Changes

- Added a module-level `_player_config_stub()` factory that builds the config side effect from optional extra keys plus `min_volume`/`max_volume`.
- Used it in the shared `mock_mass` fixture and in eight test helpers/cases that each had their own copy.
- Left one stub with its own closure: `test_protocol_redirect_forwards_scaled_volume` deliberately answers a different `max_volume` per player, which is the very thing that test pins.
- Replaced the volume-limit and control key literals with the existing `CONF_*` constants.

## Checklist

### DIFF
--- a/tests/controllers/players/test_player_controller.py
+++ b/tests/controllers/players/test_player_controller.py
@@ -44,6 +44,8 @@ from music_assistant.constants import (
     ATTR_MUTE_LOCK,
     ATTR_PREVIOUS_VOLUME,
     CONF_AUTO_PLAY,
+    CONF_MAX_VOLUME,
+    CONF_MIN_VOLUME,
     CONF_MUTE_CONTROL,
     CONF_POWER_CONTROL,
     CONF_VOLUME_CONTROL,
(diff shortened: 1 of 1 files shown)


marcelveldt:
On the suppressed comment about the helper's position: leaving it where it is. `check_method_order` only looks at class methods under `music_assistant/`, so tests are out of scope, and the guideline is about class members in the production code rather than module-level test helpers. The helper is used by the `mock_mass` fixture at the top of the file, so parking it 2000 lines further down would only make it harder to follow — this file already keeps its other module-level helper (`_set_play_media_override`) next to its callers instead of at the bottom.

MarvinSchenkel (approved):

PR #5308, opened 2026-08-04
Stop logging spurious errors when an artwork request is cancelled

marcelveldt:

Whenever a client walks away from an artwork request (a page closing while covers are still loading, the app backgrounding), the server could log a scary `ERROR: Error doing task: ... exception in shielded future` with a full traceback — even though nothing actually went wrong that we hadn't already handled.

Image fetches, thumbnail generation and colour palette extraction all share one piece of work between everyone asking for the same artwork, so that one slow request does not become ten. On Python 3.14 the mechanism used to protect that shared work from a cancelled request started reporting its failure a second time, straight into the logs. The image proxy and the palette endpoint are the busiest places where requests get cancelled, so this was a frequent source of confusing errors.

Same fix as #5288 / #5295, now applied to the artwork caches.

- n/a

## Types of changes

### Changes

- Wait for the shared work instead of shielding it in the image, thumbnail and palette caches
- Added tests covering all three, each verified to fail without the fix

## Checklist

### DIFF
--- a/music_assistant/helpers/colors.py
+++ b/music_assistant/helpers/colors.py
@@ -270,7 +270,14 @@ async def get_palette(
         task_id=f"palette.{key}",
         abort_existing=False,
     )
-    return await asyncio.shield(task)
+    # wait for the shared extraction instead of awaiting it directly: a caller awaiting a
+    # task holds it as its fut_waiter, so cancelling that caller would cancel the
+    # extraction for every other caller too. asyncio.shield achieves the same, but as of
+    # Python 3.14 a cancelled caller makes it report the extraction's exception through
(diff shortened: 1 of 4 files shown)


marcelveldt:
On Copilot's note about the `asyncio.sleep(0)` in the first test: keeping it. Nothing in `get_image_data` yields before the wait for a non-http path, so both callers are parked on the shared fetch after one loop iteration - same shape as the already merged test for `get_ip_addresses`. And the `calls == [...]` assert would fail loudly if a caller had missed the shared fetch, so it can't pass vacuously. The other two tests do use an explicit event, because those paths await before the dedup.

PR #5322, opened 2026-08-05
Sonos speakers now recover on their own after a network interruption

MarvinSchenkel:

Sonos speakers never came back on their own after an event that blackholes established
connections (a firewall state table reset, a router reboot, a VLAN hiccup). Only restarting
Music Assistant brought them back.

The reconnect handler in `_connect` caught `ClientConnectorError`, but a blackholed
connection raises `aiohttp.ConnectionTimeoutError` — a `ClientError`, not a
`ClientConnectorError`. The exception escaped the `call_later`-scheduled reconnect task
(surfacing only as "Task exception was never retrieved"), so the retry chain was never
re-armed and the player stayed unavailable indefinitely.

- Widen the reconnect catch from `ClientConnectorError` to `ClientError`
- Also catch aiosonos' `CannotConnect` for websocket handshake failures
- Add regression tests for both cases

- https://github.com/music-assistant/support/issues/5979

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos/player.py
+++ b/music_assistant/providers/sonos/player.py
@@ -14,12 +14,12 @@ import time
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, cast
 
-from aiohttp import ClientConnectorError
+from aiohttp import ClientError
 from aiosonos.api.models import Container, ContainerType, MusicService, SonosCapability
 from aiosonos.client import SonosLocalApiClient
 from aiosonos.const import EventType as SonosEventType
 from aiosonos.const import SonosEvent
-from aiosonos.exceptions import ConnectionFailed, FailedCommand
+from aiosonos.exceptions import CannotConnect, ConnectionFailed, FailedCommand
 from music_assistant_models.enums import (
     IdentifierType,
     MediaType,
@@ -736,7 +736,7 @@ class SonosPlayer(Player):
             return
         try:
             await self.client.connect()
-        except (ConnectionFailed, ClientConnectorError) as err:
+        except (ConnectionFailed, CannotConnect, ClientError) as err:
             self.logger.warning("Failed to connect to Sonos player: %s", err)
             if not retry_on_fail or not self.mass.players.get_player(self.player_id):
                 raise
--- a/tests/providers/sonos/test_player.py
+++ b/tests/providers/sonos/test_player.py
@@ -1,0 +1,71 @@
+"""Tests for the Sonos player connection/reconnect handling."""
+
+import logging
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from aiohttp import ConnectionTimeoutError
+from aiosonos.exceptions import CannotConnect
+
+from music_assistant.providers.sonos.player import SonosPlayer
+
+
+def _make_player() -> tuple[SonosPlayer, MagicMock]:
+    """Create a SonosPlayer with mocked connection dependencies."""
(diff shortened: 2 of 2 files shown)


music_assistant/providers/sonos/player.py
```diff
@@ -736,7 +736,9 @@ async def _connect(self, retry_on_fail: int = 0) -> None:
             return
         try:
             await self.client.connect()
-        except (ConnectionFailed, ClientConnectorError) as err:
+        # ClientError (not just ClientConnectorError) since a blackholed connection
+        # raises a timeout error, not a connector error
```
marcelveldt:
```suggestion
```
no need for these verbose comments

music_assistant/providers/sonos/player.py
```diff
@@ -736,7 +736,9 @@ async def _connect(self, retry_on_fail: int = 0) -> None:
...
+        except (ConnectionFailed, CannotConnect, ClientError) as err:
```
marcelveldt, line 739:
```suggestion
        except ConnectionFailed, CannotConnect, ClientError as err:
```
python 3.14+ syntax

marcelveldt: NOTE: Isn't ClientError the base exception for the others as well ?

tests/providers/sonos/test_player.py
```diff
@@ -0,0 +1,69 @@
+"""Tests for the Sonos player connection/reconnect handling."""
+
+import logging
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from aiohttp import ConnectionTimeoutError
```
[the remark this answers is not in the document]
MarvinSchenkel: Checked this against the pinned aiohttp (3.14.1) — it *is* re-exported at the top level, and it is in `__all__`:

```
>>> import aiohttp; aiohttp.__version__
'3.14.1'
>>> hasattr(aiohttp, 'ConnectionTimeoutError'), 'ConnectionTimeoutError' in aiohttp.__all__
(True, True)
```

The tests import and run fine (5 passed). Leaving the import as-is.

music_assistant/providers/sonos/player.py
MarvinSchenkel: Dropped it in 928eaf6f — the comment is gone entirely.

music_assistant/providers/sonos/player.py
MarvinSchenkel: Two things:

**Syntax:** PEP 758 keeps the parentheses mandatory when there is an `as` clause, so that suggestion does not parse on 3.14:

```
$ python3.14 -c "try: pass
except A, B as e: pass"
SyntaxError: multiple exception types must be parenthesized when using 'as'
```

Unparenthesized only works without `as`, so I kept the parens.

**Re the NOTE:** no — `ConnectionFailed` and `CannotConnect` are aiosonos exceptions (`SonosException` → `TransportError` in `aiosonos/exceptions.py`), completely unrelated to aiohttp's `ClientError`. So all three are needed: `ClientError` catches what leaks straight out of `get_discovery_info` (the `ConnectionTimeoutError` in the issue), the other two catch what aiosonos wraps itself.

If you prefer fewer names, the two aiosonos ones share a base and could collapse to `except (TransportError, ClientError)` — that also pulls in `ConnectionClosed`, which seems harmless here. Happy to change it if you like that better.

marcelveldt (approved):

PR #5327, opened 2026-08-05
Keep AirPlay speakers on the same clock when they start at different times

marcelveldt:

When several AirPlay speakers play together they must all use the same clock,
or they slowly drift apart. Speakers that are grouped through Sendspin each
picked their clock on their own, at the moment their own playback started.

That is fine when they all start together, but not when they don't: a speaker
joining minutes later, or one that keeps its connection alive across a track
change while another reconnects, could end up on a different clock than the
rest of the group and drift audibly out of sync.

A speaker now takes the clock its group is already playing on, and only decides
for itself when it is the first one up.

- Resolve the clock once per group instead of once per speaker
- A speaker joining an existing group adopts that group's clock
- Moving a speaker to a group on the other clock reconnects it rather than
  reusing its connection, so it can actually switch
- Provider README updated for the group-wide behaviour

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/airplay/README.md
+++ b/music_assistant/providers/airplay/README.md
@@ -453,6 +453,12 @@ daemon reports it is serving, attaching them to its elected clock via shared
 memory. A sync group resolves that choice once and applies it to every member,
 so a group never mixes members on the shared clock with members off it.
 
+Sendspin-bridged players hold the same line across their Sendspin group. Their
+processes are spawned independently and can outlive several tracks, so a bridge
+adopts the choice a live group member is already running with and only asks the
+daemon when the group has no such member. That keeps members which start minutes
+apart, or which keep a warm process across a track change, on one clock.
+
 The official Music Assistant container runs as root, allowing the daemon to bind
 these ports. A custom container running Music Assistant as a non-root user must
 grant the binary `CAP_NET_BIND_SERVICE` (for example,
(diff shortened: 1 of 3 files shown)


marcelveldt:
Both suppressed comments were right — same hazard, two call sites. When the process isn't kept, `_use_shared_ptp` now goes with it, so a bridge that just left a group can't hand its old flag to a sibling resolving in between. That's the one case where the stale value is actively wrong rather than merely old.

Also tightened a related window on re-read: a superseded cold start could overwrite the record belonging to the process the newer start is spawning, so the record is now guarded on still owning the bridge.

Tests added for both, and mutation-checked (dropping either clear, or the guard, fails them).

marcelveldt:
Independent review caught two more, both real:

**The ownership guard I added on the previous push was wrong.** `mass.create_task` runs eagerly (`eager_start=True`), so the start task executes up to its first await *before* `_airplay_stream_start_task` is assigned — the guard therefore saw `None` on every cold start from idle and never recorded the decision, silently disabling the whole thing. Reverted to the unconditional record. The window it was meant to close needs a superseded start to resume before its own cancellation *and* the group to change in between; not worth subtle machinery, so I've dropped it.

**A RAOP member grouped with an AirPlay 2 member could never warm-reuse.** RAOP carries no clock decision by design, so comparing its `None` against a sibling's real flag failed forever — that member respawned its CLI on every track change and reported the cold start lead, which the group takes the max of, costing everyone ~2s per track. It now short-circuits: a process with no decision can't be on the wrong one.

Also cleared the record on the two remaining release paths (failed warm handover, abandoned transport) so the invariant actually holds everywhere.

Added a regression test that drives the real `_on_audio_chunk` → eager-task path — that's what would have caught the first one, since every other test hand-set the task handle.

PR #5329, opened 2026-08-05
Make the Mono output channels option work on all players

marcelveldt:

Selecting **Mono (both channels)** for a player's output channels did nothing for most players. It only worked by accident on players streaming FLAC or raw PCM over HTTP — on Sendspin, AirPlay, Snapcast, universal groups, the MSX bridge, and on any WAV/MP3/AAC stream, the setting was silently ignored.

While fixing that it turned out the existing **Left** and **Right** options play about 3 dB too quiet on those same players, so that is corrected in the same place.

- Mono now really mixes both channels together, on every player and stream format
- Left/Right/Mono no longer lose 3 dB of volume on players that need a stereo stream
- Mono recordings no longer play as silence when Left or Right is selected
- Mono is now correctly reported as "not bit perfect" in the stream details

Note for release notes: players already set to Left or Right will sound slightly louder after updating — that is the corrected level.

The `helpers/ffmpeg.py` change makes the existing surround downmix apply to single channel outputs too, so the center and surround content is not dropped by the pan. It cannot take effect yet for multichannel sources: we hand ffmpeg `-ac 6 -channel_layout stereo` for those, the layout argument wins, and the stream is read as stereo. That is pre-existing and tracked separately.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/streams/audio.py
+++ b/music_assistant/controllers/streams/audio.py
@@ -1210,21 +1210,25 @@ class StreamsAudio:
         channel_value = self._get_output_channels(player, player_id)
         source_channel = None
         channel_mix = ""
-        if channel_value == "left":
-            source_channel = AudioChannel.FL
-            channel_mix = "FL"
-        elif channel_value == "right":
-            source_channel = AudioChannel.FR
-            channel_mix = "FR"
-        elif channel_value == "mono":
-            # both source channels feed the downmix, so report ALL to keep the
-            # output from ever being presented as bit perfect
-            source_channel = AudioChannel.ALL
-            channel_mix = "0.5*FL+0.5*FR"
+        # a single channel source is already the downmix and holds no FL/FR to select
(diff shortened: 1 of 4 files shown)


marcelveldt:
Re the suppressed comment on the surround downmix: premise is right, and it is worse than described. For multichannel PCM we pass `-ac 6 -channel_layout stereo`, and the layout argument wins — ffmpeg reads a 6ch stream as stereo and the pan degenerates to an identity (`o0 = 1 i0 + 0 i1`). Verified against ffmpeg 7.1.3; with `-channel_layout 5.1` (or none at all) it resolves correctly.

That is pre-existing and independent of this PR — the `-channel_layout` line is untouched here, and the filter is a harmless no-op in that state rather than a failure. Keeping the `<= 2` gate because it is correct and needed once the layout is fixed. Tracking the layout bug separately.

PR #5363, opened 2026-08-05
Fix surround sound sources being misread as stereo

marcelveldt:

Every PCM stream was announced to ffmpeg as stereo, whatever its real channel count. That layout wins over the channel count, so anything wider than stereo was read at the wrong width: a surround track had its length (and the play time reported to scrobbling services) come out three times too short, and a raw multichannel source such as a VBAN stream configured above 2 channels played back garbled.

Found by Copilot on #5329, left out of scope there.

## Changes

- Name the channel layout only for mono and stereo, where the channel count leaves no doubt, and let ffmpeg derive it for anything wider
- Fold surround down to stereo with ffmpeg's own rematrix instead of a fixed pan expression, which dropped the rear channels on a standard 5.1 layout, mixed in the LFE and could clip
- Pick a stereo internal stream format for surround sources everywhere they are chosen, since no output format carries more than two channels — this also keeps loudness analysis measuring the stereo fold that is actually played

- n/a

## Types of changes

## Checklist

- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked. (n/a)
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. (n/a)
- [x] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. (n/a)

### DIFF
--- a/music_assistant/controllers/streams/audio.py
+++ b/music_assistant/controllers/streams/audio.py
@@ -1325,9 +1325,9 @@ class StreamsAudio:
         rate the player supports that is <= the source rate, so the source is never
         upsampled. The bit depth follows the source unless audio processing
         (crossfade, volume normalization, DSP) is active — those need F32 headroom
-        to avoid clipping/precision loss. Realtime AudioSource items skip all
-        processing and get a pure passthrough format (source rate/bit depth when
-        the player supports them).
+        to avoid clipping/precision loss. Surround sources are folded down to stereo.
+        Realtime AudioSource items skip all processing and get a pure passthrough
+        format (source rate/bit depth when the player supports them).
 
         :param player: The player requesting the stream.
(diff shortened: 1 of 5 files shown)


marcelveldt:
Re the suppressed comment about `_split_at_input` placement: leaving it as is. The private-methods-at-the-bottom rule is about methods within a class in `music_assistant/` — `scripts/check_method_order.py` only scans the package, not `tests/`. Test helpers in this file already sit next to the tests that use them (`_output_args`, `_FakeProc`, `_rms_db`), and `_output_args` landed that way in #5362 yesterday. Moving this one 700 lines down from its only two callers would make it harder to read, not easier.

PR #5418, opened 2026-08-06
Include WebDAV and cloud filesystem providers in background audio analysis

traverseda:

> AI-derived diagnosis (Claude), included per the project's AI Policy — not a substitute for the commentary below:
>
> Audio analysis is only ever produced reactively during playback for
> `webdav`, `filesystem_google_drive`, and `filesystem_onedrive` libraries.
> `_available_filesystem_domains()` gates the nightly background scan on a
> hardcoded `FILESYSTEM_PROVIDER_DOMAINS` tuple that only lists
> `filesystem_local`/`filesystem_smb`/`filesystem_nfs`. All five
> `LocalFileSystemProvider` subclasses are equivalent for scanning purposes,
> but three were never added to that list. Fix: derive the domain list via
> `isinstance(p, LocalFileSystemProvider)` instead of the static tuple.

Could instead hard-code more subclasses of LocalFileSystemProvider in the check list, but this seems more robust. Let me know if you want me to just hard-code webdav in there instead.

- none — checked GitHub search for existing issues/PRs mentioning this, found none

## Types of changes

## Checklist

>  *(ran `pre-commit run --files` against the two changed files: 16/17 hooks pass clean. The one failure, `mypy`, checks the whole repo regardless of which files are passed — it surfaced 125 pre-existing errors across 42 unrelated files (zvuk_music, sendspin, chromecast, airplay, etc.); zero errors in either file this PR touches.)*

> *(69/69 passed in `tests/controllers/streams/test_audio_analysis.py`, including the 3 tests updated for this change — they stubbed a filesystem provider with a bare `MagicMock()`, which no longer satisfies the new `isinstance` check, so switched to `MagicMock(spec=LocalFileSystemProvider)`. Only ran this file, not the full suite.)*

- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. *(n/a — no model changes)*
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. *(n/a)*
- [ ] I have raised a PR against the documentation repository. *(n/a — no user-facing doc/behavior change beyond the bugfix. Behavior now matches documented webdav behavior.)*


OzGav:
Thanks for the PR. The problem is the network traffic this would cause. There is an argument that WebDAV might be on someone’s local network so that would be ok. So I’m not sure about this. At the very least it would need to be an option with it turned off by default. Bear in mind that the on playback scan still runs. I would have to talk to the other guys about this.

traverseda:
image

Worth noting that current behavior directly contradicts the documentation here: https://www.music-assistant.io/audio-analysis/sonic-analysis/#when-analysis-runs

This was a pretty big surprise to me, who read that documentation while designing my system. I may have set up SMB instead if I had known.

>At the very least it would need to be an option with it turned off by default.

You do you, it's your project, but personally I find those kinds of little inconsistencies maddening when working with your project. My webdav host isn't on my LAN, this is still the correct behavior because I have internet access, and my giant 10_000 song archive is only ~70gb. It is much less surprising that it would scan filesystems over the internet than that it would just not do documented behavior, and I don't see how SMB, NFS, WebDav, or even google drive should change that.

OzGav:
I don't appreciate your tone. Its not surprising I have at least one word wrong in the documentation out of 168 pages. I will fix it.

OzGav:
I want to clarify my response. I understand the frustration of documentation not matching behavior, and you're right that this is something I need to fix. My reaction was mainly to the phrasing "you do you" and "maddening" came across as a bit dismissive given this is an open-source project and not a commercial product with a documentation team.

The feedback about the inconsistency is valid, and I'll update the documentation. I just appreciate keeping the discussion focused on the issue rather than making it feel personal.

traverseda:
I'm more frustrated that making the behavior match the documentation ended up in a bit of a rug pull, that and being told you'll have to talk to a committee about it honestly. What I want is clear communication about your preference and requirements. The preference for not scanning the users curated files over the live internet is a new and surprising preference that leads to a lot of extra work on my end. Most of it communications.

Here is what I'd like to do.

>(AI) filesystem_local, webdav, filesystem_smb, filesystem_nfs, filesystem_google_drive, and filesystem_onedrive each currently define their own get_config_entries() from scratch, and five of them return an identical hardcoded list of six shared ConfigEntry constants plus their own content-type label (SMB adds one extra cache-mode entry) — pure duplication, and exactly the kind of place a new provider can silently omit something. The fix: move that shared six-entry tuple (plus the new CONF_ENTRY_BACKGROUND_ANALYSIS_ENABLED, default_value=True) into LocalFileSystemProvider.get_config_entries() on the base class, then have every subclass's own get_config_entries() call entries = await super().get_config_entries() first and append/merge only what's actually provider-specific (SMB's cache-mode entry, etc.), so the shared set — including the new toggle — is inherited automatically instead of retyped; webdav and filesystem_google_drive/filesystem_onedrive (network/cloud-backed, not local disk) additionally swap the toggle for a dataclasses.replace(CONF_ENTRY_BACKGROUND_ANALYSIS_ENABLED, default_value=False) copy after calling super(), so those three default to scan-disabled while filesystem_local/smb/nfs default to enabled.

I want to do it this way to minimize duplicate code, and defensibly program so that future network filesystems automatically get that config entry. The current duplication seems very fragile and any changes to the base class could require you to edit all the subclasses, a situation to be avoided.

Would you prefer it if I used that technique to de-duplicate the configuration entries, or would you prefer it if I duplicated that config in each of the subclass's get_config_entries?

Let me know how you'd like to have this implemented and I'll see what I can do.

MarvinSchenkel:
Agreeing with @OzGav here that we should be careful with just blindly enabling background analysis for providers like One Drive and GDrive. I like your approach for the ConfigEntries, but I would be defensive and swap the defaults, so:

- The new config entry has a default of `False`
- The local, smb and nfs providers override this default with `True`

This way, any future providers will have the option available and the default is defensive.

traverseda:
Much bigger PR than I anticipated, will need to wait until the weekend to do this. Not something I can do on a lunch break.

Where WebDavFileSystemProvider inherits from LocalFileSystemProvider there's no clean way to set the default to False, but set LocalFileSystemProvider to True.

I'd propose that I create an new ABC called "FileSystemProvider" with the defaults you want set, and all the existing classes inherit from it instead of from LocalFileSystemProvider. This is a cleaner hierarchy, but is a larger structural change than I think any of us signed up for. All the existing signatures would remain the same, but it is a complete change to your class hierarchy to provide a clean FileSystemProvider ABC that LocalFileSystemProvider and others inherit from.

MarvinSchenkel:
I'd like to prevent stacking hierarchies simply for the reason of not duplicating a single config entry. I would solve it like so:
- Set the default to False in the LocalFileSystemProvider
- Override this default and set it to True in the local, smb and nfs providers

This leaves the default behaviour to False and makes opting in to background analysis a conscious choice without introducing an unnecessary hierarchy

traverseda:
>  * Set the default to False in the LocalFileSystemProvider
>  * Override this default and set it to True in the local, smb and nfs providers

What you're asking me to do here is set the default to False on LocalFileSystemProvider and set the default to True on local, which is also LocalFileSystemProvider. That's going to require a different mechanism for inheriting settings than `super().get_config_entries()`. Note that all other classes inherit from the real LocalFileSystemProvider.

traverseda:
Doing this properly, with the defaults how you want them and a proper class hierarchy, is a bit too much code touching on a bit too much for me to be able to review properly or stand by. Needs some deeper changes to the class hierarchy so that defaults propagate up the MRO properly, and I'm not the right person to get this done.

PR #5433, opened 2026-08-07
Fix Sonos connections lingering after a provider reload

marcelveldt:

Reloading either Sonos provider left its connections to the speakers open. The provider's own cleanup ran over a player list that is already empty by the time it runs, so nothing was ever torn down.

For S1 that meant a speaker's event subscriptions kept firing for players that had already been removed. For S2 the connection was never closed at all, so the websocket and its listener task stayed alive, and any timer still pending for that speaker kept running — a reconnect that was already scheduled when the provider unloaded would fire afterwards and bring the removed player back to life on a fresh connection.

- S1: unsubscribe from the speaker's events when a player is unloaded, and drop the cleanup loop that could never run.
- S1: do that under the same lock #5432 introduced, so a subscribe that is still in flight can't put the subscriptions straight back.
- S2: close the connection when a player is unloaded, and cancel its pending reconnect and AirPlay group restore timers.
- Tests for both.

The S2 change touches connection lifecycle and couldn't be checked against real S2 hardware, so it's covered by tests only and kept deliberately small.

- n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos/player.py
+++ b/music_assistant/providers/sonos/player.py
@@ -195,6 +195,19 @@ class SonosPlayer(Player):
             CONF_ENTRY_HTTP_PROFILE_DEFAULT_2,
         ]
 
+    async def on_unload(self) -> None:
+        """Handle logic when the player is unloaded from the Player controller."""
+        await super().on_unload()
+        for task_id in (
+            f"sonos_reconnect_{self.player_id}",
+            f"restore_airplay_group_{self.player_id}",
+        ):
+            # a timer that already fired lives on as a task under the same id,
+            # so both are needed to cover the pending and the running case
+            self.mass.cancel_timer(task_id)
+            self.mass.cancel_task(task_id)
+        await self._disconnect()
+
     async def volume_set(self, volume_level: int) -> None:
         """
         Handle VOLUME_SET command on the player.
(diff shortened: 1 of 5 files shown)


marcelveldt:
Good catch on the suppressed one — guarded the disconnect in `on_unload` so a speaker that can't be closed cleanly no longer aborts the unload and strands the provider's other speakers.

The wider issue behind it (`unregister()` awaits `on_unload()` unguarded, and the unregister loop in `unload_provider` sits outside its try/except) is a controller-level fix that affects every provider, so I'll do that separately rather than bolt it onto a Sonos bugfix.

marcelveldt:
On the second suppressed one — that window doesn't exist. `Player.on_unload` has no awaits in it and `cancel_timer`/`cancel_task` are both sync, so nothing can yield to the loop between the cancellations and `_disconnect()` setting `connected = False`. A listener can't get in there to schedule anything.

It's a fair thing to worry about though, since it's an invariant that's easy to break by accident, so I've pinned it with a test: drop an `await asyncio.sleep(0)` in between and it fails.

marcelveldt:
`CancelledError` doesn't derive from `Exception` — it's been a direct `BaseException` subclass since 3.8, so `except Exception` leaves cancellation alone. Checked it on the interpreter we run:

```
>>> issubclass(asyncio.CancelledError, Exception)
False
```

Leaving it as is — same shape the WiiM provider already uses for its teardown.

PR #5439, opened 2026-08-07
Say which type failed when JSON serialization fails

OzGav:

When `json_dumps` hit a value it could not convert, the handler returned the value unchanged. orjson then called the handler again with the same value, over and over, until it gave up with "default serializer exceeds recursion limit". That message does not say what went wrong or which value caused it. Cache writes run in a background task, so all you got was that one line in the log and no way to trace it back.

The handler now raises an error naming the type that could not be converted. The type name includes the module it came from, because several providers ship classes with the same names as our own models.

Two lines in the same function never did anything and are fixed as well. The check for dict value iterators compared a class against a string, so it was always false. `dict_values` was listed twice in the type check.

The `raise_unhandled` argument is removed. Nothing used it.

I wasn’t able to test this locally but the change is a straightforward improvement to error clarity and tests have been added.

- related issue https://github.com/music-assistant/support/issues/5974

## Types of changes

## Checklist


chrisuthe (approved):
Better Error Messages are a good thing :)

PR #5455, opened 2026-08-07
Fix slow Home Assistant logins when a provider fails to finish loading

OzGav:

A provider whose loaded_in_mass() raised stayed registered and available but never got its ready event set, so every caller waiting for it paid the full timeout - forever. Home Assistant ingress logins wait on it twice per connection, which is the reported 10+ seconds per page load.

The Home Assistant provider walked into this whenever its player control settings held a value that is not an entity ID (a leftover from an earlier config): Home Assistant refuses the whole state subscription over one bad entity ID, so registering the controls failed and took the ready. signal down with it.

- signal a provider as initialized/ready even when a post load step failed, since it stays registered and available either way
- drop values that are not entity IDs from the Home Assistant player control settings, logging which one was ignored
  
- https://github.com/music-assistant/support/issues/6003

## Types of changes

## Checklist


music_assistant/mass.py
```diff
@@ -1368,9 +1368,14 @@ async def _register_loaded_provider(
...
-            provider.initialized.set()
-            self.get_provider_ready_event(provider.domain).set()
+            try:
+                await provider.loaded_in_mass()
+            finally:
+                # the provider stays registered and available even when a post load step
+                # failed, so signal it either way: an event left unset makes every waiter
+                # pay the full timeout, on every attempt, until the provider reloads
+                provider.initialized.set()
+                self.get_provider_ready_event(provider.domain).set()
```
MarvinSchenkel: The `finally` sets the events and then re-raises, so `run_provider_discovery()` and the default name persistence never run after a failed `loaded_in_mass()`. That contradicts the premise of the change: the provider is announced as ready, but its post load work stops halfway. Several player providers register their players in `loaded_in_mass` (`bose_soundtouch`, `wiim` manual IPs, `hass_players`), and `run_provider_discovery()` calls `discover_players()`. One failing step there now costs the provider its discovery pass as well, while every waiter is told it is ready.

The raised error also only reaches the generic `create_task` warning, which prints the coroutine object and no provider name.

An `except` block fixes both, and it keeps the events unset when the task is cancelled at shutdown:

```python
async def _on_provider_loaded() -> None:
    try:
        await provider.loaded_in_mass()
    except Exception as err:
        # the provider stays registered and available even when a post load step failed:
        # an event left unset makes every waiter pay the full timeout until it reloads
        LOGGER.warning(
            "Error in post load step of provider %s: %s", provider.name, err, exc_info=err
        )
    provider.initialized.set()
    self.get_provider_ready_event(provider.domain).set()
    await self.run_provider_discovery(provider.instance_id)
    ...
```

OzGav: Good catch, you’re right on both counts. It now catches the failure and carries on, so the discovery pass and the name saving still run, and the warning names the provider. A shutdown still leaves the provider unmarked.

This slipped through because nothing tested that step failing. The existing tests call it directly rather than through the real registration path. Three tests now cover the normal case, a failure and a shutdown.

MarvinSchenkel (approved):
Nice find, thanks @OzGav

PR #5461, opened 2026-08-07
Keep a playback position of zero instead of treating it as unknown

marcelveldt:

A playback position of exactly `0` was read as "no position reported" instead of as the valid position it is (the very start of a track).

For a universal group this meant the group never adopted the position its member reported when that position was `0`, and kept the optimistic anchor it set itself when playback started. Members that serve the group stream from a buffer (Snapcast, for example) always report position `0` and let the timestamp carry the correction for their buffer delay — so those groups showed a position that ran a couple of seconds ahead of the actual audio for the whole session. The AriaCast receiver had the same issue and told its senders the position was unknown at the start of every track.

## Changes

- Universal group: adopt a member's position of `0` instead of discarding it, and take the position and its timestamp as a pair so an incomplete anchor is skipped.
- AriaCast receiver: report position `0` to senders instead of "unknown", and stop dropping a position of `0` coming *from* a sender (the camelCase/snake_case fallback ran on the value instead of on the key, so a `0` fell through and the stale position was kept).
- Tests for all three.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/ariacast_receiver/__init__.py
+++ b/music_assistant/providers/ariacast_receiver/__init__.py
@@ -699,12 +699,13 @@ class AriaCastReceiver(PluginProvider):
         if "album" in data:
             m.album = data["album"]
 
-        # Accept both camelCase (Android) and snake_case (spec broadcast) per interop rule.
-        # The fallback is on the key and not on the value, so a reported 0 survives.
-        duration = data.get("durationMs", data.get("duration_ms"))
+        # Accept both camelCase (Android) and snake_case (spec broadcast) per interop rule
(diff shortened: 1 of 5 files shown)


marcelveldt:
Good catch on the asymmetry, but I went the other way and reverted the duration change instead.

A duration of 0 and an unknown duration mean the same thing (`stream_metadata.duration or current_item.duration` downstream), so nothing is gained by putting a 0 on the wire — unlike a position, where 0 is a real value. Keeping the duration handling untouched avoids changing the wire format for external senders for no benefit. This PR is about the position.

PR #5462, opened 2026-08-07
Stop the dependency check from blaming a PR for a vulnerability it did not introduce

marcelveldt:

The dependency security check installs the environment first and only resolves the target branch about a minute later. Most of the ~300 packages are not pinned to an exact version, so a release published in that window makes the two disagree: the check then sees a version the target branch "does not have" and blames the PR for a vulnerability it never touched. That produces a hard, non-overridable failure on an innocent PR, only clearable by re-running the job.

The PR's own dependency set is now resolved right next to the target branch's, and a finding only gates when that resolution carries it too. The window shrinks from about a minute to a second, and anything left over errs towards reporting a finding instead of blocking on it.

- follow-up to #5424

## Types of changes

- [x] CI / workflow change — `ci`

## Changes

- Resolve both branches back to back in the audit workflow instead of the target branch alone (~2s extra, no second install)
- Gate a finding only when the PR's own resolution contains that exact package and version
- Keep the previous behaviour when that resolution is unavailable
- Fail loudly rather than silently clearing findings when a resolution cannot be read
- Reword the report line for findings a PR does not introduce, which no longer always means the target branch has them

## Checklist

### DIFF
--- a/.github/workflows/dependency-security.yml
+++ b/.github/workflows/dependency-security.yml
@@ -142,6 +142,7 @@ jobs:
           git show origin/$BASE_BRANCH:requirements_all.txt > base_requirements.txt
           echo "pip-audit" >> base_requirements.txt
 
+          # Step 1 leaves the PR's own requirements_all.txt in place, in sync or not
           cp requirements_all.txt head_requirements.txt
           echo "pip-audit" >> head_requirements.txt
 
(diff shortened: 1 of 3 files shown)


marcelveldt:
Good catch on the suppressed one - the status list in the module docstring still described `preexisting` as "the target branch has all of them too", which is no longer the full story now that a mid-run release can also land there. Reworded it to what the status actually means. Same reason the report line in the workflow got reworded.

PR #5464, opened 2026-08-07
Share My Wave and My Mix track requests between listeners

marcelveldt:

The Yandex Music "My Wave" and KION Music "My Mix" track fetches opted out of request sharing with `single_flight=False`, so several listeners asking for the same batch at the same moment each fired their own request at the service. The reason given was that these calls advance a rotor cursor and send a one-shot "radio started" event.

Neither reason holds up. The "radio started" event is already guarded by a flag, and both methods hold a provider-wide lock for the whole fetch, so callers were queueing up behind each other anyway. On top of that, the opt-out was never even reached once an entry had been cached — serving a slightly stale batch and refreshing it in the background always used a single shared request regardless.

So the opt-out was costing extra requests against a service that is quick to throw up a captcha under bursts, without buying anything. With no callers left for it, the option is gone and sharing is now the only behaviour.

- n/a

## Types of changes

Changes:

- Yandex Music My Wave tracks and recommendations, and KION Music My Mix tracks, now share one request between callers asking for the same thing at the same time.
- Removed the now-unused `single_flight` option from the caching decorator; requests explicitly asking to bypass the cache still fetch on their own.
- Updated the cache notes and the two provider tests to match.

## Checklist


marcelveldt:
On the suppressed naming comment: `my_wave` is what the whole KION provider uses internally — `MY_WAVE_PLAYLIST_ID`, `_my_wave_lock`, `_my_wave_seen_track_ids`, `_my_wave_batch_id`, `_my_wave_radio_started_sent`. Only the user-facing name is "My Mix". Renaming just this one method would leave it out of step with the seven names around it, so I'd rather rename the whole cluster in a separate pass than half-do it here. Pre-existing either way, nothing to do with this change.

marcelveldt:
On the suppressed bypass/deepcopy comment: correct in principle, but pre-existing and deliberately left alone in #5370 — the bypass caller getting the same object the store task serialises has been true all along. If anything this PR narrows it: the `single_flight=False` methods used to share that property and no longer do, so bypass is the last one left. Copying on every force-refresh would put a copy straight on the dynamic-playlist playback path, which isn't worth it for something nobody has hit. Leaving as-is.

PR #5474, opened 2026-08-07
Play announcements on muted speakers

marcelveldt:

Announcements (TTS, doorbell, alerts) behaved differently on muted speakers depending on how they were grouped:

- An announcement to a muted group or a muted synced pair was completely silent — every member stays grouped, so nothing was ever turned up and you heard nothing at all.
- A muted single speaker did play the announcement, but was left unmuted afterwards.

An announcement is meant to be heard, so it now always overrides the mute — and the speaker is muted again once the announcement is done.

Follows on from #5465, which fixed a single muted speaker being released from its group but did not restore its mute afterwards.

## Types of changes

## Checklist

## Changes

- Muted speakers are explicitly unmuted for an announcement instead of relying on a side effect of the volume change, so a muted group is no longer silent.
- The mute state is restored afterwards, for every speaker involved.
- Speakers that use volume-based (fake) mute announce at their real volume instead of at the muted zero.
- Added test coverage for announcements on muted speakers, groups and fake-muted speakers.

### DIFF
--- a/music_assistant/controllers/players/announcements.py
+++ b/music_assistant/controllers/players/announcements.py
@@ -112,6 +112,10 @@ class AnnouncementsMixin:
 
         async def _handle_cmd_volume_set(self, player_id: str, volume_level: int) -> None: ...
 
+        async def _handle_cmd_volume_mute(
+            self, player: Player, mute_control: str, muted: bool
+        ) -> None: ...
+
         async def _handle_cmd_power(
             self, player_id: str, powered: bool, skip_auto_play: bool = False
         ) -> None: ...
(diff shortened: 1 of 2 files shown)


marcelveldt:
Checked Copilot's fake-mute flag. The providers that actually need fake mute (mpd, alexa, fully_kiosk) have no native mute and all update their volume state inline, so they're fine. The ones that don't update inline (chromecast, sonos, snapcast) all ship native mute, so you'd have to deliberately pick fake mute over a working one. Leaving as is - if it turns up in an issue we'll fix it then.

PR #5479, opened 2026-08-07
Prevent Sonos S1 speakers from reconnecting after a provider reload

marcelveldt:

A Sonos S1 speaker that was already offline could reconnect itself after its provider was reloaded, leaving a connection behind that nothing ever cleans up. This is the leak #5433 closed, reachable through a route that fix did not cover.

The player controller can still be inside `await player.poll()` for a speaker while that speaker is being unloaded — the availability ping alone suspends for up to a second per speaker. When that poll finished, a speaker that had been marked unavailable read as "it is back" and was subscribed to all Sonos events again, with auto-renew, on a player the controller no longer owns. It needs the speaker to have been offline for longer than the resubscribe cooldown, so the window is narrow, but once it happens there is nothing left to ever clean those subscriptions up.

- ignore speaker activity once the player has been unloaded, matching the guards already on the other event entry points
- add tests for a poll that reaches an unloaded speaker, and for one that reaches a speaker that recovered

- follow-up to #5433

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos_s1/player.py
+++ b/music_assistant/providers/sonos_s1/player.py
@@ -802,6 +802,8 @@ class SonosPlayer(Player):
 
     def _speaker_activity(self, source: str) -> None:
         """Track the last activity on this speaker, set availability and resubscribe."""
+        if self._unloaded:
+            return
         if self._resub_cooldown_expires_at:
             if time.monotonic() < self._resub_cooldown_expires_at:
                 self.logger.debug(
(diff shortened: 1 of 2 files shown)


marcelveldt:
The `subscribe()` argument is evaluated eagerly, so the mock is called synchronously inside `poll()` — what's deferred is `create_task` on the already-created coroutine, not a second `subscribe` call. Ran it 40x with RuntimeWarning as error, no flake and nothing left unawaited.

PR #5481, opened 2026-08-07
Stop Sonos S1 from scanning for speakers after it is turned off

marcelveldt:

A Sonos S1 provider that was unloaded kept scanning the network for speakers every 30 minutes, forever.

Each provider load runs discovery twice (once while starting up, once right after loading). Every run armed a new "scan again in 30 minutes" timer on top of the previous one without cancelling it, and unloading the provider could only cancel the one timer it still knew about. The leftover timer stayed alive, and every time it fired it started another scan and armed itself again — so an unloaded provider never stopped. If the provider had been disabled or removed, those scans also re-connected to the speakers, which is the leak that #5433 fixed coming back through another route. Each reload left another one of these behind.

Discovery is now scheduled through the server's own timer helper under a fixed id, so a new run replaces the previous one instead of stacking. Unloading cancels a pending scan as well as one that already started, and a scan that is already talking to the network can no longer schedule itself again on the way out.

- n/a — found while reviewing the discovery/unload path after #5433

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos_s1/constants.py
+++ b/music_assistant/providers/sonos_s1/constants.py
@@ -116,6 +116,7 @@ SUBSCRIPTION_SERVICES = {
 }
 
 # Timing Constants
+DISCOVERY_INTERVAL = 1800
 NEVER_TIME = 0
 RESUB_COOLDOWN_SECONDS = 10.0
 # S1 speakers apply a command a moment after acknowledging it, so the resulting state is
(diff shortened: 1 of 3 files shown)


marcelveldt:
Both suppressed comments were fair, fixed in the last push:

- `scanning.set()` ran on the worker thread — now goes through `loop.call_soon_threadsafe`. Couldn't get it to actually fail, but it's the wrong way round regardless.
- fixture teardown now only cancels the discovery reschedules it tracks, not every handle asyncio armed during the test.

Ran the file 20x after the change, no flakes.

marcelveldt:
Both flagged the same thing an independent review had just caught — a discovery started outside the provider's own task id (the one `run_provider_discovery` awaits directly after every load) resumes after the unload and re-arms the timer. Real, and reproduced it.

Already fixed in 721442fd, which landed just after that review ran: `_unloaded` is set at the top of `unload()` and checked before the reschedule, same pattern as `SonosPlayer._unloaded`. Test covers it.

PR #5503, opened 2026-08-08
Ignore incoming media update events while starting HEOS playback

Tommatheussen:

We tried different fixes based on available data to check whether we want to update the metadata on a playing HEOS device. The problem is that HEOS can send media update events out of order and potentially a little bit delayed, causing the metadata inside MA to get overriden. This mostly happens when HEOS is playing from an external source before starting playback from MA.

Because the combination with out of order events, and HEOS not providing us with any data from the media update (besides 'URL Stream') is not reliable to check whether we should accept the update, this fix adds a transition period during which those updates are ignored. The transition period is currently set to 5 seconds, in my testing my devices always seem to report back the correct state after +-1.5 seconds, but environments with all wireless devices might have slower responses.
In my opinion, 5 seconds is fine, this just means users cannot start playback from another source on the HEOS device. Well, they can but MA will just ignore those updates until a new update comes in after the 5 second period.

- related issue https://github.com/music-assistant/support/issues/5614

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -62,6 +62,7 @@ class HeosPlayer(Player):
 
         self._device: PyHeosPlayer = device
         self._ma_controls_playback = False
+        self._ma_playback_starting = False
         self._queue_cleanup_lock = asyncio.Lock()
         self._queue_cleanup_pending = False
 
@@ -220,6 +221,14 @@ class HeosPlayer(Player):
             )
             return
 
+        if self._ma_playback_starting:
+            self.logger.debug(
+                "[%s] Ignoring now playing change while MA playback starts: %s",
+                self._device.name,
+                now_playing,
+            )
+            return
+
         # Only update if we're not playing from our queue
         # HEOS does not make a distinction on source ID when playing from a DLNA server, USB stick,
         # generic URL (like MA), or other local source.
@@ -334,10 +343,12 @@ class HeosPlayer(Player):
         )
 
         url = await self.provider.mass.streams.resolve_stream_url(self.player_id, media)
+        self._ma_playback_starting = True
         self._ma_controls_playback = True
         try:
             await self._device.play_url(url)
         except HeosError as err:
+            self._ma_playback_starting = False
             self._ma_controls_playback = False
             self._queue_cleanup_pending = False
             raise PlayerCommandFailed("Failed to start playback.") from err
@@ -346,7 +357,17 @@ class HeosPlayer(Player):
         self._attr_active_source = self.player_id
         self._queue_cleanup_pending = True
 
+        self.mass.call_later(
+            5,
+            self._finish_ma_playback_transition,
+            task_id=f"heos_playback_transition_{self.player_id}",
+        )
+
         self.update_state()
+
+    def _finish_ma_playback_transition(self) -> None:
+        """Stop suppressing delayed events from the source replaced by MA playback."""
+        self._ma_playback_starting = False
 
     def _schedule_queue_cleanup(self) -> None:
         """Debounce queue cleanup so rapid queue changes only trigger one follow-up."""
--- a/music_assistant/providers/heos/player.py
+++ b/music_assistant/providers/heos/player.py
@@ -63,6 +63,7 @@ class HeosPlayer(Player):
         self._device: PyHeosPlayer = device
         self._ma_controls_playback = False
         self._ma_playback_starting = False
+        self._ma_playback_transition_timer_id = f"heos_playback_transition_{self.player_id}"
         self._queue_cleanup_lock = asyncio.Lock()
         self._queue_cleanup_pending = False
 
@@ -343,12 +344,13 @@ class HeosPlayer(Player):
         )
 
         url = await self.provider.mass.streams.resolve_stream_url(self.player_id, media)
+        self._cancel_ma_playback_transition()
         self._ma_playback_starting = True
         self._ma_controls_playback = True
         try:
             await self._device.play_url(url)
         except HeosError as err:
-            self._ma_playback_starting = False
+            self._cancel_ma_playback_transition()
             self._ma_controls_playback = False
             self._queue_cleanup_pending = False
             raise PlayerCommandFailed("Failed to start playback.") from err
@@ -360,14 +362,10 @@ class HeosPlayer(Player):
         self.mass.call_later(
             5,
             self._finish_ma_playback_transition,
-            task_id=f"heos_playback_transition_{self.player_id}",
+            task_id=self._ma_playback_transition_timer_id,
         )
 
         self.update_state()
-
-    def _finish_ma_playback_transition(self) -> None:
-        """Stop suppressing delayed events from the source replaced by MA playback."""
-        self._ma_playback_starting = False
 
     def _schedule_queue_cleanup(self) -> None:
         """Debounce queue cleanup so rapid queue changes only trigger one follow-up."""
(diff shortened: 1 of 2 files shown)


MarvinSchenkel:
Is this something that can potentially be fixed in `pyheos`? Or is this something in the HEOS firmware itself? I would expect those events coming from the firmware to carry at the very least a created_timestamp. Maybe this can be exposed from the `pyheos` package so that we know when to ignore an event?

Tommatheussen:
> Is this something that can potentially be fixed in `pyheos`? Or is this something in the HEOS firmware itself? I would expect those events coming from the firmware to carry at the very least a created_timestamp. Maybe this can be exposed from the `pyheos` package so that we know when to ignore an event?

I will take a deeper look into the actual messages received on the `pyheos` side. From memory these are processed in order, but I do remember some events call other commands to enrich data first before firing that event to MA.

1 thing I'm sure of is HEOS doesn't send timestamps when sending events, they are really barebones and mostly just hold the playerID.

music_assistant/providers/heos/player.py
```diff
@@ -330,10 +339,12 @@ async def play_media(self, media: PlayerMedia) -> None:
         )
 
         url = await self.provider.mass.streams.resolve_stream_url(self.player_id, media)
+        self._ma_playback_starting = True
```
OzGav, line 349: [PROBLEM] A second play_media within the 5s window can lose its protection entirely.
The flag is set to `True` before `await self._device.play_url(url)`, but the timer is only re-armed after the await returns. If playback A armed the timer at `t=0` and playback B calls play_media at `t≈4.8`, the pending timer fires during B's play_url await and `sets _ma_playback_starting = False`. B then re-arms the timer but nothing sets the flag back to `True`, so B runs with a live timer and no suppression at all, exactly the delayed events this PR targets will land. Two easy fixes:

Re-assert `self._ma_playback_starting = True` after the successful `play_url` (next to self._queue_cleanup_pending = True), or
Drop the timer entirely and store a deadline instead — `self._ma_playback_transition_until = self.mass.loop.time() + 5` in `play_media`, and check `self.mass.loop.time() < self._ma_playback_transition_until` in `_update_player_current_media`. No callback lifecycle, no race, less code.

music_assistant/providers/heos/player.py
OzGav, line 480:
```suggestion
        self._ma_controls_playback = False
        self._ma_playback_starting = False
```

`select_source` is an MA-initiated source change, so the transition guard has nothing left to protect and the next now-playing event is exactly what we need to update the active source and current media. Clearing the flag ends the suppression window early. Without it, selecting a source within 5 seconds of `play_media` would have that event swallowed, leaving MA showing the previous track.

music_assistant/providers/heos/player.py
```diff
@@ -342,8 +353,18 @@ async def play_media(self, media: PlayerMedia) -> None:
...
+            5,
+            self._finish_ma_playback_transition,
+            task_id=f"heos_playback_transition_{self.player_id}",
+        )
+
         self.update_state()
 
+    def _finish_ma_playback_transition(self) -> None:
+        """Stop suppressing delayed events from the source replaced by MA playback."""
+        self._ma_playback_starting = False
```
OzGav:
```suggestion
        self._ma_playback_starting = False
        self._update_player_current_media()
        self.update_state()
```
Suppressed events are dropped, not deferred but I understand `pyheos` keeps `now_playing_media` current, so re-running `_update_player_current_media()` when the window closes applies the last known state instead of losing it. If MA is still the active source this hits the existing "Url Stream" guard and is a no-op but if an external source took over during the window, its metadata lands at the 5-second mark rather than staying stale until the next event happens to arrive.

OzGav (commented):
A couple of things to consider

Tommatheussen:
> > Is this something that can potentially be fixed in `pyheos`? Or is this something in the HEOS firmware itself? I would expect those events coming from the firmware to carry at the very least a created_timestamp. Maybe this can be exposed from the `pyheos` package so that we know when to ignore an event?
> 
> I will take a deeper look into the actual messages received on the `pyheos` side. From memory these are processed in order, but I do remember some events call other commands to enrich data first before firing that event to MA.
> 
> 1 thing I'm sure of is HEOS doesn't send timestamps when sending events, they are really barebones and mostly just hold the playerID.

@MarvinSchenkel checking some of the earlier logs that we've received for various bug reports, What I've found is that the general event order that HEOS sends is not 100% consistent, 1 of the observed orders (after sending a Play command from MA) is:
    - player_queue_changed
    - player_now_playing_changed
    - player_queue_changed
    - player_state_changed=stop
    - player_state_changed=play
    - player_now_playing_changed

The other order that can be seen:
    - player_queue_changed
    - player_queue_changed
    - player_state_changed=stop
    - player_now_playing_changed
    - player_now_playing_changed
    - player_state_changed=play

For both `player_queue_changed` and `player_now_playing_changed` the only metadata we receive is the player ID. So we need to send additional commands to the device to grab the relevant information. The `player_queue_changed` is not really relevant, as we use that one to schedule the queue cleanup task only.
`player_now_playing_changed` however is relevant, `pyheos` enriches that event to make sure the player object has the latest information before sending the event to MA. Due to this enrichment, the order that MA receives the events usually is different:
    - player_queue_changed
    - player_queue_changed
    - player_state_changed=stop
    - player_now_playing_changed
    - player_state_changed=play
    - player_now_playing_changed
 
Our previous fix, to ignore the `now_playing_changed` events when the player is stopped, handled a lot of the cases, due to the event reordering from `pyheos`. The problem remains whenever stale now_playing is processed while the latest state is (still or already) PLAY. But because HEOS seems to not send the event in the same order everytime, we can't just rely on the order itself. `pyheos` just reads them from the TCP connection, logs the event and only then starts any additional processing on them.


@OzGav valid points, I will take a look at testing and  incorporating them

music_assistant/providers/heos/player.py
marcelveldt: This race is still present. I think we should cancel the existing transition timer before the `play_url()` await (or use a monotonic deadline) and cover overlapping starts with a regression test.

music_assistant/providers/heos/player.py
marcelveldt: This still applies. `select_source()` should clear and cancel the transition guard so its own now-playing update is not swallowed.

music_assistant/providers/heos/player.py
marcelveldt: This still needs the reconcile on expiry. The test calls `_update_player_current_media()` manually, but production only clears the flag, so a suppressed external update can remain stale.

marcelveldt:
@Tommatheussen I think these are the remaining bits to address before merging. I am happy to pick them up and push a fix if you want.

Tommatheussen:
> @Tommatheussen I think these are the remaining bits to address before merging. I am happy to pick them up and push a fix if you want.

You can pick it up if you want, I'm currently on vacation (with my laptop) so a little less time to look at it ;)

music_assistant/providers/heos/player.py
marcelveldt: Fixed by cancelling the previous transition timer before awaiting `play_url()`, with a regression test covering a second start during the window.

music_assistant/providers/heos/player.py
marcelveldt: Fixed by ending and cancelling the transition before `select_source()` calls HEOS, so its resulting media update is accepted immediately.

music_assistant/providers/heos/player.py
marcelveldt: Fixed by reconciling the latest HEOS media and publishing state when the transition expires. The test now verifies this without a manual refresh.

music_assistant/providers/heos/player.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, provider: HeosPlayerProvider, device: PyHeosPlayer) -> None:
 
         self._device: PyHeosPlayer = device
         self._ma_controls_playback = False
+        self._ma_playback_starting = False
+        self._ma_playback_transition_timer_id = f"heos_playback_transition_{self.player_id}"
```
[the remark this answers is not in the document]
marcelveldt: Good catch. The transition cleanup is now registered with the player's unload callbacks, with a regression test verifying the timer and guard are cleared.

PR #5510, opened 2026-08-08
Fix file shares importing nothing and hiding playlists (#6019)

OzGav:

Unfortunately the setup flow change in beta 10 dropped the `content_type` config entry and used a LABEL in its place, which has resulted in the loss of the default value for the filesystem provider content type. [#5506](https://github.com/music-assistant/server/issues/5506) restored the fallback for the local provider's own read, but the mirror on the options page still holds no value, so the sync options don't resolve their depends_on against it, and WebDAV/OneDrive/Google Drive read the content type through their own paths which were left without a fallback. This PR restores the read-only entry, so on next restart the default value of "music" will be used again and affected shares recover on their own. Shares explicitly set to audiobooks/podcasts were never affected.

- related issue https://github.com/music-assistant/support/issues/6019

## Types of changes

## Checklist


chrisuthe (approved):
LGTM!

PR #5517, opened 2026-08-08
Fix MusicCast zone2/main-sync players never refreshing outside UDP push

bsny:

Related to music-assistant/support#6014

Zone2/main-sync players (e.g. a zone joined via Main Zone Sync) basically never update outside of
a UDP push for that specific zone, or an availability flap. On a network where the receiver can't
reach the app's UDP listener, that means never.

Two separate bugs here, both found digging into the same issue:

`poll()` was only ever pushing the fetched data to the main-zone player, even though
`physical_device.fetch()` pulls fresh data for every zone on the device in one call. Sibling zones
had no periodic refresh path at all. Also dropped the `update_lock.locked()` early-return at the
top — it could silently skip the whole poll (including the new refresh) if a UDP callback happened
to be running at the exact moment the scheduled tick fired.

Separately, a main-sync zone's `state` could never report PAUSED/IDLE. `is_netusb` only matches
whichever zone is actually running the netusb session, and a main-sync zone's `source_id` is
always `main_sync`, so it fell through to PLAYING forever no matter what actually happened.
`netusb_playback` is device-wide though, so widened the check to cover main-sync zones too.

Tested both live against two TSR-700s (one on a native MA queue, one on Main Zone Sync) - zone2
entities refresh on the normal poll cycle now, and the main-synced one correctly follows the
leader through play/pause/idle/play.

## Types of changes

## Checklist


music_assistant/providers/musiccast/musiccast.py
```diff
@@ -188,12 +188,18 @@ def other_zones(self) -> list[MusicCastZoneDevice]:
...
+        """
+        Return the state of the player.
+
+        A main-sync zone mirrors the driving zone's netusb playback status, since only
+        one zone can hold the shared netusb session at a time.
+        """
         assert self.zone_data is not None
         if self.zone_data.power == "on":
-            if self.is_netusb and self.device.data.netusb_playback == "pause":
+            _tracks_netusb = self.is_netusb or self.source_id == MC_SOURCE_MAIN_SYNC
```
OzGav:
```suggestion
            _main = self.physical_device.zone_devices.get(MC_DEFAULT_ZONE)
            _tracks_netusb = self.is_netusb or (
                self.source_id == MC_SOURCE_MAIN_SYNC and _main is not None and _main.is_netusb
)
```

When a zone is set to main_sync, it mirrors whatever the main zone is playing which may not be network audio at all (tuner, HDMI/TV, analog inputs).

The receiver's netusb playback status is device-wide and sticky so once a stream stops, it reports "stop" (or "pause") until the next netusb session starts, regardless of what any zone is currently playing.

This PR makes a main_sync zone derive its state from that device-wide netusb status. That's correct while main is actually running a netusb session (e.g. playing an MA queue via the server input) but if main is on the tuner or TV, the synced zone picks up the stale "stop" and MA reports it as idle while it's audibly playing. Main itself, meanwhile, still reports playing (powered-on + non-netusb source), so the leader and its mirror disagree about the same audio stream.

So only follow the netusb status when the main zone actually holds the netusb session, otherwise fall back to the existing powered-on-means-playing behaviour

fmunkes: I think this comment is valid. Other than that, the PR looks good to me.

bsny: valid. applied.

OzGav:
Please make sure you run pre-commit before pushing as lint is failing

music_assistant/providers/musiccast/musiccast.py
```diff
@@ -188,12 +188,21 @@ def other_zones(self) -> list[MusicCastZoneDevice]:
...
+        A main-sync zone mirrors the driving zone's netusb playback status, since only
+        one zone can hold the shared netusb session at a time.
+        """
         assert self.zone_data is not None
         if self.zone_data.power == "on":
-            if self.is_netusb and self.device.data.netusb_playback == "pause":
+            _main = self.physical_device.zone_devices.get(MC_DEFAULT_ZONE)
+            _tracks_netusb = self.is_netusb or (
+                self.source_id == MC_SOURCE_MAIN_SYNC and _main is not None and _main.is_netusb
+)
```
OzGav:
```suggestion
                self.source_id == MC_SOURCE_MAIN_SYNC and _main is not None and _main.is_netusb
            )
```

OzGav (approved):
Lint is fixed. Thanks @bsny !

PR #5519, opened 2026-08-08
Name the right speaker when Sonos S1 grouping fails

marcelveldt:

When grouping Sonos S1 speakers fails, the log named the wrong speaker. Grouping is driven from the group leader, but it joins and unjoins the *other* speakers, so a speaker refusing to join was reported against the leader:

`Error calling SonosPlayer.set_members on Kitchen: <error>` — while Study was the one that actually refused.

Only the server log is affected; the error shown in the UI is unchanged.

- Join and unjoin now run as their own methods on the speaker being (un)grouped, so the error names that speaker.
- Added tests covering both the join and unjoin failure paths.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos_s1/player.py
+++ b/music_assistant/providers/sonos_s1/player.py
@@ -283,7 +283,6 @@ class SonosPlayer(Player):
         else:
             await self.stop()
 
-    @soco_error()
     async def set_members(
         self,
         player_ids_to_add: list[str] | None = None,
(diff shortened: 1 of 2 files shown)


marcelveldt:
Checked this one — the raw message doesn't reach the UI.

`PlayerCommandFailed` sets `translation_key = "player_command_failed"`, and `ErrorResultMessage.__post_serialize__` overwrites `details` with the localized string whenever a resolver is bound — which is on every outbound message (`websocket_client.py:337`). The key resolves via `common.errors.player_command_failed` in `strings.json`, so clients see "The command to the player failed." both before and after this PR. `str(err)` is only the fallback for keys that don't resolve.

So this stays a log-only change.

PR #5522, opened 2026-08-09
Clean up library items left behind without a provider

marcelveldt:

When a music provider is removed, its items are stripped from the library one by one. The provider mapping rows were deleted *before* the library item itself, so if removing the item then failed, the item stayed behind with no provider attached to it. Such an item can never be played and is not picked up by the retry, because that retry only looks at the (already deleted) mapping rows.

On top of that, the daily cleanup that removes these leftovers never covered podcasts and audiobooks, so those stayed in the library forever. Opening one resulted in a confusing internal error instead of a clear "not found" message.

- n/a

## Types of changes

Changes:

- Remove the library item first and let that clear its provider mappings, so a failed removal leaves the item intact and can simply be retried.
- Include podcasts and audiobooks in the periodic cleanup of leftover library items.
- Report a clear "not found" error for a library item that has no provider left, instead of an internal error.
- As a side effect, play history and audio analysis records belonging to the removed provider are now properly cleaned up too.

## Checklist

### DIFF
--- a/music_assistant/controllers/music/database.py
+++ b/music_assistant/controllers/music/database.py
@@ -59,7 +59,9 @@ if TYPE_CHECKING:
     from music_assistant import MusicAssistant
     from music_assistant.controllers.music.media.albums import AlbumsController
     from music_assistant.controllers.music.media.artists import ArtistsController
+    from music_assistant.controllers.music.media.audiobooks import AudiobooksController
     from music_assistant.controllers.music.media.playlists import PlaylistController
+    from music_assistant.controllers.music.media.podcasts import PodcastsController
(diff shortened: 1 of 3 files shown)


marcelveldt:
Re the suppressed comment about the already-deleted/race path: that behaviour is unchanged here - `dev` deletes the same mapping rows and returns without touching audio_analysis/playlog either, the refactor just moved it into the except block. And to reach that path the item row is already gone, which normally means `remove_item_from_library` already cleaned those rows. The one case where leftovers can survive is the orphan prune deleting item rows directly, which is a pre-existing gap in the prune itself - fixing it here wouldn't help. Follow-up.

PR #5524, opened 2026-08-09
Keep the full duration of an audiobook when resuming mid-book

marcelveldt:

Resuming a multi-file audiobook part-way through would shrink the book's stored duration to just the part that was left to play. Progress reporting then used that shortened length, so a book could be marked as finished while there was still plenty left.

When the audio source takes care of seeking itself, the seek position is passed on to that source and cleared. At the end of the stream the duration was recalculated as if playback had started at the very beginning, so only the remaining audio was counted.

This also affects normal tracks from providers that seek themselves (Spotify, for example): seeking past the first minute of a track could get it marked as played, and scrobbled, well before it actually ended.

Changes:

- Remember the position that was originally asked for, and only store a newly measured duration when the item really did play from the start.
- Added tests for both cases: a multi-file book, and a provider stream that handles its own seeking.

Present in `stable` as well, so labelled for backport.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/streams/audio.py
+++ b/music_assistant/controllers/streams/audio.py
@@ -447,6 +447,9 @@ class StreamsAudio:
         # copy: the args below are appended per call, while the StreamDetails is cached on
         # the queue item and reused across calls (retry, seek, background analysis)
         extra_input_args = list(streamdetails.extra_input_args or [])
+        # the branches below zero out seek_position where the seek is delegated to the
+        # source itself, so keep the requested position for the duration writeback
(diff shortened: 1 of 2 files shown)


marcelveldt:
Both fair, fixed:

- The docstring promised more than the test checked. Rather than water down the wording I made it true - both delegated-seek tests now also assert `-ss` never reaches ffmpeg, which is the actual proof the source consumed the seek itself.
- The 21MB allocation is gone, the fake now yields per-second chunks like the real `iter_chunked` does.

PR #5532, opened 2026-08-09
Make an author or narrator playable

fmunkes:

While we do support authors and narrators as library items now, I forgot to make an author/ narrator a playable media item. I.e. currently, any try to play an author/ narrator in the frontend results in a "no playable media items found".

This PR adds this ability. If an author/ narrator has unplayed items, these are enqueued. If some of these items are part of a collections, the collection's order is respected. If the item is part of multiple collections, they are only enqueued once. If all items of an author/ narrator are played, all of them are enqueued.

## Types of changes

## Checklist


music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -495,6 +495,62 @@ async def get_next_audiobook(
...
+            author_narrator.item_id, author_narrator.artist_type
+        )
+        for book in all_audiobooks:
+            # do not use get_resume_position here, as an artist may potentially have a lot of audiobooks,
+            # resulting in many API calls.
+            if book.fully_played:
+                continue
+            audiobooks.append(book)
+        if len(audiobooks) == 0:
+            audiobooks = all_audiobooks
```
OzGav: This all books fallback queues all the played audiobooks but does not reset the played position so they will all resume at the end and just skip through?

music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -495,6 +495,62 @@ async def get_next_audiobook(
...
+                    collection = await self.mass.music.audiobooks.get_collection(
+                        get_collection_item_id(media_item_collection.title, MediaType.AUDIOBOOK)
+                    )
+                    collections.add(collection)
+                    books_with_collection.add(book)
+                except MediaNotFoundError:
+                    continue
+
+        # remove books which are part of a collection
+        audiobooks = list(set(audiobooks).difference(books_with_collection))
```
OzGav: Both this line and the collections/books_with_collection sets on 528–529 lose the ordering. `MediaItem.__hash__` is `hash(self.uri)`, so set iteration order follows randomized string hashing and changes between restarts which means the standalone books come back in arbitrary order instead of the DB's sort order, and with more than one collection the collections themselves get shuffled, which works against the "collection's order is respected" goal. `UniqueList` is already imported here and dedupes while preserving order. Use `UniqueList()` for the two accumulators and `[x for x in audiobooks if x not in books_with_collection]` here.

music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -495,6 +495,62 @@ async def get_next_audiobook(
...
+            _author_narrator = await self.mass.music.artists.get_library_item_by_prov_id(
+                author_narrator.item_id, author_narrator.provider
+            )
+            if _author_narrator is None:
+                raise MediaNotFoundError(
+                    f"Author/ narrator {author_narrator.name} is not part of the library."
+                )
+            author_narrator = _author_narrator
+        audiobooks: list[Audiobook] = []
+        all_audiobooks = await self.mass.music.artists.get_library_author_narrator_audiobooks(
```
OzGav: `artists.audiobooks(...)` already does all of this, plus the SINGER guard and a provider fallback for artists not in the library. Calling `get_library_author_narrator_audiobooks` directly skips both. 505–513 also duplicates `_resolve_library_artist` from further down this file.

music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -495,6 +495,62 @@ async def get_next_audiobook(
...
+            author_narrator.item_id, author_narrator.artist_type
+        )
+        for book in all_audiobooks:
+            # do not use get_resume_position here, as an artist may potentially have a lot of audiobooks,
+            # resulting in many API calls.
+            if book.fully_played:
```
OzGav, line 517: `fully_played` comes from the playlog scoped to the current session user, not the `userid` this resolver was passed. With no session user (autoplay, HA) it can pick up another user's progress. Every other audiobook path here passes `userid` through but this method doesn't take one.

music_assistant/controllers/player_queues/media_resolver.py
```diff
@@ -495,6 +495,62 @@ async def get_next_audiobook(
...
+        if len(audiobooks) == 0:
+            audiobooks = all_audiobooks
+
+        # treat books part of a collection separately by keeping the collections order
+        collections: set[MediaCollection[Audiobook]] = set()
+        books_with_collection: set[Audiobook] = set()
+        for book in audiobooks:
+            for media_item_collection in book.metadata.collections or []:
+                try:
+                    collection = await self.mass.music.audiobooks.get_collection(
```
OzGav: `get_collection` is an uncached DB query, so this refetches the same collection once per book. You could collect the distinct collection IDs first, then fetch each once.

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: No. If you enqueue a fully played item, this will start from the beginning. I verified it.

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: Good catch. Fixed.

music_assistant/controllers/player_queues/media_resolver.py
OzGav: OK then thanks! I could have sworn it would have because I thought the playlog would hold the fully played duration and because I saw guards for this elsewhere but if it works it works!

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: Well, and I even added that method to the artist controller. Fixed.

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: This is true - I'm gonna add the impersonated user context handler here, for both audiobooks and get_collection. I really don't want to call get_resume_position in the author part on every item.

music_assistant/controllers/player_queues/media_resolver.py
fmunkes: good idea.

fmunkes: Alright, this got a bit more complicated, as verification if a collections exists is later now, but I got it sorted out.

marcelveldt (approved):
Thanks, @fmunkes !

PR #5533, opened 2026-08-09
Allow collapsing of collections when acquiring the audiobooks of an author or narrator

fmunkes:

Small enhancement to the collection support.
We already have everything needed to collapse the in-library collections of an author/ narrator, we just need to add the parameter to the respective method.

The frontend companion PR is equally small: https://github.com/music-assistant/frontend/pull/2351

As the collapse_collections parameter is optional and False, we can merged this PR independently of the frontend one. The frontend one needs to wait.

## Types of changes

## Checklist


music_assistant/controllers/music/media/artists.py
```diff
@@ -325,11 +326,13 @@ async def audiobooks(
         provider_instance_id_or_domain: str,
         artist_type: ArtistType = ArtistType.AUTHOR,
         in_library_only: bool = False,
-    ) -> list[Audiobook]:
+        collapse_collections: bool = False,
+    ) -> list[Audiobook] | list[Audiobook | MediaCollection[Audiobook]]:
```
chrisuthe, line 355: The other `collapse_collections` signatures in the codebase (`base.py:1270-1327`, `audiobooks.py:150-199`) make the parameter keyword-only and use `@overload` on `Literal[True]`/`Literal[False]` so callers get a precise return type. As written here, anyone leaving the default has to narrow away a `MediaCollection` that can never appear.

```suggestion
        *,
        collapse_collections: bool = False,
    ) -> list[Audiobook] | list[Audiobook | MediaCollection[Audiobook]]:
```

And to match the pattern, above `async def audiobooks(` (needs `Literal, overload` added to the `typing` import):

```python
if TYPE_CHECKING:

    @overload
    async def audiobooks(
        self,
        item_id: str,
        artist_type: ArtistType = ArtistType.AUTHOR,
        in_library_only: bool = False,
        *,
        collapse_collections: Literal[False] = False,
    ) -> list[Audiobook]: ...

    @overload
    async def audiobooks(
        self,
        item_id: str,
        artist_type: ArtistType = ArtistType.AUTHOR,
        in_library_only: bool = False,
        *,
        collapse_collections: Literal[True],
    ) -> list[Audiobook | MediaCollection[Audiobook]]: ...
```

Keyword-only is safe for the API command — the dispatcher calls `handler.target(**args)`.

music_assistant/controllers/music/media/artists.py
```diff
@@ -325,11 +326,13 @@ async def audiobooks(
...
         """
         Return audiobooks for an artist.
 
         Artist_type can be omitted for in-library artists.
+        Collapse collections only works for in library items.
```
chrisuthe: This method is registered as the `music/artists/artist_audiobooks` API command, and the docs generator only picks up `:param name:` entries (`_parse_docstring` in `controllers/webserver/api_docs.py`) — so as written this note won't reach the generated API docs. Also worth being explicit that provider items come back flat: with `in_library_only=False`, a not-in-library book from an already-collapsed series appears loose next to its own collection.

```suggestion

        :param collapse_collections: Collapse available collections. Only applies to
            in-library items; when in_library_only is False, provider items are
            appended as plain audiobooks alongside the collapsed collections.
```

music_assistant/controllers/music/media/artists.py
```diff
@@ -355,15 +358,23 @@ async def audiobooks(
...
         # return all (unique) items from all providers
         # initialize unique_ids with db_items to prevent duplicates
-        unique_ids: set[str] = {f"{item.name}.{item.version}" for item in db_items}
+        unique_ids: set[str] = set()
+        for item in db_items:
+            if isinstance(item, Audiobook):
+                unique_ids.add(f"{item.name}.{item.version}")
+            else:
+                for collection_item in item.items:
+                    unique_ids.add(f"{collection_item.name}.{collection_item.version}")
```
chrisuthe, line 403: Cleaner to branch on the type you're actually special-casing — that's also how the existing test discriminates (`test_library_listing_queries.py:561`), and it stays correct if this list ever carries summary items.

```suggestion
        unique_ids: set[str] = set()
        for item in db_items:
            if isinstance(item, MediaCollection):
                for collection_item in item.items:
                    unique_ids.add(f"{collection_item.name}.{collection_item.version}")
            else:
                unique_ids.add(f"{item.name}.{item.version}")
```

music_assistant/controllers/music/media/artists.py
```diff
@@ -391,10 +402,8 @@ async def audiobooks(
         return result
 
     async def get_library_author_narrator_audiobooks(
-        self,
-        item_id: str | int,
-        artist_type: ArtistType,
-    ) -> list[Audiobook]:
+        self, item_id: str | int, artist_type: ArtistType, collapse_collections: bool = False
+    ) -> list[Audiobook] | list[Audiobook | MediaCollection[Audiobook]]:
```
chrisuthe: Same keyword-only treatment here, and this keeps the multi-line signature the method already had rather than reflowing it.

```suggestion
        self,
        item_id: str | int,
        artist_type: ArtistType,
        *,
        collapse_collections: bool = False,
    ) -> list[Audiobook] | list[Audiobook | MediaCollection[Audiobook]]:
```

The only caller already passes it by keyword, so nothing else changes.

chrisuthe (commented):
Logic checks out — I traced `_adapt_query_for_collections` and confirmed collections are built from the already-filtered query, so a collapsed collection only ever contains that artist's audiobooks, and the reworked dedup correctly reaches into collection members.

One thing not in the diff: there's no test for the new path. `test_audiobook_collections_collapse_and_preserve_order` covers collapsing on the audiobooks controller, but nothing exercises it via the artist route — and the dedup-across-collection-members loop is new logic that would regress silently. Worth a small test that adds two collection books by the same author plus a standalone one and asserts the shape coming back from `artists.audiobooks(..., collapse_collections=True, in_library_only=True)`.

Rest of the notes are inline.

music_assistant/controllers/music/media/artists.py
fmunkes: Thank you!

music_assistant/controllers/music/media/artists.py
fmunkes: Fair, thanks for providing the signatures as well.

fmunkes:
> One thing not in the diff: there's no test for the new path. `test_audiobook_collections_collapse_and_preserve_order` covers collapsing on the audiobooks controller, but nothing exercises it via the artist route — and the dedup-across-collection-members loop is new logic that would regress silently. Worth a small test that adds two collection books by the same author plus a standalone one and asserts the shape coming back from `artists.audiobooks(..., collapse_collections=True, in_library_only=True)`.

I added a test. Sorry for the commit mess, I pushed/ pulled to the wrong branch first, but that branch was clean. Thanks for the review!

chrisuthe (approved):
Looking good to me!

PR #5549, opened 2026-08-09
Play announcements on muted speakers that announce natively

marcelveldt:

#5474 made announcements override a deliberate mute, but only for speakers that Music Assistant renders the announcement for itself. Speakers that play announcements natively (Sonos, Snapcast, Samsung WAM, Bose SoundTouch, Home Assistant players, ESPHome via Sendspin) were left out, so whether a muted speaker actually said anything was up to the speaker.

In practice that meant a muted Snapcast client stayed completely silent, a Samsung speaker played the announcement but was left unmuted afterwards, and any speaker using volume-based (fake) mute announced at the muted zero instead of at its real volume.

An announcement is meant to be heard, so it now always overrides the mute — on every speaker, and the mute is restored once the announcement is done.

## Types of changes

## Checklist

## Changes

- Speakers with native announcement support are unmuted for the announcement and muted again afterwards, the same way the other speakers already were.
- All members of a muted group hear the announcement, and each keeps its own mute setting.
- The announcement volume is now worked out after the unmute, so speakers using volume-based mute announce at their real volume.
- Bose SoundTouch now waits for the announcement to finish before handing the speaker back, like every other speaker does.
- Added test coverage for announcements on muted speakers that announce natively.

### DIFF
--- a/music_assistant/controllers/players/announcements.py
+++ b/music_assistant/controllers/players/announcements.py
@@ -258,8 +258,9 @@ class AnnouncementsMixin:
                         player.state.name,
                         url,
                     )
-                announcement_volume = self.get_announcement_volume(player_id, volume_level)
-                await announce_player.play_announcement(announcement, announcement_volume)
+                await self._play_native_announcement(
+                    player, announce_player, announcement, volume_level
+                )
                 return
             # use fallback/default implementation
             await self._play_announcement(player, announcement, volume_level)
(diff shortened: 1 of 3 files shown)


marcelveldt:
Copilot flagged the `duration or 10` fallback on the Bose sleep as critical - it's the exact line Sonos has been running for ages. It only kicks in when the render didn't finish within its own 30s window, on top of the `wait_ready` before it, which for a TTS clip means the audio is broken anyway. And before this change Bose returned instantly, so the announcement got torn down after ~0s - this is strictly better. Leaving it as-is to match Sonos; if a real truncation report ever shows up we should fix both at once.

PR #5551, opened 2026-08-10
Make failing background tasks fail the announcement tests

marcelveldt:

The two announcement test classes each carried their own copy of a `mass.create_task` stub. That stub never looked at the result of the tasks it started, so a background task that blew up only produced an asyncio warning instead of failing the test.

Both now share one fixture, built on the `use_real_create_task` helper the rest of the test suite already uses. Letting it re-raise immediately surfaced a stray artwork lookup that had been failing unnoticed in two of those tests.

- Replaced the duplicated stub with a single shared fixture that runs background tasks for real and raises whatever they raise.
- Stopped the two announcement test classes from handing the player a stand-in queue, which was triggering an unrelated (and failing) artwork colour lookup.

## Types of changes

## Checklist

### DIFF
--- a/tests/controllers/players/test_player_controller.py
+++ b/tests/controllers/players/test_player_controller.py
@@ -13,9 +13,9 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import time
-from collections.abc import AsyncIterator, Callable, Iterator
+from collections.abc import AsyncIterator, Callable, Coroutine, Iterator
 from types import SimpleNamespace
-from typing import cast
+from typing import Any, cast
 from unittest.mock import ANY, AsyncMock, MagicMock, call, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


marcelveldt:
Re the suppressed comment about done-callback timing: the callback version did actually catch it (the collect callback is queued while the loop is still running the test, so it lands before run_until_complete stops it), but the point stands that you shouldnt have to reason about tick ordering to read this. Switched to tracking the tasks and inspecting them at teardown, which is what you suggested - with a `task.done()` guard, since `exception()` throws InvalidStateError on a task that never finished.

Verified both ways: a fire-and-forget task that raises as the very last statement of a test now fails that test, and two failures come out as an ExceptionGroup.

marcelveldt:
Re the pending-task comment: checked it rather than assumed. Instrumented the fixture and ran all 21 tests in both classes - every one ends with 0 pending tasks, because TaskManager.__aexit__ awaits everything it dispatched before the code under test returns. So theres nothing for the extra check to catch here, and failing on a still-running task would be a policy that could bite a future test that legitimately fires and forgets. Leaving as is.

PR #5553, opened 2026-08-10
Clear the AirPlay session a crashed stream leaves on the device

OzGav:

When a cliairplay process dies without delivering a teardown (typically after a network blackout outlives the keepalive tolerance), the receiver still holds the armed audio session. Apple receivers audibly pop on that session's starving queue and keep doing so until a new session displaces it. A user report shows exactly this: a HomePod popping for hours overnight until it was power cycled, right after such a crash where the automatic re-join also failed and MA logged it could not stop the player.

After an unexpected stream loss the player now schedules bounded attempts to connect a short-lived stream to the device and cleanly disconnect, which forces the receiver to drop the leaked session.

This hasn't been tested locally, but the receiver behavior is documented in cliairplay.c: "a receiver whose queue underruns while the session stays armed pops audibly (the end-of-playback burst on Apple receivers)". The remedy is therefore established, what was missing is anyone to deliver it after a crash. The fix uses only the existing connect and disconnect paths that every normal playback start and stop already exercises, so no new device-facing protocol code is introduced; the new code is scheduling logic only, and that is what the unit tests cover. Risk is bounded, the attempts stand down whenever a real session owns the player, skip offline devices, are cancelled by any user action, and give up after three tries, so the worst case is identical to current behavior. Success is logged at INFO ("Cleared a leftover AirPlay session on...") so field confirmation from the reporter's next incident is directly visible in their log.

- related issue Reported here https://github.com/orgs/music-assistant/discussions/5982#discussioncomment-17945050 and maybe also here https://github.com/music-assistant/support/issues/5279#issuecomment-5234779452

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/airplay/constants.py
+++ b/music_assistant/providers/airplay/constants.py
@@ -206,6 +206,13 @@ AIRPLAY_LATE_JOIN_RING_MAX_BYTES: Final[int] = 6 * 1024 * 1024
 # Staged retries can be reintroduced by adding entries to the tuple.
 AIRPLAY_REJOIN_ATTEMPT_DELAYS: Final[tuple[int, ...]] = (5,)
 
+# Delays (seconds) between attempts to clear a session left behind on a device
+# whose cliairplay process died without a clean teardown. Apple receivers keep
+# the session armed and audibly pop until a new session displaces it, so a
+# short-lived connect/disconnect forces the drop. The increasing delays cover
+# both a brief network blip and a longer outage.
+AIRPLAY_SESSION_CLEANUP_ATTEMPT_DELAYS: Final[tuple[int, ...]] = (10, 60, 300)
+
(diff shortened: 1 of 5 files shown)


marcelveldt:
Thanks for digging into this - the diagnosis is right and the log trail you found is what pinned it.

Fixing it in the binary instead: music-assistant/airplay-cli#59. cliairplay never tells the receiver anything when it gives up on the control channel - it marks the channel dead and shuts the socket, and the send path is fail-closed after that, so the TEARDOWN at disconnect is a no-op. The device just gets a bare FIN and keeps the session armed. Writing the goodbye at the point we give up lands it within a second, while the receiver still has audio queued - the case we already measured as clean.

Doing it server-side needs a second session to displace the armed one, and your log shows that failing: the re-join 5s later got "cliairplay did not connect". And 5 minutes on we can't tell whether that HomePod is busy with someone else.

Closing this one, your analysis got us there.

PR #5559, opened 2026-08-10
Universal Players no longer come back as a brand new player

OzGav:

A Universal Player — the player Music Assistant creates for a device that has no native provider — took its player id from whichever device identifiers happened to be available at the moment it was created. Re-creating one could therefore give the same speaker a different id, for example switching from a MAC-based to a UUID-based one. Anything bound to that id then sees the player disappear and a brand new one show up: in Home Assistant the old entity is orphaned and automations referencing it stop working. A related problem made the replacement carry a stale automatically generated name, which showed up as a duplicated friendly name.

The id is now handed out once and never derived from the device again.

## Changes

- A Universal Player gets a random id the first time its device is wrapped, and that id is never recomputed
- A device's Universal Player is found through the link its protocol players store, and is restored from its configuration when it is not loaded yet — previously this case quietly created a second player
- Nothing looks a Universal Player up by rebuilding its id from the device any more
- A Universal Player keeps its settings when its protocol players disappear: it goes unavailable instead of being deleted, so the device returns as the same player. Only removing it yourself deletes it for good
- When two Universal Players for the same device merge, the one that has been around longest now wins, so the established player id is the one that survives
- A player no longer stores its automatically generated name as if you had renamed it, so a renamed device is no longer shadowed by its old name. Existing players are migrated

Existing Universal Players keep the id they already have — nothing is renumbered.

**Related issue:** https://github.com/music-assistant/support/issues/5888

<sub>Note for reviewers: the trigger in the report (changing the output protocol) could not be reproduced as a cause on either `dev` or 2.9.9 — that setting never tears a player down. The destructive path that did exist in 2.9.9 was on the startup restore and is already fixed on `dev` by #4643. What remains genuinely unstable is the derived id itself, which this PR removes. One case stays open by design: if a protocol player is permanently removed and its config deleted, a returning device has no stored link left and gets a new id.</sub>

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/players/protocol_linking.py
+++ b/music_assistant/controllers/players/protocol_linking.py
@@ -1093,7 +1093,7 @@ class ProtocolLinkingMixin:
 
         # only carry an actual user rename, not the auto-generated default name;
         # likewise a name on the native player only counts as a user override when
-        # it differs from the default name (fresh configs store name == default_name)
+        # it differs from the default name
         custom_name = source_raw.get("name")
         target_name = target_raw.get("name")
         target_has_custom_name = bool(target_name) and target_name != target_raw.get("default_name")
--- a/music_assistant/controllers/players/protocol_linking.py
+++ b/music_assistant/controllers/players/protocol_linking.py
@@ -733,8 +733,8 @@ class ProtocolLinkingMixin:
         from a different protocol (e.g., DLNA-based universal player now matches
         AirPlay-based universal player because they share the same MAC address).
 
-        The universal player with more protocol links absorbs the other one. Only one merge
-        is performed per call; re-evaluation will catch cascading merges.
+        The oldest universal player absorbs the other one. Only one merge is performed
+        per call; re-evaluation will catch cascading merges.
         """
         if not (match := self._find_mergeable_universal_player(universal_player)):
             return
@@ -801,7 +801,7 @@ class ProtocolLinkingMixin:
         """
         Return which of the two universal players absorbs the other, as (keeper, absorbed).
 
-        The player with the most protocol links wins, on a tie the oldest one.
+        The oldest player wins, on a tie the one with the most protocol links.
 
         :param universal_player: The universal player the merge was triggered for.
         :param other_player: The universal player it matched with.
@@ -815,13 +815,21 @@ class ProtocolLinkingMixin:
         return other_player, universal_player
 
     def _merge_rank(self, universal_player: Player) -> tuple[int, int, str]:
-        """Return the sort key of a universal player in a merge, lowest wins."""
+        """
+        Return the sort key of a universal player in a merge, lowest wins.
+
+        Age comes first because the keeper's player id is the one that survives, and the
+        oldest player is the one API consumers have been bound to the longest. The links
+        of the absorbed player move over either way, so nothing is lost by keeping the
+        smaller one. A player from before player ids were minted has no stored moment and
+        counts as the oldest, which is exactly what it is.
+        """
         created_at = self.mass.config.get(
             f"{CONF_PLAYERS}/{universal_player.player_id}/values/{CONF_CREATED_AT}", 0
         )
         return (
-            -len(universal_player.linked_output_protocols),
             created_at if isinstance(created_at, int) else 0,
+            -len(universal_player.linked_output_protocols),
             universal_player.player_id,
         )
 
--- a/music_assistant/controllers/players/protocol_linking.py
+++ b/music_assistant/controllers/players/protocol_linking.py
@@ -45,6 +45,7 @@ from music_assistant.helpers.util import (
     is_locally_administered_mac,
     is_valid_mac_address,
     normalize_mac_for_matching,
+    strong_identifiers_match,
 )
 from music_assistant.models.player import LinkedOutputProtocol, Player
 from music_assistant.providers.sync_group.constants import CONF_ALLOWED_MEMBERS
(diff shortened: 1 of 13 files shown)


music_assistant/providers/universal_player/provider.py
```diff
@@ -572,6 +578,71 @@ async def _create_separate_universal_player(
...
+                    normalized,
+                ) in identifiers:
+                    identifier_match = player_id
+                    break
+        if identifier_match is None:
+            return None
+        return identifier_match.removeprefix(UNIVERSAL_PLAYER_PREFIX)
+
+    @staticmethod
+    def _normalize_identifier(id_type: IdentifierType, value: str) -> str | None:
```
MarvinSchenkel: `_normalize_identifier` is a second device-matching rule set alongside `_identifiers_match` (`music_assistant/controllers/players/protocol_linking.py:1603`), and the two disagree.

`_identifiers_match` restricts matching to `MAC_ADDRESS, SERIAL_NUMBER, UUID, CAST_UUID, AIRPLAY_ID` and drops invalid MACs via `is_valid_mac_address`. This helper accepts every type except `IP_ADDRESS`, which includes `IdentifierType.UNKNOWN` — and `IdentifierType._missing_` maps *any* unrecognized string to `UNKNOWN`, so a stored garbage key lands there instead of being skipped:

```
IdentifierType('some_unknown_key')  ->  <IdentifierType.UNKNOWN: 'unknown'>
```

That also makes the `except ValueError: continue` guard in `_get_stored_device_key` dead code. And with no MAC validity check, a device reporting `00:00:00:00:00:00` normalizes to a matchable `"000000000000"`.

Deciding "are these the same device" is what `_identifiers_match` exists for, and two copies of that rule will drift. Either build a `DeviceInfo` from the stored identifiers and call the existing matcher, or at minimum mirror its type allowlist and MAC filter:

```python
_MATCHABLE_TYPES = (
    IdentifierType.MAC_ADDRESS,
    IdentifierType.SERIAL_NUMBER,
    IdentifierType.UUID,
    IdentifierType.CAST_UUID,
    IdentifierType.AIRPLAY_ID,
)

@staticmethod
def _normalize_identifier(id_type: IdentifierType, value: str) -> str | None:
    """Normalize a device identifier value for matching, or None if unusable."""
    if not value or id_type not in _MATCHABLE_TYPES:
        return None
    if id_type == IdentifierType.MAC_ADDRESS:
        if not is_valid_mac_address(value):
            return None
        return normalize_mac_for_matching(value)
    return value.replace("-", "").replace(":", "").replace("_", "").lower()
```

OzGav: Good catch, you’re right on all three points. I checked and an unrecognized stored key really does come back as UNKNOWN instead of raising an error, so that guard did nothing, and an all-zero MAC would have been treated as a real match. I’ve pushed a fix (9737d47) that follows your suggestion: the lookup now only considers the same identifier types as _identifiers_match, throws away invalid MACs using is_valid_mac_address, and drops the dead error handling. I also added two tests covering the all-zero MAC and the unknown key cases. I stopped short of calling _identifiers_match directly because it compares two live players and can fall back to matching on IP address, which is exactly what this lookup must never do, so I kept a small allowlist with a comment saying it has to stay in sync.

music_assistant/providers/universal_player/provider.py
```diff
@@ -572,6 +588,72 @@ async def _create_separate_universal_player(
...
+            if not isinstance(player_id, str) or not player_id.startswith(UNIVERSAL_PLAYER_PREFIX):
+                continue
+            raw_conf = all_player_configs[player_id]
+            if not isinstance(raw_conf, dict):
+                continue
+            values = raw_conf.get("values")
+            values = values if isinstance(values, dict) else {}
+            # a protocol player that was already a member is the strongest match
+            stored_protocol_ids = values.get(CONF_LINKED_PROTOCOL_IDS)
+            if isinstance(stored_protocol_ids, list) and protocol_ids.intersection(
```
marcelveldt: [CRITICAL] This can pick an obsolete universal-player config when multiple retained configs still list the same protocol ID, even if the child’s `protocol_parent_id` points to the current one. Please prefer that canonical parent and ignore contradictory reverse memberships, otherwise this can change the ID again and load stale settings.

music_assistant/providers/universal_player/provider.py
```diff
@@ -572,6 +588,72 @@ async def _create_separate_universal_player(
...
+        # restricted to the same identifier types _identifiers_match uses, so both
+        # apply the same notion of device identity. This notably excludes IP
+        # addresses (change with DHCP, shared by multi-instance hosts) and UNKNOWN.
+        if not value or id_type not in MATCHABLE_IDENTIFIER_TYPES:
+            return None
+        if id_type == IdentifierType.MAC_ADDRESS:
+            if not is_valid_mac_address(value):
+                return None
+            return normalize_mac_for_matching(value)
+        return value.replace("-", "").replace(":", "").replace("_", "").lower()
```
marcelveldt: [PROBLEM] This still differs from `_identifiers_match`: it strips underscores and misses the UUID `_MR` equivalence handled there. Please extract the strong-identifier matching into a shared helper, with only the unsafe persisted-IP fallback excluded here.

tests/controllers/players/test_protocol_linking.py
```diff
@@ -849,6 +849,207 @@ def test_device_key_from_no_identifiers_falls_back_to_player_id(
...
+    """
+    Tests for reusing the stored device key of an earlier universal player.
+
+    Regression tests for music-assistant/support#5888: re-creating a universal
+    player must not change its player id, even when a different set of protocol
+    players or identifiers is available at re-creation time. Consumers such as
+    the Home Assistant integration bind entities to that id.
+    """
+
+    def test_stored_key_reused_via_linked_protocol_ids(self, mock_mass: MagicMock) -> None:
```
marcelveldt: [PROBLEM] These tests only cover a single stored candidate, so they miss the stale-config case that can re-key the player. Please add a lifecycle regression with old and current universal configs where `protocol_parent_id` identifies the current parent, and assert the ID and settings stay with it.

music_assistant/providers/universal_player/provider.py
OzGav: Agreed, this was a real gap. The lookup now checks the parent link stored on the protocol player itself first and returns that universal player when it points at one of ours, before any membership lists are considered. When it does fall back to membership lists, a member whose own parent link points somewhere else is ignored, so a leftover config from before a re-key can no longer steal the id or bring back old settings. Fixed in cc1a85b with a lifecycle test where an obsolete config still lists the member, the current one is named by the parent link, and the current one must win.

music_assistant/providers/universal_player/provider.py
OzGav: You are right, my copy stripped underscores and did not know about the RINCON and _MR equivalence, so the two rule sets had already drifted. I went with the extraction you suggested: the strong identifier comparison now lives in a shared strong_identifiers_match helper in [helpers/util.py](https://github.com/music-assistant/server/blob/claude/github-issue-5888-u80s0j/helpers/util.py), used both by _identifiers_match and by the stored key lookup. Only the parts that need live players stay in _identifiers_match, meaning the reported MAC comparison and the IP fallback, and the IP fallback is exactly what the stored lookup must never use. Done in cc1a85b, with a test showing a stored RINCON uuid matches its _MR variant.

tests/controllers/players/test_protocol_linking.py
OzGav: The new test sets up two universal player configs that both list the same protocol player as a member, with the stale one sorting first, while the protocol player's stored parent link names the current one. The test asserts the lookup returns the current player's key, so the id and everything stored under it stay with the right player. I also added tests for the provider ownership check and the shared uuid equivalence rule from the other comments.

OzGav:
I did note another comment in the email I received and so made changes in line with that as well. As the prefix alone could be too weak since providers are free to choose their own id formats, the lookup now also requires the stored config to belong to the universal player provider itself, and the same check is applied when validating the parent link from the first comment. Fixed in cc1a85b, with a test where a native player config whose id happens to start with "up" is correctly ignored.

music_assistant/providers/universal_player/provider.py
```diff
@@ -116,6 +120,7 @@ async def create_universal_player(
             enabled=True,
             values={
                 CONF_LINKED_PROTOCOL_IDS: protocol_player_ids,
+                CONF_CREATED_AT: int(time.time()),
```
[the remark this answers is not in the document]
marcelveldt: Fair point, fixed in 76b84cd. Switched the stored value to `time.time_ns()` so two wrappers created in the same discovery second still order properly. Note the tiebreak was never non-deterministic (it fell through to the player id, which is stable once minted), but "oldest wins" did degrade to arbitrary, and that is the bit worth having. Added a test that two back-to-back wrappers get distinct timestamps.

music_assistant/controllers/players/protocol_linking.py
```diff
@@ -798,26 +799,30 @@ def _select_merge_winner(
...
+    def _merge_rank(self, universal_player: Player) -> tuple[int, int, str]:
+        """Return the sort key of a universal player in a merge, lowest wins."""
+        created_at = self.mass.config.get(
+            f"{CONF_PLAYERS}/{universal_player.player_id}/values/{CONF_CREATED_AT}", 0
+        )
+        return (
+            -len(universal_player.linked_output_protocols),
+            created_at if isinstance(created_at, int) else 0,
+            universal_player.player_id,
+        )
```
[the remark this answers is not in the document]
marcelveldt: Agreed on the outcome, changed in 6768782. Age is now the primary rank and the link count only breaks a tie. The "oldest wins" wording was only ever in the PR description though - the code and its docstring said "most links, oldest on a tie" - and links-first was deliberate pre-existing behaviour from #4017, so I did not want to flip it purely on the wording.

What convinced me is the upgrade path: existing wrappers have no stored timestamp so they rank oldest, but under links-first that protection did nothing. A device that forms a new two-protocol wrapper would beat a restored single-protocol legacy one and throw away exactly the id HA has been bound to. Nothing is lost by keeping the smaller player - the absorbed one's links move over regardless, and a merge with overlapping domains is refused upstream anyway. Added a test for the older-with-fewer-links case.

music_assistant/controllers/players/protocol_linking.py
```diff
@@ -50,6 +50,7 @@
 from music_assistant.providers.sync_group.constants import CONF_ALLOWED_MEMBERS
 from music_assistant.providers.universal_player import UniversalPlayer, UniversalPlayerProvider
 from music_assistant.providers.universal_player.constants import (
+    CONF_CREATED_AT,
```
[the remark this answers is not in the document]
marcelveldt: Good catch, that one was a real bug - fixed in 9748134. `created_at` is wrapper bookkeeping and had no business travelling, so it is in `UNIVERSAL_PLAYER_INTERNAL_CONF_KEYS` now, which covers both the merge and the native replacement path. Added a test that fails if it is taken out again.

marcelveldt:
On the suppressed README comment: correct, the "only ever deleted" list was missing the merge case - `_stop_and_unregister` unregisters the absorbed player permanently, which deletes its config. Reworded in 9748134 to cover all three cases and to note that the merge and native-replacement paths carry the settings over first.

PR #5563, opened 2026-08-10
Keep events and settings working when a path is not valid UTF-8

OzGav:

Files with names that are not valid UTF-8, such as a music library named before UTF-8 became standard where a letter like ä or ß was saved in the older Latin-1 encoding, reach us as text that cannot be written back out.

The scan picked such a file up, tried to add it to the library, failed, and then named the file in a task message. That message goes to two places that both have to be written out as text. One is the task update sent to every connected client. The other is the task state saved in settings.json. Both writes failed with "str is not valid UTF-8: surrogates not allowed", so the server stopped sending updates to clients and settings.json was never written again. Every setting the user changed after that point was lost on the next restart, which is what was reported.

These files can never be added to the library in any case, because the database cannot store a name it cannot read. So they are now skipped where they are first read, both during a library scan and when browsing a folder, and the name is logged with the unreadable bytes shown as escapes so the file can be found. Nothing further down ever receives a name it cannot handle.

The check only looks at names that are not plain ASCII, so it costs nothing for almost every file. The local, SMB, NFS, WebDAV and cloud providers all share this code, so one check covers all of them.

- related issue https://github.com/music-assistant/support/issues/6042

## Types of changes

## Checklist


music_assistant/controllers/tasks/controller.py
```diff
@@ -32,6 +32,7 @@
     has_scope,
 )
 from music_assistant.helpers.api import api_command
+from music_assistant.helpers.json import make_utf8_safe
```
marcelveldt: shouldn't we be fixing this at the root instead of now punishing every call though this helper ?

OzGav: Yes can do it in the file system provider scan code as that will fix the reported problem. The potential downside is if non-UTF8 ever comes in via a different path in the future then that will require another fix.

OzGav: OK switched to guarding the local file system provider

music_assistant/controllers/tasks/controller.py
```diff
@@ -574,7 +577,7 @@ def _append_task_log(self, task_id: str, line: str) -> None:
         if not (managed := self._tasks.get(task_id)):
             return
         logs = managed.task_info.logs
-        logs.append(line)
+        logs.append(make_utf8_safe(line))
```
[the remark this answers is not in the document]
OzGav: Outdated

marcelveldt:
~~This is a better fix: https://github.com/music-assistant/server/pull/5653 which fixes it at the root instead of working around it~~

actually scrap that, that is already the default in Python and won't fix this issue at all

music_assistant/helpers/json.py
```diff
@@ -45,6 +45,50 @@ def get_serializable_value(obj: Any) -> Any:
...
+    if isinstance(value, list):
+        return [make_utf8_safe(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(make_utf8_safe(item) for item in value)
+    if isinstance(value, set):
+        return {make_utf8_safe(item) for item in value}
+    if (serializable := get_serializable_value(value)) is not value:
+        # objects (such as dataclasses) only expose their strings once converted
+        return make_utf8_safe(serializable)
+    return value
```
marcelveldt: In production this only ever gets a `str` - the dict/list/tuple/set walk and the object branch are only reached from conftest and the new test.

Can we keep just `make_utf8_safe(value: str) -> str` here and leave the container walk in `tests/common.py` where it came from? That drops both overloads too and makes the diff smaller than it is now.

The object branch also returns a dict instead of the object you gave it (as the test asserts) - bit of a trap to leave in a shared helper.

music_assistant/providers/filesystem_local/helpers.py
```diff
@@ -500,7 +505,11 @@ def nat_key(name: str) -> tuple[int | str, ...]:
                 continue
             if not (is_dir or is_file):
                 continue
-            if entry.name in IGNORE_DIRS or entry.name.startswith("."):
+            if (
+                entry.name in IGNORE_DIRS
+                or entry.name.startswith(".")
+                or _skip_undecodable_name(entry.name, logger)
+            ):
```
marcelveldt: Ordering here is the other way round from `recursive_iter`, where the guard runs before `is_dir`/`is_file`.

Matters a bit because the EINVAL branch just above logs the raw `entry.name` - I checked, and logging a surrogate name to the utf8 file handler throws and the line gets dropped silently, so you'd lose exactly the diagnostic you're adding. I couldn't prove that combination actually occurs in practice, but moving this block to the top of the loop is one line and makes both call sites read the same.

music_assistant/providers/filesystem_local/helpers.py
```diff
@@ -523,6 +532,25 @@ def nat_key(name: str) -> tuple[int | str, ...]:
...
+    """
+    Return True if the given filename is not valid UTF-8 and must be skipped.
+
+    A skipped name is logged in escaped form, so the caller only has to skip it.
+
+    :param name: Name of the file or directory, as returned by the os module.
+    :param log: Logger to report a skipped name on.
+    """
+    # such a path can be neither stored in the database nor sent to a client;
+    # isascii() is O(1) and true for almost every name, keeping this off the hot path
```
marcelveldt: "true for almost every name" doesn't really hold for german/nordic/CJK libraries, most names there are non-ascii and take the encode path. It's nothing next to the file IO anyway, so I'd just drop that half of the comment.

tests/providers/filesystem/test_helpers.py
```diff
@@ -474,6 +474,36 @@ def test_scan_errors_describe_names_examples() -> None:
     assert len(errors.failed_paths) == helpers.MAX_REPORTED_FAILED_PATHS
 
 
+def test_recursive_iter_skips_names_that_are_not_valid_utf8(
```
marcelveldt, line 477: Can you add the same test for `sorted_scandir`? It's not only browse - it feeds 7 call sites via `_scandir` (browse, podcast episodes, playlist images, audiobook parsing x2, local images, chapters), so it's the busier of the two paths and the skip is untested there.

Copilot raised this one earlier and it got suppressed.

marcelveldt (commented):
Thanks for moving this into the provider, much better place for it 👍

Few things inline, main one is the missing test on `sorted_scandir` - that path is busier than the description suggests.

Also: branch is 85 commits behind dev so it needs a rebase. And the copilot comment about `last_error` is outdated now the tasks controller changes are gone, can you resolve that thread?

music_assistant/helpers/json.py
OzGav: Agreed, production only ever passes a string, so the helper now takes and returns a string and the container walk has gone back to tests/common.py where it came from, which also removes the object branch that handed you back a dict instead of what you passed in.

music_assistant/providers/filesystem_local/helpers.py
OzGav: Good catch, moved it to the top of the loop so both places read the same, and you are right that a name like this would otherwise reach the branch above and have its log line silently dropped.

music_assistant/providers/filesystem_local/helpers.py
OzGav: Fair, that is not true for those libraries, so I have dropped thatas suggested.

tests/providers/filesystem/test_helpers.py
OzGav: Added

marcelveldt (approved):

PR #5571, opened 2026-08-10
Add a configurable volume step size to the players core config

quadcom:
The four volume nudge commands (`cmd_volume_up`/`_down` and their group equivalents) each hardcoded the same adaptive ladder: step 1 below 10 or above 90, step 2 below 30 or above 70, step 3 in the middle. That ladder now lives behind a single `_get_volume_step` helper, and a new `volume_step` integer entry on the `players` core config overrides it with a flat step.

`0` is the default and means "keep the adaptive ladder", so behaviour on existing installs is unchanged. The value is read per command rather than cached, so a change takes effect without a reload.

At the request of maintainers, putting the setting on the server rather than in a front-end UI makes it apply consistently everywhere: the scroll wheel and tap-at-the-edges already call `players/cmd/volume_up`/`volume_down`, so they have always followed the server ladder, and a frontend-only setting could never reach them.

This PR supports the front-end PR (https://github.com/music-assistant/frontend/pull/2274), which adds volume control options ("absolute" and "relative") and presents the volume-steps server configuration changes made here.

## Types of changes

## Checklist

- [X] I have read and complied with the project's AI Policy for any AI-assisted contributions.


music_assistant/controllers/players/strings.json
```diff
@@ -1,4 +1,10 @@
 {
+  "config_entries": {
+    "volume_step": {
+      "label": "Volume step size",
+      "description": "How much the volume changes per step when using volume up/down, the scroll wheel, or tapping the ends of a volume slider.\n\nLeave at 0 to use the adaptive default, which takes smaller steps near the ends of the range."
```
marcelveldt:
```suggestion
      "description": "How much the volume changes per step when using volume up/down commands, For example within the web interface by using the scroll wheel, tapping the ends of a volume slider or when using Home Assistant to issue a volume up or down command.\n\nLeave at 0 to use the adaptive default, which takes smaller steps near the ends of the range."
```

marcelveldt (approved):
Nice and clean. Thanks, @quadcom !

PR #5573, opened 2026-08-10
Rework and refine the Bose SoundTouch provider

fmunkes:

I did the following things:
- Reworked the client to parse the xmls into mashumaro-based dataclasses.
- Verified, that the present tests also work with the reworked client.
- Wrote the client in a way, which could potentially move it into its own future lib.
- Using an asyncio.lock where applicable, that improved stability quite a bit.
- Switch the default audio format to mp3 - at least for me that was way more stable, though of course the other options still exist
- Removed the dependency on the dlna provider - of course, it's still a upnp player, but the few commands we need, we can just generate from present dlna helpers.
- Added the ability to store a (fake) preset on the player, and detect present ones. If the player has no preset stored at all, the button will not emit the event, thus not work.
- Added two player options, bass (writable) and network name (only read currently). Future PRs may enhance this.
- Tried to reduce redundancy/ code where possible.
- Forward preset selection to the group leader, if there is a group leader

If we want to stick with the previous approach of parsing the xmls, that is fine, and I can undo that. It was still a chance for me to get to know this API.

Future follow-up:
- Presets are still configured per provider - it should be easy to have the provider ones as a global fallback, and optional individual ones per player
- Add some more player options should that be possible.
- The ws client should become part of the client too. It hasn't been touched.

## Types of changes

## Checklist


music_assistant/providers/bose_soundtouch/client/client/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+
+    async def add_zone_members(self, zone: Zone) -> None:
+        """Add zone.members to a zone."""
+        await self._add_or_remove_zone_members(zone, add_members=True)
+
+    async def remove_zone_members(self, zone: Zone) -> None:
+        """Remove zone.members from a zone."""
+        await self._add_or_remove_zone_members(zone, add_members=False)
+
+    async def get_now_playing(self, endpoint: str = "nowPlaying") -> NowPlaying:
```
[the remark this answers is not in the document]
fmunkes: The endpoint is supported in both camel and non-camel-case. I'll add another try for a not found with non-camel-case.

music_assistant/providers/bose_soundtouch/client/client/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+    return string
+
+
+def create_zone_xml(zone: Zone, sender_ip: str | None = None) -> str:
+    """Create zone xml."""
+    assert zone.leader
+    assert zone.leader.mac is not None
+
+    if sender_ip:
+        xml = f'<zone master="{zone.leader.mac}" senderIPAdress="{sender_ip}">'
```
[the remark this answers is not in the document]
fmunkes: Good catch. Luckily, the client uses the version without send address. Fixed.

music_assistant/providers/bose_soundtouch/client/client/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+        """Get now playing."""
+        element = await self._get(endpoint)
+        d: dict[str, Any] = element.attrib
+        for el in element:
+            if el.tag == "ContentItem":
+                item_name: str | None = None
+                for sub_el in el.iter():
+                    if sub_el.tag == "itemName":
+                        item_name = sub_el.text
+                d["content_item"] = {**el.attrib, "item_name": item_name}
```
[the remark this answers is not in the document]
fmunkes: fixed.

music_assistant/providers/bose_soundtouch/client/exceptions.py
```diff
@@ -0,0 +1,13 @@
+"""Exceptions for Soundtouch."""
+
+
+class SoundtouchError(Exception):
+    """Base exception."""
```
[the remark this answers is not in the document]
fmunkes: valid and fixed.

music_assistant/providers/bose_soundtouch/client/client/session_configuration.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@dataclass(kw_only=True)
+class SessionConfiguration:
+    """Session configuration for a speaker client."""
+
+    session: ClientSession
+    ip: str
+    http_port: int = 8090
+    timeout: ClientTimeout = DEFAULT_TIMEOUT
```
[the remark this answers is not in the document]
fmunkes: good catch, fixed.

music_assistant/providers/bose_soundtouch/client/client/__init__.py
```diff
@@ -0,0 +1,334 @@
...
+    def __init__(self, session_configuration: SessionConfiguration) -> None:
+        """Initialize."""
+        self.session_config = session_configuration
+
+        if self.session_config.logger is None:
+            self.logger = logging.getLogger(__name__)
+            logging.basicConfig()
+            self.logger.setLevel(logging.DEBUG)
+        else:
+            self.logger = self.session_config.logger
```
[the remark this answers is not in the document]
fmunkes: ah, that's an issue, because I had the client separate initially. I pass the logger. In the future, I'd like to move the client code, including the websocket part, into its dedicated lib.
Fixed.

music_assistant/providers/bose_soundtouch/avt_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def get_headers(xml: str, soap_action: str) -> dict[str, str]:
+    """Get headers for MusicCast."""
+    return {
+        "Content-Type": 'text/xml; charset="utf-8"',
+        "SOAPACTION": f'"{soap_action}"',
+        "Accept": "*/*",
+        "User-Agent": "MusicAssistant",
+        "Content-Length": str(len(xml)),
```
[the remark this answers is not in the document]
fmunkes: Let's try this. This lives in MusicCast like this, so maybe I can remove it there too...

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -341,90 +544,93 @@ async def _refresh_sources(self) -> None:
...
         else:
             self._attr_group_members = []
-        self.update_state()
 
-    def _update_state_from_now_playing(self, now_playing: SoundTouchNowPlaying) -> None:
+    def _update_state_from_now_playing(self, now_playing: NowPlaying) -> None:
         """Update player state from a now_playing snapshot."""
-        self._attr_powered = now_playing.source != SOURCE_STANDBY
+        if now_playing.content_item:
+            self._attr_powered = now_playing.content_item.source != SOURCE_STANDBY
```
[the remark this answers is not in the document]
fmunkes: good catch, fixed.

music_assistant/providers/bose_soundtouch/provider.py
```diff
@@ -113,6 +115,18 @@ async def loaded_in_mass(self) -> None:
             if stripped := ip_address.strip():
                 await self.try_add_player(stripped)
 
+        self._on_unload_callbacks: list[Callable[[], None]] = []
+        self._on_unload_callbacks.append(
+            self.mass.webserver.register_dynamic_route(
+                f"/{self.instance_id}_preset", self._handle_preset_request
+            )
```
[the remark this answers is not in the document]
fmunkes: adapted it to webserver. That endpoint is probably not needed at all, but I had it in my testing to get the presets going. Let's have it for the time being, it does no harm.

music_assistant/providers/bose_soundtouch/avt_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+def get_upnp_ctrl_url(device: BoseSoundTouchPlayer) -> str:
+    """Get UPNP control URL."""
+    return f"http://{format_ip_for_url(device._client.session_config.ip)}:{UPNP_PORT}/{UPNP_CONTROL_ENDPOINT}"
+
+
+async def _post_soap(
+    client: aiohttp.ClientSession,
+    ctrl_url: str,
+    xml: str,
+    soap_action: str,
```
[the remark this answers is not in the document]
fmunkes: Moved it to the bottom of the file.

music_assistant/providers/bose_soundtouch/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def source_id(source: str, source_account: str | None) -> str:
+    """Build a stable source id from a SoundTouch source and account."""
+    return f"{source}:{source_account}" if source_account else source
+
+
+def _local_name(tag: str) -> str:
+    """Return an XML tag name without its namespace prefix."""
+    return tag.rsplit("}", 1)[-1]
```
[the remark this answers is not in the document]
fmunkes: moved it to the bottom of the file

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -239,19 +316,80 @@ def device_id(self) -> str:
...
+        base_entries = await super().get_config_entries()
+
+        default_entries = [
+            # by far the most consistent results with mp3. flac is flakier.
+            CONF_ENTRY_OUTPUT_CODEC_DEFAULT_MP3,
+            CONF_ENTRY_FLOW_MODE,
+            create_sample_rates_config_entry(max_sample_rate=192000, max_bit_depth=24),
+        ]
+
+        current_presets = await self._client.get_presets()
```
OzGav, line 340: This calls the speaker every time the settings page is rendered. If the speaker is asleep or unreachable I am thinking this will raise and stop the page opening?

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -318,21 +461,81 @@ async def _handle_preset(self, preset_id: int) -> None:
...
+            bass_capability = await self._client.get_bass_capabilities()
+            if bass_capability.available is not None and bass_capability.available:
+                self._supported_player_options.add(PlayerOptionKeys.BASS)
+        _updated_options: list[PlayerOption] = []
+        if PlayerOptionKeys.NETWORK_NAME in self._supported_player_options:
+            info = await self._client.get_info()
+            _updated_options.append(
+                PlayerOption(
+                    key=PlayerOptionKeys.NETWORK_NAME,
+                    translation_key=PlayerOptionKeys.NETWORK_NAME,
```
OzGav: I can't see a label for this one anywhere?

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -79,52 +121,41 @@ def __init__(
...
-            await self._refresh_volume()
-        except (aiohttp.ClientError, TimeoutError, OSError) as err:
-            self.logger.debug("Poll failed for %s: %s", self.name, err)
-            if self._attr_available:
+        async with self._update_lock:
+            try:
+                await self._refresh_now_playing()
+                await self._refresh_volume()
+                await self._refresh_zone()
+                await self._refresh_options()
```
OzGav: This is running on every poll and getting values that rarely change. Could you read the capabilities once and refresh bass only when it actually changes.

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -207,28 +256,56 @@ async def play_announcement(
...
         if player_ids_to_remove:
-            members = self._member_tuples(player_ids_to_remove)
-            if members:
-                await self._client.remove_zone_slaves(self._device_id, members)
+            zone = get_zone(player_ids_to_remove)
+            await self._client.remove_zone_members(zone)
+
         await self._refresh_zone()
-        for member_id in (player_ids_to_add or []) + (player_ids_to_remove or []):
-            self.mass.players.trigger_player_update(member_id)
```
OzGav, line 231: With this removed did the other speakers in the group show valid group membership?

music_assistant/providers/bose_soundtouch/avt_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+from music_assistant.providers.bose_soundtouch.const import UPNP_CONTROL_ENDPOINT, UPNP_PORT
+
+if TYPE_CHECKING:
+    from music_assistant.providers.bose_soundtouch.player import BoseSoundTouchPlayer
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_headers(soap_action: str) -> dict[str, str]:
+    """Get headers for MusicCast."""
```
OzGav: Nice, not reinventing the wheel :-) I was thinking though that since this is now in two places could it go in `helpers/upnp.py` ?

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -147,52 +178,70 @@ async def on_unload(self) -> None:
...
             # the POWER key toggles standby, so only send it when a change is needed
-            await self._client.press_key(KEY_POWER)
+            await self._client.press_key(Key.POWER)
         self._attr_powered = powered
         self.update_state()
 
     async def volume_set(self, volume_level: int) -> None:
         """Handle VOLUME_SET command on the player."""
-        await self._client.set_volume(volume_level)
+        await self._client.set_volume(volume_level, mute=False)
```
OzGav, line 194: I am wondering if `_attr_volume_muted` should be set here as well?

music_assistant/providers/bose_soundtouch/provider.py
```diff
@@ -152,7 +170,7 @@ async def try_add_player(self, ip_address: str) -> None:
 
         player = BoseSoundTouchPlayer(self, player_id, client, info)
         try:
-            await player.setup()
+            await player.setup(info)
             await self.mass.players.register_or_update(player)
         except Exception:
```
OzGav: I know this is existing code but it would be nice to narrow it if possible...?

music_assistant/providers/bose_soundtouch/strings.json
```diff
@@ -4,6 +4,42 @@
...
+    },
+    "action_overwrite_preset_4": {
+      "label": "Overwrite preset 4"
+    },
+    "action_overwrite_preset_5": {
+      "label": "Overwrite preset 5"
+    },
+    "action_overwrite_preset_6": {
+      "label": "Overwrite preset 6"
+    },
```
OzGav: These twelve entries are the same two sentences six times over. `ConfigEntry` supports `translation_params`, and you already uses it for "Assign result to preset {0}" in `config.py`. Two strings with a {0} would replace all twelve.

music_assistant/providers/bose_soundtouch/provider.py
```diff
@@ -174,3 +192,27 @@ def _get_player_by_ip(self, ip_address: str) -> BoseSoundTouchPlayer | None:
...
+        """
+        if not (preset_id := request.query.get("preset_id")):
+            return web.Response(status=400, text="Missing preset_id")
+
+        return web.json_response(
+            {
+                "audio": {
+                    "hasPlaylist": False,
+                    "isRealtime": True,
+                    "streamUrl": f"{self.mass.webserver.base_url}/{self.instance_id}_preset?preset_id={preset_id}",
```
OzGav: When the press leads to playback this is all moot, since Music Assistant grabs the source straight away. I am less sure about the cases where it does not, like a preset button that is enabled but has no media mapped to it yet. There the speaker is left holding a station whose stream URL returns JSON rather than audio. Does it error out and stop, or keep trying?

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -318,21 +461,81 @@ async def _handle_preset(self, preset_id: int) -> None:
...
-        self._attr_volume_level = volume.level
-        self._attr_volume_muted = volume.muted
-        self.update_state()
+        self._attr_volume_level = volume.actual_volume
+        self._attr_volume_muted = volume.mute_enabled
 
     async def _refresh_now_playing(self) -> None:
         """Refresh playback state from the speaker."""
-        self._update_state_from_now_playing(await self._client.get_now_playing())
+        with contextlib.suppress(ApiError):
```
OzGav: This hides every API error from the now playing refresh, including inside poll, which is what the availability tracking at line 148 depends on to notice a speaker has gone offline. What failure is this covering?

music_assistant/providers/bose_soundtouch/player.py
```diff
@@ -59,14 +71,44 @@ def __init__(
...
+
+        self._listener_task = self.mass.create_task(self._listen())
+
+    def set_static_attributes(self, info: Info) -> None:
+        """Set static attributes."""
+        self._attr_needs_poll = True
+        self._attr_poll_interval = IDLE_POLL_INTERVAL
+        self._attr_available = True
         self._attr_type = PlayerType.PLAYER
         # Playback (PLAY_MEDIA) is deliberately omitted: SoundTouch has no usable API to
```
OzGav: PLAY_MEDIA is added below now

music_assistant/providers/bose_soundtouch/const.py
```diff
@@ -2,17 +2,24 @@
...
 # Bose SoundTouch exposes a local HTTP API on port 8090 and a websocket
 # notification channel on port 8080 (the "gabbo" subprotocol).
-API_PORT = 8090
 NOTIFICATION_PORT = 8080
+# e.g. http://1.2.3.4:8091/Xml/AVTransport3.xml
+UPNP_PORT = 8091
+UPNP_CONTROL_ENDPOINT = "AVTransport/Control"
 WS_SUBPROTOCOLS = ("gabbo",)
 WS_HEARTBEAT = 30
 REQUEST_TIMEOUT = 10
```
OzGav, line 25: I dont think is used anymore

music_assistant/providers/bose_soundtouch/const.py
```diff
@@ -2,17 +2,24 @@
...
 PLAYER_ID_PREFIX = "bose_soundtouch_"
 
+IDLE_POLL_INTERVAL = 30
+PLAYBACK_POLL_INTERVAL = 20
+
 # Optional Bose SoundTouch developer app key. When configured, announcements are
 # sent natively to the speaker as an overlay that ducks and resumes playback.
 CONF_APP_KEY = "app_key"
 
 # Bose SoundTouch exposes a local HTTP API on port 8090 and a websocket
```
OzGav, line 17: You have deleted the 8090 port below?

music_assistant/providers/bose_soundtouch/manifest.json
```diff
@@ -7,6 +7,5 @@
   "codeowners": ["@Odn0"],
```
OzGav:
```suggestion
  "codeowners": ["@fmunkes", "@Odn0"],
```

OzGav (commented):
Some random things I noticed :-)

music_assistant/providers/bose_soundtouch/player.py
fmunkes: If the speaker is unreachable, it should be marked unavailable by the regular poll (which then fails, but is handled). If the speaker is asleep, the call works and is necessary, otherwise I can't update the buttons.

music_assistant/providers/bose_soundtouch/player.py
fmunkes: good catch. I added it to the general strings, like the others. It is already available in HA, but I forgot in MA.

music_assistant/providers/bose_soundtouch/player.py
fmunkes: Thank you. I forgot that part. PlayerOptions are now initialized on start, and then updated by socket.

music_assistant/providers/bose_soundtouch/player.py
fmunkes: Yes. There is a socket event emitted. At least for the two speaker I have, group membership seems correctly displayed.

music_assistant/providers/bose_soundtouch/avt_helpers.py
fmunkes: I agree that there is redundancy, and you spotted that right. However, there are some header differences. Nonetheless parts of it can be moved to general helpers. I would prefer to do that in a follow-up post-stable-release PR, since I remember MusicCast to be very picky about the correct syntax, versus Soundtouch appearing to be more relaxing. I put that onto my personal to do list.
I did correct the description though ;)

music_assistant/providers/bose_soundtouch/player.py
fmunkes: Doesn't hurt, added.

music_assistant/providers/bose_soundtouch/provider.py
fmunkes: I'm catching MusicAssistantError and SoundtouchError now. I think these are the ones.

music_assistant/providers/bose_soundtouch/strings.json
fmunkes: Guess I was a little sleepy here. Fixed.

music_assistant/providers/bose_soundtouch/player.py
fmunkes: I'll remove that.

music_assistant/providers/bose_soundtouch/player.py
fmunkes: Fixed.

music_assistant/providers/bose_soundtouch/const.py
fmunkes: ~Removed.~ And added back, it's used in the avt_helpers :)

music_assistant/providers/bose_soundtouch/const.py
fmunkes: It moved as default argument into the client. The websocket part is not yet part of the dedicated client, so that port still lives in this file. I'd like to have a separate lib as the final goal.

music_assistant/providers/bose_soundtouch/manifest.json
fmunkes: ok, I added myself :)

music_assistant/providers/bose_soundtouch/provider.py
fmunkes: I removed that entirely. In my testing, the speaker doesn't access this at all. The only important part is, that the speaker has something stored there, otherwise the socket event is not emitted. That is what config entries in player are for now (store_preset and so on). I can't tell what happens if there is something stored previously, as my speaker were reset, and the old Soundtouch app doesn't work with my phone - the whole ecosystem is not supported anymore. So in the best case, at least as far as I can see, we make these buttons usable again.

music_assistant/providers/bose_soundtouch/provider.py
```diff
@@ -152,9 +158,9 @@ async def try_add_player(self, ip_address: str) -> None:
 
         player = BoseSoundTouchPlayer(self, player_id, client, info)
         try:
-            await player.setup()
+            await player.setup(info)
             await self.mass.players.register_or_update(player)
-        except Exception:
+        except MusicAssistantError, SoundtouchError:
```
OzGav:
```suggestion
        except MusicAssistantError, aiohttp.ClientError, TimeoutError, OSError:
```
I'm not an expert on this so I checked with Claude and it said `SoundtouchError` subclasses `aiohttp.ClientError`, so this does not catch a plain connection failure or timeout. Hence the suggestion above.

music_assistant/providers/bose_soundtouch/provider.py
```diff
@@ -5,13 +5,15 @@
 from typing import TYPE_CHECKING, cast
 
 import aiohttp
+from music_assistant_models.errors import MusicAssistantError
 from zeroconf import ServiceStateChange
 
 from music_assistant.constants import CONF_ENTRY_MANUAL_DISCOVERY_IPS
 from music_assistant.helpers.util import get_primary_ip_address_from_zeroconf
 from music_assistant.models.player_provider import PlayerProvider
+from music_assistant.providers.bose_soundtouch.client.exceptions import SoundtouchError
```
OzGav:
```suggestion
```
If you agree with the suggestion below then this can be deleted.

OzGav (commented):
Just one thing looking at the changes.  Otherwise I think we can give this a go!

music_assistant/providers/bose_soundtouch/provider.py
fmunkes: Ok, doesn't hurt to be broader I think.

OzGav (approved):
Thanks Fabian!

PR #5577, opened 2026-08-10
Fix a brief audio hiccup on the first playback after starting the server

marcelveldt:

The first time a Sendspin stream starts in a running server, the audio library it uses for
resampling and encoding was loaded on the event loop, right in the middle of producing the
first audio. That froze audio production for up to half a second, which left a hole in the
play timeline — bridged players reported it as a ~560ms realign, and every other player type
would be affected the same way. It only happens once per server run, on the first play.

Loading it up front, off the event loop, removes the hiccup. Second and later playbacks were
already fine and are unchanged.

While tracking this down, the "slow SQL query" debug warning turned out to be misleading: it
timed how long the server *waited* for a query, not how long the query took. Any unrelated
freeze got reported as a slow query, which is exactly what sent the first diagnosis down the
wrong path — the query it blamed actually runs in 0.05ms. It now discounts time the server
spent blocked elsewhere.

- n/a — found while diagnosing the bridged AirPlay sync work in #5570

## Types of changes

## Changes

- Load PyAV in a thread when a Sendspin playback session starts, before the play timeline exists.
- Slow-query logging now subtracts time the event loop was blocked, so it only flags genuinely slow queries.
- Tests covering both directions of the slow-query warning.

## Checklist

### DIFF
--- a/music_assistant/helpers/database.py
+++ b/music_assistant/helpers/database.py
@@ -88,14 +88,13 @@ class _LoopStallTracker:
 
     def release(self) -> None:
         """Stop sampling once the last user has released the tracker."""
-        if not ENABLE_DEBUG:
-            return
         self._users = max(0, self._users - 1)
         if self._users == 0 and self._task is not None:
             self._task.cancel()
             self._task = None
 
     async def _sample(self) -> None:
+        """Record how late each wake-up is, which is the time the loop was unavailable."""
         loop = asyncio.get_running_loop()
         while True:
             await asyncio.sleep(_STALL_SAMPLE_INTERVAL)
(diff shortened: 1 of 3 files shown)


marcelveldt:
Good catch on the stall accounting - that one was real. The start marker read the raw sampled total while the end read total + in-flight overdue, so a stall that ended just before a query started got booked afterwards and subtracted from that query. A genuinely slow query right after a freeze would have gone unreported.

Both ends now take the same reading (`_loop_stalls.stalled`, which includes the overdue term), so two readings bracket exactly the stalls between them. Added `test_slow_query_warning_survives_a_stall_that_precedes_it` which fails on the old asymmetric version and passes now.

Didn't add the clamp - with matching readings the value is monotonic, so it can't go negative in a way that matters, and a negative just means no warning anyway.

marcelveldt:
Second round - one taken, two declined.

**`_tracking_loop_stalls` set even when tracking is off** - fair, fixed. `acquire()` now returns whether it actually took a slot and `setup()` stores that, so a connection that never started sampling can't release one later.

**The two 0.5s stalls in the tests** - keeping them. They're not arbitrary padding, the margins are what make the tests meaningful:
- The stall-discount test's residual is capped by the sampler interval (0.05s), not by how long we block, so a shorter stall doesn't buy accuracy - it just narrows the gap to the 0.2s threshold on the failing side.
- The precede test only catches the bug while the stall exceeds the query's own duration (~0.17-0.27s here). Drop it to 0.3s and on a fast machine the test passes whether or not the bug is there, which is worse than a slow test.

Costs ~1s on a suite that runs in 100s, and both fail if you revert the fix - I checked.

PR #5584, opened 2026-08-11
Add Yoto provider

pantsman0:

This allows the streaming of cards (purchased, free, or make-your-own) but not Yoto radio streams. This also does not cache content -  if a card in the yoto library is scanned into another family's library, loading the next stream will fail. This can be the end of a chapter, or for cards with multiple tracks per chapter it will cut out at a track boundary.

https://github.com/music-assistant/music-assistant.io/pull/946

## Types of changes

## Checklist

- [X] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate.


music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+DEFAULT_HEADERS = {
+    "User-Agent": (
+        "Yoto/3.37 (com.yotoplay.yoto; build:15572; Android 14) Dalvik/2.1.0 "
+        "(Linux; U; Android 14; Pixel 4 XL Build/AP2A.240805.005)"
+    ),
+    "X-Yoto-App-Version": "3.37.15572",
+    "X-Yoto-Appisalpha": "false",
+    "Accept": "application/json",
+}
```
OzGav: This extracts the first-party Yoto Android app's Auth0 client ID, spoofs its User-Agent and version headers, and uses the Resource Owner Password grant to log in as that app. 

It's important that we identify as Music Assistant. Yoto publishes a developer API at yoto.dev with its own client registration and a device/PKCE flow. Please use that with an MA-registered client and MA's own User-Agent. If there's a reason that path won't work, say so in the PR and let us make the call

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+    _access_token: str | None = None
+    _library_cards: list[dict[str, Any]] | None = None
+
+    async def get_config_entries(self) -> tuple[ConfigEntry, ...]:
+        """Return the config entries for this provider instance."""
+        return ()
+
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await self._login()
```
OzGav: Move to `handle_async_init`. `loaded_in_mass `exceptions are swallowed (mass.py:1379-1389), so a failed login still leaves the provider looking set up. Let the error propagate from `handle_async_init` so the setup flow shows it; just make sure it is an MA error (LoginFailed, plus aiohttp.ClientError/TimeoutError mapped) and not a raw aiohttp one.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+class YotoProvider(MusicProvider):
+    """Yoto music/story provider implementation."""
+
+    _access_token: str | None = None
+    _library_cards: list[dict[str, Any]] | None = None
+
+    async def get_config_entries(self) -> tuple[ConfigEntry, ...]:
+        """Return the config entries for this provider instance."""
+        return ()
```
OzGav: You can remove this as it is the same as the parent class

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+    async def loaded_in_mass(self) -> None:
+        """Call after the provider has been loaded."""
+        await self._login()
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True if the provider is a streaming provider."""
+        return True
```
OzGav: You can remove this as it is the same as the parent class

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+        return StreamDetails(
+            provider=self.instance_id,
+            item_id=item_id,
+            audio_format=AudioFormat(content_type=content_type),
+            media_type=MediaType.TRACK,
+            stream_type=StreamType.CUSTOM,
+            data={"track_urls": track_urls},
+            allow_seek=True,
+            can_seek=True,
+        )
```
OzGav: These two branches can be one. Build a list of `MultiPartPath`, one per track URL, and pass it as path with `stream_type=StreamType.HTTP`. See [audiobookshelf/__init__.py:926-954](https://github.com/music-assistant/server/blob/dev/audiobookshelf/__init__.py#L926-L954), including how it passes a plain path when there is only one file. `get_audio_stream` below can then go, since MA handles multi file items itself. That also gives you working seeking, which the current code claims to support but does not do, and drops the 30 second timeout that would cut off any chapter slower than that to fetch.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+        }
+        async with self.mass.http_session.get(url, headers=headers) as response:
+            if response.status == 401:
+                await self._login()
+                headers["Authorization"] = f"Bearer {self._access_token}"
+                async with self.mass.http_session.get(url, headers=headers) as retry_res:
+                    retry_res.raise_for_status()
+                    return cast("dict[str, Any]", await retry_res.json())
+            response.raise_for_status()
+            return cast("dict[str, Any]", await response.json())
```
OzGav: `raise_for_status()` raises an aiohttp error. Library sync only catches MA errors, so one bad response from Yoto stops the entire sync instead of skipping that one album. Convert the failures inside _get before they get out: 401 or 403 to LoginFailed, 404 to MediaNotFoundError, 429 to RateLimited, 5xx to ResourceTemporarilyUnavailable. Catch `aiohttp.ClientError` and `TimeoutError` for connection problems and raise `ResourceTemporarilyUnavailable` for those too. Please catch those two specifically rather than catching everything.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+        artists: list[Artist | ItemMapping] = []
+        if author:
+            artists.append(
+                ItemMapping(
+                    item_id=author,
+                    provider=self.instance_id,
+                    name=author,
+                    media_type=MediaType.ARTIST,
+                )
+            )
```
OzGav, line 387: The album lists an artist, but there is no `get_artist` to look one up. When someone clicks the artist, the music controller calls it and the request fails.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+    async def sync_library(self, media_type: MediaType) -> None:
+        """
+        Run library sync for this provider.
+
+        :param media_type: Media type to sync.
+        """
+        if media_type in (MediaType.ALBUM, MediaType.ALL):
+            self._library_cards = await self._fetch_library_cards()
+        await super().sync_library(media_type)
```
OzGav: This whole override can go. The base class already calls `get_library_albums` once per sync, so prefetching here buys nothing, and` MediaType.ALL` on line 92 is a list of media types rather than a single one, so that half of the check never matches anyway. Instead of holding the cards on self, put `@use_cache(...)` from `music_assistant.controllers.cache` on `_fetch_library_cards` and `_resolve_card`, then drop `_library_cards` entirely. Right now that attribute is filled once and never refreshed, so `get_album` keeps serving whatever it saw first until the provider reloads. Caching` _resolve_card` also fixes the repeat calls, since today `get_album_tracks` fetches a card and then `get_stream_details` fetches the same card again for every track played. See [pandora/provider.py](https://github.com/music-assistant/server/blob/dev/pandora/provider.py) or [ard_audiothek/__init__.py](https://github.com/music-assistant/server/blob/dev/ard_audiothek/__init__.py).

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+    async def get_track(self, prov_track_id: str) -> Track:
+        """
+        Get full track details by id.
+
+        :param prov_track_id: Track ID formatted as {card_id}:{chapter_key}.
+        """
+        if ":" not in prov_track_id:
+            raise MediaNotFoundError(f"Invalid track ID format: {prov_track_id}")
+        card_id, _chapter_key = prov_track_id.split(":", 1)
```
OzGav, line 161: InvalidProviderID fits better than MediaNotFoundError here, since the problem is the shape of the id rather than a missing item. The same split and check is repeated at line 157, so a small shared helper that raises it in one place would be cleaner

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,434 @@
...
+
+        for url in streamdetails.data.get("track_urls", []):
+            async with self.mass.http_session.get(
+                url, timeout=aiohttp.ClientTimeout(total=30)
+            ) as response:
+                response.raise_for_status()
+                async for chunk in response.content.iter_chunked(64 * 1024):
+                    yield chunk
+
+    async def _login(self) -> None:
```
OzGav: A library sync running while someone browses will call _get from two places at once, both find no token, and both log in. Put an asyncio.Lock around this and check the token again once inside it.

tests/providers/yoto/test_parse_card.py
```diff
@@ -0,0 +1,175 @@
...
+def test_parse_track_from_card() -> None:
+    """Parse an album and tracklist from a resolved card/album API response."""
+    provider = _make_provider()
+    card = _make_multitrack_chapter_card()
+    chapters = card.get("content", {}).get("chapters", [])
+    album = provider._parse_album(card)
+
+    tracks: list[Track] = []
+    for idx, chapter in enumerate(chapters):
+        tracks.append(provider._parse_track("test", chapter, idx, album))
```
OzGav: This builds tracks and then throws it away, so it only fails if something raises. Please assert the album name, artist, cover and copyright, and the tracks' ids, track numbers and durations. Worth noting the fixture sets duration on every chapter, so the fallback that adds up the individual track durations is never reached. That fallback is the most interesting logic in the parser and is the thing most worth a test.

tests/providers/yoto/test_parse_card.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def _make_multitrack_chapter_card() -> dict:
```
OzGav: Please add a docstring

music_assistant/providers/yoto/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "music",
+  "domain": "yoto",
+  "name": "Yoto",
+  "description": "Yoto Music Provider for Music Assistant.",
```
OzGav: This needs to be more descriptive so the users gets a quick understanding what this provider does. Look at the other providers for inspiration

music_assistant/providers/yoto/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "music",
+  "domain": "yoto",
+  "name": "Yoto",
```
OzGav, line 4:
```suggestion
  "name": "Yoto",
  "stage": "beta",
```
You are missing the stage key

music_assistant/providers/yoto/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "description": "Yoto Music Provider for Music Assistant.",
+  "codeowners": [
+    "@pantsman0"
+  ],
+  "requirements": [],
+  "documentation": "https://music-assistant.io/music-providers/yoto/",
+  "icon": "icon.png"
```
OzGav: Delete the icon. You have supplied an icon.png but we actually need icon_monochrome as well added to this PR. Make sure the latter is all white.

OzGav (changes requested):
Thanks for the PR. Great to see more sources for the kids. 

The main thing to address is that this doesn't use the authorised developer login flow. Yoto has self-service client registration, a device code flow, and a family:library:view scope that covers exactly what this provider needs, and `cdnninja/yoto_api` already works that way. Using the Android app's client ID and an undocumented password grant instead looks unnecessary, so please explain the reasoning. I'm also concerned this runs against "You must not: Reverse engineer, decompile, or extract source code from the APIs or underlying services", since I can't find `/card/resolve/` in the published reference either.

In terms of the actual code please see the inline comments and also note that mypy is failing. Lastly, there is no `strings.json` so we will want that added so the provider can be translated.

So in summary we would like to get this into MA but we must do this in a manner which is acceptable to Yoto.

pantsman0:
Hi @OzGav, thanks so much for looking at my PR and giving comments. I don't really do much work in Python and this is my first contribution to MA, so your comments are very much appreciated.

For the API, I didn't really know there was a sanctioned developer portal. I am a penetration tester by trade, so I just did what I normally do to understand apps and sniffed the HTTPS traffic from the Android app. I was in a bit of a grey area before, as I had not signed up for the developer portal and was subject to those terms (they do not appear in any of the account creation or product purchase terms) - but now I have created an account and I am subject to those terms. I will have to rewrite the logic to use that Oauth flow.

For the User-Agent, I saw that some providers (e.g. YTM, Apple Music) are just using generic Mozilla User-Agent strings so I did that just so I minimized the differences between my implementation and the official traffic. I didn't want something to break because I removed the `X-Yoto-Appisalpha` header for instance.
I also used the password flow since that's the simplest flow that is implemented both by the website and the mobile apps. The reason that I used the android app's client ID is that the app uses the call to `/card/resolve/` which requires a granted scope that is included in the app's JWT but not the website's. As I didn't know about the developer portal, I didn't know about `card/{cardId}` working with the `family:library:view`scope - I have tested it locally now and it produces the expected response including trackUrl links, so I will rebuild the provider based around that functionality. Once it is done, I will also document how to create a developer portal account and application, and how to run through the Oauth flows to connect your Yoto application to MA.

As for the other deficiencies in the code, I will address them during the other changes.

music_assistant/providers/yoto/manifest.json
pantsman0: Sorry, this seems to be ambiguous. Do I need icon and icon_monochrome? or should I *only* have icon_monochrome?

OzGav: Both. They are used in different places.

pantsman0:
I have checked the tests and they are working as expected. 

I think the noted question about the chapter duration and track duration may need adjustment though.

We don't know if it will happen in the real world, but how should it be handled if the chapter's duration does not match the sum of the tracks? Currently the duration is set directly from the chapter's reported duration, and each track uses its own reported duration.

OzGav:
Please comment on and resolve each of the open conversations and another review will be done when you are ready,

music_assistant/providers/yoto/manifest.json
pantsman0: This has been changed in the new commits

music_assistant/providers/yoto/manifest.json
pantsman0: Both icons are now present

tests/providers/yoto/test_parse_card.py
pantsman0: The new data generation function has a docstring

tests/providers/yoto/test_parse_card.py
pantsman0: The new test function asserts on various data model properties.

music_assistant/providers/yoto/manifest.json
pantsman0: A New longer description has been provided

music_assistant/providers/yoto/__init__.py
pantsman0: This has now been outsourced to the yoto_api package, which does not implement locking around automatic reauthentication. Should I submit a PR there before we merge here?

music_assistant/providers/yoto/__init__.py
pantsman0: This has been changed to InvalidProviderID

music_assistant/providers/yoto/__init__.py
pantsman0: It has been removed.

music_assistant/providers/yoto/__init__.py
pantsman0: This has been updated with a `get_artist` implementation

music_assistant/providers/yoto/__init__.py
pantsman0: This is still potentially an issue. Should I be wrapping all calls into the yoto_api client to catch and map exceptions?

OzGav: Yes please, and it should be a small job because the library has already done most of it. `yoto_api` catches `aiohttp `errors itself and re-raises its own types, so you only ever see `YotoError` or one of its two subclasses. `AuthenticationError` comes back on a 401, and `YotoAPIError` on anything else non-2xx, on a connection failure, and on bad JSON. `YotoAPIError` carries a `status_code` attribute so you do not have to read the message text.

So catch `YotoError` at each of the four places that call the client and map it: 403 to `LoginFailed`, 404 to `MediaNotFoundError`, 429 to `RateLimited`, anything else to `ResourceTemporarilyUnavailable`. `AuthenticationError` maps to `LoginFailed`. A single small private helper called from all four spots will do it.

One gap to cover as well. The library only catches aiohttp's own error type, but when its request timeout expires the error raised is a plain `TimeoutError`, which is a different type and slips past. So catch `YotoError` and `TimeoutError` together.

music_assistant/providers/yoto/__init__.py
OzGav: Worth doing upstream, but we won't hold this PR for it. A lock on your side cannot help because the refresh happens inside the library, so this is genuinely theirs to fix.

It is a real risk though. The library refreshes a full hour before the token expires and does so at the start of every call, so the window is wide. A library sync running while someone starts playback will have both calls decide the token is stale and send the same refresh token. Your line 82 checks whether the refresh token came back changed, so we know rotation is switched on, and providers that rotate will often reject a token that gets sent twice. Worst case the user has to run setup again.

There is a related problem that is ours rather than theirs, and I think it is the more likely one to actually bite people, so I have left a separate comment on it.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,351 @@
...
+        refresh_token = str(self.get_setup_value(CONF_REFRESH_TOKEN) or "")
+        if not client_id or not refresh_token:
+            raise LoginFailed("Missing Yoto credentials")
+
+        self.client = YotoClient(client_id=client_id, session=self.mass.http_session)
+        self.client.set_refresh_token(refresh_token)
+        try:
+            token = await self.client.check_and_refresh_token()
+            if token.refresh_token and token.refresh_token != refresh_token:
+                self._update_setup_data(CONF_REFRESH_TOKEN, token.refresh_token)
```
OzGav, line 93: This is the only place the refresh token gets saved, but it is not the only place it changes. Every later call refreshes the token inside the library when it is close to expiring, which updates it in memory and never writes it back here. After a restart we hand Yoto a token that has already been replaced, and since rotation is on, that will be rejected and the provider will not load until the user runs setup again.

The library has no callback for this, but `client.token` is public, so reading `self.client.token.refresh_token` after calls and saving it when it differs will cover it.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,351 @@
...
+        ):
+            return
+
+        for url in streamdetails.data.get("track_urls", []):
+            async with self.mass.http_session.get(
+                url, timeout=aiohttp.ClientTimeout(total=30)
+            ) as response:
+                response.raise_for_status()
+                async for chunk in response.content.iter_chunked(64 * 1024):
+                    yield chunk
```
OzGav: This can go now. `get_stream_details` always returns `StreamType.HTTP`, and the core only calls this method for `CUSTOM`, so it is unreachable. The first branch would return immediately anyway, and it reads `streamdetails.data["track_urls"]`, which nothing sets any more. Removing it also makes the `aiohttp` import on line 8 unused.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,351 @@
...
+
+    async def get_artist(self, prov_artist_id: str) -> Artist:
+        """
+        Get full artist details by id.
+
+        :param prov_artist_id: Provider's artist ID.
+        """
+        return Artist(
+            item_id=prov_artist_id,
+            provider=self.domain,
```
OzGav: Should be `self.instance_id`. Everything else in the file uses the instance id, and items coming straight from a provider are expected to carry it

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,351 @@
...
+        track_id = f"{card_id}:{chapter_key}"
+        chapter_title = chapter.title or f"Chapter {idx + 1}"
+
+        chapter_duration = chapter.duration
+        if not chapter_duration and chapter.tracks:
+            chapter_duration = sum(
+                t.duration for t in chapter.tracks.values() if isinstance(t.duration, (int, float))
+            )
+
+        chapter_cover = chapter.icon
```
OzGav: `chapter.icon` is the 16 by 16 pixel icon shown on the player screen, not artwork. Preferring it over the album cover means tracks get a tiny image blown up to fill a tile. Better to use the album cover, or leave track images off entirely so the album art shows through.

music_assistant/providers/yoto/strings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "config_entries": {
+    "intro": {
+      "label": "Create an application using the instruction in the provider page. Paste the Client ID for the application below."
+    },
+    "client_id": "Client ID"
```
OzGav: This needs to be an object with a label inside it, like intro above it, otherwise the generated key comes out as `provider.yoto.config_entries.client_id` where the UI wants` ...client_id.label`, so the field renders without a label.

Rather than adding it here though, "Client ID" is generic and soundcloud already has the identical string, so please lift it into the common strings and point both at it. Add to [music_assistant/strings.json](https://github.com/music-assistant/server/blob/dev/music_assistant/strings.json) under config_entries:

`"client_id": { "label": "Client ID" }`

then in both this file and [soundcloud/strings.json](https://github.com/music-assistant/server/blob/dev/soundcloud/strings.json):

`"client_id": { "label": "[%key:common::config_entries::client_id::label%]" }`

While you are in here, could you also add a manifest block with the description in it. That is what makes the provider description translatable, and right now it only lives in manifest.json.

OzGav (changes requested):
Looks really good now. This won't be merged before 2.10 ships next week but just fix these couple of things and then keep testing yourself to make sure it is working how you want.

Also I have written to Yoto to see if they are happy for us to just use a MA client ID. I will let you know what they come back with and if there are any limits that we will need to work to.

music_assistant/providers/yoto/__init__.py
pantsman0: I have changed this to `track_cover = album_cover if album_cover else chapter.icon`.

At least there is _something_ there if the album image fails for some reason.

music_assistant/providers/yoto/__init__.py
pantsman0: Fixed.

music_assistant/providers/yoto/__init__.py
pantsman0: Removed

music_assistant/providers/yoto/strings.json
pantsman0: Done.

music_assistant/providers/yoto/__init__.py
OzGav: To be honest I think you should just stick with None so the standard placeholder is used as a blown up icon will look terrible.

music_assistant/providers/yoto/__init__.py
pantsman0: This should now be handled in the _update_libarary and _update_card_detail wrapper functions.

music_assistant/providers/yoto/__init__.py
pantsman0: this should now be detected by the _update_library and _update_card_detail wrapper functions, which AFAIK are the only calls that can trigger a refresh token rotation.

music_assistant/providers/yoto/__init__.py
pantsman0: changed.

music_assistant/providers/soundcloud/strings.json
```diff
@@ -1,7 +1,7 @@
 {
   "config_entries": {
     "client_id": {
-      "label": "Client ID"
+      "label": { "label": "[%key:common::config_entries::client_id::label%]" }
```
OzGav: Too many labels here. Match what you have in the yoto strings.json

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,402 @@
...
+            audio_format=AudioFormat(content_type=content_type),
+            media_type=MediaType.TRACK,
+            stream_type=StreamType.HTTP,
+            duration=chapter.duration,
+            path=track_paths,
+            allow_seek=True,
+            can_seek=True,
+        )
+
+    async def _update_library(self) -> None:
```
OzGav: This block and the one in _update_card_detail below it are identical, and the copied messages still say "card library" when that one is fetching a single card. Could you pull the mapping into one helper so both call sites stay short. An async context manager around the call, or a function that takes the error and returns the right one to raise, would work.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,402 @@
...
+        """Update the library from the Yoto API and handle the yoto-api package errors."""
+        assert self.client.token
+        refresh_token = self.client.token.refresh_token
+        try:
+            await self.client.update_library()
+        except YotoAPIError as err:
+            if err.status_code is not None:
+                match err.status_code:
+                    case 403:
+                        raise PermissionError(
```
OzGav: `PermissionError` is a Python builtin rather than one of ours, so library sync will not catch it and a single 403 will still stop the whole sync, which is what we were trying to avoid. Please use `ProviderPermissionDenied`, or `LoginFailed` if you would rather send the user back to re-auth.

Same comment for line 262

OzGav (commented):
I was looking at the docs PR and noticed in the categories you had audiobooks and podcasts and I was going to say take this out as you are only using albums but then I thought I would look more into what Yoto actually provides. I see in their API that they do discriminate via categories and I wondered why you might have chosen to not steer the content into an appropriate bucket? But then I looked at the upstream library and saw they have some bugs there and so maybe that is why. The bug returns None for all categories so you wont have been able to use that metadata to date.  Specifically I noticed that:

Line 338 of its client.py reads card.metadata.stories instead of card.metadata.category, so the field is always None. 

The [Card Content Schema](https://yoto.dev/reference/card-content-schema/) gives category as one of `none`, `stories`, `music`, `radio`, `podcast`, `sfx`, `activities`, `alarms`. So stories to Audiobook, music and activities to Album, podcast to Podcast with its chapters as episodes, none to Album as the fallback, and radio, sfx and alarms out of scope for now.

The blocker is that upstream bug, since category is always `None` until the path is fixed. Once it populates, could you check how many of your cards come back as none, because that tells us whether this is worth doing properly or whether most cards fall through to the default anyway. If it does split usefully then I think it is worth doing before this ships rather than after, because changing media type later means migrating people's existing libraries. If most cards come back as none then albums for everything will be a reasonable thing to do and we leave it as is with a comment in the code and PR explaining the rationale.

The assignment of values to `series_title` and `series_order` in the upstream library also look suspicious but I would raise those upstream as a question rather than a bug. The library's own comments suggest they exist, but unlike category they are not in the published schema, so I cannot be certain the paths are wrong.

music_assistant/providers/yoto/__init__.py
pantsman0: Is there some way to signal to Music assistant that the card has been removed without waiting for a library sync?

If someone scans one of the physical cards into a different card, they will be immediately removed from the first account and sent into the library of the new account.
For now I have changed it to a temporary resource error since it could theoretically come back, but that will leave a zombie Album in the library until the next sync.

music_assistant/providers/soundcloud/strings.json
pantsman0: Fixed

music_assistant/providers/yoto/__init__.py
pantsman0: I have changed it to a wrapper function that takes an `Awaitable[None]`. Originally I split it because I would have to take a function that has no arguments, and a function that takes one; this has resolved that issue.

music_assistant/providers/yoto/__init__.py
OzGav: You should not need to signal anything. Returning `MediaNotFoundError` already does it, and it takes effect at lookup time rather than at the next sync. The core catches it, falls back to the stored library item and marks it unavailable, deliberately keeping the record so the matching logic can re-link the card if it comes back under a different id.

It is also worth knowing what a sync actually does here, because I think the expectation is off. Yoto is a streaming provider, so the deletion path does not delete the row. It sets `in_library` to `false`. The card drops out of the library listing but the record is kept, so the metadata survives and can be re-linked later. So the item you are describing as a zombie is not something a sync would clean up either, and that is intentional.

I would not treat a 403 as deletion though. It is at least as likely to mean a missing scope or a revoked consent, and if we act on it by removing things then one auth problem takes the user's whole Yoto library out of the library view at once. It would be recoverable, since the rows are kept and a later good sync puts them back, but it is not something to trigger on an ambiguous status code. Please keep 403 as `ProviderPermissionDenied` and let 404 carry the not found meaning.

Do you know which one Yoto actually returns when a card is scanned into another family? If it is a 404 then you already have the behaviour you want for free.

pantsman0: I will test the functionality and make changes based on that. Requesting a valid card ID from an account that doesn't have access causes a 404 in the private android API to resolve cards, but I haven't tested with the public query API.

pantsman0: yes, once the card is in a different account. Querying the card just returns a 404.

pantsman0:
Looks like cdnninja already fixed it last week and it's out in v4.3.4.

I have updated and mapped the types as below:
Album: ["music", "activities", "sfx", "none", None]
Audiobook: "stories"
Podcast: "podcast"
Radio: "radio"
Unhandled: "alarm"

I have noticed however that this degrades the audiobook experience - the `Audiobook` model has no chapters or tracks, so you can't easily go to different stories in the children's books. e.g. A Bear Called Paddington, Winnie the Pooh


Also I only have one "sfx" in my library, but it was ambience so I put it in with Albums since it was music-ish

OzGav:
You can fix the audiobook chapters by populating metadata.chapters. The Audiobook model does support them, it just does not derive them for you.

The core already has the navigation wired up. get_audiobook_resume_point in media_resolver.py takes a chapter, finds it in audio_book.metadata.chapters by position, and seeks to chapter.start. With the list empty it raises InvalidDataError, which is why jumping to a story on a Paddington or Winnie the Pooh card does nothing at the moment. Fill the list and that path starts working with no other changes.

You already do the hard part. The loop in get_stream_details at lines 195 to 202 walks every chapter and sums the durations. The same walk gives you the markers: keep a running offset, and for each chapter emit a MediaItemChapter with position as the 1-based index, name as the chapter title, start as the offset so far and end as the offset plus that chapter's duration. See [audiobookshelf/parsers.py:273-281](https://github.com/music-assistant/server/blob/dev/audiobookshelf/parsers.py#L273-L281) for the shape.

While you are there, please also set duration on the Audiobook. It is 0 today, so the UI has no length to draw the progress bar or remaining time against, and the total you already compute is the value you need.

This works with the streaming side rather than against it. You are passing the tracks as a MultiPartPath list with per-part durations, so MA can already seek within the concatenated stream. The chapter offsets are just the labels that let someone aim at a point it can already reach.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+        """
+        card: YotoCard | None = None
+        if prov_album_id in self.client.library:
+            card = self.client.library[prov_album_id]
+        else:
+            await self._handle_yoto_api_call(self.client.update_card_detail(prov_album_id))
+            card = self.client.library.get(prov_album_id)
+        if not card:
+            raise MediaNotFoundError(f"Card {prov_album_id} not found")
+        return self._parse_album(card)
```
OzGav: These lines are duplicated four times in the code. Lets extract them and go with a single `_get_card(card_id)` helper so each of `get_album`, `get_audiobook`, `get_podcast` and `get_radio` becomes a fetch plus a parse.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+                            MultiPartPath(path=track.trackUrl, duration=track.duration)
+                        )
+                        if track.duration:
+                            total_duration += track.duration
+
+            if not track_paths:
+                raise MediaNotFoundError(f"No audio URLs found for card {card_id}")
+
+            # Use format from first track
+            first_track = next(iter(next(iter(card.chapters.values())).tracks.values()), None)
```
OzGav: `next(iter(next(iter(...))))` is hard to follow. It is safe today because the empty case is caught by the `track_paths` check above, but the inner next has no default so it would raise `StopIteration` if that guard ever moved. Worth unpicking into a named local.

OzGav (commented):
Do your Make Your Own cards actually come through in the sync? `update_library()` only hits `/card/family/library`, and there is a separate `/content/mine` endpoint for MYO content that nothing here calls. If they do come through then the `user:content:view` scope in the authorize request looks unnecessary and we should drop it, since we ask for as little as we need. If they do not, then the manifest description is promising MYO support we are not delivering yet.

The PR description and docs also need updating with these recent changes.

Your call as to whether pulling in sfx into the library is value adding. You could skip it if you wanted to but totally up to you.

pantsman0:
Thanks again for all the help getting this up to snuff. I'll get onto those changes tomorrow.

OzGav:
All good. They have given you a good API to work with so this is going to work really well in the end i think.

pantsman0:
I've done a bit of a rework to get everything working right, but it still needs proper testing. I am going away for the weekend so that will have to be early next week.

as for the MYO, I think you're right. We'll have to take that out of the docs and manifest.

m4rkireland:
Hi @pantsman0 - I've been independently working on the Yoto implementation as well and only noticed this submission today 😅 (Go coding Dads! 😂) 

I've linked it here as a draft for your info https://github.com/music-assistant/server/pull/5903. I tried to get a AI summary of the two approaches and put that as the description so we can easily see differences and see if anything is worth merging into yours.

I honestly really don't mind which PR get's merged, but I thought you might be interested in this anyway.

Cheers!

OzGav:
@m4rkireland appreciate your message. This PR is almost ready so we will continue here. You should consider a follow up PR with any functional improvements after this is merged.

m4rkireland:
Sure no worries at all @OzGav - I've closed mine, but it can still be referenced if required.

Cheers

music_assistant/providers/yoto/__init__.py
pantsman0: This is  now removed.

music_assistant/providers/yoto/__init__.py
pantsman0: This has been implemented.

pantsman0:
@OzGav  - I don't know what is going on, but for some reason the podcast parsing was failing when using cached copies of the card from the library, so I've removed the cache calling logic from that code path.

Everything else is working for me.

You're right about the MYO too, I'm taking it out.

OzGav:
I am not sure if the cache is broken. It is a two-stage fetch and the first stage does not include chapters. So the fix is not to remove the check, it is to make the check ask the right question:

```
card = self.client.library.get(card_id)
if card and card.chapters:
    return card
```

OzGav:
Could you double check playing an audiobook on the current head. From what I can see `_parse_audiobook` gives the mapping a bare card id with no colon, and `get_stream_details` now rejects those at line 181 since the whole-card branch went in the media type rework. `media_resolver` hands the audiobook straight through without rewriting the id, and the error gets caught and logged as a warning rather than surfacing, so it may not have been obvious. If audiobooks do play then I have misread it and I would like to know what I have missed.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+    async def get_library_albums(self) -> AsyncGenerator[Album]:
+        """Retrieve library albums from the provider."""
+        await self._handle_yoto_api_call(self.client.update_library())
+        for card in filter(
+            lambda card: card.category in ["music", "activities", "sfx", "none", None],
+            self.client.library.values(),
+        ):
+            yield self._parse_album(card)
+
+    @use_cache(3600 * 7)
```
OzGav: Please review the use of the @use_cache decorators as they seem a bit inconsistent. Seven hours on get_album, seven days on get_audiobook, nothing on get_podcast or get_radio?

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+            yield self._parse_podcast(card)
+
+    async def get_podcast(self, prov_podcast_id: str) -> Podcast:
+        """Get full podcast details by id."""
+        return self._parse_podcast(await self._get_card(prov_podcast_id))
+
+    async def get_podcast_episodes(self, prov_podcast_id: str) -> AsyncGenerator[PodcastEpisode]:
+        """Get all PodcastEpisodes for given podcast id."""
+        card = await self._get_card(prov_podcast_id)
+        podcast = self._parse_podcast(await self._get_card(prov_podcast_id))
```
OzGav: `get_card` called twice in a row?

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+
+    def _parse_audiobook(self, card: YotoCard) -> Audiobook:
+        """Parse Yoto card into a Music Assistant Audiobook."""
+        card_id = card.id
+        title = card.title or f"Unknown Audiobook ({card_id})"
+
+        chapters: list[MediaItemChapter] = []
+        total_duration = 0
+        for idx, chapter in enumerate(card.chapters.values(), start=1):
+            track = next(iter(chapter.tracks.values()))
```
OzGav: There is no default here, so a chapter with no tracks raises `StopIteration` rather than hitting the warning branch below. The `if` track test after it can never be `false`. This needs the None default.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+        """Retrieve library radio stations from the provider."""
+        await self._handle_yoto_api_call(self.client.update_library())
+        for card in filter(lambda card: card.category == "radio", self.client.library.values()):
+            await self._handle_yoto_api_call(self.client.update_card_detail(card.id))
+            async for radio in self._parse_radio_stations(card):
+                yield radio
+
+    async def get_radio(self, prov_radio_id: str) -> Radio:
+        """Get full radio details by id."""
+        card_id, _station_id = prov_radio_id.split(":", 1)
```
OzGav, line 308: You should check ":" is there, so a malformed id raises `ValueError` instead of `InvalidProviderID` like the other methods.

OzGav (commented):
We are standardising podcast episode positions so that the oldest episode is always the lowest number. Your position at line 561 already looks good, 1 upwards with no duplicates, so the only thing to check is the direction. The numbers follow whatever order Yoto returns the chapters in at line 254, so could you look at a real podcast card and confirm the first chapter is the oldest episode. If Yoto lists them newest first they will need reversing.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,630 @@
...
+                return station
+        raise MediaNotFoundError(f"Radio station {prov_radio_id} not found")
+
+    async def _get_card(self, card_id: str) -> YotoCard:
+        """Get a Yoto card from the provider."""
+        if len(self.client.library) == 0:
+            await self._handle_yoto_api_call(self.client.update_library())
+
+        # if card_id in self.client.library:
+        #    return self.client.library[card_id]
```
OzGav: If this remains unneeded then it needs to be removed.

music_assistant/providers/yoto/__init__.py
pantsman0: fixed. I needed to pull the _get_card out but forgot to remove the second call.

music_assistant/providers/yoto/__init__.py
pantsman0: Yeah, I think I need to really dig into what is going on here. There's nothing really special about podcasts, so I need to figure out why removing that fixed the Story Shed podcast but having it there before didn't break all the other media types.

pantsman0: Just to make sure this is covered off, I have reintroduced the check, but also put in a check that the chapters list is populated (i.e. second stage loading has occurred). If this fails, it updates the card.

music_assistant/providers/yoto/__init__.py
pantsman0: this is being properly guarded now.

music_assistant/providers/yoto/__init__.py
pantsman0: I have just removed these for now. `get_album`, `get_audiobook` and `get_radio` should be very stable, but `get_podcast` can have rare-to-frequent updates - but probably easier to just ignore caching for now, and potentially make it user-configurable later if that's possible with decorators.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,669 @@
...
+
+            if not track_paths:
+                raise MediaNotFoundError(f"No audio URLs found for card {card_id}")
+
+            # Use format from first track
+            first_chapter = next(iter(card.chapters.values()), None)
+            assert first_chapter  # We know there are chapters due to the track enumeration above
+            first_track = next(iter(first_chapter.tracks.values()), None)
+            assert first_track  # We know there are tracks due to the track enumeration above
+            format_str = first_track.format if first_track else None
```
OzGav: The reason given for the second assert does not hold. The if not `track_paths` guard proves some chapter had a track with a `trackUrl`, not that the first chapter did, so the comment is misleading even in the cases where it never fires.

I noticed `_parse_audiobook` already handles a chapter with no tracks and logs a warning for it, so the file treats that as possible in one place and impossible here.

Rather than adding a check, this can just get smaller. You are already walking every track to build `track_paths`, so take the format from the first one you collect there. That removes both asserts and line 202, which is currently dead anyway since the assert has already guaranteed `first_track`

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,669 @@
...
+
+    async def get_podcast(self, prov_podcast_id: str) -> Podcast:
+        """Get full podcast details by id."""
+        return self._parse_podcast(await self._get_card(prov_podcast_id))
+
+    async def get_podcast_episodes(self, prov_podcast_id: str) -> AsyncGenerator[PodcastEpisode]:
+        """Get all PodcastEpisodes for given podcast id."""
+        card = await self._get_card(prov_podcast_id)
+        podcast = self._parse_podcast(card)
+        for idx, episode in enumerate(reversed(card.chapters.values()), start=1):
```
OzGav, line 282: Could you add a one line comment saying why. reversed is correct only if Yoto returns chapters newest first, and that fact is not recorded anywhere, so the next person to read this cannot tell whether it is deliberate or a bug. Something like "Yoto lists podcast chapters newest first, so reverse them to make position 1 the oldest". (I assume you found this to be the case)

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,669 @@
...
+    async def get_audiobook(self, prov_audiobook_id: str) -> Audiobook:
+        """Get full audiobook details by id."""
+        card: YotoCard | None = None
+        if prov_audiobook_id in self.client.library:
+            card = self.client.library[prov_audiobook_id]
+        else:
+            await self._handle_yoto_api_call(self.client.update_card_detail(prov_audiobook_id))
+            card = self.client.library.get(prov_audiobook_id)
+        if not card:
+            raise MediaNotFoundError(f"Card {prov_audiobook_id} not found")
```
OzGav: This is the last call site still doing the library lookup by hand, and it is the one that matters most, since it returns a chapterless audiobook when only a library sync has run. `_get_card` now has the chapters guard, so this can become `return self._parse_audiobook(await self._get_card(prov_audiobook_id))` like `get_album` and `get_podcast`.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,669 @@
...
+            can_seek=True,
+        )
+
+    async def get_library_audiobooks(self) -> AsyncGenerator[Audiobook]:
+        """Retrieve library audiobooks from the provider."""
+        await self._handle_yoto_api_call(self.client.update_library())
+        for card in filter(lambda card: card.category == "stories", self.client.library.values()):
+            # we update audiobooks explicitly as we need chapters preloaded
+            await self._handle_yoto_api_call(self.client.update_card_detail(card.id))
+            yield self._parse_audiobook(await self._get_card(card.id))
```
OzGav: `_get_card` on the next line now handles this itself, so the explicit `update_card_detail` and its comment are redundant.

music_assistant/providers/yoto/__init__.py
```diff
@@ -0,0 +1,669 @@
...
+        await self._handle_yoto_api_call(self.client.update_library())
+        for card in filter(lambda card: card.category == "radio", self.client.library.values()):
+            await self._handle_yoto_api_call(self.client.update_card_detail(card.id))
+            async for radio in self._parse_radio_stations(card):
+                yield radio
+
+    async def get_radio(self, prov_radio_id: str) -> Radio:
+        """Get full radio details by id."""
+        if ":" not in prov_radio_id:
+            raise InvalidProviderID(f"Invalid track ID format: {prov_radio_id}")
```
OzGav:
```suggestion
            raise InvalidProviderID(f"Invalid radio ID format: {prov_radio_id}")
```

OzGav (commented):
Thanks. A few comments based on the most recent commits

music_assistant/providers/yoto/__init__.py
pantsman0: removed.

music_assistant/providers/yoto/__init__.py
pantsman0: Yes, I did check that this is the case. I have added a justifying comment.

music_assistant/providers/yoto/__init__.py
pantsman0: This has been updated in line with your suggestion.

music_assistant/providers/yoto/__init__.py
pantsman0: I have moved this around so that the first track that is found with a non-Null format will set the format for the stream, and kept the AAC fallback.

pantsman0:
I've also just tested all the changes locally on my machine.

I did notice that the player UI for audiobooks has changed.
When I tested before your dev branch rebase, playing an audiobook had the whole book show up in the scrubbing bar in the player. This now just has the bar show the current chapter - I'm assuming this is intended, but I wanted to call it out in case it was something I did.

OzGav:
No that is new behaviour. If you look at the queue in the now playing view you should see all the chapters there?

music_assistant/providers/yoto/setup_flow.py
```diff
@@ -0,0 +1,120 @@
...
+
+    :param session: The setup session driving the flow.
+    :param client_id: The Yoto developer client id.
+    """
+    code_verifier, code_challenge = pkce.generate_pkce_pair()
+    redirect_uri, state = hosted_bounce_redirect(session.callback_url)
+    params = {
+        "response_type": "code",
+        "client_id": client_id,
+        "scope": "family:library:view user:content:view offline_access",
```
OzGav:
```suggestion
        "scope": "family:library:view offline_access",
```
I think we can drop that since you removed all the MYO parts?

pantsman0:
> No that is new behaviour. If you look at the queue in the now playing view you should see all the chapters there?

Yes. Instead of just having the audiobook as one queue item and a the scrubber bar sectioned for each chapter, the audiobook is started up as a queue. You can't use the scrubber to go between chapters.

OzGav (approved):
Thanks for the hard work on this @pantsman0 . This looks great!

One for later, not for this PR. The Yoto library refreshes the login token whenever it is close to expiring, and it does that at the start of every call. If two things happen at once, say a library sync while someone starts playing something, both of them can try to refresh at the same time and both send the same token. Yoto gives out a new token each time and usually rejects one that gets sent twice, so in the worst case the user is logged out and has to set the provider up again.

It needs bad timing to happen and setting up again fixes it, so it is not a blocker. You cannot fix it here because the refresh happens inside the library. Worth raising with `cdnninja/yoto_api` when you get a chance, it just needs a lock around that refresh so only one runs at a time.

pantsman0:
It's been a pleasure @OzGav, thanks so much for the repeated reviews.

As for the refresh token, I have been thinking about that too. For now, I have put in a PR so we can provide a callback to the `YotoClient` that will notify the `YotoProvider` of a token update.

I was also thinking that it would be good to put in a PR that wraps token usage in the library with a mutex - but looking at the behaviour of the Yoto API, they just reject invalid requests without invalidating the new tokens. Hopefully this will just be rare failed requests that can be rerun with the updated token.
Wrapping the token in a mutex should just be done in the update function so there's no out of order updating that ends up with an invalid refresh token being cached.

OzGav:
All good. I'll leave that with you. Just two question on the docs PR if you can have a look at that and then keep an eye out for the next beta release when this will be live!

pantsman0:
Actually, I don't think it needs fixing. If the refresh hits the path of raising an `aiohttp.ClientError` then that bubbles up to MA as a `ProviderPermissionDenied` which shouldn't put the provider into an error state. If it hits the `AuthenticationError` path, then that bubbles up to MA as a `ResourceTemporarilyUnavailable` so that definitely won't put the provider into an error state.

OzGav:
Your mapping is right on the `AuthenticationError` path, though `aiohttp.ClientError` actually comes out as `ResourceTemporarilyUnavailable` too rather than `ProviderPermissionDenied`, since the library raises `YotoAPIError` with no status code and only a real 403 reaches that branch.

The worry was not the provider dropping into an error state though. It is that the refresh token itself can be lost. I understand Yoto hands out a new refresh token each time, and providers that do that often treat a token sent twice as a sign it has been stolen and invalidate it, including the good one the first call just received. If that happens, the token we saved is lost.

Then you wouldn't see an error but at runtime every call fails, MA quietly retries every two minutes forever, and it won't be clear what the problem is. Then on the next restart `handle_async_init` turns it into a `LoginFailed`, which MA does not retry, so the provider stays down and the user has to set it up again.

All of that only matters if Yoto actually does the invalidate thing. If they just reject the second request and move on then you are completely right and there is nothing to fix. I do not know which it is, which is why this is not a blocker but will be worth remembering if we get reports saying "why am I gettting logged out all the time?".

pantsman0:
Yeah, you are absolutely right. I have tested and double submitting a refresh token will invalidate the new refresh token. I will try to craft something for a PR to the yoto_api package.

kieranhogg:
Excited to play around with this!

Just noting two bits here in terms of setting up a new provider from fresh, I appreciate this isn't support, just passing on in case it's not just me being daft:
~~1. The new UI for setup flow makes it really hard to find the instructions~~
Screenshot 2026-09-01 at 19 56 13

~~You can't see the provider page until you've set the provider up.~~

_Realised the ? will direct to the actual documentation once it has been merged, ignore this one._

2. After callback I'm getting the following, I can't see anything helpful in the logs, so let me know if I can do anything further:
image

pantsman0:
Hi Kieran, 

Did you set up for app I'm the way documented by this page? https://deploy-preview-946--ohf-music-assistant.netlify.app/music-providers/yoto/#configuration

That's the preview page built from the pull request. 

I would expect to see an error message I'm the logs, but can you try setting up the oauth app in the yoto dashboard according to that page and let me know if it works?

kieranhogg:
Yeah, I found that in the beta docs repo and followed it from there. I didn't have an option for offline in my version, is that related?

pantsman0:
offline_access is at the very bottom of the list.

image

kieranhogg:
That's embarrassing - I didn't spot it scrolled! All good adding it now offline_access is enabled

PR #5595, opened 2026-08-11
Fix sync group not re-forming after a leader switch

marcelveldt:

After a sync group switched to a new leader, the group could end up broken: Music Assistant would report one speaker as the leader while the audio session had actually moved to a different one. From that point the group could never be dissolved or re-formed, speakers kept repeating "can not be grouped with", and only a restart of Music Assistant fixed it.

This happened because the sync group picked the new leader from its own member order, while the speaker protocol hands the session to its own first remaining member. Those two orders can drift apart over time, and once they disagree the group and the actual session were out of sync for good.

- Keep the group's member order aligned with the live session, so the new leader is the speaker that actually takes over the session
- Dissolve the group from the speaker that really holds the members, instead of a leader that no longer holds any (which silently did nothing before)
- Stop a speaker that the protocol promoted on its own, so it doesn't keep playing after the group is dissolved
- Added tests covering the leader switch and both dissolve paths

- reported on 2.10.0b14 with 5 Sendspin speakers in one sync group with dynamic members enabled

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sync_group/player.py
+++ b/music_assistant/providers/sync_group/player.py
@@ -758,23 +758,31 @@ class SyncGroupPlayer(Player):
         self._cancel_reform_timer()
         self._playback_start_at = float("-inf")
         if sync_leader := self.sync_leader:
-            # dissolve the temporary syncgroup from the sync leader
+            # dissolve the temporary syncgroup from the player that holds the members:
+            # ungrouping from a leader that no longer holds them is a no-op and would
+            # leave the members grouped and streaming with no way back
+            group_leader = self._protocol_group_leader(sync_leader)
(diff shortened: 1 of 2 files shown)


marcelveldt:
Re the suppressed comment about `_dissolve_syncgroup` releasing members that aren't ours:

`_protocol_group_leader` only adopts a member whose group overlaps this syncgroup's members, so we never target a group we didn't form. And this matches what already happened for the tracked leader - `set_members` deliberately releases members that were grouped to the leader outside of MA.

Scoping the removal wouldn't buy anything either: we stop the promoted leader right after, so anything still in its group loses audio regardless. Leaving as is.

PR #5603, opened 2026-08-12
SiriusXM Library Swap

MizterB:

Moves the SiriusXM provider from `sxm==0.2.8` (unmaintained since 2021) to
[aiosxm](https://github.com/MizterB/aiosxm), which targets the edge-gateway API
the SiriusXM web player uses.

Adds artist stations and Xtra channels, which the provider did not support
before. Both are runs of discrete tracks rather than broadcasts, so they are
modelled as dynamic playlists and get per-track seek and skip; live channels
stay radio. Also adds library sync for saved stations, catalogue search,
browse by library/channel/genre, and a stream quality option.

- https://github.com/music-assistant/support/issues/5502

## Types of changes

## Checklist


music_assistant/providers/siriusxm/parsers.py
```diff
@@ -0,0 +1,315 @@
...
+                    provider=instance_id,
+                    name=track.artist,
+                )
+            ]
+        )
+    # No album mapping: SiriusXM has no album entity to resolve, and MA tries to
+    # fetch any mapping it is given. The name still reaches the UI via the
+    # track's version field.
+    if track.album:
+        mass_track.version = track.album
```
OzGav: Putting the album name into version looks like it'll cause trouble downstream. That field means things like "Remastered" or "Live", and MA leans on it in two places, `compare_version` treats a track that has a version and one that doesn't as different songs, so your SiriusXM tracks would never link up with the same song from Spotify, and the scrobbler appends it to the title, so Last.fm sees "Volare (That's Amore)". Neither shows up as an error, which is what makes it hard to catch later.

Your instinct about the album mapping is right though, and worse than the comment suggests as `get_full_track_details` resolves an album mapping by calling `get_album` on the provider, and the base class raises `NotImplementedError`, which isn't caught by the guard around that call. So that would actually crash, not just warn.

But this is the same problem you already solved for artists. `get_artist` returns a name-only stub for exactly this reason. Could we do the same for albums? The mapping stays lightweight and the stub costs no extra request, since it's built from the name you already have. Apple Music and Bandcamp both do name-as-id mappings this way. The only real cost is a thin album row if a track reaches the library, which is already true for artists.

And if you'd rather not have album rows at all, I'd still rather drop the album than carry it in version.

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+        :param media_type: The media type (only radio is supported).
+        """
+        if parsed := split_queue_item_id(prov_item_id):
+            await self.client.remove_from_library(*parsed)
+            return True
+        if media_type != MediaType.RADIO:
+            return False
+        channel = await self.get_channel(prov_item_id)
+        await self.client.remove_from_library(channel.type, channel.id)
+        return True
```
OzGav: These handle artist stations and Xtra channels, but the provider only advertises radio editing, and MA checks that list before it calls you so this branch can't actually run today. Was LIBRARY_PLAYLISTS_EDIT meant to be in SUPPORTED_FEATURES? These two methods are the right place for it, so that's the only change needed.

While you're in there could these branch on `item.media_type` rather than on whether the id splits? Deezer's version checks the media type first and raises `UnsupportedFeaturedException` for anything else. Working out the kind of item from the shape of its id means the caller's declared type gets ignored, which feels like it could bite later. The docstrings need a touch-up too, they still say only radio is supported.

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+        """
+        if track := self._tracks.get(item_id):
+            return track
+        parsed = split_track_item_id(item_id)
+        if parsed is None:
+            return None
+        entity_type, entity_id, _ = parsed
+        try:
+            stream = await self.client.get_stream(entity_type, entity_id)
+        except Exception as err:
```
OzGav: Anywhere you see this in the code you need to narrow it down. If for some reason that isn't appropriate then put a comment in explaining why.

music_assistant/providers/siriusxm/constants.py
```diff
@@ -0,0 +1,65 @@
...
+STREAM_EXPIRATION = 300
+
+# How often MA polls us for fresh now-playing data during playback.
+STREAM_METADATA_UPDATE_INTERVAL = 15
+
+# -- Caching --
+
+# The channel catalog costs ~24 sequential requests to walk, and changes
+# rarely; aiosxm caches it internally, this bounds our own lookups too.
+CACHE_TTL_CHANNELS = 3600 * 6
```
OzGav: This is unused

music_assistant/providers/siriusxm/constants.py
```diff
@@ -0,0 +1,65 @@
+"""Constants for the SiriusXM provider."""
+
+# -- Config keys --
+# Kept as-is from the previous implementation so existing setups keep working.
+
+CONF_SXM_USERNAME = "sxm_email_address"
+CONF_SXM_PASSWORD = "sxm_password"
+CONF_SXM_REGION = "sxm_region"
```
OzGav: This still appears in a couple of spots but seems not to be used

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+        return await self.client.get_genres()
+
+    async def get_library_radios(self) -> AsyncGenerator[Radio]:
+        """
+        Retrieve the live channels saved to the SiriusXM account.
+
+        Only linear channels are radio; artist stations and Xtra channels are
+        runs of discrete tracks and surface as dynamic playlists instead.
+        """
+        for channel in await self.client.get_library_channels():
```
OzGav: `get_library_channels()` is called once for radios and again for playlists, and browsing "My Library" runs both, so that's two identical requests for one screen. Could one fetch feed both?

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+    async def _search_stations(
+        self, search_query: str, talent: list[Talent], limit: int
+    ) -> list[ArtistStation]:
+        """
+        Find the artist stations a query implies.
+
+        A direct station search only matches a station's own name, so a term
+        like "holiday" finds almost nothing. SiriusXM answers the same query
+        with the artists it matched, and each of those has a station, so those
+        are resolved too — that is how a themed search finds anything.
```
OzGav: General comment through the code. Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

This is good info though it just needs to be appropriately moved inline.

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+        playlist_id = queue_item_id(entity_type, entity_id)
+        upcoming = self._upcoming.get(playlist_id)
+        if not upcoming:
+            return
+        for index, track in enumerate(upcoming):
+            if track.id == track_id:
+                del upcoming[: index + 1]
+                return
+
+    async def _search_stations(
```
OzGav: search_all already returns `artist_stations` in the same response. `SearchResults` has the field, filled from the same request you make at the top of search.` _search_stations` discards it and starts over with `search_artist_stations`, which re-runs the identical search and then walks the talent containers, three requests per name. At the default limit that's roughly nineteen requests for one search, on top of one that already had the stations.

Could this just read `results.artist_stations`? If that turns out too thin in practice, the fan-out makes sense as a fallback when the direct results come back empty, rather than as the default path and bounded, since `len(stations) < limit` doesn't currently stop the gather from firing all of them.

music_assistant/providers/siriusxm/browse.py
```diff
@@ -0,0 +1,190 @@
...
+        self.domain = provider.domain
+        self.logger = provider.logger
+        # Genre names go into the browse path, so map a URL-safe form back to
+        # the name SiriusXM knows.
+        self._genre_slugs: dict[str, str] = {}
+
+    async def browse(
+        self,
+        path: str,
+        base_browse: Callable[
```
OzGav: `base_browse` is passed in from the provider but never called, it looks like it came across from the Deezer browse manager, where it is used, along with the type signature? Since the root listing deliberately replaces the base folders with the combined "My Library" view, the base browse isn't needed here, so it would seem the parameter and its docstring line can go, and `browse()` can drop the argument at the call site.

music_assistant/providers/siriusxm/parsers.py
```diff
@@ -0,0 +1,315 @@
...
+        title=now_playing.title,
+        artist=now_playing.artist,
+        # SiriusXM reports the show (e.g. "The Weekend Countdown") rather than
+        # an album; it is the more useful third line for a radio stream.
+        album=now_playing.show,
+        image_url=now_playing.image_url(*THUMB_SIZE) or fallback_image,
+    )
+
+
+def search_result_media_types(media_types: list[MediaType]) -> bool:
```
OzGav: I can't see this used anywhere?

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,520 @@
...
+    browse_manager: SiriusXMBrowseManager
+    streaming_manager: SiriusXMStreamingManager
+    channels_by_id: dict[str, Channel]
+    # Resolved queue tracks by MA item id. A SiriusXM track is only addressable
+    # through its parent's tune, so the url has to be kept from when the batch
+    # was built; its CDN signature lasts an hour.
+    _tracks: dict[str, SxmTrack]
+    # Tuned streams held per station, so the queue cursor advances between
+    # refills instead of restarting at the same three tracks.
+    _streams: dict[str, SxmStream]
```
OzGav: _upcoming gets cleaned up as tracks play, but _tracks and _streams never shrink. Three entries go into _tracks on every pull and nothing takes them out, so on a server that runs for weeks this just keeps climbing. Would it be worth expiring them on the same hour that the track URLs expire?

music_assistant/providers/siriusxm/streaming.py
```diff
@@ -0,0 +1,157 @@
...
+            return
+        if now_playing is None:
+            return
+
+        fallback_image: str | None = None
+        if channel := self.provider.channels_by_id.get(streamdetails.item_id):
+            fallback_image = channel.image_url("tile", "1x1", THUMB_SIZE)
+
+        if metadata := parse_stream_metadata(now_playing, fallback_image):
+            streamdetails.stream_metadata = metadata
```
OzGav: When an ad is playing this returns nothing, but the `if` means the old value stays put, so the player keeps showing the previous song's title and artwork for the whole break. The comment in `parsers.py` says the intent was for the station name to show instead so should this be clearing the metadata rather than leaving it?

OzGav (dismissed):
A lot of work done here. Thanks. I have a number of questions for you to have a look at. Thanks for giving this a lot of love!

MarvinSchenkel:

music_assistant/providers/siriusxm/parsers.py
MizterB: Dropped the album rather than mapping it. Nothing in SiriusXM is album-centric — it's all live or dynamic, track and artist driven — and a queued track carries an album name with no entity behind it, so a stub would just be a library row you can't browse to or play from.

music_assistant/providers/siriusxm/provider.py
MizterB: Yes, that was meant to be there — I want stations and Xtra channels savable back to the SXM account, same as channels. Added `LIBRARY_PLAYLISTS_EDIT`.

Both methods now branch on `media_type` and raise `UnsupportedFeaturedException` for anything else, like Deezer does. Docstrings updated.

music_assistant/providers/siriusxm/provider.py
MizterB: Narrowed these to `SxmError`, which is the base aiosxm raises for everything including wrapped network errors.

Left one broad, in `update_stream_metadata`: it runs as a background task for the whole of playback and nothing there is worth surfacing as an unhandled task error. Comment explains it.

music_assistant/providers/siriusxm/constants.py
MizterB: Right, that's left over. aiosxm caches the channel catalog itself so I never needed a TTL on my side. Removed.

music_assistant/providers/siriusxm/constants.py
MizterB: Dead too. The old `sxm` library took a region; aiosxm doesn't need one, so nothing reads it anymore. I kept the config keys as-is so existing setups wouldn't have to be re-entered, and this one just came along for the ride. Removed the constant and the re-export.

music_assistant/providers/siriusxm/provider.py
MizterB: Fixed. The library listing now fetches once in `_browse_library` and splits by type, rather than driving both generators.

The generators still fetch independently — MA syncs radios and playlists separately, so that part is intentional. Worth noting the channel catalog was never fetched twice either way, aiosxm caches that internally. It was the library call that doubled up.

music_assistant/providers/siriusxm/provider.py
MizterB: Done — moved the implementation detail out of the docstrings and into inline comments across the provider. The docstrings now say what the caller gets. The reasoning sits at the line it explains.

music_assistant/providers/siriusxm/provider.py
MizterB: Good catch. Seeding from `results.artist_stations` now, so searching an artist by name is a single request instead of nineteen.

Kept the fan-out as a fallback — I want the themed searches ("holiday", "80s") to keep working, and those turn up nothing by station name. It only runs when the direct results come up short, and it is a sequential loop now that stops once the limit is filled, rather than a gather that fired everything regardless.

music_assistant/providers/siriusxm/browse.py
MizterB: Removed, along with the argument at the call site.

music_assistant/providers/siriusxm/parsers.py
MizterB: Leftover. That check is inline at the top of `search` now. Removed.

music_assistant/providers/siriusxm/provider.py
MizterB: Fixed for `_tracks`. Entries carry the time they were cached now, and anything past the signed-url lifetime gets dropped on the next pull. I measured it before and after — 40 pulls took it to 118 entries and climbing, now it settles at 15.

Left `_streams` alone. That is one entry per station played rather than per track, and dropping one would restart that listener's queue from the top. Added a comment saying so.

music_assistant/providers/siriusxm/streaming.py
MizterB: Fixed. Clearing `stream_metadata` gets there on its own — MA falls back to the channel item, so the player shows the station name and logo.

Same bug one line up when `now_playing` came back empty. A failed request still keeps the last metadata, since that tells us nothing about what is playing.

MizterB:
@OzGav Thank you for all of the feedback as I tried to match all of the latest features and coding conventions.  I've responded to everything above, if you could take another look.  Thank you!

music_assistant/providers/siriusxm/constants.py
```diff
@@ -0,0 +1,58 @@
...
+QUEUE_BATCH_SIZE = 3
+QUEUE_LOW_WATER_MARK = 3
+
+# -- Browse paths --
+
+BROWSE_LIBRARY = "library"
+BROWSE_CHANNELS = "channels"
+BROWSE_XTRA = "xtra"
+BROWSE_GENRES = "genres"
+BROWSE_STATIONS = "stations"
```
OzGav: Sorry I missed this the one the first time. This is also unused.

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,554 @@
...
+            yield parse_radio(channel, self.instance_id, self.domain)
+
+    async def get_library_playlists(self) -> AsyncGenerator[Playlist]:
+        """Retrieve the artist stations and Xtra channels saved to the account."""
+        for channel in await self.get_library_channels():
+            if channel.type not in TRACK_QUEUE_TYPES:
+                continue
+            yield parse_xtra_playlist(channel, self.instance_id, self.domain)
+        for station in await self.client.get_library_artist_stations():
+            yield parse_station(station, self.instance_id, self.domain)
```
OzGav, line 224: The fix for browsing worked, one screen now makes one request instead of two. Syncing still makes three though. The radios method and the playlists method each ask SiriusXM for the saved library, and fetching the saved artist stations asks a third time inside aiosxm. They are small requests so it may not be worth chasing, but the comment about making the call once only holds for browsing, not for a sync. Worth a look, or worth saying so in the comment?

music_assistant/providers/siriusxm/browse.py
```diff
@@ -0,0 +1,190 @@
...
+            parse_xtra_playlist(channel, self.instance_id, self.domain)
+            if channel.type in TRACK_QUEUE_TYPES
+            else parse_radio(channel, self.instance_id, self.domain)
+            for channel in channels
+        ]
+        items.extend(
+            parse_station(station, self.instance_id, self.domain)
+            for station in await self.provider.client.get_library_artist_stations()
+        )
+        return sorted(items, key=lambda item: item.name.lower())
```
OzGav, line 116: In here, the saved channels come from the provider method but the saved stations go straight to the aiosxm client, skipping the provider. Any reason for the split? Going through the provider for both would keep it consistent, and it would mean anything the provider does on the way, like keeping its channel lookup up to date, applies to both.

music_assistant/providers/siriusxm/provider.py
```diff
@@ -0,0 +1,554 @@
...
+
+        upcoming = self._upcoming.setdefault(prov_playlist_id, [])
+        if len(upcoming) < QUEUE_LOW_WATER_MARK:
+            upcoming.extend(await self._pull_tracks(prov_playlist_id, entity_type, entity_id))
+        if not upcoming:
+            return []
+
+        # Hand out the head of the buffer without removing it. The same batch is
+        # returned until one of its tracks is actually streamed, so the listing
+        # a listener sees is the listing that plays; _drop_played then advances
```
OzGav:
```suggestion
        # a listener sees is the listing that plays; drop_played then advances
```
nit

OzGav:
This looks good to me. We won't merge this in before 2.10 ships so just a couple of minor things we can tighten while we wait.

music_assistant/providers/siriusxm/constants.py
MizterB: Removed. I only rechecked the ones you flagged last time, so I missed this too. Checked the rest of the file and it's the only one.

music_assistant/providers/siriusxm/browse.py
MizterB: No reason, just left over from when I moved the browse listing off the generators. Added `get_library_artist_stations` on the provider to match `get_library_channels`, and both browse and the playlist generator go through it now.

music_assistant/providers/siriusxm/provider.py
MizterB: Done.

music_assistant/providers/siriusxm/provider.py
MizterB: The comment was wrong. Browsing reads the library once, but a sync doesn't go through that path, so each generator reads it again. Updated it to say that.

Left the request count as-is. A sync bypasses the cache, so caching doesn't help, and deduping it means holding library state on the provider for three small requests. Added a note about the extra read inside aiosxm on the new method.

OzGav (approved):
OK let's test this in the beta! Thanks @MizterB !

PR #5607, opened 2026-08-12
Restore OpenSubsonic track artwork in playlist listings

MarvinSchenkel:

Tracks in an OpenSubsonic playlist listing showed a placeholder instead of their cover art.

Since #5359 dropped the per-track album fetch, a playlist track's album is a bare `ItemMapping` with no image, and `parse_track` never set an image on the track itself, so nothing was left to fall back on. OpenSubsonic already returns a `coverArt` id on every playlist entry, so the track now carries its own thumbnail, the same way `parse_album` does. No extra requests, so resolving a playlist stays a single `getPlaylist` call.

- Add a THUMB image from the song's `coverArt` in `parse_track`
- Add a regression test asserting playlist tracks carry that image and still trigger no per-track album or lyrics fetches

- https://github.com/music-assistant/support/issues/6054

## Types of changes

## Checklist

- [x] `pre-commit run --all-files` passes. (run on the changed files; the only mypy failures are pre-existing `sendspin` ones from a stale `aiosendspin` in my local env)
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable. (opensubsonic suite: 48 passed)


khers (approved):
This is the code that I had in place before we were resolving every song in a playlist fully. The problem with it is that implementations differ on if a Child (the object representing a playable item) needs to have the `coverArt` field filled in or if that should be delegated to the `AlbumId3` object instead. Which is why we were resolving the tracks. It looks like our major two implementations (Navidrome and Gonic) do populate the `coverArt` field in the Child object so we can see if it breaks for anyone else.

PR #5621, opened 2026-08-12
Send a spoken announcement by typing the text

marcelveldt:

The play announcement action could only be given a link to an audio file. Announcing spoken text meant letting Home Assistant's text-to-speech action render it first, which loses the fine-grained options the Music Assistant action offers, such as the pre-announce chime. There was also no way to make an announcement from the Music Assistant interface at all.

An announcement can now carry a text message instead of a link. Music Assistant speaks it with a text-to-speech engine and plays the result like any other announcement.

- Play announcement now takes either a link or a message to speak (one or the other), plus an optional engine to speak it with
- New "Announcement text-to-speech engine" setting under the player settings, which adopts the first available engine when it is not set
- New command for clients to list the available text-to-speech engines, so the interface can offer an announcement action only when one is set up
- Speaking a message to a group renders the audio once and hands the result to each member
- Moved the text-to-speech rendering helper out of AI Radio so both features share it
- Side effect of that sharing: the "no engine available" notice now sits next to the engine picker it explains, instead of under Generic (also affects the AI Radio, Music Quiz and Smart Playlist pickers)

- n/a

Companion frontend PR: https://github.com/music-assistant/frontend/pull/2458

## Types of changes

## Checklist


marcelveldt:
Re the suppressed comment about rejecting `LOCAL_FILE` TTS output:

Premise is right, the contract does allow an absolute local path. Leaving it as is for now though. Both in-tree engines (hass, openai_tts) return an http url, and the local-path branch in AI Radio is itself a latent fallback that nothing in tree hits - openai_tts renders to disk but deliberately serves it back over its own route because a bare path didn't work there either.

Making it work end to end means the group forward would have to carry a local path through the public command, which today only accepts http on purpose (it's what stops a caller making the server read arbitrary files). Reopening that would need a path containment check for a case no engine actually produces, so not worth it right now.

Did make the error say what's wrong instead of just failing vaguely. If a real engine ever needs it we'll pick it up from the issue.

marcelveldt:
On the schema version comment: it is bumped, 45 -> 46 in f13d9cb68, just in a later commit than the one that got reviewed.

On rejecting a local file path from a TTS engine: covered in my earlier comment. Short version, both in-tree engines return an http url and the local-file branch in AI Radio is itself latent, so making it work end to end would mean letting a local path through the public command purely for a case nothing produces. Left as is, error now says what's wrong.

marcelveldt:
Two suppressed ones:

Local/rtsp paths - same answer as above, not changing it.

Empty PlayerCommandFailed on a provider-side timeout: premise is right, a bare TimeoutError from inside an engine propagates and the decorator stringifies it to nothing. But with the 30s cap our own timeout now always wins - the only engine-side deadline in tree is openai_tts at 120s, and hass has none at all - so that path needs a third party plugin with a sub-30s deadline to be reachable at all. Leaving it; if one turns up we'll see it in a report.

OzGav:
Hi Marcel. Are you doing the HA or MA docs changes for this one?

PR #5635, opened 2026-08-13
Album art no longer holds up the app over a remote connection

marcelveldt:

Over a remote connection, album art and preview images are fetched through the WebRTC
connection. Until now they travelled on the same data channel as the API messages, so a
screen full of album art would queue ahead of API traffic on that one ordered stream —
and every API message had to wait for the image backlog to drain before it could be sent.
The result is a UI that feels sluggish exactly when it is loading a big list.

Images now get their own data channel, so they no longer sit in front of API messages.

- Route data channels by label to a handler, so a label can be served in-process instead
  of only being bridged to a local WebSocket
- Serve proxied HTTP requests (album art, previews) on a dedicated `http_proxy` channel
- Answer every proxy request on the channel it arrived on, so clients that still proxy
  over the API channel keep working unchanged
- Bump the API schema version so clients can tell whether the channel is available

- follow-up to #4889

## Types of changes

## Checklist

- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked: https://github.com/music-assistant/frontend/pull/2503

### DIFF
--- a/music_assistant/constants.py
+++ b/music_assistant/constants.py
@@ -48,7 +48,7 @@ PLAYLIST_MEDIA_TYPES: Final[tuple[MediaType, ...]] = (
 
 # API_SCHEMA_VERSION: bump this when adding new features to the API commands (and models)
 # or small non-breaking changes to existing commands
-API_SCHEMA_VERSION: Final[int] = 48
+API_SCHEMA_VERSION: Final[int] = 49
 
 # MIN_SCHEMA_VERSION is the minimum API schema version that the current server
 # version can work with. Only bump when there are breaking changes to existing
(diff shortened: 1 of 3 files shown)


marcelveldt:
On the suppressed comment about `except Exception` in `_serve_http_proxy`: leaving it as is.

The `await self._close_channel(...)` sits after the try/except, so a narrower catch would let an unexpected error skip teardown - the entry stays in `session.channels` and the next `http_proxy` channel gets refused as a duplicate. And `logger.exception` already logs the full traceback at ERROR, so defects stay visible rather than being swallowed. Same shape as the two bridge pumps right above it.

PR #5639, opened 2026-08-13
Document the AirPlay audio buffer depth setting

marcelveldt:

The `Audio buffer depth` player setting was missing from the AirPlay README's configuration options, and the neighbouring `sync_adjust` entry still said the buffer was handled automatically by the binary.

Older LinkPlay speakers need a deep receiver queue to play at all, and that same depth is how long a pause, seek or track change takes to become audible. That is an accepted trade-off - discarding the queued audio instead is an audible noise burst on Apple receivers - so it is now written down rather than rediscovered.

- n/a

## Types of changes

- [x] Documentation only — `documentation`

Changes:

- Document `buffer_depth` under the AirPlay configuration options
- Note under Known Issues that warm boundaries wait for the audio already queued in the speaker, so the depth is also the pause/seek delay
- Correct the `sync_adjust` entry, which no longer covers the buffer
- Correct the "pause while synced" limitation, which said pause is unsupported when the session is actually parked
- Fix a code comment in `stream.py` that still said the queue depth had no user override

## Checklist

### DIFF
--- a/music_assistant/providers/airplay/README.md
+++ b/music_assistant/providers/airplay/README.md
@@ -538,7 +538,8 @@ keeps their exposed player id stable and their Universal Player merging intact.
 ### General
 - **`password`**: Device password, stored encrypted (hidden). It is entered through the player's setup flow, not the settings form: a device that announces password protection without one stored - or that rejects the stored one - is marked as needing setup, which offers the password step again
 - **`ignore_volume`**: Ignore device volume reports (default: false)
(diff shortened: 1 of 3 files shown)


marcelveldt:
Good catch on the suppressed one - the lead really is clamped to the device-reported window (`ap2_client.c` adjusts `lead_ms` from the `latencyMin`/`latencyMax` in the SETUP reply), so I dropped a true clause there. Restored it; the point of the edit was only that the queue depth is no longer un-overridable.

marcelveldt:
Also right - `constants.py` had the same lead/buffer conflation in the sync_adjust rationale. Fixed there too.

PR #5643, opened 2026-08-13
Album art uses far less data over a remote connection

marcelveldt:

Over a remote connection album art travels through the WebRTC connection. On the dedicated
image channel added in #5635 each image was hex-encoded inside a JSON message and then
base64-chunked, so a 200 KB cover cost roughly 530 KB on the wire — about 2.7x the image,
over what is usually the user's home upstream.

That channel carries nothing but proxied responses, so its format is ours to pick: images
now travel as themselves, a small JSON header followed by the body as raw binary frames.

- Send responses on the image channel as a JSON header plus raw binary body frames
- Size those frames against the channel's negotiated message limit, so a peer that
  advertises a small limit still gets frames it can accept
- Hold the channel for a whole response, since binary frames carry no request id
- Leave the hex-in-JSON response untouched for clients that proxy over the API channel
  (the mobile app and every released frontend)

No schema bump: schema 49 and the image channel are both still unreleased, so no client
in the wild expects the old format on this channel.

Pacing was measured rather than assumed, and deliberately left alone: chunk size makes no
difference (every gap sits under the run-to-run spread), and raising the buffered-amount-low
threshold *deadlocks* the sender — libdatachannel only signals on a downward crossing, which
a stop-and-wait sender never makes.

- follow-up to #5635

## Types of changes

## Checklist

- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked: https://github.com/music-assistant/frontend/pull/2507

### DIFF
--- a/music_assistant/controllers/webserver/README.md
+++ b/music_assistant/controllers/webserver/README.md
@@ -290,6 +290,17 @@ Proxied HTTP requests are answered on the channel they arrived on. That is what 
 clients working: they send `http-proxy-request` over `ma-api` and get the response back there,
 so the gateway needs no version negotiation of its own.
 
+The reply is framed to suit that channel. On `ma-api` it is one JSON message with the body
+hex-encoded, which costs about 2.7x the image once the oversized-message chunking below is
+applied on top. `http_proxy` carries nothing else, so there the reply is a JSON header
+(`type`, `id`, `status`, `headers`, `size`) followed by the body as raw binary messages — the
+image costs its own size and no more. Those binary messages carry no request id, so the
+gateway holds the channel for a whole reply: replies go out one at a time rather than
+interleaving, which a channel that sends one message at a time would do anyway.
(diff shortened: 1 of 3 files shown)


marcelveldt:
On the suppressed comment about a swallowed send failure corrupting the stream: I dont think this one is reachable.

A dropped body frame only strands that one request - the next response header resets the client state, so nothing after it is misread. The corrupting case needs a dropped *header* while the channel keeps working for the next response. But send() only raises on an invalid/closed handle or an over-limit message: the header is a few hundred bytes, and a dead handle means the next response never goes out either. So there is no transient failure that recurs on a healthy channel.

Guarding it would mean bypassing _send_on_channel and closing the channel from a spawned task while the serve loop is still iterating it, which is a real teardown risk for a path I cannot construct a trigger for. Leaving as is - happy to revisit if it ever shows up.

PR #5653, opened 2026-08-14
Set C.UTF-8 locale and Python UTF-8 mode in Docker base image

testuser7:
In minimal base images like python:3.14-slim-trixie, the default system locale falls back to POSIX/ASCII (C). When mounting external volumes (such as Local, SMB, or NFS shares) containing files or folders with non-ASCII/Unicode characters, Python handles paths using surrogate escapes.

Setting LANG=C.UTF-8 and LC_ALL=C.UTF-8 forces glibc and underlying C libraries to use multi-byte UTF-8 encoding for mounted filesystems and subprocesses. Enabling PYTHONUTF8=1 explicitly activates Python UTF-8 Mode (PEP 540), guaranteeing that standard streams, OS paths, and file operations use UTF-8 across the entire Python runtime.

Variables are set in the final stage of Dockerfile.base so all downstream server builds and development add-ons inherit them automatically.

- proper handling of non-ASCII and Unicode file and directory names across all file providers (Local, SMB, NFS)
- zero overhead on Docker image size, as C.UTF-8 is natively supported by glibc/Debian without requiring extra locale packages
- consistent filesystem path decoding regardless of host environment variables

## Types of changes

## Checklist

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -179,6 +179,11 @@ RUN set -x \
 
 FROM python:${PYTHON_VERSION}-slim-trixie
 
+# Ensure UTF-8 encoding across system tools, subprocesses, and Python runtime
+ENV LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8 \
+    PYTHONUTF8=1
+
 # Enable non-free and contrib repositories for codec libraries
 # (trixie uses the deb822 sources format)
 RUN sed -i 's/^Components: .*/Components: main contrib non-free non-free-firmware/' /etc/apt/sources.list.d/debian.sources


marcelveldt:
Thanks for digging into this, but I checked it and unfortunately this doesn't change anything.

Python enables UTF-8 mode automatically when the locale is C/POSIX (PEP 540, since 3.7), so our images are already fully UTF-8. Verified on both the published base and server images:

```
utf8_mode = 1    fsencoding = utf-8    fserrors = surrogateescape
```

Adding LANG/LC_ALL/PYTHONUTF8 gives byte-identical output.

The surrogates in #6042 don't come from the locale. They come from `surrogateescape`, which Python always uses for paths on Linux regardless of locale, and it only kicks in for bytes that are genuinely not valid UTF-8. In that report it was a Latin-1 encoded `ä` (raw byte 0xE4) - there is no UTF-8 reading of that byte, so no env var can decode it. I tested it: with your change applied, such a filename still fails in exactly the same way, while a real UTF-8 name already worked fine without it.

One other thing to be aware of: `LC_ALL` is the strongest locale override, so baking it into the image would silently ignore any `LANG` a user sets on their container.

https://github.com/music-assistant/server/pull/5563 handles the actual case. Closing this one, but thanks for the effort!

testuser7:
@marcelveldt I'm using a Docker image on a TrueNAS Scale, and without any env variables, after mounting a local resource inside the container, I get:
```
root@truenas:/media/test# ls -la
total 2
drwxrwx--- 3 root 568 3 Aug 14 12:01  .
drwxrwx--- 5  568 568 5 Aug 14 12:00  ..
drwxrwx--- 2 root 568 2 Aug 14 12:01 'A'$'\305\233\305\274\305\272\305\202'
```

After adding the `LANG=C.UTF-8` or `LC_ALL=C.UTF-8` env variable, I get the correct directory name:
```
root@truenas:/media/test# ls -la
total 2
drwxrwx--- 3 root 568 3 Aug 14 12:01 .
drwxrwx--- 5  568 568 5 Aug 14 12:00 ..
drwxrwx--- 2 root 568 2 Aug 14 12:01 Aśżźł
```

marcelveldt:
Thanks for checking! That's a real difference, but it's `ls` doing the mangling, not the filesystem or Python.

Those octal bytes are `c5 9b c5 bc c5 ba c5 82` = ś ż ź ł, so `Aśżźł` is already perfectly valid UTF-8 on disk. Under the C locale GNU ls treats every byte >= 0x80 as non-printable and shell-escapes it, with a UTF-8 locale it renders them properly. Same bytes in both cases:

```
LC_ALL=C        ->  'A'$'\305\233\305\274\305\272\305\202'
LC_ALL=POSIX    ->  'A'$'\305\233\305\274\305\272\305\202'
LC_ALL=C.UTF-8  ->  Aśżźł
```

And what Python (so MA) actually sees is identical with and without the env vars:

```
no env vars                 ->  'Aśżźł'  -> strict utf-8 encode OK
LANG+LC_ALL+PYTHONUTF8 set  ->  'Aśżźł'  -> strict utf-8 encode OK
```

So that folder was never broken for MA, valid UTF-8 names already work fine. #6042 was the opposite case: a raw Latin-1 `0xE4` byte that has no UTF-8 reading at all, which is why no locale variable can fix it and the file had to be renamed.

That said, there is something real in here: if you shell into the container to debug, `ls` output is mangled and that's just annoying. I'd happily take a PR for only:

```dockerfile
ENV LANG=C.UTF-8
```

without `LC_ALL` (it hard-overrides any locale a user sets on their own container) and without `PYTHONUTF8=1` (already on by default when the locale is C/POSIX, PEP 540). Tick `maintenance` instead of bugfix since it doesn't change behaviour, just the shell ergonomics.

Want to reopen with that?

testuser7:
> Want to reopen with that?

I created a new PR, I hope it's OK.

PR #5658, opened 2026-08-14
Add Sendspin `source@v1` role support

maximmaxim345:

Adds a Sendspin Source plugin that exposes paired source-role clients as Live Inputs, allowing audio streams coming from line-in, turntable, (and theoretically microphone and Bluetooth audio) to play on any Music Assistant player or group. A separate plugin is required because Music Assistant assigns each provider one type, so the existing Sendspin player provider cannot also expose AudioSources. Captured audio is converted into a stable PCM stream with ASRC clock-drift correction, reconnect recovery, exclusive handoffs, and optional (by default enabled) signal-driven autostart.

## Features

- Exposes each connected and paired Sendspin source as a Live Input.
- Plays line-in or turntable receiver audio on any player, stereo pair, or group. Not only on Sendspin devices.
- Corrects capture-clock drift and network jitter through configurable buffered resampling.
- Allows independent sources to stream concurrently while keeping each source exclusive to one target.
- Preserves active streams across reconnects.
- Automatically starts and stops line-in playback for clients that report signal presence.
- Configurable automatic start player target. (default is the same device if it implements the `source` and `player` roles simultaneously)

## Known limitations

- Autostart uses the target player's normal output-protocol selection. A linked multi-protocol player may therefore use its active or preferred protocol instead of forcing Sendspin.
- Capture-only Sendspin clients use `PlayerType.UNKNOWN`. A dedicated capture-only player type requires a follow-up change in `music-assistant-models` and corresponding server/frontend handling.
- Target latency is configured for the provider as a whole rather than per source.
- Audio captured during a disconnect is lost. The existing stream emits silence and resumes after reconnection and latency-buffer recovery. This is the intentional behavior following the Specification.
- Devices without `line_sense` remain manual, and the first signal report after connecting is recorded without triggering playback.
- Options refer to `line-in` instead of the technical `source` terminology to keep it simpler.

## Potential limitations

- I have not yet tested the overhead of running ASRC/soxr on a lightweight device like a Home Assistant Green or Raspberry Pi. `aiosendspin` does have a lower quality/simpler implementation that could be used instead if that poses a problem
- Did not extensively test resuming streaming of reconnected clients.

- N/A

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) - `bugfix`
- [ ] New feature (non-breaking change which adds functionality) - `new-feature`
- [ ] Enhancement to an existing feature - `enhancement`
- [x] New music/player/metadata/plugin provider - `new-provider`
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) - `breaking-change`
- [ ] Refactor (no behaviour change) - `refactor`
- [ ] Documentation only - `documentation`
- [ ] Maintenance / chore - `maintenance`
- [ ] CI / workflow change - `ci`
- [ ] Dependencies bump - `dependencies`

## Checklist

- [x] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. - https://github.com/music-assistant/music-assistant.io/pull/948


music_assistant/providers/sendspin_source/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        async for pcm, timestamp_us in handle:
+            if self._get_session(session.client_id) is not session or session.bridge is None:
+                break
+            try:
+                session.bridge.feed(pcm, timestamp_us)
+            except ValueError as err:
+                self.logger.warning("Dropping malformed chunk from %s: %s", session.client_id, err)
+                continue
+            session.last_pcm_monotonic = time.monotonic()
+            session.pcm_received.set()
```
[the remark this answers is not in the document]
maximmaxim345: Very minor issue in practice (since it's a live stream after all parts can get cut off), but pushed a fix for that.

music_assistant/providers/sendspin_source/README.md
```diff
@@ -0,0 +1,79 @@
...
+`server/command: start`. The client announces its native stream format,
+streams timestamped encoded audio up to the server, and the Sendspin server
+library decodes it back to PCM. This provider feeds that PCM into a clock
+bridge (`aiosendspin.audio.AsrcSourceBridge`) and serves Music Assistant a
+steady 48 kHz / 16-bit / stereo stream.
+
+## Design notes
+
+- **Fixed output format.** A source's native format is only known once the
+  client starts streaming, but Music Assistant needs the stream format before
```
marcelveldt, line 30: can we fix this ? so we know the source format upfront ? That we you prevent an extra conversion layer.
another approach could be to just set the contenttype to unknown, apply a wav header at stream time and let ffmpeg deal with it

maximmaxim345: Not easily. The client reports its stream format only after we request it to start streaming. That wav header trick could work, but the ASRC already performs the sample-rate conversion while correcting clock drift. Returning unknown/WAV would add an FFmpeg conversion step without removing the ASRC step.

marcelveldt: Ok let's leave it

marcelveldt (approved):
Nice work, @maximmaxim345 !

PR #5685, opened 2026-08-14
AirPlay speakers on a Sendspin bridge now report the volume and mute they are really at

marcelveldt:

An AirPlay speaker driven through the Sendspin bridge could show a volume it was not playing at, or be silent while the player reported it as unmuted.

Volume and mute only ever travelled one way: from Sendspin down to the AirPlay player. Anything that moved them on the AirPlay side - the speaker reporting its own volume back over DACP, or a mute latched there - never reached the bridge, so the visible player kept reporting the old state. A latched mute was the worst case: the AirPlay player swallows every volume command while it is set, and nothing in the UI showed why.

- follow-up on #5677

## Types of changes

Changes:

- The bridge role can now adopt volume/mute observed on the player itself, without handing it straight back to that player.
- The AirPlay bridge manager watches its players and feeds their volume/mute changes into the bridge.
- Registering a bridge now picks up the mute state the speaker is in, not just its volume, so a rebuild no longer starts out unmuted.
- A volume adopted from the parent stays on the scale a volume command arrives on, so a speaker with a volume limit set on it no longer starts a stream louder than asked.
- Tests for both directions and the no-loop guarantee.

## Checklist

### DIFF
--- a/music_assistant/providers/airplay/player.py
+++ b/music_assistant/providers/airplay/player.py
@@ -934,15 +934,21 @@ class AirPlayPlayer(Player):
             # it would start the stream hard muted, so keep our own last known
             # volume instead.
             return False
-        # The parent's state volume is logical (0-100), while a volume command reaches
-        # this player already scaled into the parent's min/max range - which is what
-        # the level we hold has to stay on for the two to agree.
-        parent_volume = self.mass.players.scale_volume_to_device(
-            parent_player.player_id, parent_player.state.volume_level
-        )
-        if self._attr_volume_level == parent_volume:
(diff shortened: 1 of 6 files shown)


marcelveldt:
Good catch on the round trip, and it's worse than a one-off: with limits like 10-90 the level walks down a step per stream, and 0-99 walks it all the way to zero.

Fixed it differently though. Reading the resolved control player's raw level would drop the case this function exists for - the parent's volume often comes from a sibling interface that isn't rendering, and adopting that is the whole point. So it now compares on the parent's scale and leaves a level that already reads back as the parent's alone. Test covers the 0-99 range.

PR #5725, opened 2026-08-15
Preserve flow source errors

teancom:

I use a friend's flaky Audiobookshelf server and will not-infrequently get ffmpeg I/O errors when reading from it due to network "blips". Those will occasionally cause my audiobook to be marked as 'complete' and removed from my queue.  I believe I tracked it down to that fact that in flow-state, errors can be swallowed and treated as natural EOFs. This PR attempts to both allow errors to propagate cleanly while also not killing flow state on blips.

## Types of changes

## Checklist


teancom:
Tagging @MarvinSchenkel as the flow-state king.

music_assistant/controllers/streams/audio.py
```diff
@@ -2350,6 +2350,33 @@ def _superseded() -> bool:
                     del crossfade_buffer[:pcm_sample_size]
                     await asyncio.sleep(0)
 
+            # An item stream can yield buffered audio and then report a source error. Do not let that
+            # swallowed error look like a naturally completed item or flow stream. Keep the existing
+            # zero-audio skip behavior below, which allows the flow to try the next queue item.
+            if first_chunk_received and queue_track.streamdetails.stream_error:
```
marcelveldt, line 2360: Copilot is right that this never fires on a default config, and I measured it to be sure. `AudioBuffer.get_stream()` puts ffmpeg in the path as soon as there is a filter or a format mismatch, and `FFMpeg._feed_stdin` catches the exception, logs a warning and then still calls `write_eof()`, so ffmpeg exits 0 and the stream looks perfectly clean:

```
no ffmpeg (same format, no filters)  -> raises RuntimeError
ffmpeg (s16 -> f32)                  -> ends cleanly, 2.00s
ffmpeg (same format + volume filter) -> ends cleanly
```

For an audiobook that is guaranteed: normalization defaults to `fallback_dynamic`, so `_pick_pcm_bit_depth()` picks F32 for the flow while the buffer holds native s16. ffmpeg is always in the path and `stream_error` stays False.

Good news is that `audio_buffer.has_error` is still True after the loop ends, so dropping the `bytes_received == 0` condition on the check at line 1623 is enough to make `stream_error` reliable on every path. That is where I would put the actual fix.

music_assistant/controllers/streams/audio.py
```diff
@@ -2350,6 +2350,33 @@ def _superseded() -> bool:
...
+                if _superseded():
```
marcelveldt, line 2361: Separate problem with the same check: `stream_error` lives on the streamdetails, and we reuse those for 600s (`get_stream_details()`) without ever resetting it. So if the user hits play again within 10 minutes, this fires on the very first chunk and we bail out again, over and over until the streamdetails expire.

Whichever form the check ends up in, we need to clear `stream_error` when an item stream starts, or keep it local instead of on the shared streamdetails.

music_assistant/controllers/streams/audio.py
```diff
@@ -2350,6 +2350,33 @@ def _superseded() -> bool:
...
+                    return
+                # Only remove the entry we appended above. A newer producer may have replaced the
+                # shared log while this source stream was awaiting data.
+                if (
+                    pq_data.flow_mode_stream_log
+                    and pq_data.flow_mode_stream_log[-1] is play_log_entry
+                ):
+                    pq_data.flow_mode_stream_log.pop()
```
marcelveldt: The audio for this item has already gone out to the player and is still going to be played out, so I would not remove its play log entry. `_get_flow_queue_stream_index()` walks that log to work out the index and elapsed time, and with a single-item queue it ends up empty, so we report elapsed 0 and the position snaps to 0:00 for the whole remaining playout.

Keeping the entry and setting `seconds_streamed = bytes_written / pcm_sample_size` stays honest about what we actually sent.

music_assistant/controllers/streams/audio.py
```diff
@@ -2350,6 +2350,33 @@ def _superseded() -> bool:
...
+                self.logger.warning(
+                    "Queue flow for %s stopped after a stream error on %s (%s)",
+                    queue.display_name,
+                    queue_track.name,
+                    queue_track.streamdetails.uri,
+                )
+                # The previous track's tail is valid audio held back for a possible crossfade.
+                # Flush it, but never flush the failed track's partial crossfade buffer.
+                if last_fadeout_part:
```
marcelveldt, line 2372: Small one: the other paths that flush the tail without a crossfade also undo the eager seek position with `queue_track.streamdetails.seek_position = raw_seek_position` (line 2398). Missing here, and since streamdetails live on for 600s that inflated value ends up back in the reported elapsed time on a retry.

music_assistant/controllers/streams/audio.py
```diff
@@ -2350,6 +2350,33 @@ def _superseded() -> bool:
...
+                    for pcm_slice in iter_pcm_slices(last_fadeout_part, pcm_format, 1000):
+                        yield pcm_slice
+                        await asyncio.sleep(0)
+                return
```
marcelveldt: I would rather not abort the whole flow here. Once `stream_error` is set reliably, `_handle_playback_progress_report()` already bails out for items with a stream error, so the audiobook stops being marked complete, which is your actual bug. Stopping the flow on top of that costs us:

- Item 2 of 50 blips and playback just stops. `_handle_end_of_queue()` returns early because `next_item` is not None, and `_resume_on_idle()` was never scheduled, so the remaining 48 items never play. Today we move on to item 3.
- Skipping `queue_buffer_completed()` leaves `flow_stream_finished()` False, which is what `_flow_stream_underrun()` in the Cast provider needs to treat a group that underruns at EOF as idle. Such a group can sit in BUFFERING indefinitely.
- That same flag drives the lead-out in `_finish_flow_stream()`, so we `force_close()` straight away and the player drops up to 8s of audio we just yielded, including the tail flushed just above.

Can you make this a `continue` instead, so we finish the item and move on? Same as the zero-audio path right below.

music_assistant/controllers/streams/audio_buffer.py
```diff
@@ -604,10 +604,11 @@ async def _get_seekable(self, chunk_number: int) -> bytes:
 
         buffer_index = chunk_number - self._discarded_chunks
         while buffer_index >= len(self._chunks):
-            if self.cancelled or self._eof_received:
-                raise AudioBufferEOF
+            # Producer errors also set EOF after buffered data; preserve the real failure.
```
marcelveldt, line 608: Copilot is right here too and I would like this fixed in the same PR: `_get()` checks `_eof_received or cancelled` before `_producer_error` when the deque is empty, so a drained rolling buffer never gets down here. On your branch:

```
ROLLING  -> ends cleanly (error swallowed)
SEEKABLE -> raises RuntimeError
```

That is radio and audio sources. The same reorder in `_get()` plus a `BufferMode.ROLLING` case in the test would cover it.

The reorder itself is safe by the way, I checked the cancelled + error combination: `clear()` sets `_cancelled` and nulls `_producer_error` under the same lock, and `_on_producer_done()` skips cancelled tasks, so a cancelled buffer never carries an error.

marcelveldt (commented):
Thanks for digging into this, the underlying problem is real and you found the right root cause. Bit of history: in #4738 we deliberately set `_producer_error` *before* the EOF signal so readers would never mistake an abort for a clean EOF, but only the analysis reader was ever taught to check in that order. The playback readers were left behind, so your `_get_seekable` change is finishing something we started.

That said, I don't think this actually fixes your Audiobookshelf case yet, and the flow-mode half has side effects I'd rather avoid. Details inline, short version:

- The error is still swallowed one step later by ffmpeg, so the new flow branch never runs on a default config. Copilot is right about that one, I measured it to be sure.
- Rolling buffers (radio, audio sources) still turn it into a clean EOF.
- Stopping the whole flow stream is a bigger hammer than we need. I'd rather skip the failed item and keep the queue going, like we already do for an item that produces no audio at all.

One note on the tests: the new flow test replaces `get_queue_item_stream` wholesale, so it can't catch the ffmpeg swallow that makes this a no-op in practice. A test that drives a real `AudioBuffer` with a failing source through to `stream_error` would be worth more here.

Good catch though, happy to help get this over the line.

teancom:
Thank you for the review, @marcelveldt ! I was a lot more confident about the diagnosis of the issue than I was about my solution, though I wasn't able to reproduce the bug for 2 days after making the changes. But proof-through-absence is not proof at all.

Anyway, I believe I've addressed your and Copilot's comments. Please let me know if there's anything else / something I missed.

marcelveldt (approved):
Looks good. Thanks @teancom !

PR #5747, opened 2026-08-16
Queueing an album over a radio no longer plays it shuffled

marcelveldt:

A queue playing a radio or dynamic playlist is an always-on smart mix, so shuffle is switched on
for it and its toggle is locked. Picking an album and choosing "Play next (replace queue)" takes
that dynamic source away, but the shuffle it imposed stayed behind — so the album was queued up in
random order, and the queue was left showing a shuffle the user never switched on.

Follow-up to #5740, which fixed this for play and replace but not for this menu item.

- n/a

## Types of changes

## Changes

- "Play next (replace queue)" over a radio or dynamic playlist now queues the new media in its own
  order, and the queue no longer reports a shuffle the user never set.
- Queueing another radio or dynamic playlist that way still keeps the smart mix, shuffle included.

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues/README.md
+++ b/music_assistant/controllers/player_queues/README.md
@@ -110,9 +110,9 @@ analogous to how the Player Controller pairs runtime state with the wire `Player
   stay in the queue are restored to their original order rather than left shuffled behind a queue
   that now reads unshuffled. The options that only stage items for later (*add* / *next* /
   *replace next*) leave the shuffle state alone. A dynamic queue is exempt: it is an always-on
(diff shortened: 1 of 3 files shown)


marcelveldt:
Good catch on the suppressed comment - checked and it's right. Every other path that switches shuffle off drops the intent stamp with it (`set_shuffle`, `_reset_shuffle`, and the tail of `_apply_shuffle_intent`), so this branch was the only exception. With a 300s intent window that's reachable enough: toggle shuffle on, add a radio, then "play next (replace queue)" an album, and the next play would have resurrected a shuffle the toggle already read as off.

Fixed in 32e6b489c - clears `shuffle_set_at` alongside the flag, with a test covering the follow-up play.

PR #5763, opened 2026-08-17
Add Library Automations plugin: generic trigger/condition/action rules

ManuelW77:

Adds a new plugin provider, `library_automations`, that lets users define generic rules
for library events: a **trigger** (e.g. item unfavorited), optional **conditions**
(name, genre, provider, explicit flag, playlist membership), and an **action**
(add to playlist, remove from playlist, remove from library).

Motivating example: automatically move an unfavorited track into a "Sorted Out"
playlist, without needing a dedicated single-purpose plugin for that one workflow.

Rules are CRUD-managed via API commands (`list_rules`, `create_rule`, `update_rule`,
`delete_rule`, `set_rule_enabled`) and persisted per-server as JSON under the
provider's storage path. A companion frontend PR adds the rule-builder UI.

Because Music Assistant has no dedicated favorite/unfavorite event, the provider
tracks each item's last-known favorite state itself (in memory) so it can detect a
real True→False/False→True transition on `EventType.MEDIA_ITEM_UPDATED`, instead of
re-firing on every unrelated metadata update to an already-unfavorited item.

- n/a

## Types of changes

## Checklist

- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. (n/a, no shared-model changes)
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. → https://github.com/music-assistant/frontend/pull/2543
- [x] I have raised a PR against the documentation repository targeting the beta branch. → https://github.com/music-assistant/music-assistant.io/pull/950


MarvinSchenkel:
I love the idea of being able to automation **all the things**, but I don't think rebuilding an automation engine inside Music Assistant is the real solution here. 

We are planning to expose A LOT more api methods in the [client](https://github.com/music-assistant/client) and ultimately as `actions` in Home Assistant, which I think is the best way forward for your use case.

ManuelW77:
Well, i don't use MA inside of my HA, because i have it running at an other machine. 
I thing, MA should be able to do some actins without HA.
It is a simple Plugin, which can be thought forward to more automatic actions.

marcelveldt:
> Well, i don't use MA inside of my HA, because i have it running at an other machine. I thing, MA should be able to do some actins without HA. It is a simple Plugin, which can be thought forward to more automatic actions.

But you can still connect HA to a MA instance on another host and vice versa, they complement eachother.
I'm with Marvin that automations is the domain of HA and not of MA.
I'd rather spend the efforts on improving the HA integration and make sure that you can fully manipulate every aspect of MA from within HA and together with HA's great automation engine you can build whatever you want.

I'm very sorry about all the efforts you already spent on this PR but I don't think we can accept it in this state.
I'll happily take up a conversation on discord to talk about your needs and if we can find a way.

MarvinSchenkel:
Closing this one as per our discussion above. As mentioned, please reach out on Discord to talk about your needs and how we could fit them into MA 🙏

PR #5803, opened 2026-08-18
Fix white noise when playing DTS 5.1 audio packed inside WAV

vintvinst:

Title:
fix(stream): resolve white noise when playing DTS 5.1 audio packed inside WAV container

### Problem

When playing DTS-WAV via network receivers (like Denon AVR-4520 etc.) using bitstream passthrough, MA playback works fine.
However, when streaming DTS 5.1 audio encapsulated within a WAV container to players (like Sendspin, Snapcast, etc.),
ffmpeg/Music Assistant misidentifies the input stream as raw PCM (s16le), resulting in loud static / white noise instead of proper DTS decoding.

```
**Before fix (misdetected directly as PCM, producing white noise):**
music-assistant  | 2026-08-18 11:12:48.236 INFO (MainThread) [ffmpeg] Detected ffmpeg version 7.1.2 with libsoxr support
DEBUG (MainThread) [ffmpeg.233] Detected input stream info: codec=s16le sample_rate=44100 bit_depth=16 bit_rate=1411 kb/s
DEBUG (MainThread) [ffmpeg.233] Detected output stream info: codec=s16le sample_rate=44100 bit_depth=16 bit_rate=1411 kb/s
```

```
**After fix (correctly identified as DTS stream first):**
DEBUG (MainThread) [ffmpeg.230] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=1411 kb/s
DEBUG (MainThread) [ffmpeg.244] Detected input stream info: codec=s16le sample_rate=44100 bit_depth=16 bit_rate=1411 kb/s
```

### Solution:
Added conditional input arguments (-probesize 5000000 and -analyzeduration 5000000) for .wav, .aiff containers.
Updated the stream controller logic in audio.py to ensure DTS streams encapsulated in WAV containers
are correctly passed and handled by ffmpeg without falling back directly to s16le PCM mode.

WAV/AIFF containers declaring standard PCM in their headers can embed non-PCM bitstreams (such as DTS 5.1 intended for S/PDIF or HDMI passthrough).
With ffmpeg's default probe settings, the demuxer does not analyze enough data to sniff the embedded bitstream and falls back to treating the payload as raw PCM.
Increasing the probe size gives ffmpeg sufficient data to correctly detect the embedded DTS track before outputting to downstream player clients.

### How Has This Been Tested?
    Tested on Music Assistant Docker container version 2.9.13.
    Tested with ffmpeg version 7.1.2 with libsoxr support.
    Tested with Sendspin and Snapcast player clients.
    Verified that DTS-in-WAV tracks that previously produced white noise now decode into clean audio.

## Types of changes

## Checklist


music_assistant/controllers/streams/audio.py
```diff
@@ -469,28 +460,219 @@ async def get_media_stream(
...
+                # As a result, DTS 5.1 packed into WAV plays back as white noise!
+                # (e.g. DTS, intended for SPDIF/HDMI passthrough) while declaring
+                # PCM in the fmt chunk. With ffmpeg's default probesize/
+                # analyzeduration, the demuxer's content-sniffing heuristic
+                # doesn't have enough data to detect the embedded bitstream and
+                # falls back to trusting the fmt chunk, causing the payload to
+                # be treated as raw PCM downstream. Only applies to WAV/AIFF -
+                # other containers don't have this ambiguity.
+                if audio_source.lower().rsplit(".", 1)[-1] in ("wav", "aiff", "aif"):
+                    extra_input_args += ["-probesize", "5000000", "-analyzeduration", "5000000"]
```
marcelveldt: this will now punish every user that uses regular WAV files with just pcm audio in them, and it's especially hurtful for live audio sources (less of an issue as it looks at the file extension but still fragile).

Why didn't you just save these files as .dts ?
This is the first time ever I have heard of someone packing DTS into a wav container

vintvinst: The decision to use the .wav container instead of a raw .dts extension is mainly driven by legacy hardware and software compatibility:

    CD Audio & S/PDIF Standard: 
DTS-WAV (Pseudo-PCM) was designed to trick standard audio CD players 
    and digital outputs (S/PDIF / Optical). 
A classic CD player sees it as regular 1411 kbps PCM stereo and passes the raw bitstream to an  external AV receiver, which then decodes the 5.1 DTS signal. 
    Raw .dts files cannot be burned to a standard Audio CD or passed over legacy S/PDIF in this way.

    **Playback Infrastructure:** 
 Many older media players and automotive systems only recognize .wav and have no idea what a .dts file is, failing to play it entirely.

**How it works under the hood:**
A multi-channel DTS 5.1 stream is encoded to match the exact bitrate of standard PCM stereo (16-bit / 44.1 kHz at 1411 kbps). The DTS stream is encapsulated inside the WAV container so that standard software treats it like stereo audio, while a compatible DTS decoder detects the header and unfolds the 5.1 sound.

**Regarding live audio and normal WAV files:**
I have tested standard 2.0 audio.wav files via MA, and they play perfectly as well.

marcelveldt: My point is that the changes to probesize and analyzeduration will induce some extra delay which will cause issues with latency sensitive sources such as live audio sources.

How did you come up with those numbers ?
Any chance you can lower them a bit more ?

vintvinst: Here are a few key points regarding your concerns:
Impact on Live Audio / Latency:
   The patch explicitly filters by file extension: 
   if audio_source.lower().rsplit(".", 1)[-1] in ("wav", "aiff", "aif"). Live streams, HTTP/RTSP audio, or pipe inputs won't match this condition, so zero latency is added to live sources.

 How I came up with 5,000,000:
  5000000 (5MB / 5s) was chosen as a conservative bulletproof value to ensure ffmpeg catches the DTS sync word in DTS-in-WAV (DTS-CD) files, which play at 1411.2 kbps.

 Lowering the values:
  We can definitely lower these numbers! 
  Since 1 second of 16-bit 44.1kHz stereo PCM is ~176 KB, 
  even 1000000 (1MB / 1s) gives ffmpeg ~5.6 seconds of audio to analyze 
  more than enough to detect the embedded DTS header without adding noticeable open/parse delay.

I can test reduced values like 2000000 on my DTS WAV library. 
Alternatively, 
if you have specific target limits in mind for local file probing, 
let me know and I'll verify them against !  
Thanks for the feedback!  ))

OzGav: Your merge of dev has deleted your code change? In terms of actual numbers I think you will need to find the smallest values that work and then Marcel will decide from there if the additional latency is justified.

marcelveldt: yeah, I would just do 256kb or 1MB - that will have almost zero effect on latency

vintvinst: Thanks for catching that! I ran into a merge conflict during the `dev` rebase 
and I've dropped the snippet. 
I’ll re-apply and push the code changes soon.

I've updated the patch to use 256 KB / 0.5s.

```
root@o5pro:/app/venv# sed -n '530,551p' /app/venv/lib/python3.14/site-packages/music_assistant/controllers/streams/audio.py
            if isinstance(streamdetails.path, list):
                # multi part stream
                audio_source = self.get_multi_file_stream(streamdetails, seek_position)
                seek_position = 0  # handled by get_multi_file_stream
            else:
                # regular single file/url stream
                assert isinstance(streamdetails.path, str)  # for type checking
                audio_source = streamdetails.path
                # WORKAROUND: WAV/AIFF containers can embed a non-PCM bitstream
                # As a result, DTS 5.1 packed into WAV plays back as white noise!
                # (e.g. DTS, intended for SPDIF/HDMI passthrough) while declaring
                # PCM in the fmt chunk. With ffmpeg's default probesize/
                # analyzeduration, the demuxer's content-sniffing heuristic
                # doesn't have enough data to detect the embedded bitstream and
                # falls back to trusting the fmt chunk, causing the payload to
                # be treated as raw PCM downstream. Only applies to WAV/AIFF -
                # other containers don't have this ambiguity.
                if audio_source.lower().rsplit(".", 1)[-1] in ("wav", "aiff", "aif"):
                    # 256 KB probesize / 0.5s duration is enough to detect embedded DTS-in-WAV
                    # while keeping I/O overhead and latency virtually zero (<1ms).
                    extra_input_args += ["-probesize", "256000", "-analyzeduration", "500000"]
```

I ran tests directly in Music Assistant with reduced values. 
Testing showed that lowering -probesize to 256000 (256 KB) 
and -analyzeduration to 500000 (0.5s) still reliably detects embedded DTS-in-WAV 
and plays it correctly without white noise.

Since standard DTS-CD PCM data flows at ~176 KB/s, 256 KB gives FFmpeg nearly 1.5 seconds of audio buffer to identify the DTS sync words, 
while taking virtually 0ms I/O time on local/network storage.

```
docker compose logs music-assistant -f | grep -iE "detected input stream"
music-assistant  | 2026-08-19 13:04:08.001 DEBUG (MainThread) [ffmpeg.242] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=None kb/s
music-assistant  | 2026-08-19 13:04:08.211 DEBUG (MainThread) [ffmpeg.256] Detected input stream info: codec=s16le sample_rate=44100 bit_depth=16 bit_rate=1411 kb/s
music-assistant  | 2026-08-19 13:04:11.149 DEBUG (MainThread) [ffmpeg.294] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=None kb/s
```

**Why this is safe for latency:**
    No impact on Live Audio:
    The condition checks file extensions (wav, aiff, aif), so live network streams/pipes bypass this completely.
    Near-zero I/O delay:
    Reading 256 KB from disk/LAN takes less than 1ms,
    while providing enough safety margin if the WAV file contains large embedded metadata/tags before the audio payload.

vintvinst:
**This PR includes two unrelated fixes discovered while working on the DTS-in-WAV issue:**
1. Main fix: WAV/AIFF probesize workaround for DTS-in-WAV misdetection (see below)
2. Incidental fix: two `except X, Y:` (Python 2 syntax) → `except (X, Y):` in the same file, which prevented this module from compiling

**Update**: 
re-tested on a freshly built image from the latest `dev` 
(Python 3.14.7, built locally via `docker buildx build`), 
to make sure the fix still holds after the recent stream-controller refactor (`_resolve_media_stream_source`).

Confirmed working the input stream is now correctly identified as `dts` on the first ffmpeg pass:

```
[ffmpeg.366] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=1411 kb/s
[music_assistant.streams.audio.media_stream] First chunk received after 0.14 seconds (codec detected: dts)
```

Ran the full `tests/controllers/streams/` suite (853 tests) against the patched image  all passing, no regressions. 
Didn't add a new dedicated test case for the DTS-in-WAV probesize branch itself 
happy to add one if maintainers want a specific regression test, 
e.g. asserting `-probesize`/`-analyzeduration` are added only for .wav/.aiff paths).

OzGav:
You must run pre-commit before pushing, lint is failing. We use Python 3.14 so your "incidental fix" is incorrect. Giving ffmpeg more data to sniff is ok but the string match is fragile and the analyzeduration arguments are a noop

vintvinst:
Tested without `-analyzeduration` on a fresh rebuild confirmed it was indeed a noop here. 
Removed it, keeping only `-probesize 256000`:

Log from a real playback test (Eagles - Hell Freezes Over, DTS-in-WAV):
```
docker compose logs music-assistant -f|grep -iE "ffmpeg"
music-assistant  | 2026-08-20 15:21:34.101 INFO (MainThread) [ffmpeg] Detected ffmpeg version 7.1.2 with libsoxr support
music-assistant  | 2026-08-20 15:23:10.836 DEBUG (MainThread) [ffmpeg.74] Detected input duration: 205 seconds
music-assistant  | 2026-08-20 15:23:10.837 DEBUG (MainThread) [ffmpeg.74] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=None kb/s
music-assistant  | 2026-08-20 15:23:10.845 DEBUG (MainThread) [ffmpeg.74] Detected output stream info: codec=s16le sample_rate=44100 bit_depth=16 bit_rate=1411 kb/s
```

So all three points addressed:
1. Also reverted the "incidental fix" change from earlier  that was valid Python 3.14 syntax, not a bug. 
My mistake, thanks for catching it.

2. Removed `-analyzeduration` confirmed via a live rebuild that -probesize 256000 alone is sufficient (ffmpeg still correctly reports `codec=dts` on the first pass).

3. Replaced the manual `.rsplit(".", 1)` with `Path(streamdetails.path).suffix.lower()` for more robust extension matching.

`ruff check`/`ruff format --check` clean on this file (aside from the pre-existing
CPY001 on line 1, unrelated to this change).

vintvinst:
Update:
switched approach entirely, based on a real edge case found during testing , see below.

**Problem with the probesize approach**:
tested against a real multi-track DTS-in-WAV file, a whole ripped CD sampler, ~2679s,
and found the DTS sync word located 14MB into the file likely due to metadata/CUE chunks or a
non-DTS lead-in track before the actual DTS content.
No reasonable probesize value would reliably catch this without adding real latency.

**New approach**:
instead of guessing a probesize, use the codec already
detected by the library scan's ffprobe pass, which has no such probe-size limitation 
and pass it explicitly to ffmpeg via `-acodec` at stream time, skipping content-sniffing entirely.
Added a `_codec_type_from_tags()` helper in `filesystem_local/__init__.py` that reads `codec_name` from the already collected ffprobe data and sets `AudioFormat.codec_type` accordingly,
for all five places `AudioFormat` gets built "track, audiobook, podcast episode, sound effect, stream details".

Confirmed working on the previously-failing 2679s multi-track file:

```
[ffmpeg.2901] Detected input stream info: codec=dts sample_rate=44100 bit_depth=None bit_rate=1411 kb/s
[music_assistant.streams.audio.media_stream] First chunk received after 0.06 seconds (codec detected: dts)
```

**Note for existing libraries**:
since `codec_type` is now part of the cached `provider_mapping.audio_format`,
files scanned before this fix keep their old incorrect value until re-scanned.
A regular "Rescan library" didn't pick up the change for me likely skipped due to unchanged file checksum,
I only saw it take effect after forcing a re-parse & renamed the folder to test.
Happy to hear if there's a preferred way to force this, or if that's expected acceptable for a fix like this.

marcelveldt:
Yes, this is indeed the much better approach. Fix it at the root instead of working around it.

music_assistant/providers/filesystem_local/__init__.py
```diff
@@ -1545,6 +1545,24 @@ def _versioned_image_path(relative_path: str, checksum: str | None) -> str:
...
+        Return the actual audio codec detected by ffprobe, if any.
+
+        WAV/AIFF containers can embed a non-PCM bitstream e.g. DTS 5.1, intended
+        for SPDIF/HDMI passthrough) while declaring PCM in the fmt chunk. Forcing
+        ffmpeg to decode with the real codec, learned from the library scan probe,
+        which uses a much larger read window than the streaming ffmpeg pass, avoids
+        relying on ffmpeg's own content-sniffing heuristic at stream time, which can
+        miss the embedded bitstream if it's located far into a large multi-track file.
+        Falls back to UNKNOWN if the probe didn't report anything usable.
+        """
```
```suggestion
        """Return the audio codec detected by ffprobe, if any."""
```

marcelveldt (approved):
Thanks, @vintvinst !

vintvinst:
Thank you too!  @marcelveldt :1st_place_medal:

PR #5868, opened 2026-08-20
Update MSX Bridge provider from v1.4.9 to v1.6.0

trudenboy:

Updates the in-tree MSX Bridge player provider from v1.4.9 to v1.6.0.

Source: [trudenboy/ma-provider-msx-bridge](https://github.com/trudenboy/ma-provider-msx-bridge), release [v1.6.0](https://github.com/trudenboy/ma-provider-msx-bridge/releases/tag/v1.6.0).

## Summary

This update reworks MSX playback around Music Assistant queues and direct Streamserver delivery. It also removes the provider-owned browser and grouping implementations in favor of Music Assistant's generic facilities.

### Queue-backed native MSX playback

- Uses the Music Assistant player queue as the authoritative source for playback started from albums, playlists, search results, recently played items, and track lists.
- Generates native MSX playlists from the active MA queue, preserving queue item identity so repeated URIs select the correct occurrence.
- Handles next, previous, repeat-one, queue completion, long queues, radio entries, and URL-backed queue items without maintaining a separate TV-only queue.
- Synchronizes position, pause, resume, and seek events from MSX back to Music Assistant while suppressing command echoes.

### Redirect-first stream delivery

- Makes direct Music Assistant Streamserver delivery the default and rewrites the advertised host to the address used by the TV, which keeps streams reachable behind Docker or NAT.
- Keeps an independent per-TV ffmpeg proxy as a compatibility fallback when a direct per-track URL cannot be resolved.
- Applies each player's output codec and DSP/output plan through Music Assistant.
- Defaults MSX players to the `forced_content_length` HTTP profile so finite redirected tracks retain progress reporting on the TV.
- Accepts explicit flow media supplied to Universal Group members while rejecting accidental queue flow URLs that break MSX per-track progress and completion behavior.

### Security and lifecycle hardening

- Adds per-player HMAC-derived tokens to audio URLs and rejects missing or invalid tokens.
- Rejects arbitrary raw stream URLs and media that is not present in the player's active queue.
- Adds Fetch Metadata/origin checks for browser control, playlist, and WebSocket requests while retaining originless native MSX compatibility.
- Bounds prebuffering and stream task/transport cleanup so stop, disable, disconnect, and provider shutdown release active work.
- Handles stale position reports, reconnects, codec differences, served-duration clamping, and unavailable-player recovery.

### Party integration

- Extracts Party status, QR generation, and QR-on-cover rendering into a dedicated adapter.
- Restricts cover sources to Music Assistant origins, does not follow redirects, and bounds download size, decoded image dimensions, render concurrency, and cache size.
- Coalesces duplicate renders and cancels in-flight work during provider shutdown.

### Breaking removals and migration

- Removes provider-native grouping, command fan-out, and shared buffered group streams. Multi-TV playback now uses Music Assistant Universal Groups.
- Removes the bundled browser web player/kiosk, Sendspin bridge, bundled Sendspin JavaScript client, kiosk configuration, and kiosk-only endpoints. Browser playback is no longer part of this provider.
- Migrates the legacy `shared` stream mode to `independent` and removes the retired `enable_player_grouping` setting. Migration failures are logged without blocking provider startup.

## Behavior and fallbacks

- Redirect resolution failures automatically fall back to the independent local proxy.
- The independent proxy supports MP3, AAC, and FLAC. An advanced option controls estimated `Content-Length` for local MP3/AAC responses.
- Universal Group flow streams remain continuous and do not receive a false per-track HTTP length.
- When Party is unavailable, it is reported as inactive; permitted QR-cover failures fall back to the original artwork.
- Disabled MSX players remain registered so they can be re-enabled without waiting for rediscovery.

## Tests

The provider suite was reorganized around concrete Music Assistant models and explicit fakes. Coverage includes:

- queue loading, duplicate items, long queues, built-in and URL-backed media, queue rotation, repeat, and completion;
- redirect delivery, independent fallback, Universal Group flow media, output pacing, prebuffering, cancellation, and `Content-Length` behavior;
- stream tokens, raw and unqueued URI rejection, cross-site controls, same-origin WebSockets, malformed input, and invalid positions;
- native pause, resume, seek, progress rebasing, stale reports, disconnects, and player recovery;
- Party availability, QR rendering, source restrictions, image/cache bounds, concurrency, and shutdown cleanup;
- legacy configuration migration, provider lifecycle, player registration, and unload behavior.

Local upstream compatibility validation against Music Assistant `dev`: **315 passed, 1 skipped**, with lint and mypy passing.

- N/A

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue) - `bugfix`
- [ ] New feature (non-breaking change which adds functionality) - `new-feature`
- [ ] Enhancement to an existing feature - `enhancement`
- [ ] New music/player/metadata/plugin provider - `new-provider`
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - `breaking-change`
- [ ] Refactor (no behaviour change) - `refactor`
- [ ] Documentation only - `documentation`
- [ ] Maintenance / chore - `maintenance`
- [ ] CI / workflow change - `ci`
- [ ] Dependencies bump - `dependencies`

## Checklist


music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -107,7 +111,10 @@ async def _is_media_item_uri(uri: str) -> bool:
     Both spellings of a raw URL — bare, and wrapped as ``builtin://<media_type>/<url>`` —
     resolve to the builtin provider, which would make the server fetch and play whatever
     the caller names, so the resolved provider is what decides rather than the uri text.
-    The bridge only ever hands out uris of library or music provider items.
+
+    The menus only ever hand out uris of library or music provider items. Whatever the
+    user queued from the MA interface is vetted against that queue instead, so this
+    answers for the uri text alone.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2743,6 +2753,30 @@ async def _ensure_player_for_request(
...
         return player_id, device_param, player
 
+    def _uri_is_in_player_queue(self, player_id: str, uri: str) -> bool:
+        """
+        Check whether the uri is already an item in the queue this player is playing from.
+
+        A queue item's uri was put there by MA rather than by the caller, so it is safe
+        to play even when it resolves to the builtin provider — that is how a radio
+        station added by URL reaches a TV. A grouped player follows the leader's queue,
+        which is why the active queue decides rather than the player's own id.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -1445,6 +1452,9 @@ async def _handle_msx_audio(self, request: web.Request) -> web.StreamResponse:
             return web.Response(status=404, text="Player not found")
         if rejected := self._reject_invalid_stream_token(request, player_id):
             return rejected
+        # the queue check runs last so an unauthorized caller learns nothing from it
+        if not await _is_media_item_uri(uri) and not self._uri_is_in_player_queue(player_id, uri):
```
OzGav: Playing the item wipes the list that just approved it.

Four lines further down, line 1484 calls play_media(player_id, uri) with no enqueue option. For a radio or audio source, queue_loader.py line 780 falls back to CONF_DEFAULT_ENQUEUE_OPTION_LIVE_SOURCES, which defaults to REPLACE in [player_queues/config.py](https://github.com/music-assistant/server/blob/dev/player_queues/config.py) line 230, and REPLACE clears the queue.

So the first builtin item in the queue plays, the queue is then cleared down to that one item, and any other builtin item further along the same queue gets a 400 when the TV asks for it. Every test here uses a queue with one item, so none of them would show this.

Can you check what happens with two non-library builtin items queued back to back?

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2743,6 +2753,30 @@ async def _ensure_player_for_request(
...
+        :param uri: The uri the caller asked to play.
+        """
+        queue = self.provider.mass.player_queues.get_active_queue(player_id)
+        if queue is None:
+            return False
+        offset = 0
+        while batch := self.provider.mass.player_queues.items(
+            queue.queue_id, limit=_QUEUE_SCAN_PAGE, offset=offset
+        ):
+            if any(item.media_item is not None and item.media_item.uri == uri for item in batch):
```
OzGav: This check already exists twice over.

PartyProvider._queue_contains_uri in [providers/party/__init__.py](https://github.com/music-assistant/server/blob/dev/providers/party/__init__.py) line 837 is the same test, and QueueItem.uri in the models package already falls back safely when there is no media item, so the is not None half is doing work the model does for you.

Two providers asking the same question about a queue points at the queue controller as the right home. Somewhere near get_item and index_by_id in [controllers/player_queues/controller.py](https://github.com/music-assistant/server/blob/dev/controllers/player_queues/controller.py) would let both callers share it.

One thing to watch if you switch to queue_item.uri: it returns the queue item id when there is no media item, and this branch is only reached for values that failed _is_media_item_uri, which includes plain strings with no scheme. That would let ?uri=<queue_item_id> through to play_media and fail there instead of returning 400. A shared helper comparing media_item.uri explicitly avoids that.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -72,6 +72,10 @@
 PARTY_CACHE_TTL = 10.0
 PARTY_CALL_TIMEOUT = 5.0
 
+# player_queues.items() serves one page at a time, so a queue longer than this is
+# walked page by page rather than trusted to arrive in one call.
+_QUEUE_SCAN_PAGE = 500
```
OzGav: This is the third copy of 500 in the tree. [ai_radio/queue_dj.py](https://github.com/music-assistant/server/blob/dev/ai_radio/queue_dj.py) line 33 has QUEUE_PAGE_SIZE and [fastmcp_server/tools/queue.py](https://github.com/music-assistant/server/blob/dev/fastmcp_server/tools/queue.py) line 35 has MAX_QUEUE_ITEMS, both the same value for the same purpose.

If the paging stays, put it in [music_assistant/constants.py](https://github.com/music-assistant/server/blob/dev/music_assistant/constants.py) and have the three callers share it.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2743,6 +2753,30 @@ async def _ensure_player_for_request(
...
+        which is why the active queue decides rather than the player's own id.
+
+        :param player_id: The player whose queue vouches for the uri.
+        :param uri: The uri the caller asked to play.
+        """
+        queue = self.provider.mass.player_queues.get_active_queue(player_id)
+        if queue is None:
+            return False
+        offset = 0
+        while batch := self.provider.mass.player_queues.items(
```
OzGav: The loop always makes one extra call that comes back empty. Break when a page is shorter than the limit, the way [ai_radio/queue_dj.py](https://github.com/music-assistant/server/blob/dev/ai_radio/queue_dj.py) line 461 does.

Better still, drop the paging. items() in [controllers/player_queues/controller.py](https://github.com/music-assistant/server/blob/dev/controllers/player_queues/controller.py) line 231 is a slice of a list already held in memory, so items(queue.queue_id, limit=queue.items) gets the whole queue in one call. That removes the loop and _QUEUE_SCAN_PAGE along with it, which is worth doing because 500 is already defined twice elsewhere in the tree for the same purpose.

tests/providers/msx_bridge/test_http_server.py
```diff
@@ -889,6 +889,220 @@ async def test_msx_audio_rejects_missing_token(
...
+        with patch(
+            "music_assistant.providers.msx_bridge.http_server.get_ffmpeg_stream",
+            return_value=_async_iter([b"encoded"]),
+        ):
+            resp = await client.get(
+                f"/msx/audio/msx_test?uri={quote(radio_uri, safe='')}&token={token}"
+            )
+            assert resp.status == 200
+
+        assert items.call_count > 1
```
OzGav: Drop this line. The 501 item queue and the 200 response on the line above are the behaviour worth testing. Counting the calls locks in how the scan is written, so any of the simpler versions above would fail this test while still being correct.

music_assistant/providers/msx_bridge/http_server.py
OzGav: Broad excepts like this are against the projects standards and need to be tightened at your next opportunity. There are 19 in this file.

OzGav (changes requested):

marcelveldt:
Putting back to draft because what I see in a quick look is that the review comments have not been addressed but dismissed instead.

music_assistant/providers/msx_bridge/http_server.py
trudenboy: moved, caller-facing

music_assistant/providers/msx_bridge/http_server.py
trudenboy: moved

music_assistant/providers/msx_bridge/http_server.py
trudenboy: fixed in 1.4.11

music_assistant/providers/msx_bridge/http_server.py
trudenboy: paging dropped

tests/providers/msx_bridge/test_http_server.py
trudenboy: assertion removed

music_assistant/providers/msx_bridge/http_server.py
trudenboy: http_server tightened

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2071,6 +1401,17 @@ def _handle_ws_message(self, player_id: str, data: str) -> None:
...
+        elif msg_type == "seek":
+            position = msg.get("position")
+            player = self.provider.mass.players.get_player(player_id)
+            if (
+                player
+                and isinstance(player, MSXPlayer)
+                and position is not None
+                and isinstance(position, (int, float))
+            ):
+                player.note_tv_seek(float(position))
```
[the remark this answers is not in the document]
trudenboy: `cmd_seek` on a queue-backed player calls `player_queues.seek` → `play_index(..., seek_position=...)`, which rebuilds the Music Assistant stream. The TV is already playing the current `/msx/audio` response (`Accept-Ranges: none`) and will not re-request that URL until the native playlist reloads. Routing the WebSocket seek through `cmd_seek` would start a new ffmpeg from the offset while the in-flight HTTP body stays at the original encoding, and a playlist reload would interrupt native next/prev.

Elapsed time is already updated via `note_tv_seek` so the MA UI tracks the native seek. Leaving TV seek as clock + native player seek inside the current progressive stream.

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,762 @@
...
+    ) -> web.StreamResponse:
+        """Serve audio from a shared group stream."""
+        player_id = player.player_id
+        media_uri = getattr(media, "uri", "") or str(media)
+        session_id = get_media_session_id(media) or getattr(media, "queue_item_id", None) or ""
+
+        existing_stream = self.provider.get_shared_stream(group_id)
+        if existing_stream is not None and (
+            existing_stream.media_uri != media_uri or existing_stream.session_id != session_id
+        ):
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.4.22. Shared-stream identity now includes the encoded content type. A grouped member whose output codec differs from the shared producer falls back to an independent stream instead of inheriting the leader's bytes with the wrong `Content-Type`.

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,803 @@
...
+                )
+                raise self.producer_error
+
+            async with self._lock:
+                buffer_snapshot = list(self.buffer)
+                already_finished = self.finished
+                self.subscribers[player_id] = q
+                if already_finished:
+                    q.put_nowait(None)
+                subscriber_count = len(self.subscribers)
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.4.23. A second subscribe for the same player now EOFs the previous queue so the old HTTP response can exit, and teardown only `pop`s the registry if it still owns that queue — so a reconnect cannot strand the first request or detach the replacement.

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,807 @@
...
+
+    async def _write_shared_response(
+        self,
+        request: web.Request,
+        player_id: str,
+        shared_stream: SharedGroupStream,
+        headers: dict[str, str],
+    ) -> web.StreamResponse:
+        """Pump a shared stream into an HTTP response and register it for cancel."""
+        response = web.StreamResponse(status=200, headers=headers)
```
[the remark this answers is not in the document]
trudenboy: Fixed in [77e29f7](https://github.com/trudenboy/ma-server/commit/77e29f7e9).

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,807 @@
...
+                    self.producer_error,
+                )
+                raise self.producer_error
+
+            async with self._lock:
+                buffer_snapshot = list(self.buffer)
+                already_finished = self.finished
+                previous = self.subscribers.get(player_id)
+                if previous is not None and previous is not q:
+                    _signal_eof(previous)
```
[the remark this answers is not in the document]
trudenboy: Fixed reconnect-specific EOF handling in [77e29f7](https://github.com/trudenboy/ma-server/commit/77e29f7e9).

music_assistant/providers/msx_bridge/player.py
```diff
@@ -395,7 +434,14 @@ async def _propagate_to_group_members(self, command: str, **kwargs: Any) -> None
...
             if tasks:
-                await asyncio.gather(*tasks, return_exceptions=True)
+                results = await asyncio.gather(*tasks, return_exceptions=True)
+                for result in results:
+                    if isinstance(result, BaseException):
+                        self.logger.warning(
+                            "Failed to propagate %s to a group member",
+                            command,
+                            exc_info=(type(result), result, result.__traceback__),
+                        )
```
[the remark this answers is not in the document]
trudenboy: Fixed in [77e29f7](https://github.com/trudenboy/ma-server/commit/77e29f7e9).

music_assistant/providers/msx_bridge/mappers.py
```diff
@@ -6,14 +6,31 @@
...
+
     from .provider import MSXBridgeProvider
 
 logger = logging.getLogger(__name__)
 
 
+def queue_nav_properties(player_id: str, prefix: str = "") -> dict[str, str]:
+    """MSX next/prev/complete must go through the MA queue, not the native list."""
+    next_action = f"execute:{prefix}/api/next/{player_id}"
+    prev_action = f"execute:{prefix}/api/previous/{player_id}"
```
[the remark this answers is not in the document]
trudenboy: Explicit-player routes no longer derive/register an IP player in [77e29f7](https://github.com/trudenboy/ma-server/commit/77e29f7e9).

music_assistant/providers/msx_bridge/party.py
```diff
@@ -0,0 +1,284 @@
...
+            headers={"Cache-Control": "no-store"},
+        )
+
+    def qr_cover_task(
+        self, cache_key: tuple[str, str], image_url: str, join_url: str
+    ) -> asyncio.Task[bytes]:
+        """Return the in-flight render task for this cover, starting one if needed."""
+        if (task := self.qr_cover_inflight.get(cache_key)) is None:
+            task = asyncio.create_task(self._fetch_and_render_cover(cache_key, image_url, join_url))
+            self.qr_cover_inflight[cache_key] = task
```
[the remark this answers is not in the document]
trudenboy: Shutdown now cancels and gathers QR tasks in [77e29f7](https://github.com/trudenboy/ma-server/commit/77e29f7e9).

music_assistant/providers/msx_bridge/VERSION
```diff
@@ -1 +1 @@
-1.4.9
+1.4.24
```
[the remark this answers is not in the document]
trudenboy: PR metadata now matches v1.4.24.

music_assistant/providers/msx_bridge/queue_handshake.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def _prepare_lock(player: Any) -> asyncio.Lock:
+    """Return the per-player prepare lock, creating it if missing (test doubles)."""
+    lock = getattr(player, "_prepare_lock", None)
+    if isinstance(lock, asyncio.Lock):
+        return lock
+    lock = asyncio.Lock()
+    player._prepare_lock = lock
+    return lock
```
OzGav: I think you can delete this and use `player._prepare_lock` directly. `MSXPlayer.__init__` already creates the lock in player.py at line 78, and test_player.py at line 23 asserts it exists, so there is no player that reaches here without one. Writing an attribute onto the player from a module function to cover a case that cannot happen is worth avoiding.

music_assistant/providers/msx_bridge/mappers.py
OzGav: This branch is unreachable and would fail if it ever ran. Every caller of `map_track_to_msx` passes `context_uri`, so playback from the menus goes through` /api/play-context`. If something did reach this branch, the URL it builds has no `queue_item_id` and names a track that is not in the queue yet, so `/msx/audio` would return 400. Please remove it, or say which caller still needs it.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -1435,119 +1202,46 @@ async def _handle_msx_audio(self, request: web.Request) -> web.StreamResponse:
...
-                if queue_item.media_item:
-                    duration = getattr(queue_item.media_item, "duration", None) or duration
-                if not duration and queue_item.duration:
-                    duration = queue_item.duration
-        return int(duration)
+        """Return the length in seconds of the audio served for the given media."""
+        return resolve_served_duration(self.provider.mass, media)
 
     @staticmethod
     def _build_audio_params(
```
OzGav: `_build_audio_params` has no callers anywhere that I could see. `_serve_shared_stream` has none either, apart from `test_group_stream.py` line 251. Both should go, along with the test that only exists to call one of them.

music_assistant/providers/msx_bridge/queue_handshake.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class PrepareFailure:
+    """HTTP-shaped failure from preparing MSX audio."""
+
+    status: int
+    text: str
```
OzGav: This puts HTTP status codes into a module that otherwise knows nothing about HTTP, and the docstring says so. The queue handshake should say what went wrong and let `_handle_msx_audio` decide the status. Raising one of the errors from `music_assistant_models.errors` and catching it in the handler keeps the status codes in the one place that serves HTTP.

OzGav (changes requested):
General comment, on the mocks

Several pieces of the production code are shaped around the test doubles rather than the real objects, and it is worth pulling that apart.

`test_http_server.py` line 1248 builds queue items with a bare `Mock`, which answers any attribute with another `Mock`. That is why `queue_item_uri` in `queue_handshake.py` at line 53 walks a chain of `getattr` calls and then checks` isinstance(candidate, str)`, and why `_queue_length` in `player.py` at line 464 reads `queue.items` through `getattr`, tests it with `isinstance`, falls back to 500, and catches `TypeError` and `AttributeError`. Against a real `QueueItem` and a real `PlayerQueue`, none of that is needed. `QueueItem.media_item` is typed and `PlayerQueue.items` is a plain int, so both would be one line.

The same issue is why `queue_handshake.py` types `mass` as `Any` on lines 63, 82 and 98. That switches off type checking for the module that decides what a TV is allowed to play. It should take the real `MusicAssistant` type.

Building the fixtures from real model objects would let all of this collapse, and it would test what actually runs.

music_assistant/providers/msx_bridge/queue_handshake.py
trudenboy: Fixed in [e5574c8](https://github.com/trudenboy/ma-server/commit/e5574c806df285721733962235a154f811d2fa1f): the helper was removed and player._prepare_lock is used directly.

music_assistant/providers/msx_bridge/mappers.py
trudenboy: Fixed in [e5574c8](https://github.com/trudenboy/ma-server/commit/e5574c806df285721733962235a154f811d2fa1f): the invalid direct-audio fallback was removed.

music_assistant/providers/msx_bridge/http_server.py
trudenboy: Fixed in [e5574c8](https://github.com/trudenboy/ma-server/commit/e5574c806df285721733962235a154f811d2fa1f): both unused wrappers were removed and the test now calls AudioPipeline directly.

music_assistant/providers/msx_bridge/queue_handshake.py
trudenboy: Fixed in [e5574c8](https://github.com/trudenboy/ma-server/commit/e5574c806df285721733962235a154f811d2fa1f): the handshake now raises typed MA errors and the HTTP handler maps them to statuses.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -3,93 +3,85 @@
...
-
 
-class PartyInfo(NamedTuple):
-    """Active-party details resolved from the MA Party plugin."""
-
-    join_url: str
-    name: str | None
-    qr_text: str | None
-    qr_version: str
+def _queue_item_limit(queue: Any) -> int:
```
OzGav: This is the helper you just removed from `player.py`, still here under another name. Any, a getattr, an isinstance, and a fallback of 500.
The only caller is line 1207, and line 1200 above it already holds the queue from `player_queues.get(queue_id)`, which is typed `PlayerQueue | None`. So the whole helper is a `if queue is None` check away from disappearing, and `limit=queue.items` reads straight off the model. Line 1204 has the same shape and can become `queue.current_index or 0` in the same edit.

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,822 @@
...
+                len(tasks),
+                len(transports),
+                player_id,
+            )
+
+    async def serve(
+        self,
+        request: web.Request,
+        player: MSXPlayer,
+        media: Any,
```
OzGav: serve, serve_shared at line 386 and serve_independent at line 497 all take media: Any, and `resolve_served_duration` at line 812 takes mass: Any. That is the module that actually serves the audio, and typing it as Any means mypy checks nothing in it.
The real type is known. `prepare_msx_audio` returns PlayerMedia, and `_handle_msx_audio` passes that straight into serve. You already type it correctly on line 812, where the same parameter is media: PlayerMedia.
Once those four are typed, the getattr calls on lines 394, 395, 400, 402, 479, 488, 513 and 515 all become direct attribute reads, because PlayerMedia has uri, source_id and `queue_item_id` as plain fields. `audio_stream.py` has eight Any annotations, more than any other file in the provider. `queue_handshake.py` now has none, so the pattern is proven.

music_assistant/providers/msx_bridge/mappers.py
```diff
@@ -122,34 +211,35 @@ def map_track_to_msx(
...
         action = f"playlist:{playlist_url}"
     else:
-        action = _build_audio_action(
-            prefix=prefix,
-            player_id=player_id,
-            track_uri=track.uri,
-            token=provider.get_stream_token(player_id),
-            device_param=device_param,
-        )
+        raise ValueError("context_uri or playlist_url is required")
```
OzGav: Raising here trades one unreachable branch for another. No caller of `map_track_to_msx` passes `playlist_url` either, only `test_mappers.py` line 102, so both the `elif` and this `else` are dead in production.
If `playlist_url` has a real future, leave it. If not, make `context_uri` a required argument and delete both branches. Either way a `ValueError` is the wrong choice for a condition that cannot occur, and MA has its own error types for the ones that

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2273,279 +1598,39 @@ async def _handle_recently_played(self, request: web.Request) -> web.Response:
...
-            target_origin = self._url_origin(image_url)
-        except ValueError:
-            return False
-        if target_origin[0] not in ("http", "https") or not target_origin[1]:
-            return False
-        allowed_bases = [self._get_prefix(request)]
+        """Serve a cover image with the party QR stamped into its corner (PNG)."""
+        extra_bases: list[str] = []
         for source in (
             getattr(self.provider.mass, "webserver", None),
```
OzGav: `mass.webserver` and `mass.streams` are always present, and both expose `base_url` as a plain string. The `getattr` pair, the `isinstance` and the `startswith("http")` check are guarding against a shape that cannot reach here.

OzGav (changes requested):

music_assistant/providers/msx_bridge/static/plugin.html
```diff
@@ -463,6 +509,17 @@
...
 
+    MAHandler.prototype.handleEvent = function(data) {
+        // Native MSX playback lives outside this plugin's DOM, so its media
+        // events arrive here rather than on a local audio/video element.
+        if (!data || typeof data.event !== "string") return;
+        if (data.event === "video:pause") {
+            onMediaPause();
+        } else if (data.event === "video:play") {
+            onMediaPlay();
+        }
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.7: `video:seek` now forwards `data.position` through the server-seek echo guard.

music_assistant/providers/msx_bridge/audio_stream.py
```diff
@@ -0,0 +1,814 @@
...
+    """
+
+    def __init__(
+        self,
+        group_id: str,
+        media_uri: str,
+        session_id: str = "",
+        content_type: ContentType | None = None,
+    ) -> None:
+        """Initialize shared stream for a group."""
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.8: shared ffmpeg output uses 16 KB chunks; catch-up and subscriber queues are capped at 37 chunks (~592 KB).

music_assistant/providers/msx_bridge/static/plugin.html
```diff
@@ -237,18 +259,38 @@
...
     }
 
+    function reportTvSeek(pos) {
+        var now = Date.now();
+        if (pendingServerSeek != null && now <= pendingServerSeekUntil) {
+            if (Math.abs(pos - pendingServerSeek) <= 2) {
+                clearPendingServerSeek();
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.8: `reportTvSeek` updates `pausedAtPosition` before the echo-suppression return.

music_assistant/providers/msx_bridge/player.py
```diff
@@ -228,6 +255,15 @@ def update_position(self, position: float) -> None:
         self._last_ws_position = time.monotonic()
         self.update_state()
 
+    def note_tv_seek(self, position: float) -> None:
+        """Trust a TV-initiated seek even before the first position report."""
+        if self._attr_playback_state != PlaybackState.PLAYING:
+            return
+        self._accepted_position = True
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.8: trusted TV seeks update PLAYING and PAUSED state; ordinary position reports remain PLAYING-only.

music_assistant/providers/msx_bridge/http_server.py
trudenboy: fixed: the helper was removed; the handler reads queue.current_index and queue.items directly

music_assistant/providers/msx_bridge/audio_stream.py
trudenboy: fixed: audio methods use PlayerMedia and MusicAssistant types, and the getattr fallbacks were removed.

music_assistant/providers/msx_bridge/mappers.py
trudenboy: fixed : context_uri is required and the unreachable playlist_url/ValueError branches were removed.

music_assistant/providers/msx_bridge/http_server.py
trudenboy: fixed : webserver.base_url and streams.base_url are now read directly.

music_assistant/providers/msx_bridge/party.py
```diff
@@ -0,0 +1,330 @@
...
+        Results are cached briefly.
+        """
+        now = time.monotonic()
+        if self.cache is not None and now - self.cache[0] < PARTY_CACHE_TTL:
+            return self.cache[1]
+        info: PartyInfo | None = None
+        try:
+            party = cast("PartyProvider | None", self.provider.mass.get_provider("party"))
+            if party is not None:
+                join_url = await asyncio.wait_for(party.get_party_url(), PARTY_CALL_TIMEOUT)
```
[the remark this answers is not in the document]
trudenboy: Related PR #6184

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2570,17 +1658,47 @@ def _reject_cross_site(request: web.Request) -> web.Response | None:
         player, MSX interaction plugin, dashboard) or non-browser clients that
         omit the header entirely — both pass.
         """
-        if request.headers.get("Sec-Fetch-Site", "").lower() == "cross-site":
+        fetch_site = request.headers.get("Sec-Fetch-Site", "").lower()
+        if fetch_site not in ("", "none", "same-origin"):
             return web.json_response({"error": "Cross-site request rejected"}, status=403)
```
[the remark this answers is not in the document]
trudenboy: out of scope this pr

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -1365,51 +1168,39 @@ async def _handle_msx_search_playlist(self, request: web.Request) -> web.Respons
...
             self.provider,
             device_param,
             qr_cover_base=await self._qr_cover_base(prefix),
         )
-        return web.json_response(playlist.model_dump(by_alias=True, exclude_none=True))
+        return web.json_response(dump_msx(playlist))
 
     # --- MSX Queue Playlist ---
 
     async def _handle_queue_playlist(self, request: web.Request) -> web.Response:
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.9: the queue-playlist handler now calls `_reject_cross_site()` before reading the queue or rendering token-bearing actions. A regression test verifies the 403 response and that no token or queue data is exposed.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2570,17 +1658,47 @@ def _reject_cross_site(request: web.Request) -> web.Response | None:
...
+            return _msx_execute_error(400, "Invalid uri")
+        start = _non_negative_int_param(request.query, "start", 0)
+        track_uri = request.query.get("track")
+        if track_uri and not await is_media_item_uri(track_uri):
+            track_uri = None
+        self.provider.on_player_activity(player_id)
+        try:
+            with player.suppress_ws_notify():
+                player.expect_new_media()
+                await self.provider.mass.player_queues.play_media(player_id, uri)
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.9: both `/api/play` and context playback now wrap `play_media` in the configured owner `ImpersonatedUser`. Regression tests cover attribution on both paths.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -1429,507 +1220,56 @@ async def _handle_msx_audio(self, request: web.Request) -> web.StreamResponse:
...
+                player,
+                uri,
+                from_playlist=from_playlist,
+                queue_item_id=requested_queue_item_id,
+            )
+        except InvalidDataError as err:
+            return web.Response(status=400, text=str(err))
+        except ResourceTemporarilyUnavailable as err:
+            return web.Response(status=504, text=str(err))
+        except Exception:
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.9: both handlers now catch only `MusicAssistantError`, `OSError`, and `TimeoutError`. Regression tests confirm unexpected programming errors still surface as HTTP 500.

music_assistant/providers/msx_bridge/provider.py
```diff
@@ -439,7 +198,7 @@ async def get_owner_username(self) -> str | None:
                         self._owner_username = user.username
                         self.logger.debug("Resolved owner username: %s", self._owner_username)
                         break
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.10 (`be862adc`): the first-enabled-user heuristic and cache were removed. Both unauthenticated MSX playback paths now use `ImpersonatedUser(self.provider.mass, None)` as external hardware; authentication and attribution remain scoped to a separate PR.

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -1365,51 +1170,41 @@ async def _handle_msx_search_playlist(self, request: web.Request) -> web.Respons
...
-        return web.json_response(playlist.model_dump(by_alias=True, exclude_none=True))
+        return web.json_response(dump_msx(playlist))
 
     # --- MSX Queue Playlist ---
 
     async def _handle_queue_playlist(self, request: web.Request) -> web.Response:
         """Return the current MA queue as an MSX native playlist."""
-        _, device_param, _ = await self._ensure_player_for_request(request)
+        if rejected := self._reject_cross_site(request):
+            return rejected
```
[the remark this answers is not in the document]
trudenboy: Fixed in v1.5.10 (`be862adc`): the CORS middleware now centrally rejects cross-site requests for both `/msx/playlist/*` and `/msx/queue-playlist/*` before any handler renders token-bearing actions. A parameterized regression test covers all six routes.

music_assistant/providers/msx_bridge/VERSION
```diff
@@ -1 +1 @@
-1.4.9
+1.6.0
```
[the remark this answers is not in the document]
trudenboy: fixed

music_assistant/providers/msx_bridge/http_server.py
```diff
@@ -2642,25 +1768,90 @@ async def _handle_next(self, request: web.Request) -> web.Response:
...
-        if self._get_msx_player(player_id) is None:
+        player = self._get_msx_player(player_id)
+        if player is None:
             return web.json_response({"error": "Unknown MSX player"}, status=404)
         self.provider.on_player_activity(player_id)
-        await self.provider.mass.players.cmd_previous_track(player_id)
-        return web.json_response({"status": "ok"})
+        with player.suppress_ws_notify():
+            await self.provider.mass.players.cmd_previous_track(player_id)
+        return _msx_execute_ok(self._queue_playlist_action(request, player_id))
```
OzGav, line 1793: `_handle_next` above got a guard for this and `_handle_previous` did not.

Lines 1775 to 1779 record the index, call the command, and return a bare `_msx_execute_ok()` when `_queue_advanced` reports no movement. This handler calls `cmd_previous_track` and then reloads the playlist unconditionally. At index 0 of a queue with repeat off, `PlayerQueuesController.previous` does nothing, so the reload rotates to the same index and the TV restarts the track the user was already on.

The helpers are already there and `_queue_advanced` at line 1835 handles repeat-one correctly, so this is the same three lines you added to `_handle_next`.

OzGav (commented):
The description covers three fixes between v1.5.10 and v1.6.0, but the PR takes the bundled provider from v1.4.9 to v1.6.0 and removes two configurable features along the way that being shared buffer streaming and player grouping. 

Please describe the full v1.4.9 to v1.6.0 change and say what happens to a user who had those enabled. I would also reconsider the enhancement label, since removing two modes people may be running is a breaking change for them.

Of the 68 automated review threads on this PR, 57 are marked resolved with no reply. I spot checked the ones still anchored to unchanged code and most of them are genuinely fixed, so the work is being done.

The problem is that a resolve with no reply gives a reviewer nothing to verify against. It costs a reader a trip through the diff to establish something you already know. You replied on eleven of them, and those were the easier ones to confirm. Please do that on the rest.

Also, this is the fourth review round and the provider has moved v1.4.10, v1.4.24, v1.5.4, v1.6.0 while under review. Each round rewrites or deletes a large part of what the previous round looked at, including whole modules and their tests.

Please pick a version, hold it, and let the review finish against that. Anything after it can go in a follow-up PR. Reviewing a moving target costs more than it finds.

MarvinSchenkel:

PR #5908, opened 2026-08-22
Dissolve sync group when a playback start never materializes

OzGav:

A sync group formed for a start that fails before the leader ever reports PLAYING (e.g. an announcement whose media URL returns 404) stayed formed forever: the idle-grace dissolve is only armed on a PLAYING/PAUSED -> IDLE transition, which never happens for a group that goes IDLE -> IDLE from inception. The lingering group kept all members captured, so a later play_media targeting a single member was redirected to the (stale) group queue.

Arm the idle-grace timer from _await_leader_playback when the leader has not confirmed playback by the time the start wait ends (including when the play command itself raises). A start that merely settles late cancels the timer again through the existing PLAYING update handling, and the grace runner re-checks state at fire time, so a healthy session is never torn down. Groups pinned via Fake power control are left alone, matching the existing idle-grace behavior.

- related issue https://github.com/music-assistant/support/issues/6159

## Types of changes

## Checklist


MarvinSchenkel (approved):
Nice find, thanks @OzGav

PR #5929, opened 2026-08-23
Fix podcast episodes showing in the wrong order

OzGav:

Podcast feeds that only number some of their episodes (like Science Vs) showed episodes in a scrambled order. The episode listing jumped between real episode numbers and feed index positions because we mixed two incompatible numbering schemes within the same show.

Episode ordering was also inconsistent between providers, so playing from a chosen episode could queue up older episodes instead of newer ones.

- Use episode numbers only when every episode in the feed has one, otherwise work the order out from the publication dates
- Oldest episode is always the lowest number, so the newest episode is always the highest
- Fixed across iTunes, Podcast Index, Podcast Feed, gPodder, Overcast, Pocket Casts, Spotify, Audiobookshelf, YouTube Music, Deezer, OpenSubsonic and Audible
- Playing from a chosen episode now queues that episode and the newer ones, on every provider
- Asking for the latest episode of an RSS feed podcast no longer plays the oldest one
- Serial podcasts, the ones that ask apps to list oldest first, are no longer shown in reverse
- Podcasts split into seasons no longer have their episodes interleaved
- Single episode and browse paths no longer invent a position when one isn't available

Now that oldest = lowest is standardised then the default sort for podcasts needs to be position descending which is done in this frontend PR https://github.com/music-assistant/frontend/pull/2620

- n/a

## Types of changes

## Checklist


fmunkes (approved):
The PR is fine, but why did we change episode ordering in the first place? The frontend remembers the selected sorting as far as I know, so the new default will not apply to already present podcasts, i.e. users will have to change the sorting of all their podcasts manually...

OzGav:
I have a frontend change to swap to descending order (which I see you just saw!) to retain the original behaviour. I just found that there were inconsistencies and ABS had the most robust solution and was going oldest to newest  (just using timestamps vs integers) so I aligned everything. This came about because I am working on a podcast details view and wanted to be able to step through the episodes with buttons and I found a podcast where the order was jumping all over the place and that led me here?

fmunkes:
> I am working on a podcast details view

Great!

I am not 100% sure, if the old sorting is stored if you opened that podcast already. I.e. your frontend change might only change the order it is displayed, but if a podcast was already subscribed to, it may stick to the old sorting and only the new one gets your updated default. And in addition, this also changes the order returned by the API, right, so if you happen to have used some rest command, you need to change your automation.

OzGav:
On Fabian's API question we discussed on Discord and the order of the episodes returned by the API remain the same just the number assigned to them changes. So someone would have to have an API call that was looking at the position values rather than just the returned list. This is ok as the PR actually fixes something for these people in that some podcasts were getting scrambled position numbers currently.

MarvinSchenkel:
Can we check if we still need to update these four providers in this PR as well?
1. deezer/media.py:657 — cached_ep.position = position over result.raw_episodes
2. server/music_assistant/providers/ard_audiothek/__init__.py:456 — offset + idx
3. audible/audible_helper.py:870 — incrementing counter across pages
4. yandex_music/provider.py:1110 — incrementing counter over album volumes

OzGav:
Yes I did think about these before but initially I was concerned I didn't know exactly what order the API supplies the episodes. But now on review I can do Deezer and OpenSubsonic because they both already give us a publish date for each episode, so we can rank on that and it doesn't matter which way round the API lists them.

I have an Audible subscription so I will check that one. We already read a release date from the same payload for audiobooks, it just isn't read for episodes, so it looks promising.

That leaves ARD Audiothek and Yandex. ARD doesn't ask for a date at all in its query, though it does fetch an episode number we currently ignore. Yandex has no per episode date anywhere in the client, so ranking by date isn't possible there. I will ask the provider maintainers if they know how their APIs order things.

OzGav:
@khers just a FYI on the change here

khers:
Please don't add the subsonic change into the RC unless we have a demonstrated case that it is broken, I would like to test it without a time crunch first. Everything is working for subonsic podcast ordering now AFAICT.

OzGav:
@khers you're right that it works today. The catch is that it only works because the frontend currently sorts position ascending, and this change comes with [frontend#2620](https://github.com/music-assistant/frontend/pull/2620) which flips the default to descending.

Your gonic fixture shows it as the newest episode gets position 1 and the oldest gets 5. Ascending that reads correctly, descending it is fully reversed. So we need to just invert the ordering. 

One thing you could confirm is we rank on publishDate, and where a server sends none we fall back to assuming the listing is newest first. Gonic is newest first so it is fine, but if you know a server that omits publishDate or lists oldest first, that is the case that would catch us out.

khers:
The good news is I don't think many (maybe any) other Open Subsonic implementations support podcasts, however the publishDate field is not required in the PodcastEpisode type so this may bite us at some point.

OzGav:
Agreed, and thanks for checking. We will have to revisit if it comes up... fingers crossed it doesnt!

PR #5955, opened 2026-08-24
Fix Flow Mode sample rate setting being locked for players with enforced flow mode

vintvinst:
### Problem

For HTTP-based players where flow mode is enforced rather than user-toggleable
(`player.requires_flow_mode == True`, e.g. Home Assistant native MediaPlayers),
the "Flow Mode sample rate" setting in Player Settings renders as permanently
disabled in the UI — even though a value (e.g. `bit_perfect`) is correctly
saved and actively applied by the backend during playback.

Confirmed via DOM inspection that the disabled dropdown correctly shows the
saved value:
```html
<option selected="" value="bit_perfect"></option>
<input disabled="" ... value="Bit-perfect">
```

And confirmed via live playback logs that this saved value is being honored a Denon AVR-4520 receiving bit-perfect DTS 5.1 passthrough with this setting locked to `bit_perfect`:

Before (Settings: disabled): 
![Screenshot] https://i.ibb.co/4w6fqcy6/00-Before-Fix.png

After (Settings: editable): 
[!screenshot] https://i.ibb.co/7Nr7XfxM/01-After-Fix.png

### Root cause

`CONF_ENTRY_FLOW_MODE_SAMPLE_RATE` has `depends_on=CONF_FLOW_MODE`. Per
`ConfigEntry.is_satisfied()`:

```python
if self.depends_on is None:
    return True
dependency = next((e for e in entries if e.key == self.depends_on), None)
if dependency is None:
    return False
```

For players with `requires_flow_mode == True`, `CONF_ENTRY_FLOW_MODE` is
intentionally omitted from the returned entries list (`controllers/config/players.py`)
since there's no user-facing toggle needed — flow mode is always on. But this
means the `depends_on` key is never present in `entries`, so `is_satisfied()`
always returns `False`, and the frontend renders the sample-rate field as
disabled — regardless of the actual (correctly applied) saved value.

### Fix

For players where flow mode is enforced, the dependency is trivially always
satisfied (flow mode is unconditionally active), so drop `depends_on` on a
copy of `CONF_ENTRY_FLOW_MODE_SAMPLE_RATE` for that branch instead of trying
to fake a `CONF_ENTRY_FLOW_MODE` entry.

### How Has This Been Tested?

- Tested on Music Assistant Docker container built from the latest `dev` and `2.10.0.dev2026082303`.
- Verified the field is now editable in Settings for a Home Assistant native
  MediaPlayer (`denon_avr_4520_perfect_ma`), with `bit_perfect` still correctly selected and saved.
- Ran `tests/controllers/config/` (307 tests) — all passing.

Fixes a UI bug where the **"Flow Mode sample rate"** setting was permanently disabled for players with enforced flow mode (e.g. Home Assistant native MediaPlayers).

## Types of changes

## Checklist


music_assistant/controllers/config/players.py
```diff
@@ -668,7 +668,13 @@ async def _get_player_config_entries(
                 # add flow mode entry for http-based players that do not already enforce it
                 if not player.requires_flow_mode:
                     default_entries.append(CONF_ENTRY_FLOW_MODE)
-                default_entries.append(CONF_ENTRY_FLOW_MODE_SAMPLE_RATE)
+                    default_entries.append(CONF_ENTRY_FLOW_MODE_SAMPLE_RATE)
+                else:
+                    # Flow mode is enforced for this HA-player. Clear depends_on so the
```
marcelveldt:
```suggestion
                    # Flow mode is enforced for this player. Clear depends_on so the
```

marcelveldt (approved):
Thanks, @vintvinst !

PR #5978, opened 2026-08-25
Add per player HEOS playback transition configuration

Tommatheussen:

**I'm currently on a business trip so can't test on actual hardware**

A previous fix made sure playback transition events are ignored while a stream was just started on HEOS from MA. Some players take a longer time to respond their initial state, falling (just) outside the 5 second window we fixed initially.
This change moves the window from a fixed 5 second interval to a per player config based option.

The alternative solution is to just bump the fixed timeout value, technically also possible as this interval is used to ignore state updates from HEOS during the playback startup window. A larger fixed window would just miss some playback updates (like starting another external source). You can argue this doesn't matter as we've just started playing from MA instead. Going with that solution means less options for a user.

- related issue https://github.com/music-assistant/support/issues/5614#issuecomment-5398603852

## Types of changes

## Checklist


OzGav (approved):
LGTM!

PR #6025, opened 2026-08-26
Hide the Sendspin legacy clients option

maximmaxim345:

The Sendspin provider shows an "Allow legacy clients" toggle that is enabled by default, and leaving it visible invites two problems: client authors read it as a promise that the spec as it stands today is the target to implement, and anyone who switches it off is left with a stored `false` that turns into a breaking change the moment a patch release bumps the supported Sendspin version.

The option is now hidden in the UI while the config entry itself stays declared, so the stored value, the API, and the provider reload on change all behave exactly as before. It gets unhidden again in the release that lands the protocol changes.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sendspin/provider.py
+++ b/music_assistant/providers/sendspin/provider.py
@@ -405,6 +405,7 @@ class SendspinProvider(PlayerProvider):
                 key=CONF_ALLOW_LEGACY_CLIENTS,
                 type=ConfigEntryType.BOOLEAN,
                 default_value=True,
+                hidden=True,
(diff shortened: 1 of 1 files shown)


maximmaxim345:
Should we add a migration that re-enables legacy compatibility for beta/nightly users?
Otherwise they can't turn it back on if they disabled it in a prior version (though I'm not sure that anyone actually did that).
Or accept that for now and just unhide it again with the first beta of 2.11?

marcelveldt (approved):

PR #6065, opened 2026-08-27
Fix Spotify not loading after updating to 2.10

kiegsgroot:

After updating to 2.10, Spotify stops loading with "Spotify playback authorization required", even on an account that was fully set up and working before. Playback is dead and every queue request returns "No playable items found". Running the Spotify setup again fixes it, which is why this looks like an ordinary re-authentication prompt.

2.10 moved provider credentials out of the plain config values into `setup_data`, and a one-off migration moves them across on the first start. That migration skipped Spotify, so the credentials stayed where 2.9 wrote them, and the new code no longer looks there.

`setup()` still finds the global token, because `seed_stored_config_values` exposes the stored values at that point, so the load starts. `rehydrate_provider_config` then re-parses the config against the declared config entries only, and none of the credentials are declared entries any more, so they all drop out. `LibrespotBackend.setup()` is the first to read one and raises.

The credentials themselves are never lost, so an install that already hit this recovers on the next start without re-authenticating.

**Changes**

- The migration now moves Spotify's `refresh_token_global`, `refresh_token_dev`, `librespot_credentials`, `client_id` and `account_id` into `setup_data`.
- The deprecated `refresh_token` deliberately stays in `values`, because `_migrate_legacy_token` reads it from there during `setup()` and clears it afterwards.
- That same function now reads `client_id` from `setup_data` first, since the migration is what moved it.

- https://github.com/music-assistant/support/issues/6208

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/config/migrations.py
+++ b/music_assistant/controllers/config/migrations.py
@@ -347,8 +347,8 @@ _LEGACY_ICON_MAP: dict[str, str] = {
 # - plex_connect's "plex_provider_id"/"mass_player_id" are not secrets, but its setup flow
 #   collects them (the options are only known from live server state), so they live in
 #   setup_data like any other flow-collected value.
-# - spotify is deliberately absent: its own _migrate_legacy_token still reads the legacy
-#   "refresh_token" and "client_id" from `values`, so this migration must not move them.
+# - spotify's deprecated "refresh_token" stays in `values`: _migrate_legacy_token reads it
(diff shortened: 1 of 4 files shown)


marcelveldt:
Thanks, @kiegsgroot !

kattcrazy:
Thank you for this fix 🙏

kiegsgroot:
No problem! It was getting real quite in my appartment haha

PR #6098, opened 2026-08-29
Add VRT MAX music provider

bollewolle:

Adds a new music provider for **VRT MAX**, the streaming platform of the Flemish
public broadcaster (VRT). It brings the VRT radio ecosystem into Music Assistant:

- **Live radio** — all VRT stations (Radio 1, Radio 2, Studio Brussel, MNM,
  Klara, VRT NWS, plus the digital-only channels) via their direct streams.
- **On-demand radio & podcasts** — listen-back shows and podcast series, browsable
  by station/programme and playable with seeking.
- **Played-songs tracklist** — radio listen-back episodes expose the songs played
  during the show as chapters in the Now Playing view.
- **Search** across stations, programmes and episodes.
- **Library sync** — VRT "Mijn lijst" favourites map to MA library podcasts, with
  add/remove written back to VRT.
- **Resume position** — playback progress is read from and written back to your
  VRT account, so listening continues across devices.

Authentication is optional: live radio works anonymously; on-demand playback,
favourites and resume require VRT account credentials configured on the provider.
The provider supports **multiple instances**, so family members can each connect
their own VRT account.

- n/a

## Types of changes

## Checklist

- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. _(n/a — no model changes)_
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. _(n/a — no UI changes)_
- [x] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. _(docs PR: music-assistant/music-assistant.io#1002, targeting `beta`)_


bollewolle:
For full transparency, this was written with Claude Code (Fable 5) under my guidance. I have some python knowledge, but it's more scripting than writing full blown apps/software/... For me the code looks good, it might be a different opinion for more knowledgeable developers, so if it turns out to be crappy then feel free to scrap this PR of course.
I have been running this for the past 1,5-2 weeks locally and it has been performing without any major issues, but it's a one sided test of course.

music_assistant/providers/vrt_max/constants.py
```diff
@@ -0,0 +1,15 @@
...
+BROWSE_PODCASTS = "podcasts"
+
+# Tile typenames that identify a program/podcast row on a landing page.
+RADIO_ROW_TYPE = "RadioProgramTile"
+PODCAST_ROW_TYPE = "PodcastProgramTile"
+
+# Only radio-archive episodes carry a played-songs tracklist (podcasts never do).
+# Their archives hold only a handful of recent broadcasts; the cap is a safety net
+# against a pathologically long archive triggering a burst of tracklist fetches.
+MAX_TRACKLIST_EPISODES = 50
```
OzGav: MAX_TRACKLIST_EPISODES is an invented ceiling of 50 to stop a runaway. Limit the concurrency properly, see the comment on provider.py line 535, and this stops being needed.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+    logo_url: str | None = None
+    tagline: str | None = None
+
+
+STATIONS: tuple[VrtStation, ...] = (
+    VrtStation(
+        id="radio1",
+        name="Radio 1",
+        stream_url="http://icecast.vrtcdn.be/radio1-high.mp3",
+        aac_url="http://icecast.vrtcdn.be/radio1.aac",
```
OzGav: `aac_url` is filled in for 20 of the 21 stations and never read anywhere. Only `stream_url` is used. Either wire it up or drop it.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+# Used with the Bearer token for user-scoped queries (Mijn lijst, progress) and
+# anonymously for everything else. A public-only mirror exists at
+# `.../vrtnu-api/graphql/public/v1` (no auth) - a fallback should VRT ever start
+# requiring auth on this endpoint for anonymous catalogue/playlist queries.
+GRAPHQL_URL = "https://www.vrt.be/vrtnu-api/graphql/v1"
+GRAPHQL_HEADERS = {
+    "Accept": "application/json",
+    "Content-Type": "application/json",
+    "x-vrt-client-name": "WEB",
+    "x-vrt-client-version": "1.5.15",
```
OzGav: x-vrt-client-version is pinned to 1.5.15 and will go stale. Add a note saying where the value comes from so the next person can refresh it.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+    "User-Agent": "Music Assistant",
+}
+GRAPHQL_TIMEOUT = aiohttp.ClientTimeout(total=25)
```
OzGav: The timeout is 25 seconds. Music Assistant only waits 8 seconds for a provider search, set in controllers/music/constants.py line 25. So on a slow response the provider carries on for another 17 seconds after it has already been given up on. A search needs a shorter timeout than the window it is given.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+# Per-user playback progress ("resume points"), keyed by the episode's aud media id.
+RESUMEPOINTS_URL = "https://ddt.profiel.vrt.be/resumePoints"
+# Snap the resume position to 0/end within this many seconds of the start/end.
+_RESUMEPOINT_MARGIN = 30
+# Upper bound on the in-memory resume-target cache (evicts oldest when exceeded).
+_RESUME_TARGET_CACHE_MAX = 512
+SSO_INIT_URL = "https://www.vrt.be/vrtnu/sso/login?scope=openid,mid"
+SSO_LOGIN_URL = "https://login.vrt.be/perform_login"
+# VRT's login endpoint expects a browser-like client.
+BROWSER_UA = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:137.0) Gecko/20100101 Firefox/137.0"
```
OzGav: Did you hit a failure that required this to be added?

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+class VrtApiError(Exception):
+    """Raised on a VRT API transport/protocol error (network, HTTP, bad payload)."""
+
+
+class VrtNotFoundError(VrtApiError):
+    """Raised when requested VRT content genuinely does not exist (empty page)."""
+
+
+class VrtAuthError(Exception):
+    """Raised when VRT authentication (SSO login / token exchange) fails."""
```
OzGav: The three custom exceptions inherit from plain `Exception` instead of the Music Assistant error types. Auth failures should be `LoginFailed`, missing content `MediaNotFoundError`, transport failures `ResourceTemporarilyUnavailable`. Doing this also lets most of the catch and re-raise code in `provider.py` go away.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+        """
+        Initialize the client.
+
+        :param session: Shared aiohttp session (use the MA http session).
+        :param logger: Logger for diagnostics.
+        """
+        self._session = session
+        self._logger = logger
+        # Cache of episode resume-point targets (stable per episode).
+        self._resume_targets: dict[str, VrtResumeTarget] = {}
```
OzGav: `get_resume_target` caches its result in a dictionary held on the client, with a hand-written size cap and oldest-first eviction. Your docstring is right that the target is stable per episode, so caching it is the correct instinct.

The catch is that it lives on the client rather than on the provider, and `@use_cache` needs a provider. So it has to carry its own cap and its own eviction, and it is discarded on every restart, meaning the first progress write-back for each episode after a restart pays for an extra request.

There is already a pattern for this in your own code. `_episode_chapters` keeps the decision on the provider and puts the fetch in `_fetch_chapters`, which is decorated with `@use_cache` and calls straight through to the client. Doing the same for the resume target would let you delete the dictionary at line 738, the `_RESUME_TARGET_CACHE_MAX` constant at line 252 and the eviction at lines 1031 to 1033, and it would survive a restart.

Low priority. Worth doing while the file is open rather than on its own.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+    async def iter_episodes(self, page_id: str) -> AsyncGenerator[VrtEpisode]:
+        """
+        Yield all listen-back episodes of a program/podcast, across every season.
+
+        :param page_id: The program/podcast page path.
+        """
+        program = await self.get_program(page_id)
+        for season in program.seasons:
+            async for episode in self.iter_season_episodes(season.component_id):
+                yield episode
```
OzGav: Not called from anywhere

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+
+    async def _ensure_login(self) -> None:
+        """Ensure a valid access + identity token, performing the SSO login if needed."""
+        if (
+            self._access_token
+            and self._identity_token
+            and time.time() < self._login_expiry - _TOKEN_EXPIRY_MARGIN
+        ):
+            return
+        jar = aiohttp.CookieJar()
```
OzGav: `_ensure_login` creates its own cookie jar and its own `aiohttp.ClientSession`. There is a helper for exactly this at helpers/aiohttp_client.py, and its docstring says it is for creating a session with cookies. Using it gives you the shared DNS resolver, the connection pool, SOCKS proxy support and the Music Assistant user agent, none of which this session currently has. It also means the login fails the same way as the rest of the provider when the network is unhappy, which it currently does not.

Change it to `create_clientsession(self.mass, cookie_jar=jar)` and pass the browser user agent on the two login requests that need it rather than as a session default. That needs mass on the auth manager, which is a small constructor change.

While you are there, the four places that set the user agent by hand to "Music Assistant", at lines 229, 915, 1069 and 1351, can all be deleted. A header set on a request overrides the session default, so those four are replacing the real user agent with a shorter one that has lost the version numbers. Delete them and the shared session supplies the correct one.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+    try:
+        payload = token.split(".")[1]
+        payload += "=" * (-len(payload) % 4)
+        claims = json.loads(base64.urlsafe_b64decode(payload))
+        exp = claims.get("exp")
+        if isinstance(exp, (int, float)):
+            return float(exp)
+    except IndexError, ValueError, binascii.Error, json.JSONDecodeError:
+        pass
+    return time.time() + 1800
```
OzGav: PyJWT is already a core dependency and is used by helpers/jwt_auth.py, so it is available to you.

Replace the body with `jwt.decode(token, options={"verify_signature": False})` and read exp from the result. That also removes the binascii import and the fallback that returns now plus thirty minutes when parsing fails, which is guessing at an expiry rather than failing.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+
+
+def _parse_iso(value: Any) -> datetime | None:
+    """Parse a VRT ISO-8601 timestamp (e.g. '2026-08-09T10:00:00.000Z')."""
+    if not isinstance(value, str) or not value:
+        return None
+    try:
+        return datetime.fromisoformat(value)
+    except ValueError:
+        return None
```
OzGav: Use the existing helper

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+
+    id: str
+    name: str
+    stream_url: str
+    aac_url: str | None = None
+    logo_url: str | None = None
+    tagline: str | None = None
+
+
+STATIONS: tuple[VrtStation, ...] = (
```
OzGav: Move this to constants.py

music_assistant/providers/vrt_max/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "type": "music",
+  "domain": "vrt_max",
+  "name": "VRT MAX",
+  "version": "0.0.1",
```
OzGav: Where has this come from? Same question for `dependencies` and `config_entries`

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+    from music_assistant_models.config_entries import ProviderConfig
+    from music_assistant_models.provider import ProviderManifest
+
+    from music_assistant.mass import MusicAssistant
+
+SUPPORTED_FEATURES = {
+    ProviderFeature.BROWSE,
+    ProviderFeature.SEARCH,
+    ProviderFeature.LIBRARY_PODCASTS,
+    ProviderFeature.LIBRARY_PODCASTS_EDIT,
```
OzGav, line 91: The supported features are a fixed constant, which causes three separate problems.

With no account configured, the podcast library features are still declared, so Music Assistant adds three sync switches to the settings that cannot do anything. One of them, "Sync library deletions", is on by default. A sync that finds nothing then tells Music Assistant that everything previously synced has gone, so if a user clears their password their synced VRT podcasts get removed on the next sync.

There is also no `supported_media_types`, so the provider never says it serves radio. That excludes it from single media type searches and from radio matching. Radio matching is what attaches your stations to a station the user already has from TuneIn or Radio Browser, so they end up with the broadcaster's own stream on the same item instead of a competing duplicate.

Make `supported_features` a property that includes the podcast library features only when a username and password are saved. filesystem_local does exactly this from a setting. Then add a `supported_media_types` property returning radio and podcast, following musicme, which is a provider in the same position as yours.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        return (
+            CONF_ENTRY_UNOFFICIAL_PROVIDER,
+            ConfigEntry(key=CONF_USERNAME, type=ConfigEntryType.STRING, required=False),
+            ConfigEntry(key=CONF_PASSWORD, type=ConfigEntryType.SECURE_STRING, required=False),
+        )
+
+    @property
+    def is_streaming_provider(self) -> bool:
+        """Return True for streaming providers."""
+        return True
```
OzGav: Delete as this is the same as the base class

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+
+    async def handle_async_init(self) -> None:
+        """Initialize the VRT MAX GraphQL client and (optional) auth manager."""
+        self._client = VrtMaxClient(self.mass.http_session, self.logger)
+        username = str(self.config.get_value(CONF_USERNAME) or "")
+        password = str(self.config.get_value(CONF_PASSWORD) or "")
+        self._auth = VrtMaxAuth(self.mass.http_session, self.logger, username, password)
```
OzGav: This stores the username and password without ever using them. A typo in the password is not discovered until the user tries to play something, and the message they get then is about logging in rather than about the password being wrong.

Call `get_access_token` here when credentials are present and raise `LoginFailed` if it fails.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        if MediaType.PODCAST in media_types:
+            podcasts: list[Podcast] = []
+            seen: set[str] = set()
+            for tile in await self._client.search_podcast_programs(query, limit):
+                if tile.page_id not in seen:
+                    seen.add(tile.page_id)
+                    podcasts.append(self._podcast_from_tile(tile))
+            # Radio archives are only searchable as episodes; fold them up to their
+            # parent programme so search returns the show, not individual broadcasts.
+            for episode in await self._client.search_radio_episodes(query, limit):
```
OzGav: Search makes two calls to VRT one after the other, at lines 155 and 161, so the second waits for the first. The client timeout is twenty five seconds, but Music Assistant only waits eight seconds for a provider search. So a slow response means the user sees nothing while the provider carries on working for another seventeen seconds.

Run the two with `asyncio.gather` and give search a timeout below eight seconds.

Separately, the radio half of this method needs no network at all, being a match against the station list already in memory, but it cannot be returned without waiting for both network calls. Consider returning it independently so live radio results are never held up by the podcast catalogue.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+                program_id = _program_id_from_episode(episode.page_id)
+                if program_id not in seen:
+                    seen.add(program_id)
+                    podcasts.append(self._podcast_base(program_id, episode.title))
+            results.podcasts = podcasts[:limit]
+
+        return results
+
+    @use_cache(3600 * 24)
```
OzGav: The method looks up a key in STATIONS_BY_ID, which is a dictionary in the code. Remove the decorator.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        """Get a single podcast / radio programme archive by page id."""
+        # Only a genuine "not found" maps to MediaNotFoundError; a transient
+        # VrtApiError propagates so the library sync aborts instead of pruning.
+        try:
+            program = await self._client.get_program(prov_podcast_id)
+        except VrtNotFoundError as err:
+            raise MediaNotFoundError(f"Podcast not found: {prov_podcast_id}") from err
+        return self._podcast_from_program(program)
+
+    async def get_podcast_episodes(self, prov_podcast_id: str) -> AsyncGenerator[PodcastEpisode]:
```
OzGav, line 225: Three things here.

Nothing in this method is cached. Every time a podcast is opened, and on every library sync, it fetches the programme page and then every season list, paginated at a hundred at a time. For a six season podcast that is seven or more requests, repeated every time. These lists change daily at most.

It also fetches the programme page a second time. `get_podcast` at line 179 fetched and cached that exact page moments before. Reuse it rather than calling `get_program` again.

The numbering at line 214 is backwards. VRT returns newest first, both within a list and across season tabs. I confirmed this against the API and it shows in the logs, where the queue worked through 29, 22, 15, 8 and 1 August in that order. Music Assistant expects the opposite, documented at helpers/podcast_parsers.py, where positions run oldest to newest so the newest episode has the highest number. Use `rank_episodes_by_date`, which is written for this and handles the no-dates case by assuming a newest-first provider, which is what you have.

Before relying on that, confirm the seasons never overlap in time, since the fix assumes the concatenated list runs strictly newest to oldest.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        try:
+            episode = await self._client.get_episode(prov_episode_id)
+        except VrtApiError as err:
+            raise MediaNotFoundError(f"Episode not found: {prov_episode_id}") from err
+        podcast_id = _program_id_from_episode(prov_episode_id)
+        podcast_mapping = ItemMapping(
+            media_type=MediaType.PODCAST,
+            item_id=podcast_id,
+            provider=self.instance_id,
+            name=episode.title,
```
OzGav: Should this be `program.title` ?

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        """Add a podcast to the user's 'Mijn lijst' (mirrors an MA library add)."""
+        if item.media_type != MediaType.PODCAST:
+            return True
+        return await self._set_favourite(item.item_id, favourite=True)
+
+    async def library_remove(self, prov_item_id: str, media_type: MediaType) -> bool:
+        """Remove a podcast from the user's 'Mijn lijst' (mirrors an MA library remove)."""
+        if media_type != MediaType.PODCAST:
+            return True
+        return await self._set_favourite(prov_item_id, favourite=False)
```
OzGav, line 278: Both return True for anything that is not a podcast rather than calling the parent, which does the right logging for an unsupported type. With no account they also return True while writing nothing anywhere, so Music Assistant believes the change reached VRT when it did not. The base class at `models/music_provider.py` logs when a change stays local. Do the same.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        media_item: MediaItemType,
+        is_playing: bool = False,
+    ) -> None:
+        """Write the playback position back to VRT ('resume point')."""
+        if media_type != MediaType.PODCAST_EPISODE or not self._auth.enabled:
+            return
+        try:
+            access_token = await self._auth.get_access_token()
+            target = await self._client.get_resume_target(prov_item_id)
+            total = target.duration or getattr(media_item, "duration", 0) or 0
```
OzGav: getattr(media_item, "duration", 0) sits after the media type has already been checked as a podcast episode, which always has a duration. Use media_item.duration so the type checker can see it.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+        except (VrtAuthError, VrtApiError) as err:
+            self.logger.warning("Failed to update Mijn lijst for %s: %s", page_id, err)
+            return False
+        return True
+
+    # ----------------------------
+    # Browse helpers
+    # ----------------------------
+
+    def _browse_root(self) -> list[BrowseFolder]:
```
OzGav, line 371: The menu always offers the radio programmes and podcasts folders. From what I can tell every listen back episode and every podcast episode is refused without an account, with a message saying the content requires authentication. Live radio is the only free part.

So with no credentials the user is offered several hundred programmes and not one of them can play.

Build these two folders only when a username and password are saved.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+            BrowseFolder(
+                item_id=BROWSE_PODCASTS,
+                provider=self.instance_id,
+                path=f"{self.instance_id}://{BROWSE_PODCASTS}",
+                name="Podcasts",
+                translation_key="podcasts",
+            ),
+        ]
+
+    async def _browse_landing(self, page_id: str, row_type: str, prefix: str) -> list[BrowseFolder]:
```
OzGav, line 402: Needs caching

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+                BrowseFolder(
+                    item_id=encoded,
+                    provider=self.instance_id,
+                    path=f"{self.instance_id}://{prefix}/{encoded}",
+                    name=row.title.strip(),
+                )
+            )
+        return folders
+
+    async def _browse_programs(self, component_id: str) -> list[Podcast]:
```
OzGav, line 421: Needs caching

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+            path=station.stream_url,
+            audio_format=AudioFormat(content_type=ContentType.try_parse("mp3")),
+            can_seek=False,
+            allow_seek=False,
+        )
+
+    async def _episode_stream_details(self, item_id: str) -> StreamDetails:
+        if not self._auth.enabled:
+            raise UnplayableMediaError(
+                "Log in with your VRT account in the provider settings to play on-demand audio."
```
OzGav, line 568: The message here is correct and so is the rule behind it. The problem is that the item was offered as playable at all, so Music Assistant only finds out when it asks for the stream, once per episode, until the queue runs out.

Alongside hiding the folders, set the provider mapping to unavailable, or `is_playable` to false, on podcasts and episodes when there is no account. Both fields exist on the models. The interface then greys them out.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+            duration=info.duration or None,
+            can_seek=True,
+            allow_seek=True,
+        )
+
+    async def _attach_tracklists(self, episodes: list[PodcastEpisode]) -> None:
+        """Concurrently attach the played-songs tracklist (as chapters) to radio episodes."""
+        targets = [ep for ep in episodes if _has_tracklist(ep.item_id)][:MAX_TRACKLIST_EPISODES]
+        if not targets:
+            return
```
OzGav, line 657: This starts up to fifty jobs at once with `asyncio.gather`. Each job runs `get_episode_chapters`, which makes one request for the episode menu, one for the playlist tab and at least one more for the paginated song list. So opening a single radio programme archive can put around a hundred and fifty requests to VRT in flight simultaneously. On a warm cache most of those are avoided, but the first visit to any programme pays the full cost, and there is no rate limiting anywhere in the provider.

Cap the concurrency with `TaskManager.create_task_with_limit`, and add a Throttler from to the client so no path can burst. With a real limit in place, MAX_TRACKLIST_EPISODES is no longer doing anything and can be removed.

Separately, worth deciding how much of the archive needs this at all. For podcast episodes the chapters are only rendered in the now playing view, so the current behaviour spends a lot of requests on something most users will not open. Fetching them for the most recent handful of episodes would give the same visible result for a fraction of the traffic.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+    return base64.urlsafe_b64encode(component_id.encode()).decode().rstrip("=")
+
+
+def _decode(encoded: str) -> str:
+    """Reverse _encode()."""
+    padded = encoded + "=" * (-len(encoded) % 4)
+    return base64.urlsafe_b64decode(padded.encode()).decode()
+
+
+def _program_id_from_episode(episode_id: str) -> str:
```
OzGav, line 717: This derives the parent programme by counting path segments and branching on whether the path starts with `/vrtmax/podcasts/`. That is correct for both of VRT's current URL shapes and I see it is covered by tests.

However, the path format comes from VRT rather than from Music Assistant, so VRT can change it. If they do, this returns a plausible-looking but wrong parent rather than failing, and the symptom will be episodes attached to the wrong podcast with nothing in the log to explain it.

Raise or log a warning when the path does not match either expected shape.

OzGav (changes requested):
Thanks for this new provider. There are a few things to fix which are below. 

In addition, neither `search` nor `browse` logs anything. When a user reports that results are missing, the log gives no way to tell whether the provider was called, what it was asked for, or what it returned.

One debug line in each, with the query or path and the number of results, would be enough.

Also helpers.py is too big so please split it up. This could be split up into, for example, models.py, api_client.py, auth.py, and parsers.py

OzGav:
Just marking as draft so we know which need attention. Please mark as ready for review when you want another look.

music_assistant/providers/vrt_max/provider.py
bollewolle: This was checked  against the live API before doing the switch. Radio programmes return exactly one season, so there's no overlap to worry about for those. Podcast seasons are listed newest-first (e.g. Seizoen 15, 14, 13…), and episodes within a season are newest-first too. The seasons also don't seem to overlap, e.g. "Onder ons: crimi" season 3 starts Aug 2026 while the previous season ends Sep 2025. So the concatenated list does run strictly newest to oldest. Now we are using rank_episodes_by_date, and the numbering is fixed.

music_assistant/providers/vrt_max/helpers.py
bollewolle: No, it was an incorrect assumption. We removed it and tested a real login with Music Assistant's own user agent and it works just fine. The four places that hand-set "Music Assistant" as the UA are removed as well, so the shared session now supplies the proper one.

music_assistant/providers/vrt_max/helpers.py
bollewolle: We've chose to actually wire it up. We did first check both endpoints: MP3 and AAC are both 128 kbps / 44.1 kHz stereo and both carry ICY metadata. we concluded that AAC is the better of the two at identical bandwidth. Live radio now prefers AAC. VRT NWS is the one station with no AAC URL, so MP3 stays as a fallback.

music_assistant/providers/vrt_max/provider.py
bollewolle: The two catalogue calls now run together via asyncio.gather under a 6s deadline, inside the 8s SEARCH_PROVIDER_SOFT_TIMEOUT. Radio is matched in memory first and is returned even if the catalogue call times out or fails, so it's never lost, however it can still be delayed by the deadline. Returning it truly independently isn't expressible while search() returns a single SearchResults; that'd need a streaming search API on the MA side from our understanding? But happy to follow up if you'd like it done that way of course.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+
+    async def get_library_podcasts(self) -> AsyncGenerator[Podcast]:
+        """Sync the user's 'Mijn lijst' favourites (podcasts + radio archives) to the library."""
+        if not self._auth.enabled:
+            return
+        try:
+            access_token = await self._auth.get_access_token()
+        except VrtAuthError as err:
+            self.logger.warning("Cannot sync VRT favourites: %s", err)
+            return
```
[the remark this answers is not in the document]
bollewolle: Confirmed that this is indeed an issue. A transient token-refresh failure returned an empty generator, which the sync reads as "no favourites", so the deletion pass would have deleted every previously synced podcast. The error now propagates and aborts the sync, with a regression test.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+            target_url = target.get("url")
+            if isinstance(target_url, str) and target_url:
+                hls_urls.append(target_url)
+        # Prefer the DRM-free variant; fall back to the first HLS url.
+        for hls_url in hls_urls:
+            if "_nodrm_" in hls_url:
+                return hls_url
+        if hls_urls:
+            return hls_urls[0]
+        raise VrtApiError("No HLS stream in aggregator response")
```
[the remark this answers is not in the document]
bollewolle: Also fixed this one. Falling back to hls_urls[0] could return a DRM-protected manifest that can never play. It now fails explicitly when no _nodrm_ variant is offered, which also makes the intent clear that this provider only ever plays the DRM-free rendition. Covered by three tests now as well.

music_assistant/providers/vrt_max/provider.py
```diff
@@ -0,0 +1,607 @@
...
+    ) -> None:
+        """Initialize the VRT MAX provider."""
+        super().__init__(mass, manifest, config, SUPPORTED_FEATURES)
+
+    async def get_config_entries(self) -> tuple[ConfigEntry, ...]:
+        """Return provider configuration entries (optional VRT account login)."""
+        return (
+            CONF_ENTRY_UNOFFICIAL_PROVIDER,
+            ConfigEntry(key=CONF_USERNAME, type=ConfigEntryType.STRING, required=False),
+            ConfigEntry(key=CONF_PASSWORD, type=ConfigEntryType.SECURE_STRING, required=False),
```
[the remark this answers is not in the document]
bollewolle: I don't think this one applies here. DEVELOPMENT.md documents this approach directly: config_entries with secure_string "(for passwords)", read via self.config.get_value(key), and SECURE_STRING values are already encrypted at rest, so moving to setup_data wouldn't add encryption. Credentials here are also deliberately optional: live radio works with no account, and a setup flow would force a dialog at add-time for people who only want the free stations.

music_assistant/providers/vrt_max/helpers.py
```diff
@@ -0,0 +1,1679 @@
...
+
+class VrtNotFoundError(VrtApiError):
+    """Raised when requested VRT content genuinely does not exist (empty page)."""
+
+
+class VrtAuthError(Exception):
+    """Raised when VRT authentication (SSO login / token exchange) fails."""
+
+
+class VrtMaxClient:
```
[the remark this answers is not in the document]
bollewolle: Fair point. I've added fixture-driven tests for the parsing layer, which takes parsers.py from 19% to 78%, 23 tests over the payload shapes including missing fields, wrong types, nested season tabs and scheduled broadcasts. The suite is now 102 tests, up from 58, and overall coverage 50% to 63%. HTTP-level tests for the SSO/token lifecycle aren't in this round; if wanted then feel free to let us know and we can add them as a follow-up.

bollewolle:
Thanks for the thorough review — all points are addressed.

helpers.py is now split into models, queries, parsers, api_client and auth, and the station table moved to constants.py. One deviation from your suggestion: I gave the GraphQL documents their own queries.py rather than putting them in api_client.py. They're about 340 lines, and folding them in would have made that file ~900, and would just be moving the size problem. Sizes now: models 125, queries 344, parsers 309, auth 190, api_client 567.

The rest: exceptions derive from Music Assistant's own error types, which let most of the catch-and-re-raise code to go away; features are gated now on configured credentials and supported_media_types now declares radio; episode ordering is fixed via rank_episodes_by_date; caching has been added for programme pages, episode lists, browse listings and the resume target; and a Throttler plus a TaskManager limit replace the ceiling that was incorrectly invented before.

Also added the search/browse debug logging like you asked. Tests are up from 58 to 102.

music_assistant/providers/vrt_max/auth.py
```diff
@@ -0,0 +1,190 @@
...
+                    info = await resp.json(content_type=None)
+                if not isinstance(info, dict) or info.get("errorCode") != 0:
+                    message = (info or {}).get("errorMessage") or "invalid credentials"
+                    raise VrtAuthError(f"VRT login failed: {message}")
+                redirect_url = info.get("redirectUrl")
+                if not redirect_url:
+                    raise VrtAuthError("VRT login returned no redirect url")
+                async with session.get(redirect_url, timeout=GRAPHQL_TIMEOUT) as resp:
+                    await resp.read()
+        except (aiohttp.ClientError, TimeoutError, ValueError) as err:
```
OzGav, line 141: Sorry, this one is my fault. I asked you to check the credentials during setup without checking how Music Assistant classifies a failure there.

The login code reports every failure as a login failure, including one caused by the network being unavailable. Music Assistant reads a login failure as a wrong password, and it deliberately never retries those, because a wrong password will not fix itself. So a brief network problem when the server starts now leaves the provider switched off until somebody reloads it by hand.

Please report only a rejected username or password as a login failure. A network or timeout failure should be reported as a temporary one instead, which does get retried automatically after a couple of minutes.

The setup check itself is right and worth keeping.

Same comment for line 158 below

OzGav (dismissed):

music_assistant/providers/vrt_max/auth.py
bollewolle: Not a problem. I've gone through every failure path in auth.py, not just the two you referenced: only "no credentials configured" and VRT actually rejecting the account stay a login failure now. Transport errors, timeouts and unexpected response shapes are all reported as temporary so they get retried. I also found that listing episodes only caught the login failure when fetching the progress token, so that would have broken on a network blip too, it now tolerates both. Added test_auth.py covering both directions. The setup check itself is unchanged.

OzGav (approved):
LGTM! Thanks @bollewolle !

PR #6101, opened 2026-08-29
Fix BBC Sounds station program metadata getting stuck from a stale cache

kieranhogg:

The BBC Sounds upstream library aimed to cache the result of station lookup to prevent a flurry of API hits on first load. This cache created two downstream bugs, so it was removed in auntie-sounds 2.0.7. This PR tidies up the provider code as a result, and adds a very short cache to prevent fetching the details every 5 seconds.

## Types of changes

## Checklist


OzGav:
The title needs reviewing and there are conflicts after your other PR was merged

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -413,12 +424,9 @@ async def _update_live_stream_metadata(
...
         else:
-            self.logger.debug(f"No song playing on {station_id}, fetching station info")
-            station = await self.client.stations.get_station(station_id)
-            if station:
-                stream_details.stream_metadata = await self._station_programme_display(
-                    station=station
-                )
+            self.logger.debug(f"No song playing on {station_id}, displaying current programme info")
+            programme = await self._station_current_programme(station_id)
+            stream_details.stream_metadata = await _station_programme_display(programme)
```
OzGav: Did dropping the `if station:` guard have a purpose I'm missing?

`_station_programme_display` returns `None` when the lookup fails, and line 415 bails out if `stream_metadata` is falsy. Since `playback_tracker.py` reuses the same `StreamDetails` for the whole item, I think one `None` freezes the display for the rest of the stream and `cache_none` defaults to True, so it sticks for 60s. Would keeping the guard cover it?

...plus `cache_none=False` on the decorator, so a transient failure retries?

music_assistant/providers/bbc_sounds/metadata.py
```diff
@@ -34,3 +34,10 @@ def _segment_to_metadata(now_playing: Segment) -> StreamMetadata | None:
             image_url = None
         return StreamMetadata(title=title, artist=artist, image_url=image_url)
     return None
+
+
+async def _station_programme_display(station: LiveStation) -> StreamMetadata | None:
```
OzGav: I don't think this needs to be async anymore?

music_assistant/providers/bbc_sounds/metadata.py
```diff
@@ -1,7 +1,7 @@
 """Metadata-related functions for BBC Sounds provider."""
 
 from music_assistant_models.streamdetails import StreamMetadata
-from sounds import Segment
+from sounds import LiveStation, Segment
```
OzGav: nit. `init.py` imports `LiveStation` from `sounds.models` it would be good to be consistent?

kieranhogg: Not a nit, I should really remove them from the main library module as VSCode defaults to that import.

music_assistant/providers/bbc_sounds/__init__.py
kieranhogg: Good catch, the intention was just to allow the None metadata to be set, but I didn't think through the logic with the early return. 

I'm not actually entirely sure what the best behaviour is here, I don't think any option is particularly great. Realistically the only scenario that would fail to fetch a valid programme is some form of API error which would likely result in an exception anyway, so I'll just add the guard back in.

music_assistant/providers/bbc_sounds/__init__.py
```diff
@@ -413,12 +420,10 @@ async def _update_live_stream_metadata(
...
-            self.logger.debug(f"No song playing on {station_id}, fetching station info")
-            station = await self.client.stations.get_station(station_id)
-            if station:
-                stream_details.stream_metadata = await self._station_programme_display(
-                    station=station
-                )
+            self.logger.debug(f"No song playing on {station_id}, displaying current programme info")
+            programme = await self._station_current_programme(station_id)
+            if programme:
+                stream_details.stream_metadata = _station_programme_display(programme)
```
OzGav:
```suggestion
            if metadata := _station_programme_display(programme):
                stream_details.stream_metadata = metadata
```

It looks like at the top of `_station_programme_display` you think it could return `None`, and since assigning None will stop your updater you should guard it here

OzGav (approved):
Thanks @kieranhogg !

PR #6109, opened 2026-08-30
Add static Zeroconf port option for go-librespot

bsod2b:

Adds a "Static Zeroconf port" setting to the Spotify Connect plugin's go-librespot engine. Today it always binds a random port for the Spotify Connect/mDNS advertisement on every restart, which is hard to reconcile with a firewall that needs a known port allow-listed in advance. The new setting (default 0 = today's random behavior, unchanged) lets you pin it to a fixed value instead.

- related [discussion](https://github.com/orgs/music-assistant/discussions/4648)

## Types of changes

## Checklist

- [X] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. music-assistant/music-assistant.io#1005

### DIFF
--- a/music_assistant/providers/spotify_connect/strings.json
+++ b/music_assistant/providers/spotify_connect/strings.json
@@ -53,6 +53,10 @@
         "very_high": "Very high (about 320 kbps)",
         "lossless": "Lossless (Spotify Soloist engine only)"
       }
+    },
+    "zeroconf_port": {
+      "label": "Static Zeroconf port",
+      "description": "The network port the Spotify app connects to when it discovers this device on your LAN. By default (0) go-librespot picks a random free port on every start, which strict firewalls (firewalld/ufw) and container setups using host networking may block. Set a fixed port here and allow it through your firewall instead."
     }
   },
   "setup_flow": {
--- a/music_assistant/providers/spotify_connect/README.md
+++ b/music_assistant/providers/spotify_connect/README.md
@@ -76,6 +76,7 @@ as `BackendEvent`s through a single async callback and answers `get_stream_sourc
 | Queue control | Session-side queue verbs (add-to-queue, shuffle, repeat) + queue/options events | Transport only |
 | Volume | Two modes: pin at 100% (default) or sync with compensation | `external_volume`: MA owns volume |
 | Risk profile | Binary downloaded from Spotify's CDN, 90-day build expiry, ToS grey area | May break when Spotify changes the protocol |
+| Static Zeroconf port | Not configurable | Configurable (`zeroconf_port`), for strict firewalls / host networking |
 
 The setup flow defaults new setups to Soloist and presents both engines as expanded
 choices. Existing pre-backend-split configs migrate to go-librespot. Loudness normalization,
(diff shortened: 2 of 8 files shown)


OzGav:
This wont work for more than one connected player which is one of the reasons why we say you can't put MA behind a firewall.

bsod2b:
@OzGav Hmm, you're right. Would a per-player port or a port range specification do it for you? 
The Connected players are already specified a few lines above so it'd just need to iterate over those.
If the concern is broader than this bug though let me know so I don't keep working on something you'd never want to ship.

music_assistant/providers/spotify_connect/strings.json
```diff
@@ -53,6 +53,10 @@
         "very_high": "Very high (about 320 kbps)",
         "lossless": "Lossless (Spotify Soloist engine only)"
       }
+    },
+    "zeroconf_port": {
+      "label": "Static Zeroconf port",
+      "description": "The network port the Spotify app connects to when it discovers this device on your LAN. By default (0) go-librespot picks a random free port on every start, which strict firewalls (firewalld/ufw) and container setups using host networking may block. Set a fixed port here and allow it through your firewall instead."
```
marcelveldt, line 59: how will this work with multiple instances ?
if you configure multiple players, multiple instances will be started.

marcelveldt:
> @OzGav Hmm, you're right. Would a per-player port or a port range specification do it for you? The Connected players are already specified a few lines above so it'd just need to iterate over those. If the concern is broader than this bug though let me know so I don't keep working on something you'd never want to ship.

The only thing I can think of is that you specify a base port and the per-player daemons just pick one, incrementing from that port. So basically it will be a range.

But, like @OzGav already mentioned, officially we simply do not support running MA behind a firewall, especially not the streaming part. The streamserver, as well as the inputsource plugins should be hosted on the same L2 network as the players/devices.

bsod2b:
I thought either that by just simply incrementing, or as you specify the players in the plugin too, or adding one UI element per player that let's you specify that port per player (or think of a new UI element that let's you match these in one single element).

For me, this was the only issue when trying to set it up behind a firewall I couldn't work around but I do realize that I have quite the simple setup and there might be more if it were more complex.

OzGav:
My question is why is MA behind a firewall? On a home LAN there's no host firewall configuration that both protects you and leaves MA working as the players connect to MA to fetch audio and to control it, over ports that protocols like AirPlay allocate at random. Blocking that traffic breaks playback and permitting it means you haven't restricted anything. Your router handles the boundary that matters.

bsod2b:
I live in a bulidng where one LAN is shared for every resident, so people I don't really know have access to local IPs. On the machine where my MA is running I am also hosting more sensitive stuff as well, so opening all ports feels wrong.

I guess the correct fix for this is to get my own router, I just thought I could save a bit of money with a simple fix.

marcelveldt:
Agreeing with @OzGav here - even if we would fix this by allowing some sort of range config - you will discover more issues of stuff that breaks.

bsod2b:
WIll close this then, thank you for your advice :)

PR #6110, opened 2026-08-30
Fix queue stalling after one track when current item is briefly unset

### LINKED ISSUE #6108: Queue with a multi-item media_id never enqueues a next track — playback stops after one song
# Queue with a multi-item `media_id` never enqueues a next track — playback stops after one song

## Summary

Calling `play_media` with a **list** of playlists (e.g. `media_id: [36, 28]`) loads the queue
correctly but never enqueues a next track. Playback stops after the first track and the queue
wedges permanently. The same playlists loaded **one at a time** work fine.

Root cause is a startup race: `track_loaded_in_buffer()` fires while `queue.current_index` is
still `None`, so the preload task hits its `if current_item is None: return` guard and exits
silently. Nothing ever re-arms it.

## Environment

| | |
|---|---|
| Music Assistant Server | 2.10.1 (schema 65) |
| Install | Home Assistant add-on (`d5369777_music_assistant`) |
| Home Assistant | 2026.8.3 |
| Player | SqueezeLite / piCorePlayer, output protocol reported as "native" |
| Music provider | Apple Music |

Reproduced on 2026-08-30, three times deterministically.

## Reproduction

1. Have a player with an active, playing queue.
2. Call the HA service `music_assistant.play_media` with a **list** media_id:

   ```yaml
   action: music_assistant.play_media
   data:
     media_id: [36, 28]      # two library playlists, 100 + 327 tracks
     media_type: playlist
     enqueue: replace
   target:
     area_id: dining_room
   ```

3. The first track plays to its exact duration, then the player goes `idle` and never advances.

**Expected:** the queue advances to track 2.
**Actual:** playback stops after one track. `repeat: all` and 427 queued items are ignored.

Loading a single playlist (`media_id: 36`) under otherwise identical conditions works correctly.

## Root cause

`track_loaded_in_buffer()` calls `_preload_next_item` unconditionally:

```python
# controllers/player_queues/controller.py:1396
def track_loaded_in_buffer(self, queue_id: str, item_id: str) -> None:
    ...
    current_index = self.index_by_id(queue_id, item_id)
    queue.index_in_buffer = current_index
    self.logger.debug("PlayerQueue %s loaded item %s in buffer", queue.display_name, item_id)
    self.signal_update(queue_id)
    self._preload_next_item(queue_id, item_id)          # line 1409
```

That reaches `_preload_streamdetails`, whose first statement is:

```python
# controllers/player_queues/stream_feeder.py
current_item = queue.current_item
if current_item is None:
    return  # guard
```

`queue.current_item` is derived from `queue.current_index`, and that index is set from whatever
track the *player* reports it is playing:

```python
# controllers/player_queues/playback_tracker.py:88-98
elif item_id := self._parse_player_current_item_id(queue_id, player):
    current_index = self.index_by_id(queue_id, item_id)
else:
    # this may happen if the player is still transitioning between tracks
    # we ignore this for now and keep the current index as is
    return False
queue.current_index = current_index        # assigns None with no guard
```

Immediately after a replace, the player is still reporting the **previous** queue's track. That id
is not in the new queue, so `index_by_id()` returns `None`, and it is assigned unguarded — note that
the sibling branch directly above deliberately protects against a transient state ("keep the current
index as is"), while this path does not.

So `queue.current_item` is briefly `None`, `_preload_streamdetails` trips its guard, and no next item
is ever preloaded or enqueued.

A multi-item `media_id` matters only because resolving several playlists takes ~10s instead of ~5s,
which widens the window in which the player is still reporting the replaced queue's track.

The early return is permanent: `current_index` is populated ~3s later, but nothing re-triggers
the preload.

### Evidence — the race window

Polling `player_queues/all` every 3s across the load. Note `current_index` is `None` at the exact
moment the buffer-load callback fires (`11:57:34.433`):

```
11:57:34 | playing | items 427 | current_index None | index_in_buffer 0 | current_item None
11:57:37 | playing | items 427 | current_index 0    | index_in_buffer 0 | Queen - Another One Bites the Dust
```

Healthy single-playlist control, same player, same session:

```
          | playing | items 403 | current_index 27   | index_in_buffer 27 | Colbie Caillat - I Won't
```

### Evidence — logs

Multi-playlist (broken). Resolving two playlists takes ~10s, and after the buffer load there is
no preload at all:

```
11:57:23.748 INFO  [player_queues.media_resolver] Fetching tracks to play for playlist Easy Rock
11:57:26.482 INFO  [player_queues.media_resolver] Fetching tracks to play for playlist Highly Rated Music
11:57:33.930 DEBUG [player_queues] (pre)loading (next) item for queue pCPDiningRoomDAC...
11:57:34.197 DEBUG [players] Starting playback on pCPDiningRoomDAC via native
11:57:34.433 DEBUG [player_queues] PlayerQueue pCPDiningRoomDAC loaded item 1f532842... in buffer
             <nothing further — no "Preloaded next item", no "Enqueued next track">
```

Single playlist (healthy), same player. The preload fires 1ms after the buffer load:

```
09:34:01.534 DEBUG [player_queues] PlayerQueue pCPDiningRoomDAC loaded item f1b182a5... in buffer
09:34:01.535 DEBUG [player_queues] (pre)loading (next) item for queue pCPDiningRoomDAC...
09:34:10.138 DEBUG [player_queues] Preloaded next item Phoebe Bridgers - Lost Boys for queue pCPDiningRoomDAC
09:34:11.144 DEBUG [player_queues] Enqueued next track Phoebe Bridgers - Lost Boys on queue pCPDiningRoomDAC
```

(Logs captured with the `player_queues` core logger at `VERBOSE`.)

### Note on a red herring

`prepare_next_audio_buffer` **does** still run and logs `Preparing audio buffer for next track X`,
so a next track appears to be getting ready. That is a different task from the preload that calls
`_enqueue_next_item`, so its success is misleading when diagnosing this.

## Results

| `media_id` | Start condition | Advances? |
|---|---|---|
| `36` (single) | from idle | yes |
| `33` (single) | from idle | yes |
| `33` (single) | replacing playing music | yes |
| `[36, 28]` | replacing playing music | **no — stops after 1 track** |
| `[36, 28]` | replacing playing music | **no — stops after 1 track** |
| `[12, 28, 21]` | replacing playing music | **no — stops after 1 track** |

Single-playlist loads succeed both from idle and while replacing active playback, so "replace while
playing" is not the trigger. The multi-item list is.

## Suggested fix

The guard treats a transient startup state as terminal. Two options:

1. **In `_preload_streamdetails` (what the linked PR does):** distinguish "no current item yet" from
   "queue drained". Before any current item has been observed, `None` is transient — keep waiting.
   After one has been observed, `None` still means the queue drained and the preload should stop.
   Self-contained and covered by a regression test.
2. **In `playback_tracker`:** do not overwrite `current_index` with `None` when the player reports an
   item that is not in the queue; keep the previous value, matching the sibling branch. This is closer
   to the true cause but `current_index` is read across the controller, so the blast radius is larger.

## Workaround

Load the first playlist alone, then append the rest:

```yaml
- action: music_assistant.play_media
  data: {media_id: 36, media_type: playlist, enqueue: replace}
- wait_template: "{{ is_state('media_player.x', 'playing') }}"
  timeout: "00:00:30"
- delay: {seconds: 15}          # let the preload + enqueue finish
- action: music_assistant.play_media
  data: {media_id: 28, media_type: playlist, enqueue: add}
```

Note that a following `shuffle_set: true` is a no-op if shuffle is already on
(`set_shuffle` returns early when the value is unchanged), so appended playlists stay at the end of
the queue. Toggling shuffle off then on forces `_apply_local_shuffle` to reorder the unplayed tail.

bcl79:

`_preload_next_item` gave up permanently when `queue.current_item` was `None`, which left the
player with no next track queued. Playback stopped at the end of the current track and the queue
wedged — 400+ items and `repeat: all` were ignored.

That `None` is transient. `queue.current_item` derives from `queue.current_index`, which
`playback_tracker` sets from whatever track the *player* reports. Immediately after a queue
replace the player is still reporting the previous queue's track, which maps to no index in the
new queue, so `current_index` is briefly `None`. The preload task hit its guard, returned
silently, and nothing re-armed it once the index populated a second or two later.

This distinguishes the two meanings of `None`:

- before any current item has been observed, it means "not populated yet" → keep waiting
- after one has been observed, it still means the queue drained → stop, as before

The retry budget now comes from the buffered item, which the caller already validates has a
duration, and those caller guards moved above the closure so the item is known non-`None` where
it's used.

It shows up most reliably with a multi-item `media_id` (e.g. `play_media` with a list of
playlists), because resolving several playlists takes noticeably longer and widens the window in
which the player is still reporting the replaced queue's track. Single-playlist loads set
`current_index` before the buffer-load callback fires and are unaffected.

Observed on 2.10.1 against a SqueezeLite player with Apple Music. With `player_queues` at
`VERBOSE`, a healthy load logs `loaded item ... in buffer` → `Preloaded next item` →
`Enqueued next track`; the failing load logs only the first and then nothing.

Note there is a plausible alternative fix a maintainer may prefer: guard `playback_tracker` so it
does not overwrite `current_index` with `None` when the player reports an item absent from the
queue, matching the sibling branch just above it that deliberately keeps the previous value. That
is closer to the true cause, but `current_index` is read throughout the controller, so I went with
the narrower change I could cover with a test. Happy to redo it that way if you'd rather.

- fixes https://github.com/music-assistant/server/issues/6108

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/player_queues/stream_feeder.py
+++ b/music_assistant/controllers/player_queues/stream_feeder.py
@@ -174,24 +174,42 @@ class StreamFeederMixin(_PlayerQueuesBase):
         """
         queue = self._queue_data[queue_id].queue
 
+        if not (buffered_item := self.get_item(queue_id, item_id_in_buffer)):
+            # this should not happen, but guard anyways
+            return
+        if buffered_item.media_type == MediaType.RADIO or not buffered_item.duration:
+            # radio items or no duration, nothing to do
+            return
+
         async def _preload_streamdetails(item_id_in_buffer: str) -> None:
             try:
                 # wait for the item that was loaded in the buffer is the actually playing item
                 # this prevents a race condition when we preload the next item too soon
                 # while the player is actually preloading the previously enqueued item.
-                current_item = queue.current_item
-                if current_item is None:
-                    return  # guard
-                retries = max(120, int(current_item.duration or 0) + 10)
+                retries = max(120, int(buffered_item.duration or 0) + 10)
+                seen_current_item = False
                 for _ in range(retries):
-                    # the queue can drain to empty while we sleep (e.g. all remaining
-                    # items skipped as unplayable); stop waiting once it has no current item
                     current_item = queue.current_item
                     if current_item is None:
-                        return
+                        if seen_current_item:
+                            # the queue drained to empty while we waited (e.g. all remaining
+                            # items skipped as unplayable): nothing left to preload
+                            return
+                        # the queue has no current item YET: right after a queue replace the
+                        # player still reports the previous queue's track, which maps to no
+                        # index in the new queue. that is transient, so wait for it to appear.
+                        # abandoning the preload here would leave nothing enqueued on the
+                        # player and playback would stop at the end of the current track.
+                        await asyncio.sleep(1)
+                        continue
+                    seen_current_item = True
                     if current_item.queue_item_id == item_id_in_buffer:
                         break
                     await asyncio.sleep(1)
+                if not seen_current_item:
+                    # the wait ran out without a current item ever appearing, so playback
+                    # never actually started - don't load (and enqueue) a next item for it
+                    return
                 if next_item := await self.load_next_queue_item(queue_id, item_id_in_buffer):
                     self.logger.debug(
                         "Preloaded next item %s for queue %s",
@@ -203,13 +221,6 @@ class StreamFeederMixin(_PlayerQueuesBase):
 
             except QueueEmpty:
                 return
-
-        if not (current_item := self.get_item(queue_id, item_id_in_buffer)):
-            # this should not happen, but guard anyways
-            return
-        if current_item.media_type == MediaType.RADIO or not current_item.duration:
-            # radio items or no duration, nothing to do
-            return
 
         task_id = f"preload_next_item_{queue_id}"
         self.mass.create_task(
--- a/tests/controllers/player_queues/test_stream_feeder.py
+++ b/tests/controllers/player_queues/test_stream_feeder.py
@@ -222,3 +222,166 @@ async def test_prepare_next_gives_up_softly_on_a_capacity_failure() -> None:
     await mass.create_task.call_args.args[0]()
 
     assert next_item.available
+
+
+async def test_preload_waits_for_a_current_item_that_is_not_set_yet(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """
+    Preload must wait out a briefly-unset current item instead of abandoning the queue.
+
+    Right after a queue replace the player still reports the previous queue's track, so
+    index_by_id() finds no index and current_item is transiently None. Giving up there leaves
+    nothing enqueued and playback stops at the end of the current track.
+    """
+    controller = PlayerQueuesController.__new__(PlayerQueuesController)
+    controller.logger = MagicMock()
+
+    buffered_item = _make_queue_item("q1", "buffered")
+    next_item = _make_queue_item("q1", "next")
+    queue = PlayerQueue(
+        queue_id="q1",
+        active=True,
+        display_name="Q1",
+        available=True,
+        items=2,
+        state=PlaybackState.PLAYING,
+        current_index=None,
+        index_in_buffer=0,
+        current_item=None,
+    )
+    controller._queue_data = {
+        "q1": PlayerQueueData(
+            queue=queue,
+            items=[buffered_item, next_item],
+            session_id="session-1",
+        )
+    }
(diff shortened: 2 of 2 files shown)


OzGav:
The PR description looks completely generated by AI.

Why did you not go with the last paragraph as that appears to be the actual fix. In `playback_tracker.py` at line 393 the code trusts the item id the player reports without checking it is still in the queue, unlike the two checks just below it, and that is what wipes the current item after a queue replace. Add that check and the `stream_feeder` changes and all three new tests can go, and the next item and buffer warm up get fixed as well.

PR #6124, opened 2026-08-31
Clean up sidebar shortcuts when a music provider is removed

OzGav:

When you remove and recreate a music provider (e.g. YT Music), stale references to the old provider instance ID cause "provider isn't currently available" popups on every page load.

This fixes the problem in two ways:

**Shortcut cleanup on provider removal:**
When a provider is removed, sidebar shortcuts that reference it are rewritten to point at the library copy of the item if one exists, or removed if it does not. Library shortcuts for items that no longer exist are also pruned after a sync finishes.

**Stale instance ID fallback in get_provider:**
When a stored reference uses an old instance ID that no longer exists, `get_provider` now extracts the domain from the ID and falls back to another instance of the same streaming provider. This handles all stored references (shortcuts, queue items, playlog) without needing to find and clean each one individually. Non-streaming providers are excluded from this fallback since their data is account-specific.

- Shortcut cleanup method on the auth manager
- Provider and library shortcut cleanup wired into the removal flow and sync completion
- get_provider handles deleted streaming provider instance IDs by domain extraction

- related issue https://github.com/music-assistant/support/issues/6256

## Types of changes

## Checklist


music_assistant/mass.py
```diff
@@ -593,6 +594,28 @@ def get_provider(
                 return prov
         return None
 
+    def resolve_provider_reference(self, provider_instance_or_domain: str) -> str:
```
marcelveldt: not sure this is the right fix.
so you pass in a specific instance id and then get returned a different one ?

OzGav: Yes from the same domain but only from a "streaming provider". So if you have YTM music then all the tracks will start working again when you add another YTM instance

marcelveldt: but it already does today because if you remove a provider, its links will be removed.
So who is calling us with a stale provider id then ?

I agree that this "fixes" it but in general its a workaround with a side effect that if somebody has multiple accounts for one provider, this workaround will steer into the wrong one without you knowing.

I would at least log a warning when this happens because this is basically a hidden "redirect my mail temporary after I was moved" situation

OzGav: The report was about builtin playlists, but when I looked into it those were actually fine, they just looked bad. That path already falls back from the instance to the provider name when it reads the file, so those entries still resolve and play.

What was actually failing is the sidebar shortcuts. As they are stored as URIs in the browser, the server never sees them and the cleanup cannot reach them. So the OP was seeing warnings in his log land a second after each browser connects, from three different devices, and it never settled down. The frontend drops a pinned item when the lookup comes back empty but not when it comes back as an error, which is this case.

So rather than extend the cleanup, which cannot reach the browser anyway, I went for making the lookup work. There is also a comment in the builtin playlist code saying entries are deliberately left in place and marked unavailable so the user can see what to replace, and I did not want to work against that.

On being steered to the wrong account, as I understand it that needs two accounts for the same provider and one of them not loaded, and I could not think of a reason that would happen on a normal system. I agree a warning is worth having so the user knows something unusual is going on, so I will add that.

What do you think?

MarvinSchenkel: I think the shortcuts are actually stored server side in user preferences. I think we should just make sure that all references are properly deleted when we delete a provider (instance) instead of this workaround.

OzGav: Yes quite right they are in user preferences.

It will be tricky to get right but sounds like I need to change tack and immediately after the removal of any music source do a scan of the shortcuts and remove any that don't resolve?

Looking at it I think there is a slightly wider bug that this will encompass and that is making sure if an item is removed from the library and it is a shortcut that it is cleaned up properly. Right now that only happens if a client is connected to catch the delete event, and it only cleans up that user's own pins. Deletions during a sync or a provider cleanup suppress the event entirely, so nothing gets cleaned.

OzGav: Reworked as you suggested. The redirect is gone. Shortcuts are now cleaned up server-side when a provider is removed, and stale library shortcuts are pruned after sync.

marcelveldt: Yes this fixes the root issue. We can always talk later if we want this auto redirect to another instance but that is more of a feature than a bugfix.

MarvinSchenkel: @OzGav are you able to drop the redirect from this PR so we can discuss it separately. That would scope this PR to the bugfix at hand

OzGav: @MarvinSchenkel The redirect from the original approach was already removed. What's in the PR now is shortcut cleanup. When a provider is removed, shortcuts pointing to it are either rewritten to their library copy if one exists, or removed if not. There's also a small fix in get_provider to extract the domain from a deleted instance ID so the domain fallback works. Is it one of those you want split out?

MarvinSchenkel (approved):
Looks good to me, thanks @OzGav 

This does need a follow up on the frontend though to be complete, we need to invalidate the userPreference cache in order to prevent old shortcuts being rewritten on a preference change. So, on PROVIDERS_UPDATED , refetch getCurrentUserInfo and replace store.currentUser.

PR #6144, opened 2026-09-01
Add Smart Fades-aware Smart Shuffle ordering

leonkdk:

Adds an optional Smart Fades-aware ordering pass to Smart Shuffle.

When enabled together with Smart crossfade, upcoming tracks are reordered using existing Smart Fades analysis for tempo, key and end-to-start energy. Recency and candidate selection stay authoritative, missing analysis is neutral, and the ordering pass never starts analysis or calls the full Smart Fades planner.

Dynamic refills are ordered only after Managed Pool has selected the batch, using the existing queue tail as the transition anchor. Close candidates keep some randomness so the result is not one fixed optimum.

Validated locally with 422 player queue tests and the full pre-commit suite.

- n/a

## Types of changes

## Checklist

Documentation PR: https://github.com/music-assistant/music-assistant.io/pull/1012


OzGav:
Please also open a docs PR

music_assistant/controllers/player_queues/controller.py
```diff
@@ -1567,8 +1567,13 @@ async def load(
             # a user-picked item must stay the one that plays, so hold it out of the shuffle
             pinned = next_items[:1] if pin_first else []
             shuffled = next_items[1:] if pin_first else next_items
+            # SFREQ-03: Keep MA's protected part of the queue fixed. Only the future part MA
```
MarvinSchenkel: Can we remove references to a requirements document that does not exist?

music_assistant/controllers/player_queues/smart_fade_ordering.py
```diff
@@ -0,0 +1,326 @@
...
+def _tempo_score(outgoing_bpm: float | None, incoming_bpm: float | None) -> float | None:
+    """SFREQ-06: Score tempo distance, including half/double-time matches."""
+    if not outgoing_bpm or not incoming_bpm or outgoing_bpm <= 0 or incoming_bpm <= 0:
+        return None
+
+    ratio = incoming_bpm / outgoing_bpm
+    diff_percent = (
+        min(
+            abs(1.0 - ratio),
+            abs(1.0 - ratio * 2.0),
```
MarvinSchenkel: Smart fades currently does not support half / double time tempo stretches, so let's score them the same as pairs we can actually beat match.

music_assistant/controllers/player_queues/smart_fade_ordering.py
```diff
@@ -0,0 +1,326 @@
...
+    from music_assistant_models.media_items import Track
+
+    from music_assistant import MusicAssistant
+    from music_assistant.models.audio_analysis import AudioAnalysisData
+
+ItemT = TypeVar("ItemT")
+
+# SFREQ-09: Keep this local. We only need a few sensible next choices, not a perfect route
+# through the whole queue.
+_LOOKAHEAD = 6
```
MarvinSchenkel: I think this is very sensible when the queue is in dynamic mode (and thus Tracks keep flowing in progressively). However, I think we should distinguish between the queue being in dynamic mode and a user just enqueueing a fixed playlist. In the latter case, we should consider the entire playlist as the population for re-sorting. This will result in considerably better sorting.

music_assistant/controllers/player_queues/strings.json
```diff
@@ -73,6 +73,10 @@
         "global": "Follow the Player Queues default."
       }
     },
+    "smart_shuffle_optimize_smart_fades": {
+      "label": "Optimize order for Smart Fades",
+      "description": "When Smart crossfade is active, reorder upcoming tracks so neighbouring tracks are more likely to support a musical blend. Uses only existing Smart Fades analysis; tracks without analysis remain neutral."
+    },
```
MarvinSchenkel, line 87: We should add 'options' and 'options_descriptions' here as well

music_assistant/controllers/player_queues/smart_fade_ordering.py
```diff
@@ -0,0 +1,326 @@
...
+        weighted += _TEMPO_WEIGHT * tempo
+
+    if (
+        outgoing.key is not None
+        and outgoing.mode is not None
+        and incoming.key is not None
+        and incoming.mode is not None
+    ):
+        # SFREQ-06: A key clash lowers the score; it does not remove the track.
+        key_score = (
```
MarvinSchenkel: Let's add a new helper to `smart_fades/helpers` to also capture relative keys etc. E.g.:

```
def camelot_affinity(
    key_a: str | None, mode_a: str | None, key_b: str | None, mode_b: str | None
) -> float | None:
    """
    Return a graded harmonic affinity between two keys on the Camelot wheel.

    1.0 same key, 0.9 one step or relative major/minor, 0.55 two steps in the
    same mode (energy boost), 0.45 one step with a mode switch, 0.1 for a clash.
    Returns None when either key is unknown so callers pick their own neutral.
    """
    a = _camelot_code(key_a, mode_a)
    b = _camelot_code(key_b, mode_b)
    if a is None or b is None:
        return None
    num_a, major_a = a
    num_b, major_b = b
    steps = min((num_a - num_b) % 12, (num_b - num_a) % 12)
    if major_a == major_b:
        if steps == 0:
            return 1.0
        if steps == 1:
            return 0.9
        if steps == 2:
            return 0.55
        return 0.1
    if steps == 0:
        return 0.9
    if steps == 1:
        return 0.45
    return 0.1
```

And then normalize that score according to your penalty/point scale. E.g. `_KEY_WEIGHT * (affinity * 1.5 - 0.5)`

music_assistant/controllers/player_queues/smart_fade_ordering.py
```diff
@@ -0,0 +1,326 @@
...
+        float(value)
+        for value in values
+        if isinstance(value, (int, float))
+        and not isinstance(value, bool)
+        and math.isfinite(value)
+        and value >= 0.0
+    ]
+    if len(clean) < 4:
+        return None
+
```
MarvinSchenkel, line 353: We should also guard for a silent tail here. `0.01` should be a good cutoff value

MarvinSchenkel (changes requested):
Really nice PR! Few comments, but all relatively small :)

music_assistant/controllers/player_queues/controller.py
leonkdk: Yep — removed all SFREQ references.

music_assistant/controllers/player_queues/smart_fade_ordering.py
leonkdk: Done — half/double time is no longer treated as a beat-matchable tempo.

music_assistant/controllers/player_queues/smart_fade_ordering.py
leonkdk: Done.
Fixed queues now consider the full movable population within each recency tier. Dynamic refills keep the bounded window.

music_assistant/controllers/player_queues/strings.json
leonkdk: Added both.

music_assistant/controllers/player_queues/smart_fade_ordering.py
leonkdk: Done.
Added camelot_affinity() to the shared Smart Fades helpers and use the graded score for ordering.

music_assistant/controllers/player_queues/smart_fade_ordering.py
leonkdk: Done.
Added the 0.01 silent-tail guard.

music_assistant/controllers/player_queues/smart_fade_ordering.py
MarvinSchenkel: Revisited this in fc6851aba: half/double time is now a discounted match (0.6) instead of a clash, so octave pairs cluster together without ever outranking a pair we can actually beat match.

music_assistant/controllers/player_queues/smart_fade_ordering.py
MarvinSchenkel: Took this one step further in e9fe521c9. With per-track features precomputed the window costs nothing to drop on the dynamic path either, so both modes now consider the whole run (a refill batch is at most 25 tracks anyway). That commit also fixes the O(N²) edge-energy recomputation the full window exposed: 11.6s → 0.14s for a 300-track queue.

music_assistant/controllers/player_queues/smart_fade_ordering.py
MarvinSchenkel: Replaced the cutoff with audible-bin filtering in ebe7d353d, so partially silent edges and silent intros measure the actual music instead of only catching the fully silent tail.

MarvinSchenkel (approved):
Awesome stuff, thanks @leonkdk 🎉

PR #6166, opened 2026-09-02
Add support to use loudness data from plex

chicco-carone:

Add support to the plex provider to use the calculated loudness data from the plex server during playback.
When adding music to plex if the user owns a plex pass on the server will calculate loudness data that can be later used by applications during playback. This pr makes music assistant aware of this data by adding it to music assistant so it can be used during playback instead of the live measurements that are used by default. If the server doesn't provide the data either for a singular track or for all of them it will fallback to using the user's configured loudness calculation for mass

## Types of changes

## Checklist


music_assistant/providers/plex/__init__.py
```diff
@@ -1014,6 +1017,14 @@ async def get_stream_details(self, item_id: str, media_type: MediaType) -> Strea
             can_seek=True,
             allow_seek=True,
         )
+        if loudness:
+            stream_details.loudness, stream_details.loudness_album = loudness
+            await self.mass.streams.audio_analysis.set_track_loudness(
```
MarvinSchenkel: Please use a task here

MarvinSchenkel:
I need to make a few changes in the core to accomodate for this PR, because we cannot simplfy hardcode the version to "2", as that will get outdated fast. Furthermore, I think your goal here is that loudness values from Plex are always used instead of MA doing the analysis?

chicco-carone:
The values from plex can't always be used reliably since if there are new tracks they will not get that data when they are added since the server takes a while to calculate it and also the server only calculates loudness data with a plex pass so if the user doesn't have one it should always fallback to the mass calculated one to allow playback correctly

MarvinSchenkel:
> The values from plex can't always be used reliably since if there are new tracks they will not get that data when they are added since the server takes a while to calculate it and also the server only calculates loudness data with a plex pass so if the user doesn't have one it should always fallback to the mass calculated one to allow playback correctly

Yep that will work out of the box, we will then check on track level whether an external loudness value exists (eg plex), if not, we calculate in MA and use that

chicco-carone:
Do i wait for #6188 to be merged and then update this one with the new field from that pr?

MarvinSchenkel (approved):
I see you already figured it out, nice work, looks good 🎉

PR #6169, opened 2026-09-02
Deezer: fix multiple instances sharing the same account

jdaberkow:

Deezer identifies a session by the sid cookie that gw-light.php sets on the first call, not by the arl. Both clients run on mass.http_session, which keeps one cookie jar for the whole server, so a second instance picked up the first one's sid and acted on the wrong account.

The gw client now sends its cookies with every request and keeps what deezer sets in its own dict instead of the shared jar. The sid defaults to empty, which makes deezer resolve the account from the arl again, so a sid left behind by another instance cannot take over.

GWClient.formats was a mutated ClassVar: one list for all instances, grown on every license refresh, leaking quality rights between accounts. It is per instance now.

- https://github.com/music-assistant/support/issues/6298

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/deezer/gw_client.py
+++ b/music_assistant/providers/deezer/gw_client.py
@@ -41,15 +41,14 @@ class GWClient:
     _license_expiration_timestamp: int
     _user_id: int
     session: ClientSession
-    formats: ClassVar[list[dict[str, str]]] = [
-        {"cipher": "BF_CBC_STRIPE", "format": "MP3_128"},
-    ]
+    formats: list[dict[str, str]]
     user_country: str
 
     def __init__(self, session: ClientSession, arl_token: str) -> None:
         """Provide an aiohttp ClientSession and the deezer ARL token."""
         self._arl_token = arl_token
         self.session = session
+        self.formats = [{"cipher": "BF_CBC_STRIPE", "format": "MP3_128"}]
 
     async def _set_cookie(self) -> None:
         cookie: Morsel[str] = Morsel()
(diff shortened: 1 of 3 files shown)


music_assistant/providers/deezer/provider.py
```diff
@@ -101,18 +104,25 @@ async def handle_async_init(self) -> None:
         """Handle async init of the Deezer provider."""
         arl_token = str(self.get_setup_value(CONF_ARL_TOKEN))
 
+        # Dedicated session with its own cookie jar to support multi-instance
```
marcelveldt: in general we don't allow providers to create their own http clientsession and everything needs to use the shared http session. You are confident you can't make this work otherwise ? For example to pass the cookiejar with every request ?

jdaberkow: From my understanding it is not possible to manage multiple cookie jars within one aiohttp session? So the other way would be to try to handle the cookie management manually for each session and then attach the cookies by hand.

I can give that a try.

jdaberkow: @marcelveldt 

You were right, works fine without an own session. The arl was never the problem, it's the sid that gw-light sets and the second instance just grabbed that one from the shared jar. Cookies passed per request do override the jar though, so I just send our own sid along now, empty by default so deezer falls back to the arl.

music_assistant/providers/deezer/gw_client.py
```diff
@@ -26,6 +25,7 @@
 )
 
 GW_LIGHT_URL = "https://www.deezer.com/ajax/gw-light.php"
+GET_URL_URL = "https://media.deezer.com/v1/get_url"
```
marcelveldt: nitpick: should this constant be GET_MEDIA_URL ? or MEDIA_GET_URL

marcelveldt (approved):
Nice work, @jdaberkow !

music_assistant/providers/deezer/gw_client.py
jdaberkow: True. MEDIA_GET_URL describes that way better. :+1:

PR #6200, opened 2026-09-04
Subsonic scrobbler: credit plays to the playing user's server account

pcc0x:

I run one Navidrome server with two accounts (one per person) and added it to MA twice, one OpenSubsonic instance per account, with each MA user filtered to their own instance. Plays from my wife's phone showed up in my Navidrome history, and the next track of the same album sometimes landed on a third account.

Cause: every library track has one provider mapping per instance, and the Subsonic scrobbler just picks the first opensubsonic mapping it finds. Since provider_mappings is a set, "first" is random.

Fix: the play report already has the userid, so the scrobbler now looks up that user's provider filter and prefers the mapping that is in it. If there is no user, no filter, or no matching instance it behaves exactly as before. Instances that are not loaded are skipped instead of hitting the assert. Plays started directly from an instance uri are untouched.

Added unit tests for the plugin (there were none): filter preferred, no user, filter without a subsonic instance, preferred instance not loaded, direct instance uri, and the full scrobble path.

- none, found it while setting up per user accounts

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue): `bugfix`

## Checklist


OzGav:
@Clusters

Clusters:
> @Clusters

Will look into it after the weekend 👍🏻

pcc0x:
@Clusters Please let me know if I should do something from my end! It's my first experience with Github. Thanks! Great application!

music_assistant/providers/subsonic_scrobble/__init__.py
```diff
@@ -119,19 +126,39 @@ async def _get_subsonic_provider_and_item_id(
...
-                    if isinstance(library_item, PodcastEpisode) and EP_CHAN_SEP in mapping.item_id:
-                        _, ret_id = mapping.item_id.split(EP_CHAN_SEP)
-                    else:
-                        ret_id = mapping.item_id
-                    return prov, ret_id
+            sonic_mappings = [
+                mapping
+                for mapping in library_item.provider_mappings
+                if mapping.provider_domain.startswith("opensubsonic")
+            ]
```
Clusters, line 133: return early if sonic_mappings is empty. This saves us the __get_user_provider_filter()_ call


```suggestion
            ]
            if not sonic_mappings:
                return None, item_id
```

Clusters (changes requested):

Clusters:
Hi @pcc0x, thanks for your fix and PR.
I think this got relevant after the scrobbler plugin got updated last. So this is a very welcome addition.

I looked through the code and it is sound.
Just that small note in my review.

@OzGav do we want test coverage for the two branches that aren't covered yet: The EP_CHAN_SEP episode-id split in the provider selection loop, and the fallback for user_id set but get_user returning None (deleted/disabled user, should keep the previous any-instance behavior)?

OzGav:
Since tests are being added here then I think yes lets improve the coverage. @pcc0x please do that and address Andreas' comment as well as resolve the copilot comments.

pcc0x:
Done. I added the early return you suggested (no Subsonic mappings, no user lookup) and five tests for the branches that were not covered: the episode id split for library and provider uris, a deleted user, a user with an empty filter, and a library item without Subsonic mappings. Copilot threads resolved. Thanks both for the review.

Clusters (approved):

Clusters:
@pcc0x @OzGav looks good to me 👍

Linting must be resolved, after that we should be good to go.

OzGav:
@pcc0x As per the checkbox at the top please always run pre-commit so you don't have these failures.

MarvinSchenkel (approved):
Thanks @pcc0x 🙌

PR #6235, opened 2026-09-09
Keep Sonos S1 playback events from blocking the event loop

OzGav:

Starting a radio station on a Sonos S1 speaker could stall for several seconds and then fail, while library tracks played fine.

When the speaker reported a playback change, Music Assistant asked it for the current track details right away, but did so in a way that froze the server while waiting for the answer. On a radio start the speaker only answers once it has connected to the stream, and that stream is served by the very server that was frozen. Both sides waited on each other until the speaker gave up.

The speaker query now runs in the background so the server keeps serving the stream, and playback events are still applied in the order they arrived.

- Handle Sonos S1 playback events without blocking the server
- Keep a burst of events applied in order
- Test that the speaker query no longer runs on the main thread

Tested by the original reporter using the DEV server

- related issue https://github.com/music-assistant/support/issues/6360

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/sonos_s1/player.py
+++ b/music_assistant/providers/sonos_s1/player.py
@@ -114,6 +114,7 @@ class SonosPlayer(Player):
         self._subscriptions: list[SubscriptionBase] = []
         self._subscription_lock: asyncio.Lock = asyncio.Lock()
         self._avtransport_event_lock: asyncio.Lock = asyncio.Lock()
+        self._avtransport_tasks: set[asyncio.Task[None]] = set()
         self._last_activity: float = NEVER_TIME
         self._resub_cooldown_expires_at: float | None = None
         self._poll_task_id: str = f"sonos_poll_{self.player_id}"
@@ -159,6 +160,9 @@ class SonosPlayer(Player):
         # the poll runs as a task under the same id and cancel_task is what stops it
         self.mass.cancel_timer(self._poll_task_id)
         self.mass.cancel_task(self._poll_task_id)
+        # playback events still queued on the lock must not query the speaker any more
+        for task in self._avtransport_tasks:
+            task.cancel()
         # unsubscribe directly: offline() skips a speaker that is already marked
         # unavailable, which would leave its subscriptions behind. The lock keeps a
         # subscribe() that is still in flight from re-populating them afterwards.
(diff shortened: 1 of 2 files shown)


music_assistant/providers/sonos_s1/player.py
```diff
@@ -717,7 +718,7 @@ def _handle_event(self, event: SonosEvent) -> None:
             self.update_player()
             return
         if service_type == "AVTransport":
-            self._handle_avtransport_event(event)
+            self.mass.create_task(self._handle_avtransport_event(event))
```
MarvinSchenkel: [PROBLEM] These AVTransport tasks are not cancelled when the player unloads. Events waiting on the lock can still start speaker queries after removal; the existing `update_player()` guard only prevents publishing the results.

Please keep references to these tasks, remove completed ones, and cancel the remaining handlers in `on_unload()`. Also check `_unloaded` inside `_avtransport_event_lock` before querying the speaker, and retain the guard against publishing results from an already-running worker thread. Avoid a single shared `task_id`: `mass.create_task()` deduplicates it and would drop subsequent events.

tests/providers/sonos_s1/test_player.py
```diff
@@ -224,16 +224,45 @@ def test_settled_state_restores_the_poll_interval(sonos_player: SonosPlayer) ->
...
-    sonos_player._handle_avtransport_event(event)
+    await sonos_player._handle_avtransport_event(event)
 
     assert sonos_player.poll_interval == TRANSITION_POLL_INTERVAL
 
 
+async def test_avtransport_event_queries_the_speaker_off_the_event_loop(
+    sonos_player: SonosPlayer,
+) -> None:
+    """Fetching track info for a playback event must not stall the event loop."""
```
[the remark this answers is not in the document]
OzGav: Not adding this. The lock is a plain asyncio lock, its ordering is guaranteed by Python itself. The test covers the actual fix, the query running off the loop.

music_assistant/providers/sonos_s1/player.py
OzGav: Done. Tasks are tracked per player and cancelled on unload, and a handler that only gets the lock after unload bails before touching the speaker. Publish guard unchanged.

marcelveldt (approved):
Thanks, @OzGav !

PR #6240, opened 2026-09-09
Update Yoto API dependency with Oauth refresh token fixes

pantsman0:

This PR updates the yoto-api package to v4.4.1. This lets a user register a refresh hook to notify a `YotoClient` holder that the refresh token has been updated.
This also includes and update which locks auth token usage during updates to prevent a race condition resulting in deauthenticated tokens being used after being refreshed.

- https://github.com/music-assistant/server/pull/5584#pullrequestreview-5072481253

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/yoto/__init__.py
+++ b/music_assistant/providers/yoto/__init__.py
@@ -39,6 +39,7 @@ from music_assistant_models.media_items import (
 from music_assistant_models.streamdetails import MultiPartPath, StreamDetails
 from yoto_api import Card as YotoCard
 from yoto_api import Chapter as YotoChapter
+from yoto_api import Token as YotoToken
 from yoto_api import YotoAPIError, YotoClient, YotoError
 
 from music_assistant.models.music_provider import MusicProvider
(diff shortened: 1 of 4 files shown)


pantsman0:
@OzGav Hopefully this should fix the potential race condition when updating the refresh token.

music_assistant/providers/yoto/__init__.py
```diff
@@ -660,3 +656,7 @@ async def _parse_radio_stations(self, card: YotoCard) -> AsyncGenerator[Radio]:
                 )
             else:
                 self.logger.warning(f"No tracks found for radio station {station.key}")
+
+    async def _refresh_hook(self, token: YotoToken) -> None:
+        """Store updated refresh token."""
+        self._update_setup_data(CONF_REFRESH_TOKEN, token.refresh_token)
```
OzGav: The refresh token on the `Token` object can be `None`, and the code this replaces checked for that before saving. Without the check, a response that comes back without a refresh token will overwrite the saved one with nothing. Next time the provider starts it finds no token and raises `LoginFailed`, which MA does not retry, so the provider stays down and the user has to go back to the Yoto dashboard and set it up again. That is the same thing this PR is trying to avoid, just arriving a different way. Please put the check back so we only save a real value.

pyproject.toml
```diff
@@ -47,7 +47,7 @@ dependencies = [
   # TEMPORARY test-drive pin — replace with PyPI release before merge (see webrtc_aiolibdatachannel_plan.md Phase C)
   "aiolibdatachannel @ git+https://github.com/music-assistant/aiolibdatachannel@feat/certpem-testdrive",
   "pyjwt[crypto]>=2.10.1",
-  "yoto-api==4.3.4",
+  "yoto-api==4.4.1",
```
OzGav: While you are in here, can `yoto-api` come out of this list altogether. This is the core dependency list so everyone installs it whether they use Yoto or not, and no other provider library is in here. It is already listed in the Yoto manifest, which is where provider requirements go, and `requirements_all.txt` is built from the manifests. My fault for not picking this up when the provider PR went through.

OzGav (commented):
Good jobn getting the upstream library fixed. Just a couple of things here,

pyproject.toml
pantsman0: it has been removed.

music_assistant/providers/yoto/__init__.py
pantsman0: Personally I don't think this is a problem right now as we don't call `device_code_flow_complete` in MA server, the only times the refresh hook is called is after a previous refresh token has been consumed. If we decide to update to `None` we will just be keeping an invalidated refresh token anyway.

I have still made the change though, as updates to yoto_api may change internal behaviour and the change doesn't result in any worse UX here.

OzGav (approved):
Thanks @pantsman0 !

PR #6261, opened 2026-09-10
Fix ariacast sticky active player on second session and beyond

arthurbre:

- Fix AriaCast receiver leaving players idle after an abrupt sender disconnect
  The /audio WebSocket handler only flipped its local is_playing flag when the
  sender vanished without a graceful is_playing=false (e.g. app killed outright, a
  network drop) instead of running the same teardown a normal stop uses. The
  owning queue/group was never told to release the source, so it stayed
  claimed and a later reconnect, or a future session, never retriggered playback, leaving the
  player(s) stuck idle, with the same symptoms as the sticky-leader bug fixed
  in the previous commit.
  Route the abrupt disconnect through _handle_playback_state(False), the same
  path a graceful stop already takes, so the source is always released
  properly regardless of how the sender goes away.

- Fix AriaCast receiver playing a group leader alone on the second (and further) session
  on_source_selected stored the protocol player actually consuming the stream
  (e.g. a sync group's leader) into _active_player_id, which also doubles as
  the play target for the next session. Once a group had formed once, every
  later session skipped the group and replayed on that one leaf player,
  silently leaving the rest of the group's members without audio.
  Track the stream-consuming player separately (_session_player_id) so it no
  longer leaks into the next session's target, and use it (falling back to
  _active_player_id) for cmd_stop on pause.

I didn't find any issue describing this.

## Types of changes

## Checklist

> I played several sessions from several senders across two days.

### DIFF
--- a/music_assistant/providers/ariacast_receiver/__init__.py
+++ b/music_assistant/providers/ariacast_receiver/__init__.py
@@ -116,6 +116,13 @@ class AriaCastReceiver(PluginProvider):
 
         # MA stream-routing state
         self._active_player_id: str | None = None
+        # The protocol player actually consuming the current stream (e.g. the
+        # sync leader MA picked for a group), used to target cmd_stop for the
+        # live session only. Kept apart from _active_player_id — the play
+        # target used to (re)start the *next* session — so a leaf picked by
+        # grouping never becomes a future session's play target and leaves
+        # the rest of a sync group unplayed.
+        self._session_player_id: str | None = None
         self._in_use_by_player: str | None = None
         self._active_session_id: str | None = None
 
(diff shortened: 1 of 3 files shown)


music_assistant/providers/ariacast_receiver/__init__.py
```diff
@@ -116,6 +116,13 @@ def __init__(
 
         # MA stream-routing state
         self._active_player_id: str | None = None
+        # The protocol player actually consuming the current stream (e.g. the
+        # sync leader MA picked for a group), used to target cmd_stop for the
+        # live session only. Kept apart from _active_player_id — the play
+        # target used to (re)start the *next* session — so a leaf picked by
+        # grouping never becomes a future session's play target and leaves
+        # the rest of a sync group unplayed.
```
OzGav, line 124: These AI generated comments are way too verbose. Comments are only needed where the next dev won't understand the code.

OzGav:
@AirPlr

AirPlr:
I approve of this changes, but the comments need to be shorter

arthurbre:
Thank you both. I cleaned vebose comments while trying to keep sufficient info level.

OzGav (approved):
Thanks @arthurbre !

PR #6271, opened 2026-09-11
Fix Sendspin metadata during group content takeover

teancom:

I have a 'screen' device in a static sendspin group alongside a WiiM, and I noticed that I would often see, at the start of playback of the group, a "flash" (or sometimes several seconds) of the metadata and album art of the WiiM's *individual* player from whatever it had played last. Then the metadata and artwork of the _group_ player would come through and overwrite it, fixing the glitch. I thought it was going to be relatively simple but a few days of testing showed there to be a ton of edge-case and race conditions to account for. Which is how I ended up with this including "generations" and serializing publication so we don't flap the state back and forth.

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/controllers/players/controller.py
+++ b/music_assistant/controllers/players/controller.py
@@ -3492,12 +3492,14 @@ class PlayerController(AnnouncementsMixin, AudioSourceMixin, ProtocolLinkingMixi
             return False
         return True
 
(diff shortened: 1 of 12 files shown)


teancom:
Feedback addressed, thank you @OzGav

PR #6272, opened 2026-09-11
Keep the MilkDrop visualizer alive on tracks longer than the audio buffer

jozefKruszynski:

On a track longer than the audio buffer the MilkDrop tap was forced to read at the eviction edge, fanned frames out tens of seconds early, and sent `stream/clear` every time eviction overtook it, which on players that fetch in gulps is every gulp, so viewers went dark for most of the track. The tap now reads everything the buffer retains as soon as it holds it, keeps frames in a bounded pending queue, and releases each one five seconds before it is audible, so every player gets the same lead. Falling behind the retained window catches the cursor up on the same timeline instead of resetting, which leaves `stream/clear` for track change, seek, speed change and stop. The realign workaround in the relay is removed with it.

- related issue https://github.com/music-assistant/support/issues/6392

## Types of changes

## Checklist


music_assistant/providers/milkdrop_visualizer/tap.py
```diff
@@ -414,10 +387,9 @@ async def _read_once(self, tap: Tap, cursor: TrackCursor | None) -> TrackCursor
             await asyncio.sleep(IDLE_POLL_SECONDS)
             return cursor
         except AudioBufferDiscarded:
-            # the retained window moved past us (a stalled tap, or a rolling
-            # buffer outrunning it); pick the timeline up again where it is now
+            # the retained window moved past us; _align catches up next pass
             await asyncio.sleep(IDLE_POLL_SECONDS)
-            return None
+            return cursor
```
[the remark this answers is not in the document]
jozefKruszynski: Not a stall: a replacement buffer for the same item keeps media-time chunk numbering, so the cursor stays valid. A seek re-anchors as before; at the same position the cursor waits for the new window to reach it, which it does as playback advances, while the pending queue keeps feeding viewers. A cancelled buffer that is never replaced just retries each poll, as the old code did, minus a clear per retry.

6181d5651 adds a test for exactly that path.

music_assistant/providers/milkdrop_visualizer/tap.py
```diff
@@ -65,17 +63,13 @@
...
-# follows the player's stream pull (readrate 2x for HTTP players, ~30s commit
-# lead for Sendspin), so the tap is forced to read - and stamp - audio well
-# ahead of the audible playhead. The ring bridges that gap for attaching
-# viewers: ~95s at ~43 frames/s of ~1KB each (~4MB per tap). Beyond it the
-# audio is already evicted server-side, so no ring size can help; long tracks
-# spend their pinned phase there and that is an accepted limitation.
-RING_FRAMES = 4096
+# Frames replayed to a viewer that attaches mid-track (~12s), and the ceiling
+# on one viewer's outbound queue.
+RING_FRAMES = 512
```
MarvinSchenkel: [PROBLEM] At 192 kHz, 512 frames cover only 2.73 seconds, less than the five-second release lead. A second viewer joining an established tap therefore receives only future frames and waits roughly two seconds before getting a drawable waveform; faster playback makes this worse. Please size the replay window to cover the release lead plus a margin, accounting for sample rate and playback speed, and add coverage for a second viewer joining during high-rate playback.

jozefKruszynski: Good catch, the count was the wrong unit. 5542969a5 sizes the ring by time instead: on each release pass it keeps frames from one second behind now through the release lead, using the frame timestamps, so a joining viewer always gets a drawable frame plus the full lead at any sample rate or playback speed. Two tests cover it, including a simulated 200 frames/s join.

music_assistant/providers/milkdrop_visualizer/tap.py
```diff
@@ -65,17 +63,14 @@
...
-# spend their pinned phase there and that is an accepted limitation.
-RING_FRAMES = 4096
+# Ceiling on one viewer's outbound queue.
 VIEWER_QUEUE_FRAMES = 1024
+# How far behind now the ring keeps released frames. An attaching viewer draws
+# the newest past frame at once, and the rest of the ring is the lead ahead.
+RING_PAST_SECONDS = 1.0
+# Ceiling on frames held back before they are due (~3 min, ~8MB per tap); past
+# it the tap stops reading and accepts a hole.
+PENDING_FRAMES = 8192
```
OzGav, line 74: Quick question on PENDING_FRAMES: this is a fixed number of frames, like the ring was before Marvin's comment. Does the same sample-rate issue apply here, since 8192 frames is a lot less than 3 minutes at 192 kHz?

jozefKruszynski: Fair question, but this one is a frame count on purpose. The ring had to span a fixed length of time, so a count was the wrong unit there. PENDING_FRAMES bounds memory, and a frame is the same size at any rate, so 8192 is ~8MB per tap regardless. What shrinks at 192 kHz is how far ahead the tap reads before pausing, from ~3 min to ~44 s (or ~22 s at 2x), and that only has to stay above one fetch burst plus the 5 s lead. The largest burst measured was 8 s. 5bbeedfbe makes the comment say so.

OzGav (approved):
Thanks for the quick turnaround on this one!

PR #6326, opened 2026-09-13
Keep retrying YouTube Music when the PO Token server is not up yet

CodeCommander:

The YouTube Music provider raises LoginFailed when the PO Token server does not answer its ping during handle_async_init. Since #5599 the core no longer schedules a retry for LoginFailed (it waits for the user to fix their credentials), so on a host reboot where the PO Token add-on/container starts after Music Assistant — which is the normal order on my Synology — the provider stays unloaded until it is reloaded by hand, and the UI reports "Login failed. Please check your credentials".

An unreachable PO Token server does not mean the credentials are wrong. The proposed fix raises SetupFailedError instead (which the core retries) and returns a more accurate error message from the provider.

Related: #6280 touches the same method and makes the Premium-check stream failure retryable, but keeps LoginFailed on the ping — this is the other half of that boot race. Same translation key used so either order rebases cleanly.

## Types of changes

## Checklist


MarvinSchenkel (approved):
Thanks @CodeCommander 🙏

PR #6327, opened 2026-09-13
Split library artists a streaming provider knows under several ids

jozefKruszynski:

Follow-up to https://github.com/music-assistant/server/pull/6324, stacked on that branch. Libraries that merged same-name artists before that fix still hold one library artist with several mappings on one streaming instance, and the relation tables only record the library artist id, so the split has to be rebuilt from the provider. The artists controller runs a one-off repair after the first completed sync: it finds such artists, keeps one mapping per instance unless a strong external id links them, removes the rest, re-fetches the artist's albums and tracks so their credits land on the right library artist, and records a done marker in the library database so it does not run again. A `music/split_merged_artists` API command runs it by hand. Providers whose ids are not stable, such as the filesystem providers, are left alone.

**This needs discussion before it goes anywhere.** Open points: whether an automatic one-off run after the first sync is acceptable or only the manual command should exist, how the kept mapping is chosen when both provider artists carry the same name, the cost of re-fetching every album and track of an affected artist, and that the done marker is written even when a provider was offline during the run, which leaves the manual command as the only retry.

- related issue: none

## Types of changes

## Checklist


jozefKruszynski:
Closing: an automated split is not safe. Checked against a real library, a streaming provider can list one artist under two ids with an identical name (Tidal does this for GUNSHIP and Seeb), so this pass would turn those into duplicate library artists. Details in the status note on #6324.

PR #6339, opened 2026-09-14
Fix auto-merge getting stuck on dependency updates that need a newer Python

MarvinSchenkel:

The auto-merge workflow for dependency bumps got stuck on `music-assistant-models` updates. Its "Wait for package availability on PyPI" step verifies a release with `pip download`, which runs on the runner's default Python (3.12). Version 1.1.212 raised its required Python to `>=3.13`, so `pip download` kept refusing it even though the package was published and fine, and the step retried for 20 minutes and then failed, blocking every models bump.

The check only needs to confirm the release exists on PyPI, not that it installs on the runner, so it now ignores the required-Python metadata.

- n/a

## Types of changes

- [x] CI / workflow change — `ci`

## Checklist


MarvinSchenkel:
Re Copilot's "cp313-only wheel could still fail" note: doesn't apply here. Both `music-assistant-models` and `music-assistant-frontend` ship pure-python `py3-none-any` wheels, so `--ignore-requires-python` is enough to fetch them on any runner Python. Parsing a file URL out of the JSON would be more than this check needs.

Also fixed the indentation from the applied suggestion (it dedented the `if` out of the `run:` block), and the flag is now covered by the workflow contract test.

PR #6345, opened 2026-09-14
Stop a Sendspin player reporting a position ahead of what it is playing

caraar12345:

After a producer stall, a Sendspin player reports a playback position ahead of the audio actually leaving the speakers, and stays wrong for the rest of the stream. Everything downstream of that position is off by the same amount.

`SendspinPlaybackSession._timeline_start_us` is the server-clock time at which flow-stream audio position 0 renders. It was assigned on the first commit and never updated — but aiosendspin rebases its scheduled play timeline forward whenever audio production stalls (`aiosendspin` 9.1.1, `server/push_stream.py`, `_resolve_channel_play_start`). The schedule moves; the anchor didn't.

I got here through debugging an issue with a custom live lyrics card I've got in my HA instance wherein music played through MA is often a second or two behind. Wanted a plugin for MA to give the card a true time anchor from Sendspin itself, but in scoping that out, found that MA had a bug itself that would contribute to the issue later down the line.

**Fix** — recompute the anchor on every commit rather than only the first:

```python
self._timeline_start_us = int(commit_start_us) - self._produced_audio_us
```

`commit_start_us` comes fresh from `commit_audio()` and is therefore always post-rebase. With no stalls the expression is constant and identical to the old value; after a rebase it absorbs the shift on the very next commit.

Three things read that anchor, so each needed follow-up:

- **History pruning** no longer derives its cutoff from the anchor at all. A chunk's `start_time_us` is already on the push stream's own clock, so the cutoff is that clock directly. Going via the anchor tied pruning to the commit tail, which `sleep_to_limit_buffer` lets run 30s ahead of the speakers, discarding committed-but-unplayed audio that a late-joining speaker is backfilled from.
- **Elapsed time** is now published on the rebasing commit. The periodic update only fires once the position advances a full second, which a backward rebase step delays for longer than the step itself.
- **Beat schedule** is re-published on a rebase, since `on_player_media_updated` fires on media identity changes, not position updates. Where the flow log cannot yet place the current track, the replacement anchor is the last published one shifted by the signed timeline delta, rather than derived from queue-backed progress, which is corrected asynchronously.

- none

## Types of changes

## Checklist

- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable — `tests/providers/sendspin/test_timeline_anchor_rebase.py` and `test_beat_anchor_rebase.py`.
- [ ] For changes to shared models, the companion PR in `music-assistant/models` is linked. — N/A
- [ ] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. — N/A
- [ ] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. — N/A

No `API_SCHEMA_VERSION` bump and no migration: nothing here touches API command signatures, config entries or database rows.

> [!NOTE]
> I was working on this PR in my fork if you wanted to see the CodeRabbit reviews for whatever reason: https://github.com/caraar12345/music-assistant-server/pull/2


OzGav:
Marking as draft due to copilot comments outstanding. Please as as ready for review when you want another look.

caraar12345:
Apologies - I had actually handled them, and the commit history showed that... until I rebased and it reordered 😅 (also forgot to click resolved). 

Sorry! Ready for review.

OzGav:
Marking this as draft as there are outstanding copilot comments

PR #6351, opened 2026-09-14
Keep one music provider from breaking every version lookup

stvncode:

Opening an album or a track page asks every music provider for other versions of
it. The helper behind that read the results straight off whatever the provider
returned, so a provider that reports `ProviderFeature.SEARCH` without
implementing it - it returns `None`, not an empty `SearchResults` - took the
whole lookup down with an `AttributeError`, discarding the answers the other
providers had already given:

```
music/albums/album_versions  →  'NoneType' object has no attribute 'albums'
music/tracks/track_versions  →  'NoneType' object has no attribute 'tracks'
```

The same helper backs the cross-provider matching during a library sync, so that
fails the same way.

- `_demo_music_provider.search()` was exactly such a provider: an unfinished stub
  marked `# type: ignore[empty-body]`. It now returns an empty `SearchResults`,
  and the ignore goes with it, so the next stub is caught by mypy instead of
  shipping. It matters beyond the demo: that file is the template a new provider
  is copied from.
- `MediaControllerBase.search()` no longer trusts the annotation. A provider that
  hands back nothing is skipped and named in the log, and the other providers
  still count - the same rule the global search already states for itself ("one
  failing provider must not break the entire search").

Only developers hit this today, since `_`-prefixed provider directories load in
dev mode only, which is why it is not flagged for backport. A third-party
provider making the same mistake would take these lookups down for real users.

Two tests fail on `dev` before this change and still do afterwards, neither
anywhere near this code path: `test_track_listing_with_pagination_matches_legacy`
and `test_verify_ssl_action_reports_the_certificate_info`.

## Types of changes

## Checklist


MarvinSchenkel:
I think you ran into this issue because of the missing return in the demo provider, there is no need to adjust the base controller here. If you just scope this PR to just the changes of the demo provider, it should be good to go (also ditch the test).

MarvinSchenkel (approved):

PR #6353, opened 2026-09-14
Load Plex library metadata in batches instead of one item at a time

RyanAtTanagra:

A Plex library sync makes one extra request for every artist, album and track, plus one per playlist entry.

Library listings return partial objects. The parsers additionally read `userRating`, `moods` and `styles`, which a listing does not include, so plexapi silently reloads that item on its own. On a large library that is tens of thousands of requests per sync.

This change loads the full metadata for each listing with multi-key requests (`/library/metadata/k1,k2,...`), 500 items at a time (different batch sizes were tested and 500 seems ideal), and parses from that. It sends the same include params plexapi uses when it reloads a single item, so the parsers see the same payload as before. Results are matched back to the listing by rating key, which keeps listing order and duplicate playlist entries intact, and an item the server no longer returns is parsed from the listing result exactly as it is today.

Measured on a library of 22,532 tracks, 2,301 albums and 923 artists: a full sync went from about 49,700 requests to about 140, and the track sync from 7.1 to 1.3 minutes. I compared the synced library field by field before and after and found no differences.

Not covered: `get_album_tracks` (on-demand, not part of a sync) and the collection branch of `get_playlist_tracks`, which mixes tracks and albums.

n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -61,7 +61,7 @@ from music_assistant_models.streamdetails import MultiPartPath, StreamDetails
 from plexapi.audio import Album as PlexAlbum
 from plexapi.audio import Artist as PlexArtist
 from plexapi.audio import Track as PlexTrack
-from plexapi.base import PlexObject
+from plexapi.base import PlexObject, PlexPartialObject
 from plexapi.myplex import MyPlexAccount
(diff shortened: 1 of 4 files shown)


MarvinSchenkel (approved):
Great improvement, thanks @RyanAtTanagra 🎉

PR #6354, opened 2026-09-14
Sync a Plex library as soon as Plex reports that it changed

RyanAtTanagra:

New music added to Plex only shows up in Music Assistant at the next scheduled sync, which is every 12 hours by default.

Plex pushes activity notifications over a websocket (`/:/websockets/notifications`), which is how its own apps pick up new music straight away. This adds an opt-in option, "Sync when the Plex library changes", that keeps that socket open and starts a sync of the provider when a library scan finishes.

The notification does not say which library was scanned, and every scan bumps `scannedAt` even when it finds nothing, so the provider compares the library's `contentChangedAt` from `/library/sections` and only syncs when it moved. If the socket drops it reconnects after 30 seconds, and the scheduled sync stays in place either way, so a broken socket only costs the faster pickup.

The token is sent in the `X-Plex-Token` header rather than the socket URL. aiohttp includes the URL in connection errors, so a token in the query string would end up in the debug log.

Tested against a real Plex server: adding an album was picked up in Music Assistant about 20 seconds later.

n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/providers/plex/__init__.py
+++ b/music_assistant/providers/plex/__init__.py
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar, cast
 from urllib.parse import urlencode
 from uuid import uuid4
 
+import aiohttp
 import plexapi.exceptions
 import plexapi.utils
 import requests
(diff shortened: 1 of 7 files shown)


RyanAtTanagra:
Added a fix for a library to be rechecked if in changed during a running sync.

Notes:
 - I had originally considered this but didn't implement it since this isn't a replacement for scheduled scans, which would pick up the change on the next run, and I was trying to keep the changes minimal.
 - With the PR that batches metadata loading and improves scans by a factor of almost 5x, the default scan interval could possibly be change from 12(?) to 1hr, making this even less critical.  https://github.com/music-assistant/server/pull/6353

However, the change did ended up being simpler than when I first considered it.

MarvinSchenkel (approved):
Looks good to me, thanks @RyanAtTanagra 🙌

PR #6355, opened 2026-09-14
Drop provider mappings for items the provider no longer has

RyanAtTanagra:
When a file is replaced, providers like Plex delete the old item and give the replacement a new id. The library item is matched on metadata, so it picks up the new mapping and keeps the old one, still marked available. Playback fails whenever the dead mapping is the one that gets picked, and a resync does not fix it because the item itself is still in the library. On my Plex library, 600 of 2,324 album mappings pointed at deleted items (after a big flac upgrade run).

The sync now records which items the provider listed, and the deletion pass removes this provider's mappings on synced items that were not listed.

It only runs for providers whose catalog is their library (`is_streaming_provider` is False). For a streaming provider an unlisted mapping only means the item is not in the user's library, so it is left alone. Because it runs inside the existing deletion pass, it follows the library sync deletions option, is held back on an incomplete run, and leaves items the provider reported as skipped alone. `filesystem_local` overrides `sync_library` and does not reach this code.

There is no schema change. The first sync after upgrading clears existing dead mappings, and later syncs find nothing more to remove.

`_run_library_sync` and `_sync_library_audiobooks` each went one statement over the ruff limit, so both have a `# noqa: PLR0915`.

 - n/a

## Types of changes

## Checklist

### DIFF
--- a/music_assistant/models/music_provider.py
+++ b/music_assistant/models/music_provider.py
@@ -42,6 +42,7 @@ from music_assistant.constants import (
     CONF_ENTRY_LIBRARY_SYNC_ALBUM_TRACKS,
     CONF_ENTRY_LIBRARY_SYNC_DELETIONS,
     CONF_ENTRY_LIBRARY_SYNC_PLAYLIST_TRACKS,
+    DB_TABLE_PROVIDER_MAPPINGS,
     PlaylistPlayableItem,
 )
 from music_assistant.controllers.tasks.context import (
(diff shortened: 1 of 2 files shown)


marcelveldt (approved):
Thanks, @RyanAtTanagra !

kieranhogg:
Could this be the fix for https://github.com/music-assistant/support/issues/6458 ?

RyanAtTanagra:
> Could this be the fix for [music-assistant/support#6458](https://github.com/music-assistant/support/issues/6458) ?

This PR only touches non-streaming providers, so it wouldn't apply to Apple Music.

PR #6356, opened 2026-09-14
Let the sample rates setting apply to Sonos players

RyanAtTanagra:

The per-player "Sample rates" setting has no effect on Sonos players, and there is no way to tell from the UI: the setting is not shown for them at all.

`SonosPlayer` sets `_attr_supported_sample_rates` in its constructor. `Player.get_supported_sample_rates()` returns a declared list as-is and only falls back to `CONF_SAMPLE_RATES` when nothing is declared, and the config controller only injects the sample rates entry for players that do not declare their rates. So for Sonos the output format always follows the source within the hardware limits, and a mixed-rate library keeps switching output format between tracks, which is audible as a gap at the track boundary on a grouped system. There is no way to pin one constant output format. Flow mode somewhat solves for this but per-track info gets lost in the stream.

This removes the hardcoded declaration and has the player offer the setting itself, built from the same hardware limits it declared before: 44.1/48 kHz, and 16 bit only for the models in `NON_HIRES_MODELS`.

The default selection covers that full range, so the resolved rates are identical to the previously declared list for every model and nothing changes for existing users until they choose otherwise. Selecting a single pair (for example 48000/16) is what pins the output to one constant format.

Only the sonos provider is changed. sonos_s1 declares [(44100, 16), (48000, 16)] in the same way, so the setting is hidden for those players too. That is the same limitation, left out here to keep this change to a single provider.

 - n/a

## Types of changes

## Checklist


marcelveldt:
Why is this needed ? This introduces a setting which is not exposed for a reason: we can just tell what the sonos speakers support.

marcelveldt (changes requested):
There is no motivation why we need to put this back

RyanAtTanagra:
On a large group (I have 13 devices and often have them all grouped), 24bit streaming runs into issues on wifi (and I have a decent multi-WAP UniFi network with good coverage), so I needed to limit to 16bit to Sonos devices.

marcelveldt:
> On a large group (I have 13 devices and often have them all grouped), 24bit streaming runs into issues on wifi (and I have a decent multi-WAP UniFi network with good coverage), so I needed to limit to 16bit to Sonos devices.

OK, it would have helped if you had described the usecase in the PR.
Since this is a advanced setting and the default is still good, we can proceed with the PR.

marcelveldt (approved):

marcelveldt:
@RyanAtTanagra do you want this backported into stable or are you running beta/nightly in your home ?

RyanAtTanagra:
> @RyanAtTanagra do you want this backported into stable or are you running beta/nightly in your home ?

I'm running stable with my changes patched in, so no need to backport on my account.

PR #6363, opened 2026-09-15
Ensure that the in-library view doesn't "lose" media items during a socket update in Audiobookshelf

fmunkes:

During a regular library sync the music provider model sets the provider mappings in_library attribute to True automatically when needed.
The abs provider supports updates outside the regular library syncs via a socket connection. However, each update resets this attribute to None, which showed itself in a suddenly vanishing media item in the respective library view if the item got updated. Only a full sync made the item reappear.

This PR ensures that in_library is always set to True during a socket update, while leaving the default regular sync route as is. Additionally, the proper add or update_item_in_library methods are used, as was only the case for playlists so far.

## Types of changes

## Checklist


fmunkes:
I'm just rethinking this, and I think currently due to the "overwrite" the item will still become lost, if the abs provider is added multiple times for different users with provider filtering. The overwrite is present in the current code as well. If I drop it, it behaves exactly like the normal sync, but some metadata (e.g. the name) can only be adjusted by deleting and resyncing the item.

I dropped the overwrite - this way, both sync paths either regular or socket should give the same result.

MarvinSchenkel (approved):
Thanks @fmunkes

PR #6364, opened 2026-09-15
Add Teufel Raumfeld player provider

Simanias:

Adds a native **player provider for Teufel Raumfeld** (Smart Speaker) multiroom devices. Communication with the Raumfeld host goes through the [`hassfeld`](https://pypi.org/project/hassfeld/) library.

Each Raumfeld **room** is exposed as a stable Music Assistant player, and MA sync-groups are mapped onto Raumfeld **zones**. Playback points the zone's UPnP AVTransport at MA's own stream URL (similar to the DLNA provider).

Highlights:

- Play / stop / volume / mute
- **Pause → resume from position.** Raumfeld renderers drop the HTTP connection on a UPnP Pause and re-fetch on Play, so an in-place pause restarts the track. The provider therefore does not advertise `PAUSE`; MA treats pause as stop and resumes by re-streaming from the stored position. The flow-mode toggle is hidden so this can't be broken (a flow stream can't be resumed by seeking).
- **Per-track (`ENQUEUE`) playback** with a smoothed elapsed-time clock (the device reports whole-second positions, so the clock only re-anchors on real divergence).
- **Multi-room grouping** via `SET_MEMBERS` → Raumfeld zones (`async_add_room_to_zone` / `async_drop_room_from_zone`).
- **Player de-duplication**: exposes the room's renderer UUID + IP as device identifiers so MA merges this native player with the same device's Chromecast/DLNA/Sendspin representations instead of listing duplicates.
- **Hi-res output** up to 24-bit / 192 kHz.
- Setup flow for the host IP, translatable strings, and a provider icon.

New dependency: `hassfeld==0.3.14`. Provider stage is `experimental`. No gapless between tracks: the Raumfeld zone renderer does not expose `SetNextAVTransportURI`, so `enqueue_next_media` is a no-op and MA plays each queue item individually. Tested live against a real 4-room Raumfeld system (setup, discovery, playback, pause/resume, grouping/ungrouping, de-duplication, hi-res).

Disclosure: this contribution was prepared with AI assistance (Claude Code) under the author's review and testing.

- n/a

## Types of changes

## Checklist


music_assistant/providers/raumfeld/constants.py
```diff
@@ -0,0 +1,44 @@
...
+    CONF_ENTRY_FLOW_MODE,
+    CONF_ENTRY_FLOW_MODE_SAMPLE_RATE,
+    create_sample_rates_config_entry,
+)
+
+DOMAIN = "raumfeld"
+
+# Config entry keys
+CONF_HOST = "host"
+CONF_PORT = "port"
```
OzGav: There are already MA constants for these

music_assistant/providers/raumfeld/helpers.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def room_to_player_id(room: str) -> str:
+    """
+    Derive a stable Music Assistant player_id from a Raumfeld room name.
+
+    NOTE: Raumfeld's control API (hassfeld) is keyed by *room name*, so we keep the
+    original room string on the player for issuing commands and only slugify it for
+    the player_id. Renaming a room in the Raumfeld app therefore creates a new
+    player; that is an accepted limitation of the room-name-based identity.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class RaumfeldPlayer(Player):
+    """
+    A single Raumfeld room, exposed to Music Assistant as a player.
+
+    Grouping: Raumfeld groups rooms into dynamic *zones*. We translate MA's
+    leader/members model onto zones via the hassfeld host. Playback and transport
+    commands always target the *zone* the room currently belongs to (a solo room is
+    treated as a single-room zone).
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+        self.update_state()
+
+    async def enqueue_next_media(self, media: PlayerMedia) -> None:
+        """
+        Handle enqueuing the next queue item.
+
+        Raumfeld zone renderers do not expose ``SetNextAVTransportURI``, so we cannot
+        pre-buffer the next item for gapless playback. ENQUEUE is still declared to keep
+        MA out of (non-resumable) flow mode; the queue controller falls back to calling
+        ``play_media`` for each item when the previous one ends.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+        self,
+        player_ids_to_add: list[str] | None = None,
+        player_ids_to_remove: list[str] | None = None,
+    ) -> None:
+        """
+        Handle SET_MEMBERS: translate MA group changes into Raumfeld zone changes.
+
+        The leader (this player) owns the zone. Adding a member adds its room to this
+        room's zone; removing a member drops its room. MA-side bookkeeping mirrors the
+        demo provider so that each member's derived ``synced_to`` recomputes.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # Playback state (per zone). GetTransportInfo -> CurrentTransportState.
+        playing = False
+        try:
+            transport = await host.async_get_transport_info(zone)
+            self._attr_playback_state = _map_transport_state(
+                (transport or {}).get("CurrentTransportState")
+            )
+            playing = self._attr_playback_state == PlaybackState.PLAYING
+        except Exception as err:
```
OzGav: Please don’t have broad catch alls like this. Narrow to expected errors

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+        self.update_state()
+
+    def _mark_play_started(self) -> None:
+        """
+        Record a play/resume command and switch to fast polling for startup.
+
+        Also reset the reported position to 0: on resume, MA re-streams the track from a
+        seek offset and adds that offset to the player's position, so a stale pre-pause
+        position here would be double-counted (a forward jump) until the device reports
+        the new stream's 0-based time. The next poll re-anchors to the real position.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+        self._attr_poll_interval = FAST_POLL_INTERVAL
+        self._attr_elapsed_time = 0.0
+        self._attr_elapsed_time_last_updated = time.time()
+
+    def _current_zone(self) -> list[str]:
+        """
+        Return the room-list identifying the zone this room currently controls.
+
+        A room that is not part of an active multi-room zone is treated as its own
+        single-room zone, which is the unit hassfeld's transport commands expect.
```
OzGav: As above and just single line docstrings for private functions

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,314 @@
...
+        """
+        return self.raumfeld.get_zone_for_room(self.room) or [self.room]
+
+    def _player_id_to_room(self, player_id: str) -> str | None:
+        """
+        Resolve a MA player_id belonging to this provider back to its Raumfeld room.
+
+        :param player_id: The Music Assistant player id to resolve.
```
OzGav: Single line docstrings for private functions

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,126 @@
...
+
+    async def handle_async_init(self) -> None:
+        """Handle async initialization of the provider."""
+        host = cast("str", self.get_setup_value(CONF_HOST))
+        port = cast("int", self.get_setup_value(CONF_PORT) or DEFAULT_PORT)
+
+        self.host = hassfeld.RaumfeldHost(host, port, session=self.mass.http_session)
+
+        if not await self.host.async_host_is_valid():
+            raise RuntimeError(f"'{host}:{port}' is not a valid Raumfeld host")
```
OzGav: You need to raise MA errors so they get handled correctly

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,126 @@
...
+        # async_update_all runs the hassfeld background update loops for the whole
+        # lifetime of the provider; keep the handle so we can cancel it on unload.
+        self._update_task = self.mass.create_task(
+            self.host.async_update_all(self.mass.http_session)
+        )
+        try:
+            async with asyncio.timeout(INITIAL_UPDATE_TIMEOUT):
+                await self.host.async_wait_initial_update()
+        except TimeoutError as err:
+            raise RuntimeError(
```
OzGav: As above

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,126 @@
...
+        return None
+
+    def resolve_room_renderer(self, room: str) -> tuple[str | None, str | None]:
+        """
+        Return the ``(renderer UUID, renderer IP)`` for a room's media renderer.
+
+        These let Music Assistant link this native player to the same device's other
+        protocol representations (DLNA/Chromecast/Sendspin) instead of listing them as
+        duplicates. The UUID is the renderer UDN without the ``uuid:`` prefix; either
+        value may be ``None`` if the host has not resolved it.
```
OzGav: Don't explain inner workings of the code in the docstrings (you can use inline comments for that if/when needed). The docstring should provide clarity to the caller of the function/method, not explain how it works technically/internally.

OzGav (changes requested):
Thanks for this. These comments are from a quick glance. Detailed reviews will follow.

Simanias:
> Thanks for this. These comments are from a quick glance. Detailed reviews will follow.

Thanks for the quick reply, the suggested comments are included in the new PR.

OzGav:
Please comment on and resolve each open conversation when it has been addressed.

music_assistant/providers/raumfeld/constants.py
Simanias: Done — switched to MA's CONF_IP_ADDRESS and CONF_PORT from music_assistant.constants; the provider-local CONF_HOST/CONF_PORT are gone.

music_assistant/providers/raumfeld/helpers.py
Simanias: Done — trimmed the docstring to a caller-focused summary and moved the room-name/slug explanation into an inline comment.Done

music_assistant/providers/raumfeld/player.py
Simanias: Done — the class docstring is now a single caller-focused line; the zone/grouping detail lives in inline comments where relevant.

music_assistant/providers/raumfeld/player.py
Simanias: Done — single-line docstring; the "no SetNextAVTransportURI, so we play the next item ourselves" note is now an inline comment.

music_assistant/providers/raumfeld/player.py
Simanias: Done — reduced to a single caller-focused line.

music_assistant/providers/raumfeld/player.py
Simanias: Done — replaced the broad except Exception with except HOST_ERRORS (a defined tuple of the aiohttp/OS/timeout/lookup errors expected from the host), used consistently across the module.

music_assistant/providers/raumfeld/player.py
Simanias: Done — single-line docstring; the resume/seek-offset reasoning is now an inline comment.

music_assistant/providers/raumfeld/player.py
Simanias: Done — single-line docstring.

music_assistant/providers/raumfeld/player.py
Simanias: Done — single-line docstring.

music_assistant/providers/raumfeld/provider.py
Simanias: Done — this no longer raises. The host connection is now owned by a background supervisor that retries instead of hard-failing, so there's no RuntimeError on an invalid/unreachable host; user-facing command failures raise PlayerCommandFailed.

music_assistant/providers/raumfeld/provider.py
Simanias: Done — the initial-update timeout now pauses via the supervisor and retries rather than raising; no RuntimeError remains.

music_assistant/providers/raumfeld/provider.py
Simanias: Done — docstring is now caller-focused (returns (renderer UUID, renderer IP) or (None, None)); the protocol-linking rationale moved to the call site.

Simanias:
> Thanks for this. These comments are from a quick glance. Detailed reviews will follow.

All has been resolved. The docstring-treads are committed in [2b62b0d](https://github.com/music-assistant/server/pull/6364/commits/2b62b0d6501099a61b1dbe72ee12c2040f068ca6)

music_assistant/providers/raumfeld/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+def room_to_player_id(room: str) -> str:
+    """
+    Derive a stable Music Assistant player_id from a Raumfeld room name.
+
+    :param room: The Raumfeld room name.
+    """
+    # hassfeld is keyed by room name (kept on the player for commands); only the
+    # player_id is slugified, so renaming a room in the app creates a new player
+    slug = _slug_re.sub("_", room.strip().lower()).strip("_")
+    return f"{PLAYER_ID_PREFIX}_{slug}"
```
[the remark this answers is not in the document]
Simanias: Fixed — player_id is now derived from the immutable room UDN (room_udn_to_player_id), not the room name. Renames and slug collisions no longer orphan config or hide a room; the room name is display-only. (The provider is still experimental/unreleased, so no migration is needed.)

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,507 @@
...
+        # only a room that is part of an active zone has something to stop; a
+        # standby/unassigned room has no zone the host can address
+        if (zone := self._active_zone()) is not None:
+            try:
+                await self.raumfeld.host.async_zone_stop(zone)
+                if hard_stop:
+                    for room in zone:
+                        await self.raumfeld.host.async_enter_manual_standby(room)
+            except HOST_ERRORS as err:
+                self.logger.debug("Failed to stop %s: %r", self.room, err)
```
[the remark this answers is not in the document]
Simanias: Fixed — stop() now raises a chained PlayerCommandFailed on a host error instead of optimistically reporting IDLE. Note: the line-348 site is poll() reading GetPositionInfo (a read, not a command); that stays tolerant on purpose so a transient poll blip doesn't crash polling or mark the player failed.
Addressed in 6c00b27

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,291 @@
...
+        ip_address = urlparse(devloc).hostname if devloc else None
+        return uuid, ip_address
+
+    async def _supervise(self) -> None:
+        """Supervise the host connection: connect when reachable, pause when it drops."""
+        while True:
+            try:
+                if not self._connected:
+                    await self._try_connect()
+                elif not await self.host.async_host_is_valid():
```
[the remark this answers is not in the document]
Simanias: Fixed — the supervisor now treats a missing/done() update task as disconnected, so when hassfeld swallows a disconnect inside async_update_all() and returns, the next cycle reconnects and recreates the task instead of leaving state frozen.

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,291 @@
...
+                identifiers.get(IdentifierType.IP_ADDRESS) == host_ip
+                and uuid
+                and uuid not in physical
+            ):
+                self.logger.info(
+                    "Disabling redundant Raumfeld host DLNA renderer '%s' (%s)",
+                    player.display_name,
+                    player.player_id,
+                )
+                await self.mass.config.save_player_config(player.player_id, {ATTR_ENABLED: False})
```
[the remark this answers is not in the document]
Simanias: Removed — dropped the cross-provider config mutation entirely. Linking now relies solely on the player controller's identifier-based protocol linking; unmatched virtual renderers are left to that mechanism rather than disabled behind another provider's back.

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,291 @@
...
+            if len(member_ids) < 2:
+                continue
+            leader = self.mass.players.get_player(member_ids[0])
+            if isinstance(leader, RaumfeldPlayer):
+                leader.set_group_members(member_ids)
+            grouped.update(member_ids)
+        # any player no longer part of a multi-room zone must be marked solo
+        for player in self.players:
+            if isinstance(player, RaumfeldPlayer) and player.player_id not in grouped:
+                player.set_group_members([])
```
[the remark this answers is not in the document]
Simanias: Fixed — _sync_groups now tracks current leader ids and clears group_members on every non-leader (solo rooms and followers), so a former leader that became a follower can't retain a stale list and form a circular group.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,507 @@
...
+        if transport_ok:
+            self._maybe_advance(playing, recently_started)
+
+        self.update_state()
+
+    def _maybe_advance(self, playing: bool, recently_started: bool) -> None:
+        """Play the next queue item when the current track has finished."""
+        # a track that was playing and is now stopped - outside the startup transition and
+        # not a user stop (which disarms) - has ended on its own
+        if self._advance_armed and self._prev_playing and not playing and not recently_started:
```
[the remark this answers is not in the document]
Simanias: Added tests/providers/raumfeld/ with tests for the end-of-track state machine (PLAYING→IDLE advancement, startup-window suppression, user-stop disarming, absent-next-item) plus helper tests (duration/line-in/player-id parsing).

music_assistant/providers/raumfeld/setup_flow.py
```diff
@@ -0,0 +1,68 @@
...
+    errors: dict[str, str] | None = None
+    setup_data = dict(session.context.setup_data)
+    while True:
+        entries = [
+            replace(entry, value=setup_data.get(entry.key, entry.value)) for entry in _ENTRIES
+        ]
+        submitted = await session.form(entries, step_id="user", errors=errors, last_step=True)
+        setup_data.update(submitted)
+        host = str(setup_data.get(CONF_IP_ADDRESS) or "").strip()
+        port = int(setup_data.get(CONF_PORT) or DEFAULT_PORT)
```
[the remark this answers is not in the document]
Simanias: Fixed — the normalized (trimmed) host and port are written back into setup_data before finish(), so the stored value matches what was validated.

music_assistant/providers/raumfeld/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+    result["album"] = album.text if album is not None else None
+    result["image_url"] = image.text if image is not None else None
+    return result
+
+
+def parse_duration(value: str | None) -> int | None:
+    """
+    Parse a UPnP ``H:MM:SS`` (or ``H:MM:SS.mmm``) duration string into whole seconds.
+
+    :param value: The duration string, or None.
```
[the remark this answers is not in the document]
Simanias: Fixed — parse_duration now wraps the shared music_assistant.helpers.util.try_parse_duration (tolerating None/NOT_IMPLEMENTED/garbage) instead of reimplementing the parser.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,507 @@
...
+        self._attr_active_source = SOURCE_LINE_IN
+        self.set_current_media(uri=url, clear_all=True, title="Line-in")
+        self._attr_playback_state = PlaybackState.PLAYING
+        self.update_state()
+
+    async def enqueue_next_media(self, media: PlayerMedia) -> None:
+        """Handle enqueuing of the next queue item."""
+        # Raumfeld zone renderers have no SetNextAVTransportURI, so we can't hand the next
+        # item to the device; keep it and play it ourselves when the current track ends.
+        self._next_media = media
```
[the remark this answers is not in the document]
Simanias: You'll want to correct the PR description here (not code): the provider does implement software-managed enqueueing — enqueue_next_media stores the next item and _maybe_advance plays it when the current track ends (Raumfeld renderers have no SetNextAVTransportURI). I've shortened the wording accordingly.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,542 @@
...
+            self.logger.debug("Failed to read position info for zone %s: %r", zone, err)
+            return None
+
+    def _maybe_advance(self, playing: bool) -> None:
+        """Play the next queue item when the current track has finished."""
+        # a track that was playing and is now stopped - and not a user stop (which
+        # disarms) - has ended on its own. _mark_play_started resets _prev_playing to
+        # False, so the not-yet-playing gap right after a play command can't look like an
+        # ended track (that gap starts from prev_playing=False).
+        if self._advance_armed and self._prev_playing and not playing:
```
OzGav, line 435: `_maybe_advance` treats any non-`PLAYING` transport reading as end-of-track, but `_map_transport_state` returns `PAUSED` for `PAUSED_PLAYBACK`. Pausing the room from the Raumfeld app while an MA queue is armed skips to the next track. Gate on `CurrentTransportState in ("STOPPED", "NO_MEDIA_PRESENT")`, ideally plus a position-near-duration check so a buffer stall doesn't skip either.

music_assistant/providers/raumfeld/setup_flow.py
```diff
@@ -0,0 +1,71 @@
...
+        # persist the normalized values so finish() stores exactly what we validated
+        setup_data[CONF_IP_ADDRESS] = host
+        setup_data[CONF_PORT] = port
+        if not await _host_reachable(session, host, port):
+            errors = {CONF_IP_ADDRESS: f"No Raumfeld host found at {host}:{port}"}
```
OzGav, line 55: Please make this translatable.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,542 @@
...
+        self._attr_playback_state = PlaybackState.PLAYING
+        self.update_state()
+
+    async def enqueue_next_media(self, media: PlayerMedia) -> None:
+        """Handle enqueuing of the next queue item."""
+        # Raumfeld zone renderers have no SetNextAVTransportURI, so we can't hand the next
+        # item to the device; keep it and play it ourselves when the current track ends.
+        self._next_media = media
+
+    async def volume_set(self, volume_level: int) -> None:
```
OzGav, line 256: This and volume_mute don’t raise `PlayerCommandFailed` like your other methods. Please wrap these as well.

music_assistant/providers/raumfeld/player.py
```diff
@@ -0,0 +1,542 @@
...
+            playing = self._attr_playback_state == PlaybackState.PLAYING
+
+        if position is not None:
+            self._apply_position(position, playing)
+
+        # Poll fast while playing - and for a short window right after a play/resume
+        # command, so MA locks onto the device's real position quickly instead of
+        # running its own clock ahead during the device's buffering/startup delay.
+        recently_started = (time.time() - self._play_started_at) < STARTUP_POLL_WINDOW
+        self._attr_poll_interval = FAST_POLL_INTERVAL if (playing or recently_started) else 15
```
OzGav, line 353: The poll interval drops to 1 second while a player is playing, and each poll sends three separate requests to the Raumfeld host, one for volume, one for transport state and one for position. Every room polls, and rooms that are grouped together all ask about the same zone, so the same two answers get fetched once per room. A group of four rooms works out at roughly twelve requests a second at the host for as long as music is playing.

Music Assistant also polls players one after another rather than all at once, so when the host is slow to answer, every other player in the setup waits behind it.

For comparison, Bluesound polls every 10 seconds during playback and Bose SoundTouch every 20.

Two changes would bring this down a lot. Let only the group leader ask for transport state and position, since the answer is the same for every room in the zone, and have the other rooms read just their own volume. Then use something like 5 seconds while playing, keeping the 1 second rate for the short window right after a play command and for the last few seconds of a track when you are watching for it to finish. You already know the track length, so you know when it is worth looking closely.

music_assistant/providers/raumfeld/provider.py
```diff
@@ -0,0 +1,273 @@
...
+                    existing.set_available(True)
+                continue
+            player = RaumfeldPlayer(provider=self, player_id=player_id, room=room)
+            await self.mass.players.register(player)
+            self.logger.debug("Registered Raumfeld room '%s' as player %s", room, player_id)
+        # a room the host no longer lists (e.g. a speaker that dropped to deep standby or
+        # left the network) is gone until it returns; show it as unavailable meanwhile
+        present = set(rooms)
+        for known in self.players:
+            if isinstance(known, RaumfeldPlayer) and known.room not in present:
```
OzGav, line 225: If someone renames a room in the Raumfeld app, that room's player gets stuck as unavailable until the provider is reloaded.

This check compares the name the player stored when it was created against the names the host currently reports. After a rename the old name is not in that list any more, so the player gets marked unavailable. The loop just above it already found the same player by its room id and marked it available, so both things happen in the same pass, and it repeats on every sync from then on.

See also line 75 in player.py, where the room name is stored. Nothing ever updates it, and it is the name used for every call to the host, so play, stop and volume would all be pointing at a name that no longer exists anyway.

The player id already comes from the room id and does not change when the name does, so the fix is to look the player up by id and refresh its stored room name and display name whenever the host reports a different one. Then only mark a player unavailable when its room id has actually gone from the host, instead of when the name stops matching.

OzGav (changes requested):

music_assistant/providers/raumfeld/player.py
Simanias: Good catch — advancement now only triggers on an actual STOPPED/NO_MEDIA_PRESENT transport state near the end of the track (a _near_end flag set from the position poll), not on any playing→stopped edge. A pause (reported as PAUSED_PLAYBACK) or a mid-track buffer stall no longer skips to the next item. Added tests covering the pause-before-end and startup-gap cases. Done in e80c5b3.

music_assistant/providers/raumfeld/setup_flow.py
Simanias: Switched the "no host found" setup error to a translatable key (no_host_found) in the provider's strings.json errors section, and the generated en.json is updated accordingly. Done in e80c5b3.

music_assistant/providers/raumfeld/player.py
Simanias: volume_set and volume_mute now wrap the host calls and raise PlayerCommandFailed on a host error, so a failed volume change surfaces to the user instead of being swallowed. Done in e80c5b3.

music_assistant/providers/raumfeld/player.py
Simanias: Reworked polling so only the zone leader does the transport/position round-trips; grouped followers poll only their own per-room volume. The three leader reads run concurrently, and the poll interval is now tiered — 1s right after a play command and in the last seconds of a track, 5s while playing, 15s when idle — so it's responsive at transitions without hammering the host the rest of the time. Done in e80c5b3.

music_assistant/providers/raumfeld/provider.py
Simanias: The player_id is now derived from the immutable room UDN rather than the (mutable) room name, so a rename no longer orphans the player. _sync_rooms matches on that stable id and calls a new set_room() to refresh the name used for host calls and display. Done in e80c5b3.

Simanias:
## Gapless playback: why it is built this way, and what it costs

Track transitions used to stop, re-point and re-play the transport, which audibly clipped the start of the next track. This is now gapless for a solo room. Getting there meant working around Raumfeld's design rather than with it, so the reasoning and the price are recorded here.

### Raumfeld keeps sequencing to itself

Three ways an external controller might hand Raumfeld a next track. All three are closed:

**1. The zone renderer omits it.** Enumerating its AVTransport gives `SetAVTransportURI`, `Play`, `Pause`, `Seek`, `Next`, `BendAVTransportURI` — but no `SetNextAVTransportURI`, and `GetMediaInfo` reports `NextURI` as `NOT_IMPLEMENTED`. This is not a gap in the firmware: the room's *own* renderer has `SetNextAVTransportURI` **and** `SetNextStartTriggerTime`, and the host uses both. A speaker's log shows it receiving them, alongside a relay stream from the host and a clock-correction loop:

```
rendererHttp: finished async http job for http://<host>:51811/<zone>/<room>/<renderer>/…
rendererTransportService::SetNextStartTriggerTime 9445:6000522
rendererRingbuffers fill state: 8810 ms
raumfeldTC: storing -0.000002027 diff … last: 0.999740899
```

The zone renderer is a control facade; the host owns sequencing and sync.

**2. The queue API is catalogue-only.** The host's `ContentDirectory` does have `CreateQueue`, `AddItemToQueue`, `AddContainerToQueue`, `MoveInQueue`, `RemoveFromQueue`. But `AddItemToQueue(QueueID, ObjectID, Position)` takes an `ObjectID` — an object already in Raumfeld's own catalogue. There is no `CreateObject`, and browsing the root shows only Raumfeld's own sources (TIDAL, TuneIn, Line-in, My Music, Playlists, Spotify, Favourites). A Music Assistant stream URL can never become one of these.

**3. The host webservice does not play.** Its endpoints are `connectRoomToZone`, `connectRoomsToZone`, `dropRoomJob`, `enterAutomaticStandby`, `enterManualStandby`, `leaveStandby`, `getHostInfo`, `getZones`, `addMusicServer`, `addResource`, `setSpotifyMode`, … — system management only. Nothing for a queue or a next track.

Music Assistant is by definition an external URL source, so it falls outside all three.

### What this implementation does

Pre-load the next item on the **room's own renderer**, which does accept `SetNextAVTransportURI`, and let the device cross the boundary itself.

The zone never reports such a hand-over — its metadata stays on the track it streamed while its clock runs on across the boundary — so the hand-over is predicted from the zone clock instead of observed, and the zone clock is translated into the item MA is actually on. Prediction is needed because MA's poll loop ticks at one second and the zone reports whole seconds, which would otherwise show the new track up to a second late.

### What it costs

| | Gapless (solo) | Raumfeld app metadata | Grouped rooms |
|---|---|---|---|
| This implementation | yes | stale | unchanged, not gapless |
| Staying on the zone only | no | correct | unchanged |

- **Grouped rooms are deliberately excluded.** In a zone the coordinator feeds the members one shared stream; a per-speaker next track would fight the host's trigger-time coordination and break sync.
- **The Raumfeld app keeps showing the previous track.** The host is unaware of a hand-over, and no UPnP action updates metadata without also restarting playback — which would defeat the point.

Both follow directly from bypassing the host. If that trade is not wanted, dropping the pre-load restores the previous behaviour and leaves the rest of this PR intact.

---

## Player ids now come from the device serial (revisits review point 5)

Your earlier review asked for ids that survive a room rename, which is why they moved to the room UDN. That fixes a rename, but it turns out the room UDN has its own failure case, and I hit it here.

Enabling SSH on one speaker meant removing and re-adding it. That rebuilt the host's room table and handed **all four** rooms a fresh UDN — so every player in MA was replaced by a new one, and volume, grouping and enabled state were left behind on ids nothing points to. Six orphaned configs accumulated across the day.

The renderer's `serialNumber` lives in the hardware and survives both cases. Through that same event the room UDNs changed three times while the renderer identity did not. So:

| Identity | Survives rename | Survives system rebuild |
|---|---|---|
| Room name | no | yes |
| Room UDN (previous) | yes | **no** |
| Device serial (now) | yes | yes |

Details that seemed worth getting right:

- The serial is read once per room from the device description and cached. A room whose serial cannot be read yet is **skipped and retried** on the next resync, rather than registered under an id that would later have to change.
- The host runs virtual renderers that inherit its own serial — there is one named "Bar" on the host carrying the TV room's serial. A room resolving to a duplicate is left out instead of quietly taking over another room's player and config.
- Existing installs are **migrated on first start**: config under the legacy room-UDN id is moved across, matched through the room the host still reports. It runs once, does nothing on a second start, and an install that cannot be matched simply starts with defaults.

The trade-off worth flagging: identity now follows the hardware, so a speaker physically moved to another room takes its settings along. That seemed the lesser evil against losing every player on a system rebuild, but say the word if you would rather have it the other way.

PR #6397, opened 2026-09-19
Remove stale author/narrator links when an audiobook is overwritten

fmunkes:

_update_library_item never forwarded its overwrite flag, so audiobook_artists rows were only ever inserted. 
Forwards it and splits the clear per artist type, since both types share one table. 

## Types of changes

## Checklist


music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -415,28 +416,41 @@ async def _set_artist_mappings(
...
-                author.artist_type = ArtistType.AUTHOR
-            await self._set_audiobook_authors_narrators(db_id, authors)
-        if item.narrators and isinstance(item.narrators[0], Artist):
-            # only for type checking
-            narrators = [narrator for narrator in item.narrators if isinstance(narrator, Artist)]
-            for narrator in narrators:
-                # just to be sure
-                narrator.artist_type = ArtistType.NARRATOR
-            await self._set_audiobook_authors_narrators(db_id, narrators)
+        await self._set_audiobook_authors_narrators(db_id, artists)
```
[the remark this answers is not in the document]
fmunkes: Good catch, the role change did lose the link. I fixed it in 206eeb56e without forwarding overwrite, though. Forwarding it would replace every contributor's name, metadata and external ids with the provider's on each audiobook overwrite, and drop anything metadata providers added. Instead, _set_audiobook_author_narrator now updates the artist only when its stored artist_type differs from the provider's. A normal (non-overwrite) update already writes artist_type, so the role is stored before the type-scoped delete runs, and everything else about the artist stays as it was. test_overwrite_keeps_an_artist_whose_role_changed covers the narrator → author case.

music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -460,7 +474,13 @@ async def _set_audiobook_author_narrator(
         ):
             db_artist = existing
 
-        if not db_artist or overwrite:
+        # stale links are removed by the stored type, so a changed role must be stored
```
OzGav: This seems complex so I am posting all of Claude’s assessment. See what you think.

An artist only has one type stored for the whole library, but whether someone is the author or the narrator depends on the book. This change rewrites that stored type whenever it doesn't match the current book, and that causes three problems.

1.	Plenty of authors narrate their own books. For those books the author step saves the person as an author, then the narrator step saves them as a narrator. They end up stored only as a narrator, so they disappear from the author list for every book they wrote. This happens on every sync, not just when overwriting.

2.	If someone wrote one book and narrated another, their type keeps switching depending on which book synced last. Each switch is a database write plus an update event, twice per sync.

3.	Say that person was just switched to narrator by another book. If this book is then overwritten with only narrators, their author link on this book is deleted.

I'd suggest leaving the artist's stored type alone and removing this check. Instead, gather the ids of every author and narrator in the update first. Then, when overwriting, delete only the links for this book whose artist is not in that list and whose type was part of the update. Someone who went from narrator to author is in the list, so their link stays. A type the update left out is still untouched.

tests/controllers/music/test_audiobook_artist_overwrite.py
```diff
@@ -0,0 +1,130 @@
...
+    db_id = await _add(mass, _book(narrators=[_artist("Old Voice", ArtistType.NARRATOR)]))
+
+    await mass.music.audiobooks.update_item_in_library(
+        db_id, _book(narrators=[_artist("New Voice", ArtistType.NARRATOR)])
+    )
+
+    assert "Old Voice" in await _linked(mass, db_id, ArtistType.NARRATOR)
+
+
+async def test_overwrite_keeps_an_artist_whose_role_changed(mass: MusicAssistant) -> None:
```
OzGav: Inline with the main comment, assuming you agreed with it, please add a test where the same person is both the author and the narrator of one book. It should check that they're still linked as the author, and that their stored type doesn't change after syncing twice. Please also add a test where one person wrote one book and narrated another, and overwriting one of the books leaves the other book's link alone. The role change test at this line only passes because of the type switching described above.

PR #6398, opened 2026-09-19
Ensure an updated author or narrator reaches the library in a sync

fmunkes:

_library_item_needs_update only compared provider mappings and date_added, so a corrected author or narrator never reached the library. Adds a details_changed hook on the sync snapshot; the audiobook snapshot now carries the stored names and compares them order-insensitively. Subsumes the ad-hoc str→Artist check.

## Types of changes

## Checklist


music_assistant/controllers/music/media/audiobooks.py
```diff
@@ -642,18 +642,20 @@ def _sync_details_query_parts(self) -> tuple[str, str, dict[str, Any]]:
             playlog_user_clause = "AND p2.userid = :playlog_userid "
             params["playlog_userid"] = session_user.user_id
         extra_columns = f"""
-            , EXISTS (
-                SELECT 1 FROM {DB_TABLE_AUDIOBOOK_ARTISTS}
+            , (
```
OzGav: This compares the names on the library's artist records with the names the provider sends, and those don't have to match. A library artist can be shared with another provider, renamed by a metadata lookup, or edited by the user. "J.R.R. Tolkien" in the library against "J. R. R. Tolkien" from the provider would count as a change on every sync, forever. For linked artists, I think compare the provider's artist IDs with the stored links. Only compare names for plain text authors.

music_assistant/models/music_provider.py
```diff
@@ -2000,6 +1986,9 @@ def _library_item_needs_update(
         if not self._check_provider_mappings(library_item, prov_item, True):
             # provider mapping doesn't match the library item
             return True
+        if not isinstance(library_item, MediaItem) and library_item.details_changed(prov_item):
```
OzGav: This is a complex potential problem so this is Claude’s complete assessment.

This spots the change, but the update it triggers mostly doesn't save it. The sync calls `update_item_in_library` without overwrite, and that keeps whatever is already stored.

•	Plain text authors. `_update_library_item` writes `cur_item.authors` or `_update_authors`, so the old name stays because the stored list isn't empty.

•	Linked artists.` _set_artist_mappings` only adds rows to `audiobook_artists` and never removes old ones. The corrected author is added next to the wrong one, the library shows both, and the names never match again.

•	Same artist renamed on the provider. `_set_audiobook_author_narrator` finds the existing library artist and doesn't change its name.

•	Provider going back from artists to plain text. The old links are never removed, so this check fires on every sync. The old check only handled the other direction, so this loop is new.

Only the switch from plain text to linked artists works. In every other case the book is rewritten on every sync from then on. When this check fires, the update should replace the authors and narrators, not add to them. That means deleting this book's author and narrator rows in `audiobook_artists` before adding the new ones, and writing the provider's plain text names into the authors and narrators columns. A full `overwrite=True` goes too far because it also replaces metadata and mappings.

tests/controllers/music/test_audiobook_author_changes.py
```diff
@@ -0,0 +1,123 @@
...
+    return Artist(
+        item_id=f"author:{name}",
+        provider=INSTANCE_ID,
+        name=name,
+        artist_type=ArtistType.AUTHOR,
+        provider_mappings={_mapping()},
+    )
+
+
+def _needs_update(stored: AudiobookSyncDetails, reported: MediaItemType) -> bool:
```
OzGav: Every test only calls `_library_item_needs_update` on a snapshot built by hand. None of them checks that the update saves the new names or that the next sync leaves the book alone, and that's where the bug above is. Add a test that goes through the database, once with plain text authors and once with linked artists. Store a book, sync a corrected author, then check the stored names and that the next check finds no change.

PR #6399, opened 2026-09-19
Add support for authors and narrators to the filesystem providers

fmunkes:

Adds support for authors and narrators to the filesystem providers via tags. Author tags are writers, album-artists, and artist, narrator tags are narrators, narrator, narrated-by and composer

## Types of changes

## Checklist


OzGav:
Thanks. Same gap as the collections PR. On m4a and m4b we do not read the narrator, narratedby or writer tags at all, and that is the format most audiobooks use. The narrator can only come from the composer tag there, and if that names two people we keep only the first, so a book with two narrators loses one. The same loss happens on mp3. Both are fixable in _parse_id3_tags and _parse_mp4_tags, which already do this for album artists.

Please add tests using the real fixture files for all four formats rather than a dictionary of values.

authors falls back to the file name or Unknown artist, and those now become real artists in the library, so that needs skipping.

The one off rescan needs some thought too. After it turns existing author names into artists, the old names are still sitting in the authors column on the audiobook, so the same thing is stored twice and the two can disagree. Can we settle that before merge.

fmunkes:
So, I obviously tested that PR previously, and it did work with my lib - ffprobe appears to keep the first value of a freeform tag, and I got lucky.
Anyhow, I let claude work on it, and all your points should be covered. Tested it again. On your question on that one-off scan. I didn't touch it, as I honestly don't know what's best. Not scanning means the provider will at some point have both, promoted and non-promoted artists. It is correct, that they can disagree, though our database queries shouldn't take the old values. Still, would you like me to wipe the old values?

I'll mark this as draft for the time being.

OzGav:
I had my Claude check and it all works now.

On the old values, leave them. They are only read when a book has no promoted artists, so they do no harm sitting there.

The thing to look at is search, which still only reads those old values. That gap already exists with Audiobookshelf so it is not something you have introduced, but it wants fixing in another PR to look at the artist table as well. We will need to cater for both ways until every provider has moved over.

fmunkes:
> I had my Claude check and it all works now.
> 
> On the old values, leave them. They are only read when a book has no promoted artists, so they do no harm sitting there.
> 
> The thing to look at is search, which still only reads those old values. That gap already exists with Audiobookshelf so it is not something you have introduced, but it wants fixing in another PR to look at the artist table as well. We will need to cater for both ways until every provider has moved over.

I believe this one is good now after having a personal discussion.

PR #6400, opened 2026-09-19
Add support for audiobook collections to the filesystem providers

fmunkes:

Stacked on https://github.com/music-assistant/server/pull/6399.

Adds support for collections (series) to the filesystem providers via tags. The series' title tag is series, and the sequence is series-part.

## Types of changes

## Checklist


OzGav:
The series and series-part tags are not standard, and every file format stores them differently. I tested it. It works for mp3, flac and wavpack, but not for m4a or m4b. ffprobe does not give us the custom tags in mp4 files, so series always comes back empty there, which is the format most audiobooks use. `_parse_mp4_tags` needs to read them in the same place it already reads `originaldate`.

We also need tests using real fixture files for all four formats rather than a dictionary of values. There are fixtures in tests/fixtures already, and `test_original_release_date_is_read_from_an_m4a_file` shows the pattern.

Please also raise a docs PR. It needs to specify the tags needed for this to work and that Picard does not write them.

fmunkes:
Fixed, thanks -  series and series-part are now read with mutagen in _parse_mp4_tags next to originaldate, and in the ID3, Vorbis and APEv2 parsers too, so nothing depends on ffprobe any more. I'll mark this as draft, like the base PR, let's keep the discussion over there.

The docs PR I'll do once we settle the implementation.

OzGav:
This looks good now as well but I think that the series is only written onto the book when the tag is present, so if someone removes the series tag from a file and rescans, the book stays in its old series with no way to get it out. Audiobookshelf sets it every time, including empty when there is no tag, so we should do the same here.

PR #6405, opened 2026-09-19
Move the ibroadcast item mapping to strings only

robsonke:

Solves a serious sync issue due to mismatches in the itemmapping between int's and string's. 

- https://github.com/music-assistant/support/issues/6463

## Types of changes

## Checklist


OzGav (commented):
This is Claude posting on Gav's behalf.

Short version is I don't think this is ready to merge yet. It's aimed at the right problem, but it breaks two other things on the way.

### What the PR gets right

The sync problem is real. When the sync works out which mappings are stale, it reads the stored provider item id out of the database, which is always text, and compares that against what the provider just listed. iBroadcast was handing back numbers, so 42 never matched "42", every mapping looked dead and got removed. iBroadcast isn't a streaming provider, so that cleanup definitely runs. Turning the ids into text in the parsers is the right answer.

The Various Artists change is right as well. The old code read `artist_obj["name"]` on an empty artist object, which raised and made the whole track get dropped. Good catch.

### Why I'd hold off

The ibroadcastaio library keeps the whole library in memory keyed by numbers. Have a look at `__json_to_dict` in its `client.py`, it does `result[main_key] = int(key)`. Every lookup after that is a plain dict get, so if you hand it text it quietly finds nothing and gives back an empty dict.

This PR turns the ids into text at the top of each parser, then passes those same variables straight into the artwork lookups. It remembers to convert back for tracks but not for the other two.

In `_parse_album` around line 341, `get_album_artwork_url` now gets `"101"`, the lookup misses, and the library raises `Album with id 101 not found`. Nothing guards that call, so it happens for every single album.

In `_parse_artist` around line 283, `get_artist_artwork_url` now gets `"42"` and hits the same wall. That one sits behind an `artwork_id` check, but most real artists do have artwork.

Both of those come out as ValueError, and `get_library_albums` and `get_library_artists` only catch KeyError, TypeError, InvalidDataError and IndexError. So it isn't a case of one item getting skipped. The error escapes the generator and the whole album and artist sync falls over on the first item. That's a worse outcome than the bug being fixed.

The fix is tiny and it's the same thing already done for tracks.

```python
path=await self._client.get_artist_artwork_url(int(artist_id)),
artwork_url = await self._client.get_album_artwork_url(int(album_id))
```

Personally I'd rather keep the number in a local variable and only make it text where the model actually needs text, rather than converting back and forth, but either way works fine.

### Why the tests went green

The fixture mocks all three artwork methods, so nothing ever checks what type of id they're handed. Given this PR is entirely about id types at that boundary, I think the fake client should behave like the real one and back its lookups with number keyed dicts instead of accepting anything. That would have failed straight away, and it would stop the same thing creeping back in later.

### Smaller things, none of them blocking

`test_various_artists_album_keeps_its_artist_id` only checks that the id is text. VARIOUS_ARTISTS_MBID is already a text constant, so that passes no matter what the code does. Worth asserting the actual value instead.

There's an older problem sitting right next to the changed lines. In `_parse_track`, `album` only gets set inside the `if track_obj["album_id"]` branch but it's read unconditionally a few lines further down, so it'll fall over with a NameError on a track with no album. Not this PR's doing, but it's close enough to be worth a separate fix.

The branch is behind dev and needs bringing up to date.

It's labelled bugfix. If this is broken in stable too then it probably wants backport-to-stable as well.

### On migrations

Worth saying out loud since the repo rules ask for it. No migration needed here. The `provider_item_id` column is TEXT NOT NULL, so SQLite had already turned those numbers into text as they were written. The rows on disk are fine, the mismatch only ever existed in memory.

### One caveat

I confirmed the two artwork problems by reading the ibroadcastaio 0.6.0 source rather than running it, because the sandbox wouldn't execute the package and the full test suite won't install in this environment. The code path is clear enough, but I haven't actually watched it fail.

---
_Generated by [Claude Code](https://claude.ai/code)_

robsonke:
Hi @OzGav, thanks for the feedback, pretty good actually. Important note: it's against ibroadcastaio 0.6.0 and we're on 0.8.0, which changed exactly those int/str lookups.

- In 0.8.0 get_artist, get_album, get_track and get_playlist all do .get(int(id), {}) — that landed in the 0.8.0 release.
- ibroadcastaio stays int-based, the provider translates to string. The library is a general iBroadcast client; the string requirement comes from MA's models and is fine, so the conversion belongs in the provider.

What I will change/am changing:

- Keep the native id in a local and call str() only where the model needs it, so the provider never passes a string to the library. That also drops the int() round-trips in _parse_track and _parse_playlist.


I'll create another PR for the album issue you spotted, plus three more places where a missing-artwork ValueError ends the whole listing instead of skipping one item (tracks, albums and playlists). And some more automated tests. But I prefer to have this one merged, it builds on top of these changes.

OzGav:
This looks good to me now Rob. Thanks. I can’t trigger a copilot review at the moment so that will happen soon as the next step.

OzGav (approved):

PR #6406, opened 2026-09-19
Use the loudness provided by Tidal

StollD:

As part of the playback info, Tidal also returns standard ReplayGain tags that bring the volume of the track to -18 LUFS. Using this information, Music Assistant can apply the correct volume normalization right away, without scanning the track. It also enables using the same gain for all tracks of an album, which (to my understanding) was not possible before.

## Types of changes

## Checklist


OzGav:
@jozefKruszynski

remon1496:
I always thought Tidal targeted -14 LUFS through their official apps. Out of curiosity, I wonder where you got the -18 from?

StollD:
I initially treated the gain values as if they normalized to -14 LUFS, but the tracks from Tidal were playing back noticeably quieter than local files or the official Tidal app. Changing it to -18 made them line up perfectly. I don't know why they use a different target in the API, but since -18 LUFS is also the loudness recommended by ReplayGain I went with it.

remon1496:
Thank you for the explanation!

music_assistant/providers/tidal/streaming.py
```diff
@@ -257,3 +259,10 @@ def _get_direct_url(self, stream_data: dict[str, Any]) -> tuple[str, str | None]
         if not urls:
             raise MediaNotFoundError("No stream URL found")
         return urls[0], codec
+
+    def _loudness_from_gain(self, value: float | None) -> float | None:
```
jozefKruszynski: -18.0 is a magic number and the method never touches self. Name the reference like Plex does with PLEX_LOUDNESS_REFERENCE_LUFS, and make it a module-level function next to the other module constants.


```py
# ReplayGain 2.0 reference level; Tidal's replay gain values are relative to it.
_REPLAYGAIN_REFERENCE_LUFS = -18.0

def _loudness_from_gain(value: float | None) -> float | None:
    """Convert a ReplayGain value to absolute loudness in LUFS."""
    if value is None:
        return None
    return _REPLAYGAIN_REFERENCE_LUFS - value
```

music_assistant/providers/tidal/streaming.py
```diff
@@ -164,6 +164,8 @@ async def _serve_manifest(_request: web.Request) -> web.Response:
             path=url,
             can_seek=True,
             allow_seek=True,
+            loudness=self._loudness_from_gain(stream_data.get("trackReplayGain")),
```
jozefKruszynski: The loudness is only set in memory on the StreamDetails and never persisted through the core's provider-loudness path. The builtin loudness analyzer skips a track only when a provider_loudness row exists in the database ([loudness_analysis/provider.py:158-166](https://github.com/music-assistant/server/blob/94b8d271f707064cb1c23a3af7283ca5a7c4441f/music_assistant/providers/loudness_analysis/provider.py#L158-L166)), so every Tidal play will still spawn an ffmpeg ebur128 process whose result is never used, because the in-memory Tidal value always wins at hydration. Plex does both, set the fields and persist ([plex/init.py:1063-1071](https://github.com/music-assistant/server/blob/94b8d271f707064cb1c23a3af7283ca5a7c4441f/music_assistant/providers/plex/__init__.py#L1063-L1071)); OpenSubsonic and filesystem persist too. Persisting also gets the isfinite and minimum-LUFS sanity guards for free, which the direct assignment bypasses.


```py
loudness = _loudness_from_gain(stream_data.get("trackReplayGain"))
loudness_album = _loudness_from_gain(stream_data.get("albumReplayGain"))
if loudness is not None:
    self.mass.create_task(
        self.mass.streams.audio_analysis.set_track_loudness(
            item_id=track.item_id,
            provider_instance_id_or_domain=self.provider.instance_id,
            loudness=loudness,
            loudness_album=loudness_album,
        )
    )
```

tests/providers/tidal/test_streaming.py
```diff
@@ -264,6 +264,45 @@ async def test_get_stream_details_with_dash_manifest_duplicate_registration(
...
+    stream_details = await streaming_manager.get_stream_details("123")
+
+    assert stream_details.loudness == -15.0
+    assert stream_details.loudness_album == -16.5
+
+
+async def test_get_stream_details_with_replaygain_track_only(
+    streaming_manager: TidalStreamingManager, provider_mock: Mock, mock_track: Mock
+) -> None:
+    """Test get_stream_details with ReplayGain values."""
```
jozefKruszynski: Copy-pasted docstring. This test covers the track-only case, so say so. If the persistence call is added, assert mass.streams.audio_analysis.set_track_loudness is scheduled with the converted values, and that it is not called when Tidal omits the keys.

jozefKruszynski (changes requested):
Hi @StollD,
First of all, thank you very much for contributing.
I left a couple of comments, so please take a look.

Thank you

music_assistant/providers/tidal/streaming.py
StollD: Done

music_assistant/providers/tidal/streaming.py
StollD: Done

tests/providers/tidal/test_streaming.py
StollD: Done. I tried to follow how the OpenSubsonic tests do it, hope that's OK.

jozefKruszynski:
Looks good, can you please update the branch

jozefKruszynski (approved):
LGTM, Thank you for the contribution

OzGav:
Please review the documentation and open a PR there as well. You should start on this page https://www.music-assistant.io/faq/tech-info/#where-the-loudness-value-comes-from

StollD:
Documentation PR opened: https://github.com/music-assistant/music-assistant.io/pull/1052

@jozefKruszynski Thank you for your feedback!

PR #6407, opened 2026-09-19
Fix content type for sync group member streams

Carunga:
Sync group member stream URLs (e.g. `/slimproto/multi?player_id=...&fmt=flac&child_player_id=...`) have no file extension, so deriving the mime type from the URL produced garbage (e.g. `audio/103:8098/slimproto/multi`). Some players - notably classic Squeezebox hardware such as the Squeezebox Radio - then fail to decode the stream and stop immediately, while squeezelite tolerates it.

Pass the member codec's mime type explicitly from the group `play_media` path (and from the single-track-repeat path), keeping the URL-derived value only as a fallback for plain file URLs.

Tested with a Squeezebox Radio (FW 8.5.3) and a wired squeezelite in a native sync group.


Carunga:
Opened by mistake - closing it. The branch stays in my fork (Carunga/server:fix/squeezelite-sync-member-mime).

Context: sync-group member stream URLs (/slimproto/multi?...&fmt=flac&child_player_id=...) have no file extension, so deriving the mime type from the URL produced garbage (e.g. audio/103:8098/slimproto/multi). Classic Squeezebox players (e.g. Squeezebox Radio) then failed to decode the group stream, while squeezelite tolerated it. Passing the member codec's mime explicitly fixed it. Tested with a Radio (FW 8.5.3) + wired squeezelite in a native sync group.

PR #6408, opened 2026-09-19
Use LMS-faithful server-side sync for squeezelite sync groups

Carunga:
Replaces the ad-hoc sync with Lyrion Music Server's server-side corrected sync, ported into aioslimproto (`SyncGroup`).

- coordinated start at a common player-clock instant (`SyncGroup.start()`, sends `strm u` per player)
- steady-state correction via `SyncGroup.check()`: recent play-points, reference-player selection (never skips a non-skippable player), deviation gates, skip-ahead / pause-for
- per-player `play_delay` mapped from the existing "Audio synchronization delay correction" setting
- drops the WiiM pause workaround and the `pause_and_unpause` helper

**Stacked / depends on:**
- music-assistant/aioslimproto#557 (library; must be released first)
- #6407 (mime fix; this branch is based on it, so the diff will shrink to the sync-only changes once #6407 merges)

**Note:** `aioslimproto` in `manifest.json` is pinned to the fork commit as a placeholder until #557 is released; it will be updated to the released version before merge. Kept as draft until then.

Tested on real hardware: wired squeezelite + Squeezebox Radio synced to ~1 ms.


Carunga:
Opened by mistake - closing it (it was a draft, stacked on the mime change). The branches stay in my fork (Carunga/server:feat/squeezelite-lms-sync, and fix/squeezelite-sync-member-mime).

Context: replaces the provider's ad-hoc sync (average-elapsed corrector + fixed pause+200 ms start) with LMS's server-side corrected sync via aioslimproto's new SyncGroup. Real-hardware verified at ~1 ms. It depends on the aioslimproto work; the manifest pin was a placeholder.

PR #6416, opened 2026-09-20
Add native FeiNiu Music provider

neqq3:

Adds an experimental, read-only provider for **FeiNiu Music**, the personal music-library service for fnOS.

This allows users to browse and play their FeiNiu-managed music library directly in Music Assistant, without deploying a separate protocol bridge. The provider connects to the music service over HTTP(S) using a regular music account and password; it does not require NAS administrator access.

### Supported features

- Library synchronization and browsing for tracks, albums, artists and playlists.
- Search, artwork, and plain-text or line-synchronized lyrics.
- Track playback and multiple provider instances with separate music accounts.

Provider authentication and caches are isolated per instance. Detail and artwork requests are checked against account-filtered library results, and server-side playback refusals are respected. The integration is read-only, with no playlist or metadata write operations.

No MA core changes or additional runtime dependencies are included.

**Related issue (if applicable):** None.

## Types of changes

## Checklist

## Testing

The final candidate passes **101 provider tests** and `pre-commit run --all-files`, including mypy. The provider tests also pass on a clean checkout with the patch applied and do not require a NAS or credentials. In addition, the upstream GitHub Actions test workflow for this PR completed successfully.

Live integration testing used **FeiNiu Music 1.0.1 (0.8.41)** with three accounts, including a restricted-library account. Testing covered synchronization, account scoping, artwork, FLAC/MP3/OGG samples, and a 150-entry playlist.

Player-state checks covered pause/resume, seeking, completion of two full tracks, and automatic advance. These checks preceded the final reconfiguration and stream-cleanup changes, which have offline regression coverage but were not followed by another live playback run. Playback was assessed through player state and progress, without a listening check for that run.

## Limitations

- This is an unofficial integration marked experimental. Only the music-service version above has been tested; large libraries, cross-process concurrent logins and long-term natural session expiry remain unverified.
- FN ID, NAS OAuth, CUE split-track playback and arbitrary reverse-proxy path prefixes are not supported.
- Changing the server URL or username requires removing the source and adding it again. Password updates for the existing account are supported.
- Metadata and artwork already cached by MA or its clients may remain after permission changes or source removal; immediate removal of previously delivered data is not guaranteed.

On the tested MA baseline, provider-only search can omit tracks already synchronized to the library. This was reproduced independently of FeiNiu and is kept separate from this PR; no core workaround is included.


music_assistant/providers/feiniu_music/client.py
```diff
@@ -0,0 +1,553 @@
...
+from weakref import WeakValueDictionary
+
+import aiohttp
+
+_LOGIN_LOCKS: WeakValueDictionary[tuple[asyncio.AbstractEventLoop, str], asyncio.Lock] = (
+    WeakValueDictionary()
+)
+
+
+class FeiNiuError(Exception):
```
OzGav: Please change these to inherit from the Music Assistant errors rather than from Exception. Each of the seven has a direct equivalent already, and `_raise_error` in `provider.py` is doing nothing but translating them back into those same errors. Inheriting from them directly is the established pattern in the codebase.

Once that's done you can delete `_raise_error` entirely. The only real work is moving the backoff values for `RateLimited` and `ResourceTemporarilyUnavailable` across to where you raise them in this file. You'll also keep the original error attached, which the from None in `_raise_error` currently discards and which makes these much harder to trace from a log.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,525 @@
...
+        for kind, attribute, parser in (
+            (MediaType.TRACK, "tracks", parse_track),
+            (MediaType.ALBUM, "albums", parse_album),
+            (MediaType.ARTIST, "artists", parse_artist),
+            (MediaType.PLAYLIST, "playlists", parse_playlist),
+        ):
+            if kind not in media_types:
+                continue
+            found = []
+            async for item in self._pages(
```
OzGav: Minor one. When the search reaches the limit it breaks out of `_pages` and leaves that generator hanging around for the garbage collector to tidy up later. You already wrap the same situation in `aclosing` over in `get_audio_stream` on line 293, so let’s have the two match.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,525 @@
...
+                if len(found) >= limit:
+                    break
+            setattr(results, attribute, found)
+        return results
+
+    async def resolve_image(self, path: str) -> bytes:
+        """Proxy authenticated artwork as bytes, never as a credential-bearing URL."""
+        if "/" not in path:
+            # Earlier development builds stored bare cover IDs in MA. Resolve
+            # those only through listed artwork; do not trust cover HTTP success.
```
OzGav: When a cover id has no slash in it, this loops over all four collections, and any one of those can read the whole library. So a single old thumbnail can cost four full library reads.

This branch only exists for ids saved by your earlier development builds, so could it just go? If you'd rather keep it, looking only at collections that are already loaded would avoid the fetches.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,525 @@
...
+        spec = data.get("audioSpec") or {}
+        return StreamDetails(
+            provider=self.instance_id,
+            item_id=item_id,
+            stream_type=StreamType.CUSTOM,
+            audio_format=audio_format(spec),
+            duration=number(track.get("duration")) // 1000 or None,
+            size=number(spec.get("size")) or None,
+            can_seek=False,
+            allow_seek=True,
```
OzGav, line 236: `can_seek` is hardcoded to False, which means every seek makes Music Assistant decode from the start of the track again. That's rough on long tracks. `media_prefix` already reads `Accept-Ranges` and reports it as `accepts_bytes` on line 445 of `client.py`, so you may already have what you need to know whether the server supports it. Happy for this to be a later PR once you've confirmed the behaviour against a real server.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,525 @@
...
+            for row in rows:
+                item = self._bind_images(parser(row, self.instance_id))
+                if item.item_id in items:
+                    raise InvalidDataError("FeiNiu returned duplicate collection identifiers")
+                items[item.item_id] = item
+            self._check_open()
+            self._collections[kind] = (monotonic(), items)
+            return items
+
+    async def _listed(self, kind: str, item_id: str) -> MediaItem:
```
OzGav: This is the one I'd most like to see addressed before merge.

`_listed` calls `_collection`, and once the 30 second cache has expired _collection reads the entire library one page at a time. `get_stream_details` goes through `_listed` on line 255, so this sits directly in front of playback.

That's the part that worries me. Tracks run to several minutes, so the 30 second cache has always expired by the time the next one starts. Playing an album straight through means reading the whole library again before every track. With the throttle at four requests a second and 100 items a page, a 5,000 track library is about 50 requests, so roughly 12 seconds of waiting before each track begins. At 10,000 tracks it's nearer 25 seconds. Browsing has the same problem, since `get_album`, `get_album_tracks` and the rest all come through here too.

I do understand why it's built this way, and you're right that the detail endpoints can't be trusted for permissions on their own. But the check only needs to know whether an id is allowed, not hold the whole parsed item. Keeping a set of allowed ids with a much longer lifetime, and fetching the item itself from `_client.detail`, would give you the same protection without putting a library read in front of every track.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,525 @@
...
+    async def get_stream_details(self, item_id: str, media_type: MediaType) -> StreamDetails:
+        """Return a server-owned stream with a stable track identity."""
+        if media_type != MediaType.TRACK:
+            raise UnplayableMediaError("FeiNiu supports track playback only")
+        await self._listed("track", item_id)
+        data = await self._call(lambda: self._client.detail("track", item_id))
+        track = data.get("track")
+        if not isinstance(track, dict) or track.get("guid") != item_id or track.get("isCue"):
+            raise UnplayableMediaError("Missing track or unsupported CUE track")
+        summary, _ = await self._call(lambda: self._probe_audio(item_id))
```
OzGav: This pulls 4096 bytes of the track and throws them away, then `get_audio_stream` opens the stream again immediately afterwards. `get_audio_stream` already runs the same check on lines 387 to 396 of `client.py`, so this is an extra request for every track to catch something that gets caught anyway. Is it needed?

OzGav (changes requested):
Thanks for the new provider! Looks pretty good already. A few things below and also we need icons in colour and monochrome SVG

neqq3:
Thanks for the detailed review — it was very helpful.

I’ve pushed the requested changes:

the FeiNiu client errors now inherit directly from the corresponding Music Assistant error types, and the provider-side conversion layer has been removed;
search pagination is explicitly closed on early exit;
the development-era bare artwork ID compatibility path has been removed;
the duplicate audio preflight request has been removed;
account-visible membership is now separated from per-item metadata, so normal per-item detail expiry no longer causes a full library scan before every track;
colour and monochrome SVG provider icons have been added.

I’ve left native seeking unchanged for now, pending verification of the real service’s Range behaviour.

One detail I’d like your opinion on is the membership refresh policy. The current implementation keeps the account-visible membership set for 15 minutes, and a normal full library sync also refreshes it. Once that membership cache itself expires, the next lookup still performs a full list refresh before continuing.

This avoids scanning the library before every track, but can still cause an occasional blocking refresh after expiry. Would that be acceptable for the initial provider, or would you prefer membership refresh to be driven by sync/background work so playback never waits on a full refresh?

OzGav:
Please comment on each conversation with your resolution and resolve it.

music_assistant/providers/feiniu_music/client.py
neqq3: Changed the client exceptions to inherit directly from the corresponding MA errors, removed `_raise_error`, and moved the backoff values to the client raise sites.

JSON decoding errors and ordinary connection failures now retain their original cause. I kept local `from None` handling for aiohttp response/payload errors and disconnects carrying incomplete response data, as these can include raw response fragments. The error messages still include the exception type and status.

music_assistant/providers/feiniu_music/provider.py
neqq3: Wrapped the search iterator in `aclosing`, matching the audio-stream code. It now closes when the search stops at the requested limit.

music_assistant/providers/feiniu_music/provider.py
neqq3: Removed the entire development-only bare cover-ID compatibility branch, including the loops over all four collections. Image resolution now uses references that identify the provider instance and owner item.

music_assistant/providers/feiniu_music/provider.py
neqq3: I confirmed that the real stream endpoint honours byte-range requests, and tested a local HTTP-based seek implementation for FLAC and OGG.

It reduced the data read for a later FLAC seek, but the prototype also involved changes to session validation, redirect handling and FFmpeg credential logging. Including session preparation, my limited local tests did not show a consistent improvement in time to first decoded audio.

For this PR, I’m keeping the existing custom stream and MA’s seek fallback. The HTTP playback change is better reviewed separately rather than added to the initial provider.

music_assistant/providers/feiniu_music/provider.py
neqq3: Further testing with a restricted account on Music 1.0.1 showed that the album, artist and playlist endpoints already reject or filter inaccessible objects. Track metadata is different: it can still be returned, but exposes an account-specific `accessStatus`, also present on playlist rows.

I’ve removed the full-library membership checks and the 15-minute refresh. Track metadata must have `accessStatus=0` before it is parsed or cached. Playlist entries use their own status without extra per-track requests. Artwork resolution checks that the owner is accessible and that the requested image belongs to it.

Full collection reads are now limited to library sync/enumeration. The normal detail, playback and artwork paths made no full-collection requests when tested against the real service.

Per-item detail results are cached briefly to avoid repeated requests for the same object; when that cache expires, only that item is fetched again, not the library.

music_assistant/providers/feiniu_music/provider.py
neqq3: Removed the separate audio preflight request. The actual stream request still validates the response before yielding audio, including JSON errors returned with HTTP 200, so the opening bytes are no longer fetched twice.

music_assistant/providers/feiniu_music/provider.py
```diff
@@ -0,0 +1,479 @@
...
+        """Return native artist details or MA's local unknown-artist placeholder."""
+        if prov_artist_id == UNKNOWN_ARTIST:
+            return unknown_artist(self.instance_id)
+        return cast("Artist", await self._get_item("artist", prov_artist_id))
+
+    async def get_playlist(self, prov_playlist_id: str) -> Playlist:
+        """Return an owned, read-only playlist from the native service."""
+        return cast("Playlist", await self._get_item("playlist", prov_playlist_id))
+
+    async def get_album_tracks(self, prov_album_id: str) -> list[Track]:
```
OzGav: This parses every row the relationship returns, but `_playlist_items` on line 393 filters the same kind of rows down to `accessStatus == 0` before keeping any of them. Both come back from `related()` and the same `/track/{kind}-detail/list` shape, so if album relations carry `accessStatus` the way playlist ones do, a restricted track would stay hidden in a playlist but still show its title in an album listing.

Playback and artwork are both safe either way, since those go through `_detail` and get the status check there. It's the metadata that would leak. Given the trouble you went to on the playlist side, including the test asserting hidden titles never reach the cache, could the same filter go on the album path? And if album relations genuinely don't return the field, a short comment saying so would save the next person checking.

OzGav (commented):
This is looking good. Just one thing to think about after your rework.

PR #6418, opened 2026-09-20
Improved syncing in case it fails halfway on a single item

robsonke:

- Improved syncing in case it fails halfway on a single item
- Improved test coverage

## Types of changes

## Checklist


music_assistant/providers/ibroadcast/__init__.py
```diff
@@ -125,6 +126,8 @@ async def get_library_artists(self) -> AsyncGenerator[Artist]:
             except (KeyError, TypeError, InvalidDataError, IndexError) as error:
                 self._report_skipped_item(MediaType.ARTIST, artist, "artist_id", error)
                 continue
+        if await self._has_various_artists():
```
OzGav, line 138: Now that Various Artists is handed out as a real library artist, people can open it in the UI, and both lookups behind that page choke on its id. `get_artist_albums` (line 133) and `get_artist` (line 162) both start with `int(prov_artist_id)`, and the Various Artists id is the MusicBrainz string, so `int()` raises a ValueError. The caller in `controllers/music/media/base.py` only swallows MediaNotFoundError, so the ValueError travels all the way up and the artist page fails to load.

music_assistant/providers/ibroadcast/__init__.py
OzGav, line 211: The playlist listing is the only one of the four that still has no per item guard. Albums,
artists and tracks all catch a bad item, report it as skipped and carry on, but a playlist
that is missing "type" or "name" raises a KeyError straight out of the generator and ends
the whole playlist sync. Since that is exactly the failure this PR is fixing everywhere
else, it would be good to close it here too.

OzGav: Albums, tracks and playlists now survive an artwork lookup that raises, but the artist one here was left out. The client raises a ValueError from `get_artwork_base_url` whenever the account settings come back without an "artwork_server" entry, and ValueError is not in the list the listing catches, so a single artist would abort the entire artist sync. Please wrap this the same way as the other three.

music_assistant/providers/ibroadcast/__init__.py
```diff
@@ -337,10 +328,11 @@ async def _parse_album(self, album_obj: dict[str, Any]) -> Album:
         # iBroadcast doesn't seem to know album type - try inference
         album.album_type = infer_album_type(name, version)
 
-        # There is only an artwork in the tracks, lets get the first track one
-        artwork_url = await self._client.get_album_artwork_url(album_id)
-        if artwork_url:
-            album.metadata.images = UniqueList([self._get_artwork_object(artwork_url)])
+        # There is only an artwork in the tracks, lets get the first track one.
+        # The client raises when it finds none, which is no reason to drop the album.
+        with suppress(ValueError):
```
OzGav, line 342: `suppress(ValueError)` here, and in the same shape at lines 403 and 439, catches more than just missing artwork. The client raises the same ValueError for "Library not loaded" and "Artwork base URL not found in settings". If either of those ever happens, every single item quietly loses its artwork and there is nothing in the log to explain it. Catching it and dropping a debug line would keep the behaviour and leave a trail.

music_assistant/providers/ibroadcast/__init__.py
```diff
@@ -443,13 +435,36 @@ async def _parse_playlist(self, playlist_obj: dict[str, Any]) -> Playlist:
...
+            playlist.metadata.images = UniqueList(
+                [self._get_artwork_object(await self._client.get_playlist_artwork_url(playlist_id))]
+            )
         if "description" in playlist_obj:
             playlist.metadata.description = playlist_obj["description"]
         return playlist
 
+    async def _has_various_artists(self) -> bool:
+        """Return whether any album or track in the library maps to Various Artists."""
+        if any(album["artist_id"] == 0 for album in (await self._client.get_albums()).values()):
```
OzGav: Both of these read the "artist_id" key directly, and neither call sits inside a try block.
The client builds its library dicts by zipping each row against the field map, so a short
row simply has no "artist_id" key at all. That is why `_parse_track` guards with
`if "artist_id" in track_obj` and why the listings catch KeyError per item. Here one such
row raises a KeyError that escapes `get_library_artists` entirely, which kills the artist
sync after every artist has already been yielded. That is the same abort this PR set out
to fix.

```
    if any(album.get("artist_id") == 0 for album in (await self._client.get_albums()).values()):
        return True
    return any(track.get("artist_id") == 0 for track in (await self._client.get_tracks()).values())
```

A test with an album and a track that have no "artist_id" key would pin this down.

tests/providers/ibroadcast/test_parsers.py
```diff
@@ -242,3 +260,243 @@ async def test_listed_item_ids_are_all_text(provider: IBroadcastProvider, listin
...
+    """An empty playlist has no track to take artwork from."""
+    provider._client.missing_artwork.add("playlist")
+
+    playlists = [item async for item in provider.get_library_playlists()]
+
+    assert [playlist.item_id for playlist in playlists] == ["5001"]
+    assert not playlists[0].metadata.images
+
+
+async def test_one_unreadable_track_does_not_end_the_listing(provider: IBroadcastProvider) -> None:
```
OzGav, line 323: The name and the docstring talk about an unreadable track, but neither track in this fixture is unreadable. One has no album and the other is a straight copy of the good one, so what this actually checks is that a track with no album does not stop the listing, which is already covered by test_track_without_an_album_is_parsed just above. Either drop it or make one of the two genuinely unparseable, for example by removing its "title".

tests/providers/ibroadcast/test_parsers.py
```diff
@@ -242,3 +260,243 @@ async def test_listed_item_ids_are_all_text(provider: IBroadcastProvider, listin
...
+    provider._client.tracks = {1001: {**TRACK, "album_id": 0}, 1002: {**TRACK, "track_id": 1002}}
+
+    tracks = [item async for item in provider.get_library_tracks()]
+
+    assert [track.item_id for track in tracks] == ["1001", "1002"]
+
+
+async def test_stream_url_asks_for_the_original_upload(provider: IBroadcastProvider) -> None:
+    """The bitrate segment defaults to 128kbps, so it is swapped for the original format."""
+    provider._client.stream_url = (
```
OzGav: This sets stream_url to the exact value the fixture already has, so the line does not do
anything. Removing it makes it clearer that the default url is the one under test.

OzGav (changes requested):
Some thoughts from me.

robsonke:
Thanks, working on them in the mean time.

music_assistant/providers/ibroadcast/__init__.py
robsonke: Agreed, made the change in the first commit, rest is pending.

music_assistant/providers/ibroadcast/__init__.py
robsonke: Agreed, made the change in the first commit, rest is pending.

music_assistant/providers/ibroadcast/__init__.py
robsonke: Instead of a debug line per item, I now check this once in handle_async_init. Should give the same effect

music_assistant/providers/ibroadcast/__init__.py
robsonke: part of the other commits

robsonke: part of the other commits

tests/providers/ibroadcast/test_parsers.py
robsonke: Fixed, track 1001 now has its title removed so it is genuinely not parsable, and the test asserts only 1002 comes back

tests/providers/ibroadcast/test_parsers.py
robsonke: Removed

robsonke:
All done, please review and thanks for the feedback!

OzGav (approved):
Thanks Rob!

PR #6422, opened 2026-09-20
Skip loudness analysis for tracks over 10 minutes up front

newman1982:
## What's broken

`LoudnessAnalysisProvider` never sets `max_analysis_duration`, so the base
class's duration gate in `start_analysis()` — the one that correctly makes
`SonicAnalysisProvider` and `SmartFadesProvider` decline multi-hour files
outright via `ACCUMULATING_ANALYSIS_MAX_DURATION_SECONDS` — never applies to
it.

The only limit loudness has is the inline check in `process_pcm_chunk()`:

```python
if data.chunks_received >= MAX_DURATION_SECONDS:
    # cap the analysis window for very long streams
    await self._send_eof(data)
```

`_send_eof()` only closes the write side of *this provider's own* internal
ffmpeg process used for the ebur128 measurement. It doesn't tell the
`AudioAnalysisController` to evict the provider from the shared streaming
session. So the session stays "active" for loudness even though it stopped
listening after ~10 minutes, and the controller keeps decoding the source
stream all the way to its real EOF.

## Impact

For a normal song this is invisible. For a multi-hour audiobook chapter or
DJ mix, it means a background-scan concurrency slot sits occupied for the
*entire* track length, decoding audio nobody is measuring anymore. On a
library with a sizeable audiobook collection filed under the regular music
path (a fairly common setup), this can end up occupying every concurrency
slot at once and stall regular track analysis for hours — we hit this
directly: 16/16 slots pinned on ~56-minute audiobook chapters, sonic/loudness
throughput for the rest of the library dropping to single digits per hour
until that queue drained.

The docs already describe the intended behavior at
https://www.music-assistant.io/audio-analysis/loudness-analysis/ ("longer
than 10 minutes of analysed audio... are skipped"), so this is a gap
against documented behavior, not a change in intent.

## Fix

Set `max_analysis_duration = MAX_DURATION_SECONDS` on
`LoudnessAnalysisProvider`, same pattern already used by the other two
accumulating-state providers. The base class then rejects the session before
it ever starts, instead of accepting it and giving up 10 minutes in.

## Testing

- Ran against the real (unmodified besides this change) `start_analysis()` /
  `LoudnessAnalysisProvider` classes with a stubbed `mass`/session context:
  a 3360s (56 min) track is now declined at `start_analysis()` before any
  session/stream work begins; a normal 180s track still proceeds past the
  duration gate unchanged.
- Reproduced live on a real deployment with ~370 hour-long audiobook files
  mixed into the music library: before the fix, all background-analysis
  concurrency slots filled with these files (ffmpeg processes alive for
  250–330s+ and climbing); after the fix and a restart, the same slots
  immediately filled with normal 3–5 minute tracks with healthy turnover
  (processes completing in 5–170s), and no further audiobook files were
  picked up by loudness analysis.
- `ruff check` passes on the changed file.

---

Small disclosure since it's relevant here: I used an AI coding assistant (Claude Code) to dig through the actual runtime behavior, trace this back to the missing `max_analysis_duration`, and put this PR together — it found the root cause and worded the fix/description, I reviewed and I'm submitting it. I wouldn't have tracked this one down on my own, so credit where it's due; happy to answer questions or make changes if the approach isn't what you'd want here.


OzGav:
Do not submit autonomous AI generated PRs. This will be closed soon unless this is fixed.

newman1982:
Thanks. I will close it right now

PR #6427, opened 2026-09-21
Keep the stale mapping pass from emptying a library on mismatched ids

RyanAtTanagra:

Note: this may "fix" what is considered a non-issue and may just get deleted, but I feel bad my previous PR caused an unintended issue and thought maybe it should have had stronger safeguards.

The stale mapping pass added in [#6355](https://github.com/music-assistant/server/pull/6355) compares the item ids a provider listed during a sync with the ids stored on its mappings. The library stores them as text, and iBroadcast listed numbers, so none of them matched. Every iBroadcast mapping looked stale and was removed, and removing an item's last mapping removes the item, which emptied the library ([music-assistant/support#6463](https://github.com/music-assistant/support/issues/6463)). [#6405](https://github.com/music-assistant/server/pull/6405) fixed the iBroadcast side. This makes the pass itself safe against the next provider that does the same.

- The listed and skipped ids are compared as text, so a provider that lists numbers matches the mappings stored for it.
- The sync has already just matched every item in the pass to an item the provider listed, so each keeps at least one mapping to it. Removing the stale ids should never take every mapping for an item, so log a warning instead of possibly emptying the library.

A genuinely stale mapping never trips the check: the item keeps its mapping to the listed item and only loses the old one.

I checked the other providers whose sync reaches this pass (Emby, Jellyfin, OpenSubsonic, Plex, TeddyCloud, Audiobookshelf, the podcast providers, builtin) and did not find another one that lists numbers, so this is a safeguard rather than a fix for a known case.

The two new tests both fail on current `dev`. With only the text comparison in place, the second one still fails, so each safeguard is covered on its own.

Tested live on a 2.10.4 install with a Plex library of about 22,000 tracks. I moved an album out of Plex and back between two syncs, so Plex gave it and its 14 tracks new ids and left the library with mappings to the deleted ones. The next sync removed those stale mappings and kept the new ones: the album and every track kept their library id, the library counts did not change, and the safeguard did not trigger.

I don't have an iBroadCast provider to test against, unfortunately.

- [music-assistant/support#6463](https://github.com/music-assistant/support/issues/6463)

## Types of changes

## Checklist


OzGav:
@robsonke

robsonke (approved):

OzGav (approved):

OzGav:
Thanks @RyanAtTanagra !

PR #6429, opened 2026-09-21
Fix resume position after fallback announcements

sickkick:

Fallback announcements interrupted playback with a device-only stop. That skipped
`player_queues._handle_stop`, so `resume_pos` was never parked. While an announcement
is in progress, player→queue updates are suppressed, so the queue can still look
`PLAYING`. `resume()` then used wall-clock `corrected_elapsed_time` (which keeps
advancing during the announcement) instead of the real media position — often seeking
past the track or effectively restarting from the start (especially on Squeezelite
and other non-native announce players).

This change:

- Stops through `player_queues._handle_stop` when an active MA queue exists before
  playing the announcement, so `resume_pos` is parked.
- During `ATTR_ANNOUNCEMENT_IN_PROGRESS`, prefers parked `resume_pos` / `elapsed_time`
  over the live corrected clock, and clamps seeks past track duration.

- related issue <link to support issue after you open it>

## Types of changes

## Checklist


OzGav:
Thanks for digging into this. I went through the chain and the root cause is real. The fallback announcement stops the device directly, so the queue never records where it was. The queue also ignores player updates while an announcement runs, so it still thinks it is playing. Resume then trusts the clock and lands too far into the track. That matches what you describe and the fix is on the right track.

A few things I would like sorted before this goes in.

1. The description still contains a placeholder where the related issue link should go. Please open that issue or remove the line. Please also give the whole description a read through. It looks like it was drafted with an AI tool and not checked afterwards.

2. The change calls a private method on the queue controller from the player controller. The player controller already stops queues through the public stop method elsewhere. If you chose the private one to skip the user permission check, say so in the comment, otherwise please use the public one.

3. Stopping through the queue does a lot more than stopping the device. It cancels the preload and enqueue next timers, tears down the stream session and cleans up the audio buffers. Before this change the announcement left all of that in place. That may well be the better behaviour, but it is a bigger change than the description suggests and nothing in the tests covers it. Please mention it in the description and add a test if you can.

4. The clamp that caps the resume position at the track length will seek to the very end of the track, which just makes the queue skip to the next item straight away. Now that the parked position is used, I am not sure this clamp is needed at all. If you want to keep it, resuming from the start would be a better fallback.

5. Because the playing branch is skipped during an announcement, the fade in is no longer forced off. A player that has been idle for more than a minute will now fade back in after an announcement. That may be fine, just flag it as intended.

6. The new check in resume only works because the restore step runs before the announcement flag is cleared. Nothing in the code makes that ordering explicit. A short comment near the flag reset in the announcement code would help the next person who touches it.

The branch is also behind dev again and needs another merge.

sickkick:
> Thanks for digging into this. I went through the chain and the root cause is real. The fallback announcement stops the device directly, so the queue never records where it was. The queue also ignores player updates while an announcement runs, so it still thinks it is playing. Resume then trusts the clock and lands too far into the track. That matches what you describe and the fix is on the right track.
> 
> A few things I would like sorted before this goes in.
> 
> 1. The description still contains a placeholder where the related issue link should go. Please open that issue or remove the line. Please also give the whole description a read through. It looks like it was drafted with an AI tool and not checked afterwards.
> 2. The change calls a private method on the queue controller from the player controller. The player controller already stops queues through the public stop method elsewhere. If you chose the private one to skip the user permission check, say so in the comment, otherwise please use the public one.
> 3. Stopping through the queue does a lot more than stopping the device. It cancels the preload and enqueue next timers, tears down the stream session and cleans up the audio buffers. Before this change the announcement left all of that in place. That may well be the better behaviour, but it is a bigger change than the description suggests and nothing in the tests covers it. Please mention it in the description and add a test if you can.
> 4. The clamp that caps the resume position at the track length will seek to the very end of the track, which just makes the queue skip to the next item straight away. Now that the parked position is used, I am not sure this clamp is needed at all. If you want to keep it, resuming from the start would be a better fallback.
> 5. Because the playing branch is skipped during an announcement, the fade in is no longer forced off. A player that has been idle for more than a minute will now fade back in after an announcement. That may be fine, just flag it as intended.
> 6. The new check in resume only works because the restore step runs before the announcement flag is cleared. Nothing in the code makes that ordering explicit. A short comment near the flag reset in the announcement code would help the next person who touches it.
> 
> The branch is also behind dev again and needs another merge.

@OzGav I believe I have addressed your concerns? If it looks good I will mark it ready for review?

OzGav:
Thanks for the quick turnaround. Most of this is sorted now.

One thing is left. I added point 7 to my comment after you replied, so you may have missed it. The comments in the code are too long, and they got longer in this round. The block above the stop call is now eight lines for a four line if and else. Please cut each comment down to a line or two that says why, not what, and only have it at all if a future dev won’t understand from just reading the code. The resume comment and the one above the flag reset can shrink the same way. Please also drop the mention of test_stop_teardown from the production code, since a test file name does not belong there. The test docstrings repeat the same explanation and can be trimmed too.

PR #6432, opened 2026-09-22
Fix Snapcast volume/mute routing to idle native player when Sendspin is active

tortfeaser:

A snapclient's native volume/mute control is a private software gain applied
only to its own decoded audio stream. It has no effect on audio actually
being rendered by another protocol (e.g. Sendspin) sharing the same physical
output. `Player.volume_control` / `mute_control` (in `models/player.py`)
unconditionally prefer a native control whenever the player advertises
`PlayerFeature.VOLUME_SET` / `VOLUME_MUTE` at all, without checking
`active_output_protocol`. Snapcast always advertises both, so when Sendspin
is the active output protocol on a Snapcast player, volume/mute commands
were silently redirected to the idle snapclient instead of the Sendspin
player actually producing sound - matching the symptom reported in the
linked issue (commands reach snapclient, zero volume-related log lines on
the Sendspin side).

`SnapCastPlayer.supported_features` now hides `VOLUME_SET` / `VOLUME_MUTE`
whenever a foreign protocol is the active output. This makes
`volume_control` / `mute_control` fall through to the existing
active-protocol resolution path (`Player._get_protocol_player_for_feature`),
which already correctly redirects to the active protocol player once native
is out of the running - no changes needed there. The externally-visible
feature set is unaffected, since `__final_supported_features` re-derives
`VOLUME_SET` / `VOLUME_MUTE` from the (now-correct) resolved control rather
than the raw flag.

This is scoped to the Snapcast provider only, since only it has this
stream-local (vs. device-wide) native volume characteristic among current
protocol-capable native players. No changes to shared player/model code.

- related issue #6468

## Types of changes

## Checklist

- [x] The code change is tested and works locally. Verified against real hardware: a physical speaker running both a snapclient and a Sendspin client simultaneously, with Sendspin as the active output. Before this change, volume/mute commands only reached the idle snapclient (no audible effect). After, MA logs show `Redirecting volume command to protocol player Sendspin` / `Redirecting mute command to protocol player Sendspin`, and the change is audible on the speaker.
- [x] `pre-commit run --all-files` passes. Confirmed via this PR's own CI `lint` job (ruff, mypy, and the rest of the repo's pre-commit hooks), which is green.
- [x] `pytest` passes, and tests have been added/updated under `tests/` where applicable. Added `tests/providers/snapcast/test_player_volume_routing.py`; confirmed passing via this PR's `test` CI job.
- [x] For changes to shared models, the companion PR in `music-assistant/models` is linked. Not applicable - no shared model changes.
- [x] For changes affecting the UI, the companion PR in `music-assistant/frontend` is linked. Not applicable - no UI changes.
- [ ] I have [raised a PR against the documentation repository](https://github.com/music-assistant/music-assistant.io/blob/main/CONTRIBUTING.md) targeting the main or beta branch as appropriate. Not applicable - internal bugfix, no user-facing documentation change.


OzGav:
This appears to be unreviewed AI generated code. Code comments are excessive, PR description is borderline unreadable, PR description has modified the template and contains information in the wrong places.

tortfeaser:
I've revised the comment, rewritten the description in plain terms, and moved test detail out of the checkboxes. Appreciate the further review @OzGav. Apologies for wasting time.

MDK

music_assistant/providers/snapcast/player.py
```diff
@@ -138,6 +138,18 @@ def group_members(self) -> list[str]:
 
         return []
 
+    @property
+    def supported_features(self) -> set[PlayerFeature]:
+        """Return the supported features of the player."""
+        if self.active_output_protocol and self.active_output_protocol != "native":
```
OzGav: This check only works once Sendspin has already been marked as the active output. Some code asks which volume control to use before that happens. Announcements do this in announcements.py line 934. An idle speaker that plays an announcement over Sendspin still gets told to use the Snapcast volume, so the announcement volume goes to the idle snapclient. That's the same bug this PR fixes, just in a different place.

A fuller fix would check which output is actually playing the audio, not whether an output has been marked active. One way is a small method on the shared Player class that Snapcast overrides, so it can say its own volume only applies to its own stream. That would reach outside the Snapcast provider. If you'd rather keep this PR small, please add a test for volume_control_for_output with the Sendspin output and mention the gap in the description.

music_assistant/providers/snapcast/player.py
```diff
@@ -138,6 +138,18 @@ def group_members(self) -> list[str]:
...
+            # native volume only affects the snapclient's own stream, not whatever
+            # protocol is actually driving the output right now (support#6468)
+            return self._attr_supported_features - {
```
OzGav: If a user has set the volume control to Native in the player settings, that setting is now ignored while Sendspin is playing. That's probably what we want, since the Snapcast volume can't change what you hear in that case. But users will see the change, so please confirm it's on purpose and add a test for it.

tests/providers/snapcast/test_player_volume_routing.py
```diff
@@ -0,0 +1,133 @@
...
+        """volume_control/mute_control redirect to Sendspin once it is active."""
+        controller._players = {
+            snapcast_player.player_id: snapcast_player,
+            sendspin_player.player_id: sendspin_player,
+        }
+        snapcast_player.set_active_output_protocol(sendspin_player.player_id)
+
+        assert snapcast_player.volume_control == sendspin_player.player_id
+        assert snapcast_player.mute_control == sendspin_player.player_id
+        assert snapcast_player.volume_control != PLAYER_CONTROL_NATIVE
```
OzGav: Please add a test that sets Sendspin as active, then switches back to native or clears it, and checks that the volume control goes back to native. That shows the Snapcast volume comes back once the snapclient is playing again.

OzGav (changes requested):

OzGav:
Marking this as draft so we know which need our attention. Mark as ready for review when you want another look.

